PR #4, opened 2020-09-26
IGAPP-3: Fix flow for web

klinzo:


Taggotty (approved):

.circleci/src/jobs/native_deliver_ios.yml
```diff
@@ -26,17 +26,19 @@ steps:
...
-        - 3-yarn-{{ arch }}-{{ checksum "yarn.lock" }}
+        - 3-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
         - 3-yarn-{{ arch }}-
   - run:
       name: Install Dependencies
       command: yarn install --frozen-lockfile
   - save_cache:
       name: Save Yarn Package Cache
-      key: 3-yarn-{{ arch }}-{{ checksum "yarn.lock" }}
+      key: 3-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
```
maxammann, line 36: We can reuse a cache here:

5-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
```suggestion
      key: 5-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
```

.circleci/src/jobs/web_build.yml
```diff
@@ -14,23 +14,25 @@ steps:
   - restore_cache:
       name: Restore Yarn Package Cache
       keys:
-        - 2-yarn-{{ arch }}-{{ checksum "yarn.lock" }}
+        - 2-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
```
maxammann, line 17:
```suggestion
        - 3-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}
```

maxammann (approved):
I think we should use the same key for every cache, regardless of platform. So just use `5-asdfjskdf` for every yarn cache key. That way it is simpler.

PR #8, opened 2020-09-28
IGAPP-3: Adjust workflows

klinzo:


klinzo:
The PR is quite big (once again). Nearly half of the changes are probably from the config.yml, which could be ignored though because its autogenerated. Anyway, I totally understand if it is not possible to review it. Thend I'd just merge it. **Deliveries itself have not been tested yet, will be done in a separate PR.**

.circleci/src/jobs/build_android.yml
```diff
@@ -0,0 +1,114 @@
+parameters:
+  build_for_delivery:
+    description: Whether builds are delivered to the store. If true, integreat and malte builds are created.
+    type: boolean
+docker:
+  - image: circleci/android:api-28-node
```
Taggotty, line 6: didn't we already upgrade to android api 29 ? Probably just a naming issue, but I am wondering

.circleci/src/jobs/build_android.yml
```diff
@@ -0,0 +1,114 @@
...
+        - native/android/vendor/bundle
+  - restore_yarn_cache
+  - run:
+      name: "[FL] Prepare Android Keystore"
+      command: bundle exec fastlane keystore
+      working_directory: native/android
+  - restore_cache:
+      keys:
+        - 1-gradle-{{ checksum "native/android/build.gradle" }}-{{ checksum "native/android/app/build.gradle" }}-{{ checksum "native/android/settings.gradle" }}
+        - 1-gradle-
```
Taggotty, line 51: Is seems like something is missing here. Or is there?

.circleci/src/jobs/deliver_android.yml
```diff
@@ -16,39 +17,36 @@ steps:
   - restore_environment_variables
   - restore_cache:
       keys:
-        - 2-gems-{{ arch }}-{{ checksum "Gemfile.lock" }}
+        - 2-gems-{{ arch }}-{{ checksum "native/Gemfile.lock" }}
         - 2-gems-{{ arch }}-
   - run:
-      command: bundle check || bundle install --path vendor/bundle
+      command: bundle config set path 'vendor/bundle'
```
Taggotty, line 23: this path didn't change?

Taggotty (approved):

.circleci/src/jobs/build_android.yml
klinzo: Good catch! I guess this should be changed, yes.

.circleci/src/jobs/deliver_android.yml
klinzo: No, because of the `working_directory` below.

.circleci/src/jobs/build_android.yml
maxammann: This is fine. Gradle will use an older cache if the gradle files change.

.circleci/src/jobs/build_android.yml
```diff
@@ -0,0 +1,114 @@
...
+      paths:
+        - native/android/vendor/bundle
+  - restore_yarn_cache
+  - run:
+      name: "[FL] Prepare Android Keystore"
+      command: bundle exec fastlane keystore
+      working_directory: native/android
+  - restore_cache:
+      keys:
+        - 1-gradle-{{ checksum "native/android/build.gradle" }}-{{ checksum "native/android/app/build.gradle" }}-{{ checksum "native/android/settings.gradle" }}
```
maxammann, line 50: I'm a bit concerned whether this is alright. We maybe should include some more checksums here:
* gradle version
* checksum of the node_modules/react-native/react-native.gradle file

.circleci/src/workflows/commit.yml
```diff
@@ -1,25 +1,49 @@
...
+        - build_web
 
-  - native_prepare_build
-  - native_build_android:
+  - check_native:
+      context:
+        - slack-app
+
+  - build_android:
+      context:
```
maxammann, line 22: Cool that it is possible to have the context here! Maybe it is a duplication though in other workflows. As long as we only have one set of contexts we should probably keep them in the job files.

native/android/app/buildConfigs.gradle
```diff
@@ -28,11 +28,13 @@ def execCommand(command) {
 }
 
 def determineBuildConfigName() {
-    if (!System.getenv()["BUILD_CONFIG_NAME"]) {
+    if (project.hasProperty('BUILD_CONFIG_NAME')) {
+        return project.BUILD_CONFIG_NAME
```
maxammann, line 32: Why should the project already have this variable?

maxammann: Oh oke because of the properties in the Fastfile. Hm.. I think we should stick to the environment variable.

Else it is unclear what happens if the environment variable and the project properties variable is set to different values.

native/package.json
```diff
@@ -119,7 +119,6 @@
     "jest-react-native": "^18.0.0",
     "jest-transform-stub": "^2.0.0",
     "jetifier": "^1.6.5",
-    "js-yaml": "^3.14.0",
```
maxammann, line 122: I'm not sure but maybe flat can be removed here. Same for web.

maxammann (approved):

.circleci/src/workflows/commit.yml
klinzo: There are no contexts on job level :)

klinzo: https://circleci.com/docs/2.0/contexts/

native/android/app/buildConfigs.gradle
klinzo: I rather find it unclear that everything else is passed as property but the build config name not. :thinking:

.circleci/src/jobs/build_android.yml
klinzo: Wasn't it enough until now? How can I include the gradle version here?

native/package.json
klinzo: Required here because its used in `native/tools/build-config`. Same for `decamelize`. Already not in the web package.json.

In fact, flat it is used in the locales repo already and hoisted, so it could be omitted for now. But I think its much clearer if we just leave it here.

Taggotty (approved):

PR #9, opened 2020-09-30
IGAPP-97: Use webkit on iOS

Schedulaar:

It seems Pois.js and PoisContainer.js had the wrong line-endings. I ended up converting them to LF. Not sure though, what happened exactly :see_no_evil: 

Also this has to be tested on iOS (maybe one needs to do `pod install` as wel...)


native/src/modules/common/components/RemoteContent.js
```diff
@@ -73,17 +69,14 @@ class RemoteContent extends React.Component<PropType, StateType> {
...
     const width = this.state.webViewWidth
     return <StyledView onLayout={this.onLayout}>
       <WebView
-        source={createHtmlSource(renderHtml(content, files, theme, language),
-          URL_PREFIX + getResourceCacheFilesDirPath(cityCode))}
-        allowFileAccess // Needed by android to access file:// urls
+        source={createHtmlSource(renderHtml(content, files, theme, language, resourceCacheUrl), resourceCacheUrl)}
         originWhitelist={['*']} // Needed by iOS to load the initial html
         javaScriptEnabled
-        useWebKit={false}
```
klinzo, line 86: 🎉

native/src/modules/common/renderHtml.js
```diff
@@ -41,28 +43,28 @@ const renderJS = (files: PageResourceCacheStateType) => `
...
   var hrefs = document.querySelectorAll('[href]')
   var srcs = document.querySelectorAll('[src]')
-  var files = ${JSON.stringify(files)}
-  
-  console.debug('Files to inject:') // TODO: remove
-  console.debug(files)
-  
+  var cacheDictionary = ${JSON.stringify(cacheDictionary)}
+
+  console.debug('Resources to inject:') // TODO: remove
```
klinzo, line 48: I think the TODO could be removed

native/src/routes/disclaimer/Disclaimer.js
```diff
@@ -25,7 +25,7 @@ class Disclaimer extends React.Component<PropsType> {
                  navigation={navigation}
                  files={{}}
                  language={language}
-                 cityCode={city}
+                 resourceCacheUrl='http://localhost:8080'
```
klinzo, line 28: Why did you choose to hardcode it here? Shouldn't you rather pass nothing? Just seems weird.

native/src/modules/static-server/containers/StaticServerProvider.js
```diff
@@ -0,0 +1,55 @@
...
+      this.setState({ errorMessage: e.message })
+    }
+  }
+
+  componentWillUnmount () {
+    this.staticServer.stop()
+  }
+
+  render () {
+    if (this.state.errorMessage != null) {
```
klinzo, line 48:
```suggestion
    if (this.state.errorMessage !== null) {
```

klinzo (changes requested):
If I am clicking on a leaf category, safari opens localhost:8080 🤔 

Happens also for events, pois and disclaimer.

Schedulaar:
@klinzo Ahh.. Yeah, fortunately, I know why this happens 🙈 I just forgot to readd a line...
Could you test it on a real iOS device again, please? (Simulator never was a problem)

klinzo:
AFAICT it still opens the browser (localhost) :/

klinzo (approved):

Taggotty (approved):
Looks alright to me, although this is definitely not my field of expertise ^^

native/src/modules/common/components/RemoteContent.js
```diff
@@ -58,9 +58,9 @@ class RemoteContent extends React.Component<PropType, StateType> {
   }
 
   onShouldStartLoadWithRequest = (event: WebViewNavigation) => {
-    const url = event.url
+    const url = event.url.endsWith('/') ? event.url.slice(0, -1) : event.url
```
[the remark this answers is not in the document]
klinzo: resourceCachUrl is `https://localhost:8080`, so if there is a query the check fails anyway, so I guess this should be fine. Right @Schedulaar ?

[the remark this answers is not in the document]
Schedulaar: Yeah, but we should probably only use the normalized url for comparing below.
We should still use `this.props.onLinkPress(event.url)` imo

maxammann (approved):

maxammann (approved):

PR #10, opened 2020-10-05
IGAPP-286: Test and fix deliveries

klinzo:

### DIFF
--- a/.circleci/src/jobs/build_ios.yml
+++ b/.circleci/src/jobs/build_ios.yml
@@ -63,17 +63,9 @@ steps:
             command: bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE} build_config_name:integreat
             working_directory: native/ios
         - run:
-            name: Create sourcemap
-            command: BUILD_CONFIG_NAME='integreat' yarn react-native bundle --platform ios --dev false --entry-file index.js --bundle-output ios/output/integreat.ios-release.bundle --sourcemap-output ios/output/integreat.ios.bundle.map
-            working_directory: native
-        - run:
             name: "[FL] malte build"
             command: bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE} build_config_name:malte
             working_directory: native/ios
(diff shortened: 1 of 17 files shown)


Taggotty (approved):

.circleci/src/workflows/commit.yml
```diff
@@ -1,49 +1,7 @@
 unless: << pipeline.parameters.api_triggered >>
```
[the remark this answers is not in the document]
klinzo: Yep, sorry! Forgot I already marked it as ready for review.

.circleci/src/jobs/build_ios.yml
```diff
@@ -63,17 +63,17 @@ steps:
             command: bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE} build_config_name:integreat
             working_directory: native/ios
         - run:
-            name: Rename sourcemap
-            command: mv index.ios.bundle.map output/integreat.ios.bundle.map
-            working_directory: native/ios
+            name: Create sourcemap
```
maxammann, line 66: I'm not sure whether we want to additinally run `react-native bundle` this takes about 1-2 minutes and slows down the CI. We can also just pass a parameter to the xcode script which then creates a source map.

maxammann (changes requested):

.circleci/src/jobs/build_ios.yml
klinzo: Could you help me with that? How would one do that?

maxammann: In fastlane there is a environment variable EXTRA_PACKAGER_PARAMS or similar. You can just append there " --sourcemap-output ios/output/integreat.ios.bundle.map"

This works for ios and android

klinzo: Ahh found the issue. Afaics, the sourcemap is generated automatically for android. For ios just the path was wrong...

maxammann (approved):
config.yml is outdated. Created a task to prevent this: https://issues.tuerantuer.org/browse/IGAPP-311

PR #12, opened 2020-10-06
IGAPP-254: Setup native locales

klinzo:


klinzo:
Script not yet run. Has to be a xcode preaction. Will do that tomorrow when I have access to a mac again.

Taggotty (approved):
Looks great :)

native/tools/native-locales
```diff
@@ -0,0 +1,41 @@
...
+  const { native: nativeLocales } = JSON.parse(fs.readFileSync('../../locales/locales.json'))
+  const languageCodes = Object.keys(nativeLocales)
+  console.warn('Creating InfoPlist.strings for the languages ', languageCodes)
+
+  languageCodes.forEach(language => {
+    const locales = nativeLocales[language]
+    const keys = Object.keys(locales)
+
+    const content = keys.map((key) => {
+      const regex = /{\{appName}}/gi
```
maxammann, line 16: Why are here double closing brackets? `}}`

native/tools/native-locales
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
```
maxammann, line 1: Having a lot of small scripts can be a bit complicated.
Should we have maybe one script which generates the bulid config and generates the locales?

Something like: `generate-configs` or similar.

native/tools/native-locales
```diff
@@ -0,0 +1,41 @@
...
+
+const fs = require('fs')
+const { program } = require('commander')
+
+const setupNativeLocales = (appName) => {
+  const { native: nativeLocales } = JSON.parse(fs.readFileSync('../../locales/locales.json'))
```
maxammann, line 7: This could be a problem because it is not clear form which current working directory this script should be run.
I think it makes sense to pass the locales.json file as paramter to this script. Then it is technically possible to run it from anywhere.

maxammann (approved):

native/tools/native-locales
klinzo: Because it should replace `{{appName}}`. The escaping is probably wrong though.

native/tools/native-locales
klinzo: What would the benefit be? I.e. why do you think a big script is better than smaller ones?

klinzo: Imo small scripts with only one purpose are better (at least as long as there are only a small amount of scripts).

native/tools/native-locales
klinzo: The escape was redundant, removed.

native/tools/native-locales
maxammann: If you have only one script then this script can be self-documenting. If there is only one script to execute then you get the documentation when you run it.

```console
$ ./integreat-setup build-config generate
$ ./integreat-setup build-config json
$ ./integreat-setup build-config xcconfig
$ ./integreat-setup locales generate-native
$ ./integreat-setup circleci trigger
```

The help documents what each command does. So it is not needed to write extensive documentation.

 The commander cli also does not allow to have a script without a subcommand afaik. This means it is not possible to do something like `./do-something --verboose`. If you use commander then you have to do  `./do-something come-command --verboose`

I think that would make it clearar

klinzo: Okay, sounds reasonable :+1: Would it be okay to do this the next time we need a script? For now, I find it clear enough what our scripts are doing. Also, we don't need to call these scripts manually, they are always called automatically :)

maxammann: Sure, this is probably a separate task: https://issues.tuerantuer.org/browse/IGAPP-310

klinzo: Thanks :+1:

PR #13, opened 2020-10-06
IGAPP-83: Show footer (and header) only once loaded

klinzo:


web/src/modules/layout/containers/LocationLayout.js
```diff
@@ -102,17 +101,14 @@ export class LocationLayout extends React.Component<PropsType, LocalStateType> {
...
     const { city, language } = location.payload
 
     const cityModel = this.getCurrentCity()
 
     if (!cityModel) {
-      return <Layout header={<GeneralHeader viewportSmall={viewportSmall} />}
-                     footer={<GeneralFooter />}>
-        {children}
-      </Layout>
+      return null
```
klinzo, line 111: Not sure whether this is the best solution. Other options:
* Show loading children (LoadingSpinner) as long as the city model is not available -> LoadingSpinner is pushed down once there is a city model
* Show only LoadingSpinner as long as it is loading -> long loading time until something happens

release-notes/unreleased/IGAPP-83-ShowFooterOnceLoaded.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: IGAPP-83
+show_in_stores: false
+platforms:
+  - web
+en: Footer and Header are only shown once everything is ready to prevent penalties by google.
```
Taggotty, line 5: to prevent penalties from google. Does that need to be part of the commit message? Sounds like we are criminals 🙈

Taggotty (approved):

web/src/modules/layout/containers/LocationLayout.js
Taggotty: I also like the first option. But maybe it would also seem a little bit too "jumpy" so i am also not quite sure...

Taggotty: Have you tried out implementing one of the other options?

release-notes/unreleased/IGAPP-83-ShowFooterOnceLoaded.yml
klinzo: Do you have a better suggestion? :D

web/src/modules/layout/containers/LocationLayout.js
klinzo: Yes, tried both. I think any option has its advantages. With the first option I don't like that the goal was to prevent jumping behavior and now we are introducing jumping behavior of the loading spinner. Not sure if that is a problem though. @Schedulaar what do you think?

release-notes/unreleased/IGAPP-83-ShowFooterOnceLoaded.yml
Taggotty: You could just remove that part. Release notes should not have to proclaim why they are there, don't they?

NKlug (approved):
I approve once the discussion about the loading spinner is resolved. I would agree with either option.

maxammann:
Did you verify that the Cumulative Layout Shift aka. CLS metric did improve with these changes?

web/package.json
```diff
@@ -93,6 +94,7 @@
     "@integreat-app/eslint-config-jest": "^0.6.0",
     "@integreat-app/eslint-config-node": "^0.6.0",
     "@integreat-app/eslint-config-react": "^0.6.1",
+    "@testing-library/react": "^11.0.4",
```
maxammann, line 97: In the web project we use enzyme which is able to render components in many different ways. Did you try enzyme first?

web/src/modules/layout/containers/LocationLayout.js
maxammann: I don't really get the issue about the cityModel. It is posisble to render the header without the cityModel. If the cityModel is not yet available then this header is shown:

image

maxammann (changes requested):

web/src/modules/layout/containers/LocationLayout.js
Schedulaar: I cannot really tell what is the best solution. You might try analyzing it with Lighthouse on several pages (e.g. /landkreisgarmischpartenkirchen/de/corona-virus/informationen-zum-corona-virus) on your own. Doing so with the desktop config gives me a 0.254 CLS value. You might want if that improved(=reduced) with your method.

web/package.json
klinzo: Yeah, I did. I think the testing library is a better solution and encourages better test. I didn't found a nice way to test the stuff I wanted to test here, so I decided to introduce the testing library in the web project as well. Imo we should switch entirely in the long run.

web/src/modules/layout/containers/LocationLayout.js
klinzo: @maxammann doesn't really look great imo. Also, if only categories are enabled, the header will collapse later.

klinzo: Tested CLS for mentioned methods. The implemented one (show nothing as long as city model is not available) ranks best with a CLS of 0.039. Option 1 (show loading spinner as long as no city model is available) has a CLS of 0.069. Before the changes CLS was 1. So I guess we should not change anything further.

web/package.json
maxammann: k, should we create a task for that? We can also just use both libraries

maxammann (approved):

web/package.json
klinzo: I think this could go hand in hand with general refactoring and could also be done incrementally? I mean it is not wrong to use enzyme, I just think the testing library encourages better tests. Also, the tests I replaced did not work at all...

klinzo: Do you think we should create a task for this?

maxammann: Probably not worth the effort. I'd say we do this when it starts to bug us. Did you check the enzyme API whether this if really not possible with enzyme?

```
it('should show correct text', () => {
  const wrapper = shallow(<MyComponent />);
  expect(wrapper.text().includes('my text')).toBe(true);
});
```

klinzo: No, did not check this api. Just tried the approach that was used before (searching for components).

PR #26, opened 2020-10-13
IGAPP-293: Setup intelliJ project(s)

maxammann:


native/.circleci
```diff
@@ -0,0 +1 @@
+../.circleci
```
klinzo, line 1: Does this work on windows?

native/android/.circleci
```diff
@@ -0,0 +1 @@
+../.circleci
```
klinzo, line 1: Why do we need this here as well?

klinzo (approved):
I think symlinks to README, version.json, package.json, root/docs and root/tools would be nice as well :)

This way it is not possible anymore to open the root project in intellij, is it?

native/android/.circleci
maxammann: I thought I'd add this here as you can have the integreat-app-android project open and nothing else. I kind of added it for consistency. But I agree that we could remove it.

native/.circleci
maxammann: :shrug: 
Probably not :D

Taggotty: Does not work on windows...

native/.idea/codeStyles/Project.xml
```diff
@@ -13,6 +13,24 @@
...
+          <package name="io.ktor" alias="false" withSubpackages="true" />
+        </value>
+      </option>
+      <option name="PACKAGES_IMPORT_LAYOUT">
+        <value>
+          <package name="" alias="false" withSubpackages="true" />
+          <package name="java" alias="false" withSubpackages="true" />
+          <package name="javax" alias="false" withSubpackages="true" />
+          <package name="kotlin" alias="false" withSubpackages="true" />
+          <package name="" alias="true" withSubpackages="true" />
```
Taggotty, line 30: why are there 2 packages `""` with different configurations?

Taggotty (approved):

native/.circleci
Taggotty: On windows there should also be symlinks, but they seem to work slightly different. Tried using a basic link, but IntelliJ doesn't know what to do with that. I only found that you could use content roots in intellij, but it doesn't really look that fancy, so I would not recommend them.

klinzo:
What is the problem with opening the whole project at once by the way? I like it personally and the performance is reasonable fast (not noticeable slower at least). More overview over the whole project, no need to have multiple windows to compare stuff and no symlinks needed :thinking:

Taggotty:
> What is the problem with opening the whole project at once by the way? I like it personally and the performance is reasonable fast (not noticeable slower at least). More overview over the whole project, no need to have multiple windows to compare stuff and no symlinks needed :thinking: 

Personally my biggest concern is that intellij takes quite some time to Index the Projekts, especially when switch Ing between Branche. Currebtly native and Web are excluded folder So they are not indexed. But if we would add then e. g. as content root that would definitely change.

maxammann:
I think it does not make sense to allow this because of flow, eslint etc.

maxammann:
We would need to merge the versions etc

PR #28, opened 2020-10-14
IGAPP-43: Remove recompose

klinzo:


Taggotty (approved):

native/src/routes/pois/containers/PoisContainer.js
```diff
@@ -133,16 +132,9 @@ const refresh = (refreshProps: RefreshPropsType, dispatch: Dispatch<StoreActionT
...
 
-type RestType = $Diff<PropsType, OwnPropsType>
-const removeOwnProps = (props: PropsType): RestType => {
-  const { t, navigation, ...rest } = props
-  return rest
-}
-
 export default withRouteCleaner<{| navigation: NavigationScreenProp<*> |}>(
-  withTranslation('pois')(
+  withTranslation('error')(
```
maxammann, line 136: Is this translation change correct? probably should be pois

maxammann (approved):

native/src/routes/pois/containers/PoisContainer.js
klinzo: No, actually this is correct. The t function is just used for the SWITCH_LANGUAGE_ACTION to alert if no internet connection. I think that should be refactored though...

PR #29, opened 2020-10-15
IGAPP-316: Tunews language selection web

klinzo:


web/src/modules/app/reducers/index.js
```diff
@@ -43,10 +42,10 @@ const endpointNames = [
...
 ]
 
-export const startFetchReducer = <T: PayloadDataType> (oldPayload?: Payload<T>, action: StartFetchActionType<T>
-): Payload<T> => action.payload
+export const startFetchReducer = (oldPayload?: Payload<*>, action: StartFetchActionType<*>
+): Payload<*> => action.payload
 
-export const finishFetchReducer = <T: PayloadDataType> (oldPayload?: Payload<T>, action: FinishFetchActionType<T>
+export const finishFetchReducer = <T: PayloadDataType> (oldPayload?: Payload<T>, action: FinishFetchActionType<T, *>
 ): Payload<T> => {
```
Taggotty, line 49: Is there a reason why you didn't remove the typing for `T` from the `finishFetchReducer`? It don't quite get why `T` actually matters here...

Taggotty (approved):
Looks great :)

web/src/modules/app/reducers/index.js
klinzo: It is necessary to indicate that the type of oldPayload is the same as the type of the return value. If we would just use `PayloadDataType`, the value of oldPayload could e.g. be `Array<EventModel>` and the return value `<Array<OfferModel>` which are both of type `PayloadDataType`.

Taggotty (approved):

maxammann (approved):

PR #34, opened 2020-10-19
IGAPP-94: Update react-native-image-zoom-viewer

Taggotty:

The issue claimed, that the package uses deprecated features, but those seem to be removed starting version 3.0.0. Convince yourself if you don't believe me though 😸

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -60,7 +60,7 @@
     "react-native-fast-image": "^7.0.2",
     "react-native-gesture-handler": "~1.5.6",
     "react-native-highlight-words": "^1.0.1",
-    "react-native-image-zoom-viewer": "^2.2.27",
+    "react-native-image-zoom-viewer": "^3.0.1",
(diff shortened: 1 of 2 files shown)


klinzo (approved):
Everything working fine? I think in the long run we have to switch nevertheless, already half a year since the last commit on master/release, 131 open issues and 17 open PRs. But I guess that is okay for now.

NKlug (approved):

Taggotty:
> Everything working fine? I think in the long run we have to switch nevertheless, already half a year since the last commit on master/release, 131 open issues and 17 open PRs. But I guess that is okay for now.

The main problem seems to be that there is only one actual maintainer who is currently also looking over several other projects. Still it seems to be maintained, even if it is at a slow pace. On the long run it will definitely make sense to find a new package

maxammann:
Tested on Android

maxammann (approved):

PR #36, opened 2020-10-19
IGAPP-167: Display old data while refreshing

NKlug:

I had to add another action for pushing refreshed content to the store. This might be not very nice, but was the easiest way.
There are some things missing, but I'd like your feedback before I make the remaining changes.
Missing:
- [x] Navigating while refreshing news, events, offers
- [x] Tests
- [x] Maybe some other things I can't remember

~I also added some dev-dependencies for flipper and redux-flipper integration.~


README.md
```diff
@@ -48,8 +48,9 @@ If you are using a different IDE, the steps might differ.
 * Import this project (VCS > Get from Version Control).
 * Run `yarn` in the terminal to install all dependencies.
 
-This is a monorepository. That means it contains more than one project. 
-All in all it contains the following projects:
+This is a monorepository. That means it contains more than one project.
+For correct project dependent IDE settings open each contained project separately in your IDE (e.g. if you are developing web, open the web folder rather than the integreat-app folder).
```
NKlug, line 52: I ran into code formatting trouble until I was told to do open the native app in a separate idea instance. Hence I added this info here.

native/src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -86,5 +86,5 @@ export function * fetchCategory (dataContainer: DataContainer, action: FetchCate
 }
 
 export default function * (dataContainer: DataContainer): Saga<void> {
-  yield takeLatest('FETCH_CATEGORY', fetchCategory, dataContainer)
+  yield takeEvery('FETCH_CATEGORY', fetchCategory, dataContainer)
```
NKlug, line 89: This is necessary for navigating while refreshing. I don't think this has any negative side effects, but am not 100% certain. The alternative would be to add a new action just for refreshing.

native/ios/Integreat.xcodeproj/project.pbxproj
```diff
@@ -486,7 +486,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 49EED0C36D66FA4AB70F7809 /* Pods-Integreat.debug.xcconfig */;
 			buildSettings = {
-				ASSETCATALOG_COMPILER_APPICON_NAME = $BUILD_CONFIG_APP_ICON;
+				ASSETCATALOG_COMPILER_APPICON_NAME = "$BUILD_CONFIG_APP_ICON";
```
NKlug, line 489: Not my changes, don't know why they pop up here...

native/src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -46,6 +46,12 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
     }
   }
 
+  // Check whether another page in the same city is loading, e.g. because it is being refreshed.
+  // This is important for displaying the loading spinner.
+  const otherPageLoading = Object.values(state.categoriesRouteMapping)
+    .filter(route => city === route.city && path !== route.path)
```
klinzo, line 52: Do we need to filter for the language as well? :thinking:

native/src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -46,6 +46,12 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
...
+    .reduce((acc, current) => (acc || current.status === 'loading'), false)
```
klinzo, line 53: Why not use `.some()`?
```suggestion
    .some(route => current.status === 'loading')
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

native/src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -46,6 +46,12 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
     }
   }
 
+  // Check whether another page in the same city is loading, e.g. because it is being refreshed.
```
klinzo, line 49: Maybe (really just maybe) it would be easier to include whether another page is loading in the action as param? But not sure if possible at all or better. This solution is perfectly fine.

native/src/modules/endpoint/reducers/refreshCategory.js
```diff
@@ -0,0 +1,114 @@
...
+const getAllAvailableLanguages = (
+  category: CategoryModel, city: string, language: string, cityLanguages: Array<LanguageModel>
+) => {
+  if (category.isRoot()) {
+    return new Map(cityLanguages.map(model => [model.code, `/${city}/${model.code}`]))
+  }
+  const allAvailableLanguages = new Map(category.availableLanguages)
+  allAvailableLanguages.set(language, category.path)
+  return allAvailableLanguages
+}
```
klinzo, line 18: Is this only used here and in `pushCategory`? Maybe this could be extracted to a separate function. If only used in these to places probably better to keep it the current way.

native/src/modules/endpoint/reducers/refreshCategory.js
```diff
@@ -0,0 +1,114 @@
...
+
+function extractResultModelsAndChildren (root: ?CategoryModel, categoriesMap: CategoriesMapModel, depth: number): {|
```
klinzo, line 20: Maybe add a comment what this does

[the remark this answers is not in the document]
klinzo: Why is root allowed to be null?

native/src/modules/endpoint/reducers/refreshCategory.js
```diff
@@ -0,0 +1,114 @@
...
+    children: resultChildren,
+    depth: depth,
+    allAvailableLanguages: getAllAvailableLanguages(root, city, language, cityLanguages),
+    language,
+    city,
+    status: 'ready'
+  }
+
+  // Update all open routes in the same city with the new content
+  Object.entries(state.categoriesRouteMapping).filter(([key, route]) => city === route.city && path !== route.path)
```
klinzo, line 87: Do we need to check for language as well here?

native/src/modules/endpoint/reducers/refreshCategory.js
```diff
@@ -0,0 +1,114 @@
...
+    .forEach(([key, route]) => {
+      const root: ?CategoryModel = categoriesMap.findCategoryByPath(route.path)
```
klinzo, line 89: Is it possible `root` is not found here? Do we need to handle this somehow?

native/src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -42,8 +43,10 @@ export function * fetchCategory (dataContainer: DataContainer, action: FetchCate
         call(dataContainer.getResourceCache, city, language)
       ])
 
+      const lastUpdate: Moment | null = yield call(dataContainer.getLastUpdate, city, language)
+
       const push: PushCategoryActionType = {
-        type: 'PUSH_CATEGORY',
+        type: loadCriterion.shouldUpdate(lastUpdate) ? 'REFRESH_CATEGORY' : 'PUSH_CATEGORY',
```
klinzo, line 49: Why is this dependent on whether it should update? Isn't this rather dependent on whether it was a user triggered refresh or not?

native/android/app/build.gradle
```diff
@@ -79,7 +79,7 @@ import com.android.build.OutputFile
 
 project.ext.react = [
         extraPackagerArgs: (System.getenv('EXTRA_PACKAGER_ARGS') ?: "").split("\\s+"),
-        enableHermes: false,  // clean and rebuild if changing
+        enableHermes: true,  // clean and rebuild if changing
```
maxammann, line 82: There is a task for that. That means we probably need to check a few settings: https://issues.tuerantuer.org/browse/IGAPP-163

If you just did this for debugging then disable it again before merging.

native/ios/Integreat.xcodeproj/project.pbxproj
maxammann: Very weird, but both is correct. With and without quotes.

native/package.json
```diff
@@ -120,10 +120,12 @@
     "jest-transform-stub": "^2.0.0",
     "jetifier": "^1.6.5",
     "metro-react-native-babel-preset": "^0.59.0",
+    "react-native-flipper": "^0.62.0",
```
maxammann, line 123: What does this dependency do?

native/package.json
```diff
@@ -120,10 +120,12 @@
...
     "react-native-git-upgrade": "^0.2.7",
     "react-native-svg-asset-plugin": "^0.4.0",
     "react-test-renderer": "16.13.1",
     "redux-devtools-extension": "^2.13.8",
```
maxammann, line 127: We should probably remvoe react-devtools-extension when switching integrating flipper.

native/src/modules/app/createReduxStore.js
```diff
@@ -60,7 +62,11 @@ const createReduxStore = (
     resourceCacheUrl: resourceCacheUrlReducer
   })
 
-  const middleware = applyMiddleware(sagaMiddleware)
+  const middlewares = [sagaMiddleware]
+  if (__DEV__) {
```
maxammann, line 66: Use the build config here: `const middleware = buildConfig().development ? flipperReduxMiddleware`

native/.eslintrc.json
```diff
@@ -2,5 +2,8 @@
   "extends": [
     "@integreat-app/eslint-config-node",
     "@integreat-app/eslint-config-jest"
-  ]
+  ],
+  "globals": {
+    "__DEV__": true
```
maxammann, line 7: `__DEV__` should no longer be needed because of the build config

native/src/modules/endpoint/sagas/watchFetchCategory.js
klinzo: There isn't any other possibility to trigger two `FETCH_CATEGORY`s except navigating after refreshing, right?
What happens if a refresh is in progress, user is navigating to another category, the refresh finishes before the push finishes? Isn't there a route with outdated content then? Since this is probably a really rare edge case and not really harmful, I guess this is fine.

native/src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -51,6 +51,10 @@ const createChangeUnavailableLanguage = (city: string, t: TFunction) => (
   dispatch(switchContentLanguage)
 }
 
+function routeHasOldContent (route: CategoryRouteStateType): boolean {
```
klinzo, line 54: Why not using arrow function?

native/src/modules/app/createReduxStore.js
```diff
@@ -43,6 +44,7 @@ const createReduxStore = (
   dataContainer: DataContainer
 ): Store<StateType, StoreActionType> => {
   const sagaMiddleware = createSagaMiddleware()
+  const flipperReduxMiddleware = createDebugger()
```
maxammann, line 47: Call this only in debug mode

native/src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -61,6 +61,10 @@ const createChangeUnavailableLanguage = (city: string, t: TFunction) =>
     })
   }
 
+function routeHasOldContent (route: CategoryRouteStateType): boolean {
```
klinzo, line 64: Why not arrow function?

native/src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: Probably not an issue. Maybe it is possible to call FETCH_CATEGORY twice very fast. But this shouldn't do harm.

native/src/routes/landing/components/CitySelector.js
```diff
@@ -54,6 +54,8 @@ class CitySelector extends React.PureComponent<PropsType> {
     const cities = this.props.cities
     if (normalizedFilter === 'wirschaffendas') {
       return cities.filter(_city => !_city.live)
+    } else if (__DEV__) {
```
klinzo, line 57: When is `__DEV__` set/present?

klinzo (commented):
Maybe it makes sense to do the refreshing of events, news, ... in another PR to keep the PRs small.

Also, could you have a look at the commit history? Looks a bit weird. Do you have the latest changes from main to prevent merge conflicts?

native/src/routes/landing/components/CitySelector.js
maxammann, line 57: Use there the flag from the build config.

maxammann: Should be replaced with build config, closing this comment

klinzo:
I think the PR already looks quite good. I don't think your changes are very hacky. Good work already!

native/.eslintrc.json
maxammann: This will cause an easy to resolve merge conflict with main

native/src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -46,6 +46,12 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
...
+  // This is important for displaying the loading spinner.
```
maxammann, line 50: Why is this important for showing the loading spinner?

native/src/modules/app/StoreActionType.js
```diff
@@ -122,7 +122,7 @@ export type FetchCategoryFailedActionType = {|
   |}
 |}
 export type PushCategoryActionType = {|
-  type: 'PUSH_CATEGORY',
+  type: 'PUSH_CATEGORY' | 'REFRESH_CATEGORY',
```
maxammann, line 125: I think you should add here a separate action. which uses the same params. You can create a type: `CategoryParamsType` and use it in the `PushCategoryActionType` and `RefreshCategoryActionType`

maxammann (commented):
I'm having quite a hard time following what is going on. Just a quick question for a way of handling this without complex logic:

Would it be possible not to update the state until the new content is ready? That means the state is not altered until the new data arrived. As soon as it arrives the content in the state is overwritten.

native/src/modules/endpoint/reducers/pushCategory.js
klinzo: Because otherwise no loading spinner is shown as the current route would be 'ready'

klinzo:
> I'm having quite a hard time following what is going on. Just a quick question for a way of handling this without complex logic:
> 
> Would it be possible not to update the state until the new content is ready? That means the state is not altered until the new data arrived. As soon as it arrives the content in the state is overwritten.

Afaics this is nearly the case. The only change on refreshing is changing the route status to 'loading', right? If this is not done, the user does not get any feedback that the content is actually refreshing.

klinzo (commented):
Is it really necessary to have a separate `refreshCategory` action/reducer? Kind of looks like it could be merged :thinking:

native/src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: This could be an issue if there are two resource cache updates. Then the app shows an error.
So probably 2 actions are safer.

klinzo: Setting the last update of the data container to the current time before something is actually loaded from the api in `loadCityContent` should resolve all issues. If it fails, it has to be set to the previous timestamp though probably.

native/package.json
NKlug: It is required by the redux-flipper plugin which allows you to see actions and state in flipper

native/package.json
NKlug: Should I remove it or will this be done in a separate task?

native/src/modules/endpoint/hocs/withPayloadProvider.js
```diff
@@ -95,7 +95,12 @@ const withPayloadProvider = <S: { dispatch: Dispatch<StoreActionType> }, R: {}>
                                                 changeLanguage={this.changeUnavailableLanguage} />
         } else if (props.status === 'loading') {
           return this.state.timeoutExpired
-            ? <ScrollView refreshControl={<RefreshControl refreshing />} contentContainerStyle={{ flexGrow: 0 }} />
+            ? <ScrollView refreshControl={<RefreshControl refreshing />} contentContainerStyle={{ flexGrow: 1 }}
+                        keyboardShouldPersistTaps='always'>
```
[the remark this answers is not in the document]
NKlug: No idea, I just copied it from the `else` block a few lines below and thought it must be there for a reason. I think it configures keyboard visibility after tapping the Scrollview or something like that.

native/src/modules/endpoint/reducers/pushCategory.js
NKlug: Didn't know this existed :smile:

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: Depends on what behavior we are looking for. This way it would update all open screens if there is new content available, so especially also when the update has been performed automatically, e.g. because of the 24h rule.
If we only check for `forceRefresh === true` this wouldn't be the case and it would only update when you refresh manually.
What do you think?

native/src/routes/landing/components/CitySelector.js
NKlug: It proved useful to me to display all cities when developing. If you want I can also remove this again.

native/src/modules/endpoint/hocs/withPayloadProvider.js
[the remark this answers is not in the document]
klinzo: I think this is probably only relevant for the search and feedback as this is the only time a keyboard is shown. Please verify this is the behavior we want

native/src/modules/endpoint/sagas/watchFetchCategory.js
klinzo: Okay, seems legit. The parameter `forceRefresh` is also taken into account, is it?

native/package.json
klinzo: If the changes with flipper are done in this issue, it should be removed. If you'll move the flipper stuff to a separate task, it should be done there I'd say.

native/src/routes/landing/components/CitySelector.js
klinzo: I like it, leave it :)

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: Yep, `shouldUpdate(...)` reads: `return this._forceUpdate || !lastUpdate ||
      lastUpdate.isBefore(moment.utc().subtract(MAX_CONTENT_AGE, 'hours'))`

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: I'll try @klinzo's solution.
@maxammann How would a separate action resolve the problem you mentioned? The new action would pretty much call the same function as FETCH_CATEGORY.

native/android/app/src/main/java/tuerantuer/app/integreat/MainApplication.java
```diff
@@ -60,7 +60,7 @@ private static void initializeFlipper(
          We use reflection here to pick up the class that initializes Flipper,
         since Flipper library is not available in release mode
         */
-                Class<?> aClass = Class.forName("com.rndiffapp.ReactNativeFlipper");
+                Class<?> aClass = Class.forName("tuerantuer.app.integreat.ReactNativeFlipper");
```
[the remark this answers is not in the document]
NKlug: Thanks :)

native/package.json
NKlug: I'm not sure what changes exactly have to be made for the flipper integration. At the current state of my PR everything works for me, but I don't know if there are any things missing.

native/src/modules/endpoint/hocs/withPayloadProvider.js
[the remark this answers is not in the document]
NKlug: I changed the parameter to different values and it makes no difference for me. Yet I am hesitant to remove it. It was introduced by @TechPiranja when fixing NATIVE-216.

native/src/modules/endpoint/reducers/pushCategory.js
NKlug: Can't hurt I guess. Routes are probably cleared anyways when switching language.

native/src/modules/endpoint/reducers/refreshCategory.js
NKlug: As far as I can tell it's only used in those two places.

native/src/modules/endpoint/reducers/refreshCategory.js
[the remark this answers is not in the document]
NKlug: I copied this code from the very similar `pushCategory`. There, `root` is also allowed to be null. I don't know when exactly this might happen, but just left it this way.

native/src/modules/endpoint/reducers/refreshCategory.js
NKlug: Added it.

native/src/modules/endpoint/reducers/refreshCategory.js
NKlug: Yep I think so. Added handling.

native/src/modules/endpoint/hocs/withPayloadProvider.js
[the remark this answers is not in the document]
klinzo: Then its fine. If we recognize faulty behavior, we can fix it later.

native/src/modules/endpoint/reducers/pushCategory.js
klinzo: I was rather thinking about peeking and linking to other cities, but should not hot hurt anyway I guess.

native/package.json
klinzo: Then its probably the best to do this in a separate task. We probably have to check if enabling hermes breaks something as well.

NKlug: Alright. Then I'll also disable hermes before merging. Leaving the redux-flipper and react-native-flipper packages shouldn't harm anything though.

native/android/app/build.gradle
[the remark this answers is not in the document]
maxammann: After enabling the app needs to be tested throughout. Compiling in release debug, native extensions etc.

native/package.json
maxammann: In that case create a task for flipper and mention the changes which have been done already.

native/src/routes/landing/components/CitySelector.js
maxammann: Can stay. The flag from the build config should be used.

native/src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: I think no new action is needed. But there are two reducers atm which contain a lot of duplicate code. They should be merged.

README.md
maxammann: I dont think so.

native/package.json
klinzo: There is already a task: https://issues.tuerantuer.org/browse/IGAPP-334

native/.idea/codeStyles/Project.xml
```diff
@@ -13,6 +13,24 @@
       <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
       <option name="SPACES_WITHIN_IMPORTS" value="true" />
     </JSCodeStyleSettings>
+    <JetCodeStyleSettings>
```
maxammann, line 16: This shouldn't be added here as there is no code in the integreat-app-native intellij project, only in the integreat-app-android project.

maxammann:
I kind of lost track about the changes here, should we move discussions to slack and create a new PR from this which only contains the minimal necessary changes? (no flipper, hermes etc)

native/.idea/codeStyles/Project.xml
NKlug: Yeah, you're right. Such unintentional changes are really annoying, I don't even know what caused them...

native/android/app/build.gradle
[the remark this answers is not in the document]
NKlug: Disabled again.

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: I merged the respective files.

native/src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -54,74 +54,88 @@ export default function * loadCityContent (
...
 
-  return true
+    return true
+  } catch (e) {
+    // If any error occurs, we have to store the old value for lastUpdate again
+    if (lastUpdate) {
+      yield call(dataContainer.setLastUpdate, newCity, newLanguage, lastUpdate)
+    }
+    throw e
+  }
```
klinzo, line 140: This should be the same as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch:
```suggestion
  } finally (e) {
    // If any error occurs, we have to store the old value for lastUpdate again
    if (lastUpdate) {
      yield call(dataContainer.setLastUpdate, newCity, newLanguage, lastUpdate)
    }
  }
```

klinzo (changes requested):
You probably tried to fix some of these flow issues already right? :/

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+        value: newResourceCache
+      },
+      searchRoute: { categoriesMap }
+    }
+  }
+
+  // Check whether another page in the same city is loading, e.g. because it is being refreshed.
+  // This is important for displaying the loading spinner.
+  const otherPageLoading = Object.values(state.categoriesRouteMapping)
+    // $FlowFixMe Flow does not support Object.values
```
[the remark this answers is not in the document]
klinzo: Fixable by using `Object.keys` instead and `state.categoriesRouteMapping` instead. Imo better than have this many FlowFixMes. Same for `Object.entries` and also other files. Bad by flow though :confounded:

Some flow errors hidden by these fix mes (e.g. path not available for all route statuses and state.categoriesRouteMapping not being writable.

native/src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -9,22 +9,38 @@ import { expectSaga, testSaga } from 'redux-saga-test-plan'
 import loadCityContent from '../loadCityContent'
 import CategoriesMapModelBuilder from '../../../../testing/builder/CategoriesMapModelBuilder'
 import ErrorCodes from '../../../error/ErrorCodes'
+import moment from 'moment'
 
 jest.mock('rn-fetch-blob')
 jest.mock('../loadCityContent')
+// $FlowFixMe Date.now is writable here
+Date.now = jest.fn().mockReturnValue(new Date('2020-01-01T12:00:00.000Z'))
```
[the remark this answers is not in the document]
maxammann: You should revert this afterwards. I think jest is able to leak this mock value to other tests.

```suggestion
const prev = Date.now
Date.now = jest.fn().mockReturnValue(new Date('2020-01-01T12:00:00.000Z'))
Date.now = prev
```

native/src/modules/app/StoreActionType.js
```diff
@@ -121,15 +121,23 @@ export type FetchCategoryFailedActionType = {|
...
+|}
+
+export type PushCategoryActionType = {|
+  type: 'PUSH_CATEGORY',
+  +params: CategoryParamsType
+|}
+
+export type RefreshCategoryActionType = {|
+  type: 'REFRESH_CATEGORY',
+  +params: CategoryParamsType
```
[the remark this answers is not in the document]
maxammann: I would prefere to have a boolean within `PUSH_CATEGORY`

```javascript
export type PushCategoryActionType = {|
  type: 'PUSH_CATEGORY',
  +params: CategoryParamsType
  refresh: boolean
|}
```

native/src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -78,12 +95,14 @@ export default (
...
+      case 'REFRESH_CATEGORY':
       case 'PUSH_CATEGORY':
-        return pushCategory(state, action)
+        return pushOrRefreshCategory(state, action)
       case 'PUSH_POI':
         return pushPoi(state, action)
+      case 'REFRESH_EVENT':
       case 'PUSH_EVENT':
-        return pushEvent(state, action)
+        return pushOrRefreshEvent(state, action)
```
maxammann, line 105: If you have a boolean in the PUSH_* actions then the name would stay: `pushCategory`

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+  // $FlowFixMe Flow can't evaluate the status as it is dynamic
+  const route: CategoryRouteStateType = {
+    path: root.path,
+    models: resultModels,
+    children: resultChildren,
+    depth: depth,
+    allAvailableLanguages: getAllAvailableLanguages(root, city, language, cityLanguages),
+    language,
+    city,
+    status: (otherPageLoading && type === 'PUSH_CATEGORY') ? 'loading' : 'ready'
```
maxammann, line 98: If the `PUSH_CATEGORY` action would have a boolean `refresh: boolean` then this would be simpler.

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  if (type === 'REFRESH_CATEGORY') {
+    // Update all open routes in the same city with the new content in case the content has been refreshed
+    Object.entries(state.categoriesRouteMapping)
+      // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 104: The next line does not contain an Object.entries

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+      .filter(([key, route]) => city === route.city && path !== route.path && language === route.language)
+      .forEach(([key: string, route]) => {
+        // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 107: The next line does not contain an Object.entries

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+        const root: ?CategoryModel = categoriesMap.findCategoryByPath(route.path)
+        if (!root) {
+          // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 110: The next line does not contain an Object.entries

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+            depth: depth,
+            language,
+            city,
+            status: 'error',
+            message: `Could not find a category with path '${path}'.`,
+            code: ErrorCodes.PageNotFound
+          }
+        } else {
+          const { resultModels, resultChildren } = extractResultModelsAndChildren(root, categoriesMap, depth)
+          // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 122: The next line does not contain an Object.entries

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+          state.categoriesRouteMapping[key] = {
+            // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 124: The next line does not contain an Object.entries

If there are that many flowfixmes it makes sense to refactor this or implement the function yourself.

maxammann: You can also extract it to a separate function and handle the flowfixme only once.

native/src/modules/endpoint/reducers/pushOrRefreshEvent.js
```diff
@@ -0,0 +1,91 @@
...
+      city
+    }
+  }
+
+  if (type === 'REFRESH_EVENT') {
+    Object.entries(state.eventsRouteMapping)
+      // $FlowFixMe Object.entries does not supply proper types
+      .filter(([key, route]) => city === route.city && path !== route.path && language === route.language)
+      .forEach(([key:string, route]) => {
+        // $FlowFixMe Object.entries does not supply proper types
```
maxammann, line 73: Same as above about the FlowFixMes

native/src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -9,22 +9,38 @@ import { expectSaga, testSaga } from 'redux-saga-test-plan'
...
+      languages: { [language]: { lastUpdate: moment('2020-01-01T00:00:00.000Z') } },
+      lastUsage: moment('2020-01-01T01:00:00.000Z')
+    }
+  }
 
   const dataContainer = new DefaultDataContainer()
   await dataContainer.setCategoriesMap(city, language, categories)
   await dataContainer.setLanguages(city, languages)
   await dataContainer.setResourceCache(city, language, resources)
+  await dataContainer._databaseConnector._storeMetaCities(metaCities)
```
maxammann, line 35: Tests should not access internals. Is there a way to test this differently or remove the _ in the function names?

native/src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -49,7 +101,7 @@ describe('watchFetchCategories', () => {
           key: 'categories-key',
           criterion: {
             forceUpdate: false,
-            shouldRefreshResources: true
+            shouldRefreshResources: false
```
maxammann, line 104: Why did this behavior change?

native/src/modules/endpoint/sagas/__tests__/watchFetchEvent.spec.js
```diff
@@ -9,9 +9,12 @@ import watchFetchEvent, { fetchEvent } from '../watchFetchEvent'
 import { expectSaga, testSaga } from 'redux-saga-test-plan'
 import loadCityContent from '../loadCityContent'
 import ErrorCodes from '../../../error/ErrorCodes'
+import moment from 'moment'
 
 jest.mock('rn-fetch-blob')
 jest.mock('../loadCityContent')
+// $FlowFixMe Date.now is writable here
+Date.now = jest.fn().mockReturnValue(new Date('2020-01-01T12:00:00.000Z'))
```
maxammann, line 17: Revert this mock after running this test.

native/src/modules/endpoint/sagas/__tests__/watchFetchEvent.spec.js
```diff
@@ -47,7 +91,7 @@ describe('watchFetchEvents', () => {
           key: 'events-key',
           criterion: {
             forceUpdate: false,
-            shouldRefreshResources: true
+            shouldRefreshResources: false
```
maxammann, line 94: Why did this change?

native/src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -42,8 +43,11 @@ export function * fetchCategory (dataContainer: DataContainer, action: FetchCate
         call(dataContainer.getResourceCache, city, language)
       ])
 
-      const push: PushCategoryActionType = {
-        type: 'PUSH_CATEGORY',
+      const lastUpdate: Moment | null = yield call(dataContainer.getLastUpdate, city, language)
+
+      // $FlowFixMe Flow can't evaluate the type as it is dynamic
```
maxammann, line 48: That actually should work. I think it is possible to refactor this that it works. You can create a function with return type `PushCategoryActionType | RefreshCategoryActionType`. Inside that function you have a if which returns one action or the other one.

native/src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -26,8 +32,11 @@ export function * fetchEvent (dataContainer: DataContainer, action: FetchEventAc
         call(dataContainer.getResourceCache, city, language)
       ])
 
-      const insert: PushEventActionType = {
-        type: 'PUSH_EVENT',
+      const lastUpdate: Moment | null = yield call(dataContainer.getLastUpdate, city, language)
+
+      // $FlowFixMe Flow can't evaluate the type as it is dynamic
```
maxammann, line 37: That actually should work. I think it is possible to refactor this that it works. You can create a function with return type PushCategoryActionType | RefreshCategoryActionType. Inside that function you have a if which returns one action or the other one.

native/src/routes/events/containers/EventsContainer.js
```diff
@@ -98,14 +100,15 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
 
   if (state.resourceCacheUrl === null || state.cities.status === 'loading' || switchingLanguage ||
-    route.status === 'loading' || languages.status === 'loading') {
+    (route.status === 'loading' && !routeHasOldContent(route)) || languages.status === 'loading') {
     return { status: 'loading' }
   }
 
   const cities = state.cities.models
 
+  // $FlowFixMe Flow can't evaluate the status as it is dynamic
```
maxammann, line 109: This is very weird...

native/src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -109,14 +112,15 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+    (route.status === 'loading' && !routeHasOldContent(route)) || languages.status === 'loading') {
     return { status: 'loading' }
   }
 
   const cities = state.cities.models
   const stateView = new CategoriesRouteStateView(route.path, route.models, route.children)
+  // $FlowFixMe Flow can't evaluate the status as it is dynamic
   return {
-    status: 'success',
+    status: route.status === 'loading' ? 'loading' : 'success',
```
maxammann, line 123: This is very weird...

maxammann (changes requested):

native/src/modules/app/StateType.js
```diff
@@ -97,7 +100,9 @@ export type EventRouteStateType = {|
   +allAvailableLanguages: $ReadOnlyMap<string, ?string>
 |} | {|
   +status: 'loading',
-  ...EventRouteConfigType
+  ...EventRouteConfigType,
```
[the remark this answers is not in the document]
NKlug: There is a task for adding Pull-to-Refresh for News. I created a task for POIs and Offers, although Offers have a low prio IMO.

native/src/modules/app/StoreActionType.js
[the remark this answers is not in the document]
NKlug: Yes, at least for updating the state with the new data. Not for refreshing itself.

native/src/modules/endpoint/hocs/withPayloadProvider.js
```diff
@@ -95,7 +99,12 @@ const withPayloadProvider = <S: { dispatch: Dispatch<StoreActionType> }, R: {}>
...
+              {/* only display content while loading if innerProps and dispatch are available */}
+              {props.innerProps && props.dispatch
```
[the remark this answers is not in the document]
NKlug: Don't know, probably not. Added it just to be sure.

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
```diff
@@ -0,0 +1,149 @@
...
+  // This is important for displaying the loading spinner.
+  const otherPageLoading = Object.values(state.categoriesRouteMapping)
+    // $FlowFixMe Flow does not support Object.values
+    .filter(route => city === route.city && path !== route.path && language === route.language)
+    // $FlowFixMe Flow does not support Object.values
+    .some(route => route.status === 'loading')
+
+  const { resultModels, resultChildren } = extractResultModelsAndChildren(root, categoriesMap, depth)
+
+  // $FlowFixMe Flow can't evaluate the status as it is dynamic
```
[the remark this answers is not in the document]
NKlug: `type` can have two values: `loading` and `ready` (and some more). As they are hard coded in the StateType.js, flow considers them "string literals". But the type of a line like e.g. `const x = condition ? 'loading' : 'ready'` is `string` instead of `string literal`, so flow complains.

There is probably a solution to this, like:
```
if (a) {
x = { type: 'loading', ...}
} else {
x = { type: 'ready', ...} 
}
```
This is more verbose than the inline solution and I don't like having to write such "unclean" code just because of flow. But can do it anyways if you want...

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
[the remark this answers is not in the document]
NKlug: In case of `REFRESH_CATEGORY` there is only `ready`, as all open routes are updated with the new content and then considered 'ready'. Meaning, in this case it's irrelevant if another page is loading or not.

native/src/modules/endpoint/reducers/pushOrRefreshEvent.js
```diff
@@ -0,0 +1,91 @@
...
+
+const pushOrRefreshEvent = (state: CityContentStateType,
+  action: PushEventActionType | RefreshEventActionType): CityContentStateType => {
+  const type = action.type
+  const { events, path, key, language, resourceCache, cityLanguages, city } = action.params
+
+  if (!key) {
+    throw new Error('You need to specify a key!')
+  }
+
```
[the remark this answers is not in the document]
NKlug: I'll try. This is a relic of the old version of this function :)

native/src/modules/endpoint/reducers/pushOrRefreshEvent.js
```diff
@@ -0,0 +1,91 @@
...
+      allAvailableLanguages,
+      language,
+      city
+    }
+  }
+
+  if (type === 'REFRESH_EVENT') {
+    Object.entries(state.eventsRouteMapping)
+      // $FlowFixMe Object.entries does not supply proper types
+      .filter(([key, route]) => city === route.city && path !== route.path && language === route.language)
```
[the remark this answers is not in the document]
NKlug: The route with `path === route.path` is treated separately in the code above/below. There is always one route for which this is the case, concretely the route in which the pull-to-refresh action was performed. So I think I don't quite understand your second question here.

native/src/modules/endpoint/sagas/loadCityContent.js
NKlug: Forgot this existed :smile:

native/src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -54,74 +54,88 @@ export default function * loadCityContent (
...
-  const cities = yield call(loadCities, dataContainer, shouldUpdate) // Refresh for flags like eventsEnabled necessary
-  const cityModel = cities.find(city => city.code === newCity)
-  if (!cityModel) {
-    throw new Error(`City '${newCity}' was not found.`)
-  }
+  // Temporarily set lastUpdate to "now" to hinder other threads from trying to update content and
+  // resources at the same time. This kind of serves as a lock.
+  yield call(dataContainer.setLastUpdate, newCity, newLanguage, moment())
+
+  try {
```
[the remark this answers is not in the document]
NKlug: Yes.

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
NKlug: Still doesn't have proper types because of the previous Object.entries. I'll try to circumvent this by @klinzo's suggestion.

native/src/modules/endpoint/sagas/__tests__/watchFetchEvent.spec.js
NKlug: We now have one test which refreshes content and one that doesn't. This is the one that doesn't.

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: Yeah I had that idea, too. Didn't like it though, as it unnecessarily makes the code more verbose and I also don't like having to adjust my code just because of flow's weaknesses :/
I also mentioned this in some comment above.

native/src/modules/app/StoreActionType.js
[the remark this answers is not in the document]
NKlug: I'll try to switch to @maxammann's suggestion. This probably should do for now.

native/src/modules/app/StateType.js
```diff
@@ -97,7 +100,9 @@ export type EventRouteStateType = {|
...
+  +models?: $ReadOnlyArray<EventModel>,
+  +allAvailableLanguages?: $ReadOnlyMap<string, ?string> // including the current content language
```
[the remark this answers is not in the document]
NKlug: Same comment is a few lines above, I just copy-pasted... I I removed it now.

native/src/modules/endpoint/reducers/pushOrRefreshEvent.js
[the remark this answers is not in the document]
klinzo: Yeah right

native/src/modules/endpoint/sagas/loadCityContent.js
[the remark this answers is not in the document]
NKlug: Actually, despite not remembering that the `finally` block existed, I was right to put it in the `catch` block, because we only want so reset `lastUpdate` to the old date if an error occurred. `finally` is always executed.

native/src/modules/endpoint/hocs/withPayloadProvider.js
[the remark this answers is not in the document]
klinzo: I think you can remove it :)

native/src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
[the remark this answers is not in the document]
NKlug: Changed mocking the date.

native/src/modules/endpoint/sagas/watchFetchCategory.js
NKlug: Irrelevant now as I unified the actions.

native/src/modules/endpoint/reducers/pushOrRefreshEvent.js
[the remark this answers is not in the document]
NKlug: Still works after removing. At least for now :sweat_smile:

maxammann (approved):
Apart from flowfixmes this is fine for me! Good work!

native/src/modules/endpoint/reducers/pushOrRefreshCategory.js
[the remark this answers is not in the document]
maxammann: I fixed this in my commit

maxammann:
I resolved all my conversations, please take a look at the latest commit from me.

native/src/modules/app/StoreActionType.js
```diff
@@ -199,20 +199,20 @@ export type ClearEventActionType = {|
   type: 'CLEAR_EVENT', +params: {| +key: string |}
 |}
 
-type EventParamsType = {|
```
klinzo: Could be removed again

native/src/utils/object.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+export const values = <T> (object: $ReadOnly<{ [string]: T }>): Array<T> => {
+  // $FlowFixMe
+  return Object.values(object)
```
klinzo, line 5: Would also be possible to use the following in order not to have $FlowFixMes, similar for entries below:
```
  return Object.keys(object).map(key => object[key])
```

klinzo (approved):
Looking good, thanks @NKlug and @maxammann :tada: Honestly not really reviewed carefully again, I think we just have to test our next development release carefully :)

PR #43, opened 2020-10-21
IGAPP-253: Malte CD

klinzo:

Will test whether this works tomorrow after the scheduled development release.


native/fastlane/Fastfile
```diff
@@ -108,30 +108,33 @@ platform :android do
...
     build_config_name = options[:build_config_name]
     apk_path = options[:apk_path]
 
     build_config = read_build_config(build_config_name)
-
-    ensure_env_vars(
-        env_vars: ["GOOGLE_SERVICE_ACCOUNT_JSON"]
-    )
-
+    skip_metadata = build_config_name == "malte"
```
maxammann, line 121: I'd favor to have this check in the circleci files. The fastline files do not need to know about which build configs are available.

maxammann (approved):

native/fastlane/Fastfile
klinzo: This is not about which build configs are available but rather for which build configs metadata is existing. Metadata is only handled in fastlane, so I think this is the clearer way. Other opinions?

.circleci/src/workflows/triggered_production_delivery.yml
```diff
@@ -114,7 +114,7 @@ jobs:
       production_delivery: true
       requires:
         - deliver_integreat_android
-  #        - deliver_malte_android
+          - deliver_malte_android
```
Taggotty, line 117: fix indent

native/fastlane/Fastfile
Taggotty: 🙋🏼‍♀️ why do we skip the metadata if the build config is malte?

Taggotty (approved):

native/fastlane/Fastfile
klinzo: If we upload all metadata on every delivery, it is not possible to change metadata in the playstore console or in appstore connect, so we would have to change it in our repo. Since we don't want to do that for the Maltesers and allow them to do it themselves, we don't upload any metadata.

Taggotty: Thanks for explaining. I agree the check should remain in the fastline files. It does seem like a realtively elegant solution to me.

.circleci/src/workflows/triggered_production_delivery.yml
klinzo: Good catch!

Taggotty (approved):

PR #45, opened 2020-10-22
IGAPP-210: Fix endless loading because of subscribing

klinzo:

### DIFF
--- a/native/src/modules/endpoint/sagas/loadCityContent.js
+++ b/native/src/modules/endpoint/sagas/loadCityContent.js
@@ -1,7 +1,7 @@
 // @flow
 
 import type { Saga } from 'redux-saga'
-import { all, call, put } from 'redux-saga/effects'
+import { all, call, put, spawn } from 'redux-saga/effects'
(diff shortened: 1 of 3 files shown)


klinzo (commented):
I am not entirely sure if this is good style, but did not find another solution. I tested this quite a bit and everything worked as it should. If you go to a city and leave it afterwords without internet, subscribing and unsubscribing is executed once a internet connection is established again. The order has been correct all the time for me.

Znippet (approved):

maxammann (approved):
Are there any timeout in firebase? I think this solves the issue, but this also means that a lot of pending promises can exist. Maybe we can check additionally whether there is internet conectivity

klinzo:
> Are there any timeout in firebase? I think this solves the issue, but this also means that a lot of pending promises can exist. Maybe we can check additionally whether there is internet conectivity
@maxammann I did not find a way to set a timeout in firebase... This was therefore the best solution I could come up with. I think this is okay since no User should be changing cities frequently. We could think about improving this at a later time though.

PR #47, opened 2020-10-26
IGAPP-335: City search crashing for query `*`

klinzo:


Taggotty (approved):
This bug was my bad. Nice catch there :)

maxammann (approved):
What does autoEscape do? Maybe add a comment why it is important. Does it use regex by default?

klinzo:
Yes, per default it is interpreted as regex. AutoEscape treets it as normal text. I think a short google search is enough to find it out, so I don't think a comment is necessary. We don't have to implement external api :)

PR #48, opened 2020-10-26
IGAPP-256: Improve remote content in category list

Taggotty:

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -64,7 +64,7 @@
     "react-native-pdf": "6.2.0",
     "react-native-permissions": "2.0.9",
     "react-native-progress": "^4.0.3",
-    "react-native-render-html": "^4.2.0",
+    "react-native-render-html": "^5.0.0-alpha.1",
     "react-native-safe-area-context": "^1.0.0",
     "react-native-safe-area-view": "2.0.0",
     "react-native-screens": "^2.11.0",
--- a/native/package.json
+++ b/native/package.json
@@ -64,7 +64,7 @@
     "react-native-pdf": "6.2.0",
     "react-native-permissions": "2.0.9",
     "react-native-progress": "^4.0.3",
-    "react-native-render-html": "^5.0.0-alpha.1",
+    "react-native-render-html": "^5.0.0-alpha.2",
     "react-native-safe-area-context": "^1.0.0",
     "react-native-safe-area-view": "2.0.0",
     "react-native-screens": "^2.11.0",
--- a/native/src/modules/categories/components/Categories.js
+++ b/native/src/modules/categories/components/Categories.js
@@ -5,7 +5,7 @@ import { View } from 'react-native'
 
 import Page from '../../common/components/Page'
 import Tiles from '../../common/components/Tiles'
-import type { CategoryListModelType } from './CategoryList'
+import type { CategoryListModelType, ListContentModelType } from './CategoryList'
 import CategoryList from './CategoryList'
 import TileModel from '../../common/models/TileModel'
 import {
@@ -132,6 +132,17 @@ class Categories extends React.Component<PropsType> {
     return categories.map(category => this.getListModel(category))
   }
 
+  getListContentModel (category: CategoryModel): ?ListContentModelType {
+    const { navigation, navigateToInternalLink, resourceCacheUrl } = this.props
+    return category.content ? {
+      content: category.content,
+      files: this.getCategoryResourceCache(category),
+      navigation: navigation,
+      navigateToInternalLink: navigateToInternalLink,
+      resourceCacheUrl: resourceCacheUrl
+    } : undefined
+  }
+
   /**
    * Returns the content to be displayed, based on the current category, which is
    * a) page with information
@@ -186,7 +197,7 @@ class Categories extends React.Component<PropsType> {
           })}
           thumbnail={this.getCachedThumbnail(category) || category.thumbnail}
           title={category.title}
-          content={category.content}
+          listContent={this.getListContentModel(category)}
           language={language}
           onItemPress={this.onItemPress}
           theme={theme} />
--- a/native/src/modules/categories/components/CategoryList.js
+++ b/native/src/modules/categories/components/CategoryList.js
@@ -37,7 +37,7 @@ export type ListContentModelType = {|
 type PropsType = {|
   categories: Array<ListEntryType>,
   title?: string,
-  listContent: ?ListContentModelType,
+  listContent?: ?ListContentModelType,
   /** A search query to highlight in the categories titles */
   query?: string,
   theme: ThemeType,
--- a/native/src/modules/categories/components/CategoryList.js
+++ b/native/src/modules/categories/components/CategoryList.js
@@ -66,12 +66,12 @@ class CategoryList extends React.Component<PropsType> {
         : file.filePath
     })
     return <CategoryListContent content={listContent.content}
-                         language={language}
-                         navigation={listContent.navigation}
-                         navigateToInternalLink={listContent.navigateToInternalLink}
-                         resourceCacheUrl={listContent.resourceCacheUrl}
-                         cacheDictionary={cacheDictionary}
-                         theme={theme} />
+                                language={language}
+                                navigation={listContent.navigation}
+                                navigateToInternalLink={listContent.navigateToInternalLink}
+                                resourceCacheUrl={listContent.resourceCacheUrl}
+                                cacheDictionary={cacheDictionary}
+                                theme={theme} />
   }
 
   render () {
--- a/native/src/modules/categories/components/CategoryList.js
+++ b/native/src/modules/categories/components/CategoryList.js
@@ -80,7 +80,7 @@ class CategoryList extends React.Component<PropsType> {
     return <>
       {thumbnail && <CategoryThumbnail source={thumbnail} />}
       {title && <CategoryListCaption title={title} theme={theme} withThumbnail={!!(thumbnail)} />}
-      {!!listContent && this.getListContent(listContent)}
+      {listContent && this.getListContent(listContent)}
       {categories.map(({ model, subCategories }) =>
         <CategoryListItem key={model.path}
                           category={model}
--- a/native/src/modules/categories/components/CategoryList.js
+++ b/native/src/modules/categories/components/CategoryList.js
@@ -2,12 +2,17 @@
 
 import * as React from 'react'
 
+import type { NavigationStackProp } from 'react-navigation-stack'
 import CategoryListItem from './CategoryListItem'
-import Html from 'react-native-render-html'
 import type { ThemeType } from '../../theme/constants'
 import styled from 'styled-components/native'
 import Image from '../../common/components/Image'
 import CategoryListCaption from '../../../modules/common/components/CategoryListCaption'
+import CategoryListContent from './CategoryListContent'
+import type { NavigateToInternalLinkParamsType } from '../../app/createNavigateToInternalLink'
+import type { PageResourceCacheEntryStateType, PageResourceCacheStateType } from '../../app/StateType'
+import { RESOURCE_CACHE_DIR_PATH } from '../../endpoint/DatabaseConnector'
+import { mapValues } from 'lodash'
 
 export type CategoryListModelType = {|
   title: string,
@@ -21,10 +26,18 @@ export type ListEntryType = {|
   subCategories: Array<CategoryListModelType>
 |}
 
+export type ListContentModelType = {|
+  navigation: NavigationStackProp<*>,
+  files: PageResourceCacheStateType,
+  navigateToInternalLink: NavigateToInternalLinkParamsType => void,
+  resourceCacheUrl: string,
+  content: string
+|}
+
 type PropsType = {|
   categories: Array<ListEntryType>,
   title?: string,
-  content?: string,
+  listContent: ?ListContentModelType,
   /** A search query to highlight in the categories titles */
   query?: string,
   theme: ThemeType,
@@ -41,20 +54,33 @@ const CategoryThumbnail = styled(Image)`
   margin: 10px;
 `
 
-const VerticalPadding = styled.View`
-  padding: 0 20px;
-`
-
 /**
  * Displays a ContentList which is a list of categories, a caption and a thumbnail
  */
 class CategoryList extends React.Component<PropsType> {
+  getListContent (listContent: ListContentModelType): React.Node {
+    const { theme, language } = this.props
+    const cacheDictionary = mapValues(listContent.files, (file: PageResourceCacheEntryStateType) => {
+      return file.filePath.startsWith(RESOURCE_CACHE_DIR_PATH)
+        ? file.filePath.replace(RESOURCE_CACHE_DIR_PATH, listContent.resourceCacheUrl)
+        : file.filePath
+    })
+    return <CategoryListContent content={listContent.content}
+                         language={language}
+                         navigation={listContent.navigation}
+                         navigateToInternalLink={listContent.navigateToInternalLink}
+                         resourceCacheUrl={listContent.resourceCacheUrl}
+                         cacheDictionary={cacheDictionary}
+                         theme={theme} />
+  }
+
   render () {
-    const { categories, title, content, query, theme, onItemPress, language, thumbnail } = this.props
+    const { categories, title, listContent, query, theme, onItemPress, language, thumbnail } = this.props
+
     return <>
       {thumbnail && <CategoryThumbnail source={thumbnail} />}
       {title && <CategoryListCaption title={title} theme={theme} withThumbnail={!!(thumbnail)} />}
-      {!!content && <VerticalPadding><Html html={content} /></VerticalPadding>}
+      {!!listContent && this.getListContent(listContent)}
       {categories.map(({ model, subCategories }) =>
         <CategoryListItem key={model.path}
                           category={model}
--- a/native/src/modules/categories/components/CategoryListContent.js
+++ b/native/src/modules/categories/components/CategoryListContent.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react'
-import { Linking, Dimensions, View, Text } from 'react-native'
+import { Dimensions, Text, View } from 'react-native'
 import { type DisplayMetrics } from 'react-native/Libraries/Utilities/NativeDeviceInfo'
 import Html, { GestureResponderEvent, type HTMLNode, type RendererFunction } from 'react-native-render-html'
 import { _constructStyles } from 'react-native-render-html/src/HTMLStyles'
@@ -9,9 +9,9 @@ import styled from 'styled-components/native'
 import { type StyledComponent } from 'styled-components'
 import type { NavigationStackProp } from 'react-navigation-stack'
 import type { NavigateToInternalLinkParamsType } from '../../app/createNavigateToInternalLink'
-import buildConfig from '../../app/constants/buildConfig'
 import type { ThemeType } from '../../../../build-configs/ThemeType'
 import { RTL_LANGUAGES } from '../../i18n/constants'
+import onLinkPress from '../../common/onLinkPress'
 
 const VerticalPadding: StyledComponent<{}, {}, *> = styled.View`
   padding: 0 20px;
@@ -29,12 +29,10 @@ type ContentPropsType = {|
   navigation: NavigationStackProp<*>,
   resourceCacheUrl: string,
   navigateToInternalLink?: NavigateToInternalLinkParamsType => void,
-  cacheDictionary: {[string]: string},
+  cacheDictionary: { [string]: string },
   language: string,
   theme: ThemeType
 |}
-
-const HIJACK = new RegExp(buildConfig().internalLinksHijackPattern)
 
 class CategoryListContent extends React.Component<ContentPropsType, {| width: number |}> {
   constructor () {
@@ -47,7 +45,7 @@ class CategoryListContent extends React.Component<ContentPropsType, {| width: nu
   onChange = (dimensionsEvent: DimensionsEventType) => {
     const defaultWidth = 200
     this.setState({ width: (dimensionsEvent.window?.width || dimensionsEvent.screen?.width || defaultWidth) })
-  };
+  }
 
   componentDidMount () {
     Dimensions.addEventListener('change', this.onChange)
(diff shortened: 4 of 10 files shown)


native/src/modules/categories/components/Categories.js
```diff
@@ -132,6 +132,17 @@ class Categories extends React.Component<PropsType> {
     return categories.map(category => this.getListModel(category))
   }
 
+  getListContentModel (category: CategoryModel): ?ListContentModelType {
+    const { navigation, navigateToInternalLink, resourceCacheUrl } = this.props
+    return category.content ? {
```
klinzo, line 137: Is it possible that content is not defined? Or is this just for the root category?

native/src/modules/categories/components/CategoryList.js
```diff
@@ -41,20 +54,33 @@ const CategoryThumbnail = styled(Image)`
...
 class CategoryList extends React.Component<PropsType> {
+  getListContent (listContent: ListContentModelType): React.Node {
+    const { theme, language } = this.props
+    const cacheDictionary = mapValues(listContent.files, (file: PageResourceCacheEntryStateType) => {
+      return file.filePath.startsWith(RESOURCE_CACHE_DIR_PATH)
+        ? file.filePath.replace(RESOURCE_CACHE_DIR_PATH, listContent.resourceCacheUrl)
+        : file.filePath
+    })
+    return <CategoryListContent content={listContent.content}
+                         language={language}
```
klinzo, line 69: Indent

native/src/modules/categories/components/CategoryList.js
```diff
@@ -41,20 +54,33 @@ const CategoryThumbnail = styled(Image)`
...
-
 /**
  * Displays a ContentList which is a list of categories, a caption and a thumbnail
  */
 class CategoryList extends React.Component<PropsType> {
+  getListContent (listContent: ListContentModelType): React.Node {
+    const { theme, language } = this.props
+    const cacheDictionary = mapValues(listContent.files, (file: PageResourceCacheEntryStateType) => {
+      return file.filePath.startsWith(RESOURCE_CACHE_DIR_PATH)
+        ? file.filePath.replace(RESOURCE_CACHE_DIR_PATH, listContent.resourceCacheUrl)
```
klinzo, line 65: Maybe we should rather put the right values directly in the cache dictionary. Perfectly fine though in this PR.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+  resourceCacheUrl: string,
+  navigateToInternalLink?: NavigateToInternalLinkParamsType => void,
+  cacheDictionary: {[string]: string},
+  language: string,
+  theme: ThemeType
+|}
+
+const HIJACK = new RegExp(buildConfig().internalLinksHijackPattern)
+
+class CategoryListContent extends React.Component<ContentPropsType, {width: number}> {
```
klinzo, line 35:
```suggestion
class CategoryListContent extends React.Component<ContentPropsType, {| width: number |}> {
```

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+import type { ThemeType } from '../../../../build-configs/ThemeType'
+import { RTL_LANGUAGES } from '../../i18n/constants'
+
+const VerticalPadding: StyledComponent<{}, {}, *> = styled.View`
+  padding: 0 20px;
+`
+type WindowEventType = {
+  window: {width: number,
+  height: number}
+}
```
klinzo, line 21:
```suggestion
type WindowEventType = {|
  window: {| width: number,  height: number |}
|}
```

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+  constructor () {
+    super()
+    this.state = {
+      width: Dimensions.get('window').width
+    }
+  }
+
+  onChange = ({ window: { width } }: WindowEventType) => {
+    this.setState({ width: width })
+  };
```
klinzo, line 45:
```suggestion
  }
```

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  componentWillUnmount () {
+    Dimensions.removeEventListener('change', this.onChange)
+  }
+
+  onLinkPress = (evt: GestureResponderEvent, url: string) => {
+    const { language, navigation, navigateToInternalLink } = this.props
+
+    console.log(url)
```
klinzo, line 58: Remove or change to debug with more information :)

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+    } else if (url.includes('.png') || url.includes('.jpg')) {
+      navigation.navigate('ImageViewModal', { url })
+    } else if (navigateToInternalLink && HIJACK.test(url)) {
+      navigateToInternalLink({
+        url,
+        language
+      })
+    } else {
+      Linking.openURL(url).catch(err => console.error('An error occurred', err))
+    }
```
klinzo, line 70: Is this copied from somewhere else? Maybe it makes sense to move this in a helper method.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+    Dimensions.removeEventListener('change', this.onChange)
+  }
+
+  onLinkPress = (evt: GestureResponderEvent, url: string) => {
+    const { language, navigation, navigateToInternalLink } = this.props
+
+    console.log(url)
+    if (url.includes('.pdf')) {
+      navigation.navigate('PDFViewModal', { url })
+    } else if (url.includes('.png') || url.includes('.jpg')) {
```
klinzo, line 61: Are there other image types we should support? E.g. svgs?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+  window: {width: number,
+  height: number}
+}
+
+type ContentPropsType = {|
+  content: string,
+  navigation: NavigationStackProp<*>,
+  resourceCacheUrl: string,
+  navigateToInternalLink?: NavigateToInternalLinkParamsType => void,
+  cacheDictionary: {[string]: string},
```
klinzo, line 28:
```suggestion
  cacheDictionary: { [string]: string },
```
Styling still not working for you?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+        language
+      })
+    } else {
+      Linking.openURL(url).catch(err => console.error('An error occurred', err))
+    }
+  }
+
+  alterResources = (node: HTMLNode) => {
+    const { cacheDictionary } = this.props
+    if (node.attribs && cacheDictionary) {
```
klinzo, line 75: Afaics, cacheDictionary is never null or undefined.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+        }
+      ]
+      const rtlMargin = 10
+      const bulletSizeRelativeToFont = 2.8
+      if (rawChild) {
+        if (rawChild.parentTag === 'ul' && rawChild.tagName === 'li') {
+          prefix = listsPrefixesRenderers && listsPrefixesRenderers.ul ? listsPrefixesRenderers.ul(...rendererArgs) : (
+            <View style={{
+              marginRight: 10,
+              width: baseFontSize / bulletSizeRelativeToFont,
```
klinzo, line 126: Did you check whether this works correctly with different font sizes?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+              height: baseFontSize / bulletSizeRelativeToFont,
+              marginTop: baseFontSize / 2,
```
klinzo, line 128: Why not just use `baseFontSize / bulletSizeRelativeToFont` as well?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+              borderRadius: baseFontSize / bulletSizeRelativeToFont,
+              backgroundColor: 'black',
```
klinzo, line 130: Do we need to use a color from our theme here?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+              marginLeft: RTL_LANGUAGES.includes(language) ? rtlMargin : 0
```
klinzo, line 131: Why do you set `marginRight` to 10 always but not `marginLeft`? You should probably use the var for `marginRight` as well.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+            }} />
+          )
+        } else if (rawChild.parentTag === 'ol' && rawChild.tagName === 'li') {
+          prefix = listsPrefixesRenderers && listsPrefixesRenderers.ol ? listsPrefixesRenderers.ol(...rendererArgs) : (
+            <Text allowFontScaling={allowFontScaling} style={{
+              marginRight: 5,
+              fontSize: baseFontSize,
+              marginLeft: RTL_LANGUAGES.includes(language) ? rtlMargin : 0
+}}>{index + 1})</Text>
```
klinzo, line 140: Linebreaks

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+              backgroundColor: 'black',
+              marginLeft: RTL_LANGUAGES.includes(language) ? rtlMargin : 0
+            }} />
+          )
+        } else if (rawChild.parentTag === 'ol' && rawChild.tagName === 'li') {
+          prefix = listsPrefixesRenderers && listsPrefixesRenderers.ol ? listsPrefixesRenderers.ol(...rendererArgs) : (
+            <Text allowFontScaling={allowFontScaling} style={{
+              marginRight: 5,
+              fontSize: baseFontSize,
+              marginLeft: RTL_LANGUAGES.includes(language) ? rtlMargin : 0
```
klinzo, line 139: Same as above

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+}}>{index + 1})</Text>
+          )
+        }
+      }
+      return (
+        RTL_LANGUAGES.includes(language) ? (
+          <View key={`list-${nodeIndex}-${index}-${key}`} style={{ flexDirection: 'row', marginBottom: 10 }}>
```
klinzo, line 146: Maybe you could make the margin 10 a global variable as you are using it here as well.

klinzo (commented):
:+1: 

Do you know some categories where this new behavior could be tested?

Taggotty:
> Do you know some categories where this new behavior could be tested?

I only know Regensburg School and Enducation -> School or University both have some content, but not for german content. for images I used the following  snippet: 

```html
const htmlContent = `
    <h1>This HTML snippet is now rendered with native components !</h1>
    <h2>Enjoy a webview-free and blazing fast application</h2>
    <img src="https://i.imgur.com/dHLmxfO.jpg?2" />
    <em style="textAlign: center;">Look at how happy this native cat is</em>
`
```

native/src/modules/categories/components/Categories.js
Taggotty: the inner categories do not always have a content, whereby content means Images, Text and stuff.

Taggotty:
Most of the styling of the innercontent could not be applied. For example I am not quite sure about font resizing and stuff. That is because this library only has limited support for styling elements. But feel free to take a look for yourself.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: Actually most of the code is more or less copy/pasted from the menioned [github issue](https://github.com/archriss/react-native-render-html/issues/286). I tried renaming the arguments a little hopefully it is more clear now.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: removed it, as this margin was related to the text distance in lists and in my opinion we do not actually need it.

native/src/modules/categories/components/Categories.js
klinzo: Yeah, so it is a empty string. But does that matter?

klinzo (approved):
Looking good :+1: There are still some minor comments which are unresolved (hidden because of too much items).

I would not merge this before our next production release on 1.11 this weekend if that is okay for you?

klinzo:
Tested on iOS and Android:
* Horizontal and vertical screen orientation
* Different font sizes in accessibility settings
* Images

The only thing I observed: Bullet points are not changing size for different font sizes. But I guess thats okay as the text itself is working fine.

native/src/modules/categories/components/Categories.js
Taggotty: Ahhh. Now I get what you mean. Somehow I missed that the content is not actually optional. Yeah, I think this can be removed then.

Taggotty: Or not. The html-lib doesn't want to render an empty string. I'll keep it and add a comment

native/src/modules/categories/components/CategoryListContent.js
Taggotty: Sure. Sounds reasonable.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: Hmm. React-native-image-viewer doesn't seem to render images from svg-uris

native/src/modules/categories/components/CategoryListContent.js
Taggotty: This margin aligns the bullets relative to the font and in this case the bullets are supposed to be centered relative to the font.

native/package.json
```diff
@@ -64,7 +64,7 @@
     "react-native-pdf": "6.2.0",
     "react-native-permissions": "2.0.9",
     "react-native-progress": "^4.0.3",
-    "react-native-render-html": "^4.2.0",
+    "react-native-render-html": "^5.0.0-alpha.1",
```
maxammann, line 67: 5.0.0-alpha.2 is out :) probably not worth upgrading though

native/src/modules/categories/components/CategoryList.js
maxammann: @klinzo do you want to create a task for this? Does this require changes in a lot of places?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+            src: newResource
+          }
+          return node
+        }
+      }
+    }
+  }
+
+  // see https://github.com/archriss/react-native-render-html/issues/286
+  renderBulletLists: RendererFunction = (htmlAttribs, children, convertedCSSStyles, passProps) => {
```
maxammann, line 93: This function is too complicated imo. There are to many conditions. This should be refactored imo.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+      const bulletSizeRelativeToFont = 2.8
+      const bulletAlignmentRelativeToFont = 2
+      if (rawChild) {
+        if (rawChild.parentTag === 'ul' && rawChild.tagName === 'li') {
+          prefix = listsPrefixesRenderers && listsPrefixesRenderers.ul ? listsPrefixesRenderers.ul(...rendererArgs) : (
+            <View style={{
+              width: baseFontSize / bulletSizeRelativeToFont,
+              height: baseFontSize / bulletSizeRelativeToFont,
+              borderRadius: baseFontSize / bulletSizeRelativeToFont,
+              marginRight: RTL_LANGUAGES.includes(language) ? rtlListIndent : textDistanceToBullet,
```
maxammann, line 126: This line can be deduplicated. See blow.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+              marginLeft: RTL_LANGUAGES.includes(language) ? textDistanceToBullet : 0,
+              marginTop: baseFontSize / bulletAlignmentRelativeToFont,
+              backgroundColor: theme.colors.textColor
+            }} />
+          )
+        } else if (rawChild.parentTag === 'ol' && rawChild.tagName === 'li') {
```
maxammann, line 132: There are some magic values.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+          return node
+        }
+      }
+    }
+  }
+
+  // see https://github.com/archriss/react-native-render-html/issues/286
+  renderBulletLists: RendererFunction = (htmlAttribs, children, convertedCSSStyles, passProps) => {
+    const { language, theme } = this.props
+    const style = _constructStyles({
```
maxammann, line 95: Is this an internal API? If yes then we should probably not use it.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+        if (newResource) {
+          console.debug(`Replaced ${node.attribs.href} with ${newResource}`)
+          node.attribs = {
+            ...(node.attribs || {}),
+            href: newResource
+          }
+          return node
+        }
+      } else if (node.attribs.src) {
+        console.debug(`Found src: ${decodeURI(node.attribs.src)}`)
```
maxammann, line 78: These debug lines are not needed imo.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+class CategoryListContent extends React.Component<ContentPropsType, {| width: number |}> {
+  constructor () {
+    super()
+    this.state = {
+      width: Dimensions.get('window').width
+    }
+  }
+
+  onChange = (dimensionsEvent: DimensionsEventType) => {
+    const defaultWidth = 200
```
maxammann, line 46: What is the effect of this defaultWidth? Is there a case when window is not defined? If yes we maybe just can skip this onChange event instead.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+  content: string,
+  navigation: NavigationStackProp<*>,
+  resourceCacheUrl: string,
+  navigateToInternalLink?: NavigateToInternalLinkParamsType => void,
+  cacheDictionary: { [string]: string },
+  language: string,
+  theme: ThemeType
+|}
+
+class CategoryListContent extends React.Component<ContentPropsType, {| width: number |}> {
```
maxammann, line 37: Does this component use the correct content font?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,181 @@
...
+      const rendererArgs = [
+        htmlAttribs,
+        children,
+        convertedCSSStyles,
+        {
+          ...passProps,
+          index
+        }
+      ]
+      const textDistanceToBullet = 10
```
maxammann, line 115: Should be extracted to the top of the file probably.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -0,0 +1,179 @@
...
+        if (newResource) {
+          console.debug(`Replaced ${node.attribs.src} with ${newResource}`)
+          node.attribs = { ...(node.attribs || {}), src: newResource }
+          return node
+        }
+      }
+    }
+  }
+
+  // see https://github.com/archriss/react-native-render-html/issues/286
```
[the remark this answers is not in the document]
maxammann: So the function below gets removed as soon as the author implemented this?

native/src/modules/common/onLinkPress.js
```diff
@@ -0,0 +1,30 @@
...
+import { Linking } from 'react-native'
+import type { NavigateToInternalLinkParamsType } from '../app/createNavigateToInternalLink'
+import buildConfig from '../app/constants/buildConfig'
+import type { NavigationStackProp } from 'react-navigation-stack'
+
+type NavigateToInternalLinkType = NavigateToInternalLinkParamsType => void
+
+const HIJACK = new RegExp(buildConfig().internalLinksHijackPattern)
+
+const onLinkPress = (url: string,
```
maxammann, line 12: This name is too generic in my opinion as it is placed in the root of `common`. Maybe a name like onInternalLinkPress or onAnchorTagPress is better.

maxammann (changes requested):

native/src/modules/categories/components/CategoryListContent.js
klinzo: It is, but there is no other workaround right now to render lists with RTL layout. As @Taggotty mentioned she well create an issue to remove this once it is fixed in the library.

maxammann: If it is really foreseeable that this function gets removed in the near future then we can merge this as is. But if not then we should improve this and not beautify the html rendering. Also the issue does not mention anything about the bullet points.

I dont think this feature is that important that it justifies to use internal API of an alpha npm module.

native/src/modules/categories/components/CategoryListContent.js
klinzo: As this is just a workaround for a not implemented functionality of the library (RTL lists) I think this is not worth the effort to be honest.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: > If it is really foreseeable that this function gets removed in the near future then we can merge this as is. But if not then we should improve this and not beautify the html rendering. Also the issue does not mention anything about the bullet points.
>

It will surely not be updated in _near_ future, as the feature seems to be added for version 6.0.0 of this npm package. 
Still you are right it is not part of the issue, but it seems to fall into the same category. The behavior without this custom renderer is that the bullet points are rendered on the left side so it is rather a beauty-related issue.

> I dont think this feature is that important that it justifies to use internal API of an alpha npm module.
the function is also used in their demo on how to implement a renderer see [here](https://github.com/archriss/react-native-render-html/blob/9f7168be5ceb1cf4de6e51d3a93bd9cca07e5c3d/Demo/src/snippets.js#L273)

Personally I don't mind leaving this part out.

native/src/modules/categories/components/CategoryListContent.js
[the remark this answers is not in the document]
Taggotty: yes. even though this release may take some time.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: you mean ul, ol and li? do you think we need to actually extract them? I think they speak for themselves

native/src/modules/categories/components/CategoryList.js
klinzo: I don't think it is worth to create a task for this. I mean we have so many tasks already, this can be refactored once we have a third occurrence of this code or we change something here anyway.

native/src/modules/categories/components/CategoryListContent.js
klinzo: Honestly I don't see why we should remove this part again as it fixes a RTL issue just for the sake of not using internal API. I don't like it either here, but I don't think we should spent more time here.

yarn.lock
```diff
@@ -13826,10 +13826,10 @@ react-native-ratings@^6.3.0:
     lodash "^4.17.4"
     prop-types "^15.5.10"
 
-react-native-render-html@^4.2.0:
-  version "4.2.4"
-  resolved "https://registry.yarnpkg.com/react-native-render-html/-/react-native-render-html-4.2.4.tgz#0b642fb72aee5935bd1cfc427a115ffdf881d522"
-  integrity sha512-OiLItEzKgS7dzD9XI5bHhjcUEfpWdzH1FgexzjbBdICPfYjmmcefpcRmLZY1+HMfxJ7wL8iF1PzTF48LchGTBA==
+react-native-render-html@^5.0.0-alpha.1:
+  version "5.0.0-alpha.2"
```
Taggotty, line 13830: wow...

native/package.json
Taggotty: will change to 5.0.0-alpha.2. See [here](https://github.com/Integreat/integreat-app/pull/48/files#r515019513)

native/src/modules/categories/components/CategoryList.js
maxammann: I'm oke with not changing it now :D but having many tasks should not be an argument :D

maxammann (approved):
There are some minor conversations open I think.
I still think it would be better to wait until the feature is implemented upstream instead of forcing it into the app right now. Probably wont harm to merge though.

maxammann (approved):

PR #49, opened 2020-10-27
IGAPP-166: Change type from Component to NavigationComponent

Taggotty:

### DIFF
--- a/native/src/modules/endpoint/hocs/withRouteCleaner.js
+++ b/native/src/modules/endpoint/hocs/withRouteCleaner.js
@@ -2,9 +2,7 @@
 
 import * as React from 'react'
 import type { NavigationStackProp } from 'react-navigation-stack'
-import type {
-  NavigationComponent
-} from 'react-navigation'
+import type { NavigationComponent } from 'react-navigation'
 /**
  * This hoc removes routes from the state when unmounting
  *
--- a/native/src/modules/endpoint/hocs/withRouteCleaner.js
+++ b/native/src/modules/endpoint/hocs/withRouteCleaner.js
@@ -2,7 +2,9 @@
 
 import * as React from 'react'
 import type { NavigationStackProp } from 'react-navigation-stack'
-
+import type {
+  NavigationComponent
+} from 'react-navigation'
 /**
  * This hoc removes routes from the state when unmounting
  *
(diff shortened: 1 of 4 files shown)


native/src/routes/pdf/components/PDFViewModal.js
```diff
@@ -51,4 +52,6 @@ class PDFViewModal extends React.Component<PropsType, StateType> {
   }
 }
 
-export default withTheme<PropsType>(PDFViewModal)
+export default withRouteCleaner(
```
klinzo, line 55: Why is this change necessary? Did you check whether there is actually an action executed on closing a pdf route? If no action is set, the app crashes afaik.

Taggotty: you're right. did not yet quite think it through. The main reason why the typing failed is because a Component cannot be interpreted as a NavigationComponent. Do you have any ideas on how to solve that in a pretty way?

Taggotty: EDIT. I think I found  the solution. Flow more or less doesn't remember the other props when inferring a Generic type too specific as e.g here: 
```javascript
function withRouteCleaner<Props: { navigation: NavigationStackProp<*> }> (
```
Later then it will only remember the specific props defined in this generic type **until** you generalize the type again, where the type is not specified as e.g. here (as there is no given return type anything could be correct):
```javascript
 const withLayout = <T>(WrappedComponent: React.ComponentType<T>) => {
  return class extends React.Component<T> {
    static displayName = wrapDisplayName(WrappedComponent, 'withLayout')

    render () {
      return <LayoutContainer>
          <WrappedComponent {...this.props} />
      </LayoutContainer>
    }
  }
}
```
To keep the type specifications now you can just add those you need as union types, similar as described [here](https://github.com/react-navigation/react-navigation/issues/3643), for the return value, which is the solution I chose as it keeps the types more specific. 
Still it is not completely correct. `withTheme` will not always return a component of type NavigationComponent - at least I think - but I think this is good enough for our purpose. Any other opinions? Those were all the findings I made at least, but I still could be wrong about sth. Feel free to correct me :)

klinzo (approved):

native/src/modules/endpoint/hocs/withRouteCleaner.js
```diff
@@ -11,7 +11,7 @@ import type { NavigationStackProp } from 'react-navigation-stack'
  */
 function withRouteCleaner<Props: { navigation: NavigationStackProp<*> }> (
   Component: React.ComponentType<Props>
-): React.ComponentType<Props> {
+): NavigationComponent {
```
maxammann, line 14: :+1:

maxammann (approved):

PR #54, opened 2020-10-30
IGAPP-338: Disable deliveries with environment variables

maxammann:


.circleci/src/jobs/check_web.yml
```diff
@@ -1,3 +1,8 @@
+parameters:
```
klinzo: Why do we need a build config name here? This is executed independently of a build config name just like e.g. check_native

klinzo (approved):

maxammann:
Variable in circleci is set

.circleci/src/jobs/check_web.yml
maxammann: I think I already removed it

NKlug (approved):

PR #55, opened 2020-10-30
IGAPP-53: Feedback Dropdown cannot handle Special Characters

maxammann:


NKlug (approved):

klinzo (approved):
Good catch! Didn't think this is because of 18n :thinking:

Taggotty (approved):

PR #57, opened 2020-10-31
IGAPP-45: Footer to bottom in IE11

Taggotty:

* Tested on IE11 and Edge


klinzo (approved):

NKlug (approved):

web/src/modules/layout/components/Layout.js
```diff
@@ -9,7 +9,7 @@ import withPlatform from '../../platform/hocs/withPlatform'
 const RichLayout = styled.div`
   position: relative;
   display: flex;
-  min-height: 100vh;
```
Schedulaar, line 12: This might (or might not) break stuff quite a lot... Do scroll bars look okay? Also when using popups (feedback, etc.)? Are pages with lots of content (overflowing everything) working as expected?

Taggotty: I actually did not encounter any issues. But if you think that this could break sth. Test it out for yourself. I already checked out  Edge, Chromium and IE11.

klinzo: Are these conflicting props? Otherwise we could just keep both :)

maxammann: I tested it in firefox, footer works fine

maxammann (approved):

native/src/routes/pois/components/PoiListItem.js
```diff
@@ -1,37 +1,37 @@
...
-                navigateTo={navigateToPoi}
-                theme={theme}>
-        {poi.location && <Description theme={theme}>{poi.location.location}</Description>}
-      </ListItem>
-    )
-  }
-}
-
-export default PoiListItem
+// @flow
```
maxammann, line 1: We have these changes quite a lot. Can you check whats wrong with the line endings or file modes?

maxammann (changes requested):

web/src/modules/layout/components/Layout.js
```diff
@@ -9,7 +9,8 @@ import withPlatform from '../../platform/hocs/withPlatform'
...
+  height: 100vh;
+  max-height: 100vh;
```
klinzo, line 13: You readded `max-height` instead of `min-height` :D

klinzo (changes requested):

web/src/modules/layout/components/Layout.js
Taggotty: 🤦🏼

native/src/routes/pois/components/PoiListItem.js
klinzo: I got this change as well. I think this is probably just a remnant from the work of Anja, I think I remember that we have problems then.

klinzo (approved):
Tested on firefox :+1:

Znippet:
Tested in Edge and IE, both look fine

Znippet (commented):
Tested in IE11 and current Edge, both look fine

Znippet (approved):

native/src/routes/pois/components/PoiListItem.js
maxammann: True, I also scanned the whole repo for wrong line endings. this is fixed now

maxammann (approved):

PR #58, opened 2020-10-31
IGAPP-339: Remove duplicated values from build configs

klinzo:


web/src/modules/layout/components/__tests__/Footer.spec.js
```diff
@@ -6,20 +6,20 @@ import { shallow } from 'enzyme'
...
 describe('Footer', () => {
   it('should insert items', () => {
+    const preVersion = global.__VERSION__
+    global.__VERSION__ = 'vX.X'
     const component = <Footer>
       <div>MockNode</div>
       <div>AnotherMockNode</div>
     </Footer>
     expect(shallow(component)).toMatchSnapshot()
+    global.__VERSION__ = preVersion
```
Taggotty, line 16: I think you can remove resetting the version here. Afaik each test suite has it own environment instance (and is therefore only set for this File. See [here](https://jestjs.io/docs/en/configuration#globals-object)

Taggotty (approved):

maxammann:
I will wait with reviewing until the review of the branch IGAPP-271-merge-build-configs is finished.

web/src/modules/layout/components/__tests__/Footer.spec.js
maxammann: I think it is good practice to clean up after setting something global. This could easily affect tests in other tests.

maxammann (approved):

maxammann:
The whole build config is currently accessible in web and native. Should this be done in a separate task?

klinzo:
> The whole build config is currently accessible in web and native. Should this be done in a separate task?

Yep, here: https://issues.tuerantuer.org/browse/IGAPP-364

web/src/modules/layout/components/__tests__/Footer.spec.js
klinzo: Agreeing with @maxammann. In this case it is not necessary, but won't hurt and potentially prevent wrong behavior/bugs.

klinzo:
@maxammann @Taggotty could you review the last commit again please? Does this work for you?
https://github.com/Integreat/integreat-app/pull/58/files/d264171f0fbc8d7ea3662ed2fbecc50ad1b135e9..a01a4c90e4b7308c9585901eb20efcff8c6f1a9f

web/tools/webpack.config.babel.js
```diff
@@ -111,7 +111,7 @@ const createConfig = (env = {}) => {
     // What information should be printed to the console
     stats: 'minimal',
     performance: {
-      hints: isProductionBuild ? 'error' : false,
+      hints: isProductionBuild && !devServer ? 'error' : false,
```
maxammann, line 114: Isn't !devServer always true when isProducitonBuild is true?

maxammann (approved):

web/tools/webpack.config.babel.js
klinzo: Not for malte, we only have one build config here. Maybe we should add one though at some point.

maxammann: But why do we not want to show hints when we are running as dev server?

maxammann: Oh because it is bigger, right?

klinzo: Yes, otherwise compiling fails.

maxammann: Hm maybe we should add this configh in this pr. I think it is not good that we can run a dev-server and buildConfig().development is false. Or should we do this in the Bug you opened this week about the bundle size?

klinzo: Separate PR please because other PRs depending on this :) Will do in the opened bug

PR #59, opened 2020-11-02
IGAPP-163: Switch to Hermes

maxammann:


maxammann:
I tested all screens on Android. So far no problems. Had to dig a bit deep to find out why there were stacktraces when starting the app. Actually it is react-natives fault.

Znippet (approved):

klinzo (approved):
Tested all screens on iOS (simulator).

Taggotty (approved):
Also tested on Android. Found a few issues, but they were all not related to this PR (as they also occur on main)

PR #63, opened 2020-11-04
IGAPP-169: Add logic for progress bar

Znippet:


native/src/modules/endpoint/hocs/withPayloadProvider.js
```diff
@@ -94,6 +94,7 @@ const withPayloadProvider = <S: { dispatch: Dispatch<StoreActionType> }, R: {}>
           return <LanguageNotAvailableContainer languages={props.availableLanguages}
                                                 changeLanguage={this.changeUnavailableLanguage} />
         } else if (props.status === 'loading') {
+          // TODO: console.log('render loading', props.progress)
```
Znippet, line 97: Also here? [IGAPP-292](https://issues.tuerantuer.org/secure/RapidBoard.jspa?rapidView=14&view=detail&selectedIssue=IGAPP-292): Notify user if offline when refreshing

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,45 +1,55 @@
...
 import { isEmpty } from 'lodash'
 
 export type TargetFilePathsType = { [path: string]: string }
 
 export type FetchResultType = { [path: string]: {| lastUpdate: string, url: string, errorMessage: ?string |} }
 
-export type ProgressCallbackType = (progress: number) => void
-
 class FetcherModule {
   // TODO NATIVE-264: Correctly handle already fetching
```
Znippet, line 13: Also: [IGAPP-217](https://issues.tuerantuer.org/secure/RapidBoard.jspa?rapidView=14&view=detail&selectedIssue=IGAPP-217): Fix the error "Already fetching!" when switching cities

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,45 +1,55 @@
...
   static currentlyFetching = false
 
-  async fetchAsync (
-    targetFilePaths: TargetFilePathsType,
-    progress: ProgressCallbackType
-  ): Promise<FetchResultType> {
+  fetchAsync = (targetFilePaths: TargetFilePathsType): [Promise<FetchResultType>, Channel] => {
```
klinzo, line 16: I think the code in this method could be improved further :)

klinzo (commented):
Anything special you want to know?

native/src/modules/endpoint/hocs/withPayloadProvider.js
Znippet: Yes, true, I just saw that this is the place for implementingIGAPP-292. But it's a completely different Task and does not belong here...

native/src/modules/fetcher/FetcherModule.js
Znippet: That belongs to the existing comment. just had a look,, NATIVE-264 does not exist any more, It is renamed to  IGAPP-217
If corrected the Task number

klinzo: Ahhh right, thanks :+1:

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,45 +1,55 @@
...
     FetcherModule.currentlyFetching = true
 
-    const subscriptions = []
-    subscriptions.push(NativeFetcherModuleEmitter.addListener('progress', progress))
+    const progressSubscription = NativeFetcherModuleEmitter.addListener('progress', emit)
 
+    let emit
+    const progressChannel = eventChannel<number>(emitter => {
+      emit = emitter
+      return () => {}
```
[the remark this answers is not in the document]
Znippet: Not sure... I don't want the channel to be able to quit my load-cache-subscription.

I think this is only necessery if the subscrption ist started in the channel, so itself must quit it.
In my case the channel is independent from the subscription to load-cache. I both subscribe and unsubscribe outside of the channel method.

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,45 +1,55 @@
...
+          // why this happened. As this cause an inconsistent state it is better to throw an error in this case.
+          throw new Error('Fetch failed for some reason!')
+        }
+      })
+      return [fetchPromise, progressChannel]
+    } catch {
+      progressSubscription.remove()
+      emit(END)
       FetcherModule.currentlyFetching = false
     }
```
Znippet, line 54: I think I do not need the try catch block, but I'm afraid to remove it.

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -19,6 +19,28 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
     `${message}'Failed to download ${result.url} to ${path}': ${result.errorMessage}\n`, '')
 }
 
+function * watchOnProgress (chan: Channel<number>): Saga<void> {
+  let prevStep = 0
+  while (true) {
```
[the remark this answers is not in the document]
Znippet: That's true. Any suggestions how I could handle cases if step 100 is never reached (not sure if it ever may happen)

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -19,6 +19,28 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
     `${message}'Failed to download ${result.url} to ${path}': ${result.errorMessage}\n`, '')
 }
 
+function * watchOnProgress (chan: Channel<number>): Saga<void> {
```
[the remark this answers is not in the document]
klinzo: Why are you hiding comments? :D Imo this is kind of confusing, just resolving is the better option imo.

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -19,6 +19,29 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
     `${message}'Failed to download ${result.url} to ${path}': ${result.errorMessage}\n`, '')
 }
 
+function * watchOnProgress (channel: Channel<number>): Saga<void> {
+  let prevStep = 0
+  while (prevStep < 1) {
+    const progress = yield take(channel)
+    const stepWidth = 10
+    const newStep = Math.floor(progress * stepWidth) / stepWidth
```
klinzo, line 27: Couldn't this same behavior achieved by rounding to one decimal place?

native/src/modules/fetcher/FetcherModule.js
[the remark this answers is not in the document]
klinzo: I think there is no reason to have the subscription outside of the channel, you are just subscribing and unsubscribing anyway or am I missing something here?

Just close the channel in the `fetchPromise.then()` block.

native/src/modules/fetcher/FetcherModule.js
klinzo: But this was already the case before already that we don't have a catch block :thinking:

maxammann (commented):
Looks good, looking forward for review!

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,47 +1,42 @@
...
         throw new Error('Fetch failed for some reason!')
       }
-
-      return result
-    } finally {
-      subscriptions.forEach(sub => sub.remove())
-      FetcherModule.currentlyFetching = false
-    }
+    })
+    return [fetchPromise, progressChannel]
```
maxammann, line 39: Better use an object here with proper names.

maxammann: You can also split this funciton into two. One that returns just the progressChannel. The other function takes the channel and awaits the promise. Not sure about naming though

native/src/modules/fetcher/FetcherModule.js
Znippet: I removed the catch block (previously "finally") so the "throw error" works again

native/src/modules/fetcher/FetcherModule.js
Znippet: Extracted creationChannel from fetchAsync, channel is opened and closed now independent of fetching itself

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: Yes, it could - but plain javascript does not support truncation to a given number of decimal places

klinzo:
@Znippet just open it for review once we should have a final look :)

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -19,6 +19,26 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
     `${message}'Failed to download ${result.url} to ${path}': ${result.errorMessage}\n`, '')
 }
 
+function * watchOnProgress (): Saga<void> {
+  const channel = new FetcherModule().createProgressChannel()
+  try {
+    let progress = 0
+    while (progress < 1) {
```
klinzo, line 26: Shouldn't we dispatch an action with `progress: 1` as well?

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -32,12 +52,18 @@ export default function * fetchResourceCache (
       return acc
     }, {})
 
-    const results = yield call(new FetcherModule().fetchAsync, targetFilePaths, progress => console.log(progress))
+    if (FetcherModule.currentlyFetching) {
+      throw new Error('Already fetching!')
+    }
```
klinzo, line 57: Why is this necessary here?

native/src/modules/fetcher/FetcherModule.js
```diff
@@ -1,47 +1,53 @@
...
+        }
+        prevStep = newStep
+        emitter(newStep)
+      })
+
+      return () => subscription.remove()
+    })
+  }
+
+  fetchAsync = (targetFilePaths: TargetFilePathsType): Promise<FetchResultType> => {
```
klinzo, line 34: This is not really async anymore, could we revert this function to some extend (without the listener of course)?

klinzo (commented):
I guess you ran your linter at some time which fixed all warnings (lines to long). I think this is fine for now, but in general not to many files should be changed that are not relevant for the PR. Good work :+1:

native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -19,6 +19,26 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
     `${message}'Failed to download ${result.url} to ${path}': ${result.errorMessage}\n`, '')
 }
 
+function * watchOnProgress (): Saga<void> {
```
Taggotty, line 22: Please follow the Airbnb Style Guide:   https://github.com/airbnb/javascript#generators--spacing

Taggotty (approved):
Quite a lot of changes that were not related to this PR. I hope I didn't miss anything. But great work nevertheless.

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: Exactly, if the content is loaded an event to show the page is already dispatched, so if event if there is 100% there is little time to see it before the loading display is hidden.
I wanted to be sure that this event really stops BEFORE the "Loading Finished" event is dispatched.
When the Loading Finished event is dispatched progress will be set to 1 anyways

Perhaps it would be nicer in the way we already discussed, an own Type "loading" beside of "ready" and "error".
The cache "loading" type would contain "progress: number",  "ready" would only contain the loaded value (like before), not the progress value

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: I could not throw the error in FetcherModule.createProgressChannel itself because I'm forking a new process here.
Throwing an error in createProgressChannel itself did not cancel this workflow here.
 
So I decided to check it before creating the progress channel here. 

Perhaps you have a suggestion how to do it inside of createProgressChannel itself, it would be nicer.

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: My auto formatter does not know that, perhaps we could teach that eslint?

Either way, done.

native/src/modules/fetcher/FetcherModule.js
Znippet: We can rename it to fetch (without async), depends how we define "async". Inside of the method itself no await is used, so the "async" tag is not needed, that is true. But the result still as async.
The method returns no value itself but a promise, so it behaves in the view of the caller in the same way as a method which awaits.

Did I understand your question correctly? If not perhaps you could give me a suggestion / proposal how to "refactor without listener".

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: ... but I suggest this is "future music" like we Germans say ;)

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: Our linter does not allow this:
// good
function* foo() {
  // ...
}

I have to use this (to make the linter happy):
// bad
function * foo() {
  // ...
}

native/src/modules/endpoint/sagas/fetchResourceCache.js
klinzo: Why do we have to throw an error at all in createProgressChannel if it is already fetching? Isn't this already handled in `fetchAsync`?

klinzo: Ah I see, you removed it in fetchAsync. Would it be a problem if a channel is created while a fetch is in progress?

native/src/modules/fetcher/FetcherModule.js
klinzo: You can just await the result here. E.g.:
```js
  fetchAsync = async (targetFilePaths: TargetFilePathsType): Promise<FetchResultType> => {
    if (isEmpty(targetFilePaths)) {
      return {}
    }
    FetcherModule.currentlyFetching = true

    const result = await NativeFetcherModule.fetchAsync(targetFilePaths)

    FetcherModule.currentlyFetching = false

    if (!result) {
      // While testing the app I noticed that the FetchResultType of this function was empty. I am absolutely not sure
      // why this happened. As this cause an inconsistent state it is better to throw an error in this case.
      throw new Error('Fetch failed for some reason!')
    }
    return result
  }

```

klinzo (approved):

native/src/modules/common/renderHtml.js
```diff
@@ -85,7 +85,13 @@ const renderJS = (cacheDictionary: { [remoteUrl: string]: string }) => `
 `
 
 // language=HTML
-const renderHtml = (html: string, files: PageResourceCacheStateType, theme: ThemeType, language: string, resourceCacheUrl: string) => {
+const renderHtml = (
```
maxammann, line 88: Please avoid reformats. If this is due vscode then we somehow need to adjust our formatters.

native/src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
```diff
@@ -61,19 +62,12 @@ describe('fetchResourceCache', () => {
...
-          if (effect.fn.name === 'fetchAsync') {
-            throw new Error('Jemand hat keine 4 Issues geschafft!')
-          }
-          return next()
-        }
-      })
       .put({
         type: 'FETCH_RESOURCES_FAILED',
         params: {
-          message: 'Error in fetchResourceCache: Jemand hat keine 4 Issues geschafft!',
```
maxammann, line 76: I think this test was different before. The test should provide what happens when `fetchAsync` is called. Then it asserts that the `FETCH_RESOURCES_FAILED` is returned.

Now it returns for some reason Already fetching.

native/src/modules/fetcher/FetcherModule.js
maxammann: The proposal by klinzo is equivalent, but more readable. It uses async/await syntax instead of then/catch

maxammann (approved):
We need to come up with a way of using a similar formatter. IntelliJ and vscode is doing stuff differently. If we continue like this we get huge diffs with no changes.

native/src/modules/common/renderHtml.js
klinzo: In fact I think this are all changes due to max line length so I think we don't have a problem. IntelliJ won't reformat these changes. I was playing with the idea of fixing this warnings for some time already as well :D

klinzo:
> We need to come up with a way of using a similar formatter. IntelliJ and vscode is doing stuff differently. If we continue like this we get huge diffs with no changes.

@maxammann I think this is fine. See answer from above:
>In fact I think this are all changes due to max line length so I think we don't have a problem. IntelliJ won't reformat these changes. I was playing with the idea of fixing this warnings for some time already as well :D

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: ... I'm not sure^^
If an error happens in the "FetcherModule().fetchAsync" call then "yield cancel(progressTask)" would not be called.
But altogether is surrounded by a try catch, does the forked task also gets closed also if the try fails?

native/src/modules/fetcher/FetcherModule.js
Znippet: Yes, looks much cleaner, thanks!

native/src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
Znippet: It can not reach "fetchAsync" any more, because before creating the channel I added a new "throw error"   

````
if (FetcherModule.currentlyFetching) {
  throw new Error('Already fetching!')
}
const progressTask = yield fork(watchOnProgress)
const results = yield call(new FetcherModule().fetchAsync, targetFilePaths)
````

native/src/modules/common/renderHtml.js
Znippet: Yes and yes - I tried to fix a linting error in my code and started from top of my linting errors
So I accidently solved the warnings (with manual autoformat). After solving my "real" error I decided to commit the solutions to the warnings also.
Sorry for the many changed files because of this

native/src/modules/endpoint/sagas/fetchResourceCache.js
Znippet: I will leave it this way, just to be sure

PR #66, opened 2020-11-10
IGAPP-197: Resolve Flowfixme in navigator

Taggotty:

see https://reactnavigation.org/docs/4.x/navigation-actions

### DIFF
--- a/native/src/routes/landing/containers/LandingContainer.js
+++ b/native/src/routes/landing/containers/LandingContainer.js
@@ -7,7 +7,7 @@ import type { Dispatch } from 'redux'
 import type { StoreActionType } from '../../../modules/app/StoreActionType'
 import Landing from '../components/Landing'
 import type { NavigationStackProp } from 'react-navigation-stack'
-import { type NavigationReplaceAction, StackActions } from 'react-navigation'
+import { StackActions } from 'react-navigation'
 import { generateKey } from '../../../modules/app/generateRouteKey'
 import type { StatusPropsType } from '../../../modules/endpoint/hocs/withPayloadProvider'
 import withPayloadProvider from '../../../modules/endpoint/hocs/withPayloadProvider'
@@ -62,8 +62,10 @@ class LandingContainer extends React.Component<ContainerPropsType> {
     const path = `/${cityCode}/${language}`
     const key: string = generateKey()
 
+    navigation.navigate('CityContent')
+
     // $FlowFixMe newKey missing in typedef
-    const action: NavigationReplaceAction = StackActions.replace({
+    navigation.dispatch(StackActions.replace({
       routeName: 'Dashboard',
       params: {
         cityCode,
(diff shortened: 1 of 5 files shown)


native/src/routes/landing/containers/LandingContainer.js
```diff
@@ -62,22 +62,18 @@ class LandingContainer extends React.Component<ContainerPropsType> {
...
       },
       newKey: key
-    })
-
-    // $FlowFixMe For some reason action is not allowed to be a StackAction
-    navigation.navigate({
-      routeName: 'CityContent',
-      action: action
-    })
+    }))
```
Schedulaar, line 76: This seems to be wrong... I guess we should dispatch the StackAction in the StackNavigator, not in the top level SwitchNavigator...  (Just an educated guess)

native/src/routes/landing/containers/LandingContainer.js
```diff
@@ -62,22 +62,18 @@ class LandingContainer extends React.Component<ContainerPropsType> {
     const path = `/${cityCode}/${language}`
     const key: string = generateKey()
 
+    navigation.navigate('CityContent')
+
     // $FlowFixMe newKey missing in typedef
-    const action: NavigationReplaceAction = StackActions.replace({
+    navigation.dispatch(StackActions.replace({
```
maxammann, line 68: What is the difference between dispath and nativate?

maxammann (approved):

native/src/routes/landing/containers/LandingContainer.js
klinzo: I think I did this change some time back as well and it led to problems... not sure of which kind though anymore...

native/src/routes/landing/containers/LandingContainer.js
Taggotty: Dispatch takes an action, while navigate is the function. Furthermore, dispatch can take any Stack/Navigation Action. In our case what we want is described [here](https://reactnavigation.org/docs/nesting-navigators#navigator-specific-methods-are-available-in-the-navigators-nested-inside).

native/src/routes/landing/containers/LandingContainer.js
Taggotty: It does make sense, but I think it is not possible, as there is no real entry point to a city before you actually enter the city (with specific city and language code). Before you are just somewhere in the switch navigator (in this case in the landing container) which itself is only a basic Navigator.

release-notes/unreleased/IGAPP-166-ResolveFlowFixMeInNavigator.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-166
```
klinzo, line 1: imo this should be removed :)

klinzo (approved):

PR #67, opened 2020-11-10
IGAPP-334: Setup flipper

klinzo:


NKlug (approved):

native/docs/debugging.md
```diff
@@ -1,31 +1,22 @@
...
+* \[optional\]: Enable plugins for installation
 
-Shake the device or press Ctrl+M to enter the menu. If you are connected through adb to your android device you can also 
-run: `adb shell input keyevent 82`
+### iOS
 
-## Reload the app
-You can [enter the development menu](#enter-the-development-menu) and click "Reload" or use adb: `adb shell input text "RR"`
+For iOS, some additional steps are required:
+* `bundle exec pod install`
```
maxammann, line 18: I think this could be excluded, this should be done even if you do not use flipper.

native/docs/debugging.md
```diff
@@ -1,31 +1,22 @@
...
+* Install [idb (iOS Development Bridge)](https://github.com/facebook/idb#quick-start):
+    * `brew tap facebook/fb`
+    * `brew install idb-companion`
+    * `pip3 install fb-idb`
```
maxammann, line 22: Is this really needed? I just downloadeed the fliipper app and it worked out of the box.

native/package.json
```diff
@@ -122,9 +122,11 @@
     "jest-transform-stub": "^2.0.0",
     "jetifier": "^1.6.5",
     "metro-react-native-babel-preset": "^0.59.0",
+    "react-native-flipper": "^0.64.0",
```
maxammann, line 125: Flipper is already integrated in our current react-native setup: https://fbflipper.com/docs/getting-started/react-native

maxammann (changes requested):

native/ios/Podfile.lock
```diff
@@ -469,6 +471,7 @@ DEPENDENCIES:
   - React-jsi (from `../node_modules/react-native/ReactCommon/jsi`)
   - React-jsiexecutor (from `../node_modules/react-native/ReactCommon/jsiexecutor`)
   - React-jsinspector (from `../node_modules/react-native/ReactCommon/jsinspector`)
+  - react-native-flipper (from `../node_modules/react-native-flipper`)
```
maxammann, line 474: Why are ther changes without changes to the Podfile?

maxammann (changes requested):
Here is the setup guide for RN 0.63+ https://fbflipper.com/docs/getting-started/react-native-ios#react-native-063

native/docs/debugging.md
klinzo: Where? On MacOS?

native/package.json
klinzo: Does not hurt though, does it?

native/docs/debugging.md
maxammann: Yes

native/package.json
maxammann: This is not how flipper should be integreated after 0.63 afaik. This could lead to duplicate binaries or whatever

native/docs/debugging.md
klinzo: It did not for me... or at least I was getting a message that it does not work without, but I did not try it

Taggotty:
I've had the the following issue 
```
java.lang.UnsatisfiedLinkError: couldn't find DSO to load: libflipper.so caused by: dlopen failed: library "libevent_extra-2.1.so" not found result: 0
```
with this solution: https://github.com/facebook/flipper/issues/824#issuecomment-611868592 + restarting the emulator

Taggotty (approved):

klinzo:
@maxammann from the page you linked: `These instructions are aimed at people manually adding Flipper to a React Native 0.62+ app. This should only be necessary if you have an existing app that cannot be upgraded with the React Native Upgrade tool.`

This is not necessary for us :) Or rather, it is already done.

native/package.json
klinzo: redux-flipper does not work without an explicit dependency.

react-native-flipper is also included in the official example here: https://github.com/facebook/flipper/blob/master/react-native/ReactNativeFlipperExample/package.json
and mentioned here as well: https://github.com/facebook/flipper/tree/master/react-native/react-native-flipper#installation

maxammann: Alright I checked what is happening here. react-native-flipper just exposes bindings for flipper. That means flipper works without that package well, but redux-flipper requires the bindings.

native/package.json
```diff
@@ -122,9 +122,11 @@
     "jest-transform-stub": "^2.0.0",
     "jetifier": "^1.6.5",
     "metro-react-native-babel-preset": "^0.59.0",
+    "react-native-flipper": "^0.54",
```
maxammann, line 125: Should we use version 0.63 here? https://www.npmjs.com/package/react-native-flipper/v/0.63.0

https://github.com/jk-gan/redux-flipper/blob/master/package.json#L31

maxammann (approved):

native/package.json
klinzo: The diff log says to use flipper v0.54: https://react-native-community.github.io/upgrade-helper/?from=0.63.2&to=0.63.3

I guess we should use the same version here

klinzo:
@maxammann any idea why no sourcemap is generated on ios here? react native upgrade?

native/package.json
klinzo: @maxammann what do you think?

maxammann: oh alright :+1:

PR #68, opened 2020-11-10
IGAPP-266: Configurable imprint, about and privacy

klinzo:

### DIFF
--- a/build-configs/package.json
+++ b/build-configs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "build-configs",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "Build configs for the integreat project",
   "main": "index.js",
   "license": "MIT",
--- a/native/src/routes/settings/createSettingsSections.js
+++ b/native/src/routes/settings/createSettingsSections.js
@@ -84,7 +84,8 @@ const createSettingsSections = ({ setSetting, t, languageCode, cityCode }: Creat
           accessibilityRole: 'link',
           title: t('about', { appName: buildConfig().appName }),
           onPress: () => {
-            const aboutUrl = buildConfig().aboutUrls[languageCode] || buildConfig().aboutUrls.default
+            const aboutUrls = buildConfig().aboutUrls
+            const aboutUrl = aboutUrls[languageCode] || aboutUrls.default
             Linking.openURL(aboutUrl)
           }
         },
--- a/native/src/routes/settings/createSettingsSections.js
+++ b/native/src/routes/settings/createSettingsSections.js
@@ -84,11 +84,8 @@ const createSettingsSections = ({ setSetting, t, languageCode, cityCode }: Creat
           accessibilityRole: 'link',
           title: t('about', { appName: buildConfig().appName }),
           onPress: () => {
-            if (languageCode === 'de') {
-              Linking.openURL('https://integreat-app.de/about/')
-            } else {
-              Linking.openURL('https://integreat-app.de/en/about/')
-            }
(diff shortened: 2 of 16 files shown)


web/src/routes/main-disclaimer/components/MainDisclaimerPage.js
```diff
@@ -2,82 +2,17 @@
...
 import React from 'react'
 import Caption from '../../../modules/common/components/Caption'
+import buildConfig from '../../../modules/app/constants/buildConfig'
 
 type PropsType = {||}
 
 class MainDisclaimerPage extends React.Component<PropsType> {
   render () {
     return (
+
```
Taggotty, line 12: Remove newline

native/tools/build-config
```diff
@@ -19,7 +19,11 @@ const loadBuildConfigAsXCConfig = buildConfigName => {
     delimiter: '_',
     transformKey: key => decamelize(key).toUpperCase()
   })
-  const prefixed = Object.keys(xcconfigOptions).map(key => `BUILD_CONFIG_${key} = ${xcconfigOptions[key]}`)
+  const prefixed = Object.keys(xcconfigOptions).map(key => {
+    const value = xcconfigOptions[key]
+    const escaped = typeof xcconfigOptions[key] === 'string' ? value.replace(/\n/g, '\\n') : value
```
Taggotty, line 24: why do we need to do that?

Taggotty (approved):

native/tools/build-config
klinzo: Now specifically: Because of the imprint (which has multiple line breaks)
In general: xcconfig files are key value files with one key/value per line. If there are linebreaks in the value, the next line is treated as another key value pair which it obviously is not, which leads to errors.

build-configs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "build-configs",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
maxammann, line 3: Do we really need to push the version here always? Is this because it is not symlinked?

native/src/routes/settings/createSettingsSections.js
```diff
@@ -84,11 +84,8 @@ const createSettingsSections = ({ setSetting, t, languageCode, cityCode }: Creat
           accessibilityRole: 'link',
           title: t('about', { appName: buildConfig().appName }),
           onPress: () => {
-            if (languageCode === 'de') {
-              Linking.openURL('https://integreat-app.de/about/')
-            } else {
-              Linking.openURL('https://integreat-app.de/en/about/')
-            }
+            const aboutUrl = buildConfig().aboutUrls[languageCode] || buildConfig().aboutUrls.default
```
maxammann, line 87:
```suggestion
            const aboutUrls = buildConfig().aboutUrls
            const aboutUrl = aboutUrls[languageCode] || aboutUrls.default
```

native/tools/build-config
maxammann: isn't `typeof xcconfigOptions[key] === 'string'` always true?

maxammann (approved):

native/tools/build-config
klinzo: No, there are e.g. boolean values as well (feature flags for example).

build-configs/package.json
klinzo: Yes, probably because of missing symlinks.

native/tools/build-config
maxammann: Maybe use lodashs function then. I dont unterstand the implementation but probably it is better: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L12135

PR #69, opened 2020-11-13
IGAPP-374: Enable coverage results

maxammann:


.circleci/src/jobs/check_native.yml
```diff
@@ -10,13 +10,13 @@ steps:
   - checkout
   - check_config
   - restore_yarn_cache
-  - run:
-      name: Linting check
-      command: yarn workspace native lint
+  - lint:
+      directory: native
+  - store_test_results:
+      path: native/reports
```
klinzo, line 16: Why don't you include this in the command `lint` as well?

.circleci/src/jobs/check_web.yml
```diff
@@ -10,16 +10,15 @@ steps:
...
-      command: yarn workspace web test --ci --maxWorkers ${TOTAL_CPUS}
-  - run:
-      name: Test coverage # Results are not used right now. Run it anyway to (possibly) discover more problems.
-      command: yarn workspace web test:coverage --ci --maxWorkers ${TOTAL_CPUS}
+  - unit_test:
+      directory: web
+  - lint:
+      directory: api-client
+  - unit_test:
+      directory: api-client
```
klinzo, line 23: I'd rather just include api-client tests in both the web and native test suite directly via the roots config option: https://jestjs.io/docs/en/configuration.html#roots-arraystring

Advantages:
* No need to have jest setup files, configs scripts and dependencies in the api-client
* All relevant tests are executed when I run native/web tests locally
* The api-client tests belong to both or neither to web or native -> I don't see a reason why we'd execute api-client tests here but not in native
* Only babel-runtime is required in the api-client package

native/package.json
```diff
@@ -20,11 +20,13 @@
...
     "test": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js",
-    "test:coverage": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js --coverage",
+    "test:coverage": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js --coverage",   
+    "test:ci": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js --coverage --ci --testResultsProcessor=jest-junit",
     "test:watch": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js --watchAll",
     "test:update": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' jest --config ./jest.config.js -u",
     "test:e2e": "cross-env BUILD_CONFIG_NAME='integreat-e2e' jest --config jest.e2e.config.js --maxWorkers=4",
-    "lint": "eslint .",
+    "lint": "yarn run eslint",
+    "eslint": "eslint .",
```
klinzo, line 29: Why do we need two commands for this?

.circleci/src/jobs/check_web.yml
maxammann: I'm not a 100% sure what is the best solution. It could also be strange to execute the tests for the api-client with two different babel configs. But maybe we can not avoid this because metro and webpack will use different babel-configs nonetheless.

I'm also not sure which flow-config is used in the api-client. Again there are two flow-configs available and there could be differences.

maxammann: But I would be fine with your solution. I just dislike a bit how we manage a monorepo. The packages in there are not really packages which could life on its own. @Schedulaar what do you think?

native/package.json
maxammann: Because I want to have the "eslint" command available in every script. In the webapp "lint" executes eslint and stylelint.

.circleci/src/jobs/check_web.yml
Schedulaar: My gut feeling tells me, that it would be better to have a separate test environment for the api client.
Probably that feeling is based upon the benefits of modularity (i.e. if the api client is a separate package, we should treat it like that)

Also usually, if I run the tests for the webapp, because I changed something there, I'm not interested in the tests of the api client; vice versa, if I only make changes to the api client, I only want to run the tests for the api client.
(Imo, it makes sense to apply changes only to the api client: E.g. if new endpoints are introduced, there should probably be an issue first, that only adds the endpoint (and corresponding tests) to the api-client)

But this is nevertheless just a feeling :D Of course, as there are different babel and flow configs in web- and native app, testing the api-client with both gives the benefit of testing these configuration / glue code as well, but the tests in the api-client should not depend on that (optimally :sweat_smile: )

klinzo: I guess then we should have a separate circle ci job for api-client test though :thinking: Also, how does flow and linting fit into this? Should be made api-client specific as well then, right?

maxammann: I think when going the route of treating the api-client as separate module then we loose the benefits of a monorepo. This kind of sucks but I think we should go the easier route which probably works "well enought"®

klinzo: @maxammann which is the easier route which works well enough for you?

maxammann: Oke, there is the first problem with not handling module separation seriously.. I'm unable to collect the coverage results from the api-client because Jest only looks at the coverage results in the `rootDir` and not `roots`

maxammann: @klinzo I think this is a topic worth discussing, there are pro and cons on both sides

* Option A: No separation -> easy and hacky
* Option B: Total separation -> very hard to make it usable and not stress developers

I'd love to find a solution inbetween

web/package.json
```diff
@@ -14,12 +14,15 @@
...
     "build:integreat-test-cms": "webpack --config tools/webpack.config.babel.js --env.config_name=integreat-test-cms",
     "check:build": "yarn es-check es5 ./dist/**/*.js --verbose",
-    "test": "jest --config jest.config.json",
-    "test:coverage": "jest --config jest.config.json --coverage",
-    "test:watch": "jest --config jest.config.json --watchAll",
-    "test:update": "jest --config jest.config.json -u",
-    "lint": "eslint . && yarn run stylelint",
+    "test": "jest --config jest.config.js",
+    "test:coverage": "jest --config jest.config.js --coverage",
+    "test:ci": "yarn run test:coverage --ci",
```
Taggotty, line 19: what does the `--ci` do?

web/jest.config.js
```diff
@@ -0,0 +1,34 @@
...
+    hashids: 'hashids/dist/hashids.min.js'
+  },
+  moduleFileExtensions: [
+    'js',
+    'json'
+  ],
+  moduleDirectories: [
+    'node_modules'
+  ],
+  coverageDirectory: '<rootDir>/../reports/coverage',
```
Taggotty, line 24: so there is a shared coverage directory for web and native? would it make sense to separate it to `'<rootDir>/../reports/web/coverage'` and `'<rootDir>/../reports/native/coverage'`?

native/jest.config.js
```diff
@@ -14,6 +14,8 @@ const transformNodeModules = [
 ]
 
 module.exports = {
+  rootDir: 'src',
+  roots: ['<rootDir>', '<rootDir>/../../api-client/src'],
```
Taggotty, line 18: This means that jest runs both the native and the api clients tests, right? Does that mean in the ci the api client is executed twice (once for web and once for native)?

Taggotty (approved):

.circleci/src/jobs/check_web.yml
klinzo: Should we discuss this in the call today?

native/jest.setup.js
```diff
@@ -12,8 +14,12 @@ if (typeof window !== 'object') {
...
 jest.mock('rn-fetch-blob')
 
-// FormData is not part of the jest react-native preset: https://github.com/jefflau/jest-fetch-mock/issues/23
-global.FormData = require('testing/FormDataMock.js')
+// We polyfill FormData here because react-native uses 'node' as 'testEnvironment' option in jest:
+// https://jestjs.io/docs/en/configuration#testenvironment-string
+// Importing it from jsdom allows us to import stuff selectively
+const jsdom = new JSDOM()
+const { FormData } = jsdom.window
+global.FormData = FormData
```
klinzo, line 22: Good solution. I think you can remove my custom mock then.

web/tools/webpack.config.babel.js
```diff
@@ -63,6 +64,7 @@ const createConfig = (env = {}) => {
   const wwwDirectory = path.resolve(__dirname, '../www')
   const distDirectory = path.resolve(__dirname, `../dist/${buildConfigName}`)
   const srcDirectory = path.resolve(__dirname, '../src')
+  const bundleReportDirectory = path.resolve(__dirname, '../reports/bundle')
```
klinzo, line 67: This is written to `web/reports/bundle`. Other reports are written to the rootDir, aren't they? Should we make this consistent, i.e. only rootdir/reports or rootDir/web/reports and rootDir/native/reports?

klinzo (approved):

.circleci/src/jobs/check_web.yml
maxammann: We will use a separate config

web/tools/webpack.config.babel.js
maxammann: Nothing is written to the root directory of the project. Reports are in web/reports and native/reports

web/jest.config.js
maxammann: No there is no shared directory. rootDir is src/

web/package.json
maxammann: Disables generation of snapshots and throws an error afaik

PR #70, opened 2020-11-15
IGAPP-368: Bundle size to big

klinzo:

The max bundle size has to be increased since the api-client package in the monorepo than the built version from npm was before. This is because previously the files were built, which does not happen anymore now. The api-client package has now a size of 100.93 KiB in comparison to 33.58 KiB before. The increase of the budget by 0.6 MiB is enough since the budget wasn't used completely before.

### DIFF
--- a/web/tools/webpack.config.babel.js
+++ b/web/tools/webpack.config.babel.js
@@ -22,7 +22,7 @@ const SHORT_COMMIT_SHA_LENGTH = 8
 
 // A first performance budget, which should be improved in the future: Maximum bundle size in Bytes; 2^20 = 1 MiB
 // eslint-disable-next-line no-magic-numbers
-const MAX_BUNDLE_SIZE = 1.56 * Math.pow(2, 20)
+const MAX_BUNDLE_SIZE = 1.62 * Math.pow(2, 20)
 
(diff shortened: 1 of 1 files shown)


maxammann:
Hm this should not be neccassary. Can you rebase on this branch: https://github.com/Integreat/integreat-app/pull/69

I fixed there how webpack bundles js.

maxammann:
webpack should bundle and minify the api-client like before. This is very strange.

klinzo:
> Hm this should not be neccassary. Can you rebase on this branch: #69
> 
> I fixed there how webpack bundles js.

Also not working, size of the bundle is still 1.62 MiB. You can try yourself, just execute `yarn build` or `yarn build:malte`

maxammann:
I think this is fine, the reason for this is maybe that we excluded all dependencies in the build of "integreat-api-client". Now we have all depenencies in the yarn.lock and webpack is propably packaging something twice. this is fine though as code is correct and this is just how webpack works and our dependencies are.

maxammann (approved):

klinzo:
@maxammann should I close this as you did it in your PR already?

maxammann:
Yeah I think it is fine, you aswell, so yeah :D

PR #72, opened 2020-11-15
IGAPP-364: Prevent passing of other platforms build config values

klinzo:


klinzo:
Any suggestions how this could be made better? I don't like the current implementation to be honest. Dimensions in the themes are also still not platform specific and will add even more complexity :/

maxammann:
I changed the title. It just didn't tell me anything :smile:

build-configs/integreat-e2e/index.js
```diff
@@ -14,18 +18,36 @@ const IntegreatE2eBuildConfig: BuildConfigType = {
...
-  ios: {
-    ...IntegreatTestCmsBuildConfig.ios,
-    bundleIdentifier: 'de.integreat-app',
-    provisioningProfileSpecifier: 'match Development de.integreat-app',
-    googleServices: null
   }
 }
 
-export default IntegreatE2eBuildConfig
+const CommonIntegreatE2eBuildConfig: CommonBuildConfigType = {
```
maxammann, line 24: I think these should not be capitalized as they are just variables.

```suggestion
const commonIntegreatE2eBuildConfig: CommonBuildConfigType = {
```

native/tools/build-config
```diff
@@ -57,10 +57,10 @@ program
   })
 
 program
-  .command('to-json <build_config_name>')
+  .command('to-json <build_config_name> <platform>')
   .description('outputs the specified build config as JSON')
-  .action(async buildConfigName => {
-    const buildConfig = loadBuildConfig(buildConfigName)
+  .action((buildConfigName, platform) => {
+    const buildConfig = loadBuildConfig(buildConfigName, platform)
```
maxammann, line 63: Does the programm exit here with exit code 1 if an exception is thrown?

web/tools/webpack.config.babel.js
```diff
@@ -15,14 +15,15 @@ const MomentLocalesPlugin = require('moment-locales-webpack-plugin')
...
+const { WEB } = require('build-configs')
 
 const currentYear = new Date().getFullYear()
 
 const SHORT_COMMIT_SHA_LENGTH = 8
 
 // A first performance budget, which should be improved in the future: Maximum bundle size in Bytes; 2^20 = 1 MiB
 // eslint-disable-next-line no-magic-numbers
-const MAX_BUNDLE_SIZE = 1.56 * Math.pow(2, 20)
+const MAX_BUNDLE_SIZE = 1.62 * Math.pow(2, 20)
```
maxammann, line 26: Not sure whether we want this, see other PR with same changes.

maxammann (approved):
Good work! Look s nice and clean

klinzo:
@maxammann what about theme dimensions? Problem here is the following: Currently in react-native js I just use the common build config which is enough. However, if dimensions are only passed for native or web, either we need a `common native` build config or we have to load android or ios build config depending on the platform.

web/tools/webpack.config.babel.js
klinzo: Yes, not sure why this shows up here... Will revert depending on other PR

maxammann:
~~I think the question is whether "dimensions" are a property of a specific device brand (ios/android) or whether they are a property of "web vs native".  ~~

Actually I think that dimensions should not be part of `theme`. There are no platform differences for the dimensions for ios/android. Also no build-config chooses different dimensions. I think a theme should look like this:

```
export const lightTheme: ThemeType = {
  colors: matleColors,
  fonts: malteFonts,
}

```

`helpers` can be removed from the theme and the build config and can be added to the native js project hardcoded. Same for `dimensions`

maxammann:
If we really have to share configuration between ios and android then I would go with the option of having to load android or ios build config depending on the platform. That option allows more freedom of configuration.

Znippet:
@klinzo:

It think i repaired "yarn android" unter windows on my machine. This PR concerns the Configs, so I tried it on this branch. Could you please confirm if this is really an error in the scripts (error happens on windows):

AndroidManifest.xml:17:5-27:16: AAPT: error: resource string/BUILD_CONFIG_APP_NAME (aka tueruer.app.integreat:string/BUILD_CONFIG_APP_NAME) not found.

In the list of the variables in the output of buildConfigs.gradle in line 45 the key "BUILD_CONFIG_APP_NAME" does indeed not exist
There is only:
BUILD_CONFIG_APPNAME = IntegreatTestCms 
and
BUILD_CONFIG_NAME = integreat-test-cms

I created a new Task, we can solve the problem also in a new branch: https://issues.tuerantuer.org/browse/IGAPP-375

maxammann:
I suppose this is just a simple typo somewhere

Taggotty (approved):

native/tools/build-config
klinzo: Yes, there is no async function here :)

PR #73, opened 2020-11-18
IGAPP-342: Fix sharing of PDFs and Images

Taggotty:


native/src/modules/layout/components/TransparentHeader.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react'
-import Share from 'react-native-share'
+import { Share } from 'react-native'
```
klinzo, line 4: Can we remove `react-native-share` then?

klinzo (approved):
Though I don't think this is because of this PR, language change does not change the share url on the dashboard. I.e. if you change language from `de` to `en`, the share url is still https://integreat.app/altmuehlfranken/de`. Could you look into this if this easily fixable? Otherwise I'll just open a new issue.

Taggotty:
> Though I don't think this is because of this PR, language change does not change the share url on the dashboard. I.e. if you change language from `de` to `en`, the share url is still [https://integreat.app/altmuehlfranken/de`](https://integreat.app/altmuehlfranken/de%60). Could you look into this if this easily fixable? Otherwise I'll just open a new issue.

The solution for that should not be that difficult, but I still struggled to make this work today. I'd prefer if you'd create a separate issue for this.

maxammann (approved):
Tested on Android

maxammann:
@Taggotty @klinzo did anyone of you create the task?

PR #75, opened 2020-11-18
IGAPP-209: Fix state update warning in withPayloadProvider

klinzo:

### DIFF
--- a/native/src/modules/endpoint/hocs/withPayloadProvider.js
+++ b/native/src/modules/endpoint/hocs/withPayloadProvider.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as React from 'react'
+import { useEffect, useState } from 'react'
 import { LanguageModel } from 'api-client'
 import { RefreshControl, ScrollView, View } from 'react-native'
 import LanguageNotAvailableContainer from '../../common/containers/LanguageNotAvailableContainer'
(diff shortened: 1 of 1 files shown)


native/src/modules/endpoint/hocs/withPayloadProvider.js
```diff
@@ -60,11 +60,18 @@ const withPayloadProvider = <S: { dispatch: Dispatch<StoreActionType> }, R: {}>
   return (Component: React.ComponentType<S>): React.ComponentType<PropsType<S, R>> => {
     return class extends React.Component<PropsType<S, R>, {| timeoutExpired: boolean |}> {
       static displayName = wrapDisplayName(Component, 'withPayloadProvider')
+      timeout = null
 
       state = { timeoutExpired: false }
 
       componentDidMount () {
-        setTimeout(() => this.setState({ timeoutExpired: true }), LOADING_TIMEOUT)
+        this.timeout = setTimeout(() => this.setState({ timeoutExpired: true }), LOADING_TIMEOUT)
```
Taggotty, line 68: wouldn't it make more sense if `timeout` was part of the state?

maxammann, line 68: I propose to use a hook here. It offers functionality for cleaning up: 

```
  useEffect(() => {
    const timer = setTimeout(() => {
      setCount('Timeout called!');
    }, 1000);
    return () => clearTimeout(timer);
  }, []);
```

maxammann (approved):

native/src/modules/endpoint/hocs/withPayloadProvider.js
klinzo: I kind of was expecting this :D However, to use a hook, we'd have to convert this to a functional component which I was currently not willing to do as I think this would have needed quite a lot of other changes.

klinzo: But definitely, with a hook it would be way cleaner :+1:

maxammann: That is a task of 3 minutes usually :) I think we should use new framework features if they make sense.

maxammann: Class Components are not used anymore in modern react

Taggotty (approved):

maxammann (approved):

PR #78, opened 2020-11-19
IGAPP-375: Fix babel transpile error in decamelize npm

Znippet:

### DIFF
--- a/native/babel.config.js
+++ b/native/babel.config.js
@@ -11,6 +11,8 @@ module.exports = {
       plugins: ['transform-remove-console']
     }
   },
+  // prevent babel error - only windows (using backslash)
+  exclude: [/node_modules\\decamelize.*/],
   plugins: [
     [
       'transform-inline-environment-variables',


Znippet:
@Taggotty: Please confirm

native/babel.config.js
```diff
@@ -11,6 +11,8 @@ module.exports = {
       plugins: ['transform-remove-console']
     }
   },
+  // prevent babel error - only windows (using backslash)
+  exclude: [/node_modules\\decamelize.*/],
```
Znippet, line 15: Instead of excluding one package we could include only the packages babel should transpile. This would lead to faster transpilation for the whole project

klinzo (approved):

Znippet:
> I dont really get which issue this exactly is fixing. Is it fixing `yarn android:integreat-test-cms`? because you say in the ticket that deleting the cache directory fixed this.

This analysis was a mistake, I will fix the description. I commented out a "throw error" in the node_modules and yarn android compiled. So I commented it in again, but yarn seems to have used cached code, it compiled without the "throw error"
I will correct what I wrote in the task, good hint

native/babel.config.js
maxammann: I'm not sure whether that would work with metro. I think React Native packages do not always transpile properly. So babel is kind of required on the node_modules directory. Did you try it?

maxammann (approved):

native/babel.config.js
Znippet: I tried my solution and it works.
If we would replace it with "includes" instead of the "exclude" then we have to try it out. 
So it is by far more work and maybe we still would have to include many packages to the babel transpiler

Taggotty:
> @Taggotty: Please confirm

Tested on my system. Works fine :) Nice work.

PR #81, opened 2020-11-23
IGAPP-19: Setup Sentry

maxammann:


native/package.json
```diff
@@ -42,14 +42,14 @@
...
-    "build-configs": "file:../build-configs",
+    "api-client": "7.0.0",
+    "build-configs": "2.0.0",
     "hashids": "^2.2.1",
     "htmlparser2": "^4.0.0",
     "i18next": "^19.1.0",
     "is-url": "^1.2.4",
     "js-md5": "^0.7.3",
-    "locales": "file:../locales",
+    "locales": "1.0.0",
```
maxammann, line 52: I did this task within here as I already started with it: https://issues.tuerantuer.org/browse/IGAPP-366

klinzo: Do we still have to bump versions? Otherwise it's probably better to just set all monorepo package versions to `0.0.1` or something.

web/tools/webpack.config.babel.js
```diff
@@ -143,6 +144,7 @@ const createConfig = (env = {}) => {
       new webpack.DefinePlugin({
         'process.env.NODE_ENV': NODE_ENV,
         __VERSION__: JSON.stringify(version),
+        __BUILD_CONFIG_NAME__: JSON.stringify(buildConfigName),
```
klinzo, line 147: What do we need that for?

web/src/modules/app/initSentry.js
```diff
@@ -0,0 +1,27 @@
...
+
+  try {
+    const Sentry = await import(
+      /* webpackChunkName: "sentry" */
+      '@sentry/react'
+    )
+
+    Sentry.init({
+      dsn: 'https://f07e705b25464bbd8b0dbbc0a6414b11@sentry.tuerantuer.org/2',
+      release: `${__BUILD_CONFIG_NAME__}@${__VERSION__}`
```
klinzo, line 19: I wonder if it wouldn't be better to use the `appName` from the build config instead of the name.

klinzo: I also guess that the version won't be correct here. You will have to pass it from the environment variables by CircleCI to the webpack config.

tools/sentry-release
```diff
@@ -0,0 +1,151 @@
+#!/bin/env bash
```
klinzo, line 1: Does this have to be reviewed? :pray:

web/babel.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   presets: [
-    ['@babel/preset-env', { modules: 'commonjs' }],
+    ['@babel/preset-env', { modules: 'auto', loose: true }],
```
Taggotty, line 3: changed it to `commonjs` in [this issue](https://issues.tuerantuer.org/browse/WEBAPP-525). @Znippet would you mind checking out whether the webapp still compiles after this change?

Taggotty (approved):

tools/sentry-release
maxammann: only the lower part

web/babel.config.js
maxammann: Using "auto" here is at least the "correct" way of doing this here :) That means webpack decides what it wants to have

web/src/modules/app/initSentry.js
maxammann: The thing here is that the appName does not have to stay constant. appName describes only the name of the app on the launch screens on phones afaik.

tools/sentry-release
```diff
@@ -0,0 +1,151 @@
...
+if [ -z "$SENTRY_AUTH_TOKEN" ]
+then
+  echo "SENTRY_AUTH_TOKEN is not set"
+  exit 1
+fi
+
+version="${package_name}@${version_name}"
+
+
+if [ ! -z "$version_code" ]
```
klinzo, line 141: I think there should be a version code all the time as well.

klinzo (approved):

web/src/modules/app/initSentry.js
klinzo: It is also used for translations. I don't think we will and should ever change this. I see your point though.

tools/sentry-release
maxammann: I excluded the version_code from the version in the webapp as it has no valuable information

In our projects version_name = version_code

web/src/modules/app/initSentry.js
maxammann: That is alright though. Or at least it does not harm anything.

If we decide to show "Integreat - Great Integreation" on the launch screens, then it would be okish to include that in the translations.

klinzo (approved):

PR #82, opened 2020-11-23
IGAPP-388: Replace the development flag with a logging feature flag

maxammann:


web/src/modules/layout/components/Footer.js
```diff
@@ -38,10 +37,7 @@ const FooterContainer = styled.footer`
  */
 class Footer extends React.PureComponent<PropsType> {
   static getVersion (): Node {
-    if (buildConfig().development) {
-      return <span>{__VERSION__}</span>
-    }
-    return null
+    return <span>{__VERSION__}</span>
```
maxammann, line 40: I think we can show it always. I kind of this it is nice to know this in production aswell

web/tools/webpack.config.babel.js
```diff
@@ -1,3 +1,5 @@
+// @flow
```
maxammann, line 1: Lets flow type everything! :)

native/src/routes/landing/components/CitySelector.js
```diff
@@ -60,8 +60,6 @@ class CitySelector extends React.PureComponent<PropsType> {
     const cities = this.props.cities
     if (normalizedFilter === 'wirschaffendas') {
       return cities.filter(_city => !_city.live)
-    } else if (buildConfig().development) {
```
maxammann, line 63: Should we add a new flag for this? Not sure how to handle this. I think we can remove it actually.

Not sure how to check whether metro is running as server or whether `react-native bundle` has been called.

klinzo: I think it was nice to have this, but I guess it isn't a real problem if the effort is to high as you can just enter 'wirschaffendas' as well. I really wouldn't mind just using the logging flag with a comment here as well... :innocent: I mean, we just log all cities... to the screen... visible to the user... :grin:

web/src/modules/layout/components/Footer.js
klinzo: Did you check how it looks on mobile? To be honest I'm not sure if that is really a good idea, I don't remember seeing this on other sites.

klinzo: This is just a unnecessary information for all our users and only useful for us, so I don't think we should show it.

web/tools/webpack.config.babel.js
```diff
@@ -154,14 +165,21 @@ const createConfig = (env = {}) => {
         filename: 'assets.json',
         prettyPrint: true
       }),
+      // $FlowFixMe Unable to find "optimize" in "webpack"
```
klinzo, line 168: No `optimize` used here, the fixme should be change I think

web/src/modules/layout/components/Footer.js
klinzo: Apart from that the version is not correct if the version got bumped before. The version bump is just made in git and not to the files in the CI, but we just read the version from the files in the webpack config. So if there was a version bump and we create a new production release, it shows an old version (the previous one).

Taggotty (approved):

native/src/routes/landing/components/CitySelector.js
Taggotty: I also think this was a quite nice feature, but I agree with @klinzo here. No need to put in too much effort.

maxammann:
I switched froom loggin to a more generic developerFriendly feature flag

klinzo (approved):

PR #84, opened 2020-11-25
IGAPP-400: SyntaxError: The keyword 'const' is reserved

klinzo:

Background: Webpack copies all files in the assets folder to the dist folder in order to enable delivery to the server. After building the app the generated java script files are checked for invalid js syntax (i.e. es6 syntax). If some is found, the `check:build` step errors which happened during our earlier production release, because the `mainImprint.js` which is used in the build configs contains such syntax. So I just moved this to a different folder. If you have any other idea how we could/should define what needs to be copied and what not feel free to suggest :) The only way is probably to differ between web and native assets again


klinzo:
Made draft again because base branch is not uptodate :)

SimonVortkamp (approved):

maxammann (approved):
I thin this is the correct was. mainImprint is not an asset in that sense that it is delivered "as-is" in the fontend. It is included in the build and therefore not really a static asset.

PR #85, opened 2020-11-26
IGAPP-107: Add Feedback to Disclaimer and Sprungbrett

Taggotty:


native/package.json
```diff
@@ -38,6 +38,7 @@
     "@react-native-community/async-storage": "^1.9.0",
     "@react-native-community/geolocation": "^2.0.2",
     "@react-native-community/netinfo": "^5.9.7",
+    "@react-native-community/picker": "^1.8.1",
```
Taggotty, line 41: Fixes IGAPP-348

native/package.json
```diff
@@ -60,7 +61,7 @@
     "react": "16.13.1",
     "react-i18next": "^11.3.1",
     "react-native": "0.63.3",
-    "react-native-elements": "^1.2.7",
+    "react-native-elements": "^2.0.3",
```
Taggotty, line 64: Fixes IGAPP-349

api-client/src/errors/ResponseError.js
```diff
@@ -33,10 +33,12 @@ class ResponseError extends Error {
   }
 
   createMessage ({ requestOptions, url, endpointName, response }: ResponseErrorParamsType): string {
+    // This is a react-native FormData
```
klinzo, line 36: Does this still work for web?

native/src/modules/categories/components/Categories.js
```diff
@@ -60,34 +55,19 @@ class Categories extends React.Component<PropsType> {
...
+      title: !category.isRoot() ? category.title : undefined,
+      path: !category.isRoot() ? category.path : undefined,
+      isPositiveFeedback
     }
 
-    navigation.navigate('FeedbackModal', {
-      isPositiveFeedback,
-      feedbackItems
-    })
+    navigation.navigate('FeedbackModal', { ...feedbackInformation })
```
klinzo, line 70: Why don't you use `createNavigateToFeedbackModal` here?

native/src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -144,7 +144,7 @@ const ThemedTranslatedDashboard = withTranslation('dashboard')(
   withTheme(Dashboard)
 )
 
-const DashboardContainer = (props: ContainerPropsType) => {
+const DashboardContainer = (props: ContainerPropsType, state) => {
```
klinzo, line 147: Missing type annotation

native/src/routes/disclaimer/DisclaimerContainer.js
```diff
@@ -53,6 +55,19 @@ class DisclaimerContainer extends React.Component<DisclaimerPropsType, Disclaime
...
+    const { navigation, city, language } = this.props
+    const feedbackInformation: FeedbackInformationType = {
+      type: 'Disclaimer',
+      cityCode: city,
+      language,
+      isPositiveFeedback,
+      path: this.state.disclaimer?.path
+    }
+
+    navigation.navigate('FeedbackModal', { ...feedbackInformation })
```
klinzo, line 68: Same as above.

native/src/routes/disclaimer/DisclaimerContainer.js
```diff
@@ -53,6 +55,19 @@ class DisclaimerContainer extends React.Component<DisclaimerPropsType, Disclaime
     this.state = { disclaimer: null, error: null, timeoutExpired: false }
   }
 
+  navigateToFeedback = (isPositiveFeedback: boolean) => {
```
klinzo, line 58: Nice and clean :+1:

native/src/routes/events/components/Events.js
```diff
@@ -63,39 +57,31 @@ class Events extends React.Component<PropsType> {
...
-      feedbackItems: [
-        createFeedbackVariant(t('feedback:contentOfCity', { city: cityTitle }), EVENTS_FEEDBACK_TYPE,
-          CONTENT_FEEDBACK_CATEGORY),
-        createFeedbackVariant(t('feedback:technicalTopics'), CATEGORIES_FEEDBACK_TYPE, TECHNICAL_FEEDBACK_CATEGORY)
-      ]
-    })
+    const { navigation, cityCode, language } = this.props
+
+    const feedbackInformation: FeedbackInformationType = {
+      type: 'Event',
```
klinzo, line 78: Is the type `Event` correct here for multiple events?

klinzo: I think it should be `Events`, right? https://github.com/Integreat/cms/wiki/REST-APIv3-Documentation#events

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+type FeedbackType = 'Category' | 'Event' | 'Pois' | 'Offers' | 'Search' | 'Disclaimer'
+
+export type FeedbackInformationType = {
+  type: FeedbackType,
+  isPositiveFeedback: boolean,
+  language: string,
+  cityCode?: string,
+  path?: string,
+  title?: string,
+  feedbackAlias?: string,
```
klinzo, line 37: Is this for offers? Is it correctly named `feedbackAlias`?

native/src/routes/wohnen/containers/WohnenOfferContainer.js
```diff
@@ -14,23 +14,24 @@ import type { NavigationStackProp } from 'react-navigation-stack'
 import FailureContainer from '../../../modules/error/containers/FailureContainer'
```
klinzo, line 14: This would not have been necessary, we generally don't want to put any effort in wohnen anymore

web/src/modules/feedback/FeedbackInformation.js
```diff
@@ -0,0 +1,26 @@
+// @flow
```
klinzo, line 1: What is the difference to FeedbackVariant/Why do we need both?

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+  constructor (props: ContainerPropsType) {
+    super(props)
+    const feedbackOptions = this.getFeedbackOptions()
+    this.state = { feedbackOptions, selectedFeedbackIndex: 0, comment: '', sendingStatus: 'idle' }
+  }
+
+  getCityName = (): string => {
+    const { cities, navigation } = this.props
+    const cityCode = navigation.getParam('cityCode')
+    return cityCode ? CityModel.findCityName(cities, cityCode) : INTEGREAT_INSTANCE
```
klinzo, line 112: How does this work with Malte? Feedback is probably sent to Integreat as well, right? Maybe this is okay though because it should never happen, right?

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+    if (currentPageFeedbackOption) {
+      options.push(currentPageFeedbackOption)
+    }
+
+    const contentFeedbackOption = this.getContentFeedbackOption()
+    if (contentFeedbackOption) {
+      options.push(contentFeedbackOption)
+    }
+
+    this.getOffersFeedbackOptions().forEach(option => options.push(option))
```
klinzo, line 127: I think push also works for multiple options and empty arrays: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push.
```suggestion
    options.push(...this.getOffersFeedbackOptions())
```

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+      feedbackType: CATEGORIES_FEEDBACK_TYPE,
+      feedbackCategory: TECHNICAL_FEEDBACK_CATEGORY
+    })
+  }
+
+  getContentFeedbackOption = (): ?FeedbackVariant => {
+    const { navigation, t } = this.props
+    const cityName = this.getCityName()
+
+    if (cityName) {
```
klinzo, line 145: I think this if is not required as it always returns a string, doesn't it?

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+      const label = t('contentOfCity', { city: cityName })
+      const feedbackCategory = CONTENT_FEEDBACK_CATEGORY
+      const feedbackType = navigation.getParam('type')
+      switch (feedbackType) {
```
klinzo, line 149: ~~I think if you pass the right feedback type directly you could get rid of this, don't you?~~
Probably I am wrong. However, I think you have duplicated `EVENTS` feedback if you are at the events route. Not sure about that either though.

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+
+  handleSubmit = async () => {
+    const { selectedFeedbackIndex, feedbackOptions, comment } = this.state
+    const feedbackItem = feedbackOptions[selectedFeedbackIndex]
+    const feedbackData = this.getFeedbackData(feedbackItem, comment)
+    this.setState({ sendingStatus: 'sending' })
+    try {
+      const apiUrl = await determineApiUrl()
+      const feedbackEndpoint = createFeedbackEndpoint(apiUrl)
+      console.log(feedbackEndpoint.request)
```
klinzo, line 258: Could be removed
```suggestion

```

klinzo (commented):
Nicely done, very clean (apart from the FeedbackModalContainer but that is mostly the fault of the api and too much cases I guess)!

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,296 @@
...
+        feedbackType: PAGE_FEEDBACK_TYPE,
+        feedbackCategory
+      })
+    } else if (type === 'Offers' && feedbackAlias && title) {
+      return new FeedbackVariant({
+        label: t('contentOfOffer', { offer: title }),
+        feedbackType: OFFER_FEEDBACK_TYPE,
+        feedbackCategory
+      })
+    } else if (type === 'Search' && query) {
```
Taggotty, line 207: Seems that this endpoint is never called. Same goes for web (as this part is just copy pasted). Will remove this.

native/package.json
maxammann: I closed the issue

native/package.json
maxammann: I closed the issue

maxammann (approved):

api-client/src/errors/ResponseError.js
Taggotty: No it actually does not. Any ideas?

native/src/routes/events/components/Events.js
Taggotty: the type here is not the type that is send to the feedback endpoint. It is just loosely used instead of CATEGORIES_ROUTE, EVENT_ROUTE, ... in web

native/src/routes/feedback/containers/FeedbackModalContainer.js
Taggotty: Yes it is used for offers. Renamed to alias.

native/src/routes/feedback/containers/FeedbackModalContainer.js
Taggotty: This should actually never happen. The only problem I see here if cities are not available. But maybe you can tell me how that is. Do I need to check whether the cities are available in refresh and then force refresh if they are not? As far as I understood the cities are loaded in the LandingContainer and should be in the redux store all along.

native/src/routes/wohnen/containers/WohnenOfferContainer.js
Taggotty: it was more or less just copy pasting stuff. But if you think it's better to remove it to avoid future errors resulting from the "new" code I will do so :)

web/src/modules/feedback/FeedbackInformation.js
Taggotty: aactually something must have happened here. I thought I replaved this FeedbackInformation file with the FeedbackVariant again. But to explain the current structure: 
The Pages, Categories, Offers, ... etc navigate to the FeedbackModal with the FeedbackInformation. There the feedbackVariants are created depending on the properties of the FeedbackInformation. And finally the FeedbackParams are created when a specific feedbackVariant was selected and submitted.

native/src/routes/feedback/containers/FeedbackModalContainer.js
klinzo: Yes, understood correctly :+1: There should be no way to end up here with no cities or no city with the code as cities are fetched beforehand and one cannot select feedback if no city is found. So maybe throwing an error would be the better (more explicit) variant.

native/src/routes/wohnen/containers/WohnenOfferContainer.js
klinzo: No, the work is already done :D Putting even more work into it to remove it again would not make sense imo :grin:

api-client/src/errors/ResponseError.js
klinzo: Without changing this feedback does not work for native? Or what was the reason for this?

web/src/modules/feedback/FeedbackInformation.js
klinzo: So this has to be changed? Or is this correct/necessary now? Not sure if I understand the first part of your explanation correctly. Thanks!

native/src/routes/feedback/containers/FeedbackModalContainer.js
klinzo: You mean it is never called here, right? There is a separate place handling search feedback, isn't it?

api-client/src/errors/ResponseError.js
Taggotty: Found it. See https://github.com/taylesworth/react-native-mock/blob/06bb395c8f2b7228716db4bb29330d10a60b65bb/src/react-native.js#L127 . Maybe there is a way to keep the original form data. But as everything else seems to be unaffected it may be enough to provide a functionality for iterating the form data object depending on the platform.

native/src/modules/categories/components/Categories.js
Taggotty: good question. Earlier I was also fetching the cities in createNavigateToFeedbackModal. As I removed it, I will do as you say.

native/src/routes/feedback/containers/FeedbackModalContainer.js
Taggotty: Yes, that's what I meant.

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,33 +1,287 @@
...
+    const feedbackCategory = CONTENT_FEEDBACK_CATEGORY
+    if (type === 'Category' && path && title) {
+      return new FeedbackVariant({
+        label: t('contentOfPage', { page: title }),
+        feedbackType: PAGE_FEEDBACK_TYPE,
+        feedbackCategory
+      })
+    } else if (type === 'Pois' && path && title) {
+      return new FeedbackVariant({
+        label: t('contentOfPoi', { poi: title }),
```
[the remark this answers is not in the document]
Taggotty: This is also related to to IGAPP-21. Which is about creating the translations for the pois page.

klinzo (approved):

native/src/routes/feedback/containers/FeedbackModalContainer.js
Taggotty: Tested. There is only on Event feedback.

native/src/routes/feedback/containers/FeedbackModalContainer.js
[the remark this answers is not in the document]
klinzo: I think this would normally be part of IGAPP-337. You can add the key with a dummy translation already if you want though. We don't know what to call POIs yet though afaik.

web/src/modules/feedback/FeedbackInformation.js
Taggotty: Ahh. I accidentally created the class in web. That's why I didn't find it ^^ 
Removed as not used.

native/src/routes/feedback/containers/FeedbackModalContainer.js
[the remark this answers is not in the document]
maxammann: I'd suggest to just add it here to the translations as it is just one line. That way we do not forget it.

api-client/src/stringifyFormData.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+type FormDataType = FormData
+  | {| getParts: () => Array<{ string: string, fieldName: string }> |}
+
+export function stringifyFormData (formData: FormDataType): string {
```
maxammann, line 6: Maybe a cmment would be nice here: `stringifies FormData from react-native or the browser variant of it`

maxammann (approved):

native/jest.config.js
```diff
@@ -10,6 +10,7 @@ const transformNodeModules = [
   'hashids',
   '@react-native-community/progress-bar-android',
   '@react-native-community/progress-view',
+  '@react-native-community/picker',
```
klinzo, line 13: Do we use `@react-native-community/picker` or `@react-native-picker/picker` now? There are usages of both

klinzo (approved):

native/jest.config.js
Taggotty: The latter one. They recently moved to another repo. Must have missed the occurence of `@react-native-community/picker`

klinzo (approved):

PR #87, opened 2020-11-27
IGAPP-215: Changed the notFound key to "Page not found"

SimonVortkamp:
https://issues.tuerantuer.org/browse/IGAPP-215

### DIFF
--- a/locales/locales.json
+++ b/locales/locales.json
@@ -39,7 +39,7 @@
         "disclaimer": "Imprint and contacts",
         "mainDisclaimer": "Imprint and contacts - {{appName}}",
         "pois": "Points of interest",
-        "notFound": "Page not located",
+        "notFound": "Page not found",
         "tunews": "tünews",
         "localNews": "Local news"
       },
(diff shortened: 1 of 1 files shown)


klinzo (approved):
So you did check the translations of other languages as well, right?

SimonVortkamp:
Now the translations should be right.

locales/locales.json
```diff
@@ -216,7 +216,6 @@
         "disclaimer": "ናይ ደራሲ ዝርዝራት ሓበሬታን ናይ ርክብን",
         "mainDisclaimer": "ናይ ደራሲ ዝርዝራት ሓበሬታን ናይ ርክብን - {{appName}}",
         "pois": "ዘገድሱ ነጥብታት",
-        "notFound": "ገጽ ኣይተረኽበን",
```
maxammann, line 219: I think all translators used german as their source. This means the "Page not located" is an exception.

Did you use google translator here? If yes then I woudl say we leave the translation here.

maxammann (approved):

locales/locales.json
SimonVortkamp: Yes, google translates this as "Sagen Sie uns nicht die Seite" or "Don't tell us the page", that's why I deleted it. I'm not sure how to check this translation.

maxammann: Me neighter :D but I'm oke with removing if we are not sure :+1:

PR #90, opened 2020-11-30
IGAPP-270: Migrate to monorepo and unify eslint configs

klinzo:


.eslintrc.js
```diff
@@ -0,0 +1,103 @@
+// @flow
```
klinzo: Merged all eslint configs of the eslint repo along with those in the project in this config. I am open for suggestions if you think some of these are overkill or you are missing some more rules.

.eslintrc.js
```diff
@@ -0,0 +1,103 @@
...
+    'flowtype/object-type-delimiter': ['error', 'comma'],
+    'flowtype/require-exact-type': ['warn', 'always'],
+    'flowtype/require-parameter-type': ['error', { excludeArrowFunctions: true }],
+    'flowtype/require-return-type': ['error', 'always', {
+      excludeArrowFunctions: true,
+      annotateUndefined: 'never',
+      excludeMatching: ['^render$']
+    }],
+    'flowtype/semi': ['error', 'never'],
+    'flowtype/type-id-match': ['error', '^(iOS)?([A-Z][a-z0-9]+)+Type$'],
```
klinzo, line 67: Allowing starting with `iOS` now. Imo `IOS` and `Ios` are looking really weird.

.eslintrc.js
```diff
@@ -0,0 +1,103 @@
...
+    'jest/no-focused-tests': 'error',
+    'jest/no-identical-title': 'error',
+    'jest/no-test-prefixes': 'error',
+    'jest/prefer-to-be-null': 'error',
+    'jest/prefer-to-have-length': 'error',
+    'jest/valid-describe': 'error',
+    'jest/valid-expect': 'error'
+  },
+  overrides: [{
+    files: ['**/src/**/*.js'],
```
klinzo, line 81: If you think there is a better file matcher, let me know.

.eslintrc.js
```diff
@@ -0,0 +1,103 @@
...
+    'react/no-unused-prop-types': 'warn',
+    'react/no-typos': 'error',
+    'react/prefer-es6-class': ['error', 'always'],
+
+    'flowtype/delimiter-dangle': ['error', 'never'],
+    'flowtype/newline-after-flow-annotation': ['error', 'always'],
+    'flowtype/no-primitive-constructor-types': 'error',
+    'flowtype/no-weak-types': 'warn',
+    'flowtype/object-type-delimiter': ['error', 'comma'],
+    'flowtype/require-exact-type': ['warn', 'always'],
```
klinzo, line 59: Added this rule. It is responsible for most warnings. I think we should gradually fix all inexact types. Will close https://issues.tuerantuer.org/browse/LINT-8

klinzo:
We still have some linting issues:
* I think we can reject this (restrict import of compose) as I think we don't use it: https://issues.tuerantuer.org/browse/LINT-11
* Stylelint preset: https://issues.tuerantuer.org/browse/LINT-6
* Babel preset: https://issues.tuerantuer.org/browse/LINT-5

web/src/modules/app/containers/Switcher.js
```diff
@@ -129,7 +129,8 @@ export class Switcher extends React.Component<PropsType> {
         <Layout footer={showFooter && <GeneralFooter language={language} />}
                 header={(error || routeConfig.requiresHeader) && <GeneralHeader viewportSmall={viewportSmall} />}
                 darkMode={darkMode}>
-          {error ? <LanguageFailure cities={citiesPayload.data}
+          {error
+            ? <LanguageFailure cities={citiesPayload.data}
```
Taggotty, line 133: If you do add this new Format then also align the attributes below.

.eslintrc.js
Taggotty: why do we want this rule only for specific files?

klinzo: Otherwise I had to type all config files etc. which would lead to lots of errors and necessary fixes. Also we don't have flow in the root project at all, so it would not make sense at all to type these files.

Taggotty: Thanks for explaining. I think the matcher should be sufficient in that case. 👍🏼

Taggotty (commented):
Just out of interest. Do you use intelliJs built in formatter for this or via cmd? And is there any autoformat on save or are you manually executing formating? I often forget to format before merging :/

klinzo:
> Just out of interest. Do you use intelliJs built in formatter for this or via cmd? And is there any autoformat on save or are you manually executing formating? I often forget to format before merging :/

Right now I am mostly doing it manually to be honest... IntelliJ should pickup most of the stuff automatically though already and display errors. There is also a shortcut (cmd + alt + L) to reformat selected code (also possible for directories afaik).

There are even more options described here: https://www.jetbrains.com/help/idea/eslint.html
`You can also configure ESLint to fix all the problems in a file when this file is saved. To configure such behavior, select the Run eslint --fix on save checkbox on the ESLint page of the Settings dialog as described in Activating and configuring ESLint in IntelliJ IDEA below. ` Corresponding link: https://www.jetbrains.com/help/idea/eslint.html#ws_js_eslint_activate

.eslintrc.js
maxammann: But some config files already benefit from linting. That way we ignore linting. Previously we had separate configs for linting and source.

maxammann (approved):
Does all of this work with IntelliJ autoformatter?

.eslintrc.js
klinzo: This only disables one line which is missing flow annotation. Every other rule is enforced on all files.

klinzo:
> Does all of this work with IntelliJ autoformatter?

Mostly the same way as before I think. I only added one new rule (exact flow types).

.eslintrc.js
maxammann: oh sorry did not get the indentation :)

Taggotty (approved):

PR #91, opened 2020-11-30
IGAPP-319: Change application/bundle id for integreat-test-cms build config

klinzo:
…taneous installations


klinzo:
@maxammann should we only do this for Android or should we create another provisioning profile? Not sure what to do with our e2e builds. We could use another bundle identifier/application id or use that of integreat-test-cms or integreat.

Taggotty (approved):

maxammann (approved):
E2E should have the same Id as the production config

PR #92, opened 2020-12-01
IGAPP-175: Add missing attributes to database connector

klinzo:


klinzo:
WebApp tests and flow errors will be fixed in #93 .

klinzo:
Actually testing this (best upgrading a previous installation) would probably be good :)

klinzo:
Also, some native tests in the DataConnector are failing locally for me because of moment. Can you reproduce this? In the CI the tests all pass.

native/src/modules/endpoint/DatabaseConnector.js
```diff
@@ -5,7 +5,7 @@ import {
   CategoryModel,
   CityModel,
   DateModel,
-  EventModel,
+  EventModel, FeaturedImageModel,
```
Taggotty, line 8:
```suggestion
  EventModel, 
  FeaturedImageModel,
```

native/src/modules/endpoint/DatabaseConnector.js
```diff
@@ -466,6 +479,7 @@ class DatabaseConnector {
         pushNotificationsEnabled: jsonObject.pushNotificationsEnabled,
         tunewsEnabled: jsonObject.tunewsEnabled,
         offersEnabled: jsonObject.extras_enabled,
+        poisEnabled: jsonObject.pois_enabled || false,
```
Taggotty, line 482: pois_enabled should never be undefined, should it.

Taggotty (approved):

native/src/modules/endpoint/DatabaseConnector.js
klinzo: Hmm I admit this should be redundant most of the time. However, after upgrading the app without a fetch/store of the cities the old cities won't have this flag, so I am setting it to false otherwise here. Kind of does not matter because undefined is falsy as well. I don't mind removing it again.

maxammann:
What are there so many changes related to linting in this PR?

native/src/modules/endpoint/DatabaseConnector.js
maxammann: If linting does not complain then both is fine ;)

maxammann (approved):
Not sure whether everything is already, not easy to review with all these changes.

The only missing one was `poisEnabled` right?

maxammann (approved):
This file could possibly need some updates: https://github.com/Integreat/integreat-app/blob/823894cdb32b23d0a0cde0ca5c40cdb17a8e53f5/native/docs/persistence.md#L37

native/src/modules/endpoint/DatabaseConnector.js
Taggotty: then I would prefer if you remove it. Otherwise it would be inconsistent with offersEnabled and tunewsEnabled

native/src/modules/endpoint/DatabaseConnector.js
klinzo: Agreeing with @Taggotty even though we don't have a rule for this :D Will fix in #93

native/src/modules/endpoint/DatabaseConnector.js
klinzo: Will do in #93

PR #93, opened 2020-12-01
IGAPP-403: Move builders to api-client

klinzo:


klinzo:
Use of the builders can be increased in future PRs if there are changes anyway. I think this could be a gradual change as well. Otherwise this PR would have been gotten much bigger than it is already.

klinzo:
Some native tests in the DataConnector are failing because of moment. Any idea why this happens? I don't have a clue... the iso strings are the same and moment versions as well :disappointed:

maxammann (approved):
nice removals :)

klinzo:
@maxammann I think I found a relatively nice solutions for the overrides :)

Taggotty (approved):
Nice and tidy. 👍🏼

Taggotty (approved):

PR #94, opened 2020-12-02
IGAPP-109: Use URL to get Pathnames

Taggotty:

### DIFF
--- a/native/src/modules/app/createNavigateToInternalLink.js
+++ b/native/src/modules/app/createNavigateToInternalLink.js
@@ -28,20 +28,19 @@ export const createNavigateToInternalLink = ({
   const pathnameParts = pathname.split('/').filter(Boolean)
   const newCity = pathnameParts[0]
   const newLanguage = pathnameParts[1]
-  const normalizedPathname = pathnameParts.reduce((acc, part) => `${acc}/${part}`, '')
 
   if (!newCity) { // '/'
     navigateToLanding()
   } else if (pathnameParts[2] === 'events') {
     if (pathnameParts[3]) { // '/augsburg/de/events/some_event'
-      navigateToEvent({ cityCode: newCity, language: newLanguage, path: normalizedPathname })
+      navigateToEvent({ cityCode: newCity, language: newLanguage, path: pathname })
     } else { // '/augsburg/de/events'
       navigateToEvent({ cityCode: newCity, language: newLanguage, path: null })
     }
   } else if (pathnameParts[2]) { // '/augsburg/de/willkommen'
-    navigateToCategory({ cityCode: newCity, language: newLanguage, path: normalizedPathname })
+    navigateToCategory({ cityCode: newCity, language: newLanguage, path: pathname })
   } else { // '/augsburg/de' or '/augsburg'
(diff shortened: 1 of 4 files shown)


native/src/modules/app/createNavigateToInternalLink.js
```diff
@@ -22,24 +23,25 @@ export const createNavigateToInternalLink = ({
...
 }) => ({ url, language }: NavigateToInternalLinkParamsType) => {
-  const parts = url.split('/').filter(segment => segment)
-  const pathnameParts = parts.splice(2)
+  const parsedUrl = new URL(url)
+  const pathname = parsedUrl.pathname
+  const pathnameParts = pathname.split('/').filter(Boolean)
   const newCity = pathnameParts[0]
   const newLanguage = pathnameParts[1]
-  const pathname = pathnameParts.reduce((acc, part) => `${acc}/${part}`, '')
+  const normalizedPathname = pathnameParts.reduce((acc, part) => `${acc}/${part}`, '')
```
Taggotty, line 31: Added this as double slashes are actually allowed in URL to current standard. See [here](https://url.spec.whatwg.org/#dom-url-pathname). Alternatively I tried using `react-native-url-polyfill` and I tried using  `ẁhatwg-url`, but as they both also follow the URL standard directly the result did not change. Do we really need a handling for double slashes in URLs?

I do not feel that this PR is a real improvement. Other opinions?

klinzo: I am quite sure that this was not to normalize double slashes, but rather to just have a pathname at all. So this line is not necessary imo, just use the pathname from above. Previously we did not have a pathname at all which is why we had to create it manually by splitting and splicing from the url.
The only improvement should be that we don't have to split the url manually to remove the `https://integreat.app` to get the pathname which is admittedly really a minor improvement.

klinzo: The filter above is probably not needed as well.

klinzo (commented):
Is this behavior tested? Could you add some tests if not please?

native/src/modules/app/createNavigateToInternalLink.js
Taggotty: The filter is needed, as every pathname starts with a `/`. `/augsburg.split('/')` would therefore result in `['', 'augsburg']`

klinzo (approved):

native/src/modules/app/createNavigateToInternalLink.js
```diff
@@ -22,11 +23,11 @@ export const createNavigateToInternalLink = ({
   navigateToCategory: NavigateToCategoryParamsType => void,
   navigateToDashboard: NavigateToCategoryParamsType => void
 }) => ({ url, language }: NavigateToInternalLinkParamsType) => {
-  const parts = url.split('/').filter(segment => segment)
-  const pathnameParts = parts.splice(2)
+  const parsedUrl = new URL(url)
+  const pathname = parsedUrl.pathname
+  const pathnameParts = pathname.split('/').filter(Boolean)
```
maxammann, line 28: This filter filters empty parts, right?

maxammann (approved):

PR #95, opened 2020-12-07
IGAPP-57: External links should open in the same tab

klinzo:

### DIFF
--- a/web/src/modules/common/components/CleanAnchor.js
+++ b/web/src/modules/common/components/CleanAnchor.js
@@ -9,7 +9,6 @@ const StyledAnchor = styled.a`
 `
 
 type PropsType = {|
-  newTab?: boolean,
   href: string,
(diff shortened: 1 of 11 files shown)


web/src/modules/common/components/CleanAnchor.js
```diff
@@ -9,22 +9,17 @@ const StyledAnchor = styled.a`
...
-  newTab?: boolean,
   href: string,
   children: React.Node,
   ariaLabel?: string,
   className?: string
 |}
 
 class CleanAnchor extends React.PureComponent<PropsType> {
-  static defaultProps = {
-    newTab: true
```
maxammann, line 21: For other reviewers: This is the line which is responsible for opening the link in the same tab

maxammann (approved):

Taggotty (approved):

PR #96, opened 2020-12-07
IGAPP-124: Disable data detection in iOS webview

klinzo:

See discussion here: https://tuerantuer.slack.com/archives/CDV9QF1GA/p1587050318013000
This seems not to happen anymore, but we think disabling is still the best option.

API reference: https://github.com/react-native-webview/react-native-webview/blob/d4ab332891af2242e663ac2e25f4aa3317945399/docs/Reference.md#datadetectortypes


maxammann (approved):
I'm not entirely sure why we do not want the webview to detect phone number for example.

But the issue clearly states it so lets disable it :D

klinzo:
> I'm not entirely sure why we do not want the webview to detect phone number for example.
> 
> But the issue clearly states it so lets disable it :D

Have a read at the linked slack thread: https://tuerantuer.slack.com/archives/CDV9QF1GA/p1587050318013000
Main reason is WYSIWYG and that it is possible to highlight it in the cms.

Taggotty (approved):

maxammann (approved):

PR #98, opened 2020-12-09
IGAPP-177: Add support for custom tabs, IGAPP-191: Fix webview back navigation

klinzo:

Not sure about everything but I think I really like the custom tabs :blush: Using the theme color may be to much...
The lehrstellenradar does only support post which is supported neither by the react native linking nor by the inapp browser library :/


native/src/modules/common/components/RemoteContent.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import * as React from 'react'
+import React, { useState, useEffect, type Element } from 'react'
```
klinzo, line 3: Only converted this to functional component because I initially thought the work has to be done here...

native/src/modules/common/components/RemoteContent.js
```diff
@@ -32,70 +31,65 @@ type PropType = {|
...
+  }, [webViewHeight])
+
+  function onLayout (event: ViewLayoutEvent) {
     const { width } = event.nativeEvent.layout
-    this.setState({ webViewWidth: width })
+    setWebViewWidth(width)
   }
 
-  onMessage = (event: WebViewMessageEvent) => {
+  function onMessage (event: WebViewMessageEvent) {
```
maxammann, line 47: Why are you using `function`s here? lambdas would be nicer

maxammann: If this is because of eslint then read this here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#react-hooks

native/src/routes/external-offer/components/ExternalOffer.js
```diff
@@ -1,34 +1,56 @@
...
+  return <WebView
+    source={postData ? createPostSource(url, body) : createGetSource(url, body)}
+    javaScriptEnabled
+    dataDetectorTypes={['all']}
+    domStorageEnabled={false}
+    ref={webviewRef}
+    /* The Lehrstellenradar does not work in the webview if you are navigating back (probably because of FormData?).
+        Without this the webview itself shows an error which looks more like it is our fault.}
+        Also, without disabling cache it is not possible to navigate back with hardware buttons anymore. */
+    cacheEnabled={false}
```
maxammann, line 49: :+1:

native/src/modules/common/openExternalUrl.js
```diff
@@ -0,0 +1,19 @@
...
+import { InAppBrowser } from 'react-native-inappbrowser-reborn'
+import buildConfig from '../app/constants/buildConfig'
+
+const openExternalUrl = async (url: string) => {
+  try {
+    if (await InAppBrowser.isAvailable()) {
+      await InAppBrowser.open(url, {
+        toolbarColor: buildConfig().lightTheme.colors.themeColor
+      })
+    } else { Linking.openURL(url) }
```
maxammann, line 13: Should we check here whether we can open it like in tunews?
```
    const supported = await Linking.canOpenURL(tunewsWebsiteUrl)
    if (supported) {
      await Linking.openURL(tunewsWebsiteUrl)
    }
```

maxammann (approved):

native/src/modules/common/openExternalUrl.js
klinzo: Sure, we could. But what happens if it is now supported? Imo better to just open it and the user gets an error there.

native/src/modules/common/components/RemoteContent.js
klinzo: Good suggestion :+1:

klinzo:
Imo we can reject this: https://issues.tuerantuer.org/browse/IGAPP-130. I think the new/current behavior is fine.

native/src/modules/common/openExternalUrl.js
maxammann: I suppose that Linking.open will crash the app :D But I'm not sure what the behavior is.

We could show a snackbar when an error occurrs.

maxammann: Oh oke then the catch block should hit.
I think a good implementation would be to check `canOpenURL()`. If the result is false then show "You do not have an app installed to open this link" in a snackbar.

If an error occurs when calling `canOpenURL` or `openLink` then show some other snackbar.

Do we already have a snackbar for this usecase? If yes then I would say that we use it here, if not then we can create a separate task.

klinzo: We don't have a snackbar yet for such a usecase yet, nope. Will only be implemented later.

klinzo: I don't think this is not relevant at all. The only case this can occur is if the user does not have custom tabs support (or the pendant on iOS which is probably never the case because of apple) AND no browser is installed. Or is there any other usecase where this will happen?

maxammann: For example if somebody in the backend embeds an email link, or geo link and there is no app.
This is just proper/better error handling. Not much benefit, but also not much work.

klinzo: Okay, so just wrapping in canOpenURL and adding a todo for the snackbar?

maxammann: I think that would be good!

Taggotty (approved):

maxammann (approved):

PR #100, opened 2020-12-11
IGAPP-386: Add a way to normalize language codes

maxammann:


maxammann:
This PR is a preparation for fixing our language problems. This PR introduces the primitives to support languages better:

* Native module to get locales
* Fallbacks specified in config.js
* Flow types in translations package
* Fixes of babel versions accross the project
* Fixing how I18NextProvider works in web

api-client/babel.config.js
```diff
@@ -2,8 +2,5 @@ module.exports = {
   presets: [
     '@babel/preset-env',
     '@babel/preset-flow'
-  ],
-  plugins: [
-    '@babel/plugin-proposal-class-properties'
```
maxammann, line 7: This is already applied in preset-env

api-client/package.json
```diff
@@ -16,8 +16,8 @@
     "eslint": "eslint ."
   },
   "dependencies": {
-    "@babel/polyfill": "^7.8.3",
-    "@babel/runtime": "^7.8.3",
+    "@babel/polyfill": "^7.12.1",
```
maxammann, line 19: Caret matching for dependencies and exact version for dev dependencies

translations/config.js
```diff
@@ -0,0 +1,78 @@
...
+    sr: {},
+    ps: { rtl: true },
+    kmr: { rtl: true },
+    am: {},
+    bg: {},
+    el: {},
+    it: {}
+  },
+  // Fallbacks for unnormalized language codes from our backend
+  fallbacks: {
```
maxammann, line 52: This is the core of this PR ;)

api-client/package.json
klinzo: Probably its fine to use different babel packages with different versions, right? Because we now have e.g. `7.12.1`, `7.12.10`, ...

package.json
```diff
@@ -11,8 +11,6 @@
       "build-configs"
     ],
     "nohoist": [
-      "build-configs/**",
```
klinzo, line 14: Why don't we need this?

translations/config.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+
+type ConfigType = {
```
klinzo, line 3: Make this strict

translations/config.js
```diff
@@ -0,0 +1,78 @@
...
+  sourceLanguage: string,
+  supportedLanguages: { [languageCode: string]: string },
```
klinzo, line 5: I think this type is not correct, is it? Shouldn't the type be `{ [languageCode: string]: {| rtl?: boolean |} }`? Or depending on my following suggestion: `{ [languageCode: string]: {| rlt: boolean |} }`

translations/config.js
```diff
@@ -0,0 +1,78 @@
...
+
+  return config
+}
+
+module.exports = checkConsistency({
+  // The language from which we translate
+  sourceLanguage: 'de',
+  // The languages into which we translate from 'sourceLanguage' including the sourceLanguage
+  supportedLanguages: {
+    de: {},
```
klinzo, line 29: I think I would do:
```suggestion
    de: { rtl: false },
```

translations/package.json
```diff
@@ -4,7 +4,18 @@
...
+  "scripts": {
+    "manage": "babel-node tools/manage.js"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.12.5"
+  },
   "devDependencies": {
+    "@babel/core": "7.12.10",
+    "@babel/node": "7.12.10",
+    "@babel/plugin-proposal-class-properties": "7.12.1",
```
klinzo, line 16: Are these really needed? At least the plugin should not be necessary (as it is used nowhere).

web/package.json
```diff
@@ -28,7 +28,8 @@
     "flow:install-types": "flow-typed install --skip"
   },
   "dependencies": {
-    "@babel/polyfill": "^7.8.3",
+    "@babel/polyfill": "^7.12.1",
+    "@babel/runtime": "^7.12.5",
```
klinzo, line 32: Why do we need babel runtime here?

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -47,37 +47,54 @@ export class I18nProvider extends React.Component<PropsType, StateType> {
...
-      }
+      this.updateLanguageState(targetLanguage)
     })
   }
 
+  updateLanguageState = (targetLanguage: string) => {
+    const fonts = I18nProvider.getSelectedFonts(targetLanguage)
+    this.setState(prevState => ({ ...prevState, language: targetLanguage, fonts }))
+
+    this.props.setUiDirection(RTL_LANGUAGES.includes(targetLanguage) ? 'rtl' : 'ltr')
```
klinzo, line 86: Is `RTL_LANGUAGES` replaced in a different PR?

translations/src/loadTranslations.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+import { merge } from 'lodash'
+import defaultTranslations from '../translations.json'
+import transformTranslations from './transformTranslations'
+import type { TranslationsType, TransformedTranslationsType } from './types'
+
+const loadTranslations = (translationsOverride: TranslationsType): TransformedTranslationsType => {
```
klinzo, line 8: Isn't this duplicated code?

translations/src/transformTranslations.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { forEach, reduce } from 'lodash'
+import type { TransformedTranslationsType, TranslationsType } from './types'
+
+/**
+ * Transform translation resources from our internal translations format to be i18next compatible.
+ * @param {object} translations in our format: namespace -> languageCode -> key -> value
+ * @returns {object} transformed translations in the format: languageCode -> namespace -> key -> value
+ */
+export default (translations: TranslationsType): TransformedTranslationsType => reduce(
```
klinzo, line 11: Isn't this duplicated code?

klinzo (changes requested):

api-client/package.json
maxammann: yeah this is something which you do not see when using caret versions.
Some libraries are not yet updates in babel

web/package.json
maxammann: Yes the webapp should have this runtime as dependency

web/src/modules/i18n/containers/I18nProvider.js
maxammann: Yes

translations/src/loadTranslations.js
maxammann: Yes this will be replaced in an other PR

web/package.json
klinzo: Yeah okay, but was is the reason why we need this? :D

native/src/modules/endpoint/reducers/pushEvent.js
```diff
@@ -4,7 +4,7 @@ import type { CityContentStateType, EventRouteStateType } from '../../app/StateT
 import type { PushEventActionType } from '../../app/StoreActionType'
 import { EventModel } from 'api-client'
 import ErrorCodes from '../../error/ErrorCodes'
-import { entries, values } from '../../../utils/object'
+import { values, entries } from 'translations'
```
klinzo, line 7: Why is this now part of `tranlsations`?

klinzo (approved):

native/src/modules/endpoint/reducers/pushEvent.js
maxammann: I moved it as it is used in both projects. translations is used in every other package in the repo.

I think this is better instead of creating a new utils package

web/package.json
maxammann: before this PR it probably was used inexplicitly. I added it for consistency.

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -47,37 +47,54 @@ export class I18nProvider extends React.Component<PropsType, StateType> {
...
+      debug: buildConfig().featureFlags.developerFriendly,
+      detection: {
+        order: ['localStorage', 'navigator']
+      }
     })
+
     this.state = {
-      language: DEFAULT_LANGUAGE,
+      language: this.i18n.language,
       fonts: I18nProvider.getSelectedFonts(DEFAULT_LANGUAGE),
```
Taggotty, line 58: shouldn't this be
```suggestion
      fonts: I18nProvider.getSelectedFonts(this.i18n.language),
```

Taggotty (approved):

web/src/modules/i18n/containers/I18nProvider.js
maxammann: Good catch, probably overwritten by the other pr thought!

PR #101, opened 2020-12-14
IGAPP-387: Choose translation based on normalized language code

maxammann:


web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -2,135 +2,102 @@
```
maxammann, line 2: I refactored this to a function component to be same as in web

web/src/modules/i18n/BrowserLanguageDetector.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+      if (navigator.userLanguage) {
+        bcp47Tags.push(navigator.userLanguage)
+      }
+      if (navigator.language) {
+        bcp47Tags.push(navigator.language)
+      }
+    }
+
+    const iso639Codes = bcp47Tags.map(bcp47Tag => {
```
maxammann, line 51: We get BCP47 codes from the browser. This is a simple way of extracting the primary language code.

We need to do this for now as we can not match en-US to the backend language en

web/src/modules/i18n/BrowserLanguageDetector.js
```diff
@@ -0,0 +1,65 @@
+let hasLocalStorageSupport = null
+
+const localStorageAvailable = () => {
```
maxammann, line 3: I copied this from the external module. I'd keep it as it is to avoid problems with browsers. This code is quite well tested.

native/src/modules/i18n/components/I18nProvider.js
```diff
@@ -1,100 +1,97 @@
 // @flow
```
maxammann, line 1: I refactored this to a function component to be same as in web

web/src/modules/i18n/BrowserLanguageDetector.js
maxammann: We need to decide whether this is oke we whether we implement code to skip the "Language not available" dialog by automatically matching en-US to en. We can use the languages endpoint to implement a heuristic for choosing a language.

web/src/modules/i18n/BrowserLanguageDetector.js
klinzo: Why did you chose to copy this? Probably it would be good to add a comment. Did you modify something? Should we include the tests as well then?

web/src/modules/i18n/BrowserLanguageDetector.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (localStorageAvailable()) {
+      const localStorageLanguage = window.localStorage.getItem(LANGUAGE_LOCAL_STORAGE)
+      if (localStorageLanguage) {
+        bcp47Tags.push(localStorageLanguage)
+      }
+    }
+
+    if (typeof navigator !== 'undefined') {
+      if (navigator.languages) { // chrome only; not an array, so can't use .push.apply instead of iterating
```
klinzo, line 38: Why not just use `apply` then?

web/src/modules/i18n/BrowserLanguageDetector.js
klinzo: I think there should not be a `Language not available` dialog for `en-US`. I think this is okay for now.

web/src/modules/i18n/containers/I18nProvider.js
klinzo: This is not exactly the same though, is it?

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -2,135 +2,102 @@
...
+
+type PropsType = {| children: React.Node |}
+
+export default ({ children }: PropsType) => {
+  const [language, setLanguage] = useState<string>(config.defaultFallback)
+  const [errorMessage, setErrorMessage] = useState<?string>(null)
+  const [i18nextInstance, setI18nextInstance] = useState(null)
+
+  const dispatch = useDispatch()
+  const contentLanguage = useSelector((state: StateType) => state.location.payload.language)
```
klinzo, line 24: I like this :+1:

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -2,135 +2,102 @@
...
+    initI18Next().catch((e: Error) => {
+      setErrorMessage(e.message)
+      console.error(e)
     })
-  }
-
-  updateLanguageState = (targetLanguage: string) => {
-    const fonts = I18nProvider.getSelectedFonts(targetLanguage)
-    this.setState(prevState => ({ ...prevState, language: targetLanguage, fonts }))
+  }, [])
```
klinzo, line 59: Shouldn't there be a dependency?

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -2,135 +2,102 @@
...
+        interpolation: {
+          escapeValue: false /* Escaping is not needed for react apps:
+                                https://github.com/i18next/react-i18next/issues/277 */
+        },
+        debug: buildConfig().featureFlags.developerFriendly
+      })
+
+      setI18nextInstance(i18nextInstance)
+
+      // Apply ui language as language
```
klinzo, line 49: If the browser language is changed?

Taggotty (approved):

Taggotty:
Looks really fine so far. I really like reviewing the hook syntax. looks a lot cleaner now. Don't forget to add/adjust the tests, as this is new behavior.

web/src/modules/i18n/containers/I18nProvider.js
maxammann: I think it is. The behavior of the app stayed the same in my tests.

web/src/modules/i18n/BrowserLanguageDetector.js
maxammann: I did some adjustments so I think there is no reason to add a comment. I create tests myself.

native/src/modules/i18n/components/I18nProvider.js
```diff
@@ -1,100 +1,97 @@
...
+      const resources = loadTranslations()
+      const i18nextInstance = await i18next
+        .createInstance()
+        .use(NativeLanguageDetector)
+
+      await i18nextInstance.init({
+        resources,
+        fallbackLng: {
+          ...config.fallbacks,
+          default: [config.defaultFallback]
```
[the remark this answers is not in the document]
maxammann: defaulFallback is the fallback which is taken when there is no fallback in the fallback array.

translations/src/config.js
```diff
@@ -12,22 +13,35 @@ class Config {
...
+    },
     fr: { rtl: false },
     ro: { rtl: false },
     tr: { rtl: false },
     pl: { rtl: false },
     ti: { rtl: false },
-    ku: { rtl: false },
+    ku: {
+      rtl: false,
+      additionalFont: 'lateef'
```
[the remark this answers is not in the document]
maxammann: Yes

web/src/modules/i18n/BrowserLanguageDetector.js
maxammann: what do you mean with apply?

I will not change this code as it is tested and works in all browsers. I will add tests though.

web/src/modules/i18n/containers/I18nProvider.js
klinzo: Yes, functionality, but you wrote to be the same as in native and not to have the same functionality as before :)

web/src/modules/i18n/BrowserLanguageDetector.js
klinzo: Hmm your call. I think for future reference it wouldn't be bad but never mind :shrug:

web/src/modules/i18n/BrowserLanguageDetector.js
klinzo: I don't understand the comment: `// chrome only; not an array, so can't use .push.apply instead of iterating`. Or rather, why there is a comment instead of just applying the suggestion from the comment.

translations/jest.setup.js
```diff
@@ -0,0 +1,2 @@
+// @flow
```
klinzo, line 1: What is this file for?

translations/package.json
```diff
@@ -19,7 +22,10 @@
     "commander": "^4.0.1",
     "csv": "^5.3.0",
     "flat": "^5.0.0",
-    "lodash": "^4.17.19"
+    "lodash": "^4.17.19",
+    "jest": "^25.5.4",
+    "jest-fetch-mock": "^3.0.3",
```
klinzo, line 27: Why do we need `jest-fetch-mock` here?

klinzo (approved):

translations/jest.setup.js
maxammann: I added it to have exactly the same config in translations, api-client etc.

web/src/modules/i18n/containers/I18nProvider.js
maxammann: For example yes

web/src/modules/i18n/BrowserLanguageDetector.js
maxammann: This is from the upstream repoistory, not my comment :D

klinzo: I think we should decide if this is our own code now or just copied. If it is just copied, add a reference. If it is our own code it should also look like our own code :stuck_out_tongue:

translations/src/config.js
```diff
@@ -44,20 +58,7 @@ class Config {
     fa_pr: ['fa'],
```
Taggotty, line 58: Please be patient with me if I still didn't undestand. But why `fa_pr` and not `fa-pr`?

web/src/modules/i18n/containers/__tests__/I18nProvider.spec.js
```diff
@@ -1,135 +1,181 @@
...
+      render(<Provider store={store}><I18nProvider>Hello</I18nProvider></Provider>)
+    })
+
+    expect(store.getActions()).toEqual([{ payload: 'ltr', type: 'SET_UI_DIRECTION' }])
   })
+
+  // We can not switch the language right now because it is bound to redux-first-router, we would need to trigger a
+  // state update through the router which is very difficult.
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('should remember the language over sessions', async () => {})
```
Taggotty, line 180: This does only apply in test, right?

Taggotty (approved):

native/src/modules/i18n/components/__tests__/I18nProvider.spec.js
```diff
@@ -168,10 +168,11 @@ describe('I18nProvider', () => {
       return <Text>{format(moment('2020-12-21T14:58:57+01:00'), {})}</Text>
     }
 
-    const { getByText } = render(<Provider store={store}><I18nProvider>
+    const { getByText, debug } = render(<Provider store={store}><I18nProvider>
```
klinzo, line 171: What is this good for?

translations/src/config.js
```diff
@@ -44,20 +58,7 @@ class Config {
...
     per: ['fa'],
     de: ['de'],
```
klinzo, line 60: This can be removed as well, same goes for `ku` and `fa`.

translations/src/config.js
klinzo: This are just shitty codes from the cms that could possibly occur, thats why we handle them here afaik.

klinzo (approved):

klinzo:
@maxammann this does fix IGAPP-178 (Support Deutsch-Einfach), right?

web/src/modules/i18n/containers/I18nProvider.js
maxammann: no it should behave exactly like before in web. The behaviour is still different between web and native

PR #103, opened 2020-12-16
IGAPP-194: Stabilize e2e tests

Taggotty:

Most of the time the ios e2e tests timed out. After looking at the appium logs of the failed tests you could see the error message `Capability 'waitForQuiescence' changed from string to boolean. This may cause unexpected behavior`. After some search I also found that this capability can lead to very slow or even stuck tests (more details [here](https://github.com/appium/appium-xcuitest-driver#desired-capabilities)). I don't think there is a good way to test whether the tests will work more stable after this change though. :)

### DIFF
--- a/native/e2e/config/configs.js
+++ b/native/e2e/config/configs.js
@@ -71,15 +71,15 @@ exports.browserstack_ci_android = {
   prefix: 'IG CI',
   platform: 'android',
   caps: {
+    'browserstack.appium_version': '1.18.0',
+    'browserstack.debug': true,
     'browserstack.user': process.env.E2E_BROWSERSTACK_USER,
     'browserstack.key': process.env.E2E_BROWSERSTACK_KEY,
     project: 'integreat-react-native-app',
     os_version: '9.0',
     device: 'Google Pixel 3',
-    real_mobile: 'true',
-    'browserstack.appium_version': '1.17.0',
     app: process.env.E2E_BROWSERSTACK_APP,
-    'browserstack.debug': true
+    real_mobile: 'true'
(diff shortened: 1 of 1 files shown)


klinzo (approved):

maxammann (approved):
That is quite weird as we did not pass `waitForQuiescence` before. Is there maybe a default which uses string? If yes then we should maybe update the webdriver dependency

Taggotty:
Hmm. Somehow the parameter is still not set correctly. I will take another look at that.

Taggotty:
> That is quite weird as we did not pass `waitForQuiescence` before. Is there maybe a default which uses string? If yes then we should maybe update the webdriver dependency

Normally this parameter should be set by browserstack, I think. Webdriver shouldn't have anything to do with that...But I think you can also pass Automate Capabilities to browserstack. As stated above, I will take another look at that.

Taggotty:
> Hmm. Somehow the parameter is still not set correctly. I will take another look at that.

Fixed. You have to pass it as a string so it is converted to a boolean later, otherwise the boolean will be interpreted as an object an therefore as a astring 🤦🏼

maxammann:
@Taggotty but probably then this PR will do nothing right? I mean we are passing it now as string to browserstack. Or did you test it again?

Taggotty:
> @Taggotty but probably then this PR will do nothing right? I mean we are passing it now as string to browserstack. Or did you test it again?

I tested it and now it works fine. A string will be parsed to bool, but a bool wirl be parsed to string. That was what I tried to explain.

PR #104, opened 2020-12-18
IGAPP-114: Upgrade react navigation

klinzo:


klinzo:
Another quite big PR from my side... there have been quite a lot necessary changes after upgrading react-navigation. Most of the changed lines is because of flow-typed though. I think reviewing the PR is not too bad. There is a separate PR with some fixes I did not want to do in this PR as well in order to keep it a little bit smaller.
Actually testing the code is therefore not necessary here.

Taggotty (approved):

native/src/modules/app/components/App.js
```diff
@@ -15,29 +14,30 @@ import type { DataContainer } from '../../endpoint/DataContainer'
...
                   <NavigatorContainer />
-                </IOSSafeAreaView>
-              </>
-            </SafeAreaProvider>
-          </I18nProviderContainer>
-        </StaticServerProvider>
-      </Provider>
-    )
-  }
+                </NavigationContainer>
```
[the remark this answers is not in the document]
klinzo: One is a `NavigatioNContainer` from the library which is definitely needed. The other one is a self defined `NavigatoRContainer` which is a wrapper (with connect) around our Navigator defining all routes. I don't mind changing the naming though if you think there is a better name.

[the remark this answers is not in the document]
Taggotty: ah okay. i missed `Navigator` and `Navigation`
Maybe we could call the `NavigatorContainer` something like `RoutesContainer` if that makes sense? Otherwise it is difficult to understand whats going on here

native/src/modules/app/components/NavigationTypes.js
```diff
@@ -0,0 +1,109 @@
+// @flow
```
maxammann, line 1: should this file go into `constants` and have a lower case name as it exports no class?

native/src/modules/categories/components/CategoryList.js
```diff
@@ -43,7 +41,8 @@ type PropsType = {|
   theme: ThemeType,
   onItemPress: (tile: { title: string, thumbnail: string, path: string }) => void,
   language: string,
-  thumbnail?: string
+  thumbnail?: string,
+  navigation: NavigationPropType<T>
```
maxammann, line 45: The navigation prop should not leak through to a standalone component. That way it gets less reusable.

native/src/modules/endpoint/hocs/withPayloadProvider.js
```diff
@@ -41,31 +41,33 @@ export type SuccessType<S: {}, R: {}> = {|
...
-export type PropsType<S: { dispatch: Dispatch<StoreActionType> }, R: {}> = {|
+export type PropsType<S: { dispatch: Dispatch<StoreActionType> }, R: {}, T: RoutesType> = {|
   ...StatusPropsType<S, R>,
   dispatch: Dispatch<StoreActionType>,
-  navigation: NavigationScreenProp<*>,
+  navigation: NavigationPropType<T>,
+  route: RoutePropType<T>,
   t?: TFunction
-|} | {|
+|} | {| // Necessary because of weird flow error saying t is missing in containers
```
maxammann, line 57: I dont understand this comment. I think it can be removed.

The types here could be extracted to type variables `const type Type = ...` though for better readability.

native/src/modules/app/createNavigateToCategory.js
```diff
@@ -2,38 +2,41 @@
...
+const createNavigateToCategory = <T: RoutesType>(
+  routeName: CategoriesRouteType | DashboardRouteType,
   dispatch: Dispatch<StoreActionType>,
-  navigation: NavigationStackProp<*>
+  navigation: NavigationPropType<T>
 ) => ({ cityCode, language, path, key = generateKey(), forceRefresh = false }: NavigateToCategoryParamsType) => {
-  navigation.navigate({
-    routeName,
-    params: {
-      onRouteClose: () => dispatch({ type: 'CLEAR_CATEGORY', params: { key } }),
```
maxammann, line 20: Is it oke that this is removed here?

native/src/modules/layout/components/Header.js
```diff
@@ -56,50 +57,38 @@ const BoxShadow: StyledComponent<{}, ThemeType, *> = styled.View`
...
+  ...StackHeaderProps,
   t: TFunction,
   theme: ThemeType,
   peeking: boolean,
   categoriesAvailable: boolean,
-  navigateToLanding: () => void,
   goToLanguageChange?: () => void,
-  routeCityModel?: CityModel
+  routeCityModel?: CityModel,
+  dispatch: Dispatch<StoreActionType>
```
maxammann, line 67: I think dispatch should not leak into standalone components. That way this is only usable with redux.

native/src/modules/layout/components/Header.js
```diff
@@ -142,8 +131,8 @@ class Header extends React.PureComponent<PropsType> {
   }
 
   render () {
-    const { routeCityModel, navigation, t, theme, goToLanguageChange, peeking, categoriesAvailable } = this.props
-    const sharePath = navigation.getParam('sharePath')
+    const { routeCityModel, scene, t, theme, goToLanguageChange, peeking, categoriesAvailable } = this.props
+    const sharePath = scene.route.params?.sharePath || ''
```
maxammann, line 135: I think it makes more sense to pass null if the share path is undefined instead of an empty string.

native/src/modules/layout/components/TransparentHeader.js
```diff
@@ -60,13 +59,13 @@ const MaterialHeaderButtons = props => {
...
-    this.props.navigation.goBack(null)
+  goBackInStack = () => {
+    this.props.navigation.goBack()
   }
 
   onShare = async () => {
-    const { navigation, t } = this.props
-    const shareUrl: string = navigation.getParam('shareUrl')
+    const { scene, t } = this.props
+    const shareUrl = scene.route.params?.shareUrl || ''
```
maxammann, line 68: I think it makes more sense to pass null if the share path is undefined instead of an empty string.

native/src/modules/layout/components/TransparentHeader.js
```diff
@@ -85,18 +84,20 @@ class TransparentHeader extends React.PureComponent<PropsType> {
   }
 
   render () {
-    const { theme, float, navigation, t } = this.props
-    const shareUrl = navigation.state.params.url
+    const { theme, float, scene, t } = this.props
+    const shareUrl = scene.route.params?.shareUrl || ''
```
maxammann, line 88: I think it makes more sense to pass null if the share path is undefined instead of an empty string.

native/src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -101,10 +123,13 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
   const resourceCacheUrl = state.resourceCacheUrl
-  if (resourceCacheUrl === null || state.cities.status === 'loading' || switchingLanguage ||
-    (route.status === 'loading' && !routeHasOldContent(route)) || languages.status === 'loading') {
+  const { models, children, allAvailableLanguages } = route
+  if (resourceCacheUrl === null || state.cities.status === 'loading' || languages.status === 'loading' ||
+    (route.status === 'loading' && (!models || !allAvailableLanguages || !children))) {
     return { status: 'loading', progress: resourceCache.progress }
   }
+  // $FlowFixMe Flow does not get that models and children cannot be undefined as it is already checked above
+  const stateView = new CategoriesRouteStateView(route.path, route.models, route.children)
```
maxammann, line 132:
```suggestion
  const stateView = new CategoriesRouteStateView(route.path, models, children)
```

Using this should fix the flowfixme for models and children.

maxammann (approved):
I fucking hate router upgrades and changes...

native/src/modules/app/createNavigateToCategory.js
klinzo: Yes, this functionality is now handled in the payload provider as well.

native/src/modules/layout/components/Header.js
klinzo: Will fix in #105

native/src/modules/layout/components/TransparentHeader.js
klinzo: Will fix in #105

native/src/modules/layout/components/TransparentHeader.js
klinzo: Will fix in #105

native/src/modules/app/components/NavigationTypes.js
klinzo: Hm not sure. Until now we have used uppercase naming all the time for flow type files. Moving to constants would make sense though. Will do in #105.

native/src/modules/endpoint/hocs/withPayloadProvider.js
klinzo: It is already specified as separate type.
The comment means that 
```
{|
  ...StatusPropsType<S, R>,
  dispatch: Dispatch<StoreActionType>,
  navigation: NavigationPropType<T>,
  route: RoutePropType<T>
|}
```
should already be handled by the type above since the only difference is an optional `t` which means there does not have to be one:
```
{|
  ...StatusPropsType<S, R>,
  dispatch: Dispatch<StoreActionType>,
  navigation: NavigationPropType<T>,
  route: RoutePropType<T>,
  t?: TFunction
|}```

Comment (klinzo) [reply native/src/modules/endpoint/hocs/withPayloadProvider.js:57]: Changed the comment to ` // Necessary because of weird flow error saying t is missing even though t is already optional in the first type`

File: native/src/routes/categories/containers/CategoriesContainer.js (hunk 10 above)
Comment (klinzo) [reply native/src/routes/categories/containers/CategoriesContainer.js:132]: Sadly not :/

File: native/src/modules/app/components/NavigationTypes.js (hunk 2 above)
Comment (klinzo) [reply native/src/modules/app/components/NavigationTypes.js:1]: I think in the long term it should be moved (or at least some parts of it) to the api-client to unify route configs.

File: native/src/modules/layout/components/Header.js (hunk 6 above)
Comment (klinzo) [reply native/src/modules/layout/components/Header.js:67]: Agreed. However, this is similar to the behavior it was before (navigation was always passed) and I would refactor this separately.

File: native/src/modules/app/components/App.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (klinzo) [reply native/src/modules/app/components/App.js:33]: @maxammann what do you think?

File: native/src/modules/categories/components/CategoryList.js (hunk 3 above)
Comment (klinzo) [reply native/src/modules/categories/components/CategoryList.js:45]: Done in #105 . Quite some more changes though.

File: native/src/modules/app/components/NavigationTypes.js [hunk 11]
```diff
@@ -0,0 +1,109 @@
...
+    languages: Array<LanguageModel>,
+    availableLanguages: Array<string>
+  |},
+  PDFViewModal: {| url: string, shareUrl?: string |},
+  ImageViewModal: {| url: string, shareUrl?: string |},
+  FeedbackModal: FeedbackInformationType
+|}
+
+export type RoutePropType<T: RoutesType> = RouteProp<RoutesParamsType, T>
+export type NavigationPropType<T: RoutesType> = StackNavigationProp<RoutesParamsType, T>
```
Taggotty, line 109: Can you shortly explain again whats the difference between the `RouteProp` and `StackNavigationProp` here?

Taggotty (approved):

native/src/modules/app/components/NavigationTypes.js
klinzo: RouteProp contains the state of the react-navigation router, e.g.: all routes in the stack, the name of the current route as well as params (like cityCode, languageCode, categoriesPath)
StackNavigationProp contains methods to change this state, e.g. navigate, replace, dispatch,...

native/src/routes/categories/containers/CategoriesContainer.js
maxammann: Maybe you also need to check route.path above.

klinzo: No, path is not the problem

PR #105, opened 2020-12-19
IGAPP-114: Fixes for upgrade react-navigation

klinzo:


klinzo:
Fixes for #104 . These fixes include:
* Background color for all routes
* PermissionSnackbar
* Sharing for all routes
* Correct mapping from react-navigation routes to our redux categories state mapping
Still missing (will do in another separate PR):
* Tests
* pod install for iOS
* Testing on iOS

Imo the upgrade makes a lot of stuff easier, safer and much more readable. But sadly also some things more complicated :/

klinzo:
Please be sure to test everything quite carefully, especially navigating between the routes and sharing. Maybe stuff is slower in development mode, I could not experience any issues if building in release mode.

klinzo:
This upgrade possibly also fixes the following issues. Could you verify?
* Glitching header on iOS: https://issues.tuerantuer.org/browse/IGAPP-201
* Back Navigation on Dashboard on Android: https://issues.tuerantuer.org/browse/IGAPP-103
* Replace hard coded string routes with variables: https://issues.tuerantuer.org/browse/IGAPP-135
* Implement sharing for news and pois: https://issues.tuerantuer.org/browse/IGAPP-143

native/src/modules/app/components/App.js
```diff
@@ -15,11 +15,31 @@ import NavigatorContainer from '../containers/NavigatorContainer'
...
+  const [cityCode, setCityCode] = useState<?string>(null)
+  const [languageCode, setLanguageCode] = useState<?string>(null)
+
+  const onStateChange = useCallback(state => {
+    if (state) {
+      const route = state.routes[state.index]
+      setRouteName(route.name)
+      setRouteKey(route.key)
+      if (route.params?.cityCode && route.params?.languageCode) {
+        // $FlowFixMe cityCode is not of type mixed
```
maxammann, line 35: Is there a way of type the parameters? That is possible with typescript at least :sweat:

native/src/modules/app/components/App.js
```diff
@@ -15,11 +15,31 @@ import NavigatorContainer from '../containers/NavigatorContainer'
...
+import PermissionSnackbarContainer from '../../layout/containers/PermissionSnackbarContainer'
 
 const dataContainer: DataContainer = new DefaultDataContainer()
 const store: Store<StateType, StoreActionType> = createReduxStore(dataContainer)
 
 const App = () => {
+  const [routeName, setRouteName] = useState<?string>(null)
+  const [routeKey, setRouteKey] = useState<?string>(null)
+  const [cityCode, setCityCode] = useState<?string>(null)
+  const [languageCode, setLanguageCode] = useState<?string>(null)
```
maxammann, line 27: Shouldn' these two be in the redux state instead of the component state here? Isn't there duplicated data now?

native/src/modules/app/components/NavigationTypes.js
```diff
@@ -5,59 +5,59 @@ import type { RouteProp } from '@react-navigation/native'
 import type { FeedbackInformationType } from '../../../routes/feedback/containers/FeedbackModalContainer'
 import { LanguageModel, OfferModel } from 'api-client'
 
-export type IntroRouteType = 'Intro'
```
maxammann, line 8: minor: From my react-navigation experience it is common to use uppercase route names

native/src/modules/app/components/Navigator.js
```diff
@@ -123,18 +126,52 @@ const Navigator = (props: PropsType) => {
...
-  // TODO Snackbar
+  const dashboardParams = initialRoute.name === DASHBOARD_ROUTE
+    ? {
+        cityCode: initialRoute.cityCode,
+        languageCode: initialRoute.languageCode,
+        shareUrl: cityContentUrl({ cityCode: initialRoute.cityCode, languageCode: initialRoute.languageCode })
+      }
+    : {}
+
+  // Keeps our previous transition on android. Fixes weird showing of splash screen on every navigate.
```
maxammann, line 164:
```suggestion
  // Keeps our previous transition we used in v4 of react-navigation on android. Fixes weird showing of splash screen on every navigate.
```

native/src/modules/app/components/Navigator.js
```diff
@@ -123,18 +126,52 @@ const Navigator = (props: PropsType) => {
     initialize().catch(error => setErrorMessage(error.message))
   }, [])
 
+  // The following is used to have correct mapping from categories route mapping in redux state to the actual routes
+  useEffect(() => {
```
maxammann, line 130: pew can't we just fetch the new categories in when pressing the button on the landing page? For what did you use the key variable previously? Can we maybe just move the key to the params of a route?

native/src/modules/common/components/LayoutedScrollView.js
```diff
@@ -0,0 +1,19 @@
+// @flow
```
maxammann, line 1: LayoutContainer is a container. That means this should also be a container. Maybe you can just move this component to the top of withPayloadProvider.js as this is really not that reusable.

native/src/routes/disclaimer/DisclaimerContainer.js
```diff
@@ -102,20 +104,18 @@ class DisclaimerContainer extends React.Component<DisclaimerPropsType, Disclaime
     const { disclaimer, error, timeoutExpired } = this.state
 
     if (error) {
-      return <ScrollView refreshControl={<RefreshControl onRefresh={this.loadDisclaimer} refreshing={false} />}
-                         contentContainerStyle={{ flexGrow: 1 }}>
+      return <LayoutedScrollView refreshControl={<RefreshControl onRefresh={this.loadDisclaimer} refreshing={false} />}>
         <FailureContainer error={error} tryAgain={this.loadDisclaimer} />
-      </ScrollView>
+      </LayoutedScrollView>
```
maxammann, line 109: Why is LayoutedScrollView.js used here? before the change there way no LayoutContainer

native/src/routes/settings/components/Settings.js
```diff
@@ -134,7 +135,8 @@ export default class Settings extends React.Component<PropsType, StateType> {
       return null
     }
 
-    return <SectionList
+    return <LayoutContainer>
```
maxammann, line 138: Adding a container here means that Settings becomes a Container. This should be moved to the upper level in the component tree.

native/src/routes/sprungbrett/containers/SprungbrettOfferContainer.js
```diff
@@ -110,29 +112,29 @@ class SprungbrettOfferContainer extends React.Component<SprungbrettPropsType, Sp
     const { jobs, error, timeoutExpired } = this.state
 
     if (error) {
-      return <ScrollView refreshControl={<RefreshControl onRefresh={this.loadSprungbrett} refreshing={false} />}
-                         contentContainerStyle={{ flexGrow: 1 }}>
-        <FailureContainer errorMessage={error.message} tryAgain={this.loadSprungbrett} />
-      </ScrollView>
+      return (
+        <LayoutedScrollView refreshControl={<RefreshControl onRefresh={this.loadSprungbrett} refreshing={false} />}>
```
maxammann, line 116: LayoutContainer was added here implicitly. Is that correct?

maxammann (approved):

native/src/modules/app/components/App.js
klinzo: Yes, I know. Did not work for me at least.

native/src/modules/app/components/App.js
klinzo: Yes, they are. But then we need the whole mapping similar to DashboardContainer and need to check every route (Category, Event, ....). It is already an ugly hack, so imo this does not make it worse.

native/src/modules/app/components/Navigator.js
klinzo: We need the key of the route. Otherwise the mapping does not work.

klinzo: Moving the key to params does not work as the key is already a property of the route property and would therefore lead to way more confusion. This whole workaround is needed because the ability to replace a route with a route with predefined key was removed.

native/src/routes/disclaimer/DisclaimerContainer.js
klinzo: Otherwise the background/surroundings have the wrong color as nothing is rendered there. Did not find a better solution.

native/src/routes/sprungbrett/containers/SprungbrettOfferContainer.js
klinzo: Yes, see answer above.

native/src/modules/app/components/NavigationTypes.js
klinzo: Hmmm. I changed to lowercase to have a direct mapping to urls.

native/src/modules/common/components/LayoutedScrollView.js
klinzo: Moved to `containers`. It is reusable and is reused, see e.g. `SprungbrettContainer`.

native/src/routes/settings/components/Settings.js
klinzo: What do you mean by `should be moved to upper level in the component tree`?

native/src/routes/disclaimer/DisclaimerContainer.js
klinzo: Wrapping in a `LayoutContainer` was already done implicitly before as well, just not for all components. Not sure why that worked at all.

klinzo:
@maxammann I really like your comment to not pass the navigation/route to components, thanks :+1: 
Quite some stuff to review again though :/

klinzo:
This whole stuff should be merged asap imo.

Taggotty:
Only found 1 warning when navigating through the app, which seemed to occur for non-external offers. Henceforth, Wohnraumbörse and Sprungbrett. Please check this out again :)

Warning: Non-serializable values were found in the navigation state. Check:
sprungbrett > params.offers[1]._postData ([object Map])

This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you can use 'navigation.setOptions' instead.

native/src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -166,19 +169,27 @@ const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>): DispatchPropsT
...
-    cityCode, language, path, forceRefresh: true, key: route.key
+    cityCode, language, cityContentPath: path, forceRefresh: true, key: route.key
   })
 }
 
 class CategoriesContainer extends React.Component<ContainerPropsType> {
+  navigateToLinkProp = (url: string, language: string, shareUrl: string) => {
+    const { dispatch, navigation } = this.props
+    const navigateToInternalLink = createNavigateToInternalLink(dispatch, navigation)
+    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl || url)
```
Taggotty, line 182:
```suggestion
    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl)
```

native/src/routes/disclaimer/DisclaimerContainer.js
```diff
@@ -62,6 +66,12 @@ class DisclaimerContainer extends React.Component<DisclaimerPropsType, Disclaime
     this.state = { disclaimer: null, error: null, timeoutExpired: false }
   }
 
+  navigateToLinkProp = (url: string, language: string, shareUrl: string) => {
+    const { dispatch, navigation } = this.props
+    const navigateToInternalLink = createNavigateToInternalLink(dispatch, navigation)
+    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl || url)
```
Taggotty, line 72:
```suggestion
    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl)
```

native/src/routes/events/containers/EventsContainer.js
```diff
@@ -182,19 +184,26 @@ const ThemedTranslatedEvents = withTranslation('events')(
 )
 
 class EventsContainer extends React.Component<ContainerPropsType> {
+  navigateToLinkProp = (url: string, language: string, shareUrl: string) => {
+    const { dispatch, navigation } = this.props
+    const navigateToInternalLink = createNavigateToInternalLink(dispatch, navigation)
+    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl || url)
```
Taggotty, line 190:
```suggestion
    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl)
```

native/src/routes/search/containers/SearchModalContainer.js
```diff
@@ -58,13 +61,24 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType) => {
...
+const ThemedTranslatedSearch = withTheme(
+  withTranslation('search')(SearchModal)
 )
+
+const SearchModalContainer = (props: PropsType) => {
+  const { dispatch, navigation, ...rest } = props
+
+  const navigateToLinkProp = useCallback((url: string, language: string, shareUrl: string) => {
+    const navigateToInternalLink = createNavigateToInternalLink(dispatch, navigation)
+    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl || url)
```
Taggotty, line 75:
```suggestion
    navigateToLink(url, navigation, language, navigateToInternalLink, shareUrl)
```

Taggotty (approved):

native/src/routes/sprungbrett/containers/SprungbrettOfferContainer.js
```diff
@@ -110,29 +112,29 @@ class SprungbrettOfferContainer extends React.Component<SprungbrettPropsType, Sp
...
-      </ScrollView>
+      </LayoutContainer>
     }
 
     if (!jobs) {
       return timeoutExpired
-        ? <ScrollView refreshControl={<RefreshControl refreshing />} contentContainerStyle={{ flexGrow: 1 }} />
-        : null
+        ? <LayoutedScrollView refreshControl={<RefreshControl refreshing />} />
+        : <LayoutContainer />
```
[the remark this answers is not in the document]
klinzo: LayoutContainer does not display anything (or rather: just fills the screen with the right background color). LayoutedScrollView has scrolling and also push to refresh enabled and shows a loading spinner if refreshing is true.

native/src/modules/app/components/App.js
klinzo: Nice, good idea. Thanks!

native/src/routes/categories/containers/CategoriesContainer.js
klinzo: Thanks, good catches!

klinzo:
> Only found 1 warning when navigating through the app, which seemed to occur for non-external offers. Henceforth, Wohnraumbörse and Sprungbrett. Please check this out again :)
> 
> Warning: Non-serializable values were found in the navigation state. Check:
> sprungbrett > params.offers[1]._postData ([object Map])
> 
> This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you can use 'navigation.setOptions' instead.

@Taggotty thanks, completely forgot about that error! To keep changes small I just changed the following: https://github.com/Integreat/integreat-app/pull/105/commits/21ed3911d1159963457dcec760d39a9f7f701eea
This prevents passing all offers (which may contain non serializable post data) by just passing the relevant offer.

native/src/routes/settings/components/Settings.js
maxammann: Settings is in the components directory. By using LayoutContainer the component is connected to the redux state. So eithger move Settings to container directory or fixis by not using layoutcontainer there.

This already was a problem before so I'm fine with not fixing it.

klinzo: Actually is LayoutContainer just there because of `withTheme`, no redux.

PR #108, opened 2021-01-05
IGAPP-194: Update Version of iOS in E2E test

Taggotty:

See discussion here: https://chat.tuerantuer.org/digitalfabrik/pl/e7nedbz6g3869m376e9n3t86tr

I also tried the version 1.18.0 and 1.19.1 but both seemed worse performance wise, as they both seemed to depend on some cached files on the device. Somthing that somehow still throws me off is [this build](https://app.circleci.com/pipelines/github/Integreat/integreat-app/965/workflows/bec0546d-bfbb-4438-a429-d78e555d6daa/jobs/5219) where an unexpected timeout occurs. Besides that the native build failed I cannot see any reason why there would be a timout :/. Maybe one of you has an idea.

### DIFF
--- a/native/e2e/config/configs.js
+++ b/native/e2e/config/configs.js
@@ -57,12 +57,13 @@ exports.browserstack_dev_ios = {
     'browserstack.user': process.env.E2E_BROWSERSTACK_USER,
     'browserstack.key': process.env.E2E_BROWSERSTACK_KEY,
     project: 'integreat-react-native-app',
-    os_version: '11',
+    os_version: '12',
     device: 'iPhone 8',
     real_mobile: 'true',
-    'browserstack.appium_version': '1.16.0',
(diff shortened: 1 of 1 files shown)


klinzo:
The build mentioned in the PR description had the old appium version though, right? So maybe this would not have been happened with these changes?

klinzo (approved):

maxammann (approved):

Taggotty:
> The build mentioned in the PR description had the old appium version though, right? So maybe this would not have been happened with these changes?

Yep. But this change mainly focuses on fixing the error that sometimes occured. The last PR, which added waitingForQuiescence, should have fixed timeouts. But as it already got a lot better and the mentioned build also had errors. That may have impacted the tests. So maybe it's still best to wait and see here, whether another timeout occurs.

PR #110, opened 2021-01-06
IGAPP-37: Social media preview

klinzo:


web/src/modules/common/containers/Helmet.js
```diff
@@ -27,12 +28,20 @@ class Helmet extends React.PureComponent<PropsType> {
...
     const { pageTitle, cityModel, metaDescription } = this.props
+    const previewImageUrl = new URL(buildConfig().shareBaseUrl)
+    previewImageUrl.pathname = buildConfig().icons.socialMediaPreview
 
     return <ReactHelmet>
       {pageTitle && <title>{pageTitle}</title>}
       {cityModel && !cityModel.live && <meta name='robots' content='noindex' />}
       {metaDescription && <meta name='description' content={metaDescription} />}
       {this.getLanguageLinks()}
+      {/* https://developers.facebook.com/docs/sharing/webmasters */}
```
SteffiStuffel, line 39: Is this comment helpful / necessary ?

maxammann: I tihnk it is good as it shows how the og: tags work.

maxammann (approved):
Does this also work for twitter an dothe rsocial meida pages? Which do we want to support?

web/src/modules/common/containers/Helmet.js
klinzo: I'll add a little more info here.

klinzo:
> Does this also work for twitter and other social media pages? Which do we want to support?

According to this post it works for facebook, twitter and linkedin at least. I think we just have to test afterwards whether this is working for the platforms we need as I don't see a way to test this without deploying first.

maxammann:
Oke cool, did not read the post :) that is very fine

klinzo:
> Oke cool, did not read the post :) that is very fine

Sorry, forgot to include the post :facepalm: https://ahrefs.com/blog/open-graph-meta-tags/

Taggotty (approved):

PR #111, opened 2021-01-07
IGAPP-411: Fix dashboard navigation items cut on small screens

SteffiStuffel:


klinzo:
This PR does fix this issue, right? https://issues.tuerantuer.org/browse/IGAPP-367

SteffiStuffel:
> This PR does fix this issue, right? https://issues.tuerantuer.org/browse/IGAPP-367

Yeah, seems to be the same issue as 411, if i get it right.

klinzo:
> > This PR does fix this issue, right? https://issues.tuerantuer.org/browse/IGAPP-367
> 
> Yeah, seems to be the same issue as 411, if i get it right.

Nice, then I'll reject it.

web/src/modules/layout/components/Header.js
```diff
@@ -92,7 +92,6 @@ const NavigationBar = styled.div`
   flex: 1 1 0%; /* The % unit is necessary for IE11 */
   align-items: stretch;
   justify-content: center;
-  overflow-y: hidden;
```
klinzo, line 95: AFAICS (according to git blame) this line was added to hide the scrollbar. Do you think there is another solution to this problem?

maxammann: I found this snipped to hide scrollbars but keep functionality.

maybe that works better for some weird css reason:

```
 /* Hide scrollbar for Chrome, Safari and Opera */
.example::-webkit-scrollbar {
  display: none;
}

/* Hide scrollbar for IE, Edge and Firefox */
.example {
  -ms-overflow-style: none;  /* IE and Edge */
  scrollbar-width: none;  /* Firefox */
} 
```

SteffiStuffel: I added the code Max suggested, but i can't see any difference. Tested on Safari on iOS, local Chrome-Browser and in the Android Emulator.
Which solution do you prefer / should we merge or rather do you see any difference?

Schedulaar: If I remember correctly, some scroll bar was shown on android... Not sure though

SteffiStuffel: Thanks for the hint with firefox. Tested this on ios and android emulator, but there is also not scrollbar.

ios firefox
android firefox

Taggotty: Could also not reproduce any scrollbar. @klinzo how old was the change with the scrollbar? Maybe we fixed it sometime in between.

klinzo: Okay, nice :+1: I just used firefox and set dimension to an iPhone :) But I guess this is fine then.

klinzo (approved):
You can write a release note if you want :)

release-notes/unreleased/IGAPP-411-fix-menu-bar-on-small-screens.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-363
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: Hide redundant navigation items.
+de: Redundante Navigationselemente werden nun nicht mehr angezeigt.
```
klinzo, line 7: Nothing against copy and paste but I guess you should change this :D

klinzo (changes requested):

release-notes/unreleased/IGAPP-411-fix-menu-bar-on-small-screens.yml
SteffiStuffel: How does this happen?? Sorry. I fixed it.

klinzo (approved):

maxammann (approved):

PR #112, opened 2021-01-14
IGAPP-373: Add search string to url

SteffiStuffel:


web/src/routes/search/containers/SearchPage.js
```diff
@@ -69,7 +73,10 @@ export class SearchPage extends React.Component<PropsType, LocalStateType> {
       .concat(categoriesWithContent)
   }
 
-  handleFilterTextChanged = (filterText: string) => this.setState({ filterText: filterText })
+  handleFilterTextChanged = (filterText: string) => {
+    this.setState({ filterText: filterText })
+    history.replaceState(null, '', `${this.props.location.pathname}?query=${filterText}`)
```
klinzo, line 78: Not sure whether something like `query-string.stringify({query: filterText}` would be better.

Additional observation: If filterText is empty, the url becomes `integreat.app/augsburg/de/search?query=`. Not sure if this should be handled separately...

klinzo (commented):
Works like a charm!
Do you think you could test this new behavior as we currently try to increase coverage. Not sure how difficult this is though.
Also a release note would be nice as this is a nice feature :+1:

web/src/routes/search/containers/SearchPage.js
maxammann: I think we  should remove the whole query string if query is empty.

maxammann (approved):

maxammann (approved):

release-notes/unreleased/IGAPP-373-add-search-query-to-url.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 373
+show_in_stores: true
```
klinzo, line 2: JFYI: As there are no stores on web, you can just set this to false and omit the german release note text below :)

web/src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -157,4 +157,30 @@ describe('SearchPage', () => {
...
+      )
+      searchPage.instance().handleFilterTextChanged('ChangeToThis')
+      expect(global.window.location.href).toMatch(/\?query=ChangeToThis/)
+    })
+    it('should remove ?query= when filteredText is empty', () => {
+      const searchPage = shallow(
+        <SearchPage categories={categories} location={location} t={t}/>
+      )
+      searchPage.instance().handleFilterTextChanged('')
+      expect(global.window.location.href).toMatch(/^((?!\?query=).)*$/)
```
klinzo, line 185: You like regex? :eyes: :+1:

release-notes/unreleased/IGAPP-373-add-search-query-to-url.yml
SteffiStuffel: Yeah i was wondering about that already. Thanks for the information :)

web/src/routes/search/containers/__tests__/SearchPage.spec.js
```diff
@@ -157,4 +157,30 @@ describe('SearchPage', () => {
       location
     })
   })
+
+  describe('Tests for url query', () => {
+    it('should set state from url', () => {
+      const location = createLocation({
+        type: SEARCH_ROUTE, payload: { city, language }, query: { query: 'SearchForThis' }
+      })
+      const searchPage = shallow(
```
klinzo, line 168: We want to convert our tests from using enzyme to those using the react testing library: https://testing-library.com/docs/react-testing-library/intro/. This is fine for now, in the future best use this library though.

klinzo (approved):

PR #114, opened 2021-01-15
IGAPP-51: Remove translations:missing Script

Taggotty:


translations/tools/manage
```diff
@@ -4,11 +4,11 @@ const program = require('commander')
 const fs = require('fs')
```
Taggotty, line 4: Only applied formating to this file

klinzo (approved):

maxammann (approved):

PR #115, opened 2021-01-15
IGAPP-429: Add Last Update

Taggotty:


maxammann (approved):
I will create a task today in the afternoon which will fix date formatting for de-si. This could break this this. So be aware to merge early (or later) :D

klinzo (commented):
Are you sure this works? I don't see any last update :see_no_evil: 
image

Also a release note would be nice :)

Taggotty:
> Are you sure this works? I don't see any last update  I think I am on the right branch...

Hmm. At least for me it worked. I will take another look on it. Thx for checking.

> Also a release note would be nice :)

👍🏼

Taggotty:
Will be closed for now as this should have been the implementation for native and not web.

PR #117, opened 2021-01-18
IGAPP-21: Add Feedback for Pois

Taggotty:


web/src/modules/layout/containers/LocationLayout.js
```diff
@@ -72,7 +73,7 @@ export class LocationLayout extends React.Component<PropsType, LocalStateType> {
     }
 
     const { cities, location, feedbackTargetInformation } = this.props
-
+    console.log(location.type)
```
klinzo, line 76: Could be removed

native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -144,6 +144,8 @@ class FeedbackModalContainer extends React.Component<ContainerPropsType, Feedbac
     const label = t('contentOfCity', { city: cityName })
     const feedbackCategory = CONTENT_FEEDBACK_CATEGORY
     const feedbackType = route.params.type
+
+    // TODO: add 'POIS' in IGAPP-438
```
klinzo, line 148: So feedback is just possible for single pois now, right?

maxammann (approved):

Znippet (approved):

native/src/routes/feedback/containers/FeedbackModalContainer.js
Taggotty: No. It is possible for both. But feedback for pois overall cannot differed from General feedback from the Category Route.

PR #119, opened 2021-01-18
IGAPP-424: ChromeCustomTabs crashing on mailto links

klinzo:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ commands:
                 default: app
                 type: string
             failure_mentions:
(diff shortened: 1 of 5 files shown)


Znippet (approved):
Unit tested - nice  :)

Is there already a task for the snackbar which should be shown?

klinzo:
> Unit tested - nice :)
> 
> Is there already a task for the snackbar which should be shown?

I don't think so, no. I guess Michelle will create the tasks for the snackbar. After the implementation we can use it here.

Taggotty (approved):

PR #120, opened 2021-01-18
IGAPP-18: Extend and use rtl array from locales

maxammann:
Instead of extending the array I went an other route and made sure we only use set the direction for language we know which script the people use.

That means:
* on Android we use what the operating system tells us
* in the browser we only set the direction flag if we know whether it is rtl or not
* I also removed ui direction form the state


translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
+   * If you do not know whether to display UI elements RTL/LTR you can use this method. Browsers for example, do not
+   * disclose the system script direction. On Android you can use the system libraries instead of this method. This only
+   * works for {@link #supportedLanguages} and not for arbitrary ones. This is because we simply do now know it and can
+   * not know it.
+   *
+   * @see http://www.i18nguy.com/temp/rtl.html
+   * @param languageCode for the check
+   * @returns {*} whether script is RTL
+   */
+  hasRTLScript (languageCode: string): boolean {
```
klinzo, line 106: I don't like the name script to be honest, but probably that's just because I am not a native speaker and haven't ever used the term script in this context.

translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
+    const language = this.getSupportedLanguage(languageCode)
+
+    if (!language) {
+      throw new Error(`Unable to determine whether ${languageCode} uses a RTL script.
```
klinzo, line 110: This should never happen because if we don't support a language we use our fallback, right?

translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
+  }
+
+  isSupportedLanguage (languageCode: string): boolean {
+    return !!this.getSupportedLanguage(languageCode)
   }
 
-  isRTLLanguage (languageCode: string): boolean {
-    return this.getRTLLanguages().includes(languageCode)
+  /**
+   * Checks whether the languageCode belongs has an RTL script. This decision is made by the project "Integreat".
```
klinzo, line 91:
```suggestion
   * Checks whether the languageCode has an RTL script. This decision is made by the project "Integreat".
```

translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
+   * Writing direction is not an attribute of "language", but of "scripts". That means that there are language which
+   * can have RTL and LTR scripts.
```
klinzo, line 93: Are there really? If there are and we use one of these languages, this makes not really sense I think. Why do we show a language with both rtl and ltr scripts as rtl? I also don't really see a point in having to scripts for one language in the apps as mixing is a really bad idea, so we should just decide for each language whether to use the rtl or ltr script.

translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
 
-  getRTLLanguages (): string[] {
-    return entries(this.supportedLanguages)
-      .filter(([_, languageConfig]) => languageConfig.rtl)
-      .map(([languageCode, _]) => languageCode)
+  getSupportedLanguage (languageCode: string): LanguageType | void {
+    const fallbacks = this.fallbacks[languageCode]
+
+    if (fallbacks) {
+      const found = fallbacks.find(fallback => !!this.supportedLanguages[fallback])
```
klinzo, line 76: So basically this just checks that there are not fallbacks specified which are not supported? Shouldn't we rather ensure (e.g. with tests or flow types) that fallbacks are always supported?

klinzo: Also, why are there (possibly) more than one fallback for one language?

klinzo:
> Instead of extending the array I went an other route and made sure we only use set the direction for language we know which script the people use.
> 
> That means:
> 
>     * on Android we use what the operating system tells us
> 
>     * in the browser we only set the direction flag if we know whether it is rtl or not
> 
>     * I also removed ui direction form the state

Why did you chose this solution? Because there are languages which have a rtl and a ltr script?

What about iOS?

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -88,9 +86,9 @@ export default ({ children }: PropsType) => {
     <I18nextProvider i18n={i18nextInstance}>
       <div
         data-testid={'direction'}
-        style={{
-          direction: config.isRTLLanguage(language) ? 'rtl' : 'ltr'
-        }}>
+        dir={config.isSupportedLanguage(language)
+          ? config.hasRTLScript(language) ? 'rtl' : 'ltr'
+          : undefined}>
```
Schedulaar, line 91: I don't really see a reason, why we want to remove the layout direction from the state. As you can see, this might very well be useful for some parts of the layout... or is there a replacement for it?
At least it should always match the direction that is set here by the I18nProvider.

maxammann: I removed it for simplicity as I thought there is no use of it anymore.

maxammann: I'll see whether I can refactor the only usage by using flexbox

translations/src/config.js
maxammann: i18next can work with multiple fallbacks. That means if a string is not availabel then it will look through all languages in that array.

maxammann: Yes this checks whether the requestsed language code is is a fallback of a supported language. That way it is also supported.

translations/src/config.js
maxammann: See the attached link. Scripts have directions not languages. I went this this name to highlight this.

translations/src/config.js
maxammann: It can happen if this method is used in a wrong way. If you pass the content language to this function then you can hit this exception.

translations/src/config.js
maxammann: We are deciding for each language whether we think that our uses have a ltr/rtl script.

It is very complicated to get to a list of language -> direction. In fact there is no such list. The linked article states this. We have to decide for our "Integreat" app specificly whether we thinkg users are rtl/ltr.

translations/src/config.js
```diff
@@ -68,18 +69,53 @@ class Config {
...
-    return this.getRTLLanguages().includes(languageCode)
+  /**
+   * Checks whether the languageCode belongs has an RTL script. This decision is made by the project "Integreat".
+   * Writing direction is not an attribute of "language", but of "scripts". That means that there are language which
+   * can have RTL and LTR scripts.
+   *
+   * <b>Typical Usage</b>
+   *
+   * If you do not know whether to display UI elements RTL/LTR you can use this method. Browsers for example, do not
+   * disclose the system script direction. On Android you can use the system libraries instead of this method. This only
```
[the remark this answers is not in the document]
maxammann: Same afaik. I will add this here.

translations/src/config.js
klinzo: Yeah but I think all language codes should only have one fallback. i.e. `de-si` should not have a fallback of `de` and `de-de`

maxammann: yes thats how it is configured atm. This is just how the config in translations js was defined. We don't have to use the full potential.

maxammann: That way we can simply pass it to i18next

web/src/modules/i18n/containers/I18nProvider.js
maxammann: I reverted the changes :)

web/src/modules/common/components/Breadcrumbs.js
```diff
@@ -33,24 +27,21 @@ const OrderedList = styled.ol`
...
-  static defaultProps = { direction: 'ltr' }
-
   render () {
-    const { direction, ancestorBreadcrumbs, currentBreadcrumb } = this.props
+    const { ancestorBreadcrumbs, currentBreadcrumb } = this.props
     // The current page should not be listed in the UI, but should be within the JsonLd.
     const jsonLdBreadcrumbs = [...ancestorBreadcrumbs, currentBreadcrumb]
 
-    return <Wrapper direction={direction}>
+    return <Wrapper>
```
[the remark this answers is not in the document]
maxammann: I reverted the changes

klinzo (approved):

Taggotty (approved):

PR #121, opened 2021-01-19
IGAPP-297: Add build config for aschaffenburg app

SteffiStuffel:


build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+  WebBuildConfigType
+} from '../BuildConfigType'
+import { ASCHAFFENBURG_ASSETS } from '../AssetsType'
+import mainImprint from './mainImprint'
+
+const APPLE_APP_ID = ''
+
+const commonAschaffenburgBuildConfig: CommonBuildConfigType = {
+  appName: 'Aschaffenburg App',
+  appIcon: 'app_icon_integreat',
```
klinzo, line 17: I think we should TODOs in places where not the right data is not available yet.

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+  assets: ASCHAFFENBURG_ASSETS,
+  cmsUrl: 'https://cms.integreat-app.de',
+  switchCmsUrl: 'https://cms-test.integreat-app.de',
+  shareBaseUrl: 'https://aschaffenburg.app',
+  allowedHostNames: ['cms.integreat-app.de', 'cms-test.integreat-app.de'],
+  internalLinksHijackPattern: 'https?:\\/\\/(cms(-test)?\\.integreat-app\\.de|web\\.integreat-app\\.de|integreat\\.app)(?!\\/[^/]*\\/(wp-content|wp-admin|wp-json)\\/.*).*',
+  featureFlags: {
+    pois: false,
+    newsStream: true,
+    pushNotifications: true,
```
klinzo, line 29: Currently we don't plan on providing this functionality for Aschaffenburg.
```suggestion
    newsStream: false,
    pushNotifications: false,
```

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+  googleServices: {
+    googleAppId: '',
+    gcmDefaultSenderId: '',
+    defaultWebClientId: '',
+    gaTrackingId: null,
+    firebaseDatabaseUrl: '',
+    googleApiKey: '',
+    googleCrashReportingApiKey: '',
+    projectId: 'aschaffenburg-2021'
+  }
```
klinzo, line 57: This is only necessary for push notifications. As we probably won't include this in the Aschaffenburg app, you can just set this to null:
```suggestion
  googleServices: null
```

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+    projectId: 'aschaffenburg-2021',
+    storageBucket: '',
+    isAdsEnabled: false,
+    isAnalyticsEnabled: false,
+    isAppInviteEnabled: true,
+    isGCMEnabled: true,
+    isSigninEnabled: true,
+    googleAppId: '',
+    databaseUrl: ''
+  },
```
klinzo, line 81: Same as above:
```suggestion
  googleServices: null,
```

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+  privacyUrls: {
+    default: 'https://integreat-app.de/datenschutz/',
+    en: 'https://integreat-app.de/en/privacy/'
+  }
+}
+
+export const androidAschaffenburgBuildConfig: AndroidBuildConfigType = {
+  ...commonAschaffenburgBuildConfig,
+  splashScreen: true,
+  applicationId: 'app.aschaffenburg',
```
klinzo, line 47: Not sure but I think I'd rather use `de.aschaffenburgapp` similar to malte. But let's wait for the opinion of @maxammann

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+    firebaseDatabaseUrl: '',
+    googleApiKey: '',
+    googleCrashReportingApiKey: '',
+    projectId: 'aschaffenburg-2021'
+  }
+}
+
+export const iosAschaffenburgBuildConfig: iOSBuildConfigType = {
+  ...commonAschaffenburgBuildConfig,
+  bundleIdentifier: 'app.aschaffenburg',
```
klinzo, line 62: Same as above: Not sure but I think I'd rather use `de.aschaffenburgapp` similar to malte

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+  },
+  privacyUrls: {
+    default: 'https://integreat-app.de/datenschutz/',
+    en: 'https://integreat-app.de/en/privacy/'
+  }
+}
+
+export const androidAschaffenburgBuildConfig: AndroidBuildConfigType = {
+  ...commonAschaffenburgBuildConfig,
+  splashScreen: true,
```
klinzo, line 46: Similar to malte this should be false as we don't want to design a separate splash screen :D

build-configs/aschaffenburg/index.js
```diff
@@ -0,0 +1,110 @@
...
+    appLogo: '/app-logo.png',
+    locationMarker: '/location-marker.svg',
+    appleTouchIcon: '/apple-touch-icon.png',
+    socialMediaPreview: '/social-media-preview.png',
+    favicons: '/favicons/'
+  },
+  splashScreen: {
+    backgroundColor: lightTheme.colors.themeColor,
+    imageUrl: '/app-icon-cornered.jpg'
+  }
```
klinzo, line 100: Same as above, we probably won't design a splash screen.

build-configs/aschaffenburg/mainImprint.js
```diff
@@ -0,0 +1,72 @@
+const mainImprint = `
```
klinzo, line 1: Probably it would be good to add a TODO.

klinzo (changes requested):

native/package.json
```diff
@@ -7,13 +7,16 @@
     "start:integreat-test-cms": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' react-native start --reset-cache",
     "start:integreat": "cross-env BUILD_CONFIG_NAME='integreat' react-native start --reset-cache",
     "start:malte": "cross-env BUILD_CONFIG_NAME='malte' react-native start --reset-cache",
+    "start:aschaffenburg": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native start --reset-cache",
```
klinzo, line 10: Not sure if we rather should use a abbreviation for these commands to make them shorter. Not sure about whether there is a reasonable one though.

SteffiStuffel: changed it to abapp

build-configs/aschaffenburg/index.js
SteffiStuffel: In the Jira ticket it is specified:

App bundle id (app.aschaffenburg)
App domain (aschaffenburg.app)

klinzo (approved):

build-configs/aschaffenburg/index.js
maxammann: app.aschaffenburg is fine as aschaffenburg.app will be the offical app domain

build-configs/aschaffenburg/index.js
maxammann: app.aschaffenburg is fine as aschaffenburg.app will be the offical app domain

klinzo: Then we did it wrong with malte :D But fine for me

native/package.json
```diff
@@ -7,13 +7,16 @@
     "start:integreat-test-cms": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' react-native start --reset-cache",
     "start:integreat": "cross-env BUILD_CONFIG_NAME='integreat' react-native start --reset-cache",
     "start:malte": "cross-env BUILD_CONFIG_NAME='malte' react-native start --reset-cache",
+    "start:abapp": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native start --reset-cache",
```
maxammann, line 10: I think we should use start:aschaffenburg here. At least I dont have a problem with the length.

maxammann (approved):

native/package.json
maxammann: Also abapp is short for aschaffenburg app not aschaffenburg :sweat_smile:

klinzo:
I think you still have to add the assets here: https://github.com/Integreat/integreat-app/blob/main/native/src/modules/app/constants/buildConfig.js

SteffiStuffel:
> @SteffiStuffel I think you still have to add the assets here: https://github.com/Integreat/integreat-app/blob/main/native/src/modules/app/constants/buildConfig.js

Yeah good point. Added this.

PR #124, opened 2021-01-20
IGAPP-429: Last update for inner categories

Taggotty:

### DIFF
--- a/native/src/modules/categories/components/CategoryList.js
+++ b/native/src/modules/categories/components/CategoryList.js
@@ -10,6 +10,9 @@ import CategoryListContent from './CategoryListContent'
 import type { PageResourceCacheEntryStateType, PageResourceCacheStateType } from '../../app/StateType'
 import { RESOURCE_CACHE_DIR_PATH } from '../../endpoint/DatabaseConnector'
 import { mapValues } from 'lodash'
+import Moment from 'moment'
+import { CategoryModel } from 'api-client'
+import { URL_PREFIX } from '../../platform/constants/webview'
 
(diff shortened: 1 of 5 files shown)


native/src/modules/categories/components/CategoryList.js
```diff
@@ -54,6 +58,17 @@ const CategoryThumbnail = styled(Image)`
  * Displays a ContentList which is a list of categories, a caption and a thumbnail
  */
 class CategoryList extends React.Component<PropsType> {
+  getCachedThumbnail (category: CategoryModel): ?string {
```
klinzo, line 61: Why did you have to add this? is this used anywhere?

klinzo (commented):
Hopefully we won't have to many merge conflicts with @maxammann s PR...

klinzo (approved):

native/src/modules/categories/components/CategoryList.js
Taggotty: I think I just forgot to remove it again.

Taggotty:
@klinzo @maxammann wouldn't it make sense to separate the search category list from the normal category list. First because it seams that they require different props and work differently. But mostly I think that this would generalize the search overall if we want to add something in the future.

klinzo:
@Taggotty if you think that makes sense sure. Haven't looked into it myself, but your points seem reasonable :+1: Maybe we should do that in a separate PR after @maxammann merged his big refactoring PR though, what do you think?

maxammann (approved):
will merge asap

PR #127, opened 2021-01-21
IGAPP-288: Skip landing aschaffenburg

klinzo:


web/src/modules/app/containers/Switcher.js
```diff
@@ -32,8 +32,10 @@ import toggleDarkModeAction from '../../theme/actions/toggleDarkMode'
 import LanguageFailure from '../../common/containers/LanguageFailure'
```
klinzo, line 32: If I see this I really think we have to rewrite the webapp at some point...

klinzo:
Please really checkout the app and let me know if you find any way to still access any other city in the aschaffenburg app. I'll fix internal linking in a separate PR.

SteffiStuffel:
Just seeing the lots of commits here, if you want to avoid them you can append changes to a pushed commit by:

git add .
git commit --amend --no-edit
git push

in my opinion this is clearer sometimes.

klinzo:
> Just seeing the lots of commits here, if you want to avoid them you can append changes to a pushed commit by:
> 
> git add .
> git commit --amend --no-edit
> git push -f
> 
> in my opinion this is clearer sometimes.

Actually I like having a good commit overview with detailed information on why changes have been made. Imo this information gets lost if commits are squashed. I also never found it confusing because of lots of commits yet. But I guess this is just my personal opinion. We could discuss this in the app channel or in the next team call of course though.

SteffiStuffel:
> Actually I like having a good commit overview with detailed information on why changes have been made. Imo this information gets lost if commits are squashed. I also never found it confusing because of lots of commits yet. But I guess this is just my personal opinion. We could discuss this in the app channel or in the next team call of course though.

Yeah you are right. Can also make it more clear because of the additional information.

build-configs/BuildConfigType.js
```diff
@@ -3,11 +3,21 @@
 import type { ThemeType } from './ThemeType'
 import type { TranslationsType } from 'translations'
 
+// Prevent enabled intro slide in combination with a preselected city.
+// If you change this make sure you are not navigating to the landing screen upon closing the intro slides.
+export type SelectedCityType = {|
+  introSlides: false,
+  selectedCity: string | null // If set, the city is automatically preselected. Changing it is not possible for users.
```
maxammann, line 10: Should we name this `preselectedCity` maybe? Or `fixedCity`?

maxammann: This could case confusion with the name in our State which is also `selectedCity`

build-configs/BuildConfigType.js
```diff
@@ -3,11 +3,21 @@
...
+|} | {|
+  introSlides: true,
+  selectedCity: null
```
maxammann, line 13: maybe `undefined` is better here. Thay way you can just omit the prop

web/src/modules/app/containers/Switcher.js
```diff
@@ -109,31 +111,43 @@ export class Switcher extends React.Component<PropsType> {
...
+      const showHeader = invalidLanguage || routeConfig.requiresHeader
+      const showFooter = (invalidLanguage || routeConfig.requiresFooter) && !isLoading
       return (
         <Layout footer={showFooter && <GeneralFooter language={language} />}
-                header={(error || routeConfig.requiresHeader) && <GeneralHeader viewportSmall={viewportSmall} />}
+                header={showHeader && <GeneralHeader viewportSmall={viewportSmall} />}
                 darkMode={darkMode}>
-          {error
-            ? <LanguageFailure cities={citiesPayload.data}
+          {invalidCity
```
maxammann, line 144: I really dont like these double ternaries :D

maxammann: But this is just personal preference :D

web/src/modules/layout/containers/LocationHeader.js
```diff
@@ -50,14 +50,21 @@ export class LocationHeader extends React.Component<PropsType> {
...
     const { city, language } = location.payload
-    return [
+    const actionItems = [
       <HeaderActionBarItemLink key='search' href={new SearchRouteConfig().getRoutePath({ city, language })}
                                text={t('search')} iconSrc={searchIcon} />,
-      <HeaderActionBarItemLink key='location' href={new LandingRouteConfig().getRoutePath({ language })}
-                               text={t('changeLocation')} iconSrc={landingIcon} />,
       <LanguageSelector key='language' languageChangePaths={languageChangePaths} isHeaderActionItem
                         location={location} />
     ]
```
maxammann, line 58:
```suggestion
      ...(!buildConfig().featureFlags.selectedCity ? : <HeaderActionBarItemLink key='location' href={new LandingRouteConfig().getRoutePath({ language })}
                                 text={t('changeLocation')} iconSrc={landingIcon} /> : [])
      <LanguageSelector key='language' languageChangePaths={languageChangePaths} isHeaderActionItem
                        location={location} />
    ]
```

maxammann (approved):
Not yet tested

maxammann:
Do we already have build configs and intellij configs for aschaffenburg? If not can you add them here?

build-configs/BuildConfigType.js
klinzo: I thought about `preselectedCity` as well, the reason I did not went for it is that it sounds like it is selected on first start but can be changed afterwards. I'm fine with `fixedCity` though.

klinzo:
> Do we already have build configs and intellij configs for aschaffenburg? If not can you add them here?

What do you mean by `build configs`? @SteffiStuffel added an aschaffenburg build config in the target branch.

Will create an issue for intelliJ configs.

build-configs/BuildConfigType.js
klinzo: Actually flow is stupid and gives you this anyway:
Error:(25, 62) Cannot assign object literal to `commonIntegreatE2eBuildConfig` because property `fixedCity` is missing in  object literal [1] but exists in  `FeatureFlagsType` [2] in property `featureFlags`.

maxammann:
Tested this on web and it works. On android I get this error:

image

klinzo:
> Tested this on web and it works. On android I get this error:
> 
> image

Has to be fixed in the target branch by @SteffiStuffel :)

SteffiStuffel:
> > Tested this on web and it works. On android I get this error:
> > image
> 
> Has to be fixed in the target branch by @SteffiStuffel 

Fixed it. :)

SteffiStuffel:
> 
> Has to be fixed in the target branch by @SteffiStuffel :)

Done :)

web/src/modules/app/containers/Switcher.js
```diff
@@ -49,7 +51,7 @@ type PropsType = {|
   sprungbrettJobsPayload: Payload<Array<SprungbrettOfferPage>>,
   wohnenOffersPayload: Payload<Array<WohnenOfferModel>>,
   disclaimerPayload: Payload<PageModel>,
-  languages: ?Array<LanguageModel>,
+  languagesPayload: Payload<Array<LanguageModel>>,
```
SteffiStuffel, line 54: Why did you change this? Why is Payload now necessary or why wasn't if before?

SteffiStuffel:
> Please really run web and native and let me know if you find any way to still access any other city in the aschaffenburg app. I'll fix internal linking in a separate PR. Please try some routes in web as well (e.g. no path, landing, other cities,...).
> I think I also fixed invalid cities in web: https://webnext.integreat.app/asdf/as

Tested it and found no way of changing the city.

web/src/modules/app/containers/Switcher.js
klinzo: Good question. In fact, it is not really necessary now as it is only checked on the data anyway. However, I though we need to check for errors/loading of the languages as well, which was not the case. Left it like this for consistency though.

Taggotty (approved):

PR #128, opened 2021-01-22
IGAPP-40: Build config documentation

klinzo:

### DIFF
--- a/build-configs/BuildConfigType.js
+++ b/build-configs/BuildConfigType.js
@@ -3,11 +3,15 @@
 import type { ThemeType } from './ThemeType'
 import type { TranslationsType } from 'translations'
 
+// Build Configs
+// These are the flow types of our build configs and therefore define the structure and available options.
+// Each build config (e.g. integreat, malte) is available per platform (android, ios, web) with some shared options.
+
 // Prevent enabled intro slide in combination with a fixed city.
 // If you change this make sure you are not navigating to the landing screen upon closing the intro slides.
 export type FixedCityType = {|
-  introSlides: false,
-  fixedCity: string | null // If set, the city is automatically preselected. Changing it is not possible for users.
+  introSlides: false, // Shows intro slides to the users on first app start.
+  fixedCity: string | null // Preselects a city without showing a selection, changing it is not possible for users.
 |} | {|
   introSlides: true,
   fixedCity: null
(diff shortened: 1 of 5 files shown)


klinzo:
I figured an documentation directly in the code would be better in this case as you really have to change it if you change behavior. Would you rather like a separate documentation?

build-configs/BuildConfigType.js
```diff
@@ -3,61 +3,70 @@
 import type { ThemeType } from './ThemeType'
 import type { TranslationsType } from 'translations'
 
+/**
```
maxammann, line 6:
```suggestion
/*
```

As this is not a jsdoc comment

build-configs/BuildConfigType.js
```diff
@@ -3,61 +3,70 @@
...
+ * Build Configs
+ * These are the flow types of our build configs and therefore define the structure and available options.
```
maxammann, line 8: I think we should explain here the difference between build options and feature flags

build-configs/BuildConfigType.js
```diff
@@ -3,61 +3,70 @@
...
 |}
 
 export type FeatureFlagsType = {|
   ...SelectedCityType,
-  pois: boolean,
-  newsStream: boolean,
-  pushNotifications: boolean,
-  sentry: boolean,
-  developerFriendly: boolean
+  pois: boolean, // Enables POIs and maps, can be disabled by cities.
```
maxammann, line 24:
```suggestion
  pois: boolean, // Enables POIs and maps. It is also possible to disable pois via our API for each city individually.
```

build-configs/BuildConfigType.js
```diff
@@ -3,61 +3,70 @@
...
+// Available on all platforms
 export type CommonBuildConfigType = {|
   appName: string,
   appIcon: string,
   cmsUrl: string,
-  switchCmsUrl?: string,
-  shareBaseUrl: string,
-  allowedHostNames: Array<string>,
-  internalLinksHijackPattern: string,
+  switchCmsUrl?: string, // Secondary api url to use, selectable by clicking ten times on the location marker (native).
```
maxammann, line 36:
```suggestion
  switchCmsUrl?: string, // Secondary api url to use, selectable by clicking ten times on the location marker (only works for native right now).
```

maxammann (approved):
If you want you can use JSDoc instead of plain comments. Apart from that this documentation should be in code :+1: 

https://jsdoc.app/tags-param.html

Taggotty (approved):

build-configs/README.md
```diff
@@ -7,24 +7,38 @@ We use the concept of build configs to build different versions of the app. A bu
...
+* [Structure and Available Options](#structure-and-available-options)
+* [Technical Information](#technical-information)
+* [Whitelabelling](#whitelabelling)
+
 ## Available Build Configs
 
 * **integreat**: Build config for [Integreat](https://integreat-app.de).
-* **malte**: Build config for the [Malte](https://www.malteser-werke.de/malte-app.html).
+* **malte**: Build config for [Malte](https://www.malteser-werke.de/malte-app.html).
+* **aschaffenburg/abapp**: TODO
```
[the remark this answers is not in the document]
maxammann: aschaffenburg!

PR #129, opened 2021-01-27
IGAPP-15: Setup E2E Tests

Taggotty:

Does not yet include CI.


web/e2e/pages/landing/LandingPage.js
```diff
@@ -0,0 +1,26 @@
...
+  driver: WebDriver
+  language: string
+
+  constructor (driver: WebDriver, langugage: 'de' | 'en' = 'de') {
+    this.driver = driver
+    this.language = langugage
+  }
+
+  ready (): Promise<void> {
+    return this.driver.get(`http://localhost:9000/landing/${this.language}`)
```
maxammann, line 16: I think we should make `http://localhost:9000` configurable in this task or the next one

web/e2e/driver/driver.js
```diff
@@ -0,0 +1,124 @@
+// @flow
+
+import { type WebDriver, Builder } from 'selenium-webdriver'
+import fetch from 'node-fetch'
+import childProcess from 'child_process'
+import serverConfigs from '../config/configs'
+import { clone } from 'lodash'
+import type { EndToEndConfigType } from '../config/configs'
+
+const BROWSERSTACK_EXHAUSTED_MESSAGE = 'All parallel tests are currently in use, including the queued tests. ' +
```
maxammann, line 10: Should we include browserstack stuff here? I suspect that no one will execute this with browserstack.

maxammann: Maybe it is cool for safari, IE etc.
Can we maybe refactor this to share the code with native?

web/docs/e2e-tests.md
```diff
@@ -0,0 +1,46 @@
+# End-to-End Tests (Web)
```
maxammann, line 1: I think we should add a short section here about how to run E2E tests locally.

maxammann (approved):

web/e2e/config/configs.js
```diff
@@ -0,0 +1,124 @@
...
+/*
+  LOCAL
+ */
+
+exports.local_chrome = ({
+  url: 'http://localhost:4444/wd/hub',
+  prefix: 'IG LOCAL',
+  caps: {
+    browserName: 'chrome', // selenium has lowercased browsernames
+    browserVersion: '88'
```
maxammann, line 38: Probably should be removed as we probably do not want to test against different versions of chrome

web/e2e/config/configs.js
```diff
@@ -0,0 +1,124 @@
...
+/*
+  CI
+ */
+
+exports.ci_chrome = ({
+  url: 'http://localhost:4444/wd/hub',
+  prefix: 'IG DEV',
+  caps: {
+    browserName: 'chrome',
+    browserVersion: '88',
```
maxammann, line 51: Probably should be removed as we probably do not want to test against different versions of chrome

maxammann (approved):

web/docs/e2e-tests.md
```diff
@@ -0,0 +1,46 @@
...
+If you don't have this access you can still [execute the tests locally](#Local-Testing). 
+Browserstack also interacts with selenium internally, but due to limited resources we decided against it for our web E2E tests.
+
+### Development Testing
+Development testing includes running the tests in browserstack. For that you need to have a browserstack account.
+Furthermore, you need to define a few Environment variables. It is recommended to create your own build configuration in IntelliJ for that. When editing the configuration you can now also set Environment Variables in the Environment Field.
+Required are:
+- `E2E_BROWSERSTACK_KEY` and `E2E_BROWSERSTACK_USER`, which you can find when logging in to Browserstack and selecting either the Automate or AppAutomate tab
+- `E2E_CONFIG` e.g. browserstack_dev_ie11 for more configs see [E2E-configurations](../e2e/config/configs.js)
+  After that you need to initiate the BrowserStackBinary server. For this download the recent Browserstack binary appropriate for your system and run the binary as [described here](https://www.browserstack.com/local-testing/automate). Finally, you can run the E2E-tests.
```
klinzo, line 33: You are a little inconsistent with upper/lowercase and `-` I think. But it's okay for me :)

web/e2e/config/configs.js
```diff
@@ -0,0 +1,124 @@
+// @flow
+
+export type SeleniumCapabilitiesType = {|
```
klinzo, line 3: maybe it would make sense to include a type string, e.g. `type: 'selenium'`

web/e2e/pages/landing/LandingPage.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+import type { WebDriver, WebElement } from 'selenium-webdriver'
+import { By } from 'selenium-webdriver'
+
+export default class LandingPage {
+  driver: WebDriver
+  language: string
+
+  constructor (driver: WebDriver, langugage: 'de' | 'en' = 'de') {
```
klinzo, line 10: Why should this just be `de` and `en`

web/e2e/tests/routes/Landing.e2e.js
```diff
@@ -0,0 +1,33 @@
+// @flow
```
klinzo, line 1: I think it should be lowercase: `landing.e2e.js`

web/e2e/driver/driver.js
```diff
@@ -0,0 +1,124 @@
...
+
+  if (!config) {
+    console.error(`Server config ${configName} not found!`)
+    return null
+  }
+
+  return config
+}
+
+const initDriver = async (config: EndToEndConfigType, desiredCaps): Promise<WebDriver> => {
```
klinzo, line 35: Why no type for desiredCaps?

klinzo (approved):
Nice! :tada:

klinzo:
Should we create issues to create more e2e tests? I think the list by @SteffiStuffel would be a good orientation: https://chat.tuerantuer.org/digitalfabrik/pl/48918pes7tnxjd1d73uwu3z3er

web/e2e/pages/landing/LandingPage.js
Taggotty: I kind of agree. But as I thought we agreed that with starting the server on a local port, even in the ci it is actually not necessary. What would be the actual advantage of setting `http://localhost:9000` configurable?

maxammann: If you run the tests in a separate selenium container then the app is no longer running on localhost form the perspective of selenium.

Then it is running e.g. on 172.17.0.5

web/e2e/pages/landing/LandingPage.js
maxammann: I think this is fine for now as this task is only about the setup.

web/e2e/config/configs.js
maxammann: All of this are selenium configs. So we rather name this `DefaultCapabilitesTypes`?

web/e2e/tests/routes/Landing.e2e.js
maxammann: I think this is fine for now as this task is only about the setup.

web/e2e/driver/driver.js
maxammann: They can be arbitrary objects. But we could define a type for them. If so then this change should also be done in native

web/e2e/config/configs.js
SteffiStuffel: Wouldn't this be helpful, when we get a bug report for a specific browser-version?

web/docs/e2e-tests.md
Taggotty: Something differen from `### Local testing` below? Please add more infos. IS it to texty too short, or should I just include docker as a solution?

web/e2e/config/configs.js
Taggotty: > All of this are selenium caps. So we rather name this DefaultCapabilitesTypes?

I prefer having the name explicitly here as otherwise this information is lost, as the capabilities are only used for setup and shouldn't be touched later in the tests.

> maybe it would make sense to include a type string, e.g. type: 'selenium'

What would be the advantage of that? :)

web/e2e/driver/driver.js
Taggotty: for now we do not have any use case for desired caps in web anyway. I just added them as native also has them. Should I remove them for now?

web/e2e/config/configs.js
klinzo: I thought it would make it clearer which kind of caps it is. But nvm, everything good as it is.

web/e2e/driver/driver.js
klinzo: Your call :) I think leaving it makes sense for consistency.

web/e2e/config/configs.js
Taggotty: Removed the browser version for local testing, as probably everyone has a different version. For the ci and browserstack I kept it, as there the version should be fixed (and for browserstack required).

web/package.json
```diff
@@ -127,6 +128,7 @@
     "raf": "^3.4.1",
     "react-hot-loader": "^4.12.19",
     "redux-mock-store": "^1.5.3",
+    "selenium-webdriver": "^3.6.0",
```
maxammann, line 131: Is there a reason for selenium-webdriver instead of https://webdriver.io/ ?
In native we use the outdated wd dependency which I may want to replace.

Taggotty: `webdriver.io` also wraps a testing framework. As we are already using jest for testing I didn't consider it. As for `selenium-webdriver` my main reasoning is based on the fact that it is managed by `Selenium` itself and should therefore be up-to-date for the most time. If there is a general consensus to use `webdriver.io` we can of course switch.

Taggotty:
@maxammann  and I agreed that webdriver.io probably is simpler to use and gets rid of setup in web and native. Therefore, we will change the implementation to use webdriver.io.

PR #131, opened 2021-01-27
IGAPP-186: Changed every occurence of Image to FastImage.

SimonVortkamp:
You can find it here: https://issues.tuerantuer.org/browse/IGAPP-186

### DIFF
--- a/native/src/modules/common/components/NewsHeader.js
+++ b/native/src/modules/common/components/NewsHeader.js
@@ -12,8 +12,9 @@ import { LOCAL_NEWS, TUNEWS } from '../../endpoint/constants'
 import activeInternational from '../../../routes/news/assets/tu-news-active.svg'
 import inactiveInternational from '../../../routes/news/assets/tu-news-inactive.svg'
 import { TFunction } from 'i18next'
+import FastImage from 'react-native-fast-image'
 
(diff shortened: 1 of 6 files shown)


klinzo (changes requested):
This somehow broke the images, e.g.
* IntroSlide app icon, is not using the right icon anymore, same for integreat icon on dashboard (or: the wrong format)
* Second intro slide icon is a little bit cut in left bottom corner
* No location marker is shown anymore on the landing page, same for navigation tiles on dashboard and feedback icons

klinzo:
There is already a custom Image component, maybe you can use it. But not sure about that.

maxammann:
Seems like RN fast image does not support this: https://github.com/DylanVann/react-native-fast-image/issues/248
I think we should revert this to Image for everything which uses images from an `import`. 

We could also refactor our Image component and then use this one everywhere:
`./native/src/modules/common/components/Image.js`

There is a check: typeof uri === 'number'
If that is true then we simply could use the Image component from React-Native. If that is false then continue and use FastImage.


We could also just remove FastImage everywhere and just use Image to simplify our code.

maxammann (changes requested):
I'm sorry simon, but I think we have to adjust here as the solution in the issue does not work :D

PR #133, opened 2021-01-29
IGAPP-299: Aschaffenburg Icons and Assets

Taggotty:


build-configs/aschaffenburg/assets/app-logo-alone.svg
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
klinzo, line 1: What does the `alone` mean in the file name?

build-configs/aschaffenburg/assets/location-marker.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 150">
```
klinzo, line 1: Do we require this with flow? Because it won't be used anyway as there won't be a landing page for aschaffenburg.

klinzo (approved):

build-configs/aschaffenburg/assets/app-logo-alone.svg
klinzo: I think we should use the same naming for this and for the inverted integreat G. I don't mind which one though.

build-configs/aschaffenburg/index.js
```diff
@@ -62,9 +62,8 @@ export const webAschaffenburgBuildConfig: WebBuildConfigType = {
   ...commonAschaffenburgBuildConfig,
   mainImprint,
   itunesAppId: APPLE_APP_ID,
-  manifestUrl: '/manifest.json',
   icons: {
-    appLogo: '/app-logo.png',
+    appLogo: '/app-logo.svg',
```
maxammann, line 66: Are we sure an svg works here? not sure where this is used.

maxammann (approved):

klinzo (approved):

build-configs/aschaffenburg/assets/location-marker.svg
Taggotty: removed it :)

build-configs/aschaffenburg/index.js
Taggotty: hmm changed it to png. Just in case. But I think this is only used in web an should not cause many problems, as the other assets also contain svgs

PR #134, opened 2021-01-29
IGAPP-136: Internal linking - pathname parser

klinzo:

First part of a refactored, consistent and generic internal linking. This moves all route types to the api-client so they can be used at a later point in web as well. It also adds an InternalRouteParser which parses internal urls and returns information for the corresponding route.


native/src/modules/navigation/InternalPathnameParser.js
```diff
@@ -0,0 +1,199 @@
...
+    this._parts = this.pathnameParts(pathname)
+    this._length = this._parts.length
+    this._languageCode = languageCode
+  }
+
+  pathnameParts = (pathname: string) => {
+    return pathname.split('/').filter(Boolean)
+  }
+
+  isFixedCity = () => {
```
maxammann, line 44: Why is this called fixedCity and not just city? Should this be `isCityFixed`?

native/src/modules/navigation/InternalPathnameParser.js
```diff
@@ -0,0 +1,199 @@
...
+const FEATURE_ID_INDEX = 3
+
+class InternalPathnameParser {
+  _pathname: string
+  _parts: Array<string>
+  _length: number
+  _languageCode: string
+  _fixedCity: string | null
+
+  constructor (pathname: string, languageCode: string, fixedCity: string | null) {
```
maxammann, line 32: Why is languageCode a part of the constructor? Shouldn't his be choosed automatically by parsing the pathname?

native/src/modules/navigation/InternalPathnameParser.js
```diff
@@ -0,0 +1,199 @@
...
+  SPRUNGBRETT_OFFER_ROUTE,
+  TU_NEWS_TYPE
+} from 'api-client/src/routes'
+import type {
+  LocalNewsType,
+  TuNewsType
+} from 'api-client/src/routes'
+import type { RouteInformationType } from './createNavigate'
+
+const FEATURE_ID_INDEX = 3
```
maxammann, line 23: What does Feature ID mean here? It is used as type in some places. Is it the route type?

native/src/modules/navigation/InternalPathnameParser.js
maxammann: Is this maybe some kind of fallback for routes which do not contain a language?

maxammann (approved):

maxammann:
We could move this parser to the api-client. I think it would make sense there because path names are actually part of an API.

Maybe we can also make use of it in the web.

native/src/modules/navigation/InternalPathnameParser.js
klinzo: The latter, yes. There are pathnames without a language in it. And I did not want to have to check afterwards for all routes if there is already a language included.

native/src/modules/navigation/InternalPathnameParser.js
klinzo: No, it is the id of a single event/poi/..., e.g. for `/augsburg/de/events/1234` it is `1234.

native/src/modules/navigation/InternalPathnameParser.js
klinzo: Because this is just used to check if the city is the same as the fixed city from the build config (i.e. the one passed in the constructor). I think the terminology is right :see_no_evil:

SteffiStuffel:
Is this only for linking form native to web or should / can it also be used for navigation (issues) in native?

klinzo:
> Is this only for linking form native to web or should / can it also be used for navigation (issues) in native?

@SteffiStuffel currently this is meant to parse web urls/pathnames (e.g. https://integreat.app/augsburg/de/events/2134) and return all relevant information to be able to open the corresponding page in the native app. However, of course it can be used for other things if it helps there. Not sure if I understand the question though :)

SteffiStuffel:
> > Is this only for linking form native to web or should / can it also be used for navigation (issues) in native?
> 
> @SteffiStuffel currently this is meant to parse web urls/pathnames (e.g. https://integreat.app/augsburg/de/events/2134) and return all relevant information to be able to open the corresponding page in the native app. However, of course it can be used for other things if it helps there. Not sure if I understand the question though :)

oke i see. I asked because I'm struggling a bit with the navigation stuff especially the routes in the store and was wondering if this PR is related to my troubles ;)

klinzo:
@SteffiStuffel not sure, but probably notthen. What are you struggling with?

SteffiStuffel:
> @SteffiStuffel not sure, but probably not, then. What are you struggling with?

May lets have a call about this or talk about it on coworking-wednesday?

klinzo:
> We could move this parser to the api-client. I think it would make sense there because path names are actually part of an API.
> 
> Maybe we can also make use of it in the web.

will do in #135 . Somehow git makes strange things here.

native/src/modules/navigation/InternalPathnameParser.js
maxammann: Ah I did not get that fixedCity is the naming from the buildConfig

native/src/modules/navigation/InternalPathnameParser.js
maxammann: Can you rename this to fallback? like this it is very unclear to me

native/src/modules/navigation/InternalPathnameParser.js
maxammann: Minor: So `ENTITY_ID_INDEX`?

native/src/modules/navigation/InternalPathnameParser.js
klinzo: Will do in #135

native/src/modules/navigation/InternalPathnameParser.js
klinzo: Will do in #135

Znippet (approved):

PR #135, opened 2021-01-29
IGAPP-136: Internal linking

klinzo:

This PR generalizes internal linking and navigating by using the types and pathname parser from #134.
It also generalizes the way news types are used.


klinzo:
Testable at testumgebung > Test mit Links

native/src/routes/news/containers/NewsContainer.js
```diff
@@ -75,15 +75,17 @@ const onRouteClose = (routeKey: string, dispatch: Dispatch<StoreActionType>) =>
...
-    forceRefresh: true,
-    key: route.key
-  })
+    newsType: selectedNewsType,
+    languageCode: language,
+    newsId: newsId || undefined
+  },
+  route.key,
+  true
+  )
```
maxammann, line 88: The formatting looks off for me here. I hink the `route.key` and `true` should be intended

maxammann:
```
  navigateTo({
    route: POIS_ROUTE,
    cityCode,
    languageCode: language,
    cityContentPath: path || undefined
  }, route.key, true)
```

This formatting is used elsewhere in this PR.

maxammann (approved):

klinzo:
Most of our routes do not support linking to other cities or languages (except from categories, events and pois). I asked @dkehne and he said cities never (intentional) link to other cities and languages though. Should we handle this and prevent this somehow or do you think it is just not supported and crashes/wrong behavior is acceptable if it is done anyway?

Also: I did encounter some crashes when navigating to the search, but were not able to reproduce it the last times I tried.

native/package.json
```diff
@@ -82,7 +82,7 @@
     "react-native-static-server": "^0.5.0",
     "react-native-svg": "^12.1.0",
     "react-native-vector-icons": "^7.1.0",
-    "react-native-webview": "^10.9.1",
+    "react-native-webview": "10.8.2",
```
Znippet, line 85: Do we have a bug-task for this? Will hurt us if we try to upgrade native dependencies

Znippet (approved):

native/package.json
klinzo: Will create one :+1:

PR #136, opened 2021-01-30
IGAPP-173: Upgrade webpack and some minor packages

Znippet:


maxammann (approved):
The build is currently failing because jest is looking in the node_modules directory to run tests.

Maybe excluding node_modules is enough to fix it.

web/package.json
```diff
@@ -95,47 +95,47 @@
...
-    "url-loader": "^3.0.0",
-    "webpack": "^4.43.0",
-    "webpack-bundle-analyzer": "^3.8.0",
-    "webpack-cli": "^3.3.12",
-    "webpack-dev-server": "^3.11.0"
+    "url-loader": "^4.1.1",
+    "webpack": "^5.16.0",
+    "webpack-bundle-analyzer": "^4.3.0",
+    "webpack-cli": "^4.4.0",
+    "webpack-dev-server": "^3.11.2"
```
klinzo, line 139: Is this still needed as you use `webpack serve` now instead?

web/package.json
```diff
@@ -69,8 +69,8 @@
     "redux-thunk": "^2.3.0",
     "styled-components": "^5.1.1",
     "translations": "0.0.1",
-    "url-polyfill": "^1.1.5",
-    "whatwg-fetch": "^3.0.0"
+    "url-polyfill": "^1.1.12",
```
klinzo, line 72: What do we need this for? IE?

klinzo (approved):

web/package.json
Znippet: To be honest - no idea. But I just had a look into the git history:
image

web/package.json
Znippet: > The build is currently failing because jest is looking in the node_modules directory to run tests.
> 
> Maybe excluding node_modules is enough to fix it.

Added a new commit, but only for the problematic files

Znippet:
> The build is currently failing because jest is looking in the node_modules directory to run tests.
> 
> Maybe excluding node_modules is enough to fix it.

Added a new commit, but only for the problematic files

web/package.json
Znippet: the "serve" command also uses webpack-dev-server

web/package.json
klinzo: Haha yeah, good find :+1:

maxammann:
The ES-Check step failed because it detected features which are not supported by ES5. @Schedulaar do  you remember why we have this check? I think we added this because of IE compatability.

Schedulaar:
> The ES-Check step failed because it detected features which are not supported by ES5. @Schedulaar do  you remember why we have this check? I think we added this because of IE compatability.

That's correct.

Schedulaar:
> The ES-Check step failed because it detected features which are not supported by ES5. @Schedulaar do  you remember why we have this check? I think we added this because of IE compatability.

That's correct.

klinzo:
@Znippet what was the problem now?

Znippet:
> @Znippet what was the problem now?

The check failes in the paket "query-string", so I reverted its upgrade. But this is not the only problem I had. Most importantly, we can't update flow, suppress_comment is not supported any more

.flowconfig:37 Unsupported option specified! (suppress_comment)

pakets that didn't work / led to problems:
"copy-webpack-plugin", "webpack", "webpack-cli", "history"

I will update everything that is possible and does not lead to errors and after that upgrade this packages step by step

klinzo:
Okay, thanks for your explanation. Flow must not be updated without React. Each react version has a fixed flow version with which it should be used together.

PR #138, opened 2021-02-02
IGAPP-289: Deactivate nearby city feature

SteffiStuffel:


native/src/modules/app/LocationPermissionManager.js
```diff
@@ -2,15 +2,22 @@
 
 import { check, PERMISSIONS, request, RESULTS } from 'react-native-permissions'
 import { Platform } from 'react-native'
+import buildConfig from './constants/buildConfig'
 
 export const checkLocationPermission = async (): Promise<RESULTS> => {
+  if (buildConfig().featureFlags.fixedCity) {
+    throw Error('Can not check location permission when city is fixed')
+  }
```
klinzo, line 10: I think these methods are still called in the PermissionSnackbarContainer even if there is a fixed city. So I guess this may crash the app. Did you check that the app does not crash?

klinzo: I think you may have to check the `updateSettingsAndPermissions` method in `PermissionSnackbarContainer`

maxammann (approved):

native/src/modules/app/LocationPermissionManager.js
maxammann: Another idea: We could also extend the `RESULTS` type from react-native-permissions instead of throwing an exception

maxammann (approved):

native/src/modules/app/LocationPermissionManager.js
maxammann: I think this does not crash the app because this code if not reachable if `buildConfig().featureFlags.fixedCity`

klinzo: > I think this does not crash the app because this code if not reachable if `buildConfig().featureFlags.fixedCity`

Actually I think it is since the PermissionSnackbarContainer is still used :D But I may be wrong.

SteffiStuffel: You were both right 'buildConfig().featureFlags.fixedCity' prevented the PermissionSnackbar from showing, but the app crashed anyways because updateSettingsAndPermissions of PermissionSnackbarContainer was calling checkLocationPermission.

Anyway, I fixed it as suggested by Max by extending the RESULTS type.

native/src/modules/app/LocationPermissionManager.js
```diff
@@ -2,15 +2,22 @@
 
 import { check, PERMISSIONS, request, RESULTS } from 'react-native-permissions'
 import { Platform } from 'react-native'
+import buildConfig from './constants/buildConfig'
 
 export const checkLocationPermission = async (): Promise<RESULTS> => {
+  if (buildConfig().featureFlags.fixedCity) {
+    return RESULTS.DISABLED_FOR_FIXED_CITIES
```
klinzo, line 9: This option does not exist on `RESULTS`, does it? Even though it works, I think it should be defined. So imo you should create a custom result, e.g.:
```
const PERMISSION_RESULTS = {
  ...RESULTS,
  FORBIDDEN: 'forbidden'
}
```

klinzo (changes requested):

native/src/modules/app/LocationPermissionManager.js
SteffiStuffel: UNAVAILABLE, DENIED and BLOCKED do exist.

Is any of UNAVAILABLE suitable?

klinzo: I think that would work as well :+1:

SteffiStuffel: done

native/src/modules/layout/containers/PermissionSnackbarContainer.js
```diff
@@ -53,7 +53,7 @@ class PermissionSnackbarContainer extends React.Component<PropsType, StateType>
     const pushNotificationStatus = await checkPushNotificationPermission()
 
     const showLocationSnackbar = settings && settings.proposeNearbyCities === true &&
-      [RESULTS.BLOCKED, RESULTS.DENIED].includes(locationStatus) && this.landingRoute()
+      [RESULTS.BLOCKED, RESULTS.DENIED, RESULTS.UNAVAILABLE].includes(locationStatus) && this.landingRoute()
```
klinzo, line 56: Not sure if we want this. What should the user do about it? There is nothing the user can do about it as this is because
* the feature is not available at all on the device
* or we restrict it
I don't think it really matters, so good to go!

klinzo (approved):

native/src/modules/layout/containers/PermissionSnackbarContainer.js
maxammann: I would not show it for the unavailable state. So I would leave it as it is.

klinzo (approved):

maxammann:
Merge in main :)

PR #140, opened 2021-02-02
IGAPP-307: Aschaffenburg ci

klinzo:


klinzo:
I did already test development web delivery which can be viewed here: https://webnext.aschaffenburg.app.

maxammann (approved):

klinzo:
Native delivery seems to work fine as well. Since the app is not ready yet, store informations are missing and the app is apparently crashing on some devices, I'll skip native delivery in the CI for now.

.circleci/src/workflows/commit.yml
```diff
@@ -15,6 +15,20 @@ jobs:
         - mattermost
       requires:
         - bump_version
+  - build_web:
```
SteffiStuffel, line 18: Why do we specity the workflow here in commit.yml and also is config.yml? Is this duplication necessary?

SteffiStuffel:
Maybe it would also be good to updaten cicd.md (add aschaffenburg)

and also (although it's not part of this ticket) you maybe can remove "old" slack stuff (eg in 
- cicd.md (SLACK_URL and part about deliverino)
- config.yml/orbs (line 794)
- @common.yml/orbs
- notify_android.yml, notify_ios.yml and notify_web.yml in the comment) or we create a ticket for this.

SteffiStuffel (approved):

.circleci/src/workflows/commit.yml
klinzo: `commit.yml` is autogenerated from all files in `.circleci/src`. You should never modify `commit.yml` directly therefore.
The split in multiple files makes it way clearer instead of having one big config file.

klinzo:
> Maybe it would also be good to updaten cicd.md (add aschaffenburg)
> 
> and also (although it's not part of this ticket) you maybe can remove "old" slack stuff (eg in
> 
>     * cicd.md (SLACK_URL and part about deliverino)
> 
>     * config.yml/orbs (line 794)
> 
>     * @common.yml/orbs
> 
>     * notify_android.yml, notify_ios.yml and notify_web.yml in the comment) or we create a ticket for this.

Done with exception of adding aschaffenburg to our documentation as there is a separte issue: https://issues.tuerantuer.org/browse/IGAPP-432

PR #141, opened 2021-02-03
IGAPP-451: Redirect landing route to fixed city

klinzo:

### DIFF
--- a/web/src/modules/app/createReduxStore.js
+++ b/web/src/modules/app/createReduxStore.js
@@ -4,7 +4,7 @@ import type { Store } from 'redux'
 import { applyMiddleware, combineReducers, compose, createStore } from 'redux'
 import thunkMiddleware from 'redux-thunk'
 import type { RoutesMap } from 'redux-first-router'
-import { connectRoutes } from 'redux-first-router'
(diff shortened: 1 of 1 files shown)


maxammann (approved):
Isn't this already working when going to https://webnext.aschaffenburg.app/ ?

klinzo:
> Isn't this already working when going to https://webnext.aschaffenburg.app/ ?

Only partly. It's working for `/`, `/landing` and `/aschaffenburg`, but not for `/landing/de`. E.g. https://webnext.aschaffenburg.app/landing/de/ is not redirecting.

maxammann:
I actually would be fine as how it already is. I mean there should be no way of knowing that link. And if you manually try it then this is acceptable behavior imo.

As long as we do not have a href to that location we do not require that complexity.

SteffiStuffel:
Tested this, works as expected

klinzo:
I'd merge it, not too much complexity imo and I think unexpected behavior is never good. Other opinions?

SteffiStuffel (approved):

klinzo:
Won't do as this is not a url someone lands accidentally and it adds complexity.

PR #142, opened 2021-02-03
IGAPP-427: Themable fonts for body and heading

SteffiStuffel:


web/src/index.ejs
```diff
@@ -48,6 +48,8 @@
   <link rel="preload" href="/fonts/raleway/raleway-v12-latin_latin-ext-regular.woff2" as="font" type="font/woff2"
         crossorigin />
 
+  <link rel="stylesheet" href="/fonts/varela-round/varela-round.css" type="text/css" />
```
klinzo, line 51: Are fonts only loaded on demand?

klinzo (commented):
This was really easy :thinking: Did you check whether it works everywhere and the right fonts are used? AFAICS at least for web the fonts look very similar

web/src/index.ejs
SteffiStuffel: Yes they are. I was not sure whats the best way to handle it. It would be nice if Integreat would not load Aschaffenburg fonts.

SteffiStuffel:
> This was really easy  Did you check whether it works everywhere and the right fonts are used? AFAICS at least for web the fonts look very similar

Yes i tested on web an native, but not "everywhere". Seems fine to me, but feed free to test, too.

build-configs/aschaffenburg/theme/fonts.js
```diff
@@ -4,9 +4,9 @@ import type { FontsType } from '../../common/theme/fonts'
 import { commonFonts } from '../../common/theme/fonts'
 
 const fonts: FontsType = {
-  decorativeFontBold: 'Raleway-Bold',
-  decorativeFontRegular: 'Raleway-Regular',
-  decorativeFontFamily: '\'Raleway\', \'Lateef\', sans-serif',
+  decorativeFontBold: 'VarelaRound-Regular',
+  decorativeFontRegular: 'VarelaRound-Regular',
+  decorativeFontFamily: '\'Varela Round\', \'Raleway\', \'Lateef\', sans-serif',
```
maxammann, line 9:
```suggestion
  decorativeFontFamily: '\'Varela Round\', \'Lateef\', sans-serif',
```

Having Raleway here is equal to using the system default. So I'd omit it here.

web/src/index.ejs
maxammann: I think we should remove the preload for all fonts for now until we have a good solution.

maxammann (approved):

klinzo:
@SteffiStuffel @maxammann where is this font used? :thinking: I.e. where can I see it is used? :eyes:

SteffiStuffel:
> @SteffiStuffel @maxammann where is this font used?  I.e. where can I see it is used?

It is used in aschaffenburg app native and web? build-configs/aschaffenburg/theme/fonts?
You can just start the web-app inspect a text, select 'Elements' and 'computed' and on the bottom it shows 'rendered fonts'

not sure if i unterstood your question.

klinzo:
Yes sure. But I don't find text in web which actually uses Valera Round afaics, which has nothing to do with your PR I guess. Maybe we need to do some work here. In native everything looks as expected.

SteffiStuffel:
@klinzo That's confusing. I checked it out again and double checked the font. It uses Varela Round in webapp.
VarelaRound

klinzo (approved):
Okay no idea why it does not work for me. But nevermind then, good work :+1:

klinzo:
@Taggotty @maxammann @Znippet does it work for you?

Znippet:
> @Taggotty @maxammann @Znippet does it work for you?

Sorry, I didn't read this...  No, doesn't seem so:
image

I don't know if this helps:
image

So I got the font "varela-round-regular.woff2"
But seems I didn't get the "varela-round.css"

SteffiStuffel:
Thanks for testing and your feedback. Again I checked it on Chrome, Chrome private Tab, Firefox, and on my iPhone on Safari, everywhere the correct font is used. After some research I found if may depend on computer or the internet connection. 

https://stackoverflow.com/questions/12009423/what-does-status-canceled-for-a-resource-mean-in-chrome-developer-tools

Cause it does not break anything, i would suggest we merge this to main and see what happens on webnext?

klinzo (approved):
Now it works, good catch :+1:

PR #143, opened 2021-02-03
IGAPP-285: Firebase not initializing

maxammann:


maxammann:
So the fix was to include a firebase config for the test env. I explained why this happens here: https://github.com/invertase/react-native-firebase/issues/360

For this I also had to adjust the bundle ids as they were invalid under certain circumstances. I also added development certificates using fastlane match.

maxammann:
The cool this is: you can test now push notifications with test builds.

native/ios/Podfile.lock
```diff
@@ -11,38 +11,37 @@ PODS:
...
+    - GoogleUtilities/UserDefaults (~> 7.0)
+  - FirebaseMessaging (7.4.0):
+    - FirebaseCore (~> 7.0)
+    - FirebaseInstanceID (~> 7.0)
+    - GoogleUtilities/AppDelegateSwizzler (~> 7.0)
+    - GoogleUtilities/Environment (~> 7.0)
+    - GoogleUtilities/Reachability (~> 7.0)
+    - GoogleUtilities/UserDefaults (~> 7.0)
   - Flipper (0.54.0):
     - Flipper-Folly (~> 2.2)
```
maxammann, line 46: @Znippet if you also do some upgrades, do not upgrade this one :D it makes the app crash when developing

maxammann: Oh and also: can you set your first name on github? that way I can tag you more easily using @

klinzo (approved):

SteffiStuffel:
looks sane, but don't know...

SteffiStuffel (approved):

PR #145, opened 2021-02-05
IGAPP-380: Dashboard share path wrong after language change

SteffiStuffel:

I added the routeMappingType to HeaderContainer as a quick-fix. In [IGAPP-450](https://issues.tuerantuer.org/browse/IGAPP-450) we can move this information to route object in the redux store. I haven't done this now because i would have to update all the NewsRouteStateType, CategoryRouteStateType, ... and in 450 they are removed / changed to a more general RouteStateType. So seems to be unnecessary effort from my perspective. I updated the 450-ticket so adjusting the header and headercontainer can't be overlooked.


klinzo:
Can't we just reset the shareUrl param on language change? See https://reactnavigation.org/docs/navigation-actions/#setparams

SteffiStuffel:
> Can't we just set the shareUrl param on language change correctly? See https://reactnavigation.org/docs/navigation-prop/#setparams or https://reactnavigation.org/docs/navigation-actions/#setparams

Yes, was my first idea, too. But we need to set it, after the path is translated (after morphContentLanguage() is executed). But morphContentLanguage does not know the Navigation and acutally there is no navigation event triggered after path translation, because we are already on the correct page.
Another advantage of this approach is, we are removing having the same information on two locations. At the moment information about current route is in shareUrl in the navigation and in the currentPath of routeMapping in redux store. Therefore on every navigation event, pathchange whatever, we have to take care of keeping this information in sync.

klinzo:
Okay right, I see. It's not possible with a route param as currently :+1: 
I think the current proposal adds a lot additional complexity though and leads to inconsistencies as sometimes the share path is handled in the corresponding route mapping and sometimes still in the route params.
I think the problem with wrong share url after language change exists for all routes, i.e. also for offers, search, disclaimer, ...
Do you already have a plan on how we could fix that as well?

Do you think it would make sense/would be possible to add a separate proprety `shareUrl` (or the necessary parts) to our redux state? This property could then be updated in `morphContentLanguage` for all routes and just used in the Header without having to check anything else.

SteffiStuffel:
You are right, this fix only helps for routes, which are already saved in the redux store. Others are still wrong :(

Your solution is more or less an extension of what we planned with the IGAPP-450, where the navigation information is moved (or copied) to the redux store. I'm not really happy with this solution, because the navigation is still in the wrong state and we always have to take care of keeping them in sync. And altough we do not see the wrong navigation state it in the app, i would prefer a proper solution, where the navigation is set correctly too. 

Maybe adding a root navigation could do the job?
https://reactnavigation.org/docs/navigating-without-navigation-prop/

release-notes/unreleased/IGAPP-380-adjust-share-url.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-380
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: The url for sharing content is now correctly set after a language change
+de: Die url, mit der Inhalte geteilt werden können, wir nach dem Ändern der Sprache nun korrekt gesetzt.
```
maxammann, line 7:
```suggestion
de: Die URL zum Teilen von Inhalten wir nach dem Ändern der Sprache nun korrekt gesetzt.
```

release-notes/unreleased/IGAPP-380-adjust-share-url.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-380
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: The url for sharing content is now correctly set after a language change
```
maxammann, line 6:
```suggestion
en: The URL for sharing content is now correctly set after a language change
```

native/src/modules/layout/components/Header.js
```diff
@@ -70,35 +71,47 @@ type PropsType = {|
...
+    props.navigation.navigate(SETTINGS_ROUTE)
+  })
 
-  onShare = async () => {
-    const { scene, t } = this.props
-    const shareUrl = scene.route.params?.shareUrl
-    if (!shareUrl) { // The share option should only be shown if there is a shareUrl
+  const onShare = useCallback(async () => {
+    const { scene, t, path, language, routeMappingType } = props
+    const shareUrlFromScene = scene.route.params?.shareUrl
```
maxammann, line 100: ~~I think we should remove any trace of sharing from the navigation props and only use the state.
I think we can just get the current route_key and then use it to search the state. If we do not find the proper information then do not enable the sharing.~~

native/src/modules/layout/components/Header.js
```diff
@@ -70,35 +71,47 @@ type PropsType = {|
...
     }
 
+    let shareUrl
+    if (path) {
+      shareUrl = url(path)
+    } else if (routeMappingType && routeCityModel) { // for overview pages of news, events and pois, information should be moved to routeMappings in redux store TODO in 450
+      shareUrl = cityContentUrl({ cityCode: routeCityModel.name, languageCode: language, route: routeMappingType, path: null })
+    } else { // for routes not saved in redux store eg. offers
+      shareUrl = shareUrlFromScene
+    }
```
maxammann, line 113: All of this should happen in the container. We also have there access to the navigation prop.

native/src/modules/layout/components/Header.js
maxammann: Oh oke we have some route which are not in the state correct...
I think we als have there a way of changing the language right? Does this actually work like this? Or do we have a problem with not-updating share urls in non-state routes too?

maxammann (dismissed):

maxammann:
@klinzo I think we should not add the shareUrl to the redux state as it will add redundancy. We could add the key of the current route to the state. But this is not required I think as we always have access to the current route key through the scene.

native/src/modules/layout/components/Header.js
SteffiStuffel: Yeah it is not working, too.

klinzo:
Adding the key of the current route does not make sense imo as it is redundant information to the route. Okay, I am fine with not adding shareUrl to the redux state. But I think the share functionality should be consistent among all routes. Can be done in a separate issue though.

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -51,7 +71,12 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     !languages || languages.status !== 'ready') {
     // Route does not exist yet. In this case it is not really defined whether we are peek or not because
     // we do not yet know the city of the route.
-    return { language: state.contentLanguage, routeCityModel, peeking: false, categoriesAvailable: false }
+    return {
```
klinzo, line 74: I thought we fixed these :see_no_evil: Did we set the wrong option? Could you check if the option set is `chop down if to long`? Thanks

native/src/modules/layout/components/Header.js
klinzo: Routes without a mapping in the redux state are working fine already as it is not possible to change languages there.

klinzo (commented):
As discussed: I am fine with this version as it is not a beatufiul workaround but probably the easiest thing to do.

native/src/modules/layout/components/Header.js
SteffiStuffel: moved it.

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: We fixed it, this branch is just older ;)

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -23,21 +24,50 @@ type StatePropsType = {|
...
+    shareUrl = url(path)
+  } else if (routeMappingType && routeCityModel) { // for overview pages of news, events and pois, information should be moved to routeMappings in redux store TODO in 450
+    shareUrl = cityContentUrl({ cityCode: routeCityModel.name, languageCode: route.language, route: routeMappingType, path: null })
+  } else if (shareUrlFromScene && typeof shareUrlFromScene === 'string') { // for routes not saved in redux store eg. offers
+    shareUrl = shareUrlFromScene
+  }
+
+  return shareUrl
+}
+
```
klinzo, line 61: How about:
```suggestion
const getShareUrl = (params: {| city: string, language: string, path: ?string, routeName: string |}): string | null => {
  const { city, language, path, routeName } = params
  if (path) {
    return url(path)
  } else if ([EVENTS_ROUTE, NEWS_ROUTE, POIS_ROUTE].includes(routeName)) {
    const params = { cityCode: city, languageCode: language, path: null }
    return cityContentUrl({ route: routeName, ...params })
  }
  return null
}
```
Together with the suggestion below this should work.

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -67,8 +103,10 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     })
   }
   const peeking = isPeekingRoute(state, { routeCity: route.city })
+  const shareUrlFromScene = ownProps.scene.route.params?.shareUrl
+  const shareUrl: any = getShareUrl(state.cityContent, route, routeKey, routeCityModel, shareUrlFromScene)
```
klinzo, line 107: Together with the suggestion above this should work:
```suggestion
   // $FlowFixMe shareUrl is a string if set
  const shareUrlFromScene: string | null = ownProps.scene.route.params?.shareUrl || null
  const { city, language } = route
  const path = route.path || null
  const shareUrl = shareUrlFromScene || getShareUrl({ city, language, path, routeName: ownProps.scene.route.name })
```
You have to remove the shareUrl from the categories, pois, news and events routes though in `NavigationTypes.js` (which you should do anyway).
Imo this way it is much cleaner. Please double check that `ownProps.scene.route.name` is really what it is used here as.

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -23,21 +24,50 @@ type StatePropsType = {|
...
+
 const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsType => {
   const routeKey = ownProps.scene.route.key
 
   const route = state.cityContent
     ? state.cityContent.categoriesRouteMapping[routeKey] ||
-      state.cityContent.eventsRouteMapping[routeKey] ||
-      state.cityContent.newsRouteMapping[routeKey]
+    state.cityContent.eventsRouteMapping[routeKey] ||
+    state.cityContent.newsRouteMapping[routeKey]
```
klinzo, line 68: Could you add pois here as well please?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -67,8 +104,21 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     })
   }
   const peeking = isPeekingRoute(state, { routeCity: route.city })
+  const shareUrlFromScene = ownProps.scene.route.params?.shareUrl
+  const shareUrl: any = getShareUrl(state.cityContent,
```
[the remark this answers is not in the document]
SteffiStuffel: Thats not possible, because we set shareUrl to null in getShareUrl() line 50. Cause we need this to check if the share-Button should be shown.

We could do something link:
- adding a shareShouldBeDisplayed variable
- only passing shareUrl if shareShouldBeDisplayed is true
- check in Header if shareSouldBeDisplayed is true and shareUrl is available and then show share-Button

But I think it is ok if we just leave string | null.

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: done

native/src/modules/app/constants/NavigationTypes.js
```diff
@@ -50,10 +50,10 @@ export type RoutesParamsType = {|
...
   landing: void,
   dashboard: {| ...ShareUrlType, cityCode: string, languageCode: string |},
-  categories: {| ...ShareUrlType, cityCode: string, languageCode: string |},
-  pois: ShareUrlType,
-  events: ShareUrlType,
-  news: ShareUrlType,
+  categories: {| cityCode: string, languageCode: string |},
+  pois: { },
+  events: { },
+  news: { },
```
klinzo, line 56:
```suggestion
  pois: void,
  events: void,
  news: void,
```
The share url type should also be removed from the dashboard types.

klinzo (approved):

klinzo:
Somehow does not work for disclaimer and offers yet. Specific news are also not shareable right now. You can do this in a separate PR though to get this merged.

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -67,8 +88,12 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     })
   }
   const peeking = isPeekingRoute(state, { routeCity: route.city })
+  const shareUrlFromScene: string | null = typeof ownProps.scene.route.params?.shareUrl === 'string' ? ownProps.scene.route.params?.shareUrl : null
```
Taggotty, line 91: Just curious. This doesn't work?

```suggestion
  const shareUrlFromScene: ?string = typeof ownProps.scene.route.params?.shareUrl === 'string' ? ownProps.scene.route.params?.shareUrl : null
```

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -23,21 +24,35 @@ type StatePropsType = {|
...
-  routeCityModel?: CityModel
+  routeCityModel?: CityModel,
+  shareUrl: string | null
 |}
 
 type DispatchPropsType = {| dispatch: Dispatch<StoreActionType> |}
 
 type PropsType = {| ...OwnPropsType, ...StatePropsType, ...DispatchPropsType |}
 
+const getShareUrl = (params: {| city: string, language: string, path: ?string, routeName: string |}): string | null => {
```
Taggotty, line 35: Please add tests for this functionality if they are not yet present. Also again can't you say `?string` here?

native/src/modules/layout/components/Header.js
```diff
@@ -70,31 +67,32 @@ type PropsType = {|
...
 |}
 
-class Header extends React.PureComponent<PropsType> {
-  canGoBackInStack (): boolean {
-    return !!this.props.previous
-  }
+const Header = (props: PropsType) => {
+  const canGoBackInStack = useCallback((): boolean => {
+    return !!props.previous
+  })
```
maxammann, line 77: I think the `props` or `previous` dependency missing here

native/src/modules/layout/components/Header.js
```diff
@@ -70,31 +67,32 @@ type PropsType = {|
...
 
-  goBackInStack = () => {
-    this.props.navigation.goBack()
-  }
+  const goBackInStack = useCallback(() => {
+    props.navigation.goBack()
```
maxammann, line 80: I think the `props` or `navigation` dependency missing here.

native/src/modules/layout/components/Header.js
```diff
@@ -70,31 +67,32 @@ type PropsType = {|
...
+  })
 
-  goToLanding = () => {
-    const { navigation, dispatch } = this.props
+  const goToLanding = useCallback(() => {
```
maxammann, line 83: I think the `navigation` or `dispatch` dependency missing here.

native/src/modules/layout/components/Header.js
```diff
@@ -70,31 +67,32 @@ type PropsType = {|
...
+    const { navigation, dispatch } = props
     // $FlowFixMe Navigation type of the header does not match that of screens.
     navigateToLanding({ dispatch, navigation })
-  }
+  })
 
-  goToSettings = () => {
-    this.props.navigation.navigate(SETTINGS_ROUTE)
-  }
+  const goToSettings = useCallback(() => {
```
maxammann, line 89: I think the `navigation` dependency missing here.

native/src/modules/layout/components/Header.js
```diff
@@ -70,31 +67,32 @@ type PropsType = {|
...
+    props.navigation.navigate(SETTINGS_ROUTE)
+  })
+
+  const onShare = useCallback(async () => {
```
maxammann, line 93: I think the `t` `shareUrl` dependency missing here.

maxammann: A lot more callback deps are missing here.

native/src/modules/layout/containers/HeaderContainer.js
maxammann: `?string` is different to `string | null` atleast as it includes undefined.

maxammann: What type can shareUrl have? Can't it be only null/undefined or string?

maxammann (approved):
Looks good! A lto of stuff is simpler now!

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: changed it to ?string and will add a test.

klinzo (changes requested):
As previously mentioned:
> Somehow does not work for disclaimer and offers yet. Specific news are also not shareable right now. You can do this in a separate PR though to get this merged.

Should either be fixed or moved to a new issue.

native/src/modules/layout/components/Header.js
SteffiStuffel: Good catch. Maybe we can add this to linting?

klinzo: There is already a linting rule but just a warning yet. I think there are cases where not including the dependencies is intentional so I am not sure if we want to change it.

SteffiStuffel:
> As previously mentioned:
> 
> > Somehow does not work for disclaimer and offers yet. Specific news are also not shareable right now. You can do this in a separate PR though to get this merged.
> 
> Should either be fixed or moved to a new issue.

Opened a new issue for it.

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const ownProps = {
+      scene: {
+        route: {
+          key: 'routeKey1'
+        }
+      }
+    }
+
+    const result = TestRenderer.create(
```
klinzo, line 92: Next time pls use testing-library instead of the test renderer.

klinzo (approved):

PR #146, opened 2021-02-05
IGAPP-345: Display offers during loading

klinzo:

This PR contains multiple changes:
* Fix displaying of errors for disclaimer, offers and sprungbrett
* Display error and disclaimer/offers/sprungbrettJobs during refreshing
* Refactoring and simplifying of routing to internal offers

### DIFF
--- a/native/src/modules/endpoint/loadFromEndpoint.js
+++ b/native/src/modules/endpoint/loadFromEndpoint.js
@@ -1,0 +1,33 @@
+// @flow
+
+import determineApiUrl from './determineApiUrl'
+import { Payload } from 'api-client'
+
+const loadFromEndpoint = async <T>(
+  request: (apiUrl: string) => Promise<Payload<T>>,
+  setData: ?T => void,
+  setError: ?Error => void,
+  setLoading: boolean => void
+) => {
+  setLoading(true)
+
+  try {
+    const apiUrl = await determineApiUrl()
+    const payload: Payload<T> = await request(apiUrl)
+
+    if (payload.error) {
+      setError(payload.error)
+      setData(null)
+    } else {
+      setData(payload.data)
+      setError(null)
+    }
+  } catch (e) {
+    setError(e)
+    setData(null)
+  } finally {
+    setLoading(false)
+  }
+}
+
+export default loadFromEndpoint
--- a/native/src/modules/theme/hocs/withTheme.js
+++ b/native/src/modules/theme/hocs/withTheme.js
@@ -5,7 +5,7 @@ import type { ThemeType } from '../constants'
 import { lightTheme } from '../constants'
 import wrapDisplayName from '../../common/hocs/wrapDisplayName'
 
-function withTheme<Props: { theme: ThemeType }> (
+function withTheme<Props: { theme: ThemeType, ... }> (
   Component: React.AbstractComponent<Props>
 ): React.AbstractComponent<$Diff<Props, {| theme: ThemeType |}>> {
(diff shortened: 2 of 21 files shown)


native/src/routes/wohnen/containers/WohnenOfferContainer.js
```diff
@@ -6,12 +6,11 @@ import type { StateType } from '../../../modules/app/StateType'
 import { connect } from 'react-redux'
```
klinzo, line 6: Arghhh. Can't wait for the shutdown of this wohnen stuff...

native/src/modules/endpoint/loadFromEndpoint.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import determineApiUrl from './determineApiUrl'
+import { Payload } from 'api-client'
+
+const loadFromEndpoint = async <T>(
```
maxammann, line 6: You could refactor this to a hook:

```suggestion
const useEndpointFetch = () => {

   const [data, setData] = useState()
   const [error, setError] = useState()
   
   useEffect(() => {
      //fetch and set state
   })

  return error, data
}
```

maxammann (approved):

native/src/modules/endpoint/loadFromEndpoint.js
klinzo: Thought about that as well, but not sure how I would be able to call it on refresh then? Any ideas?

maxammann: You could create a second hook. Something like:

```suggestion
const useRefresh = (): [boolean, () => void] => {
  const dispatch = useDispatch()
  const range = useSelector<RootState, Range | null>(
    (state) => state.days.range,
  )
  const loading = useSelector<RootState, boolean>((state) => state.days.loading)

  return [
    loading,
    useCallback(() => {
      if (range) {
        dispatch(
          fetchRange({
            from: range.from,
            to: range.to,
          }),
        )
      }
    }, [dispatch, range]),
  ]
}

export default useRefresh

```

klinzo: Okay, sorry, I meant on pull to refresh. E.g. here: https://github.com/Integreat/integreat-app/pull/146/files#diff-d46e02df50fb1c491b623b34aeacad7ca09df8050dc6e8d0634332db3fbf018eR78

klinzo: Or can I call the hook normally even though it contains `useEffect` and `useState`? That sounds weird because we need to share state between both calls...

maxammann: I think in that case it is fine as you did it.

native/src/modules/theme/hocs/withTheme.js
```diff
@@ -5,7 +5,7 @@ import type { ThemeType } from '../constants'
 import { lightTheme } from '../constants'
 import wrapDisplayName from '../../common/hocs/wrapDisplayName'
 
-function withTheme<Props: { theme: ThemeType }> (
+function withTheme<Props: { theme: ThemeType, ... }> (
```
Taggotty, line 8: why is that change?

Taggotty (approved):

native/src/modules/theme/hocs/withTheme.js
klinzo: This indicates to flow that the object contains more properties without having to specify all props (which is not possible here anyway). It makes warnings go away :grin:

PR #147, opened 2021-02-06
IGAPP-305: Prepare and translate override translations

maxammann:

This is only configuration and preparing. The translations will follow in a separate PR.


maxammann:
The translation files are here: https://drive.google.com/drive/folders/1gimQDlovKYpFASRX5_GMfbF5Nqc-MCBM?usp=sharing

klinzo (approved):

translations/tools/manage.js
```diff
@@ -202,17 +212,28 @@ program
...
-      }
-    } else if (targetFormat === 'json') {
-      writeJsonFromCsv(fromPath, toPath, sourceLanguage)
+      }],
+      ['csv-json', () => {
+        writeJsonFromCsv(fromPath, toPath, sourceLanguage)
+      }]
+    ])
+
+    const convert = converter.get(`${sourceFormat.toLowerCase()}-${targetFormat.toLowerCase()}`)
```
Znippet, line 230: Sweet, didn't know they have Map in ES6 until now

Znippet (approved):

translations/tools/manage.js
maxammann: Also use them with caution. This is a bad use it it actually :D first wanted to make the key an array. But it does not work.

Will revert this to a plain object.

SteffiStuffel:
I tried it, but i guess i'm missing something.

I exported translations with 
yarn export:ods:translations

edited the en.ods in the third line (changed Events to EventsEn)

imported it with  
yarn import:ods:translations

exits properly without errors. translation.json is updated, but neither native (although reinstalled) nor webapp is showing 'EventsEN' in the heading?

PR #149, opened 2021-02-08
IGAPP-457: Try to prevent oom errors during android builds

klinzo:

I obviously do not know whether this is enough to prevent these errors due to oom.
Possible solution is from here, maxWorkers is already set: https://stackoverflow.com/questions/56002938/reactnative-0-59-x-build-fails-on-circleci-with-exit-value-137
Possibly more infos: 
https://circleci.com/blog/how-to-handle-java-oom-errors/
and another approach we could try if this does not work:
https://medium.com/@stevewestwood/building-android-projects-on-circleci-using-their-machine-executor-b731b386a12d


Znippet (approved):

.circleci/src/@common.yml
```diff
@@ -1,5 +1,10 @@
 version: 2.1
 
+# Try to prevent oom errors during android builds
+# https://stackoverflow.com/questions/56002938/reactnative-0-59-x-build-fails-on-circleci-with-exit-value-137
+JAVA_OPTS: '-Xms512m -Xmx2g'
+GRADLE_OPTS: '-Xmx3g -Dorg.gradle.daemon=false -Dorg.gradle.jvmargs="-Xmx2g -XX:+HeapDumpOnOutOfMemoryError"'
```
maxammann, line 6: These variables should be put to the circleci jobs where they are used. Also they seem a bit weird as `org.gradle.jvmargs` and `JAVA_OPTS` is set. What should gradle choose?

We already configure this in Fastfiles: https://github.com/Integreat/integreat-app/blob/51a2f5f9db596569f639144ee9b0a932851d7eaa/native/android/fastlane/Fastfile#L55-L58

Also make sure to leave some memory for the node process (bundler). I would estimate it requires at least 500MB of ram.

maxammann (changes requested):

SteffiStuffel:
Why haven't you set the maxWorkers in metro.config.js as suggested in the stackoverflow link you shared?

SteffiStuffel:
Where is defined ci should use the metro.config.ci.js ?

maxammann:
In the Fastfile I posted: EXTRA_PACKAGER_ARGS

maxammann (approved):

PR #151, opened 2021-02-08
IGAPP-173: Upgrade web-dependencies flow-bin and flow-types

Znippet:
I was not able to upgrade flow-bin to the current version 1.44 because of the breaking change "flow Types-First"
https://flow.org/en/docs/lang/types-first/

Beside of that we have to upgrade https://github.com/integreat/react-sticky-headroom to fix some flow errors

I also updated the flow-types.
"yarn flow:install-types" added the following additional types. 
But I couldn't solve the flow errors, so i didn't commit them:

- styled-components_v5.x.x
- @babel/register_v7.x.x
- @testing-library/react_v11.x.x


web/.flowconfig
```diff
@@ -27,6 +29,10 @@
 module.system.node.resolve_dirname=node_modules
 esproposal.optional_chaining=enable
 
+; TODO Remove and fix flow errors to upgrade to v0.143.0
+; https://github.com/facebook/flow/releases/tag/v0.143.0
+types_first=false
```
Znippet, line 34: We have to solve this when we want to upgrade to current version

web/.flowconfig
```diff
@@ -4,6 +4,8 @@
 <PROJECT_ROOT>/lib-dist/.*
 .*/node_modules/.*/__tests__/.*
 .*/build-configs/.*/assets/manifest\.json
+; TODO Fix flow error
+.*/node_modules/@integreat-app/react-sticky-headroom/index\.js\.flow
```
Znippet, line 8: Seems to be our library? :) So we should fix it there

web/.flowconfig
maxammann: I would say we switch to typscript before doing this flow upgrade :D

klinzo: Don't we also have to use a specific flow version with each react version?

maxammann: yes, but this change is for web, right?

web/src/routes/wohnen/components/OfferDetail.js
```diff
@@ -3,13 +3,12 @@
...
 import styled from 'styled-components'
-import type { TFunction } from 'react-i18next'
-import { withTranslation } from 'react-i18next'
+import { withTranslation, TFunction } from 'react-i18next'
 import Caption from '../../../modules/common/components/Caption'
 
 type PropsType = {|
   offer: WohnenOfferModel,
-  t: TFunction
+  t: typeof TFunction
```
maxammann, line 11: Isn't TFunction already a type? Seems weird that you have to do typeof.

I think we declared this type manually in the flow-typed directory.

maxammann (commented):
I hope this PR was not too much work. I think we should stick to our flow versions and only upgrade if really really neccassary. In the long run flow is ready dead imho and sooner or later we should migrate

web/.flowconfig
klinzo: Ahh I see, so it is only important for react-native to stick to the same version?

web/src/routes/wohnen/components/OfferDetail.js
klinzo: I think there are no types yet for react-i18next. https://issues.tuerantuer.org/browse/IGAPP-203

web/.flowconfig
klinzo: It is. You could create an issue for this.

web/src/__mocks__/react-i18next.js
```diff
@@ -1,18 +1,16 @@
-// @flow
-
 import * as React from 'react'
```
klinzo, line 1: Why did you remove the types here?

web/src/__mocks__/react-onclickoutside.js
```diff
@@ -1,15 +1,13 @@
-// @flow
-
 import * as React from 'react'
```
klinzo, line 1: Why did you remove flow here?

web/src/modules/app/containers/Switcher.js
```diff
@@ -48,15 +47,15 @@ type PropsType = {|
   tunewsLanguagesPayload: Payload<Array<LanguageModel>>,
   tunewsElementPayload: Payload<TunewsModel>,
   offersPayload: Payload<Array<OfferModel>>,
-  sprungbrettJobsPayload: Payload<Array<SprungbrettOfferPage>>,
+  sprungbrettJobsPayload: Payload<Array<typeof SprungbrettOfferPage>>,
```
klinzo, line 50: I don't think this type is correct, is it? I think it should be something like `Payload<Array<SprungbrettJobModel>>`

klinzo (approved):

klinzo:
> I hope this PR was not too much work. I think we should stick to our flow versions and only upgrade if really really neccassary. In the long run flow is ready dead imho and sooner or later we should migrate

Why not just merge this and upgrade if the work is already done and this does not lead to more work right now? We don't have to upgrade further to the current version though.

maxammann:
I never meant not to merge this even if it probably sounded like that :D I just would avoid setting that flat to true.

Merge in main :)

web/src/routes/wohnen/components/OfferDetail.js
Znippet: Added info to remove " t: typeof TFunction" after adding types to Task IGAPP-203

web/.flowconfig
Znippet: Added Issue https://issues.tuerantuer.org/browse/HEADROOM-9

web/src/modules/app/containers/Switcher.js
Znippet: fixed

web/src/modules/app/containers/Switcher.js
```diff
@@ -2,7 +2,7 @@
 
 import * as React from 'react'
 import { connect } from 'react-redux'
-import SprungbrettOfferPage from '../../../routes/sprungbrett/containers/SprungbrettOfferPage'
+import SprungbrettJobModel from 'api-client/src/models/SprungbrettJobModel.js'
```
klinzo, line 5:
```suggestion
import SprungbrettJobModel from 'api-client/src/models/SprungbrettJobModel'
```

klinzo (approved):

klinzo:
@Znippet why do I need to update the podfile? All checks are already succeeding which totally makes sense as you only changed web dependencies.

Znippet:
> @Znippet why do I need to update the podfile? All checks are already succeeding which totally makes sense as you only changed web dependencies.

I only saw that the IOS tests failed and thaught this was the reason. If they are succeeding now all is fine, we can merge

maxammann (approved):

PR #155, opened 2021-02-14
IGAPP-448: Support html entities for local news

klinzo:

Git fails to track changes if files are renamed.
Renamed the following files:
`NewsList` -> `News` (as this is not just rendering the list but also the details, consistent with other routes now)
`List` -> `NewsList` (as this is only a list used for news)
`NewsItemDetails` -> `NewsDetail`


native/package.json
```diff
@@ -53,8 +53,9 @@
     "@sentry/react-native": "^1.7.2",
     "api-client": "0.0.1",
     "build-configs": "0.0.1",
+    "events": "^3.2.0",
```
klinzo, line 56: Somehow needed for htmlparser2 with version >= 5.0. Otherwise the app crashes with `module events not found`

native/src/routes/news/components/TuNewsFooter.js
```diff
@@ -0,0 +1,65 @@
+// @flow
```
klinzo, line 1: Extracted from `NewsDetail` (former `NewsItemDetail) with a little refactoring.

Taggotty (approved):

native/package.json
```diff
@@ -75,7 +76,7 @@
     "react-native-permissions": "2.0.9",
     "react-native-progress": "^4.0.3",
     "react-native-reanimated": "^1.13.2",
-    "react-native-render-html": "^5.0.0-alpha.2",
+    "react-native-render-html": "^5.1.0",
```
maxammann, line 79: @Taggotty does this version already fix the issues with RTL/LTR?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -155,11 +134,10 @@ const CategoryListContent = ({
 
   return <SpaceBetween>
     <Container>
-      <Html html={content}
+      <Html source={{ content }}
             onLinkPress={onLinkPress}
-            contentWidth={width}
-            allowFontScaling
-            textSelectable
+            contentWidth={useWindowDimensions().width}
```
maxammann, line 139: Put the `useWindowDimensions` to the top of the component. This is better style and also the app would crash if in one render this line is called and in an other not.

native/src/routes/news/components/NewsDetail.js
```diff
@@ -0,0 +1,88 @@
...
+        }}>
+        {selectedNewsItem instanceof TunewsModel && (
+          <HeaderImageWrapper>
+            <HeaderImage source={headerImage} />
+          </HeaderImageWrapper>
+        )}
+        <Container>
+          <NewsHeadLine theme={theme}>{selectedNewsItem.title}</NewsHeadLine>
+          <Html source={{ html: content }}
+                contentWidth={useWindowDimensions().width}
```
maxammann, line 68: Move to start of render function

maxammann (approved):
I'm not sure whether this solves our problem. So Push Notifications are now allowed to include html entities right?
But there are multiple views of the push notifications. One in the app (which was fixed in this PR) and the message in the noticiation area. The actual message which is diplayed in the notification area must not contain html entities.

native/package.json
klinzo: No, the issue is still open: https://github.com/meliorence/react-native-render-html/issues/286

klinzo:
> I'm not sure whether this solves our problem. So Push Notifications are now allowed to include html entities right?
> But there are multiple views of the push notifications. One in the app (which was fixed in this PR) and the message in the noticiation area. The actual message which is diplayed in the notification area must not contain html entities.

Oh okay, didn't know that. Then this is a cms issue because we are not able to fix it, right?

PR #156, opened 2021-02-14
IGAPP-477: Support links in local news detail view

klinzo:


native/src/routes/news/components/NewsDetail.js
```diff
@@ -43,11 +44,23 @@ const NewsHeadLine: StyledComponent<{||}, ThemeType, *> = styled.Text`
...
 
-const NewsDetail = ({ theme, selectedNewsItem, language }: PropsType) => {
-  const content = selectedNewsItem.content || selectedNewsItem.message || ''
+const NewsDetail = ({ theme, newsItem, language, navigateToLink }: PropsType) => {
+  const localNewsContent = newsItem instanceof LocalNewsModel ? newsItem.message : ''
+  const tuNewsContent = newsItem instanceof TunewsModel ? newsItem.content : ''
+  const content = localNewsContent || tuNewsContent
+
+  // https://stackoverflow.com/a/3809435
+  const regex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/g
```
klinzo, line 57: Not sure if it is good style to have regexes like this. Any other ideas?

SteffiStuffel: You can use URL() constructer:
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

If the given base URL or the resulting URL are not valid URLs, the JavaScript TypeError exception is thrown.

But not sure if it is supported by enough browsers.

klinzo: Hmm but this is native and not web, we'd have to use `url-parse` there. And I don't think this could work as we'd have to try to create a URL from all possible partitions of the string (or at least from all sequences divided by whitespaces...

Taggotty: how about sth like this: https://www.npmjs.com/package/linkify-it ? seems to be used quite a lot on npm and there is also an implementation, that provides a Wrapper for react: https://www.npmjs.com/package/react-linkify

Taggotty (approved):

native/src/routes/news/components/NewsDetail.js
klinzo: I like linkify :thinking: I think I'll play around with it and look if it makes sense to use it in this context. I think using it would make it a little bit more complicated because we first need to check with linkify where links are and afterwards have to replace it, but I guess this is better than having an ugly regex we don't understand :smile: 
Regarding the react wrapper: I don't think this works for us in the native context. We'd need a custom onClick handler which is not supported afaics. Nice suggestion though :+1:

maxammann: I think it is fine to use a regex here. I mean matching an URL is not that critical. Sometimes other clients get it wrong and mark stuff which is not a URL as URL and have some misses.

Actually we do not want to match URL but something which is similar to an URL like: `This is an url: helloworld.dev` .
But what should happen if a user does mistake like this: `This is not an URL.sexy developers ftw!`
Is URL.sexy a url? Yes! http://www.url.sexy/
Should it be matched? no!

So a best effort regex is fine here.

maxammann: image


image

maxammann, line 57: Move this to a constant variable for performance reasons. That way the RegExp object gets created only once.

maxammann (approved):

native/src/routes/news/components/NewsDetail.js
klinzo: Sounds good to me :+1:

klinzo: :+1: Moved the whole thing to the api-client. Better for testing, modularity and reusability anyway.

PR #160, opened 2021-02-16
IGAPP-340: Create manifest.json from build-config

SteffiStuffel:


web/tools/webpack.config.babel.js
```diff
@@ -35,6 +35,43 @@ const readVersionName = () => {
...
+  manifest.related_applications = [{
+    platform: 'play',
+    id: androidBuildConfig.applicationId,
+    url: `https://play.google.com/store/apps/details?id=${androidBuildConfig.applicationId}`
+  }]
+
+  switch (buildConfigName) {
+    case 'integreat':
+    case 'integreat-test-cms':
+      manifest.name = 'Integreat'
```
klinzo, line 56: I think we don't have to hardcode this, should be included in the build config already as name. Regarding the description: Getting it from the locales is probably to complex, right?

web/tools/webpack.config.babel.js
```diff
@@ -35,6 +35,43 @@ const readVersionName = () => {
   return versionFile.versionName
 }
 
+const generateManifest = (content: string, buildConfigName: string): string => {
```
klinzo, line 38: Maybe extract this to a separate script file, what do you think?

web/tools/manifest.json
```diff
@@ -1,9 +1,9 @@
 {
```
klinzo, line 1: Did you check whether there are other properties a manifest.json should include?

web/package.json
```diff
@@ -103,7 +103,7 @@
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-styled-components": "^1.10.0",
     "clean-webpack-plugin": "^3.0.0",
-    "copy-webpack-plugin": "^5.1.1",
+    "copy-webpack-plugin": "6.2.1",
```
klinzo, line 106: Does this already fix this issue: https://issues.tuerantuer.org/browse/IGAPP-481
@Znippet

web/tools/webpack.config.babel.js
SteffiStuffel: Manifest is generated, when building the app. So we don't know the locales of the user at this point.

web/tools/webpack.config.babel.js
SteffiStuffel: Also thought about it, but now i removed the switch-case and therefore its a lot shorter. Should i move it nevertheless?

web/tools/manifest.json
SteffiStuffel: I think this is enough and checked the docs, but maybe I miss something.

web/package.json
SteffiStuffel: Yeah api changed and leads to this error. I fixed it, so issue should be solved. 
I updated to 6.2.1. cause 7.0.0 is not working properly yet.

web/tools/manifest.json
```diff
@@ -20,17 +20,17 @@
...
-  "homepage_url": "https://integreat-app.de/",
+  "homepage_url": "",
   "related_applications": [
     {
-      "platform": "play",
-      "id": "tuerantuer.app.integreat",
-      "url": "https://play.google.com/store/apps/details?id=tuerantuer.app.integreat"
+      "platform": "",
+      "id": "",
+      "url": ""
```
maxammann, line 28: Maybe fill these empty string with something like: `"This is replaced by webpack.config.babel.js"`

maxammann (approved):

web/tools/webpack.config.babel.js
klinzo: No I guess it is fine then.

web/package.json
klinzo: Okay I guess then we can close the issue and just upgrade further once the version is stable.

web/tools/webpack.config.babel.js
klinzo: We could just german as default though as we do now as well I think? But probably the complexity with overrides is too much.

build-configs/BuildConfigType.js
```diff
@@ -30,6 +30,7 @@ export type FeatureFlagsType = {|
 // Available on all platforms
 export type CommonBuildConfigType = {|
   appName: string,
+  appDescription: string,
```
klinzo, line 33: Maybe we should move this to web. I think we don't want to use it for native. I think we should also add a comment why and where this is used.

web/tools/manifest.json
SteffiStuffel: I thought if anythings not working with writing a value it would be better to just have an empty string there?

build-configs/BuildConfigType.js
```diff
@@ -115,5 +116,6 @@ export type iOSBuildConfigType = {|
   bundleIdentifier: string, // iOS application identifier.
   provisioningProfileSpecifier: string, // Provisioning profile to sign the app.
   appleId: string, // Id of the app in the Apple App Store
+  itunesAppName: string,
```
klinzo, line 119: A comment with more information on what this is would be nice I think.

build-configs/aschaffenburg/index.js
```diff
@@ -14,6 +14,7 @@ const APPLE_APP_ID = '1551810291'
 
 const commonAschaffenburgBuildConfig: CommonBuildConfigType = {
   appName: 'Aschaffenburg App', // TODO maybe we have to change this to 'AB App'
+  appDescription: 'App of the city Aschaffenburg to inform about social topics.',
```
klinzo, line 17: Where did you get this from? Probably this has to be changed later, right? If so, please add a TODO (maybe in https://issues.tuerantuer.org/browse/IGAPP-453?)

web/tools/webpack.config.babel.js
SteffiStuffel: We can keep it in mind, and change it if somebody ever encounters this string and wonders or wants them to be in the correct language. I do not even know where (if ever) they are displayed?

build-configs/integreat/index.js
```diff
@@ -14,6 +14,7 @@ const APPLE_APP_ID = '1072353915'
 
 const commonIntegreatBuildConfig: CommonBuildConfigType = {
   appName: 'Integreat',
+  appDescription: 'Daily Guide for Refugees. Digital. Multilingual. Free.',
```
klinzo, line 17: Maybe we should use german here?

klinzo (approved):

maxammann:
@SteffiStuffel just a note about force pushing: If you force push after a PR is open Github can not show the diff between my last review. So better not rebase after opening a PR imo :)

web/tools/manifest.json
SteffiStuffel: Changed it, as you suggested

build-configs/BuildConfigType.js
SteffiStuffel: done

build-configs/BuildConfigType.js
SteffiStuffel: done

build-configs/aschaffenburg/index.js
SteffiStuffel: added it to the description of #453

build-configs/integreat/index.js
SteffiStuffel: What would be the correct german slogan? 
Should i use this: Integreat – die lokale und mehrsprachige Integrations-Plattform für Zugewanderte (its from the integreat homepage)

SteffiStuffel:
> @SteffiStuffel just a note about force pushing: If you force push after a PR is open Github can not show the diff between my last review. So better not rebase after opening a PR imo :)

Sorry about this. Didn't know.

PR #163, opened 2021-02-17
IGAPP-203: Add and fix flow types for react-i18next (native)

klinzo:

Will do web in a separate PR. I think it is quite nice that we have correct types everywhere now :tada:


native/src/routes/wohnen/containers/WohnenOfferContainer.js
```diff
@@ -148,12 +153,13 @@ class WohnenOfferContainer extends React.Component<WohnenPropsType, WohnenStateT
...
       <WohnenOffer title={title} offerHash={offerHash} navigateToOffer={navigateToOffer} offers={offers}
                           t={t} theme={theme} language={language} />
-      <SiteHelpfulBox navigateToFeedback={navigateToFeedback} theme={theme} t={t} />
+      <SiteHelpfulBox navigateToFeedback={navigateToFeedback} theme={theme} />
     </ScrollView>
   }
 }
 
 export default connect<PropsType, OwnPropsType, _, _, _, _>(mapStateToProps)(
+  // $FlowFixMe
```
maxammann, line 162: Add reason

maxammann (approved):
Nice PR! All these type fixes are just because of withTheme, right?

We should think about upstreaming that one such that the community can benefit

klinzo:
> Nice PR! All these type fixes are just because of withTheme, right?
> 
> We should think about upstreaming that one such that the community can benefit

Mainly because of withTranslation from `react-i18next`. If we want to upstream it we'd have to put more effort in the type definitions I guess. I'll look into it.

native/src/routes/wohnen/containers/WohnenOfferContainer.js
klinzo: It is the wohnen offer, thats reason enough :stuck_out_tongue:

native/src/routes/disclaimer/DisclaimerContainer.js
```diff
@@ -70,13 +70,15 @@ const DisclaimerContainer = ({ theme, resourceCacheUrl, navigation, route, dispa
     await loadFromEndpoint<PageModel>(request, setDisclaimer, setError, setLoading)
   }, [cityCode, languageCode, setError, setDisclaimer, setLoading])
 
+  const tryAgain = useCallback(() => { loadDisclaimer().catch(e => setError(e)) }, [loadDisclaimer])
+
   useEffect(() => {
     loadDisclaimer().catch(e => setError(e))
```
Taggotty, line 76: Do you still need this then?

native/src/routes/search/containers/SearchModalContainer.js
```diff
@@ -59,22 +59,24 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType) => {
...
-  return <ThemedTranslatedSearch
-    {...rest}
-    navigateToLink={navigateToLinkProp}
-    navigateTo={createNavigate(dispatch, navigation)} />
+  return cityCode
+    ? <ThemedTranslatedSearch cityCode={cityCode}
+                              navigateToLink={navigateToLinkProp}
+                              navigateTo={createNavigate(dispatch, navigation)}
+                              {...rest} />
+    : null
```
Taggotty, line 79: Does this case actually occur?

Taggotty (approved):

native/src/routes/disclaimer/DisclaimerContainer.js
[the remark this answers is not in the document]
klinzo: Yes in order to fetch on mount. I think this could be refactored at some point so that there are less duplicates though.

native/src/routes/search/containers/SearchModalContainer.js
klinzo: I don't think so, no. Necessary for flow though.

native/src/routes/intro/IntroContainer.js
```diff
@@ -238,10 +242,10 @@ class Intro extends React.Component<PropsType, StateType> {
...
 const mapStateToProps = (state: ReduxStateType): {| language: string |} => ({ language: state.contentLanguage })
-type ConnectType = {| language: string, dispatch: () => void |}
+type ConnectType = {| ...OwnPropsType, language: string, dispatch: () => void |}
 
-export default connect<ConnectType, {||}, _, _, _, _>(mapStateToProps)(
-  withTranslation(['intro', 'settings'])(
-    withTheme(Intro)
+export default connect<ConnectType, OwnPropsType, _, _, _, _>(mapStateToProps)(
+  withTranslation<$Diff<PropsType, {| theme: ThemeType |}>>(['intro', 'settings'])(
+    withTheme<PropsType>(Intro)
```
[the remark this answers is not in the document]
klinzo: Actually it's not that complicated. You just have to specify what props the component will receive.Therefore in this case the component wrapped in `withTranslation` will receive all props with exception of the `theme` because that is only injected by `withTheme` later. On the other hand `Intro` wrapped in `withTheme` will receive all props.

I think if we gradually remove unnecessary `withThemes` and switch to hooks, these will vanish sooner or later.

PR #165, opened 2021-02-19
IGAPP-203: Type styled components and withTheme

klinzo:

While typing react-i18next I noticed that our styled components haven't been typed as well.
Apart from typing all styled components I removed unnecessary `withTheme` hocs and theme dependencies. I think we should do the same in native (gradually). Styled components access the theme provider by themselves. Explicitly passing the `theme` prop is only intended if one wants to override the default theme (https://styled-components.com/docs/advanced#the-theme-prop).


web/src/modules/feedback/components/__tests__/FeedbackBoxContainer.spec.js
```diff
@@ -68,8 +65,7 @@ describe('FeedbackBoxContainer', () => {
           onSubmit={() => {}}
           closeFeedbackModal={() => {}}
           offers={null}
-          theme={theme}
-          sendingStatus='SUCCESS'
+            sendingStatus='SUCCESS'
```
maxammann, line 68: Wrong formatting?

web/src/modules/layout/containers/LocationHeader.js
```diff
@@ -63,7 +61,7 @@ export class LocationHeader extends React.Component<PropsType> {
     ]
   }
 
-  getNavigationItems (): Array<Element<typeof HeaderNavigationItem>> {
+  getNavigationItems (): Array<React$Node> {
```
maxammann, line 64: Isn't the correct type React.Node here? I think this is the first time we use `React$Node`. Maybe I'm wrong :D

web/src/routes/categories/components/CategoryEntry.js
```diff
@@ -99,8 +99,11 @@ class CategoryEntry extends React.PureComponent<PropsType> {
     )
   }
 
-  getMatchedContent (numWordsSurrounding: number): typeof ContentMatchItem {
+  getMatchedContent (numWordsSurrounding: number): React$Node {
```
maxammann, line 102: Maybe React.Node is better, not sure.

maxammann (approved):
I think it is fine not to pass `theme` through the component tree if the child components are able to render without a ThemeProvider.

Do you know whether that is the case here? Like what happens if you render a styled component without using a ThemeProvider around it? Will it crash? Will it work?

klinzo:
> I think it is fine not to pass `theme` through the component tree if the child components are able to render without a ThemeProvider.
> 
> Do you know whether that is the case here? Like what happens if you render a styled component without using a ThemeProvider around it? Will it crash? Will it work?

If a styled component accesses the theme but there is no ThemeProvider around it, it'll crash. I think anyway that this is fine as this is just how styled components are used and I don't see a point in adding additional complexity here. We didn't pass the theme consistent anyway.

web/src/modules/layout/containers/LocationHeader.js
klinzo: Both is correct. `Reac$Node` can be used without importing anything. It is an official type though.
`/**
 * Copyright (c) 2013-present, Facebook, Inc.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

/**
 * A UI node that can be rendered by React. React can render most primitives in
 * addition to elements and arrays of nodes.
 */
`

web/src/modules/layout/components/GeneralHeader.js
```diff
@@ -13,22 +13,22 @@ type PropsType = {|
...
                                iconSrc={landingIcon}
                                text={t('changeLocation')} />]
-      : []
+    : []
 
-    return <Header viewportSmall={viewportSmall}
-                   logoHref={getPath({})}
-                   actionItems={actionItems} />
-  }
+  const onStickyTopChanged = useCallback(() => {}, [])
```
Taggotty, line 26: What's that for?

Taggotty (approved):

web/src/modules/layout/components/GeneralHeader.js
klinzo: Required prop in the Header component which was not recognized previously because of untyped hocs. In order not to change the behaviour I use just a function doing nothing.

web/src/modules/layout/containers/LocationHeader.js
klinzo: Changed anyway for consistency

PR #168, opened 2021-02-23
IGAPP-332: Add pull to refresh for news

SteffiStuffel:


klinzo (changes requested):
I think there was a reason why I implemented it this way. If I remember correctly, this leads to two scroll views nested which is not supported and leads to undefined behavior. This is because the news component renders a scroll view as well to support infinite scrolling behavior. Please check if everything works as desired.

klinzo:
image

klinzo:
I think you either have to refactor and add the pull to refresh functionality to the news list scroll view, or you have to remove the news list scroll view and add the infinite scrolling behavior to the withPayloadProvider scroll view. I'd favor the latter, but this may prove to be quite complicated.

maxammann:
I'd be fine with both as long as it does not get hacky :)

maxammann (approved):

klinzo (approved):
Did you try the other way?

klinzo:
Tested on real android 10 device. I figure we'll have to refactor it at some point nevertheless and make it consistent with other routes. But fine for now.

SteffiStuffel:
> Did you try the other way?

I first tried as you suggested by putting the infinite scroll behaviour to the payloadProvider, but this would be a lot more stuff in payload provider and would not be reused at any other point. Think this makes sense, when we have different containers using this feature otherwise we would have code in a HOC which is only used by one component and as far as i understood HOCs (still struggling a bit with them), this wouldn't be a valid use case. But on the other hand, having the refresh functionality in the NewsList isn't a nice solution either. 

If you want i can create an issue 'Move infinite scroll behaviour to PayloadProvider' or something like this. Are there other components with infitinte scroll?

klinzo:
> > Did you try the other way?
> 
> I first tried as you suggested by putting the infinite scroll behaviour to the payloadProvider, but this would be a lot more stuff in payload provider and would not be reused at any other point. Think this makes sense, when we have different containers using this feature otherwise we would have code in a HOC which is only used by one component and as far as i understood HOCs (still struggling a bit with them), this wouldn't be a valid use case. But on the other hand, having the refresh functionality in the NewsList isn't a nice solution either.
> 
> If you want i can create an issue 'Move infinite scroll behaviour to PayloadProvider' or something like this. Are there other components with infitinte scroll?

No, no other components with that behaviour. But let's just keep it this way for now, that is definitely fine. Once we run into problems or we have another infinite scroll component we can think about changing it too.

PR #169, opened 2021-02-23
IGAPP-15: Setup e2e tests

Taggotty:


e2e-tests/web/browserstack/wdio-browserstack-chrome.conf.js
```diff
@@ -0,0 +1,7 @@
+const defaultConfig = require('./wdio-browserstack.conf')
+
+const config = Object.assign({}, defaultConfig.config)
+config.capabilities = config.capabilities.filter(cap =>
+    cap.browserName.toLowerCase() === 'chrome'
```
Taggotty, line 5: This is probably done best with args passed to the config. But for an initial setting I think this is enough

e2e-tests/web/browserstack/wdio-browserstack-chrome.conf.js
```diff
@@ -0,0 +1,7 @@
+const defaultConfig = require('./wdio-browserstack.conf')
```
klinzo, line 1: Shouldn't we type these files as well?

e2e-tests/web/test/pageobjects/landing.page.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    public constructor(language: string = 'en') {
+        super();
+        this.language = language
+    }
+
+    get cities () { return $$('//main//a') }
+    get search () { return $('//main//input') }
+
+    city(name: string) {
```
klinzo, line 18: I think you should apply formatting to all files. Does eslint not work here?

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,34 @@
+import LandingPage from '../pageobjects/landing.page';
```
klinzo, line 1: Why `test/specs` directory instead of just `test` or `specs`?

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+
+describe('Landing Page', () => {
+    it('Filter and Navigate to City', async () => {
```
klinzo, line 4:
```suggestion
describe('LandingPage', () => {
    it('should filter and navigate to city', async () => {
```

klinzo: Also I am not really sure if we shouldn't name the test case (`Landing Page`) the same way as the file name (`navigateToDashboard`) or the other way round :thinking:

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+        // given
+        const language = 'en'
+        const filter = 'Augsburg'
+        const dashboardPath = `/${filter.toLowerCase()}/en`
+        await browser.setTimeout( {'implicit': 80000})
```
klinzo, line 9:
```suggestion
        await browser.setTimeout({ 'implicit': 80000 })
```

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+        LandingPage.language = language
+        await LandingPage.open();
```
klinzo, line 11:
```suggestion
        await LandingPage.open()
```

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
klinzo: Isn't there already a timeout in `wdio.conf.js`?

klinzo (changes requested):
Looking really easy to write tests now, well done!
Things I am currently missing:
* Linting does not seem to work
* Lots of files are not typed

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
Taggotty: This is the autogenerated structure. But I think the main reason ist to create a container for both pageobjects and specs

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
Taggotty: I think especially for the E2E tests it wouldn't make sense to use `Landin Page` as E2E tests should enclose UserStories, which include multiple use cases. But maybe a better naming would be sth. like `navigation`, `search`, ... which would couple multiple cases

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
Taggotty: not for the implicit wait. it seems that the timeout is not set when you add this timeout to the capabilities

docs/e2e-tests.md
```diff
@@ -0,0 +1,25 @@
...
+
+> **Note**
+> WebdriverIO provides selectors for react components. However, in they do not work properly for internet explorer in web. Therefore, you should use more primitive selectors instead (like xpath).
+
+## Browserstack
+Browserstack provides you with a large set of available operating systems, browser, and browser versions.
+To see and run the E2E tests on Browserstack you need a Browserstack account, with access to our organization. 
+You will need to set `E2E_BROWSERSTACK_KEY` and `E2E_BROWSERSTACK_USER` as environment variables, which you can find when logging in to Browserstack and selecting either the Automate or AppAutomate tab.
+If you don't have this access you can still execute the tests locally with `yarn e2e:web`. 
+For the browserstack tests in web there are several configurations to run the tests in the specified browsers. Preferably you should run these from the commandline, e.g. `npx wdio run web/browserstack/wdio-browserstack-safari.conf.js` from the project root. There is also a configuration in the `package.json`, which runs all of the browserstack configurations sequentially.
```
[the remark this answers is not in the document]
Taggotty: I did not know that npx is basically the same as yarn but for npm. Never have I worked with that.

Taggotty:
> * Linting does not seem to work

Fixed 👍🏼 
> * Lots of files are not typed

I tried to find a solution to implement the config files as .ts files, but couldn't make it work :/ (https://github.com/webdriverio/webdriverio/issues/1509) Maybe @maxammann want's to experiment with that? 
Apart from that the browserstack configs could be merged at some point. But that I would do in separate prs

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
klinzo: Yeah I agree. I just think file name should be the same as describe

klinzo (approved):

e2e-tests/package-lock.json
```diff
@@ -0,0 +1,3165 @@
+{
```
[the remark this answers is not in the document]
klinzo: Probably still relevant

docs/e2e-tests.md
```diff
@@ -0,0 +1,29 @@
...
+
+Browserstack provides you with a large set of available operating systems, browser, and browser versions.
+To see and run the E2E tests on Browserstack you need a Browserstack account, with access to our organization. 
+You will need to set `E2E_BROWSERSTACK_KEY` and `E2E_BROWSERSTACK_USER` as environment variables, which you can find when logging in to Browserstack and selecting either the Automate or AppAutomate tab.
+If you don't have this access you can still execute the tests locally with `yarn e2e:web`. 
+For the browserstack tests in web there are several configurations to run the tests in the specified browsers. Preferably you should run these from the commandline, e.g. `yarn wdio run web/browserstack/wdio-browserstack-safari.conf.ts` from the project root. There is also a configuration in the `package.json`, which runs all of the browserstack configurations sequentially. 
+
+## Writing Tests
+
+The E2E-tests project does not use flow, but implements everything using [typescript](https://www.typescriptlang.org/).
```
maxammann, line 23:
```suggestion
The E2E-tests project does not use flow, but implements everything using [TypeScript](https://www.typescriptlang.org/).
```

maxammann:
@Taggotty just read this issue https://github.com/webdriverio/webdriverio/issues/1509 Then I saw the backlink at the bottom to this issue :D

e2e-tests/web/browserstack/wdio-browserstack-chrome.conf.js
maxammann: Flow should not be added to this project as we have TypeScript. Also writing this in TS is not really possible: https://github.com/rah-foundation/website/pull/10

I think we should not hack our way around because in the end there are probably no types available for the configs.

maxammann: @Taggotty did you try ES6 syntax and modules though?

e2e-tests/web/browserstack/wdio-browserstack.conf.js
```diff
@@ -0,0 +1,72 @@
...
+      debug: true,
+      projectName: 'integreat-app-web'
+    },
+    browserName: config.browserName,
+    acceptInsecureCerts: true
+  }
+}
+
+exports.config = merge(defaultConfig.config, {
+  maxInstances: 1,
```
maxammann, line 30: Can we set this to 2 or even 5? Did you try this? 

Asking because of native e2e tests

e2e-tests/web/test/pageobjects/page.ts
```diff
@@ -0,0 +1,15 @@
...
+ * main page object containing all methods, selectors and functionality
+ * that is shared across all page objects
+ */
+
+export default class Page {
+  /**
+     * Opens a sub page of the page
+     * @param path path of the sub page (e.g. /path/to/page.html)
+     */
+
```
maxammann, line 11:
```suggestion
  /**
   * Opens a sub page of the page
   * @param path path of the sub page (e.g. /path/to/page.html)
   */
```

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,33 @@
+import LandingPage from '../pageobjects/landing.page'
+
+describe('navigate to dashboard', () => {
+  it('Filter and Navigate to City', async () => {
+    // given
```
maxammann, line 5:
```suggestion
```

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+    const language = 'en'
+    const filter = 'Augsburg'
+    const dashboardPath = `/${filter.toLowerCase()}/en`
+    await browser.setTimeout({ implicit: 80000 })
```
maxammann, line 9: Why do we need this timeout? Should we set this globally?

e2e-tests/web/wdio.conf.js
```diff
@@ -0,0 +1,33 @@
...
+  specs: [
+    './web/test/specs/**/*.ts'
+  ],
+  exclude: [],
+  maxInstances: 10,
+
+  capabilities: [{
+    maxInstances: 5,
+    browserName: 'chrome',
+    acceptInsecureCerts: true
```
maxammann, line 12: Probably fine as these are just tests, but is this required?

e2e-tests/web/wdio.conf.js
```diff
@@ -0,0 +1,33 @@
...
+  framework: 'jasmine',
+  reporters: ['junit'],
+
+  jasmineNodeOpts: {
+    defaultTimeoutInterval: 120000
+  },
+
+  onPrepare: async function () {
+    const startupDelay = 20000
+    await new Promise(resolve => setTimeout(resolve, startupDelay))
```
maxammann, line 31: Are we always waiting 20s before running tests? If yes, why?

e2e-tests/package-lock.json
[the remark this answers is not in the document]
maxammann: Yep, this should be removed

maxammann (approved):

Taggotty:
🥳 @maxammann @klinzo Since two weeks or so it is possible to use typescript config files. Feel free to _shortly(!)_ rereview the config files

e2e-tests/web/browserstack/wdio-browserstack-chrome.conf.js
Taggotty: See [comment](https://github.com/Integreat/integreat-app/pull/169#issuecomment-789753282)

e2e-tests/web/browserstack/wdio-browserstack.conf.js
Taggotty: yes. Just for the web e2e tests I don't think it makes sense as we will not use browserstack in the ci

e2e-tests/.eslintrc.js
```diff
@@ -0,0 +1,48 @@
+module.exports = {
```
klinzo, line 1: Why does this not use the main `.eslintrc.js` file? Or does it?

klinzo: Is linting part of CI checks? Does not has to be done in this PR, maybe just add a note to the e2e CI issue.

klinzo (approved):
:tada:

e2e-tests/.eslintrc.js
Taggotty: as we want linting for typescript files the rules differ. I think I will look into inheriting `eslintrc` files and check whether they make sense here

Taggotty: > Is linting part of CI checks? Does not has to be done in this PR, maybe just add a note to the e2e CI issue.

at least there currently is a command for linting in the ci. But I didn't find where it was actually executed.

e2e-tests/yarn.lock
```diff
@@ -2926,11 +2926,6 @@ to-regex-range@^5.0.1:
   dependencies:
```
klinzo, line 2926: So this is not a monorepo workspace, right?

e2e-tests/web/wdio.conf.js
Taggotty: To wait for the webpack-server, as yarn start is executed beforehand. As the jasmine timeout is quite large it might be unnecessary, but I think this way is more correct.

e2e-tests/web/test/specs/navigateToDashboard.e2e.ts
Taggotty: I moved it into the config. It specifies how long we wait for location strategies such as findElement, $, $$, ...

e2e-tests/yarn.lock
Taggotty: now it is. Kind of missed that

e2e-tests/web/wdio.conf.js
Taggotty: removed. was the result of some copy pasting. But might be necessary for native.

e2e-tests/.eslintrc.js
klinzo: It seems like it has to be called separately for every workspace. See `check_web` and `check_native`

e2e-tests/yarn.lock
klinzo: Can't we remove this yarn.lock then?

maxammann (approved):

PR #173, opened 2021-02-26
IGAPP-433: Add imprint

klinzo:


klinzo:
This is just accessible by going to exactly this url though as the city specific disclaimer is fetched from the cms.

Taggotty (approved):

SteffiStuffel (approved):

PR #175, opened 2021-02-26
IGAPP-453: Missing store data

klinzo:


web/src/modules/layout/components/LocationToolbar.js
```diff
@@ -22,10 +22,6 @@ class LocationToolbar extends React.PureComponent<PropsType> {
         {children}
         <FeedbackToolbarItem isPositiveRatingLink openFeedbackModal={openFeedbackModal} viewportSmall={viewportSmall} />
         <FeedbackToolbarItem isPositiveRatingLink={false} openFeedbackModal={openFeedbackModal} viewportSmall={viewportSmall} />
-        {/* todo: Add these functionalities:
```
Taggotty, line 25: Did we decide against implementing this or is this just not relevant anymore?

Taggotty (approved):

web/src/modules/layout/components/LocationToolbar.js
klinzo: This Todo is there for probably more then two years now and we have never acted on it, so I just figured I'll remove it. If this is functionality that should be added it'll come up again. The right place would be in the issue tracker anyway.

SteffiStuffel (approved):

PR #178, opened 2021-03-01
IGAPP-492: Fix share url for offers and news pages

SteffiStuffel:


native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -102,4 +105,118 @@ describe('HeaderContainer', () => {
...
+    const ownProps = {
+      scene: {
+        route: {
+          name: EVENTS_ROUTE,
+          key: 'routeKeyEvent1'
+        }
+      }
+    }
+
+    const result = TestRenderer.create(
```
klinzo, line 136: Is there a relatively easy way to test the shareUrl via the react native testing library instead of the TestRenderer? Asserting on props is not really good style. Probably quite difficult in this case (or rather: for containers in general).

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -102,4 +105,118 @@ describe('HeaderContainer', () => {
...
+    )
+
+    const header = result.root.findByType(Header)
+
+    expect(header.props).toEqual(
+      expect.objectContaining({ shareUrl: cityContentUrl({ cityCode: city.code, languageCode: language._code, route: NEWS_ROUTE, path: LOCAL_NEWS_TYPE }) })
+    )
+  })
+
+  it('shareUrl should be set correctly for offers route', () => {
```
klinzo, line 194: Tests for pois and disclaimer are missing, could you add some? I think you could also generalize this kind of test by creating helper functions to render and assert. I think it is also not necessary to create the state separately each time, you can just set all mappings.

native/src/modules/layout/components/Header.js
```diff
@@ -66,7 +65,7 @@ export type PropsType = {|
   categoriesAvailable: boolean,
   goToLanguageChange?: () => void,
   routeCityModel?: CityModel,
-  language: string,
+  language: mixed,
```
klinzo, line 68: Why mixed?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -9,18 +9,25 @@ import withTheme from '../../theme/hocs/withTheme'
...
-import { CHANGE_LANGUAGE_MODAL_ROUTE, EVENTS_ROUTE, NEWS_ROUTE, POIS_ROUTE } from 'api-client/src/routes'
 
 type OwnPropsType = {|
   ...StackHeaderProps,
   t: TFunction
 |}
 
 type StatePropsType = {|
-  language: string,
+  language: mixed,
```
klinzo, line 30: Why mixed?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -32,19 +39,9 @@ type DispatchPropsType = {| dispatch: Dispatch<StoreActionType> |}
...
-  } else if ([EVENTS_ROUTE, NEWS_ROUTE, POIS_ROUTE].includes(routeName)) {
-    const params = { cityCode: city, languageCode: language, path: null }
-    return cityContentUrl({ route: routeName, ...params })
-  }
-  return null
-}
-
 const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsType => {
   const routeKey = ownProps.scene.route.key
+  const shareUrlFromScene: ?string = typeof ownProps.scene.route.params?.shareUrl === 'string' ? ownProps.scene.route.params?.shareUrl : null
```
klinzo, line 44: I think this is no longer relevant as you removed it completely from the route params, is it?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -53,14 +50,37 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     state.cityContent.poisRouteMapping[routeKey]
     : null
 
+  if (([OFFERS_ROUTE, DISCLAIMER_ROUTE, SPRUNGBRETT_OFFER_ROUTE]).includes(ownProps.scene.route.name) || shareUrlFromScene) {
```
klinzo, line 53: Why do you handle this partly here and partly below? It's a little bit confusing this way

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -53,14 +50,37 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+
   const languages = state.cityContent?.languages
 
   // prevent re-rendering when city is there.
   const cities = state.cities.models || []
   const stateCityCode = state.cityContent?.city
   const categoriesAvailable = state.cityContent?.searchRoute !== null
 
-  const routeCityModel = route ? cities.find(city => city.code === route.city) : undefined
+  const routeCityModel = route ? cities.find(cityElem => cityElem.code === route.city) : undefined
```
klinzo, line 83: Why this change?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -87,13 +107,21 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
-  const shareUrlFromScene: ?string = typeof ownProps.scene.route.params?.shareUrl === 'string' ? ownProps.scene.route.params?.shareUrl : null
-  const { city, language } = route
-  const path = route.path || null
-  const shareUrl = shareUrlFromScene || getShareUrl({ city, language, path, routeName: ownProps.scene.route.name })
+  const path = route.path || undefined
+  const language = route.language
+  let shareUrl = url(path) // getShareUrl({ city, language, path, routeName: ownProps.scene.route.name })
+
+  if (ownProps.scene.route.name === NEWS_ROUTE && route?.type && stateCityCode) {
+    shareUrl = cityContentUrl({ cityCode: stateCityCode, languageCode: language, route: NEWS_ROUTE, path: route.type })
```
klinzo, line 117: So sharing a single news is still not possible, right? Could you fix it or create an issue please?

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -87,13 +107,21 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+  }
+
+  if (ownProps.scene.route.name === EVENTS_ROUTE && !path && stateCityCode) {
```
klinzo, line 120: Why do you check for path here while you don't for news? I think it can be handled the same way, same for pois.

klinzo (changes requested):

native/src/modules/layout/__tests__/HeaderContainer.spec.js
SteffiStuffel: Yeah i tried with react testing library a bit, but wasn't successful. If you have a better idea, how to do this, let me know.

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: no it is still relevant, e.g. for externalOffers

klinzo: External offers should be external and not have a screen, therefore normally no shareUrl should be necessary :eyes:

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: forget to undo... reseted it

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: created an issue

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: In pois and events the path is set to null, which means, we are on the overview page, therefore the shareUrl can not be url(path), but must explicity set to 'city/lang/events' or 'city/lang/pois'.

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: you are completly right it is a bit confusing. I did it because afterwards by checking if (!route ...) (line 85) it would exit for every route not being in an routeMapping without setting the route correctly.

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: I meant for example for Lehrstellenradar

Just found: https://www.lehrstellen-radar.de/5100,0,lsrlist.html

native/src/modules/layout/__tests__/HeaderContainer.spec.js
SteffiStuffel: added tests for disclaimer and pois and also added a render and assert function and you are completly right, now it is a lot simpler. :)

native/src/modules/layout/components/Header.js
SteffiStuffel: Flow is complaining, because 

Cannot return object literal because mixed [1] is incompatible with string [2] in property language.

flow-typed/npm/@react-navigation/stack_v5.x.x.js
[1] 428│   declare export type ScreenParams = { +[key: string]: mixed, ... };

Any better idea to fix this, than chaning to mixed?

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -56,74 +114,47 @@ describe('HeaderContainer', () => {
...
+    return TestRenderer.create(
       <Provider store={store}>
         {/* $FlowFixMe not all props passed */}
-        <HeaderContainer {...ownProps}/>
+        <HeaderContainer {...props}/>
       </Provider>
     )
+  }
 
+  const assert = (result, expected) => {
```
klinzo, line 135:
```suggestion
  const assertProps = (result, expected) => {
```
Also not sure if `result` is a good param name.

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -56,74 +114,47 @@ describe('HeaderContainer', () => {
...
-        allAvailableLanguages: new Map()
+    const ownProps = {
+      scene: {
+        route: {
+          key: 'routeKey1'
+        }
       }
     }
+    const result = render(ownProps)
+    const expectedShareUrl = url(`${city.code}/${language._code}/abc`)
```
klinzo, line 152: Just use `cityContentUrl`

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -132,41 +163,28 @@ describe('HeaderContainer', () => {
...
-      </Provider>
-    )
-
-    const header = result.root.findByType(Header)
-
-    expect(header.props).toEqual(
-      expect.objectContaining({ shareUrl: cityContentUrl({ cityCode: city.code, languageCode: language._code, route: EVENTS_ROUTE, path: null }) })
-    )
+    const result = render(ownProps)
+    const expectedShareUrl = cityContentUrl({ cityCode: city.code, languageCode: language._code, route: EVENTS_ROUTE, path: null })
```
klinzo, line 167:
```suggestion
    const expectedShareUrl = cityContentUrl({ cityCode: city.code, languageCode: language.code, route: EVENTS_ROUTE, path: null })
```
Same for other occurrances

native/src/modules/layout/containers/HeaderContainer.js
klinzo: I see, this thing again :cry: I'd omit the shareUrl here to be honest. Reason 1: The url does not work anyway as it has to be opened with a post request, so if you just klick on it you just get HTTP_METHOD_NOT_SUPPORTED. Reason 2: Even if it would work, the url would be wrong once the user navigates anywhere. Probably we should look into whether it is possible to support sharing the url some other way or just ignore it.

native/src/modules/layout/components/Header.js
klinzo: I'd rather use a typeof check (or a flow fix me) than having a mixed type for something that is always a string. What do you think?

native/src/modules/layout/__tests__/HeaderContainer.spec.js
klinzo: Not really. @maxammann any ideas?

native/src/modules/layout/containers/HeaderContainer.js
klinzo: But we still need it for `image` and `pdf` routes, nvm. But it is not relevant in the HeaderContainer since those routes are using a transparent header.

klinzo:
@SteffiStuffel see my proposal in #186. You don't have to use it but I think it makes some logic a little bit simpler and easier to understand.

native/src/modules/layout/components/Header.js
SteffiStuffel: You are completly right, a typeof check is much nicer. I added it.

native/src/modules/layout/__tests__/HeaderContainer.spec.js
SteffiStuffel: done

native/src/modules/layout/__tests__/HeaderContainer.spec.js
SteffiStuffel: done

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: Should i open a ticket, so we can discuss this in the grooming ?

SteffiStuffel:
> @SteffiStuffel see my proposal in #186. You don't have to use it but I think it makes some logic a little bit simpler and easier to understand.

I'm with you and merged it to my branch.

native/src/modules/layout/containers/HeaderContainer.js
klinzo: Yeah sure, opening a ticket for sharing the Lehrstellenradar makes sense. I think I'd remove the logic anyway for now as it does not work.

klinzo (approved):
Well done :+1:

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -87,13 +92,20 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
     })
   }
+
   const peeking = isPeekingRoute(state, { routeCity: route.city })
-  const shareUrlFromScene: ?string = typeof ownProps.scene.route.params?.shareUrl === 'string' ? ownProps.scene.route.params?.shareUrl : null
-  const { city, language } = route
-  const path = route.path || null
-  const shareUrl = shareUrlFromScene || getShareUrl({ city, language, path, routeName: ownProps.scene.route.name })
+  const language = route.language
+  const path = route.path || undefined
```
Taggotty, line 98: why can't you just write `const path = route.path` ?

Taggotty (approved):

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -53,6 +47,17 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     state.cityContent.poisRouteMapping[routeKey]
     : null
 
+  const simpleRoutes = [OFFERS_ROUTE, DISCLAIMER_ROUTE, SPRUNGBRETT_OFFER_ROUTE]
```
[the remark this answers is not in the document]
SteffiStuffel: routes from saved in routeMappings, eg. categoiresRouteMapping. Also events, news and pois

native/src/modules/layout/containers/HeaderContainer.js
SteffiStuffel: @klinzo Its part of you code, isn't it?

klinzo: Possibly :thinking: Not sure if there was a reason for this, but if there is a reason it's because of flow. Just try it and simplify it as suggested if it works.

SteffiStuffel: Yeah its Flow.

Cannot get route.path because property path is missing in object type [1].
where [1] is NewsRouteState Type.

PR #179, opened 2021-03-01
IGAPP-405: Wrong screenshot order

klinzo:

Seems like we have to set `overwrite_screenshots` to true for one delivery in order to force the right order. Currently screenshot upload is skipped:
[05:43:48]: Starting with the upload of screenshots...
[05:43:49]: Previous uploaded. Skipping './ios/fastlane/integreat/screenshots/en-US/01_ipadPro129_1.iPAd129_portrait_01.jpg'...
...
[05:43:49]: Waiting for all the screenshots processed...
[05:43:50]: Successfully uploaded all screenshots
[05:43:50]: Sorting screenshots uploaded...
[05:43:51]: Successfully uploaded screenshots to App Store Connect

The flag (https://docs.fastlane.tools/actions/deliver/#parameters):
overwrite_screenshots | Clear all previously uploaded screenshots before uploading the new ones | false


native/fastlane/Fastfile
```diff
@@ -286,6 +286,7 @@ platform :ios do
         force: true,
         skip_screenshots: skip_screenshots,
         skip_metadata: false,
+        overwrite_screenshots: true, # TODO Remove again after next delivery
```
maxammann, line 289: We can also just keep this flag set to true

maxammann (approved):

SteffiStuffel (approved):

native/fastlane/Fastfile
klinzo: You are probably right, would be the more intuitive behavior. Removed the TODO.

PR #180, opened 2021-03-01
IGAPP-489: Only show timestamp without text for local news

zankerbe:
I modified Page.js so that the text "last update" is shown by default, but can be turned off. Timestamps for local news are now shown without this text.


web/src/modules/common/components/Page.js
```diff
@@ -25,6 +25,7 @@ type PropsType = {|
   thumbnailSrcSet?: ?string,
   content: string,
   lastUpdate?: Moment,
+  showUpdateText?: boolean,
```
klinzo, line 28: I'd rename this to `showLastUpdateText` as that is what the component is called, but apart from that nicely done :+1:

klinzo (approved):
:tada:

klinzo:
Tested on firefox (testumgebung)

web/src/modules/common/components/Page.js
zankerbe: Makes sense, i updated it

SteffiStuffel:
Nice job. Tested on Chrome.

Just a little note: normally we name our branches with dashes (-) and not with unterscore (_)

SteffiStuffel (approved):

PR #181, opened 2021-03-01
IGAPP-491: Optimize linting for hooks

klinzo:


SteffiStuffel (approved):

SteffiStuffel:
Seems valid. 

But i think we use useCallback way to often. Actually useCallback makes a component slower and only makes sense if we prevent (heavy) child components to rerender by memorizing the function or if the function is a dependency to other hooks.

klinzo:
> Seems valid.
> 
> But i think we use useCallback way to often. Actually useCallback makes a component slower and only makes sense if we prevent (heavy) child components to rerender by memorizing the function or if the function is a dependency to other hooks.

I guess you are totally right... https://dmitripavlutin.com/dont-overuse-react-usecallback/
So probably we have nearly no valid usecase of useCallback at all. So probably we should remove our useCallbacks (nearly) completely. Our options then are:
* Removing the linting rule and not caring at all
* Refactoring arrow functions to actual functions
* Extracting functions from our components
Am I missing anything here? Other opinions? @maxammann @Taggotty

maxammann:
> I guess you are totally right... https://dmitripavlutin.com/dont-overuse-react-usecallback/

I think that useCallback does not make the rendering slower. At least that post does not mention this.
It is right that it is added complexity though.

The question is: Is it worth thinking about whether to use `useCallback`. Or do we just blindly use it everywhere.

I'm more on the side of "thinking" which means not using useCallback at all. Any if we have performance problems then profile our app and fix issues.

maxammann:
So I'm fine with removing that rule which prohibits to pass lambdas to child components.

klinzo:
I don't think it makes rendering much slower but it sure does as it adds additional overhead. But I guess we agree on just removing the rule. Then we can gradually remove useCallbacks again.

maxammann:
Maybe we can remove some excessive use of useCallback in this PR. I remember some file which had over 10 usages I think.

klinzo:
> Maybe we can remove some excessive use of useCallback in this PR. I remember some file which had over 10 usages I think.

Removed some :grin:

Taggotty (approved):

PR #182, opened 2021-03-03
IGAPP 173: (Re-)Update non-breaking libraries

Znippet:

@react libraries will be updated in a different branch
(https://issues.tuerantuer.org/browse/IGAPP-484)


api-client/package.json
```diff
@@ -20,7 +20,7 @@
     "@babel/runtime": "7.13.7",
     "js-md5": "^0.7.3",
     "entities": "^2.0.0",
-    "moment": "2.25.3",
+    "moment": "2.29.1",
```
klinzo, line 23: :fearful: Hopefully this does not break anything :D

native/package.json
```diff
@@ -56,9 +56,9 @@
     "events": "^3.2.0",
     "hashids": "^2.2.8",
     "htmlparser2": "^6.0.0",
-    "i18next": "^19.1.0",
+    "i18next": "19.9.1",
```
klinzo, line 59: Why did you remove `^`?

klinzo: Same for most of the other dependencies

klinzo (approved):

native/package.json
klinzo: I just recognized you did the same for babel dependencies in a previous PR, right?

api-client/package.json
Taggotty: So I checked the release notes and it seems, that there were no breaking changes. However I found [this](https://momentjs.com/docs/#/-project-status/) interesting article about moment becoming a legacy project o.o

klinzo:
Tested on firefox and real android 10 device.

Taggotty (approved):

api-client/package.json
klinzo: Yeah crazy, right? We already have an issue to switch away from moment but we wanted to wait until we get correct timestamps from the cms which will only be the case with django.

Taggotty:
roughly tested on android simulator.

native/package.json
Znippet: I thing it doesn't make much of a difference. With the ^ yarn upgrade will only do minor version upgrades.
But with the lockfile it won't do anything automatically anyway. So if we upgrade we should try to upgrade to the latest version because testing is necessary either way.

The standard behavior of installing a new packet via yarn upgrade is adding the "^" by default. 
I can also re-add it, I don't care

PR #183, opened 2021-03-03
IGAPP-487: Support html entities for news

klinzo:


Taggotty (approved):

SteffiStuffel (approved):
Renaming really makes sense here. Tested and workes fine.

PR #184, opened 2021-03-04
IGAPP-490: Add Timestamp for Local News

Taggotty:

### DIFF
--- a/native/src/modules/common/components/TimeStamp.js
+++ b/native/src/modules/common/components/TimeStamp.js
@@ -29,18 +29,19 @@ type PropsType = {|
   formatter: DateFormatter,
   t: TFunction,
   language: string,
-  theme: ThemeType
+  theme: ThemeType,
+  useText?: boolean
 |}
 
 export class TimeStamp extends React.PureComponent<PropsType> {
   render () {
-    const { lastUpdate, formatter, t, language, theme } = this.props
+    const { lastUpdate, formatter, t, language, theme, useText } = this.props
     // only show day, month and year
     const dateText = formatter.format(lastUpdate, {
       format: 'LL'
     })
     return <DirectionContainer language={language} theme={theme}>
-      <TimeStampText theme={theme}>{t('lastUpdate')} </TimeStampText>
+      {useText !== false && <TimeStampText theme={theme}>{t('lastUpdate')} </TimeStampText>}
       <TimeStampText theme={theme}>{dateText}</TimeStampText>
     </DirectionContainer>
   }
--- a/native/src/routes/news/components/NewsListItem.js
+++ b/native/src/routes/news/components/NewsListItem.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as React from 'react'
+import { useContext } from 'react'
 import { LocalNewsModel, TunewsModel } from 'api-client'
 import styled from 'styled-components/native'
 import { type TFunction } from 'react-i18next'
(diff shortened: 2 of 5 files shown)


native/src/modules/common/components/TimeStamp.js
```diff
@@ -29,18 +29,19 @@ type PropsType = {|
   formatter: DateFormatter,
   t: TFunction,
   language: string,
-  theme: ThemeType
+  theme: ThemeType,
+  useText?: boolean
```
klinzo, line 33: I think `showText` or similar would be better. This looks like a hook :thinking:

native/src/modules/common/components/TimeStamp.js
```diff
@@ -29,18 +29,19 @@ type PropsType = {|
...
 |}
 
 export class TimeStamp extends React.PureComponent<PropsType> {
   render () {
-    const { lastUpdate, formatter, t, language, theme } = this.props
+    const { lastUpdate, formatter, t, language, theme, useText } = this.props
```
klinzo, line 38:
```suggestion
    const { lastUpdate, formatter, t, language, theme, useText = true } = this.props
```
With this default value you could just use `useText && <TimeStampText ... ` below.

native/src/routes/news/components/__tests__/NewsDetail.spec.js
```diff
@@ -44,6 +44,7 @@ describe('NewsDetail', () => {
     expect(getByText('mailto:app@integreat-app.de')).toBeTruthy()
     expect(getByText(' with lots of links ')).toBeTruthy()
     expect(getByText('https://integreat.app')).toBeTruthy()
+    expect(getByText('January 20, 2020')).toBeTruthy()
```
klinzo, line 47: Maybe you could add a check that no `lastUpdate` text is shown?

native/src/routes/news/components/__tests__/NewsListItem.spec.js
```diff
@@ -43,7 +43,7 @@ describe('NewsListItem', () => {
 
     expect(getByText(localNews.title)).toBeTruthy()
     expect(getByText('Some "test text with lots of "html entities" which won\'t be displayed.')).toBeTruthy()
-
+    expect(getByText('January 20, 2020')).toBeTruthy()
```
klinzo, line 46: Maybe you could add a check that no `lastUpdate` text is shown?

native/src/routes/news/components/NewsListItem.js
```diff
@@ -99,9 +110,11 @@ export const ReadMore: StyledComponent<{| isTunews: ?boolean |}, ThemeType, *> =
 `
 
 const NewsListItem = ({ newsItem, language, navigateToNews, theme, t, isTunews }: PropsType) => {
+  const formatter = useContext(DateFormatterContext)
   const localNewsContent = newsItem instanceof LocalNewsModel ? newsItem.message : ''
   const tuNewsContent = newsItem instanceof TunewsModel ? newsItem.content : ''
   const content = localNewsContent || tuNewsContent
+  const created = newsItem instanceof LocalNewsModel ? newsItem.timestamp : null
```
klinzo, line 117:
```suggestion
  const timestamp = newsItem instanceof LocalNewsModel ? newsItem.timestamp : null
```

klinzo (approved):

klinzo:
Tested on android 10 device. Maybe you could add a little more space between the content and the date on the local news detail?
Also do you think it would make sense to show a time as well? Not sure here myself :thinking:

klinzo (approved):
You decided to show the time as well, right? Then we should probably create an issue to do the same for web :+1:

SteffiStuffel (approved):

PR #186, opened 2021-03-04
IGAPP-492: Proposal

klinzo:

### DIFF
--- a/native/src/modules/layout/containers/HeaderContainer.js
+++ b/native/src/modules/layout/containers/HeaderContainer.js
@@ -11,13 +11,10 @@ import { type Dispatch } from 'redux'
 import type { StoreActionType } from '../../app/StoreActionType'
 import {
   CityModel,
-  DISCLAIMER_ROUTE,
-  OFFERS_ROUTE,
   CHANGE_LANGUAGE_MODAL_ROUTE,
-  EVENTS_ROUTE,
-  NEWS_ROUTE,
-  SPRUNGBRETT_OFFER_ROUTE,
-  POIS_ROUTE
+  OFFERS_ROUTE,
+  DISCLAIMER_ROUTE,
+  SPRUNGBRETT_OFFER_ROUTE
 } from 'api-client'
 import isPeekingRoute from '../../endpoint/selectors/isPeekingRoute'
 import { cityContentUrl, url } from '../../navigation/url'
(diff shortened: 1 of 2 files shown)


native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -84,15 +68,15 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+    !languages || languages.status !== 'ready' || !stateCityCode) {
     // Route does not exist yet. In this case it is not really defined whether we are peek or not because
     // we do not yet know the city of the route.
     return {
       language: state.contentLanguage,
       routeCityModel,
       peeking: false,
       categoriesAvailable: false,
-      shareUrl: null
+      shareUrl: simpleRouteShareUrl
```
klinzo, line 79: Optional: Keep the `shareUrlFromScene` above and change this to
```suggestion
      shareUrl: shareUrlFromScene || simpleRouteShareUrl
```.
It is not really needed though as the only routes with share url still in the route params will be image and pdf routes which are not using the normal header and therefore this header container.

**Comment by klinzo** [comment]
@SteffiStuffel I thought about some further refactoring of your code regarding the share url. What do you think? Imo this makes it a little bit easier to understand and simpler code. You don't have to use this though. Maybe some comments would be helpful.

File: native/src/modules/layout/containers/HeaderContainer.js [hunk 2]
```diff
@@ -51,28 +47,16 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
-
-    return {
-      language: propsRoute.params?.languageCode,
-      routeCityModel: undefined,
-      peeking: false,
-      categoriesAvailable: false,
-      shareUrl: shareUrlFromScene || shareUrl
-    }
-  }
+  const simpleRoutes = [OFFERS_ROUTE, DISCLAIMER_ROUTE, SPRUNGBRETT_OFFER_ROUTE]
```
SteffiStuffel, line 50: makes it really clearer to give them a name

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -110,17 +94,16 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
-  }
+  const path = route.path || undefined
+  const shareUrl = path
+    ? url(path)
+    : cityContentUrl({
+      cityCode: stateCityCode,
+      languageCode: route.language,
+      route: routeName,
+      path: route.type || null
+    })
```
SteffiStuffel, line 106: This is a lot handier than my solution.

SteffiStuffel (commented):
I tested this and it works perfectly, (unless i miss something). Merged it into my branch

PR #188, opened 2021-03-05
IGAPP-511: Change logo for world womens day

klinzo:

Tried to make it very generic so we can reuse the logic at some later point. I decided to make it date based in order not to be dependent on a delivery to start/stop the campaign.


maxammann (approved):

web/src/modules/layout/components/Header.js
```diff
@@ -100,41 +100,40 @@ const NavigationBar: StyledComponent<{||}, ThemeType, *> = styled.div`
...
       <Headroom onStickyTopChanged={onStickyTopChanged}
                 scrollHeight={scrollHeight}
                 height={height}
                 positionStickyDisabled={platform.positionStickyDisabled}>
         <HeaderContainer>
           <Row hasTitle={!!cityName}>
-            <HeaderLogo link={logoHref}
-                        src={buildConfig().icons.appLogo}
-                        alt={buildConfig().appName} />
+            <HeaderLogo link={logoHref} buildConfig={buildConfig()} />
```
Taggotty, line 128: Imo it is better to call `buildConig()` more locally, as it prevents those blown up snapshots. Is there a reason to pass it as property?

Taggotty (approved):

web/src/modules/layout/components/Header.js
klinzo: Testing was easier that way. Do you think it is better not to pass the build config as property apart from those blown up snaps? Those will be removed anyway sooner or later, so I figured it does not really matter.

Taggotty: If it simplifies testing then I guess it's fine. As stated, it's just my personal preference.

PR #189, opened 2021-03-05
IGAPP-407: Only deliver releases to production after testing in development

klinzo:


klinzo:
I tested most of the lanes and commands already. Exception: Actual promotion from testflight to app store connect.

Promotion will fail for aschaffenburg if we don't create production releases beforehand but I think that will still take some time. For the ios release notes I had to make things a little bit hacky as one can't upload release notes and metadata for testflight builds.

.circleci/src/workflows/scheduled_delivery.yml
```diff
@@ -249,7 +249,7 @@ jobs:
       context:
         - mattermost
         - deliverino
-      production_delivery: true
+      production_delivery: false
```
maxammann, line 252: This flag is now always false, right? Should we remove that functionality?

maxammann (approved):

.circleci/src/workflows/scheduled_delivery.yml
klinzo: Not for triggered deliveries. I think we should still keep this as we may need it sometime.

Taggotty (approved):
Seems actually not that complicated. Although of course I am not that experienced with this release stuff.

PR #190, opened 2021-03-05
IGAPP-511: Fix womens day start date

klinzo:


klinzo:
@Taggotty you were totally right, thought about it a little more and was not good style. I noticed it is not at all difficult to test it as well since modifying the build config is easily possible

web/src/modules/layout/components/__tests__/HeaderLogo.spec.js
```diff
@@ -16,49 +16,47 @@ describe('HeaderLogo', () => {
...
+    config = previousConfig
   })
 
   it('should show the regular header app icon if there is no campaign', () => {
     clock = lolex.install({ now: 1615374110000, toFake: [] }) // Wed Mar 10 2021 11:01:50 GMT+0000
     config.campaign = undefined
-    config.icons.appLogo = 'my-regular-logo'
-    const { getByAltText } = render(<HeaderLogo link='https://som.el.ink/tomy/home/page' buildConfig={config} />)
+    config.icons.appLogo = '/my-regular-logo'
+    const { getByAltText } = render(<HeaderLogo link='https://som.el.ink/tomy/home/page' />)
```
svenseeberg, line 32: I'd recommend to use the domain example.com to prevent any "accidents", i.e. dataleaks to not trusted domain owners.

svenseeberg (approved):
LGTM

PR #191, opened 2021-03-08
IGAPP-514: Show Time for Local News Details

Taggotty:


klinzo (approved):
Shouldn't we show the time in the list as well? How did you handle this for native?

Taggotty:
> Shouldn't we show the time in the list as well? How did you handle this for native?

So in native I only added the time for the news detail in native. 
Imo this is fine as most of the time the time is not relevant when looking for news. What do you think?

klinzo:
> > Shouldn't we show the time in the list as well? How did you handle this for native?
> 
> So in native I only added the time for the news detail in native.
> Imo this is fine as most of the time the time is not relevant when looking for news. What do you think?

Probably you are right. You could ask this in mattermost as well though for more opinions.

klinzo:
> I tested this locally on my IPhone, system language german and content language german, but the shown date string is: 'March 10, 2021' shouldn't this be german?

No, this is a known bug with moment... it crashes for e.g. german language. See https://issues.tuerantuer.org/browse/IGAPP-399 and https://issues.tuerantuer.org/browse/IGAPP-376

maxammann (approved):

PR #192, opened 2021-03-09
IGAPP-452: Create prettier config

SteffiStuffel:

I also already added the configs for Intellij because it is not possible to configure prettier to exactly match our current formatting rules, due to their [option philosophy](https://prettier.io/docs/en/option-philosophy.html). So we have to reformat all files with prettier. (e.g. it is not possible to have a space between function name and braces -> render () becomes render().) 

Now our linting rules only care about code-quality rules, and prettier takes care of formatting rules (overriding every es-lint rule about formatting). This avoids conflicts of prettier and eslint. Prettier runs autmatically on save in Intellij and vs code. 

If a prettier rule is violated it is not visible in the IDE, we can add this feature by adding eslint-plugin-prettier package, which added prettier as a linting rule, but this is not recommended and therefore i did not add it, but wanted to mention it. So it would make sense to add prettier checks to ci in the future, when all files are fixed.


.circleci/src/commands/lint.yml
```diff
@@ -6,5 +6,8 @@ steps:
   - run:
       name: Linting check
       command: yarn workspace << parameters.directory >> run lint:ci
+  # - run: TODO
```
klinzo, line 9: Should be enabled before merging together with a PR applying all new prettier rules.

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "semi": false,
+  "trailingComma": "none",
+  "singleQuote": true,
+  "arrowParens": "avoid",
+  "printWidth": 120,
+  "jsxBracketSameLine": false,
```
klinzo, line 7: I think this should be true.

README.md
```diff
@@ -57,6 +57,7 @@ We suggest **[IntelliJ IDEA Ultimate](https://www.jetbrains.com/idea/)** as IDE.
 If you are using a different IDE like Visual Studio Code ([./docs/vscode.md](./docs/vscode.md)) the steps might differ.
 
 * Import this project (VCS > Get from Version Control).
+* Install Prettier Plugin (inteillij should recogzie everything and format code with prettier, if not got to Settings -> Languages and Framworks | Javascript | Prettier and select the installed prettier package)
```
klinzo, line 60:
```suggestion
* Install Prettier Plugin (IntelliJ should recognize everything and format code with prettier, if not go to Settings > Languages and Frameworks > Javascript > Prettier and select the installed prettier package)
```

docs/conventions.md
```diff
@@ -8,6 +8,7 @@ We follow the [airbnb style](https://github.com/airbnb/javascript/tree/master/re
 
 We mainly follow the [standard style](https://standardjs.com/rules) for JavaScript, although there may be smaller deviations.
 Run the linter (`yarn lint`) to show problems. Our eslint config can be found [here](https://github.com/Integreat/eslint-config-integreat).
+We use prettier to format code. Run (`yarn workspace web (or native) prettier --check .`) to show formatting problems.
```
klinzo, line 11:
```suggestion
We use prettier to format code. Run (`yarn workspace <workspace> prettier --check .`) to show formatting problems.
```

.circleci/src/commands/lint.yml
SteffiStuffel: done in second PR

.prettierrc
SteffiStuffel: changed it.

Taggotty (approved):

klinzo (approved):

PR #195, opened 2021-03-10
IGAPP-488: Support links in local news detail view

zankerbe:
Links in local news are detected and clickable. However, this only seems to work when the link of the form https://google.de.
Links like www.google.de do not work and therefore can't be supported.


web/src/routes/news/containers/LocalNewsDetailsPage.js
```diff
@@ -35,11 +35,12 @@ export const LocalNewsDetailsPage = ({
     const error = new NotFoundError({ type: LOCAL_NEWS_TYPE, id, city, language })
     return <FailureSwitcher error={error} />
   }
+  const linkedContent=replaceLinks(localNewsElement.message)
```
klinzo, line 38: Take care of correct formatting.
```suggestion
  const linkedContent = replaceLinks(localNewsElement.message)
```
Either use auto formatting (possibly `STRG` + `ALT` + `L`?) or fix errors showing up in the IDE or after running `yarn lint` manually.

api-client/src/replaceLinks.js
```diff
@@ -3,11 +3,13 @@
 // https://stackoverflow.com/a/3809435
 const regex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/g
```
klinzo, line 4: To get rid of matching `www.example.com` change the regex to the following:
```suggestion
const regex = ((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|[-;:&=+$,\w]+@[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)
```
This will just match urls with protocol which is exactly what we want.

klinzo (changes requested):
Looking good!
Not sure why you are having such a huge change with yarn.lock. I think those changes should not be part of the PR.

klinzo:
Please make sure to write good commit messages prefixed with `IGAPP-<your-issue-number>`. Have a look at this guide https://github.com/erlang/otp/wiki/Writing-good-commit-messages mentioned in our convention docs.

SteffiStuffel (approved):
Works as expected, despite what @klinzo mentioned about www-starting links

klinzo:
> Works as expected, despite what @klinzo mentioned about www-starting links

@SteffiStuffel not for me. Did you try with e.g. www.example.com in the content?

web/src/routes/news/containers/LocalNewsDetailsPage.js
zankerbe: Changed the regex, thx!

klinzo (approved):

maxammann (approved):

PR #196, opened 2021-03-10
IGAPP-449: Add support for simplified chinese

maxammann:


native/src/modules/i18n/components/I18nProvider.js
```diff
@@ -59,7 +59,9 @@ export default ({ children }: PropsType) => {
         debug: buildConfig().featureFlags.developerFriendly
       })
 
-      await setContentLanguage(i18nextInstance.language).catch(e => {
+      // A language mentioned in the supportedLanguages array of the config.js in the translations package
+      const matchedLanguage = i18nextInstance.languages[0]
+      await setContentLanguage(matchedLanguage).catch(e => {
```
klinzo, line 64: I don't understand this change. Could you explain why it was necessary?

translations/translations.json
```diff
@@ -1182,9 +1252,7 @@
       "thanksMessage": "Vielen Dank für Ihr Feedback. Damit helfen Sie uns, unsere Inhalte und die {{appName}}-App zu verbessern. Wir würden uns über weiteres Feedback freuen.",
       "close": "Schließen",
       "negativeRating": "Diese Seite ist nicht hilfreich",
-      "positiveRating": "Diese Seite ist hilfreich",
-      "contactMailAddress": "E-Mail für Rückfragen",
-      "optionalInfo": "optional"
```
klinzo, line 1187: This should probably not get removed.

web/src/modules/i18n/BrowserLanguageDetector.js
```diff
@@ -51,14 +51,7 @@ export default {
...
 
-    const iso639Codes = bcp47Tags.map(bcp47Tag => {
-      if (bcp47Tag.includes('-')) {
-        return bcp47Tag.split('-')[0]
-      }
-      return bcp47Tag
-    })
-
-    return iso639Codes.length > 0 ? iso639Codes : undefined // Returning array: new i18next v19.5.0
+    return bcp47Tags.length > 0 ? bcp47Tags : undefined // Returning array: new i18next v19.5.0
```
klinzo, line 54: I don't think this comment is necessary in the code.

web/src/modules/i18n/containers/I18nProvider.js
```diff
@@ -48,8 +48,10 @@ export default ({ children }: PropsType) => {
       setI18nextInstance(i18nextInstance)
 
       // Apply ui language as language
-      i18nextInstance.on('languageChanged', uiLanguage => {
-        setLanguage(uiLanguage)
+      i18nextInstance.on('languageChanged', () => {
+        // A language mentioned in the supportedLanguages array of the config.js in the translations package
+        const matchedLanguage = i18nextInstance.languages[0]
```
klinzo, line 53: Same as above, why was this change necessary?

klinzo (commented):
1) Is it correct that the font included for web is named `latin chinese`?
2) Is there any instance with chinese so we could test this?
3) Does it work for native if no fonts are included?

native/src/modules/i18n/components/I18nProvider.js
maxammann: So the thing is chinese is our first language using a language tag instead of a code. That means we do not want to throw away everything except the code when detecting languages from browsers for example.

This means it is now possible to set de-DE as language. I use react-i18next to match de-DE automatically to a language we support (based on the config.js). This allows us to match 'zh-hans' to 'zh-CN' for example.

maxammann:
>     1. Is it correct that the font included for web is named `latin chinese`?

Yes, this includes latin and chinese chars. Size is not notably different

>     2. Is there any instance with chinese so we could test this?

Yes testumgebung on cms-test

>     3. Does it work for native if no fonts are included?

Not yet tested :)

build-configs/aschaffenburg/theme/fonts.js
```diff
@@ -4,14 +4,17 @@ import type { FontsType } from '../../common/theme/fonts'
...
-  decorativeFontBold: 'VarelaRound-Regular',
-  decorativeFontRegular: 'VarelaRound-Regular',
-  decorativeFontFamily: '\'Varela Round\', \'Lateef\', sans-serif',
-  contentFontRegular: 'OpenSans-Regular',
-  contentFontBold: 'OpenSans-Bold',
-  contentFontFamily: '\'Open Sans\', \'Lateef\', sans-serif',
-  arabicWebviewFontFamilies: '\'OpenSans\', \'Lateef\', sans-serif',
-  webviewFontFamilies: '\'OpenSans\', sans-serif',
+  native: {
+    decorativeFontBold: 'VarelaRound-Regular',
```
klinzo, line 8: Does this work on iOS for you? App crashing for me here.

klinzo:
Tested on firefox, looking good :earth_asia:

build-configs/aschaffenburg/theme/fonts.js
maxammann: not part of this task :P

maxammann:
This is ready for review. Tested on iOS, Android and Firefox. Tested with chinese as system language on iOS and Android.

Make sure to clear your local storage when testing with a browser. That triggers the language detection.

klinzo (approved):

SteffiStuffel (approved):
Tested this on android emulator.

When opening the language selection, option for chinese is always displayed as "Chinesisch", in german, also when changing system language.

PR #199, opened 2021-03-11
IGAPP-93: Deep linking

klinzo:


klinzo:
Deep linking :tada: 

Please review before I write tests. Some things may not be solved optimally.
You can test it either by adding `uri-scheme` library and executing commands as specified here https://reactnavigation.org/docs/deep-linking#test-deep-linking-on-android or by clicking on corresponding links/opening them in a browser.

Things worth mentioning:
* Previously city selections are not handled, the city/language of the deep link is always permanently selected. To change this, the user has to navigate back to the city selection and select the previous city. Not really sure but I'd say that is fine as it occurs probably more often that a user wants to select the city when opening a deep linking than the opposite.
* Universal links (https:// deep links) will probably not work on iOS as so called `apple-app-site-association` files are needed, see https://issues.tuerantuer.org/browse/IGAPP-520

native/src/modules/app/components/Navigator.js
```diff
@@ -143,25 +144,15 @@ const Navigator = (props: PropsType) => {
...
-      // https://github.com/react-navigation/react-navigation/issues/7685
-      if (
-        routeName === DASHBOARD_ROUTE &&
-        previousRouteName.current === LANDING_ROUTE &&
-        routeKey &&
-        cityCode &&
-        languageCode
-      ) {
-        fetchCategory(cityCode, languageCode, routeKey, true)
-      }
```
klinzo, line 161: Solved by using `navigation.reset` which allows us to open a route with a specific key.

native/ios/Integreat/Integreat.entitlements
```diff
@@ -4,5 +4,9 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>com.apple.developer.associated-domains</key>
```
maxammann, line 7: This will need a change in the profiles of iOS. Did you try to create a production build?

native/src/modules/app/components/App.js
```diff
@@ -14,8 +14,22 @@ import NavigatorContainer from '../containers/NavigatorContainer'
...
+  // $FlowFixMe redirect is part of available routes
+  config: {
+    screens: {
+      [REDIRECT_ROUTE]: '*'
+    }
+  },
+  getStateFromPath: path => {
+    return { index: 0, routes: [{ name: REDIRECT_ROUTE, params: { url: `https://${path}` } }] }
+  }
+}
```
maxammann, line 32: I think this should be part of the NavigationContainer

native/src/modules/app/containers/RedirectContainer.js
```diff
@@ -0,0 +1,37 @@
...
+  const { language } = i18n
+  const { url } = route.params
+
+  useEffect(() => {
+    // If actions are dispatched/navigate is called to early it fails and nothing happens
+    // Therefore we wait for a short time period
+    // https://github.com/react-navigation/react-navigation/issues/8537
+    const timeout = setTimeout(() => {
+      navigateToDeepLink(dispatch, navigation, url, language)
+    }, TIMEOUT)
```
maxammann, line 30: Is it possible to do this in a loop and check whether navigation is ready?

A timeout with 50ms will certainly fail on older devices.

native/src/modules/navigation/navigateToCategory.js
```diff
@@ -23,13 +24,23 @@ const navigateToCategory = <T: RoutesType>({
   languageCode: string,
   cityContentPath: string,
   key?: string,
-  forceRefresh?: boolean
+  forceRefresh?: boolean,
+  reset?: boolean
```
maxammann, line 28:
```suggestion
  resetNavigation?: boolean
```

native/src/modules/navigation/navigateToDeepLink.js
```diff
@@ -0,0 +1,66 @@
...
+      // Dashboard and landing route were already handled with reset above
+      if (routeInformation && ![LANDING_ROUTE, DASHBOARD_ROUTE].includes(routeInformation.route)) {
+        createNavigate(dispatch, navigation)(routeInformation, undefined, false)
+      }
+    } else {
+      navigation.replace(LANDING_ROUTE)
+    }
+
+    if (!routeInformation) {
+      // TODO show snackbar route not found
```
maxammann, line 61: Can you link here a task where this should be done?

maxammann (approved):

maxammann:
>     * Previously city selections are not handled, the city/language of the deep link is always permanently selected. To change this, the user has to navigate back to the city selection and select the previous city. Not really sure but I'd say that is fine as it occurs probably more often that a user wants to select the city when opening a deep linking than the opposite.

I dont really agree here. If you say that integreat should handle integreat.app links then you open ALL links found in google for example.

That means you google something and click a link and suddenly you change your city.

native/ios/Integreat/Integreat.entitlements
klinzo: No, didn't know about that. What do I have to do?

native/src/modules/app/components/App.js
klinzo: That is not possible. NavigationContainer is a component from react-navigation.

klinzo:
> I dont really agree here. If you say that integreat should handle integreat.app links then you open ALL links found in google for example.
> 
> That means you google something and click a link and suddenly you change your city. Can't we use the peek api here?
Thought about it a little more and I guess you are right. Also finally a use case for peeking :tada:

maxammann (approved):

native/src/modules/app/components/Navigator.js
```diff
@@ -110,7 +111,7 @@ const Navigator = (props: PropsType) => {
         throw Error('The contentLanguage has not been set correctly by I18nProvider!')
       }
 
-      if (!buildConfig().featureFlags.introSlides) {
+      if (!buildConfig().featureFlags.introSlides && !introShown) {
         await appSettings.setIntroShown()
         await appSettings.setSettings({
           errorTracking: false,
```
klinzo, line 117: Should we set this to true? I.e. should we enable sentry error reports per default for Aschaffenburg? I am inclined to say yes as this is apparently not a data privacy issue.

SteffiStuffel: makes sense to me to set it to true.

SteffiStuffel (approved):
Tested this on Android:

If uninstalling Integreat and reinstalling it, clicking through the intro slides and accepting location, etc. I get an warning:
Bildschirmfoto zu 2021-03-15 13-13-49
It is not possible to dismiss the warning and go on. Have to restart the app.

Opening of links in aschaffenburg and integreat works as expected.

native/ios/Integreat/Info.plist
```diff
@@ -70,5 +70,14 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+    <key>CFBundleURLTypes</key>
+    <array>
+        <dict>
+            <key>CFBundleURLSchemes</key>
+            <array>
+                <string>integreat</string>
```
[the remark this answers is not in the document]
klinzo: See discussion here: https://chat.tuerantuer.org/digitalfabrik/pl/e1xqjg7sdpn6i89i3rxnrem8hr
The protocol will be `integreat` for malte and aschaffenburg as well. So e.g. the links integreat://malteapp.de/ and integreat://aschaffenburg.app should be opened in the corresponding apps.

I don't know but this section here just allows the "integreat" protocol, "universal urls" aka https/http links are specified in the app entitlements.

klinzo:
@SteffiStuffel good catch, thanks! Could you check whether this still happens?

native/ios/Integreat/Info.plist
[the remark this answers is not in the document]
SteffiStuffel: Thanks for explanation. Makes more sense now.

SteffiStuffel:
> @SteffiStuffel good catch, thanks! Could you check whether this still happens?

Nope. Is fixed now.

SteffiStuffel:
Maybe it would be nice, if we add an information on how to test deep linking in readme?

maxammann:
I updated the branch and retriggered a new build. iOS should work now

PR #201, opened 2021-03-14
IGAPP-507: Specific news should be sharable

klinzo:


SteffiStuffel (approved):
Tested. Works as expected.

Taggotty (approved):

native/src/modules/layout/__tests__/HeaderContainer.spec.js
```diff
@@ -206,6 +206,32 @@ describe('HeaderContainer', () => {
     assertProps(result, { shareUrl: expectedShareUrl })
   })
 
+  it('shareUrl should be set correctly for local news details route', () => {
+    jest.doMock('../components/Header', () => Header)
```
Taggotty, line 210: Is the header mock at the top of the file even used anymore?

PR #202, opened 2021-03-14
IGAPP-522: Show errors for disabled routes

klinzo:

@maxammann you were right that the loading behavior can be reused completely, you can just return functions from hooks as well :clown_face: 

Opening sprungbrett routes via links in content or deep links is now possible as well.


native/src/routes/news/components/News.js
```diff
@@ -41,63 +41,54 @@ export type PropsType = {|
...
+          navigateToNews={navigateToNews}
           t={t}
         />
       )
     },
-    [selectedNewsType, navigateToNews, theme, t]
+    [selectedNewsType, theme, t, navigateTo]
   )
 
+  if (selectedNewsType === LOCAL_NEWS_TYPE ? !cityModel.pushNotificationsEnabled : !cityModel.tunewsEnabled) {
```
Taggotty, line 78: for local news: why should ther be a NotFounderror if the pushNotifications aren't enabled?

native/src/routes/news/containers/NewsContainer.js
```diff
@@ -39,15 +39,16 @@ type ContainerPropsType =
...
+      page: null
     |}
   | {|
       ...NavigationPropsType,
       ...DispatchPropsType,
       status: 'ready',
       news: NewsModelsType,
       hasMoreNews: ?boolean,
-      page: ?number,
+      page: number | null,
```
Taggotty, line 51: why this change?

native/src/routes/news/containers/NewsContainer.js
```diff
@@ -190,7 +192,7 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
       navigation,
       route: ownProps.route,
       hasMoreNews: route.hasMoreNews || null,
-      page: route.page || null,
+      page: route.page !== undefined ? route.page : null,
```
Taggotty, line 195: why this change? Is it possible thate `route.page` is an empty string

native/src/routes/offers/containers/OffersContainer.js
```diff
@@ -94,17 +84,20 @@ const OffersContainer = ({ theme, t, navigation, route }: OffersPropsType) => {
...
+        <FailureContainer code={fromError(error)} tryAgain={refresh} />
       </LayoutedScrollView>
     )
   }
 
   return (
-    <LayoutedScrollView refreshControl={<RefreshControl onRefresh={loadOffers} refreshing={loading} />}>
-      {offers && (
+    <LayoutedScrollView refreshControl={<RefreshControl onRefresh={refresh} refreshing={loading} />}>
+      {cities && offers && (
```
Taggotty, line 100: So shouldn't there be a `NotFoundError` if cities are not defined?

Taggotty (approved):

native/src/routes/news/components/News.js
klinzo: This is just shitty cms naming. This is whether the city has enabled local news or not. It has nothing (at least not directly) to do with actual push notifications.

native/src/routes/news/containers/NewsContainer.js
klinzo: Page is either a number or undefined. It is undefined if currently a fetch is in progress. This is to avoid sketchy 0 check, i.e. evaluation to null if page is 0.

native/src/routes/news/containers/NewsContainer.js
klinzo: This is to avoid sketchy 0 check, i.e. evaluation to null if page is 0.

native/src/routes/offers/containers/OffersContainer.js
klinzo: Hmmm normally it isn't possible that cities are undefined here as the fetch failed earlier already. I'd say if there are no cities it is not a NotFoundError.

maxammann (approved):

native/src/routes/offers/containers/OffersContainer.js
klinzo: Hmm yeah good question. If there are no cities it is not possible to check whether the feature is enabled. But I guess I'll just remove it as this edge case should not happen at all anyway.

PR #204, opened 2021-03-16
IGAPP-425: Linebreaks in navigation items on small devices

SteffiStuffel:


klinzo (approved):
On which devices did you test that there are no linebreaks anymore?

SteffiStuffel:
> On which devices did you test that there are no linebreaks anymore?

Only device i know having this bug is iPhons 6s, so i tested this on an Android Emulator with same resolution / size

Taggotty (approved):

PR #205, opened 2021-03-16
IGAPP-527: Persist tracking code

SteffiStuffel:

### DIFF
--- a/native/src/modules/settings/AppSettings.js
+++ b/native/src/modules/settings/AppSettings.js
@@ -12,7 +12,9 @@ export type SettingsType = {|
   errorTracking: boolean | null,
   allowPushNotifications: boolean | null,
   proposeNearbyCities: boolean | null,
-  apiUrlOverride: string | null
+  apiUrlOverride: string | null,
+  jpalTrackingCode: string | null,
+  jpalTrackingEnabled: boolean | null
 |}
 
 export const defaultSettings: SettingsType = {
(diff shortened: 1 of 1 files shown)


native/src/modules/settings/AppSettings.js
```diff
@@ -23,7 +25,9 @@ export const defaultSettings: SettingsType = {
   errorTracking: null,
   allowPushNotifications: null,
   proposeNearbyCities: null,
-  apiUrlOverride: null
+  apiUrlOverride: null,
```
klinzo, line 28: For all other settings we have specific getters and setters, but probably okay to just be able to set this with setSettings

klinzo (approved):

native/src/modules/settings/AppSettings.js
```diff
@@ -23,7 +25,9 @@ export const defaultSettings: SettingsType = {
...
+  jpalTrackingCode: null,
+  jpalTrackingEnabled: false
```
klinzo, line 30: For all other settings we have specific getters and setters, but probably okay to just be able to set this with setSettings

klinzo: Maybe it would also make sense to keep everything as generic as possible, i.e. just use `trackingCode` and `trackingEnabled`. @maxammann what do you think?

Taggotty (approved):

SteffiStuffel:
Will close this, as these changes are also done in IGAPP-524

PR #214, opened 2021-03-20
IGAPP-129: Use new netinfo api to only force a content refresh if there is the internet is reachable

klinzo:


native/src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -13,14 +13,100 @@ import loadLanguages from './loadLanguages'
...
 import loadPois from './loadPois'
-import type { SettingsType } from '../../settings/AppSettings'
+import { CategoriesMapModel, EventModel } from 'api-client'
+
+/**
+ * Persists the newCity as selected city and subscribes to the corresponding push notifications.
+ * @param newCity
+ * @param newLanguage
+ */
+function* selectCity(newCity: string, newLanguage: string): Saga<void> {
```
klinzo, line 31: Extracted some functionality to make it better readable and understandable

native/src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -40,20 +126,15 @@ export default function* loadCityContent(
...
   }
 
   const lastUpdate: Moment | null = yield call(dataContainer.getLastUpdate, newCity, newLanguage)
-  console.debug('Last city content update on ', lastUpdate ? lastUpdate.toISOString() : 'never')
+  console.debug('Last previous city content update: ', lastUpdate ? lastUpdate.toISOString() : 'never')
 
-  const shouldUpdate = criterion.shouldUpdate(lastUpdate)
+  const netInfo = yield call(NetInfo.fetch)
+  // Only try to update if there is an internet connection
+  const shouldUpdate = criterion.shouldUpdate(lastUpdate) && netInfo.isInternetReachable
```
klinzo, line 137: Not sure if this makes sense. That way a force refresh does not fail if there is no internet connection won't fail and no warning will be shown.

maxammann: This is indeed problematic. Can we check here whether it was a force refresh and then yield an error action?

native/src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -68,40 +149,9 @@ export default function* loadCityContent(
...
-          newCity,
-          dataContainer,
-          shouldUpdate
-        ) /* The languages for a city get updated if a any
-                                                                       language of the city is:
-                                                                          * older than MAX_CONTENT_AGE,
-                                                                          * has no "lastUpdate" or
-                                                                          * an update is forced
-                                                                        This means the loading of languages depends on
-                                                                        language AND the city */
```
maxammann, line 83: This comment should stay I think

maxammann (approved):

native/src/modules/endpoint/sagas/loadCityContent.js
klinzo: I think it is self explanatory and the same for all data fetched from the endpoints with an exception from languages. But I can also readd it.

native/src/modules/endpoint/sagas/loadCityContent.js
klinzo: I think I'll just revert the change. An error action is already yielded if there is no internet connection.

native/src/modules/app/components/App.js
```diff
@@ -16,12 +16,26 @@ import StaticServerProvider from '../../static-server/containers/StaticServerPro
...
+  config: {
+    screens: {
+      [REDIRECT_ROUTE]: '*'
+    }
+  },
+  getStateFromPath: path => {
+    return { index: 0, routes: [{ name: REDIRECT_ROUTE, params: { url: `https://${path}` } }] }
+  }
+}
+
```
maxammann: Should these changes be in this PR? a little bit confused because this is only about netinfo isnt it?

maxammann (approved):

native/src/modules/app/components/App.js
klinzo: I think you are viewing a single commit/outdated version :thinking: I just see very few changes in App.js.

SteffiStuffel (approved):
Tested on android emulator. Works as expected, when only mobile data is available a previously opend city data is not downloaded, when the city was not visited before data is downloaded.

In the ticket is metioned to: 'Also check use the `.internetReachable` api to check for connectivity in the sagas.' Think this is not done, but do not know if necessary.

klinzo:
> Tested on android emulator. Works as expected, when only mobile data is available a previously opend city data is not downloaded, when the city was not visited before data is downloaded.
> 
> In the ticket is metioned to: 'Also check use the `.internetReachable` api to check for connectivity in the sagas.' Think this is not done, but do not know if necessary.

@SteffiStuffel you are definitely right, I did not follow the issue completely since no internet reachability is already handled correctly with our current errors and changing it would not have improved anything and required lots of code changes. Good observation though!

PR #216, opened 2021-03-22
IGAPP-515: Disable Sentry Sourcemap releases for webnext

maxammann:
# This PR should serve as a discussion

The original issue is because we are not bumping versions when we are doing development releases to webnext.
That means we are uploading different sourcemaps to the same release in sentry.
This is not really a big issue as sentry will figure out on which sourcemap the issue happened and use the correct one.

I think we should close the issue and NOT merge this. In fact this is only an issue for aschaffenburg and malte as we do not have different build configs for releases to webnext.malte.app and webnext.aschaffenburg.app.

The proper fix to this issue is probably to include the deployment information for web in the build config. That way we would have different build configs for integreat, malte and aschaffenburg for webnext. Like malte-test-cms and aschaffenburg-test-cms. What do you think about this proper fix?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -679,15 +679,6 @@ jobs:
                     - run:
                         command: sftp -b - -v -o StrictHostKeyChecking=no web@web.integreat-app.de:/webnext.aschaffenburg.app \<<< "put -r $HOME/attached_workspace/dist/aschaffenburg/*"
                         name: Aschaffenburg development delivery
-                    - run:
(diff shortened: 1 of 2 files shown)


klinzo:
If it is not a problem for sentry we can just keep it as it is. I don't think a malte-test-cms and aschaffenburg-test-cms build config really makes sense. There is not test cms for those yet and I think we don't need one either. So we could create separate build configs with different feature flags enabled and so on, but imo that just increases complexity without any benefit. So I'd just keep it as is I guess.

SteffiStuffel:
I agree with Steffen, do not really see a problem with leaving it as it is.

PR #217, opened 2021-03-22
IGAPP-509: Tooltips are opened on click

maxammann:

**Opening tooltips on click seems fine for me, so I just fixed the weird clipping**

### DIFF
--- a/web/src/modules/common/components/Tooltip.js
+++ b/web/src/modules/common/components/Tooltip.js
@@ -183,10 +183,11 @@ type PropsType = {|
   text: ?string,
   flow: FlowType,
   mediumViewportFlow?: FlowType,
-  smallViewportFlow?: FlowType
+  smallViewportFlow?: FlowType,
+  className?: string
 |}
 
-export default ({ children, text, flow, mediumViewportFlow, smallViewportFlow }: PropsType) => {
+export default ({ children, text, flow, mediumViewportFlow, smallViewportFlow, className }: PropsType) => {
   const platform = useContext(PlatformContext)
 
   if (!text) {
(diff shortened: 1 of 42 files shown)


maxammann:
I had to pass through direction through a lot of components... Changes which are relevant are only in Toolbar and CSS rules.

web/src/modules/common/components/Tooltip.js
```diff
@@ -1,41 +1,28 @@
...
-      return 'inset-inline-start'
-    case 'border-left-width':
-      return 'border-inline-start-width'
-    case 'border-right-width':
-      return 'border-inline-end-width'
-    case 'border-left-color':
-      return 'border-inline-start-color'
-    case 'border-right-color':
-      return 'border-inline-end-color'
+// The previous implementation of this function used css logical properties. Therefore the name
```
klinzo, line 9: Otherwise it looks like the comment is incomplete.
```suggestion
// The previous implementation of this function used css logical properties. Therefore the name.
```

klinzo (commented):
Can't we just access the state directly in `Tooltip`? E.g. with the `useSelector` hook? https://react-redux.js.org/api/hooks#useselector

maxammann:
I think I'll try to find another solution using JS. Maybe I can figure out whether there is more free space on the left or the right in the DOM.

maxammann:
> Can't we just access the state directly in `Tooltip`? E.g. with the `useSelector` hook? https://react-redux.js.org/api/hooks#useselector

That would violate our rules of components vs containers. This is indeed still a good rule. You would not expect to create a whole redux store to render a Tooltip.

maxammann:
replaced by https://github.com/Integreat/integreat-app/pull/256

klinzo:
@maxammann can we delete the branch?

PR #218, opened 2021-03-22
IGAPP-503: Jpal signal endpoint

klinzo:


native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -47,12 +49,8 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
   const simpleRouteShareUrl =
     typeof cityCode === 'string' && simpleRoutes.includes(routeName)
-      ? cityContentUrl({
-          cityCode,
-          languageCode: state.contentLanguage,
-          route: routeName,
-          path: null
-        })
+      // $FlowFixMe Flow does not understand life and just hates the whole world :/
+      ? urlFromRouteInformation({ cityCode, languageCode: state.contentLanguage, route: routeName })
```
Taggotty, line 53: to get rid of the flow error you could write: 
```suggestion
      ? urlFromRouteInformation({ cityCode, languageCode: state.contentLanguage, route: (routeName: $Values<typeof simpleRoutes> })
```

native/src/modules/layout/containers/HeaderContainer.js
```diff
@@ -97,17 +95,16 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
-        languageCode: route.language,
-        route: routeName,
-        path: newsPath
-      })
+  const { routeType, language, city } = route
+
+  const routeInformation: RouteInformationType =
+    routeType === NEWS_ROUTE
+      // $FlowFixMe Flow does not get anything and hates the whole world :/ I am already checking the routeType!
+      ? { route: NEWS_ROUTE, languageCode: language, cityCode: city, newsType: route.type, newsId: route.newsId }
```
Taggotty, line 103: to fix the flow error check the RouteInformationType. `newsId` does not seem to be nullable, but is for the route mapping type.

Taggotty: also to help flow write `route.routeType === NEWS_ROUTE`

Taggotty (approved):

native/src/modules/layout/containers/HeaderContainer.js
klinzo: Nice, thanks :+1:

native/src/modules/layout/containers/HeaderContainer.js
klinzo: Another good suggestion :D thanks :+1: The last fixme is only solvable with a lot more complicated code which is not worth it imo as cityContentPath has to be defined for category/dashboard routes but is optional for events/pois.

klinzo: Or do you have any more tips for me @Taggotty :P

native/src/modules/endpoint/sendTrackingSignal.js
```diff
@@ -0,0 +1,46 @@
...
+          appSettings: {
+            allowPushNotifications,
+            proposeNearbyCities,
+            errorTracking
+          }
+        }
+      })
+    }
+  } catch (e) {
+    console.error(e)
```
maxammann, line 42: Should we maybe send this to sentry?

maxammann: Also not sure whether console.error crashes the app in release mode.

api-client/src/routes/RouteInformationTypes.js
```diff
@@ -63,4 +63,5 @@ export type RouteInformationType =
   | NewsRouteInformationType
   | SimpleCityContentFeatureType
   | EventsPoisRouteInformationType
-  | null
+
+export type NullableRouteInformationType = RouteInformationType | null
```
maxammann, line 67: Should we name this RouteInformationType and the above NonNullRouteInformationType?

maxammann (approved):

native/src/__mocks__/@react-native-community/async-storage.js
```diff
@@ -1,5 +0,0 @@
-// @flow
```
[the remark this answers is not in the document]
klinzo: Because I just added it directly to the jest setup as there is never a case where it makes sense to have the unmocked async storage in tests. It was also just a reference to the actual mock.

native/src/modules/endpoint/sendTrackingSignal.js
klinzo: How can I do that? And no, it does not crash the app in release mode afaik but will test again. We have console.errors on quite a few places actually.

PR #221, opened 2021-03-23
IGAPP-526: Implement jpal settings screen

SteffiStuffel:


native/src/routes/jpal-tracking/components/JpalTracking.js
```diff
@@ -0,0 +1,119 @@
...
+  useEffect(() => {
+    const loadSettings = async () => {
+      try {
+        const loadedSettings = await appSettings.loadSettings()
+        setSettingsLoaded(true)
+        setSettingsUpdated(false)
+        setSettings(loadedSettings)
+        setDisplayedTrackingCode(loadedSettings.jpalTrackingCode)
+      } catch (e) {
+        console.error('Failed to load settings.')
```
maxammann, line 69: Use an error state variable here to show the user that an error happened.

native/src/routes/jpal-tracking/containers/JpalTrackingContainer.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import type { TFunction } from 'react-i18next'
+import { withTranslation } from 'react-i18next'
+import withTheme from '../../../modules/theme/hocs/withTheme'
+import JpalTracking, { type PropsType as JpalTrackingPropsType } from '../components/JpalTracking'
+
+export default withTheme<$Diff<JpalTrackingPropsType, {| t: TFunction |}>>(
+  withTranslation<JpalTrackingPropsType>('settings')(JpalTracking)
+)
```
maxammann, line 10: I'm white sure you could use here hooks of react-i18n and our theme.
That would make the tracking screen a container, but I thinkg this is fine.

native/src/routes/settings/components/Settings.js
```diff
@@ -78,6 +83,10 @@ export default class Settings extends React.Component<PropsType, StateType> {
     }
   }
 
+  componentWillUnmount() {
+    this.unsubscribeNavigationListener()
+  }
```
maxammann, line 88: Why is this needed? why do we need to load settings when the comonent unmounts?

maxammann (approved):

version.json
```diff
@@ -1 +1,4 @@
-{"versionName":"2021.3.4","versionCode":100042155}
\ No newline at end of file
+{
+  "versionName": "2021.3.4",
+  "versionCode": 100042155
+}
```
klinzo, line 4: Revert this change please, probably because of prettier, right? I added it to ignored files somewhere already, not sure in which PR though.

translations/translations.json
```diff
@@ -2994,6 +2994,8 @@
       "version": "Version: {{version}}",
       "openSourceLicenses": "Open-Source-Lizenzen",
       "tracking": "J-PAL Evaluation",
+      "trackingAllowed": "Tracking zulassen",
```
klinzo, line 2997:
```suggestion
      "allowTracking": "Tracking zulassen",
```

native/src/routes/jpal-tracking/components/JpalTracking.js
```diff
@@ -0,0 +1,119 @@
...
+
+export type PropsType = {|
+  theme: ThemeType,
+  t: TFunction,
+  route: RoutePropType<JpalTrackingRouteType>,
+  navigation: NavigationPropType<JpalTrackingRouteType>
+|}
+
+const JpalTracking = (props: PropsType, state: StateType) => {
+  const [appSettings] = useState(new AppSettings())
```
klinzo, line 47: This seems weird, I think keeping the settings in the state should be enough. Although I think you should type `useState<type>`.

native/src/routes/jpal-tracking/components/JpalTracking.js
```diff
@@ -0,0 +1,119 @@
...
+        setSettings(loadedSettings)
+        setDisplayedTrackingCode(loadedSettings.jpalTrackingCode)
+      } catch (e) {
+        console.error('Failed to load settings.')
+      }
+    }
+    loadSettings()
+  }, [appSettings, settingsUpdated])
+
+  const onPress = () => {
```
klinzo, line 75: I think a more speaking name like e.g. `toggleTrackingEnabled` would make sense if this is what it does.

native/src/routes/jpal-tracking/components/JpalTracking.js
```diff
@@ -0,0 +1,119 @@
...
+        setSettingsLoaded(true)
+        setSettingsUpdated(false)
+        setSettings(loadedSettings)
+        setDisplayedTrackingCode(loadedSettings.jpalTrackingCode)
+      } catch (e) {
+        console.error('Failed to load settings.')
+      }
+    }
+    loadSettings()
+  }, [appSettings, settingsUpdated])
```
klinzo, line 73: I don't think we have to reload the settings every time we change something. I think it is enough to just load the settings once initially and keep only relevant settings in a local state. Then you could just catch possible errors during setting of the tracking code or tracking enabled to the global app settings and change the local state if everything succeeds and no error is thrown. Than you can safely assume that the local state always matches the global state.

native/src/routes/jpal-tracking/components/JpalTracking.js
klinzo: Possibly also just with a todo to add a snackbar at a later point of time.

native/src/routes/jpal-tracking/components/__tests__/JpalTracking.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  const route: RoutePropType<JpalTrackingRouteType> = {
+    key: 'route-id-0',
+    params: { trackingCode: 'abc' },
+    name: JPAL_TRACKING_ROUTE
+  }
+
+  it('Textfield should not be editable, when tracking is disabled', async () => {
+    const { findByTestId } = render(<JpalTracking theme={lightTheme} t={t} route={route} navigation={navigation} />)
+
+    AsyncStorage.setItem('jpalTrackingEnabled', true, async () => {
```
klinzo, line 31: I'd rather use our AppSettings instead of AsyncStorage directly. I also think it is weird to render the component before you set the setting initially. Maybe it also makes sense to assert on a few more things like that jpal tracking is enabled in the AppSettings if you toggled the switch.

native/src/routes/jpal-tracking/components/JpalTracking.js
```diff
@@ -0,0 +1,119 @@
...
+  const { jpalTrackingEnabled } = settings
+
+  return (
+    <View style={{ padding: 40 }}>
+      <Caption title={t('tracking')} theme={theme} />
+      <Text>{t('trackingDescription')}</Text>
+
+      <DescriptionContainer>
+        <ThemedText theme={props.theme}>{t('trackingAllowed')}</ThemedText>
+        <Switch
```
klinzo, line 99: Maybe it makes sense to switch on a click on the text as well? I.e. wrap the text in the switch.

klinzo (changes requested):

native/src/routes/settings/components/Settings.js
SteffiStuffel: In settings screen we see if tracking is enabled or disabled, when clicking on tracking-settings and navigate to jpal-settings-screen, we can update if tracking is enabled and when nativate.goBack to general settings, there is still the old value, because settings are only loaded in constructor. Therefore i add an listener for "on-focus" so component loads settings when opend (also on goBack()). We have to unsubscribe from this listener when unmounting.

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: done

translations/translations.json
SteffiStuffel: done

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: done

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: So clicking on the text field would be a consent for tracking? Not sure if this is a good idea.

klinzo: I think that is the default behavior in apps, clicking somewhere in the row always toggles settings. Maybe we need a dialog at a later time depending on what texts, terms and conditions jpal gives us.

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: done

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: if just putting it like:

const appSettings = new AppSettings() it get:

ESLint: The 'appSettings' object construction makes the dependencies of useEffect Hook (at line 73) change on every render. To fix this, wrap the initialization of 'appSettings' in its own useMemo() Hook.(react-hooks/exhaustive-deps)

i now changed it to useMemo()

native/src/routes/jpal-tracking/containers/JpalTrackingContainer.js
SteffiStuffel: changed it for translations, for theme we would need a ThemeProvider, we do not have yet, as Steffen poineted out.

native/src/routes/jpal-tracking/components/JpalTracking.js
SteffiStuffel: I think we talk about different things. 
But now i get what you mean: The whole row and not only the switch should be clickable?

native/src/routes/jpal-tracking/__tests__/JpalTracking.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  const route: RoutePropType<JpalTrackingRouteType> = {
+    key: 'route-id-0',
+    params: { trackingCode: 'abc' },
+    name: JPAL_TRACKING_ROUTE
+  }
+
+  // Error: Unable to find node on an unmounted component.
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('Textfield should not be editable, when tracking is disabled', async () => {
+    await act(async () => {
```
SteffiStuffel, line 35: Unfortunately I was not able to get this test running, need some help here.

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+    })
+  }
+
+  const setTrackingCode = (value: string) => {
+    setDisplayedTrackingCode(value)
+    appSettings.setJpalTrackingCode(value).catch(() => {
+      setError(true)
+      setSettingsUpdated(true)
+    })
+    props.navigation.setParams({ trackingCode: value })
```
klinzo, line 103: I think it is enough to just set the tracking code from the navigation params initially. No need to keep the tracking code in the navigation params consistent afterwards.

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+    loadSettings()
+  }, [appSettings, settingsUpdated])
+
+  const toggleTrackingEnabled = () => {
+    setDisplayedTrackingEnabled(!displayedTrackingEnabled)
+    appSettings.setJpalTrackingEnabled(!settings.jpalTrackingEnabled).catch(() => {
+      setError(true)
+      setSettingsUpdated(true)
```
klinzo, line 93: Did you try to get it working without this `settingsUpdated` thingy?

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+
+export type PropsType = {|
+  theme: ThemeType,
+  route: RoutePropType<JpalTrackingRouteType>,
+  navigation: NavigationPropType<JpalTrackingRouteType>
+|}
+
+const errorDisplayTime = 5000
+
+const JpalTracking = (props: PropsType, state: StateType) => {
```
klinzo, line 54:
```suggestion
const JpalTracking = (props: PropsType) => {
```

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+import styled from 'styled-components/native'
+import LayoutContainer from '../../modules/layout/containers/LayoutContainer'
+import withTheme from '../../modules/theme/hocs/withTheme'
+import Caption from '../../modules/common/components/Caption'
+
+const ThemedText = styled.Text`
+  display: flex;
+  text-align: left;
+  color: ${props => props.theme.colors.textColor};
+  font-family: ${props => props.theme.fonts.decorativeFontRegular};
```
klinzo, line 21: Otherwise the app crashes:
```suggestion
  font-family: ${props => props.theme.fonts.native.decorativeFontRegular};
```

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  if (!settingsLoaded) {
+    return <LayoutContainer />
+  }
+
+  const { theme } = props
+
+  return (
+    <View style={{ padding: 40 }}>
```
klinzo, line 113: Should be wrapped in a LayoutContainer to ensure the right background color and layout :)

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,141 @@
...
+  const [displayedTrackingCode, setDisplayedTrackingCode] = useState('')
+  const [displayedTrackingEnabled, setDisplayedTrackingEnabled] = useState(false)
+  const [error, setError] = useState(false)
+  const routeTrackingCode = props.route.params.trackingCode
+
+  const { t } = useTranslation(['settings', 'error'])
+
+  useEffect(() => {
+    if (routeTrackingCode) {
+      appSettings.setJpalTrackingCode(routeTrackingCode)
```
klinzo, line 68: Shouldn't we display the preset tracking code then? What happens if the route params contains a code different from one already saved in the settings?

native/src/routes/jpal-tracking/components/JpalTracking.js
klinzo: You could also just move it out of the component, i.e.:
```
const appSettings = new AppSettings()

const JpalTracking = (props: PropsType) => {
...
```
Then you also don't have to put it into the dependency of the hooks.

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: yes works now

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: It gets changed. Thought this would make sense. If one enables tracking and enters a code, and afterwards gets an letter from jpal an opens a deeplink with tracking code, i would expect it to update.

klinzo: It gets changed through loading it from the global app settings afterwards, right? So it is a little race condition, correct?

Yeah, probably you are right as there is no use case where a user clicks a jpal link with tracking code and does not want to change the tracking code. Other opinions @maxammann @Taggotty @svenseeberg? Do you think it is a problem to overwrite a previously set tracking code if a user clicks on a link containing a different tracking code?

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,138 @@
...
+        setError(true)
+      }
+    }
+    loadSettings()
+  }, [])
+
+  const toggleTrackingEnabled = () => {
+    setDisplayedTrackingEnabled(!displayedTrackingEnabled)
+    appSettings.setJpalTrackingEnabled(!settings.jpalTrackingEnabled).catch(() => {
+      setError(true)
```
klinzo, line 90: Should we reset the displayed/local settings to the value it was previously if an error happens here?

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,138 @@
...
+  navigation: NavigationPropType<JpalTrackingRouteType>
+|}
+
+const errorDisplayTime = 5000
+
+const appSettings = new AppSettings()
+
+const JpalTracking = (props: PropsType) => {
+  const [settings, setSettings] = useState(defaultSettings)
+  const [settingsLoaded, setSettingsLoaded] = useState(false)
```
klinzo, line 57: As mentioned in the coworking: I think those could be omitted :)

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,138 @@
...
+          />
+        </DescriptionContainer>
+
+        <ThemedText theme={props.theme}>{t('trackingCode')}</ThemedText>
+        <Input
+          value={displayedTrackingCode}
+          onChangeText={setTrackingCode}
+          theme={theme}
+          editable={displayedTrackingEnabled}
+          testID='input'></Input>
```
klinzo, line 132:
```suggestion
        <Input
          value={displayedTrackingCode}
          onChangeText={setTrackingCode}
          theme={theme}
          editable={displayedTrackingEnabled}
          testID='input' />
```

klinzo: Could I have a grey border if it is not editable please? :stuck_out_tongue:

klinzo (approved):

native/src/routes/jpal-tracking/JpalTracking.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }, [routeTrackingCode])
+
+  useEffect(() => {
+    const loadSettings = async () => {
+      try {
+        const loadedSettings = await appSettings.loadSettings()
+        setSettingsLoaded(true)
+        setSettings(loadedSettings)
+        setTimeout(() => setError(false), errorDisplayTime)
```
klinzo, line 75: Use effect should return a function removing all timeouts and pending promises

klinzo: Also, why do we have a timeout for the error just here?

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: removed the race condition.

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: done

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: We talked abour settingsUpdated, i removed it. But i think we should keep settingsLoaded, otherwise the page is rendered, when the date is not present yet and then (after a short time), when the data is present, the switch and the textfield get updated, what looks weird

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: It is there already :)

native/src/routes/jpal-tracking/JpalTracking.js
SteffiStuffel: cause it is the timeout to "unset" the error, so the error is displayed until the settings are reloaded correctly plus errorDisplayTime (3s). Think it does not make sens anywhere else.

maxammann:
> Do you think it is a problem to overwrite a previously set tracking code if a user clicks on a link containing a different tracking code?

I dont think that is a problem.

klinzo (approved):

PR #223, opened 2021-03-23
IGAPP-567: Dont do production releases for aschaffenburg

klinzo:

### DIFF
--- a/.circleci/src/workflows/scheduled_native_promotion.yml
+++ b/.circleci/src/workflows/scheduled_native_promotion.yml
@@ -18,12 +18,12 @@ jobs:
       context:
         - mattermost
         - tuerantuer-google-play
-  - promote_android:
-      name: promote_aschaffenburg_android
-      build_config_name: aschaffenburg
-      context:
-        - mattermost
-        - tuerantuer-google-play
+#  - promote_android: # TODO IGAPP 567 Uncomment
+#      name: promote_aschaffenburg_android
+#      build_config_name: aschaffenburg
+#      context:
+#        - mattermost
+#        - tuerantuer-google-play
 
   - promote_ios:
       name: promote_integreat_ios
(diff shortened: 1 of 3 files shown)


Taggotty (approved):

SteffiStuffel (approved):

.circleci/src/workflows/scheduled_native_promotion.yml
```diff
@@ -18,12 +18,12 @@ jobs:
       context:
         - mattermost
         - tuerantuer-google-play
-  - promote_android:
-      name: promote_aschaffenburg_android
-      build_config_name: aschaffenburg
-      context:
-        - mattermost
-        - tuerantuer-google-play
+#  - promote_android: # TODO IGAPP 567 Uncomment
```
maxammann, line 21: IGAPP-567 is the current task, I'm a bit confused :thinking:

maxammann (approved):
Can't this be done via our environment variables?

.circleci/src/workflows/scheduled_native_promotion.yml
klinzo: :D we have to disable it first before we can enable it again... I think "Arbeitsbeschaffungsmaßnahme" is the right term for this.

klinzo:
@maxammann it can, but not depending on production/development. So we can only not deliver at all or deliver all the time. This just adds an additional security that we don't accidentally release to produciton before we should

maxammann:
> @maxammann it can, but not depending on production/development. So we can only not deliver at all or deliver all the time. This just adds an additional security that we don't accidentally release to produciton before we should

Cant we just disable the promotion and continue to deliver? e.g. by setting SKIP_BUILD_promote_ios to aschaffenburg.

maxammann (approved):

klinzo:
Discussed this on mattermost. Disabling promotions does not cover manually triggered production releases and as this are very few changes the benefits of additional security of not releasing aschaffenburg to early is more important.

PR #225, opened 2021-03-24
IGAPP-370: City name in header is not centered vertically

klinzo:

### DIFF
--- a/native/src/modules/layout/components/Header.js
+++ b/native/src/modules/layout/components/Header.js
@@ -40,8 +40,6 @@ const Icon = styled.Image`
 const HeaderText: StyledComponent<{||}, ThemeType, *> = styled.Text`
   flex: 1;
   flex-direction: column;
-  text-align-vertical: center;
-  height: 50px;
   font-size: 20px;
   color: ${props => props.theme.colors.textColor};
   font-family: ${props => props.theme.fonts.native.decorativeFontBold};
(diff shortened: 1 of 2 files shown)


SteffiStuffel (approved):
Tested on Android emulator

Taggotty (approved):
Also tested on Android emulator

native/src/modules/layout/components/Header.js
```diff
@@ -40,8 +40,6 @@ const Icon = styled.Image`
 const HeaderText: StyledComponent<{||}, ThemeType, *> = styled.Text`
   flex: 1;
   flex-direction: column;
-  text-align-vertical: center;
```
maxammann, line 43: Shouldn't this actually center it vertically? I trust you if you say yes :D

maxammann (approved):

native/src/modules/layout/components/Header.js
klinzo: It did not, at least not on iOS apparently :D I just removed the height and let the parent container do the work

maxammann: But continues to work on android?

klinzo: That is the plan, yes. @Taggotty and @SteffiStuffel already tested but I will test again myself to make sure.

klinzo: Yes it does :)

PR #226, opened 2021-03-25
IGAPP-70: Implement test for internal linking

Taggotty:


web/src/modules/common/components/__tests__/RemoteContent.spec.js
```diff
@@ -1,8 +1,10 @@
 // @flow
 
 import React from 'react'
-import { shallow } from 'enzyme'
+import { mount, shallow } from 'enzyme'
```
klinzo, line 4: Could you solve this with the testing library or test renderer as well? Preferably the first?

klinzo (approved):

klinzo (approved):

native/android/.idea/misc.xml
```diff
@@ -4,7 +4,7 @@
   <component name="FrameworkDetectionExcludesConfiguration">
     <file type="web" url="file://$PROJECT_DIR$" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_13" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
SteffiStuffel, line 7: Is it intended that language level is 13, but project-jdk-name is still 8.

And why have you updated this?

SteffiStuffel (approved):

native/android/.idea/misc.xml
maxammann: This should either be JDK 8 or JDK 15. JDK 15 is the current version of java whereas jdk8 is the legacy version

maxammann (approved):

native/android/.idea/misc.xml
Taggotty: will revert. IntelliJ updated it by itself, as I did have another jdk on my system

PR #229, opened 2021-03-26
IGAPP-546: Remove require cicle in category model

zankerbe:


klinzo (approved):
:+1:

SteffiStuffel (changes requested):
I think it would make sense to use the isLeaf method also in Categories.js (line 137) (native). Or is there a reason, why you didn't use it there?

klinzo:
> I think it would make sense to use the isLeaf method also in Categories.js (line 137) (native). Or is there a reason, why you didn't use it there?

Probably makes sense to use it there, yes. This Issue wasn't about using the `isLeaf` function wherever possible but instead just to refactor previous usages, so I'd say it is @zankerbe 's decision if he wants to use it there as well.

maxammann (approved):

Taggotty (approved):

zankerbe:
> 
> 
> I think it would make sense to use the isLeaf method also in Categories.js (line 137) (native). Or is there a reason, why you didn't use it there?

Since Categories.js doesn't use an instance of CategoriesMapModel the isLeaf function cannot be directly used here. I'd prefer to leave it as it is.

SteffiStuffel (approved):

PR #234, opened 2021-03-30
IGAPP-563: Send signal if user shares a page

SteffiStuffel:


klinzo (approved):

Taggotty (approved):
If your motivated you could potentially also add a test for this :)

PR #237, opened 2021-03-30
IGAPP-553: Change ku to ckb

Taggotty:


translations/src/config.js
```diff
@@ -33,8 +33,8 @@ class Config {
     tr: { rtl: false },
     pl: { rtl: false },
     ti: { rtl: false },
-    ku: {
-      rtl: false,
+    ckb: {
+      rtl: true,
```
Taggotty, line 37: Sorani, which uses the persian alphabet is in fact a RTL language. However the  translations we have currently don't have the translations in persian alphabet. So do the translations also have to change then?

Taggotty:
In our [wiki](https://wiki.tuerantuer.org/integreat-languages) kurmandschi (kmr) is marked as ltr, but [wikipedia](https://de.wikipedia.org/wiki/Hilfe:Asiatische_Schriften) says differently. Therefore, I didn't change the tag (which was rtl to begin with)

klinzo:
> In our [wiki](https://wiki.tuerantuer.org/integreat-languages) kurmandschi (kmr) is marked as ltr, but [wikipedia](https://de.wikipedia.org/wiki/Hilfe:Asiatische_Schriften) says differently. Therefore, I didn't change the tag (which was rtl to begin with)

Actually I think our wiki is correct and although it is possible to write kurmandschi with an arabic alphabet from right to left, the standard seems to be a modified latin alphabet:
https://de.wikipedia.org/wiki/Kurdische_Schriftsysteme
https://de.wikipedia.org/wiki/Kurmandschi-Alphabet

translations/src/config.js
klinzo: Our current translations are probably `kmr` then. Could you verify that and rename them to `kmr`? `ckb` translations have to be created with our next translation process then.

klinzo (changes requested):
Lets make sure that the default for `ku` should really be `ckb`. Sorry for this here.

translations/src/config.js
Taggotty: Ah thank you very much. I was really confused by those alphabets.

translations/src/config.js
```diff
@@ -63,7 +63,7 @@ class Config {
   // Fallbacks for unnormalized language codes from our backend
   fallbacks: FallbacksType = {
     kmr: ['kmr'],
-    ckb: ['ku'],
+    ku: ['kmr'],
```
Taggotty, line 63: as discussed in the mattermost discussion I changed the fallback to Kurmandschi

translations/src/config.js
```diff
@@ -63,7 +60,7 @@ class Config {
   // Fallbacks for unnormalized language codes from our backend
   fallbacks: FallbacksType = {
     kmr: ['kmr'],
```
maxammann, line 62: Can be removed

maxammann (approved):

translations/src/config.js
```diff
@@ -44,10 +44,7 @@ class Config {
     es: { rtl: false },
     sr: { rtl: false },
     ps: { rtl: true },
-    kmr: {
-      rtl: true,
-      additionalFont: 'lateef'
-    },
+    kmr: { rtl: false },
```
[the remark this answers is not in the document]
klinzo: Actually I would stick with `kmr`:
> Script subtags should only be used as part of a language tag when the script adds some useful distinguishing information to the tag. Usually this is because a language is written in more than one script or because the content has been transcribed into a script that is unusual to the language (so one might tag Russian transcribed into the Latin script with a tag such as ru-Latn).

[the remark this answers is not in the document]
maxammann: I think kmr or kmr-Latn is fine as there is a alias defined in the "one and only locale repository": https://github.com/unicode-org/cldr-json/blob/872e62d6ca2343412311520909ee0c1f78e93d99/cldr-json/cldr-core/supplemental/aliases.json#L838

 @klinzo I did not find hints that kmr can be written RTL. Where did you find that?

translations/src/config.js
```diff
@@ -34,8 +34,8 @@ class Config {
     tr: { rtl: false },
     pl: { rtl: false },
     ti: { rtl: false },
-    ku: {
-      rtl: false,
+    ckb: {
```
maxammann, line 37: Does it make sense to use ckb-Latn. ckb also uses a non latin alphabet in our translations.

translations/src/config.js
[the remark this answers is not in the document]
klinzo: https://de.wikipedia.org/wiki/Kurmandschi
> Kurmandschi wird seit den 1930er Jahren vorwiegend im kurdisch-lateinischen Alphabet geschrieben.

Yeah even though we could do that, I think it does not mean we should do it : D I mean as quoted w3.org says it should only include the script if it adds useful distinguishing information, which is not the case here imo as Kurmandschi is only written in latin.

Did I really say krm can be written in RTL? I don't think so :see_no_evil:

translations/src/config.js
klinzo: It may make sense for Sorani.
From https://de.wikipedia.org/wiki/Sorani:
> Zur Schreibung des Sorani wird meist die Persische Schrift mit Sonderzeichen verwendet, zunehmend aber auch das kurdisch-lateinische Alphabet.

Lets discuss this in the App Call later.

Taggotty: does that mean we already have ckb translations? Where could I find them? And of course if theire implementing the latin alphabet it should be ckb-Latn. Even though in our wiki we explicitly write ckb with persian alphabet.

klinzo: I don't think we have ckb translations and I also don't think we will have ckb-Latn. But how should we know? :D

translations/src/config.js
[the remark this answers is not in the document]
Taggotty: So we decided to used `kmr` for northern Kurdish and `ckb-Latn` and `ckb-Arab` for the different Sorani versions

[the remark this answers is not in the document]
klinzo: Is `ckb-Arab` correct? I think it is a persian alphabet :thinking: 
https://de.wikipedia.org/wiki/Sorani#Verbreitung_und_Dialekte
> Zur Schreibung des Sorani wird meist die Persische Schrift mit Sonderzeichen verwendet

klinzo:
Could you please update the wiki with our newest decisions :D?

translations/src/config.js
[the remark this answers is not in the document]
Taggotty: I think there is no subtag for the persian alphabet explicitly. But as the persian and arab alphabet are almost similar I guess the tag is `ckb-Arab`. I also found an old issue in the firefox dev history where they say ckb-Arab is correct: https://bugzilla.mozilla.org/show_bug.cgi?id=580798#c5

klinzo (approved):

translations/src/config.js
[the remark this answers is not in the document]
klinzo: Seems correct, did not find one here: https://en.wikipedia.org/wiki/ISO_15924

translations/src/config.js
```diff
@@ -34,20 +34,18 @@ class Config {
     tr: { rtl: false },
     pl: { rtl: false },
     ti: { rtl: false },
-    ku: {
-      rtl: false,
+    'ckb-Arab': {
+      rtl: true,
       additionalFont: 'lateef'
     },
+    'ckb-Latn': { rtl: false },
```
maxammann, line 41: Oh oke with these changes the build fails again. Did not think about it.

translations/src/config.js
```diff
@@ -34,20 +34,18 @@ class Config {
     tr: { rtl: false },
     pl: { rtl: false },
     ti: { rtl: false },
-    ku: {
-      rtl: false,
+    'ckb-Arab': {
```
maxammann, line 37: Should we maybe go here with just ckb? I'm not 100% whether that tag is good. Never seen -Arab

klinzo: It's the same as -Latn, i.e. ISO1592 (https://en.wikipedia.org/wiki/ISO_15924). As long as we don't have latin ckb it would be fine as well, but not sure.

klinzo: @Taggotty please don't loose your mind and go crazy over this mess of language codes and tags and stuff :pray: :grin:

Taggotty: Hmm. I think we need to make a decision here. After reading through this article about the kurdish alphabets and this article about the Sorani language, I feel like Central Kurdish still heavily relies on the Persian alphabet. At least at school or at home they seem to learn this alphabet. The latin alphabet might be an adaption to make Central Kurdish more accessible to foreigners to learn their language, similar to the Romaji in Japan (which is also learned at school there, but no native person would use it). In that case it might be quite strange to add translations for `ckb-Latn`. Therefore, I think this part really needs some more discussion and more expertise than we can attain from reading through wikipedia articles... But for the other version using the persian alphabet, I think we can agree to implement it, using the `ckb` tag, as already discussed.

Some other references about the alphabet:
https://www.mustgo.com/worldlanguages/kurdish/#writing
https://sites.fas.harvard.edu/~iranian/Sorani/sorani_1_grammar.pdf
Kurdish news pages also seem to only use the arab alphabet:
https://www.kurdistan24.net/ckb
https://kurdistantv.net/index.php/ku (here ku seems to mean ckb)

translations/src/config.js
Taggotty: reverted to ckb again.

translations/src/config.js
Taggotty: And to throw everyone off i also found this reddit post on this: https://www.reddit.com/r/kurdistan/comments/ax883g/sorani_written_in_latin_alphabet/?utm_source=share&utm_medium=web2x&context=3

klinzo: :smile: nice, no we have got all possible variants I guess... There is also a kmr version though: D https://www.kurdistan24.net/kmr
But I think you are right and `ckb` is (nearly) exclusively arabic/persian font, so `kmr´ and `ckb` should be enough. As long as it is clear for translators, cities and ourselves what each tag means I think we are good.

klinzo:
@maxammann ping you need to approve because you requested changes. @Taggotty if the build still fails let me know!

maxammann (approved):

PR #238, opened 2021-03-30
IGAPP-496: Connect snackbar to redux state

klinzo:


klinzo:
image

Taggotty (approved):

Taggotty:
Snackbar 🥳  
Tested on Android Emulator.

SteffiStuffel (approved):
Code looks good. 

How can i test the PermissingSnackbarContainer or the SnackbarContainer? When do they appear?

klinzo:
> Code looks good.
> 
> How can i test the PermissingSnackbarContainer or the SnackbarContainer? When do they appear?

Permission snackbars are shown on the landing page if nearby locations are enabled but the permission wasn't granted. Same for iOS for push notifications on the dashboard.

native/src/modules/app/StoreActionType.js
```diff
@@ -341,7 +341,20 @@ export type SetResourceCacheUrlActionType = {|
   +params: {| +url: string |}
 |}
 
+// Pushes a new snackbar to the queue
+export type PushSnackbarActionType = {|
+  type: 'PUSH_SNACKBAR',
```
maxammann, line 346: Should we maybe use `SET_SNACKBAR` here? There can only be one snackbar visible, according to the material design guidelines I suppose. Also `PUSH` is used with our action terminology already in a very specific sense.

maxammann: If we want to keep the array of snackbars, then I would suggest `APPEND` which is a synonym

maxammann (approved):

SteffiStuffel:
> > Code looks good.
> > How can i test the PermissingSnackbarContainer or the SnackbarContainer? When do they appear?
> 
> Permission snackbars are shown on the landing page if nearby locations are enabled but the permission wasn't granted. Same for iOS for push notifications on the dashboard. The first new snackbar (via SnackbarContainer) is added with #239.

Oke thanks. Tested it, works as expected.

native/src/modules/app/StoreActionType.js
klinzo: I'd like to keep the array, but `APPEND` would work for me. Would we stick with `POP` or e.g. use `REMOVE` @maxammann ?

klinzo: Renamed to enqueue/dequeue as push/pop/FILO is really used in terms of stacks while this is a queue/FIFO.

Taggotty (approved):

PR #239, opened 2021-03-30
IGAPP-498: Show snackbar if language change fails

klinzo:


Taggotty (approved):

SteffiStuffel (approved):
Tested and it works, if i add a 'throw new Error'. 

Should the console.warns('Route was not ready when translating..') in morphContantantLanguage maybe also be displayed in the Snackbar?

klinzo:
> Tested and it works, if i add a 'throw new Error'.
> 
> Should the console.warns('Route was not ready when translating..') in morphContantantLanguage maybe also be displayed in the Snackbar?

@SteffiStuffel not sure about that, never had that warning yet and it would be a separate issue. Is it possible that this warning does only occur because of the throw new error you added?

SteffiStuffel:
> > Tested and it works, if i add a 'throw new Error'.
> > Should the console.warns('Route was not ready when translating..') in morphContantantLanguage maybe also be displayed in the Snackbar?
> 
> @SteffiStuffel not sure about that, never had that warning yet and it would be a separate issue. Is it possible that this warning does only occur because of the throw new error you added?

No this warning is never showed in the Snackbar, my suggestion was that it maybe would make sense to show it in a snackbar if it occurs?

klinzo:
As discussed: As long as nobody every experiences this it is not relevant.

native/src/modules/endpoint/sagas/watchContentLanguageSwitch.js
```diff
@@ -75,10 +74,8 @@ export function* switchContentLanguage(
     }
     yield put(insert)
   } catch (e) {
-    if (fromError(e) === ErrorCodes.NetworkConnectionFailed) {
-      // TODO IGAPP-498 The alert should be replaced with a snackbar, hence the TFunction should also be removed.
-      Alert.alert(t('languageSwitchFailedTitle'), t('languageSwitchFailedMessage'))
```
maxammann, line 80: Should we remove this translation?

maxammann (approved):

native/src/modules/endpoint/sagas/watchContentLanguageSwitch.js
klinzo: In general yes, but does this even work if the translations for malte are reimported afterwards?

PR #241, opened 2021-03-31
IGAPP-565: Send signal if search is finished

SteffiStuffel:


native/src/routes/search/components/NothingFoundFeedbackBox.js
```diff
@@ -65,16 +65,19 @@ const NothingFoundFeedbackBox = ({ resultsFound, query, t, theme, sendFeedback }
   if (['idle', 'failed'].includes(sendingStatus)) {
     return (
       <Container>
-        <FeedbackBox>
-          <TitleText>{resultsFound ? t('feedback:informationNotFound') : t('feedback:nothingFound')}</TitleText>
+        <FeedbackBox theme={theme}>
```
klinzo, line 68: Why these changes? I just removed it everywhere :see_no_evil:

native/src/routes/search/components/SearchHeader.js
```diff
@@ -28,18 +28,24 @@ const BoxShadow: StyledComponent<{||}, ThemeType, *> = styled.View`
...
-    <HorizontalLeft theme={theme}>
-      <HeaderBackButton onPress={closeSearchBar} labelVisible={false} />
-      <ThemedSearchBar theme={theme} onChangeText={onSearchChanged} value={query} autofocus t={t} />
-    </HorizontalLeft>
-  </BoxShadow>
-)
+const SearchHeader = ({ theme, query, closeSearchBar, onSearchChanged, t }: PropsType) => {
+  const onClose = () => {
+    closeSearchBar(query)
+  }
```
klinzo, line 39: Not sure if this is necessary, the query is available in the parent component as well (and only passed as prop here).

native/src/routes/search/components/SearchModal.js
```diff
@@ -130,7 +134,7 @@ class SearchModal extends React.Component<PropsType, StateType> {
         <View style={{ minHeight: minHeight }}>
           <CategoryList
             categories={filteredCategories}
-            navigateToLink={navigateToLink}
+            navigateToLink={navigateToLink} // Do i have to add changes to navigateToLink too? When ist this used?
```
klinzo, line 137: I think it is not really used anywhere for search but only for links in 2nd level content, so probably no changes required. But please make sure this is really the case by checking usages of `navigateToLInk` in `CategoryList`.

native/src/routes/search/components/SearchModal.js
```diff
@@ -99,6 +100,9 @@ class SearchModal extends React.Component<PropsType, StateType> {
 
   onItemPress = (category: { path: string, ... }) => {
     const { cityCode, language, navigateTo } = this.props
+    const { query } = this.state
+
+    sendTrackingSignal({ signal: { name: SEARCH_FINISHED_SIGNAL_NAME, query, url: category.path } })
```
klinzo, line 105: I think `category.path` is not an url, is it?

native/src/routes/search/containers/SearchModalContainer.js
```diff
@@ -36,7 +42,8 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType) => {
     categories: state.cityContent?.searchRoute?.categoriesMap || null,
     language: state.contentLanguage,
     cityCode,
-    closeModal: () => {
+    closeModal: (query: string) => {
+      sendTrackingSignal({ signal: { name: SEARCH_FINISHED_SIGNAL_NAME, query, url: null } })
```
klinzo, line 46: Would it maybe make more sense to handle this in `SearchModal` as the other tracking signal is sent there as well?

native/src/routes/search/components/__tests__/SearchModal.spec.js
```diff
@@ -0,0 +1,42 @@
...
+        navigateToLink={dummy}
+        t={t}
+        sendFeedback={dummy}
+      />
+    )
+    const categoryListItem = getAllByTestId('CategoryListItem')[1]
+    fireEvent.press(categoryListItem)
+    expect(sendTrackingSignal).toHaveBeenCalledTimes(1)
+    expect(sendTrackingSignal).toHaveBeenCalledWith({
+      signal: { name: SEARCH_FINISHED_SIGNAL_NAME, query: '', url: Array.from(categoriesMapModel._categories)[2][0] }
```
klinzo, line 39:
```suggestion
      signal: { name: SEARCH_FINISHED_SIGNAL_NAME, query: '', url: categoriesMapModel.toArray()[2][0] }
```

native/src/routes/search/components/__tests__/SearchModal.spec.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('SearchModal', () => {
+  beforeEach(() => {})
+
+  const dummy = jest.fn()
+  const t = key => key
+  const categoriesMapModel = new CategoriesMapModelBuilder('augsburg', 'de').build()
+
+  it('search modal should send tracking signal on Press', () => {
+    const { getAllByTestId } = render(
```
klinzo, line 22: If possible using test ids should be avoided. Instead it would be better to query by e.g. text, see https://testing-library.com/docs/queries/about/#priority

native/src/routes/search/components/NothingFoundFeedbackBox.js
SteffiStuffel: I added a test, and flow was complaining that the theme could not be found.

native/src/routes/search/components/SearchModal.js
SteffiStuffel: fixed it.

native/src/routes/search/components/NothingFoundFeedbackBox.js
klinzo: Just checked it out, I don't get any errors if I just revert this file.

native/src/routes/search/containers/SearchModalContainer.js
SteffiStuffel: I decided to sendSignal where the navigation action is executed, but i can also be done as you suggested. changed it.

native/src/routes/search/components/__tests__/SearchModal.spec.js
SteffiStuffel: Yeah, i coundn't find it byText. Title of the generated Categories must be `Category with id ${id}`, but GetAllByText('Category') does not return anything.

native/src/routes/search/components/NothingFoundFeedbackBox.js
SteffiStuffel: sorry my fault, it is not flow. The test is failing when reverting this:

SearchModal › search modal should send tracking signal on Press

    TypeError: Cannot read property 'backgroundAccentColor' of undefined

klinzo: Yes, you need to wrap it in a ThemeProvider, example from web: https://github.com/Integreat/integreat-app/blob/a32091c33c10148a3617d6c873a304877788f96d/web/src/modules/layout/containers/__tests__/LocationHeader.spec.js

native/src/routes/search/components/__tests__/SearchModal.spec.js
klinzo: I think you need to set `exact` to false as you are just querying by a substring:
https://testing-library.com/docs/queries/about#precision
Did not find docu for native.

native/src/routes/search/components/NothingFoundFeedbackBox.js
SteffiStuffel: I think this is currently only possible for web?

https://github.com/Integreat/integreat-app/pull/221#discussion_r600587803

klinzo: Done here: https://github.com/Integreat/integreat-app/pull/227/files#r601429935

native/src/routes/search/components/__tests__/SearchModal.spec.js
```diff
@@ -0,0 +1,55 @@
...
+      route: CATEGORIES_ROUTE,
+      cityContentPath: categoriesMapModel.toArray()[2].path,
+      cityCode,
+      languageCode: language
+    }
+    const expectedUrl = urlFromRouteInformation(routeInformation)
+
+    expect(sendTrackingSignal).toHaveBeenCalledWith({
+      signal: { name: SEARCH_FINISHED_SIGNAL_NAME, query: '', url: expectedUrl }
+    })
```
Taggotty, line 53: could you also add a test where the query isn't empty? you can use something like `fireEvent.changeText(getByPlaceholder('Search'), 'query')` to change the text.

native/src/routes/search/components/__tests__/SearchModal.spec.js
```diff
@@ -22,22 +25,26 @@ describe('SearchModal', () => {
...
-        navigateTo={dummy}
-        theme={lightTheme}
-        language={language}
-        cityCode={cityCode}
-        closeModal={dummy}
-        navigateToLink={dummy}
-        t={t}
-        sendFeedback={dummy}
-      />
+  it('search modal should send tracking signal on Press', async () => {
```
klinzo, line 28:
```suggestion
  it('should send tracking signal when opening a search result', async () => {
```

klinzo (approved):
Is there already a test that a tracking signal is sent if the search is closed without clicking on a result?

native/src/routes/search/components/__tests__/SearchModal.spec.js
SteffiStuffel: done

native/src/routes/search/components/__tests__/SearchModal.spec.js
```diff
@@ -0,0 +1,80 @@
...
+      <ThemeProvider theme={lightTheme}>
+        <SearchModal {...props} />
+      </ThemeProvider>
+    )
+    const button = getAllByRole('button')[0]
+    const searchBar = getByPlaceholderText('searchPlaceholder')
+    await fireEvent.changeText(searchBar, 'Category')
+    await fireEvent.press(button)
+    expect(sendTrackingSignal).toHaveBeenCalledTimes(1)
+
```
SteffiStuffel, line 50: Added this test for closing the search, but it fails an i have no idea why. It says, that sendTrackingSignal is not called, but when debugging it the code is executed, but somehow the test does not register it.

Taggotty: And even though the test below doesn't fail, you basically create the same situation with `onItemPress`. The function execution order is pretty unforseeable for asynchronous functions, when they are not awaited. That's why I'd propose to also apply the solution to the other test.

Taggotty (approved):

maxammann (approved):

native/src/routes/search/components/__tests__/SearchModal.spec.js
SteffiStuffel: Thank you very much. Now it works :)

PR #244, opened 2021-03-31
IGAPP-560: Send signal if page is closed

klinzo:

This is definitely not an optimal solution but I did not find an official react-navigation way to do this and it is relatively simple.
Other ways I could think of:
- Add a hoc to all route containers and send on unmount
- Somehow add react-navigation focus listener to all screens and send if page is focused again: https://reactnavigation.org/docs/function-after-focusing-screen/#triggering-an-action-with-a-focus-event-listener
Not sure how we could exclude app state changes then though...

What do you thin about this? This does not handle navigation from city content to landing and vice versa, i.e. all navigating that either removes more than one screen at once or just replaces the current screen. Not sure if I should include it.


native/src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -72,7 +72,7 @@ export default function* fetchResourceCache(
       // TODO: we might remember which files have failed to retry later
       // (internet connection of client could have failed)
       const message = createErrorMessage(failureResults)
-      console.warn(message)
+      console.log(message)
```
klinzo, line 75: What do you think about this, especially @maxammann? I think it is really annoying to have these warnings that a file could not be downloaded all the time during development. But if you think I should keep it I'll revert this.

SteffiStuffel: also think it is annoying, but do not know if we should keep it.

native/src/modules/app/components/App.js
```diff
@@ -45,22 +46,22 @@ const store: Store<StateType, StoreActionType> = createReduxStore(dataContainer)
...
-    if (state) {
-      const route = state.routes[state.index]
-      setRouteName(route.name)
-      setRouteKey(route.key)
-      if (route.params?.cityCode) {
-        setCityCode(typeof route.params.cityCode === 'string' ? route.params.cityCode : null)
+  const onStateChange = useCallback(
+    state => {
+      if (state) {
+        if (state.index === routeIndex - 1) {
```
SteffiStuffel, line 54: Is there a reason, why you decided to work with index instead of route name.

In the docu it is done with the route name and i am not sure it the index is less robust and there may be less suitable.

native/src/modules/app/components/Navigator.js
```diff
@@ -65,9 +65,7 @@ type PropsType = {|
   fetchCategory: (cityCode: string, language: string, key: string, forceUpdate: boolean) => void,
   fetchCities: (forceRefresh: boolean) => void,
   routeKey: ?string,
-  routeName: ?string,
-  cityCode: ?string,
-  languageCode: ?string
+  routeName: ?string
```
SteffiStuffel, line 68: Why have you removed city and language? Is this part of the ticket and needed or refactoring?

SteffiStuffel (approved):
With short research i found it would be possible to do screen tracking with

- use onStateChange in the NavigationContainer [here](https://reactnavigation.org/docs/screen-tracking/) -> mostly what you have done, and i think it is the best solution, maybe modify and use routeName instead of index, if this works also and/or better
- or use a redux middleware for screen tracking [here](https://reactnavigation.org/docs/2.x/screen-tracking/#screen-tracking-with-redux)

I guess you thought about it, but just wanted to drop it. 


Your other suggestions:
1) Add a hoc to all route containers and send on unmount: Think this could be a nice solution too, would also handle app closes properly i guess?

2) Somehow add react-navigation focus listener to all screens and send if page is focused again: This would not handle if the app is closed, but it would be possible to handle this seperatly i guess. Do not like this option though, because we do not really want to listen on "new focues".


Also the Offers-Page is not working properly.

Open Offers -> GoBack ('close' Signal is sent)

Open Offers -> Click on e.g. Sprungbrett ('close' signal is not sent)

Taggotty (approved):
I think this is probably the cleanest solution possible. I would also agree to not using a focus listener, as it seems to listen to more than originally intended by us.

native/src/modules/app/components/Navigator.js
klinzo: Refactoring, was not used anywhere anymore.

klinzo:
> With short research i found it would be possible to do screen tracking with
> 
>     * use onStateChange in the NavigationContainer [here](https://reactnavigation.org/docs/screen-tracking/) -> mostly what you have done, and i think it is the best solution, maybe modify and use routeName instead of index, if this works also and/or better
> 
>     * or use a redux middleware for screen tracking [here](https://reactnavigation.org/docs/2.x/screen-tracking/#screen-tracking-with-redux)
> 
> 
> I guess you thought about it, but just wanted to drop it.
> 
> Your other suggestions:
> 
>     1. Add a hoc to all route containers and send on unmount: Think this could be a nice solution too, would also handle app closes properly i guess?
> 
>     2. Somehow add react-navigation focus listener to all screens and send if page is focused again: This would not handle if the app is closed, but it would be possible to handle this seperatly i guess. Do not like this option though, because we do not really want to listen on "new focues".
> 
> 
> Also the Offers-Page is not working properly.
> 
> Open Offers -> GoBack ('close' Signal is sent)
> 
> Open Offers -> Click on e.g. Sprungbrett ('close' signal is not sent)

@SteffiStuffel  So I think we are talking about little different things here. Our current tracking approach is to send two signals for user navigation:
* OPEN_PAGE if the user navigates to a NEW route (e.g. by clicking on a category or the events or offers navigation item
* CLOSE_PAGE if the user navigates BACK, i.e. closing a route that was opened previously
This differs from the approach in the docu as there is just one signal sent if the currently opened screen changes. This is e.g. visible in the example you described when opening the sprungbrett. No screen is closed but the offers screen just goes to the background, therefore no close signal is sent.

Maybe (or even probably) the approach from the docs (screen tracking) would make more sense, but I think it is very difficult to implement (how do we get the url of the new page?) and conflicts with the agreement with JPAL.

SteffiStuffel:
> We can discuss this in a call as well if you want :)

No need. I just unterstood it wrong (cause i do not really get the usecase of this, but none of my buisness ;) ) and if this is the intened behaviour, i'm perfectly fine with your solution.

klinzo:
> > We can discuss this in a call as well if you want :)
> 
> No need. I just unterstood it wrong (cause i do not really get the usecase of this, but none of my buisness ;) ) and if this is the intened behaviour, i'm perfectly fine with your solution.
> 
> Just one other thing i noticed: Open langugage change does not send a "Open-Page" signal, but when navigating back a "close-page" signal is sent. (If this is also intended, just start ignoring me smile )

Not intended, good catch. Will fix it. And don't get me wrong, I think your proposed solution would normally be the better approach. But our architecture is sometimes a bit complicated and kind of prevents it.

native/src/modules/endpoint/sagas/fetchResourceCache.js
maxammann: We can remove it if it bugs you. Never had more than one message.

PR #245, opened 2021-04-02
IGAPP-576: Remove feedback type selection

Taggotty:


web/src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -1,55 +1,40 @@
...
 type PropsType = {|
-  cities: ?Array<CityModel>,
-  title?: string,
   alias?: string,
   path?: string,
-  query?: string,
   isPositiveRatingSelected: boolean,
   location: LocationState,
-  offers: ?Array<OfferModel>,
-  postFeedbackDataOverride?: FeedbackParamsType => void,
```
Taggotty, line 43: removed unused properties.

web/src/modules/feedback/components/FeedbackBoxContainer.js
```diff
@@ -60,189 +45,62 @@ type StateType = {|
...
-      feedbackType: selectedFeedbackOption.feedbackType,
-      feedbackCategory: selectedFeedbackOption.feedbackCategory,
+      feedbackType: feedbackType,
       isPositiveRating: isPositiveRatingSelected,
       comment,
       permalink: path,
       city: city || INTEGREAT_INSTANCE,
       language: language || DEFAULT_FEEDBACK_LANGUAGE,
-      alias: feedbackAlias,
-      query
```
klinzo, line 223: Why did you remove this? Because we have no feedback for the search, right?

web/src/modules/feedback/components/FeedbackBox.js
```diff
@@ -66,8 +58,6 @@ export const FeedbackBox = ({
 }: PropsType) => (
   <StyledFeedbackBox>
     <ModalHeader t={t} closeFeedbackModal={closeFeedbackModal} title={t('feedback')} />
-    <Description>{t('feedbackType')}</Description>
-    <Dropdown items={feedbackOptions} selectedItem={selectedFeedbackOption} onOptionChanged={onFeedbackOptionChanged} />
```
klinzo, line 70: Do we still use this dropdown somewhere or could we remove it? Does it use a library which could be removed?

klinzo (approved):
Tests, tests, tests :tada: :t-rex:

web/src/modules/feedback/FeedbackVariant.js
```diff
@@ -1,28 +1,16 @@
 // @flow
```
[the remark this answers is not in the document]
Taggotty: actually it is not used anymore. but in the dropdown component. I removed both the `Dropdown` component (as it was only used for feedback and is not used anymore) and the `FeedbackVariant`.

web/src/modules/feedback/components/FeedbackBox.js
Taggotty: Removed with unused libraries.

web/src/modules/feedback/components/FeedbackBoxContainer.js
Taggotty: `SearchFeedback.js` implements this property, but here it is not necessary.

SteffiStuffel (approved):

PR #250, opened 2021-04-14
IGAPP-493: Setup e2e tests native

Taggotty:


Taggotty:
Currently in draft mode, so I can test ios properly.

docs/e2e-tests.md
```diff
@@ -28,3 +28,21 @@ The [E2E test folder](../e2e-tests/web/test) contains two important subdirectori
...
+Add the accessibility-id to a react component using `testID('Example-Component')`. You can query this component with `$('~Example-Component')` in your test.
+
+For more complex queries you should add/use a custom [Selector]('../e2e/native/Selector.ts) using [predicate strings](https://github.com/facebookarchive/WebDriverAgent/wiki/Predicate-Queries-Construction-Rules) for iOS and [UiSelectors](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector) for Android.
+
+## Troubleshooting
+
+### Cannot hide Keyboard on iOS
+
+https://stackoverflow.com/a/54995267
+>The Appium method hideKeyboard() is known to be unstable when used on iPhone devices, as listed in Appium’s currently known open issues. Using this method for an iOS device may cause the Appium script to hang. Appium identifies that the problem is because - "There is no automation hook for hiding the keyboard,...rather than using this method, to think about how a user would hide the keyboard in your app, and tell Appium to do that instead (swipe, tap on a certain coordinate, etc..
```
klinzo, line 48:
```suggestion
>The Appium method hideKeyboard() is known to be unstable when used on iPhone devices, as listed in Appium’s currently known open issues. Using this method for an iOS device may cause the Appium script to hang. Appium identifies that the problem is because "There is no automation hook for hiding the keyboard". Rather than using this method just think about how a user would hide the keyboard in your app, and tell Appium to do that instead (swipe, tap on a certain coordinate, etc..
```

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    switch (direction) {
+        case "UP":
+            startPoint = {
+                x: width * anchorPercentage / 100,
+                y: height * lowerPercentage / 100
+            }
+            endPoint = {
+                x: width * anchorPercentage / 100,
+                y: height * upperPercentage,
```
klinzo, line 52: Is this missing?
```suggestion
                y: height * upperPercentage / 100,
```

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const anchorPercentage = 50;
+
+    let startPoint: { x: number, y: number };
+    let endPoint: { x: number, y: number };
+
+    switch (direction) {
+        case "UP":
+            startPoint = {
+                x: width * anchorPercentage / 100,
```
klinzo, line 47: Couldn't you just do as `anchorPercentage` is always constant, so this would be easier to understand?
```suggestion
                x: width * 0.5,
```

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
...
+                y: height * upperPercentage / 100
+            }
+            endPoint = {
+                x: width * anchorPercentage / 100,
+                y: height * lowerPercentage / 100,
+            }
+    }
+    await driver.touchPerform([].concat(... new Array(repeat).fill(swipeMove(startPoint, endPoint))))
```
klinzo, line 65: Interesting :thinking: Do you really need the concat or wouldn't this be enough as well?
```suggestion
    await driver.touchPerform(Array(repeat).fill(swipeMove(startPoint, endPoint)))
```

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const hideKeyboard = async () => {
+    if (driver.isAndroid) {
+        await driver.hideKeyboard()
+    } else {
+        await driver.sendKeys(['\n'])
```
klinzo, line 72: Why not use `swipe` directly here? Isn't that the way to close the keyboard on iOS?

e2e-tests/native/test/pageobjects/landing.page.ts
```diff
@@ -0,0 +1,22 @@
+import {Page} from "./page";
+import {Selector} from "../Selector";
+
+class LandingPage extends Page{
+    constructor() {
+        super('Landing-Page');
+    }
+
+    get cities() {
+        return $$('~City-Entry')
```
klinzo, line 10: What is the difference between `$$` and `$`?

docs/e2e-tests.md
```diff
@@ -28,3 +28,21 @@ The [E2E test folder](../e2e-tests/web/test) contains two important subdirectori
...
+### Native
+
+#### Selectors
+
+Selectors are used to select an element in the app programatically.
+However, during native development most of the common selectors are not available. 
+Therefore, you should use the accessibility identifier for this.
+Add the accessibility-id to a react component using `testID('Example-Component')`. You can query this component with `$('~Example-Component')` in your test.
+
+For more complex queries you should add/use a custom [Selector]('../e2e/native/Selector.ts) using [predicate strings](https://github.com/facebookarchive/WebDriverAgent/wiki/Predicate-Queries-Construction-Rules) for iOS and [UiSelectors](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector) for Android.
```
klinzo, line 41: The `WebDriverAgent` repo is archived, is that intended?

e2e-tests/native/test/Selector.ts
```diff
@@ -0,0 +1,21 @@
+
+export class Selector {
+    private queries: Array<string> = new Array<string>()
+
+    public ByText(text: string): Selector {
```
klinzo, line 5: Not sure but shouldn't this be lowercase?
```suggestion
    public byText(text: string): Selector {
```

e2e-tests/native/test/pageobjects/landing.page.ts
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    get search() {
+        return $('~Search-Input')
+    }
+
+    async city(name: string) {
+        return $(new Selector().ByText(name).build())
```
klinzo, line 18: So we need the Selector class for text and normal strings for accessibility ids, right?

native/package.json
```diff
@@ -8,7 +8,7 @@
     "start:integreat": "cross-env BUILD_CONFIG_NAME='integreat' react-native start --reset-cache",
     "start:malte": "cross-env BUILD_CONFIG_NAME='malte' react-native start --reset-cache",
     "start:aschaffenburg": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native start --reset-cache",
-    "start:integreat-e2e": "cross-env BUILD_CONFIG_NAME='integerat-e2e' react-native start --reset-cache",
+    "start:integreat-e2e": "cross-env BUILD_CONFIG_NAME='integreat-e2e' react-native start --reset-cache",
```
klinzo, line 11: who doesn't know integreat :smile:

e2e-tests/native/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const cities = await LandingPage.cities
+    const search = await LandingPage.search
+
+    expect(cities.length).toBeGreaterThan(0)
+    await search.click()
+    await search.addValue(filter)
+
+    await hideKeyboard()
+    await swipe('DOWN')
```
klinzo, line 20: This looks kind of weird if you don't know why this is the case, I think it would be clearer to just to a swipe in hideKeyboard for iOS. Or is it needed on android as well?

e2e-tests/native/wdio-browserstack.conf.ts
```diff
@@ -0,0 +1,74 @@
+import {execSync} from 'child_process'
```
klinzo, line 1: Is this copied from web? Looks familiar

e2e-tests/native/wdio-browserstack.conf.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const browserstackCaps = (config: Capabilities): Capabilities => {
+    const isCi = !!process.env.E2E_CI
+    const prefix = isCi ? 'IG CI' : 'IG DEV'
+    const app = config.platformName === 'android' ? process.env.E2E_BROWSERSTACK_APP_ANDROID : process.env.E2E_BROWSERSTACK_APP_IOS
+    return {
+        'bstack:options': {
+            buildName: `${prefix}: ${getGitBranch()}`,
+            sessionName: `${config.browserName?.toLowerCase()}: ${getGitHeadReference()}`,
+            projectName: 'integreat-react-native-app',
```
klinzo, line 20: Is this the right project name? I think this was the name from the previous multi repo setup, right?

klinzo (changes requested):
Looks quite straightforward, nice :tada: To 50% e2e test coverage :rocket: :wine_glass:

e2e-tests/native/test/pageobjects/landing.page.ts
Taggotty: that's correct. It is possible to implement selection also with the selector class. Then we could chain `ByText` and `ByAccessibilityId`. I did not implement it, as I thought that we would like to avoid platform specific selectors as much as possible. But it's probably interesting to think about this now to avoid later refactoring.

e2e-tests/native/wdio-browserstack.conf.ts
Taggotty: yep, it is

e2e-tests/native/wdio-browserstack.conf.ts
Taggotty: it is the name which is currently used in the e2e tests. Do you think just `integreat-app` makes more sense?

e2e-tests/native/test/pageobjects/landing.page.ts
klinzo: I agree totally with you, if it works directly without custom selectors I don't see a reason why we should make it more complicated. But I am not really familiar with the topic...

e2e-tests/native/wdio-browserstack.conf.ts
klinzo: Do we need both or could we use just one for both projects?

e2e-tests/native/wdio-browserstack.conf.ts
klinzo: No, I was just asking :D Maybe `integreat-app-native` and `integreat-app-web` would make sense and be consistent with our project setup. But I don't mind keeping this reminder to old times :grin:

e2e-tests/native/test/Action.ts
SteffiStuffel: Really? If there is just 0.5, i would wonder why it is like this. Think the original solution prevents magic numbers and is more understandable.

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
...
+            options: endPoint,
+        },
+        {
+            action: 'release',
+            options: {},
+        }
+    ]
+}
+
+export const swipe = async (direction: SwipeDirectionType, repeat: number = 1, amount: number = 80) => {
```
SteffiStuffel, line 28: I get an eslint error here for 'no-inferrable-types'. Is not linting done for these files here?

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,74 @@
+type SwipeDirectionType = 'UP' | 'DOWN'
```
SteffiStuffel, line 1: Typescript supports enums and i would really recommend using them.

So this would change to:
enum SwipeDirectionType {
  Up,
  Down,
}
and us it later on swipte(SwipeDirectionType.Down) or is the switch/case. This allows auto-complete and  checks for wrong code.

SteffiStuffel (changes requested):
About the documentation: Why are we having two files for e2e-test documentation. I do not really get the difference, can we merge them into one file about e2e-testing?
in e2e-test.ms there is the command yarn e2e:web mentioned, but i can not find where this should be? 

- I tried to run the e2e-test locacally but failed, with 'UI langugae > should match the system language'. I installed and started appium and started an emulator and run: E2E_CONFIG=local_android yarn test:e2e, as described in the docs. Are any driver specific setups needed for Appium Server?
Running both prepare tasks (prepare:android:native and prepare:start:native and then run test:native) -> Works fine :) 

- yarn test:web -> works fine.

- Tried to install appium-desktop -> followed the instructions, but got: electron: Befehl nicht gefunden.

- have not tried the browserstack stuff, what would be a usecase for this? (I would expect the workflow to be: write an e2e test -> run it locally -> upload it -> CI runs it in browserstack)

- is iOS E2E testing only possible with a mac?

docs/e2e-tests.md
Taggotty: Actually no it isn't. But personally I liked that this guide is pretty short and only showed the relevant predicates. A full guide would can be found in the [Apple Documentation Archive](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html). Would you prefer if I replace it?

klinzo: No, I guess it's fine to keep it. If you want, you can add apple as second resource, your choice!

e2e-tests/native/test/Action.ts
```diff
@@ -0,0 +1,77 @@
+export enum SwipeDirection {
```
klinzo, line 1: Indentation seems off here, looks like it uses 4 spaces.

e2e-tests/native/test/Selector.ts
maxammann: I suppose the convention is to use uppercase names for selectors.

e2e-tests/native/test/pageobjects/page.ts
```diff
@@ -0,0 +1,13 @@
+
+export class Page {
+    readonly pageId
+
+    constructor(pageId: string) {
+        this.pageId = pageId
+    }
+
+    public async exists(): Promise<boolean | void> {
```
maxammann, line 9: Better return type would be just Promise<boolean>. Can we map void to boolean here?

e2e-tests/native/wdio-browserstack.conf.ts
maxammann: I'm fine with both, copying or keeping it separate. There will be only 2 configs and 2 copies are still fine imo.

e2e-tests/native/wdio.conf.ts
```diff
@@ -0,0 +1,42 @@
+import {Capabilities} from '@wdio/types/build/Capabilities'
+
+const androidCapabilities: Capabilities = {
+    platformName: 'android',
+    'appium:app': '../native/android/app/build/outputs/apk/debug/app-debug.apk',
```
maxammann, line 5: Do we really want to use the app-debug here? Make release it more appropriate.

maxammann (approved):

e2e-tests/native/wdio-browserstack.conf.ts
Taggotty: For now I will leave it where it is. Due to @maxammann argumentation

Taggotty:
> About the documentation: Why are we having two files for e2e-test documentation. I do not really get the difference, can we merge them into one file about e2e-testing?

One documentation is the legacy documentation of the old e2e tests (the on in the native subdirectory) and the other one refers to the more current documentation.

> in e2e-test.ms there is the command yarn e2e:web mentioned, but i can not find where this should be?

It's supposed to be yarn `test:web`. I messed this one up. Will be fixed.

> I tried to run the e2e-test locacally but failed, with 'UI langugae > should match the system language'. 
> I installed and started appium and started an emulator and run: E2E_CONFIG=local_android yarn test:e2e, as described in the docs. Are any driver specific setups needed for Appium Server?

I think you were not in the correct project here. We still have our old e2e tests as long as the CI is not ready. The new tests can be found in the e2e-tests folder.

>  Tried to install appium-desktop -> followed the instructions, but got: electron: Befehl nicht gefunden.

Weird. For me it did work. But in general you wont

> have not tried the browserstack stuff, what would be a usecase for this? (I would expect the workflow to be: write an e2e test -> run it locally -> upload it -> CI runs it in browserstack)
>  is iOS E2E testing only possible with a mac?

Yes. iOS devices can afaik only be tested on a mac and maybe on real devices. But that I could not test. That is also the reason why we use browserstack, as our CI will run in browserstack.

e2e-tests/native/test/Action.ts
klinzo: Could you explain please, I don't get it...

klinzo (approved):

e2e-tests/native/test/Action.ts
klinzo: But this closes the keyboard now, right?

e2e-tests/native/test/Action.ts
klinzo: Okay, I am stupid. This is just about no magic number and speaking variables, right? And not that the resulting value is not `0.5`...

e2e-tests/native/test/specs/navigateToDashboard.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const cities = await LandingPage.cities
+    const search = await LandingPage.search
+
+    expect(cities.length).toBeGreaterThan(0)
+    await search.click()
+    await search.addValue(filter)
+
+    await hideKeyboard()
+    await swipe(SwipeDirection.Down)
```
klinzo, line 20: Do we still need both? Why?

e2e-tests/native/wdio.conf.ts
Taggotty: This config should manly be used for local testing. Therefore, I think the better workflow is to develop the tests on the e2e debug apk locally and then let it execute the release config in the CI. Imo there is no difference between the release and the e2e build config (which skips the intro slides) here.

e2e-tests/native/test/specs/navigateToDashboard.e2e.ts
Taggotty: Because, the swiping does not work to hide the keyboard on iOS. At least I could not make it work. On Android you can actually say `driver.hideKeyboard('swipeDown')`, but iOS devices don't support this.

klinzo: Okay, not sure I understand but I guess you know what you do here :D

SteffiStuffel (approved):

e2e-tests/native/test/Action.ts
Taggotty: It should yes. I added the option to send any KeyChar instead of always enter, as i am not sure whether enter always works...

PR #251, opened 2021-04-14
IGAPP-551 Add language Serbian (for Malte)

Znippet:
Tested:
image
image


klinzo (approved):
I don't get your screenshot tbh. AFAIK we don't have any Cyrillic UI app translations for Serbian and no Latin Serbian translations in the CMS, but your screenshot shows it the other way round :thinking: 

_I added Cyrillic UI app translations but I didn't commit them.
I renamed the old "sr" to latin-sr.
Then I added some manual translations for "sr-Cyrl" so the old "sr" could fall back to it. 
So what you see are my (uncommited) test-translations for cyrillic_

klinzo (changes requested):
We have to adjust the translations in the malte.json and aschaffenburg.json overwrite files as well, forgot that myself as well probably. We may also have to adjust it for the whole bunch of Persian and Kurdish languages as well, must not be done in your PR though.
Tested on android and web, apart from the thing mentioned above everything works as expected.

_Updated malte.json. aschaffenburg.json doesn't have serbian at all until now so no changes here_

maxammann (approved):
looks good

maxammann:
@klinzo fixing malte.json and aschaffenburg.json should be done soon I think. At least before merging in new translations. Else this gets quite complicated later and nobody remembers.

klinzo:
@maxammann did not find any things that need fixing, so I guess we are fine.

klinzo (approved):

PR #252, opened 2021-04-15
IGAPP-559: Additional field for feedback mail in search feedback

SteffiStuffel:

NothingFoundFeedbackBox is deleted. Both Feedbacks (from search and form useful / not-useful - Buttons) are using the same components. 
Is only the native part of IGAPP-559. Web will follow in a seperate PR.


translations/translations.json
```diff
@@ -1227,32 +1227,20 @@
   },
   "feedback": {
     "de": {
-      "technicalTopics": "Technische Funktionen",
```
klinzo, line 1230: :+1: Thanks!

translations/translations.json
```diff
@@ -1227,32 +1227,20 @@
...
       "useful": "Hilfreich",
       "notUseful": "Nicht hilfreich",
       "nothingFound": "Es wurden leider keine passenden Ergebnisse gefunden.",
-      "yourFeedback": "Ihr Feedback...",
-      "informationNotFound": "Nicht das Richtige gefunden?",
       "wantedInformation": "Welche Informationen fehlen hier?",
       "feedbackSent": "Feedback abgeschickt",
       "thanksMessage": "Vielen Dank für Ihr Feedback. Damit helfen Sie uns, unsere Inhalte und die {{appName}}-App zu verbessern. Wir würden uns über weiteres Feedback freuen.",
       "close": "Schließen",
-      "negativeRating": "Diese Seite ist nicht hilfreich",
```
klinzo, line 1254: I think this is still used on web, best check other translations as well before deleting.

native/src/routes/search/components/SearchModal.js
```diff
@@ -35,14 +37,17 @@ export type PropsType = {|
...
+  cities: $ReadOnlyArray<CityModel>
 |}
 
-type StateType = {|
+type SearchStateType = {|
   query: string
 |}
 
-class SearchModal extends React.Component<PropsType, StateType> {
+const TranslatedFeedbackContainer = withTranslation<FeedbackContainerPropsType>('feedback')(FeedbackContainer)
```
klinzo, line 48: I don't think this is the right place for this, is it? Can't you just pass the TFunction to the component? Is it because of namespaces? Maybe using `feedback:<your-translation-key>` would work.

native/src/routes/search/components/SearchModal.js
```diff
@@ -123,6 +128,8 @@ class SearchModal extends React.Component<PropsType, StateType> {
     }
 
     const filteredCategories = this.findCategories(categories)
+    const feedbackOrigin = filteredCategories.length !== 0 ? 'searchInformationNotFound' : 'searchNothingFound'
```
klinzo, line 131: According to some article I saw a while ago negated expressions are harder to understand than non-negated things. I don't mind keeping it this way though, just wanted to share this.

native/src/routes/search/containers/SearchModalContainer.js
```diff
@@ -83,4 +106,12 @@ const SearchModalContainer = (props: PropsType) => {
...
-export default connect<PropsType, OwnPropsType, _, _, _, _>(mapStateToProps, mapDispatchToProps)(SearchModalContainer)
+const refresh = (refreshProps: OwnPropsType) => {
+  const { navigation } = refreshProps
+  navigation.navigate({ name: SEARCH_ROUTE })
+}
+
+export default connect<PropsType, OwnPropsType, _, _, _, _>(
+  mapStateToProps,
+  mapDispatchToProps
+)(withPayloadProvider<InnerPropsType, OwnPropsType, SearchRouteType>(refresh, () => {}, true)(SearchModalContainer))
```
klinzo, line 117: Why did you choose to add the `withPayloadProvider? For refreshing behavior or to show a loading spinner while cities are not available?

native/src/modules/feedback/Feedback.js
```diff
@@ -46,47 +48,87 @@ const Description = styled(ThemedText)`
...
 |}
 
-const FeedbackModal = (props: PropsType) => {
+const Feedback = (props: PropsType) => {
   const renderBox = (): React.Node => {
-    const { theme, t, isPositiveFeedback, comment, contactMail, sendingStatus } = props
+    const { theme, t, feedbackOrigin, comment, contactMail, sendingStatus } = props
+    let title
+    let isSearchFeedback = false
+    switch (feedbackOrigin) {
```
klinzo, line 82: Maybe we could just let the parent component decide what the title is? I.e. we pass the (translated or untranslated) title key and whether it is positive or negative feedback to the component instead of doing the switch here? I mean feedback in the search because of missing content is also negative feedback, so maybe that would make sense? What do you think?

native/src/modules/feedback/Feedback.js
```diff
@@ -46,47 +48,87 @@ const Description = styled(ThemedText)`
...
         <>
-          <Caption theme={theme} title={t('feedback')} />
-
+          {!isSearchFeedback && <Caption theme={theme} title={t('feedback')} />}
+          {feedbackOrigin === 'searchNothingFound' && (
+            <>
+              <SadIconContainer source={SadIcon} />
+              <Heading theme={theme}>{t('nothingFound')}</Heading>
+            </>
+          )}
```
klinzo, line 108: I think this maybe should be part of the search component directly as it is not directly related to the feedback but rather just the message appearing if there are no results.

native/src/modules/feedback/Feedback.js
```diff
@@ -46,47 +48,87 @@ const Description = styled(ThemedText)`
...
           </DescriptionContainer>
-          <Input theme={theme} onChangeText={props.onCommentChanged} value={comment} multiline numberOfLines={3} />
-
+          <Input
+            theme={theme}
+            onChangeText={props.onCommentChanged}
+            value={comment}
+            multiline
+            numberOfLines={3}
+            autoFocus={!isSearchFeedback}
```
klinzo, line 119: :+1:

native/src/modules/feedback/FeedbackContainer.js
```diff
@@ -0,0 +1,167 @@
+// @flow
```
klinzo, line 1: What changes have been made here? Sadly there is no nice git diff :/

klinzo: You just moved stuff from `FeedbackModalContainer` here, right?

klinzo (changes requested):
A relasenote would be nice I think.
So you are really picking the complicated PRs, right? Well done! I really like the behavior and design, did not find any major bugs or problems. Did not check the sent feedback to the cms though.

Minor observation: Part of the feedback is hidden by the keyboard (at least on android) again if there are no search results even though there would be enough space above. I think I fixed this in a PR not that long ago, could you check please?

translations/translations.json
SteffiStuffel: good point, forgot about checking usages in web. Done now.

native/src/routes/search/containers/SearchModalContainer.js
SteffiStuffel: For loading spinner while cities are not available and error handling if cities-fetch failed.

native/src/routes/search/components/SearchModal.js
SteffiStuffel: you are right. changed it.

native/src/routes/search/containers/SearchModalContainer.js
klinzo: Why do we need the cities?

SteffiStuffel: To get the cityName See FeedbackContainer (line 83ff)

  getCityName = (): string => {
    const { cities, cityCode } = this.props
    return CityModel.findCityName(cities, cityCode)
  }

native/src/routes/search/components/SearchModal.js
SteffiStuffel: I agree with you, that using everywhere the feedback:...-Prefix would be a lot more complicated. Also thought this would be the correct place because SearchModal is passing the t-prop and there seems to be responsible to set it correctly. But thinking a bit about it, the FeedbackContainer does not need a t-Function, only the Feedback.js needs it, therefore it can be moved to FeedbackContainer and the SearchModal and the FeebackModalContainer are no longer responsible for this.

native/src/modules/feedback/__tests__/Feedback.spec.js
```diff
@@ -0,0 +1,111 @@
...
+    const { getByText } = render(<Feedback {...buildProps('positive', '')} />)
+    expect(getByText('send')).not.toBeDisabled()
+  })
+
+  it('button should be enabled for negative Feedback and input', async () => {
+    const { getByText } = render(<Feedback {...buildProps('negative', 'comment')} />)
+    expect(getByText('send')).not.toBeDisabled()
+  })
+
+  it('correct Text should be displayed for positive Feedback and input', async () => {
```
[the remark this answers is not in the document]
SteffiStuffel: done

native/src/modules/feedback/__tests__/Feedback.spec.js
```diff
@@ -0,0 +1,111 @@
...
+      sendingStatus: 'idle',
+      onCommentChanged,
+      onFeedbackContactMailChanged,
+      onSubmit,
+      theme: lightTheme,
+      t
+    }
+  }
+
+  it('button should be disabled for negative Feedback and no input', async () => {
```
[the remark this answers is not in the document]
SteffiStuffel: yeah was like this before, changed it.

native/src/routes/search/containers/SearchModalContainer.js
klinzo: Oh... and the git blame says these are my changes... I am quite sure this is wrong and the city code should be used instead. I think I messed up while removing the feedback type here... AFAICS the city is only used to call the correct endpoint and we use the city code everywhere for that, so using the city name may even break stuff...Could you check? Sorry for that, but it probably makes things a lot easier, right?

native/src/modules/feedback/FeedbackContainer.js
SteffiStuffel: Yeah, would be nice to have the diff in git... :(

Before my changes:
- FeedbackModalContainer had functions for sending feedback of positive/negative feedback
- SearchModalContainer had the functions for sending feedback of search feedback
- FeedbackModal had UI for positive/negative feedback
- NothingFoundFeedbackBox had the UI for search feedback

Now:
- FeedbackContainer has functions for every feedback (positve/negative and search)
- Feedback has UI for every feedback
- FeedbackModalContainer is the page with FeedbackContainer as a child
- SearchModal is the page for the Search including FeedbackContainer

FeedbackContainer and Feedback are moved to modules-folder because they are shared now.
Logic and UI only exists once for both. 

FeedbackContainer has the functions from FeedbackModal Container and
- the FeedbackOriginType is added, so FeedbackContainer can also be used for Search-Feedback
- Search_Route-case is added to getFeedbackType-function
- getFeedbackData() is adjusted to work with FeedbackOriginType

Logic about the "page-stuff", like connect, refresh, mapStateToProps, withPayloadProvider... still are in FeedbackModalContainer.

native/src/modules/feedback/Feedback.js
SteffiStuffel: Makes sense. Moved it to SearchModal

native/src/modules/feedback/FeedbackContainer.js
klinzo: Thanks for summarizing!

native/src/modules/feedback/Feedback.js
SteffiStuffel: I think in terms of dependency inversion principle it would really make sense to keep it like this. Otherwise the high-level components would depend on the low level components. Also in case of single-responsibility it would make sense to have all the code concerning displaying the feedback-stuff to be in Feedback.js.

native/src/routes/search/containers/SearchModalContainer.js
SteffiStuffel: Yeah then we would not need withPayloadProvider here. For testumgebung it was working, but there cityCode and cityName are the same. But for other cities it is broken. I will remove it.

native/src/modules/feedback/Feedback.js
klinzo: I don't really agree. Passing a title as a prop is just a customization to the `Feedback` component the way React works by passing props to child components imo, e.g. like setting custom styles, formats or texts. I don't really see a dependency of higher to lower level components here as well. Maybe the point is just that I don't like these switch statements if they are not completely necessary. I just think it is much more straight forward to pass the title directly instead of having to introduce a new feedback origin type and adjusting the switch in this component here when maybe introducing a new feedack type in the future (Open-Closed Principle).

But probably this is just a matter of preference. @Taggotty @maxammann what is your opinion on this?

translations/translations.json
Taggotty: Why did you only remove the German translations?

release-notes/unreleased/IGAPP-559-additional-field-for-feedback-mail-in-search-feedback.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-559
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: You now have the opportunity to add your mail adress when giving feedback.
+de: Du hast nun die Möglichkeit deine E-Mail Adresse beim Geben von Feedback zu hinterlassen.
```
Taggotty, line 7: Imo `Geben` sounds pretty strange here.
```suggestion
de: Du hast nun die Möglichkeit deine E-Mail Adresse beim Senden von Feedback zu hinterlassen.
```

native/src/modules/feedback/Feedback.js
Taggotty: Probably the best way to conform to both Dependency Inversion and Open-Closed Principle would be to create a subclass for each FeedbackType, which is probably too complex in this context. 
Besides that I think it is cleaner to set the translations here, as they are in fact `feedback` translations, even though I also think that a switch case such as this is not the most flexible and inherently consistent approach (as the feedbackOrigin has to be conformant to isPositiveFeedback). Maybe instead of passing `feedbackOrigin` as a type you could infer the title by `isPositiveFeedback` and `routeType` instead. Maybe @maxammann also has an opinion on this 🤷🏼‍♀️

Taggotty (changes requested):

native/src/modules/feedback/FeedbackContainer.js
```diff
@@ -0,0 +1,162 @@
...
+      alias,
+      query
+    }
+  }
+
+  onFeedbackCommentChanged = (comment: string) => this.setState({ comment })
+
+  onFeedbackContactMailChanged = (contactMail: string) => this.setState({ contactMail })
+
+  handleSubmit = async () => {
```
maxammann, line 127: This could silently consume exceptions. It is better to do:

```
handleSubmit = () => {
  const request = async () => {
 ...  
  }
  
  request().catch(e => ...)
}
```

maxammann (approved):

klinzo (approved):

translations/translations.json
SteffiStuffel: I thought they are the "source", when adding translations we also add only german. Is it necessary to manually remove all languages?

native/src/modules/feedback/Feedback.js
SteffiStuffel: so i will keep it, as it is.

native/src/modules/feedback/FeedbackContainer.js
SteffiStuffel: Thanks for the hint, fixed it.

translations/translations.json
klinzo: I think so, yes.

native/src/modules/feedback/Feedback.js
SteffiStuffel: oke i kept the functionallity here, but removed switch case, instead using conditional rendering for title.

translations/translations.json
SteffiStuffel: done

native/src/modules/feedback/FeedbackContainer.js
```diff
@@ -0,0 +1,163 @@
...
+    const request = async () => {
+      const apiUrl = await determineApiUrl()
+      const feedbackEndpoint = createFeedbackEndpoint(apiUrl)
+      await feedbackEndpoint.request(feedbackData)
+      this.setState({ sendingStatus: 'successful' })
+    }
+    request().catch(err => {
+      console.log(err)
+      this.setState({ sendingStatus: 'failed' })
+    })
```
Taggotty, line 140: I really like the usage of Promise here instead of the async `handleSubmit` 👍🏼

native/src/modules/feedback/FeedbackContainer.js
```diff
@@ -0,0 +1,163 @@
...
+export type PropsType = {|
+  routeType: RouteType,
+  feedbackOrigin: FeedbackOriginType,
+  language: string,
+  cityCode: string,
+  cities: $ReadOnlyArray<CityModel>,
+  path?: string,
+  alias?: string,
+  query?: string,
+  offers?: Array<OfferModel>
```
Taggotty, line 73: cities and offers are not used anymore. Could you remove them here?

Taggotty (approved):

PR #254, opened 2021-04-16
feat[IGAPP-592] Replace Title & Content of Tunews DetailPage: Replace…

f1sh1918:
Replaced title & content with page component. Added sanitizer to prevent xss attacks.

https://issues.tuerantuer.org/browse/IGAPP-592

### DIFF
--- a/web/src/modules/common/components/Page.js
+++ b/web/src/modules/common/components/Page.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react'
-import { sanitize } from "dompurify";
+import { sanitize } from "dompurify"
 import styled, { type StyledComponent } from 'styled-components'
 import RemoteContent from './RemoteContent'
 import Caption from './Caption'
--- a/web/src/modules/common/components/Page.js
+++ b/web/src/modules/common/components/Page.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as React from 'react'
+import { sanitize } from "dompurify";
 import styled, { type StyledComponent } from 'styled-components'
 import RemoteContent from './RemoteContent'
 import Caption from './Caption'
@@ -53,7 +54,7 @@ const Page = ({
       {defaultThumbnailSrc && <Thumbnail alt='' src={defaultThumbnailSrc} srcSet={thumbnailSrcSet} />}
       <Caption title={title} />
       {children}
-      <RemoteContent dangerouslySetInnerHTML={{ __html: content }} onInternalLinkClick={onInternalLinkClick} />
+      <RemoteContent dangerouslySetInnerHTML={{ __html: sanitize(content) }} onInternalLinkClick={onInternalLinkClick} />
       {lastUpdate && (
         <LastUpdateInfo
           lastUpdate={lastUpdate}
--- a/web/src/routes/news/containers/TunewsDetailsPage.js
+++ b/web/src/routes/news/containers/TunewsDetailsPage.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as React from 'react'
+import { push } from 'redux-first-router'
 import styled, { type StyledComponent } from 'styled-components'
 import TunewsIcon from './../assets/TunewsActiveLogo.png'
 import { CityModel, NotFoundError, TunewsModel } from 'api-client'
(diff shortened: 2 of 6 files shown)


web/src/modules/common/components/Page.js
```diff
@@ -53,7 +54,7 @@ const Page = ({
       {defaultThumbnailSrc && <Thumbnail alt='' src={defaultThumbnailSrc} srcSet={thumbnailSrcSet} />}
       <Caption title={title} />
       {children}
-      <RemoteContent dangerouslySetInnerHTML={{ __html: content }} onInternalLinkClick={onInternalLinkClick} />
+      <RemoteContent dangerouslySetInnerHTML={{ __html: sanitize(content) }} onInternalLinkClick={onInternalLinkClick} />
```
klinzo, line 57: Nice :+1:

web/src/routes/news/containers/TunewsDetailsPage.js
```diff
@@ -74,6 +69,10 @@ export const TunewsDetailsPage = ({ tunewsElement, language, id, city, cities }:
   }
 
   const { title, content, date, eNewsNo } = tunewsElement
+  // Removes Last Update Timestamp in content
+  const filterPattern = `<p>${eNewsNo}</p>`
+  const filteredContent = content.replace(filterPattern,"")
```
klinzo, line 74: Sorry, I guess the issue was ambiguous here. It was meant as no dedicated timestamp should be displayed in the Page component, i.e. no last update should be passed to the page component as it is already included in the TüNews content as you noticed correctly. So I'd probably just render the HTML content 1:1 without modifying it like you did as it is external content and may be subject to changes we don't have any influence on in the future. Sorry for the misunderstanding here!

Or was there any other reason why you decided to do it this way?

web/src/routes/news/containers/TunewsDetailsPage.js
```diff
@@ -82,8 +81,15 @@ export const TunewsDetailsPage = ({ tunewsElement, language, id, city, cities }:
...
+          title={title}
+          content={filteredContent}
+          formatter={formatter}
+          lastUpdateFormat={lastUpdateDateFormat}
+          lastUpdate={date}
+          showLastUpdateText={false}
+          onInternalLinkClick={push}
+        />
       </StyledWrapper>
       <TunewsDetailsFooter eNewsNo={eNewsNo} date={date} formatter={formatter} />
```
klinzo, line 94: JFYI: The custom footer will be removed in a separate issue, just in case you were wondering why there is duplicated information here.

klinzo (commented):
Nicely done, thank you very much!

We have one small problem with the fork though, CI checks are not run. Not sure how we should handle this. @maxammann any ideas?

web/src/routes/news/containers/TunewsDetailsPage.js
f1sh1918: Okay i thought this eNewsNo tag should be removed from content.
Already removed the filter and updated pr.
I mean showLastUpdate text is already set to false, so it should be fine now?

maxammann:
> We have one small problem with the fork though, CI checks are not run. Not sure how we should handle this. @maxammann any ideas?

This is how it should be, else anyone could steal our secrety by foking and creating a PR. So there is no workaround. Before merging just push that branch to the Integreat/integeat-app repo

web/src/modules/common/components/Page.js
maxammann: We already sanitize content :sweat_smile: 

https://github.com/Integreat/integreat-app/blob/main/api-client/src/endpoints/createCategoriesEndpoint.js#L30

I kind of favor dompurify though. We should look into which library we should use though.

Implementing it in the api-client has the advantage that the native app is also automatically sanitized. Maybe we should create a separate task though for sanitisation again. Do we sanitize the news content for example?

maxammann (changes requested):
_

web/src/modules/common/components/Page.js
f1sh1918: oh :D i forgot to check all the deps in package.json.
But i def would recommend using a lib which is getting frequently updated, even there are lots of options regarding sanitizers.
Implementing in api client def makes sense maybe defining which fields can contain html content instead of regex testing

f1sh1918:
> > We have one small problem with the fork though, CI checks are not run. Not sure how we should handle this. @maxammann any ideas?
> 
> This is how it should be, else anyone could steal our secrety by foking and creating a PR. So there is no workaround. Before merging just push that branch to the Integreat/integeat-app repo

Fine. Gonna do that. But looks like i didn't get the access permission yet.

f1sh1918:
Create PR in the integreat repo

PR #255, opened 2021-04-17
IGAPP-592: Html support TüNews

f1sh1918:
Replaced title & content with page component. Added sanitizer to prevent xss attacks.

https://issues.tuerantuer.org/browse/IGAPP-592

This pull request belongs to an issue on our bugtracker.
It starts with the keyword IGAPP.


klinzo (approved):
Good work :muscle:

klinzo:
Builds are currently failing because of prettier. You could fix this by running `yarn prettier --write .`.

web/src/modules/common/constants/news.js
```diff
@@ -0,0 +1 @@
+export const lastUpdateDateFormat = 'LLL'
```
SteffiStuffel, line 1: think it would be good to add this file to flow checks.

SteffiStuffel (approved):
nice. :)

Taggotty (approved):
Looks fine to me 👍🏼

web/src/modules/common/components/Page.js
```diff
@@ -53,7 +54,10 @@ const Page = ({
       {defaultThumbnailSrc && <Thumbnail alt='' src={defaultThumbnailSrc} srcSet={thumbnailSrcSet} />}
       <Caption title={title} />
       {children}
-      <RemoteContent dangerouslySetInnerHTML={{ __html: content }} onInternalLinkClick={onInternalLinkClick} />
+      <RemoteContent
+        dangerouslySetInnerHTML={{ __html: sanitize(content) }}
```
maxammann, line 58: I can no longer see the previous discusison on this. Do we want to remove the sanitization in the api-client? I think we should avoid it here and do all sanitisation in the api-client. That way we easily support native and web. Also there are other fields which could require sanitisation.

I'm open for both, just sanitisating twice it a little bit overkill.

maxammann (approved):

web/src/modules/common/components/Page.js
klinzo: Removed it here and created a separate issue to check our currently sanitizing and add it/improve it wherever necessary,The cms team also mentioned that our current sanitizing is not good enough.
Issue: https://issues.tuerantuer.org/browse/IGAPP-619

PR #256, opened 2021-04-17
IGAPP-509: Tooltips are opened on click

maxammann:
…detection


web/src/modules/common/components/Tooltip.js
```diff
@@ -52,15 +26,15 @@ const pseudosMixin = (flow: FlowType, supportsLogicalProperties: boolean) => css
     `}
     ${flow === 'left' &&
     `
-      ${toLogicalProperty('border-right-width', supportsLogicalProperties)}: 0;
-      ${toLogicalProperty('border-left-color', supportsLogicalProperties)}: #333;
-      ${toLogicalProperty('left', supportsLogicalProperties)}: -5px;
+      border-right-width: 0;
+      border-left-color: #333;
```
klinzo, line 30: Shouldn't we use a color from our theme?

klinzo (approved):
Eazzzzzzyyy :thinking: Seems to work nicely :+1:
The spacing for header navigation item is a little bit off, but even though it is because of tooltips I'll create a separate issue as this was already the case before.

klinzo (changes requested):
Tested on Firefox and Chrome (Desktop) and Firefox (Mobile), seems like the tooltips are still opened on click on mobile.

Taggotty (approved):
Seems to work fine :)
However I wonder whether It is intentional, that longer text is cropped.
E.g. "Diese Seite ist nicht hilfr..."

SteffiStuffel (approved):

klinzo:
@maxammann you mentioned in the call yesterday that we decided to open tooltips on click on mobile anyway, I can't remember that. Do you have a link? I don't really like it tbh.

maxammann:
> @maxammann you mentioned in the call yesterday that we decided to open tooltips on click on mobile anyway, I can't remember that. Do you have a link? I don't really like it tbh.

Sadly I dont remember where, but tooltips work find on mobile. If we do not want to show them, then we would need a separate way of handling it on mobile. As they work fine on mobile I'd just keep them.

klinzo (approved):

klinzo:
I just find it weird that tooltips are opened on click on mobile, this is not the case on web as well, it only appears on hovering there. But I guess that is just my opinion as I don't know any websites having tooltips on mobile at all.

maxammann:
> Also: On small screens no tooltips are shown for the toolbar but tooltips are shown for the action items (search, location change, ...). Do you know why this is as it is? But this is also no new behavior, so nothing to fix here.

Looks fine for me, I already fixed some spacing there.

PR #257, opened 2021-04-19
IGAPP-562: Send signal if deep link is opened

zankerbe:


klinzo (approved):
Looking good! Tested on Android.

native/src/modules/navigation/__tests__/navigateToDeepLink.spec.js
```diff
@@ -440,6 +457,7 @@ describe('navigateToDeepLink', () => {
         undefined,
         false
       )
+      expect(sendTrackingSignal).toHaveBeenCalledTimes(1)
```
SteffiStuffel, line 460: Can you add a test to check if sendTrackingSignal is called with correct parameters?

SteffiStuffel (approved):

native/src/modules/navigation/__tests__/navigateToDeepLink.spec.js
zankerbe: Thanks for the feedback, i added checks for that.

PR #259, opened 2021-04-19
IGAPP-608: Remove nearby location setting

klinzo:


release-notes/unreleased/IGAPP-608-remove-nearby-locations-setting.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-588
+show_in_stores: false
+platforms:
+  - android
+  - ios
+en: The now unused setting proposeNearbyCities has been removed.
```
SteffiStuffel, line 6: I think this message should be understandable for users and why only in english?
Something with more account to what the user gets/sees like: You are no longer asked to allow access to your location, unless you use/activate specific features.

SteffiStuffel (changes requested):
When I open the app and go through the intro screens, for the last intro screen the button is named "Überspringen", should be something like "continue", because there are no longer any settings / permissions requested that can be "übersprungen" (jeeah denglish).

Also when selecting a city, i do not really think the "refresh" icon is an intutive symbol for using location. Is this a design or have you selected it?

release-notes/unreleased/IGAPP-608-remove-nearby-locations-setting.yml
klinzo: We don't show this message to any users but only include in the release notes posted on mattermost/github as `show_in_stores` is set to false, therefore only English is relevant. I think removing a settings option is not anything that is relevant to users and is more confusing than helping.

klinzo:
> When I open the app and go through the intro screens, for the last intro screen the button is named "Überspringen", should be something like "continue", because there are no longer any settings / permissions requested that can be "übersprungen" (jeeah denglish).
> 
> Also when selecting a city, i do not really think the "refresh" icon is an intutive symbol for using location. Is this a design or have you selected it?

Regarding `Überspringen` vs `Weiter`: Will fix that
Regarding the refresh icon: That icon was already there, imo it would be a separate task to find a more suitable design.

klinzo:
> > When I open the app and go through the intro screens, for the last intro screen the button is named "Überspringen", should be something like "continue", because there are no longer any settings / permissions requested that can be "übersprungen" (jeeah denglish).
> > Also when selecting a city, i do not really think the "refresh" icon is an intutive symbol for using location. Is this a design or have you selected it?
> 
> Regarding `Überspringen` vs `Weiter`: Will fix that
> Regarding the refresh icon: That icon was already there and used previously in the same place, imo it would be a separate task to find a more suitable design.

maxammann (approved):
I don't

SteffiStuffel (approved):

PR #260, opened 2021-04-20
IGAPP-378: remove rtl handling for list in second level content

Taggotty:

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -73,7 +73,7 @@
     "react-native-permissions": "3.0.1",
     "react-native-progress": "^4.0.3",
     "react-native-reanimated": "^2.0.0",
-    "react-native-render-html": "^5.1.0",
+    "react-native-render-html": "6.0.0-alpha.24",
     "react-native-safe-area-context": "^3.2.0",
     "react-native-safe-area-view": "2.0.0",
     "react-native-screens": "^2.18.1",
--- a/native/src/modules/categories/components/CategoryListContent.js
+++ b/native/src/modules/categories/components/CategoryListContent.js
@@ -76,8 +76,8 @@ const CategoryListContent = ({
           alterNode={alterResources}
           renderersProps={{
             a: { onPress: onLinkPress },
-            ol: { enableExperimentalRtl: true, direction: 'rtl' },
-            ul: { enableExperimentalRtl: true, direction: 'rtl' }
+            ol: { enableExperimentalRtl: true },
+            ul: { enableExperimentalRtl: true }
           }}
           baseFontStyle={{
             fontSize: 14,
--- a/native/src/modules/categories/components/CategoryListContent.js
+++ b/native/src/modules/categories/components/CategoryListContent.js
@@ -3,7 +3,7 @@
 import * as React from 'react'
 import { useCallback, useContext } from 'react'
 import { useWindowDimensions } from 'react-native'
-import RenderHTML, { GestureResponderEvent, type HTMLNode } from 'react-native-render-html'
+import HTML, { GestureResponderEvent, type HTMLNode } from 'react-native-render-html'
 import DateFormatterContext from '../../i18n/context/DateFormatterContext'
 import styled from 'styled-components/native'
 import type { ThemeType } from 'build-configs/ThemeType'
@@ -53,13 +53,13 @@ const CategoryListContent = ({
   const alterResources = useCallback(
     (node: HTMLNode) => {
       if (node.attribs) {
-        const newHref = node.attribs.href ? { href: cacheDictionary[decodeURI(node.attribs.href)] } : {}
-        const newSrc = node.attribs.src ? { src: cacheDictionary[decodeURI(node.attribs.src)] } : {}
+        const newHref = node.attribs.href && cacheDictionary[decodeURI(node.attribs.href)]
+        const newSrc = node.attribs.src && cacheDictionary[decodeURI(node.attribs.src)]
         if (newHref || newSrc) {
           node.attribs = {
             ...node.attribs,
-            ...newHref,
-            ...newSrc
+            ...(newHref && { href: newHref }),
+            ...(newSrc && { src: newSrc })
           }
           return node
         }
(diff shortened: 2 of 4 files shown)


native/package.json
```diff
@@ -73,7 +73,7 @@
     "react-native-permissions": "3.0.1",
     "react-native-progress": "^4.0.3",
     "react-native-reanimated": "^2.0.0",
-    "react-native-render-html": "^5.1.0",
+    "react-native-render-html": "6.0.0-alpha.24",
```
Taggotty, line 76: As you can see it is still an alpha version. If you prefer waiting for the major release I will delay this PR. When I was testing the renderer I didn't encounter any actual problems.

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -31,19 +32,14 @@ type ContentPropsType = {|
...
-  lastUpdate,
-  theme
-}: ContentPropsType) => {
+                               content,
+                               navigateToLink,
+                               cacheDictionary,
+                               language,
+                               lastUpdate,
+                               theme
+                             }: ContentPropsType) => {
```
klinzo, line 42: Indentation

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -57,124 +53,41 @@ const CategoryListContent = ({
...
-              src: newResource
-            }
-            return node
+        const newHref = node.attribs.href ?? cacheDictionary[decodeURI(node.attribs.href)]
+        const newSrc = node.attribs.src ?? cacheDictionary[decodeURI(node.attribs.src)]
+        if (newHref || newSrc) {
+          node.attribs = {
+            ...node.attribs,
+            ...newHref && { href: newHref },
+            ...newSrc && { src: newSrc },
```
klinzo, line 62: Are you sure this is doing what you expect it to do? It looks wrong to me, i.e. href is set if there was no href previously, same for src (at least afaict). I think the `??` operator is wrong, my suggestion would be (without having it tested):
```suggestion
        const newHref = node.attribs.href ? { href: cacheDictionary[decodeURI(node.attribs.href)] } : {}
        const newSrc = node.attribs.src ? { src: cacheDictionary[decodeURI(node.attribs.src)] }: {}
        node.attribs = {
          ...node.attribs,
          ...newHref,
          ...newSrc
```
Is there already a test that this works as expected?

native/src/modules/categories/components/CategoryListContent.js
```diff
@@ -57,124 +53,41 @@ const CategoryListContent = ({
...
             fontSize: 14,
             fontFamily: theme.fonts.native.contentFontRegular,
             color: theme.colors.textColor
           }}
           tagsStyles={{
-            p: { textAlign: config.hasRTLScript(language) ? 'right' : 'left' },
-            img: { align: config.hasRTLScript(language) ? 'right' : 'left' }
+            ul: { direction: config.hasRTLScript(language) ? 'rtl' : 'ltr' },
+            ol: { direction: config.hasRTLScript(language) ? 'rtl' : 'ltr' },
+            p: { textAlign: contentAlignment(language) }
```
klinzo, line 90: The `align` property for images is not needed anymore?

klinzo (commented):
I think if we test this really carefully we maybe can already merge this. We just have to keep in mind that we won't be able to do much bug fixes during refactoring in may, so it may make sense to move this to afterwards, but not really sure about this.

At the least we should probably include a development release note in order for our testers to test this really carefully as well.

klinzo:
According to this it seems like a beta/stable release is not really far away as just? docs and code coverage is missing:
https://github.com/meliorence/react-native-render-html/pull/434#issue-527747509

native/src/modules/categories/components/CategoryListContent.js
Taggotty: The align property for images never actually existed or is not supported anymore. I think you can still set the alignment with proper html container and the text align property, but that's nothing we should need to worry about.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: This

> According to this it seems like a beta/stable release is not really far away as just? docs and code coverage is missing:
> [meliorence/react-native-render-html#434 (comment)](https://github.com/meliorence/react-native-render-html/pull/434#issue-527747509)

That's also what I expect. However this could mean 1 month or 3 month still. Therefore this PR proposal.

Taggotty:
This

> According to this it seems like a beta/stable release is not really far away as just? docs and code coverage is missing:
> [meliorence/react-native-render-html#434 (comment)](https://github.com/meliorence/react-native-render-html/pull/434#issue-527747509)

That's also what I expect. However this could mean 1 month or 3 month still. Therefore this PR proposal.

native/src/modules/categories/components/CategoryListContent.js
Taggotty: There is a test for navigating to a link. Anything else is not typically done by react-testing-library (such as checking html props)

Taggotty: Still I had to replace the `??` with `&&` to make it work

release-notes/unreleased/IGAPP-378-remove-rtl-workaround.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-378
+show_in_stores: false
+platforms:
+  - android
+  - ios
+en: Second level content has now proper rtl/ltr support in lists.
+de: Inhalte auf zweitem Level unterstützen rtl/ltr nun nativ in Listen.
```
SteffiStuffel, line 7:
```suggestion
de: Inhalte auf zweitem Level unterstützen rtl/ltr nun in Listen.
```

SteffiStuffel (approved):

native/package.json
klinzo: Not urgent at all, right @Taggotty? More of a refactoring issue of removing our custom workarounds.
So I think I tend to agree with @SteffiStuffel especially because of our refactoring in two weeks.

Taggotty: That's fine to me. Then I will close this issue for now and keep the branch locally.

PR #261, opened 2021-04-21
IGAPP-550: Add georgian and georgian fonts

klinzo:


klinzo:
Testable here: https://malteapp.de/nrw/ka
Fonts only added for web.

Taggotty (approved):
Looks fine 👍🏼

release-notes/unreleased/IGAPP-550-add-georgian.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-550
+show_in_stores: false
+platforms:
+  - web
+  - android
+  - ios
+en: Support added for georgian language and alphabet.
```
SteffiStuffel, line 7: I think it is only for web, isn't it?
Why no german release note?

SteffiStuffel (approved):
It does not render the correct font when i start the app?

Bildschirmfoto zu 2021-04-22 12-10-18

release-notes/unreleased/IGAPP-550-add-georgian.yml
klinzo: See answer in other PR, if `show_in_stores` is set to true, it is not shown to users and therefore no german version is needed. I don't want to show this to users as there is only one city (and that is not even live and a malteser one) that uses this new language and as a user I would really be confused why this language is not available in my city anyway.

Regarding only web: It is native as well, we just dont have an additonal font there. But the language (should) work on native as well, which it did not previously to this issue.

klinzo:
@SteffiStuffel I first thought it is because raleway (our decorative font) does not support Georgian, which is the case actually. But it uses Noto for me anyway? Could you refresh and try again please?
image

Taggotty:
> @SteffiStuffel I first thought it is because raleway (our decorative font) does not support Georgian, which is the case actually. But it uses Noto for me anyway? Could you refresh and try again please?

You forgot to add the font to `I18nProvider.js`. After adding it works for me :)

klinzo:
> > @SteffiStuffel I first thought it is because raleway (our decorative font) does not support Georgian, which is the case actually. But it uses Noto for me anyway? Could you refresh and try again please?
> 
> You forgot to add the font to `I18nProvider.js`. After adding it works for me :)

Thanks, did not know I had to add it there. Do you know by any chance why it was working for me anyway?

Taggotty:
> Do you know by any chance why it was working for me anyway? 

We still set the font-family to Noto-Sans-Georgian in Layout.js . Therefore, I think the browser checks whether the System can provide this font and if yes it will use this one. If you by by any chance have the font installed on the system it will use this one, I guess.

PR #262, opened 2021-04-21
IGAPP-552: Add dari

klinzo:

### DIFF
--- a/release-notes/unreleased/IGAPP-552-add-dari.yml
+++ b/release-notes/unreleased/IGAPP-552-add-dari.yml
@@ -1,0 +1,7 @@
+issue_key: IGAPP-550
+show_in_stores: false
+platforms:
+  - web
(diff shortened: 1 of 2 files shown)


SteffiStuffel (approved):
Still do not get how these language adding is working and do not know where to test, so just drop my "approve" with no value^^

klinzo:
> Still do not get how these language adding is working and do not know where to test, so just drop my "approve" with no value^^

If the fonts are already there, it's just about adding it to the config and setting the right direction attribute. On native, the app crashes if the language is not included in the config, so it is enough to run the app, choose the new language and if everything works fine then we are good to go.

Dari you can test e.g. here: https://malteapp.de/nrw/prs

Taggotty (approved):
Tested on Web and Android Emulator.

PR #263, opened 2021-04-21
IGAPP-615: Replace dashes in keys

klinzo:


SteffiStuffel (approved):
Should we - in long term - think aboud changing the keyes everywhere like this?

klinzo:
> Should we - in long term - think aboud changing the keyes everywhere like this?

Yeah maybe, at least for everything not directly in js it may make sense. I have to look into how we can solve this for iOS/XCode anyway as we have the same problem there.

Taggotty (approved):

Taggotty (approved):

PR #264, opened 2021-04-21
IGAPP-593: Remove permission snackbar and fix iOS PN permission handling

klinzo:


native/src/routes/settings/components/SettingItem.js
```diff
@@ -1,24 +1,29 @@
 // @flow
```
klinzo, line 1: Refactoring only, moved some functionality from `Settings` here: badges and switches

native/src/routes/settings/components/Settings.js
```diff
@@ -1,16 +1,15 @@
 // @flow
```
klinzo, line 1: Refactoring only, moved some functionality to `SettingsItems` here: badges and switches

native/src/routes/settings/__tests__/createSettingsSections.spec.js
```diff
@@ -0,0 +1,183 @@
+// @flow
```
SteffiStuffel, line 1: nice! Go for the OKR :rocket:

SteffiStuffel (approved):
Did regression test for settings page.

native/src/routes/settings/components/Settings.js
```diff
@@ -56,135 +50,97 @@ const SectionHeader = styled.Text`
...
 
-    this.unsubscribeNavigationListener = this.props.navigation.addListener('focus', () => {
-      this.loadSettings()
+  useEffect(() => {
+    // Reload settings if navigating back from another route
+    return navigation.addListener('focus', () => {
+      loadSettings()
     })
-  }
+  }, [navigation])
```
Taggotty, line 67: No need to remove the listener on unmount?
Any reason why the [useFocusEffect](https://reactnavigation.org/docs/function-after-focusing-screen/#triggering-an-action-with-the-usefocuseffect-hook) hook doesn't work here?

klinzo: It is removed on unmount, `navigation.addListener` directly returns the function to remove the listener as mentioned here: https://reactnavigation.org/docs/navigation-events/#navigationaddlistener
`useFocusEffect` seems fine, I just didn't look into refactoring this even more and also did not know about this one. Will try it :+1:

klinzo: Works nicely, thanks!

PR #268, opened 2021-04-25
IGAPP-610: Weird language item spacing if some are not available

maxammann:
… text when hovering


maxammann:
This PR also fixes an issue with the hover effect. If you hover an element it becomes bold and increases the width. The BoldSpacer fixes this jumping.

maxammann:
I also increased the max width of tooltip content

SteffiStuffel (approved):

SteffiStuffel:
Tested on Chrome

web/src/modules/common/components/Selector.js
```diff
@@ -30,22 +30,25 @@ const Element: StyledComponent<{||}, ThemeType, *> = styled(Link)`
     font-size: 1em;
     line-height: ${dimensions.headerHeightSmall}px;
   }
-`
+  
+  ${props =>
+    props.enabled ? `color: ${props.theme.colors.textColor};` : `color: ${props.theme.colors.textDisabledColor};`}
```
klinzo, line 35:
```suggestion
  color: ${props => props.enabled ? props.theme.colors.textColor : props.theme.colors.textDisabledColor};
```

web/src/modules/common/components/Selector.js
```diff
@@ -30,22 +30,25 @@ const Element: StyledComponent<{||}, ThemeType, *> = styled(Link)`
...
-const ActiveElement: StyledComponent<{| selected: boolean |}, ThemeType, *> = styled(Element)`
-  color: ${props => props.theme.colors.textColor};
   ${props =>
     props.selected
       ? 'font-weight: 700;'
       : `:hover {
           font-weight: 700;
           border-radius: 0;
         }`}
+}
```
klinzo, line 44: I think this bracket is to much:
```suggestion
```

web/src/modules/common/components/Tooltip.js
```diff
@@ -152,13 +152,16 @@ const TooltipContainer: StyledComponent<
...
+type PropsType = {
   children: React.Node,
   text: ?string,
   flow: FlowType,
   mediumViewportFlow?: FlowType,
-  smallViewportFlow?: FlowType
-|}
+  smallViewportFlow?: FlowType,
+  className?: string
+}
```
klinzo, line 164:
```suggestion
  className?: string,
  ...
}
```

web/src/modules/common/components/Tooltip.js
```diff
@@ -152,13 +152,16 @@ const TooltipContainer: StyledComponent<
   }
 `
 
-type PropsType = {|
+// We do not want exact types here as we forward props
+// eslint-disable-next-line flowtype/require-exact-type
```
klinzo, line 156:
```suggestion
```

klinzo (approved):
I think you could add a release note for this :)

One minor observation: The tooltip for chinese (http://0.0.0.0:9000/testumgebung/de/news/local/157) was hidden behind the langauage change icon for me once, but I was not able to reproduce it again :/ So I think you can just ignore it.

Taggotty (changes requested):
I get the following error in the web console: `Warning: Received 'true' for a non-boolean attribute 'enabled'.`

klinzo:
> I get the following error in the web console: `Warning: Received 'true' for a non-boolean attribute 'enabled'.`

I think we had this already, the reason are styled component props if I remember corectly: https://styled-components.com/docs/api#transient-props

maxammann:
@klinzo I imporoved the positioning of tooltips

klinzo (approved):
Tested, looks really good!

Taggotty (approved):
For the language tooltip I still get an overflow on the right side. Personally I think that is no big problem, but if you find a quick fix I am also content :)

maxammann:
> For the language tooltip I still get an overflow on the right side. Personally I think that is no big problem, but if you find a quick fix I am also content :)


Fixed, missed to remove a `/2`

PR #269, opened 2021-04-28
IGAPP-604: Single page endpoints

klinzo:


web/.idea/codeStyles/Project.xml
```diff
@@ -2,20 +2,42 @@
   <code_scheme name="Project" version="173">
```
klinzo, line 2: Code style imported from prettier, if you don't like it I could revert it.

Taggotty (approved):
The tests look much cleaner than before. 👍🏼

api-client/src/endpoints/createCategoryParentsEndpoint.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+      return `${baseUrl}/${city}/${language}/wp-json/extensions/v3/parents?&url=${cityContentPath}`
+    })
+    .withMapper((json: Array<JsonCategoryType>, params: ParamsType): Array<CategoryModel> => {
+      const basePath = `/${params.city}/${params.language}`
+
+      const parents = json.map(category => mapCategoryJson(category, basePath))
+
+      parents.push(
+        new CategoryModel({
```
maxammann, line 30: I'm not sure whether it is wise to craft these weird CaregoryModels. We could also just make this endpoint return an array of paths. What do you think?

maxammann (approved):

api-client/src/endpoints/createCategoryParentsEndpoint.js
klinzo: We'd need at least the name and the path though to show breadcrumbs. We could do that though,yes. Or we just don't include the root at all. Not sure about that. On the other hand this just follows the behavior of the already existing categories endpoint, which is not really optimal though. So not sure.

klinzo: I think I'd leave this for now, we can refactor this at a later point of time together with the other endpoints using these weird models.

PR #270, opened 2021-04-28
IGAPP-566: Pass system language for signals

klinzo:

Any better ideas on how to handle this?


native/src/modules/i18n/components/I18nProvider.js
```diff
@@ -16,6 +16,8 @@ import DateFormatter from 'api-client/src/i18n/DateFormatter'
 
 type PropsType = {| children: React.Node |}
 
+export let i18n
```
maxammann, line 19: This is quite hacky. This introduces a global state. The `sendTrackingSignal` wont be notified about changes in languages. There is no guarantee that this will continue to work. For example I18nProvider is able to discard the old instance of i18n and create a new one.

I suppose you did it that way such that we do not have to explicitly pass the language to `sendTrackingSignal`. I think it would be oke to do some hack here as signal reporting is not that important compared to core functionalities of the app.

I think we should invert the way it works though in this implementation. We could keep the state of the current language in `sendTrackingSignal` instead of the I18nProvider. The render function of I18nProvider would then call the function `sendTrackingSignal.setLanguage(language);`. This can even be called in a callback which is called each time the language is changed. The `sendTrackingSignal` module keeps a global state object `{language: 'de'}`. 

This is better because we can change the sent language in signal and do not depend that there is only one i18n instance.

maxammann (approved):

native/src/modules/i18n/components/I18nProvider.js
klinzo: Nice suggestion, thanks. Way better this way! The callback isn't even needed, the system language is already set correctly after system language changes.

PR #271, opened 2021-04-28
IGAPP-573: Save and send offline signals later

klinzo:


native/src/modules/endpoint/sendTrackingSignal.js
```diff
@@ -5,38 +5,58 @@ import AppSettings from '../settings/AppSettings'
...
+const sendSpecificSignal = async ({
+  signal: specificSignal,
+  offline = false
+}: {|
+  signal: SpecificSignalType,
+  offline?: boolean
+|}) => {
+  const appSettings = new AppSettings()
+  try {
+    const settings: SettingsType = await appSettings.loadSettings()
```
Taggotty, line 39: Do you really need this type?
```suggestion
    const settings = await appSettings.loadSettings()
```

Taggotty (approved):
The naming of this PR seems a little bit off.

Not tested.

klinzo:
> The naming of this PR seems a little bit off.
> 
> Not tested.

Copied from the wrong issue :grin:

native/src/modules/endpoint/sendTrackingSignal.js
```diff
@@ -5,38 +5,58 @@ import AppSettings from '../settings/AppSettings'
 import moment from 'moment'
 import type { SettingsType } from '../settings/AppSettings'
 import buildConfig from '../app/constants/buildConfig'
+import ErrorCodes, { fromError } from '../error/ErrorCodes'
+import type { SignalType } from 'api-client'
 
-const sendTrackingSignal = async ({ signal, offline = false }: {| signal: SpecificSignalType, offline?: boolean |}) => {
+export const sendCompleteSignal = async (signal: SignalType) => {
```
maxammann, line 11: Not sure what this name means, are there incomplete signals? Mabye `sendRequest` would be better and below just continue to use `sendTrackingSignal` instead of `sendSpecificSignal`

maxammann (approved):

PR #276, opened 2021-05-03
IGAPP-614: Adjust docs for service account permissions

SteffiStuffel:


docs/whitelabelling.md
```diff
@@ -88,6 +88,7 @@ Things the AppTeam has to do and therefore probably create issues for.
 
 - Add build config
 - Setup app store presence
+- Add new App to permissions of ci service account
```
klinzo, line 91: Is this google play specific? Or applies for apple as well?

klinzo (approved):

docs/whitelabelling.md
SteffiStuffel: only for google play.

Taggotty (approved):

docs/whitelabelling.md
klinzo:
```suggestion
- Add new App to permissions of GooglePlay CI service account  
```

PR #277, opened 2021-05-04
IGAPP-594: Sort by name and prefix

Taggotty:

"Landkreis sollte immer vor Stadt sein."

I will create another issue for native as it is also inconsistent there.


web/src/routes/landing/components/CitySelector.js
```diff
@@ -49,8 +49,17 @@ class CitySelector extends React.PureComponent<PropsType> {
     }
   }
 
+  // Landkreis should come before Stadt
+  sort(cities: Array<CityModel>): Array<CityModel> {
+    return cities.sort(
+      (a, b) =>
+        a.sortingName.localeCompare(b.sortingName) || (a.prefix && b.prefix ? a.prefix.localeCompare(b.prefix) : 0)
```
klinzo, line 56: This does not handle comparison of prefix and no prefix, right? I.e. what would happen if there would be `Augsburg` and `Stadt Augsburg`. How about
```suggestion
        a.sortingName.localeCompare(b.sortingName) || (a.prefix || '').localeCompare(b.prefix || ''))
```
I think that should handle all cases correctly.

klinzo (commented):
So we did not sort at all before? :astonished:

klinzo (approved):
Tested on firefox.

web/src/routes/landing/components/CitySelector.js
Taggotty: Great idea. Didn't think about this

Taggotty:
> So we did not sort at all before? astonished

At least I didn't find any sorting. Maybe somewhere in the backend 🤷🏼‍♀️

SteffiStuffel (approved):
Tested on chrome. :+1:

PR #278, opened 2021-05-04
IGAPP-564: Send signal on feedback

Taggotty:


native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,81 +1,21 @@
...
-      cities={props.cities}
-    />
+    <FeedbackContainer routeType={routeType} feedbackOrigin={feedbackOrigin} language={language} cityCode={cityCode} />
   )
 }
 
-export default connect<PropsType, OwnPropsType, _, _, _, _>(
-  mapStateToProps,
-  mapDispatchToProps
-)(withPayloadProvider<InnerPropsType, OwnPropsType, FeedbackModalRouteType>(refresh)(FeedbackModalContainer))
```
Taggotty, line 81: Am I correct here? Without cities, we don't need `connect` and `wihtPayloadProvider`, right?

SteffiStuffel: yes thats right. We had a own ticket for it IGAPP-612

SteffiStuffel (approved):
Tested on Android.

klinzo (approved):
Tested on Android :+1:

SteffiStuffel:
Not sure if it is supposed to be like this, but no information is send where the feedback was send (offers, categories, search.. ) but we send signals for every page opened so it should be oke, i guess?

PR #281, opened 2021-05-04
IGAPP-612: Refactor feedback

klinzo:


native/src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -1,81 +1,30 @@
...
-const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>): DispatchPropsType => ({ dispatch })
-
-const refresh = (refreshProps: OwnPropsType) => {
-  const { navigation } = refreshProps
-  const navigateToFeedback = createNavigateToFeedbackModal(navigation)
-  navigateToFeedback(refreshProps.route.params)
-}
-
-const FeedbackModalContainer = (props: InnerPropsType) => {
+const FeedbackModalContainer = (props: PropsType) => {
```
klinzo, line 14: I really like functional components :heart_eyes:

native/src/routes/search/containers/SearchModalContainer.js
```diff
@@ -1,92 +1,28 @@
...
+import { useSelector, useDispatch } from 'react-redux'
 
-type OwnPropsType = {|
+export type PropsType = {|
   route: RoutePropType<SearchRouteType>,
   navigation: NavigationPropType<SearchRouteType>
 |}
 
-type DispatchPropsType = {| dispatch: Dispatch<StoreActionType> |}
+const SearchModalContainer = ({ navigation }: PropsType) => {
```
klinzo, line 18: I really like functional components :heart_eyes:

klinzo:
I think I tested everything, but please try to give feedback and to search a little bit to make sure this still works as expected. Possibly also with an eye on the feedback tracking signals implemented by @Taggotty in #278 .

native/src/modules/feedback/Feedback.js
```diff
@@ -57,28 +57,26 @@ export type PropsType = {|
   sendingStatus: SendingStatusType,
   onCommentChanged: (comment: string) => void,
   onFeedbackContactMailChanged: (contactMail: string) => void,
-  feedbackOrigin: FeedbackOriginType,
+  isSearchFeedback: boolean,
+  isPositiveFeedback: boolean,
```
SteffiStuffel, line 61: So we have no less information now? The info if there was nothing found or not enough found got lost, right? 

I think we should keep this functionality, because i think for the future it would be nice to get as much information as possible, so cities can optimize their content and Integreat has more impact.

SteffiStuffel (approved):
Tested on Android.

Not sure if this is supposed, but no matter on which Category i give Feedback, it is always listed as "Hauptmenü / Integreat Allgemein". It would really make sense to have the information which page/information the user liked/disliked, wouldn't it?

And i would recommend to use a email-keyboard type for the email-field:
https://reactnative.dev/docs/textinput#keyboardtype

Taggotty (approved):

klinzo:
> Tested on Android.
> 
> Not sure if this is supposed, but no matter on which Category i give Feedback, it is always listed as "Hauptmenü / Integreat Allgemein". It would really make sense to have the information which page/information the user liked/disliked, wouldn't it?
> 
> And i would recommend to use a email-keyboard type for the email-field:
> https://reactnative.dev/docs/textinput#keyboardtype

@SteffiStuffel fixed and done. Really good catch. This was broken before already though :/ Looks like we missed it in #252.

PR #284, opened 2021-05-11
[PROPOSAL] IGAPP-595: Pass build config name by resolving via metro config

klinzo:

Definitely quite hacky, got the idea from here:https://stackoverflow.com/a/60099655
It works like a charm though. I like it better than having to use babel just to pass one single environment variable.


klinzo:
Rejected in favor of #285.

PR #287, opened 2021-05-18
IGAPP-605: Hooks for endpoints

klinzo:

This PR allows creating hooks in the api-client as well as using them in the native and web-ts project. The `useLoadFromEndpoint` hook was moved from native to the api-client.


api-client/src/stringifyFormData.ts
```diff
@@ -1,14 +1,14 @@
...
 
   if (formData?.getParts) {
     for (const part of formData.getParts()) {
-      entries[part.fieldName] = part.string
+      entries[part.fieldName as string] = part.string
     }
   } else {
     for (const [key, value] of formData.entries()) {
-      entries[key] = value
+      entries[key as string] = value
```
Taggotty, line 11: We probably should add tests for this. Especially as the types are a little bit hacky.

Taggotty (approved):

api-client/src/stringifyFormData.ts
klinzo: This is only used for showing a nice error, so I don't think this is critical. Maybe we could just use `JSON.stringify()` as well instead of this?

Taggotty: Sure, it is not critical. Imo just JSON.stringify would be better, especially as we just print this message to the console right?

klinzo: Replaced with JSON.stringify()

api-client/src/endpoints/hooks/useLoadFromEndpoint.ts
```diff
@@ -1,24 +1,24 @@
...
-  setLoading: (arg0: boolean) => void
-) => {
+  setData: (data: T | null) => void,
+  setError: (error: Error | null) => void,
+  setLoading: (loading: boolean) => void
+): Promise<void> => {
   setLoading(true)
 
   try {
-    const apiUrl = await determineApiUrl()
```
SteffiStuffel, line 14: We used determineApiUrl for choosing test-cms or cms. Is this no longer needed? Or where gets it checked now?

package.json
```diff
@@ -30,7 +30,8 @@
     "test-ts": "rm -rf */dist && tsc --build web-ts native"
   },
   "resolutions": {
-    "moment": "2.29.1"
+    "moment": "2.29.1",
+    "react": "16.13.1"
```
SteffiStuffel, line 34: Why is this necessary? Would really love it to be removed, if possible.

api-client/src/endpoints/hooks/useLoadFromEndpoint.ts
SteffiStuffel: Oke I see it is in the Containers now. I think it would be a lot nicer if it would be central and only in one place. A container should not be in charge for determining the correct cms-url, should it? Is there a better place to put it?

SteffiStuffel (approved):

api-client/src/endpoints/hooks/useLoadFromEndpoint.ts
klinzo: I moved `useLoadFromEndpoint` to the api-client. Therefore it is not possible to keep this functionality in the `useLoadFromEndpoint` hook. In theory it would be possible to create another wrapper around it in the native project though. I would not do this now already though.

package.json
klinzo: It is necessary to define hooks in the api-client.

api-client/src/stringifyFormData.ts
```diff
@@ -1,16 +0,0 @@
-// stringifies FormData from react-native or the browser variant of it
```
maxammann, line 1: Are we sure this if fine to just use JSON.stringify?

PR #289, opened 2021-05-19
IGAPP-639: Fix issues with push notifications

klinzo:

This PR should fix the following things:
    * The abapp does not start
    * The app says push notifications don't work on iOS and android with google play services
    * There is a weird jumping behavior of the settings item

I removed the wrong check whether google play services are available as I misread the docu and the check was just checking whether the Google Play Store is available and not the play services. There is no easy way to detect missing play services/firebase otherwise and since it is an edge case I removed it again completely.


SteffiStuffel (approved):
Tested on android.

PR #291, opened 2021-05-19
IGAPP-641: Move common module to web-ts

klinzo:


web-ts/src/components/Selector.tsx
```diff
@@ -88,9 +84,10 @@ const Selector = ({ items, activeItemCode, verticalLayout, closeDropDown, disabl
           return (
             <Element
               key={item.code}
-              as={Link}
+              // TODO As link
+              // as={Link}
+              // to={item.href}
```
Taggotty, line 89: Why is this not yet working?

web-ts/src/components/__tests__/ScrollingSearchBox.spec.tsx
```diff
@@ -56,6 +55,7 @@ describe('ScrollingSearchBox', () => {
     ).find(ScrollingSearchBox)
 
     component.find(SearchInput).prop('onFilterTextChange')('test')
+    // @ts-ignore
     component.find(Headroom).prop('onStickyTopChanged')(5)
```
Taggotty, line 59: Why ts-ignore?

Taggotty (approved):
Did you now use git move or not?
Some files actually preserved their history and some didn't. I personally don't really mind ^^

klinzo:
> Did you now use git move or not?
> Some files actually preserved their history and some didn't. I personally don't really mind ^^

I did not use git move. I guess if there are too many changes, git does not recognize it as the same file anymore. I just did the same we did in native, first moving and renaming, commit and then changes and commit again.

web-ts/src/components/FailureSwitcher.tsx
```diff
@@ -1,73 +1,71 @@
...
-  static renderContentNotFoundComponent(error: NotFoundError): React.Node {
+  static renderContentNotFoundComponent(error: NotFoundError): ReactNode {
     const { city, language } = error
     switch (error.type) {
       case 'category':
         return (
           <Failure
-            goToPath={new CategoriesRouteConfig().getRoutePath({ city, language })}
+            // TODO use right redirect path
+            // goToPath={new CategoriesRouteConfig().getRoutePath({ city, language })}
```
maxammann, line 23: Is there a ticket for this?

maxammann (approved):

web-ts/src/components/FailureSwitcher.tsx
klinzo: Not yet but we have to check all todos afterwards anyway.

PR #293, opened 2021-05-21
IGAPP-661: Move jest config

Taggotty:


web-ts/jest.config.ts
```diff
@@ -24,5 +27,10 @@ module.exports = {
       }
     ]
   ],
-  snapshotSerializers: ['enzyme-to-json/serializer']
+  snapshotSerializers: ['enzyme-to-json/serializer'],
+  globals: {
+    __BUILD_CONFIG__: webIntegreatTestCmsBuildConfig
+  }
```
Taggotty, line 33: as `__BUILD_CONFIG__` is a const we cannot (at least not without typescript error)  override it in the `jest.setup.ts` another solution is to override it right here. This will not throw any error (as jest probably does the parsing) and works perfectly fine (which means I tested it 👍🏼 )

maxammann (approved):

maxammann:
I think the package.json still contains a wrong config name as it ends now with .ts

web-ts/jest.setup.ts
```diff
@@ -28,6 +24,7 @@ function walkDir(dir: string, callback: (dir: string) => void) {
 // modules automatically as soon as they are found
 const mocksPath = 'src/__mocks__/'
 const jsPath = '.js' // This only unmocks .js files not .json for example
+
 walkDir(mocksPath, name => {
   if (name.endsWith(jsPath)) {
     jest.unmock(name.substring(mocksPath.length, name.length - jsPath.length))
```
klinzo, line 30: This should be adjusted the same way as in the native project. Otherwise jest will automatically mock everything with a file in __mocks__

web-ts/src/__mocks__/redux-first-router-link.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
```
klinzo, line 1: Delete, we won't use redux-first-router (-link) anymore.

klinzo (approved):

web-ts/src/components/__tests__/Failure.spec.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { Failure } from '../Failure'
+import { MemoryRouter } from 'react-router-dom'
 
 describe('Failure', () => {
   const mockTranslate = key => key
 
   it('should render a simple failure and match snapshot', () => {
     const errorMessage = 'error message'
-    const { getByText } = render(<Failure errorMessage={errorMessage} t={mockTranslate} />)
+    const { getByText } = render(<Failure errorMessage={errorMessage} t={mockTranslate} />, { wrapper: MemoryRouter })
```
klinzo, line 11: Why is this needed?

web-ts/src/components/__tests__/FailureSwitcher.spec.tsx
```diff
@@ -31,19 +30,19 @@ describe('FailureSwitcher', () => {
     it('should call render content not found component and create a Failure component', () => {
       const error = new NotFoundError({ type, id, language, city })
       const spy = jest.spyOn(FailureSwitcher, 'renderContentNotFoundComponent')
```
klinzo, line 32: I think we could remove this, can't we? The check seems quite unnecessary.

klinzo (approved):

PR #296, opened 2021-05-24
IGAPP-626: Sort by name and prefix

zankerbe:


Taggotty (approved):

klinzo (approved):

klinzo:
Tested :+1:

native/src/routes/landing/components/CitySelector.tsx
```diff
@@ -66,8 +66,16 @@ class CitySelector extends React.PureComponent<PropsType> {
     }
   }
 
+  // Landkreis should come before Stadt
+  _sort(cities: Array<CityModel>): Array<CityModel> {
+    return cities.sort(
+      (a, b) => a.sortingName.localeCompare(b.sortingName) || (a.prefix || '').localeCompare(b.prefix || '')
```
maxammann, line 72: This involves quite some JS knowledge :D such that 0 evaluates to false. But I dont know a simler yet expressive way.

Maybe just replace `||` with `??` in the last two cases.

maxammann (approved):

native/src/routes/landing/components/CitySelector.tsx
klinzo: Would not make a difference as the only case where `??` differs from `||` is empty strings, which are used as default anyway.

PR #298, opened 2021-05-26
IGAPP 634: Replace image params with props-object to prevent ts error

Znippet:

### DIFF
--- a/native/src/modules/common/components/ProgressSpinner.tsx
+++ b/native/src/modules/common/components/ProgressSpinner.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import { TFunction } from 'react-i18next'
 import { Dimensions } from 'react-native'
-import Svg, { Circle, Image } from 'react-native-svg'
+import Svg, { Circle, Image, G } from 'react-native-svg'
 import styled from 'styled-components/native'
 import { buildConfigAssets } from '../../app/constants/buildConfig'
 import { ThemeType } from '../../theme/constants'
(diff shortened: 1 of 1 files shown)


native/src/modules/common/components/ProgressSpinner.tsx
```diff
@@ -38,17 +38,16 @@ export type PropsType = {
 class ProgressSpinner extends React.Component<PropsType> {
   render() {
     const { t, progress, theme } = this.props
+    const imageProps = { 
+      width: logoSize,
+      height: logoSize,
+      transform: `translate(${logoXY}, ${logoXY})`,
+      xlinkHref: buildConfigAssets().loadingImage
+    }
```
Taggotty, line 46: Isn't this basically just avoiding the issue? Does transform do anything?

klinzo: I don't even get why ts does not throw an error anyway tbh :D

Znippet: Yes, it's only a workaround. I tried to add
declare module '*.svg' {
  import { SvgProps } from "react-native-svg";
  const content: React.FC<SvgProps>;
  export default content;
}
to assets.d.ts to have the types, but that didn't help.

I wouldn't replace transform because as I saw it is a standard svg action.

But I will have another look how we can convince typescript that transform exiists..

Znippet: And yes, it only does a little bit of centering, perhaps we could also find another way

klinzo: It does seem like it does not exist as prop: https://github.com/react-native-svg/react-native-svg/blob/develop/src/elements/Image.tsx. Maybe you can just use custom styling with a StyledComponent?

Znippet: Easy and clean solution: Added a wrapper for transform

klinzo (approved):
Tested on Android.

maxammann (approved):

PR #300, opened 2021-05-26
IGAPP-643: Move i18n module

klinzo:


web-ts/src/components/I18nProvider.tsx
```diff
@@ -1,52 +1,46 @@
...
-import BrowserLanguageDetector from '../BrowserLanguageDetector'
-import DateFormatter from 'api-client/src/i18n/DateFormatter'
+import { DateFormatter } from 'api-client'
 import DateFormatterContext from '../context/DateFormatterContext'
+import buildConfig from '../constants/buildConfig'
+import BrowserLanguageDetectorService from '../services/BrowserLanguageDetector'
 
-type PropsType = {| children: React.Node |}
+type PropsType = {
+  contentLanguage: string | undefined
```
klinzo, line 12: `contentLanguage` is now passed as prop.

Taggotty (approved):

web-ts/src/components/I18nProvider.tsx
```diff
@@ -70,29 +63,28 @@ export default ({ children }: PropsType) => {
...
     if (document.documentElement) {
       document.documentElement.lang = language
     }
 
-    // Only change ui direction from default ('ltr') if the language is supported
-    if (config.isSupportedLanguage(language)) {
-      dispatch(setUiDirection(config.hasRTLScript(language) ? 'rtl' : 'ltr'))
-    }
-  }, [dispatch, language])
+    // TODO Remove uiDirection from redux state
```
maxammann, line 72: I think we will need it in the redux state, but maybe I'm wrong.
I think for this case it would be simpler to keep it in the state

web-ts/src/services/BrowserLanguageDetector.ts
```diff
@@ -9,32 +11,33 @@ const localStorageAvailable = () => {
...
+
   return hasLocalStorageSupport
 }
 
 const LANGUAGE_LOCAL_STORAGE = 'i18nextLng'
-
-export default {
+const languageDetector: LanguageDetectorModule = {
   type: 'languageDetector',
-  async: false,
```
maxammann, line 26: Why is this removed? are you sure we dont need it?

maxammann (approved):

web-ts/src/components/I18nProvider.tsx
klinzo: I think we don't need it :) Basically we can just pass uiDirection or language as prop and check in each component separately, can't we? If we need redux just for this tiny prop something is wrong imo. But lets just try without and if we need it we can still add it.

web-ts/src/services/BrowserLanguageDetector.ts
klinzo: @Taggotty

web-ts/src/components/I18nProvider.tsx
klinzo: Is this what you tried already for the tooltips, right? Why did you revert it again? If this is not possible/leads to too many changes, I think e.g. using react context would be the better alternative.

web-ts/src/services/BrowserLanguageDetector.ts
klinzo: Seems correct:
Type definition: https://github.com/i18next/i18next/blob/master/index.d.ts#L785
Only usages: https://github.com/i18next/i18next/blob/2482b6117f7aa6a17ee04d24c4dfc6e46cc2138b/i18next.js#L2281

klinzo: It also works correctly AFAICT

PR #301, opened 2021-05-26
[Proposal] IGAPP-664: Only build after checks succeeded

klinzo:

**This should be discussed.**

Builds will take even longer with this change (normally ~5mins, with outliers up to 9mins, so probably 50mins instead of 43mins or something like that), but we would probably reduce our credit usage significantly (especially iOS builds are very expensive) if we only build our apps once tests succeeded. I would not merge this if anyone thinks even longer builds are really annoying and not worth the reduced credit usage because it makes it more tedious and time intensive for us. Personally I think I don't really mind since we have the automerge feature enabled anymore but really not very sure about this.

Some numbers:
Total credits used: 1,100,000, with 990,000 (90%) for the commit workflow.
Of these 990,000 credits of the commit workflow the `build_e2e_ios` job is using 805,000 (81%) and the `build_e2e_android` job 95,000 (9.6%) credits. The success rate of commit workflows is 43%. By only running builds after tests we could therefore reduce credit and resource usage quite significantly.
200,000 credits are 120$.

Some more numbers and insights:
https://app.circleci.com/insights/github/Integreat/integreat-app/workflows/commit/jobs?reporting-window=last-90-days&branch=b1d8acd4-8701-4c76-8a87-7ff2320aa95f

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -957,6 +957,7 @@ workflows:
                 name: build_e2e_android
                 requires:
                     - bump_version
+                    - check_native
             - e2e_android:
                 context:
                     - mattermost
@@ -972,6 +973,7 @@ workflows:
                 name: build_e2e_ios
                 requires:
                     - bump_version
+                    - check_native
             - e2e_ios:
                 context:
                     - mattermost
--- a/.circleci/src/workflows/commit.yml
+++ b/.circleci/src/workflows/commit.yml
@@ -15,6 +15,7 @@ jobs:
 #        - mattermost
 #      requires:
 #        - bump_version
+#        - check_web
 #  - build_web:
 #      name: build_malte_web
 #      build_config_name: malte
@@ -22,6 +23,7 @@ jobs:
 #        - mattermost
 #      requires:
 #        - bump_version
+#        - check_web
 #  - build_web:
 #      name: build_aschaffenburg_web
 #      build_config_name: aschaffenburg
@@ -29,6 +31,7 @@ jobs:
 #        - mattermost
 #      requires:
 #        - bump_version
+#        - check_web
 #  - deliver_web:
 #      filters:
 #        branches:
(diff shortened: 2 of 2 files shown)


Taggotty:
I personally don't have any issues with slightly longer build time, as most of us now use the auto-merge feature. 
Apparently it seems to be a known issue that react-native build times are somewhere between 20-45 minutes. 
On a quick search I found that you can [disable flipper in the ci](https://fbflipper.com/docs/troubleshooting/#disable-flipper-on-ci-builds-ios), which apparently should reduce the build time by about 10 minutes. I think that could be worth trying.

Taggotty (approved):

maxammann:
Fine for me! Also running our tests should not take 8 minutes to be honest :D so there is also optimization possibilities.

In general I think we should not need to care about 100-200€ per month. Thats peanuts compared to the effort we have to do here.

maxammann (approved):

klinzo:
> Fine for me! Also running our tests should not take 8 minutes to be honest :D so there is also optimization possibilities.
> 
> In general I think we should not need to care about 100-200€ per month. Thats peanuts compared to the effort we have to do here.

What do you mean by `the effort we have to do here`? This PR? Or the general work we put into this project?

I think we should care about money spent a tiny little bit, but I agree with you in general. However, a little bit more important to me is small resource usage/environmental impact and if we can reduce that a little bit I am all in. Sure, it's just a very tiny factor but it is something :man_shrugging: 

Also agreeing with the point that test should not run for 8 minutes :astonished: I think 4 mins are yarn alone...

maxammann:
>  What do you mean by the effort we have to do here? This PR? Or the general work we put into this project?

The discussions and this PR, I think investing time into saving money/resources here is lost time.

> I think we should care about money spent a tiny little bit, but I agree with you in general. However, a little bit more important to me is small resource usage/environmental impact and if we can reduce that a little bit I am all in. Sure, it's just a very tiny factor but it is something

Sure, every tiny bit helps, but as Digitalfabrik we haven't invested in "green computing" at all, so if we want to do it, then we should do it seriously.

PR #302, opened 2021-05-26
IGAPP-658: CI checks

klinzo:


Taggotty (commented):
What's the plan here? Just adding the checks to the ci or also successfully running the ci for web? :)

Taggotty:
You could probably also update `restore_yarn_cache.yml` and replace `web` with `web-ts` there and I also found occurences of the old path in `build_web.yml`

Znippet (approved):

maxammann (approved):

.circleci/src/jobs/check_web.yml
```diff
@@ -10,15 +11,28 @@ steps:
...
+      name: TS check
+      command: yarn workspace translations ts:check
+  - run:
+      name: TS check
+      command: yarn workspace build-configs ts:check
   - run:
-      name: Check flow
-      command: yarn workspace web flow:check-now --max-workers ${TOTAL_CPUS}
+      name: TS check
+      command: yarn workspace web-ts ts:check
```
maxammann, line 33: I think we should remove this check. It will be checked during compilation later. This is just checking it twice and will increase build time.

maxammann (changes requested):

maxammann (approved):

.circleci/src/jobs/check_web.yml
klinzo: But don't we just check with our webpack tsconfig? So I think one could error while one does not? Or am I wrong here?

maxammann: No, the webpack ts config is for the webpack config which is written in ts.

Taggotty (approved):

PR #305, opened 2021-05-29
IGAPP-645: Refactor app and use layout

klinzo:

This PR got bigger than planned, sorry for that. Things done here:
* Nest the routing (`RootSwitcher` and `CityContentSwitcher`)
* Handle loading of cities and languages in route switchers
* Render layout (header + footer)
* Add constants for navigation to routes
* Add ThemeProvider and I18nProvider

Important to review:
* `App`
* `CityContentSwitcher`
* `RootSwitcher`

As these are very important design decisions let me know what you think!


web-ts/src/routes/CityContentSwitcher.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return (
+    <LocationLayout
+      cities={cities}
+      categories={null}
+      viewportSmall={false}
+      feedbackTargetInformation={null}
+      languageChangePaths={null}
+      isLoading={false}
```
klinzo, line 100: Not sure about how we should handle this. Maybe keep a state here and pass callbacks to all routes to set it?

web-ts/src/routes/CityContentSwitcher.tsx
```diff
@@ -0,0 +1,118 @@
...
+// TODO pass right props instead of constants
+const CityContentSwitcher = ({ cities, match, location }: PropsType): ReactElement => {
+  const { cityCode, languageCode } = match.params
+  const cityModel = cities.find(it => it.code === cityCode)
+
+  const requestLanguages = useCallback(async () => {
+    return createLanguagesEndpoint(cmsApiBaseUrl).request({ city: cityCode })
+  }, [cityCode])
+  const { data: languages, loading, error: languagesError } = useLoadFromEndpoint<LanguageModel[]>(requestLanguages)
+  const languageModel = languages?.find(it => it.code === languageCode)
```
klinzo, line 48: Loading and finding of cities and languages are entirely handled here (and in `RootSwitcher`), so no more nasty checks in all routes.

web-ts/src/routes/RootSwitcher.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const { i18n } = useTranslation()
+  // LanguageCode is always the second param (if there is one)
+  const languageCode = useRouteMatch<{ languageCode?: string }>('/:slug/:languageCode')?.params.languageCode
+
+  const detectedLanguageCode = i18n.language
+  const language = languageCode ?? detectedLanguageCode
+
+  if (language !== detectedLanguageCode) {
+    setContentLanguage(language)
+  }
```
klinzo, line 65: We have to update the language of the `I18nProvider` if the content language was changed to update our translations.

web-ts/src/components/LocationHeader.tsx
```diff
@@ -1,47 +1,63 @@
...
-export class LocationHeader extends React.Component<PropsType> {
-  getActionItems(): Array<ReactNode> {
-    const { languageCode, pathname, languageChangePaths, t } = this.props
+const LocationHeader = (props: PropsType): ReactElement => {
+  const { viewportSmall, onStickyTopChanged, cityModel, languageCode, pathname, languageChangePaths } = props
+  const { eventsEnabled, poisEnabled, offersEnabled, tunewsEnabled, pushNotificationsEnabled } = cityModel
 
+  const params = { cityCode: cityModel.code, languageCode }
+  const categoriesPath = generatePath(RoutePatterns[CATEGORIES_ROUTE], params)
+  const eventsPath = generatePath(RoutePatterns[EVENTS_ROUTE], params)
```
maxammann, line 44: Same as above it would be better if `generatePath` would do the lookup in `RoutePatterns` and handle errors there.

web-ts/src/components/LocationHeader.tsx
```diff
@@ -1,47 +1,63 @@
...
+  const offersPath = generatePath(RoutePatterns[OFFERS_ROUTE], params)
+  const poisPath = generatePath(RoutePatterns[POIS_ROUTE], params)
+  const newsPath = generatePath(RoutePatterns[pushNotificationsEnabled ? LOCAL_NEWS_TYPE : TU_NEWS_TYPE], params)
```
maxammann, line 47: I think `generate` is not the right term here as it referes to randomness somehow.

Maybe just `create`.

web-ts/src/routes/categories/CategoriesPage.tsx
```diff
@@ -1,6 +1,10 @@
-import React from 'react'
```
maxammann, line 1: I think we can flatten this structure to `src/routes/CategoriesPage.tsx`.

web-ts/src/routes/CityContentSwitcher.tsx
maxammann: I think we should not handle this in a centralized way. I think this code is likely to change with new requirements.

Right now we do things this way:

```
<Router>
<SomeLayout>
<RouteXY>
</RouteXY>
</SomeLayout>
</Router>
```

I think this is the wrong way around. It should be:

```
<Router>
<RouteXY>
<SomeLayout>
</SomeLayout>
</RouteXY>
</Router>
```

This is because "The route determines the layout" and not the other way around: The layout determines the route.

The nasty checks you mentioned come from the fact that we used the state and called the functions which fill the state from a centralized code in the app. If you do this, then you have to check in every components whether the state is "ready".

So I propose to create a handful of reusable hooks like: `useCity`, `useCategories`, `useEvents`. They are called in each route. Until they are not finished display a loading spinner. As soon as they are finished display the content or an error. The type of layout rendering is also happen in each route. This will lead to some duplicate, but very simple code. Also then you see the dependencies of each route very clearly. Like CategoriesRoute requires cities, page and events and therefore sends 3 fetch requests.

If we add the new MapsView for the maps feature which requires a special Layout then there is no way of implementing it without touching this file. That should not be neccassary.

web-ts/src/routes/CityContentSwitcher.tsx
maxammann: To what do you refer here?

web-ts/src/routes/RootSwitcher.tsx
maxammann: Looks fine, the language detection should happen somewhere before rendering the individual routes.

maxammann (changes requested):
See comments

klinzo:
> Should we adjust the implementation of `CleanLink`? Here is an example on how to create a custom link:

I do not know what what links you mean. Do you mean our NavigationItems in the Header, i.e. Categories, Events, ...?
But yes, why not, but not in this issue/PR, so feel free to create an issue :+1:

api-client/src/endpoints/createLanguagesEndpoint.ts
```diff
@@ -1,21 +1,17 @@
...
-  city: string | null | undefined
+  city: string
 }
+
 export default (baseUrl: string): Endpoint<ParamsType, Array<LanguageModel>> =>
   new EndpointBuilder<ParamsType, Array<LanguageModel>>(LANGUAGES_ENDPOINT_NAME)
-    .withParamsToUrlMapper(params => {
-      if (!params.city) {
-        throw new ParamMissingError(LANGUAGES_ENDPOINT_NAME, 'city')
-      }
```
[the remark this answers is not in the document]
klinzo: Because it is handled by typescript that it is not undefined. Otherwise we would need this check whenever we are using props or variables as well.

web-ts/src/components/GeneralFooter.tsx
```diff
@@ -20,7 +22,7 @@ class GeneralFooter extends React.PureComponent<PropsType> {
 
     return (
       <Footer>
-        <CleanLink to='/disclaimer'>{t('imprintAndContact')}</CleanLink>
+        <CleanLink to={RoutePatterns[MAIN_DISCLAIMER_ROUTE]}>{t('imprintAndContact')}</CleanLink>
```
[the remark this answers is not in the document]
klinzo: This would work for main disclaimer, but not for other routes with params, would it?

web-ts/src/components/GeneralHeader.tsx
```diff
@@ -6,17 +6,14 @@ import { withTranslation, TFunction } from 'react-i18next'
 import buildConfig from '../constants/buildConfig'
 
 type PropsType = {
+  landingPath: string
```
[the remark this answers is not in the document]
klinzo: Not really. I first thought it may be better not to have dependencies to the `RoutePatterns` in components, but I think that is okay. Will use `RoutePatterns` here directly.

web-ts/src/components/LocationHeader.tsx
klinzo: `generatePath` is a function of `react-router`, see https://reactrouter.com/web/api/generatePath. We could introduce a custom wrapper though, I think I would like that :+1: Would do that in a separate PR though. We also need checks whether the current route is a specific route.

web-ts/src/components/LocationHeader.tsx
klinzo: See answer above, `generatePath` is a function of `react-router`, see https://reactrouter.com/web/api/generatePath.

web-ts/src/routes/categories/CategoriesPage.tsx
klinzo: Sure, but lets just wait for https://issues.tuerantuer.org/browse/IGAPP-640 as I did not create this in this PR.

web-ts/src/components/LocationHeader.tsx
```diff
@@ -138,10 +142,9 @@ export class LocationHeader extends React.Component<PropsType> {
       items.push(
         <HeaderNavigationItem
           key='offers'
-          // TODO Use right path and check
-          href='/'
           active={false}
-          // href={new OffersRouteConfig().getRoutePath({ city, language })}
+          href={offersPath}
+          // TODO Use right check
```
Znippet, line 147: Shall we add a task for this to keep this in mind and link it here and at the other "TODO user right checks"?

web-ts/package.json
```diff
@@ -48,6 +48,7 @@
     "moment": "2.29.1",
     "normalize-strings": "^1.1.1",
     "normalize.css": "^8.0.1",
+    "path-to-regexp": "^1.7.0",
```
Znippet, line 51: Do we use this or is this only used by react-router-dom generatepath?
Say: Do we have to import this our own?

Znippet (approved):
Approved with comment of Max pending (could lead to heavy changes:)

web-ts/package.json
klinzo: This is only used by `react-router-dom`, however there seems to be an issue that `<Redirect>` does not work without this:
https://github.com/ReactTraining/react-router/issues/6840#issuecomment-541039824

web-ts/src/components/LocationHeader.tsx
klinzo: Created an issue: https://issues.tuerantuer.org/browse/IGAPP-668

web-ts/src/routes/categories/CategoriesPage.tsx
```diff
@@ -1,6 +1,10 @@
...
+import React, { ReactElement } from 'react'
+import { RouteComponentProps } from 'react-router-dom'
 
-const CategoriesPage = (props: void) => {
+type PropsType = RouteComponentProps<{ cityCode: string; languageCode: string; categoriesId: string | undefined }>
+
+const CategoriesPage = ({ match }: PropsType): ReactElement => {
+  console.log(match.params)
```
Taggotty, line 7: can be removed

web-ts/src/routes/CityContentSwitcher.tsx
klinzo: Where this props come from and how to set/pass them. Of course becomes easy if we move the layout to the routes like you suggested.

web-ts/src/routes/CityContentSwitcher.tsx
klinzo: So I think you are actually mentioning two issues here @maxammann, which are somehow related though.
1) Where the loading happens, i.e. whether we load cities and languages in the switcher and pass those to the routes or load it separately in each route:
I actually don't really see a problem with my approach here, dependencies are clear as those are passed as props and there is just no need to reload this whenever a user navigates. The fetching of the specific content like categories, events and so on should definitely happen in the routes though and not in that central place.

2) Whether the layout is wrapped around the route or the route around the layout:
The problem with rendering the app (as @Taggotty mentioned) was the reason why I chose to implement it this way, otherwise the layout is umounted and remounted whenever one navigates, which will give us the feeling of a simple old website imo.
New specific layouts for new routes aren't a real problem either as we can just not wrap it in the LocationLayout and render a layout in the route. Not very nice to handle layouts in different places but I think not to rerender the layout all the time is more important. Or do you see another way to prevent this?

web-ts/src/routes/categories/CategoriesPage.tsx
klinzo: Created an issue: https://issues.tuerantuer.org/browse/IGAPP-671

web-ts/src/components/LocationHeader.tsx
klinzo: Will do this in https://issues.tuerantuer.org/browse/IGAPP-668

web-ts/src/routes/CityContentSwitcher.tsx
klinzo: Other opinions?

maxammann (approved):

klinzo:
Will apply @maxammann suggestion to decide the layout in each route in IGAPP-676

web-ts/src/routes/CityContentSwitcher.tsx
klinzo: So actually rerendering is not a problem, react detects that it is the same component automatically and keeps it rendered. Therefore done in #310.

PR #306, opened 2021-05-29
IGAPP-666: Use tsc --build

maxammann:


maxammann:
I moved the IntelliJ run configs and changed some commands in the package.json

native/package.json
```diff
@@ -11,13 +11,13 @@
     "start:aschaffenburg": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native start --reset-cache",
     "start:integreat-e2e": "cross-env BUILD_CONFIG_NAME='integreat-e2e' react-native start --reset-cache",
     "android": "yarn android:integreat-test-cms",
-    "android:integreat-test-cms": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' react-native run-android --no-packager",
-    "android:malte": "cross-env BUILD_CONFIG_NAME='malte' react-native run-android --no-packager",
-    "android:aschaffenburg": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native run-android --no-packager",
+    "android:integreat-test-cms": "cross-env BUILD_CONFIG_NAME='integreat-test-cms' react-native run-android --no-packager --appId app.integreat.test",
```
klinzo, line 14: Where/why is this appId used/needed?

klinzo (approved):

native/package.json
maxammann: For these builds the app id is different because of the build configs. This fixes the long issue that the app does not start on the device when running these commands

Using constants from the build config would be overkill imo

api-client/package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "yarn run eslint",
     "lint:fix": "yarn run eslint --fix",
     "lint:ci": "yarn run eslint --quiet --format junit -o reports/lint/junit-lint.xml",
-    "ts:check": "yarn tsc --project tsconfig.json"
+    "ts:check": "tsc --build"
```
Taggotty, line 16: Maybe it would still make sense to add the path to the tsconfig, as tsc --build by default [always selects the tsconfig in the current directory](https://www.typescriptlang.org/docs/handbook/project-references.html#tsc--b-commandline). So when running yarn workspace web ts:check I am not sure what tsconfig would be selected.

native/package.json
```diff
@@ -11,13 +11,13 @@
...
+    "android:malte": "cross-env BUILD_CONFIG_NAME='malte' react-native run-android --no-packager --appId de.malteapp",
+    "android:aschaffenburg": "cross-env BUILD_CONFIG_NAME='aschaffenburg' react-native run-android --no-packager --appId app.aschaffenburg",
     "android:integreat-e2e": "cross-env BUILD_CONFIG_NAME='integreat-e2e' react-native run-android --no-packager",
     "android:integreat:production": "cross-env BUILD_CONFIG_NAME='integreat' react-native run-android --no-packager --variant=release",
```
Taggotty, line 18: no `appId` here?

Taggotty (approved):

native/package.json
maxammann: no, as the default is fine with the production

PR #307, opened 2021-06-01
IGAPP-646: Platform todos

Taggotty:

Unfortunately @klinzo and I barely missed each other when merging the related branches. Therefore, this PR.


web-ts/src/components/Layout.tsx
```diff
@@ -69,10 +70,9 @@ const Main = styled.main`
   }
 `
 
-// TODO IGAPP-646
-// const Aside = withPlatform(styled.aside<{ platform: Platform }>`
-//   position: ${props => (props.platform.positionStickyDisabled ? 'static' : 'sticky')};
```
Taggotty, line 74: Could not reproduce this with any valid typing.

klinzo: I like it better this way anyway. Maybe even just pass `positionStickyDisabled` instead of the whole `platform`.

web-ts/src/components/Layout.tsx
```diff
@@ -112,31 +112,31 @@ type PropsType = {
  * If a footer is supplied and there's not enough content (in header and children) to fill the viewbox, the footer will
  * always stick to the bottom of the viewbox.
  */
-class Layout extends React.PureComponent<PropsType> {
-  static defaultProps = {
-    asideStickyTop: 0,
-    darkMode: false
-  }
+const Layout = ({ asideStickyTop, footer, header, toolbar, modal, children }: PropsType) => {
+  const platform = useContext(PlatformContext)
```
klinzo, line 116: :+1:

web-ts/src/components/Layout.tsx
```diff
@@ -112,31 +112,31 @@ type PropsType = {
...
-          </div>
-          {modal}
-          <div aria-hidden={modalVisible}>{footer}</div>
-        </RichLayout>
-      </FlexWrapper>
-    )
-  }
+Layout.defaultProps = {
+  asideStickyTop: 0,
+  darkMode: false
```
klinzo, line 139: Can't we just use the following syntax above:
```js
const Layout = ({ asideStickyTop = 0, footer, header, toolbar, modal, children }: PropsType) => {
```
`Did remove darkMode` in #305, so we will probably get merge conflicts here.

klinzo (approved):

maxammann (approved):

klinzo (changes requested):
Conflicting with #314, probably we can remove the whole platform thing completely as we only use it to check for sticky top disabled, right? We can merge the renaming anyway to `contexts` though.

web-ts/src/components/Layout.tsx
```diff
@@ -69,10 +69,8 @@ const Main = styled.main`
   }
 `
 
-// TODO IGAPP-646
-// const Aside = withPlatform(styled.aside<{ platform: Platform }>`
-//   position: ${props => (props.platform.positionStickyDisabled ? 'static' : 'sticky')};
```
klinzo, line 74: In #314 `position` is just set to `sticky`

klinzo (approved):

PR #308, opened 2021-06-02
IGAPP-669: Native file structure

klinzo:


klinzo:
TODO:
- [ ] adjust documentation
- [ ] routes file structure?
- [ ] services?

klinzo:
Best checkout this branch and have a look at the new file structure and let me know if you don't like something/have better ideas.
Things I am not sure about:
* Quite some stuff in `services`, should we split that further? Do you think some files do not belong there?
* Do the folders `redux` and `navigation` make sense as they are currently?
* Should we move all components from `routes` to `components`, i.e. just keep one container/component per route?

Taggotty:
Imo there are still a few files which could move from routes to services/utils (getNearbyPlaces/openProvacyPolicy/...). 
For now I would flatten the routes to e.g. `routes/offers/{Offers.tsx, OffersContainer.tsx}`, as those components should not be used anywhere else. However I would refactor some components to the `components` directory such as EastereggImage or SearchInput.

Taggotty (approved):

maxammann (approved):

PR #309, opened 2021-06-02
IGAPP-632: Enable rule noImplicitAny in api-client

Taggotty:


api-client/src/@types/sanitize-html-react.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'sanitize-html-react' {
+  interface IOptions {
+    allowedAttributes?: Record<string, string[]> | false
+    allowedSchemes?: string[] | boolean
+    allowedTags?: string[] | false
+  }
+
+  export default function sanitize(dirty: string, options?: IOptions): string
+}
```
Taggotty, line 9: Partial copy of `@types/sanitize-html`

klinzo (approved):

maxammann (approved):

PR #310, opened 2021-06-02
IGAPP-676: Move rendering of layout to routes

klinzo:


web-ts/src/routes/RootSwitcher.tsx
```diff
@@ -30,23 +43,26 @@ import LoadingSpinner from '../components/LoadingSpinner'
 import { useTranslation } from 'react-i18next'
 
 export const cityContentPattern = `/:cityCode/:languageCode`
+type cityContentPatternType = typeof cityContentPattern
 
 export const RoutePatterns = {
-  [LANDING_ROUTE]: `/${LANDING_ROUTE}/:languageCode`,
-  [MAIN_DISCLAIMER_ROUTE]: `/${MAIN_DISCLAIMER_ROUTE}`,
-  [NOT_FOUND_ROUTE]: `/${NOT_FOUND_ROUTE}`,
+  [LANDING_ROUTE]: `/${LANDING_ROUTE}/:languageCode` as `/${LandingRouteType}/:languageCode`,
```
klinzo, line 49: The type of template strings are always just `string`. By adding types here we have type security in the components (i.e. which parameters are available for the route, e.g. `cityCode` or `eventId`).

Taggotty: you could do the same as @SteffiStuffel did in her previous PR and declare RoutePatterns `as const` this way typescript is also able to correctly infer the corresponding types and all values would become readonly.

Taggotty (approved):

web-ts/src/routes/RootSwitcher.tsx
klinzo: :pray: thanks! Works like a charm! Applied in #312

maxammann: Very cool! :)

web-ts/src/routes/sprungbrett/SprungbrettOfferPage.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ReactElement } from 'react'
```
maxammann, line 1: I'm not sure whether we discussed this somewhere already, but would it be possible to remove this one layer of directories here?

To have just a directory with an index.ts: `src/routes/SprungbrettOfferPage/index.ts`
or just have a single js file per route? `src/routes/SprungbrettOfferPage.ts?

maxammann (approved):

web-ts/src/routes/sprungbrett/SprungbrettOfferPage.tsx
klinzo: Done in #311, you just reviewed it :D

maxammann: oh oke :D quite a lot of refactoring :P

PR #311, opened 2021-06-04
IGAPP-671: Flatten routes directory and move central components to src

klinzo:


maxammann (approved):

SteffiStuffel (commented):
I am not really sure about this?

Where do you plan to put route-specific components?
We planned to put them into e.g. src/routes/offers/components/OffersListItem.tsx
But now the offers-folder is removed?

klinzo:
> I am not really sure about this?
> 
> Where do you plan to put route-specific components?
> We planned to put them into e.g. src/routes/offers/components/OffersListItem.tsx
> But now the offers-folder is removed?

@SteffiStuffel we decided all components should go to `src/components` directly to keep the folder structure as flat as possible. This is already done for native (with the exception that we have a container and a component for most routes in the `routes` folder, e.g. both `DashboardContainer` and `Dashboard` live in routes there (may be refactored in https://issues.tuerantuer.org/browse/IGAPP-636). This won't be the case in web though, we'll only have one page per route. All other components go to to `src/components`.

SteffiStuffel:
> > I am not really sure about this?
> > Where do you plan to put route-specific components?
> > We planned to put them into e.g. src/routes/offers/components/OffersListItem.tsx
> > But now the offers-folder is removed?
> 
> @SteffiStuffel we decided all components should go to `src/components` directly to keep the folder structure as flat as possible. This is already done for native (with the exception that we have a container and a component for most routes in the `routes` folder, e.g. both `DashboardContainer` and `Dashboard` live in routes there (may be refactored in https://issues.tuerantuer.org/browse/IGAPP-636). This won't be the case in web though, we'll only have one page per route. All other components go to to `src/components`.
> 
> Edit: Also, if we need to or decide otherwise later, we can just readd the folders :)

puh.. oke. So please adjust the docs.

SteffiStuffel (approved):

PR #313, opened 2021-06-04
IGAPP-651: Move landing to typescript

klinzo:


web-ts/src/RootSwitcher.tsx
```diff
@@ -63,7 +63,7 @@ const RootSwitcher = ({ setContentLanguage }: PropsType): ReactElement => {
 
   return (
     <Switch>
-      <Route exact path={RoutePatterns[LANDING_ROUTE]} component={LandingPage} />
+      <Route exact path={RoutePatterns[LANDING_ROUTE]} render={props => <LandingPage cities={cities} {...props} />} />
```
maxammann, line 66: I think this could cause performance issues as we are passing here a closure instead of a static component. We once had a very restrictive eslint rule for this. Maybe its worth here as we are quite low in the component tree.

maxammann (approved):

web-ts/src/RootSwitcher.tsx
klinzo: Actually not sure why we don't just use normal react children for rendering as recommended since both `component` and `render` are apparently more or less legacy since hooks came out:
https://reactrouter.com/web/api/Route/route-render-methods

However, it seems like `render` has no performance problems even though supplying a function:
`This allows for convenient inline rendering and wrapping without the undesired remounting explained above.`

But I guess we should use 
```
<Route ...>
  <LandingPage {...props} />
</Route>
```
anyway...

web-ts/src/components/__tests__/FilterableCitySelector.spec.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import CityModelBuilder from 'api-client/src/testing/CityModelBuilder'
+import { renderWithRouter } from '../../testing/render'
+import buildConfig from '../../constants/buildConfig'
+import FilterableCitySelector from '../FilterableCitySelector'
+import { ThemeProvider } from 'styled-components'
+import { fireEvent } from '@testing-library/react'
+
+describe('FilterableFilterableCitySelector', () => {
```
SteffiStuffel, line 9: Should be FilterableCitySelector,i guess.

web-ts/src/components/FilterableCitySelector.tsx
```diff
@@ -1,28 +1,25 @@
-// @flow
```
SteffiStuffel, line 1: I do not really like the naming of these components. For me it is not clear, what file contains what. maybe we can rename them?
I also think somewhere in the filename there should be the keyword "List" as it is mostly a List?
Maybe you can have a look at this:
https://medium.com/@wittydeveloper/react-components-naming-convention-%EF%B8%8F-b50303551505
So maybe CitySelector should be LandingCityList or so? Not sure about the best solution, but the current one seems confusing - at least for me.

SteffiStuffel (approved):

web-ts/src/components/FilterableCitySelector.tsx
klinzo: Agreed, naming is not very good in quite some components. Also, many components are too big and legacy. Anyway, I think it is a very big task to adjust all naming and refactor the components as well, so imo it is out of the scope of this issue. I guess we also should get the web app up and running again relatively fast. Do you want to create an issue for this? Than you can choose a naming of your choice :)

Taggotty (approved):

web-ts/src/RootSwitcher.tsx
klinzo: Seems like we loose the types which params are available/passed. Will leave this for now as is.

PR #314, opened 2021-06-04
HEADROOM-10: Add types to headroom

Schedulaar:
Missing:
* [x] Make an npm release including https://github.com/Integreat/react-sticky-headroom/pull/26
* [x] Upgrade package in this repository
* [ ] Close IGAPP-646 as this is not necessary imo.


web-ts/src/components/__tests__/ScrollingSearchBox.spec.tsx
```diff
@@ -55,8 +55,11 @@ describe('ScrollingSearchBox', () => {
     ).find(ScrollingSearchBox)
 
     component.find(SearchInput).prop('onFilterTextChange')('test')
-    const prop: (stickyTop: number) => void = component.find(Headroom).prop('onStickyTopChanged')
-    prop(5)
+    const prop: ((stickyTop: number) => void) | undefined = component.find(Headroom).prop('onStickyTopChanged')
+    expect(prop).toBeTruthy()
+    if (prop) {
+      prop(5)
+    }
```
Znippet, line 62: As this is only a unit test you could also tell the typescript compiler that it is defined with an exclamation mark and leave the old implementation:
```
const prop: ((stickyTop: number) => void) = component.find(Headroom).prop('onStickyTopChanged')!
prop(5)
```

Znippet (approved):

klinzo (approved):
The check for `positionStickyTop` is not needed anymore since there is only one old specific edge version not handling it correctly, right?
Then we can probably remove the whole platform thing altogether as it seems to not being used for any other purpose. We should probably not merge #307 then.

Schedulaar:
@klinzo I have run prettier again, I hope the CI succeeds now.

>The check for positionStickyTop is not needed anymore since there is only one old specific edge version not handling it correctly, right?
Then we can probably remove the whole platform thing altogether as it seems to not being used for any other purpose. We should probably not merge #307 then.

Correct, that's what I'd suggest... The benefit is near zero if we want to support that specific edge version I think.

PR #315, opened 2021-06-05
IGAPP-665: Disable flipper builds in CI

maxammann:


Taggotty:
Saved around 10 mins, as expected. Might be worth it 🤷🏼‍♀️. But only if there actually is an easy workaround.

maxammann:
Yep, worth it, not sure yet how to handle the lock file.

* One solution would be to have two lockfiles.
* The other is to remove the --deployment as I did for testing

klinzo:
The option with two lockfiles would provide additional security as it is always clear which dependencies are used, right? Could lead to diverging dependencies though...

maxammann:
> The option with two lockfiles would provide additional security as it is always clear which dependencies are used, right? Could lead to diverging dependencies though...

Actually this is not possible with cocoapods, so this is only a theoretical option :D

Taggotty:
Actually why would you want to enable flipper on main? As far as I know flipper is/should also not be linked in [release mode](https://github.com/react-native-community/upgrade-support/issues/28#issuecomment-652186772). Probably I am missing something, right?

klinzo:
It is more about enforcing the correct dependencies/lockfile and flipper on main is just a drawback.

native/ios/Podfile.lock
```diff
@@ -732,6 +732,6 @@ SPEC CHECKSUMS:
   Yoga: 4bd86afe9883422a7c4028c00e34790f560923d6
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
-PODFILE CHECKSUM: 9f4f5fc0d4639de209a92fd46798bf1ba48b1f6e
+PODFILE CHECKSUM: 26c04831545c22b7526a71e0f520caedf9e5fafb
```
maxammann, line 735: I update this manually, it would be cool if somebody with a mac could run `pod exec pod install` and see whether there are changes.

maxammann:
I decided to just remove the `--deployment` flag, because:

* Conditionally using the flag or not adds a lot of complexity to CI scripts
* The benefit is not clear. 99% of people do not use these flags
* If you forget to update the lock file then somebody will eventually find it or the build will fail in other ways.

native/ios/Podfile.lock
klinzo: This seems to be fine, no changes after executing pod install :+1:

klinzo (approved):

Taggotty (approved):

PR #317, opened 2021-06-06
IGAPP-630: Remove ts-ignore for mock functions

klinzo:


api-client/src/endpoints/__tests__/createCategoryChildrenEndpoint.spec.ts
```diff
@@ -18,29 +18,26 @@ describe('createCategoryChildrenEndpoint', () => {
...
     const children = new CategoriesMapModelBuilder(params.city, params.language).build().toArray()
 
-    ;((mapCategoryJson as unknown) as jest.Mock<
-      (json: JsonCategoryType, basePath: string) => CategoryModel
-      // @ts-ignore
-    >).mockImplementation((json: string) => {
-      if (json === 'myFirstCategory') {
-        return children[0]
-      }
+    ;((mapCategoryJson as unknown) as jest.Mock)
```
maxammann, line 37: Here is a weird semicolon, is this some weird ts syntax?

maxammann (approved):

Taggotty:
Just found https://kulshekhar.github.io/ts-jest/docs/guides/test-helpers/ . Maybe looks a little bit better :)

api-client/src/endpoints/__tests__/createCategoryChildrenEndpoint.spec.ts
Taggotty: Yes, as otherwise typescript will evaluate the bracket as function call or sth like that. The semicolon asserts that the previous expression is finished and has nothing to do with the brackets.

native/src/services/__tests__/createSettingsSections.spec.ts
```diff
@@ -81,17 +81,19 @@ describe('createSettingsSections', () => {
...
-      // @ts-ignore
-      expect(pushNotificationSection.getSettingValue(settings)).toBeTruthy()
-      expect(changedSettings2.allowPushNotifications).toBeFalsy()
+      if (!pushNotificationSection?.getSettingValue) {
+        expect(true).toBeFalsy()
+      } else {
+        const settings = defaultSettings
+        settings.allowPushNotifications = false
+        const changedSettings = changeSetting(settings)
+        expect(pushNotificationSection.getSettingValue(settings)).toBeFalsy()
```
Znippet, line 90: If I interpret this right:
```
if (!pushNotificationSection?.getSettingValue) {
  expect(true).toBeFalsy()
}
```
you want to tell typescript that pushNotificationSection is set (probably otherwise typescript says it can be undefined)
You could simply tell typescript that it is set:
```
pushNotificationSection!.getSettingValue(settings)
```
This unit test will fail with "Cannot read property 'getSettingValue' of undefined" if this is not the case

Znippet (approved):

api-client/src/endpoints/__tests__/createCategoryChildrenEndpoint.spec.ts
klinzo: The alternative would be:
```suggestion
    const mockMapCategoryJson = ((mapCategoryJson as unknown) as jest.Mock)
    mockMapCategoryJson.mockImplementation()
```

klinzo: But becomes with redundant with the `mocked` utility by @Taggotty.

PR #318, opened 2021-06-07
IGAPP-673: Add eslint rule to prevent unused vars and expressions

klinzo:


native/src/__mocks__/react-i18next.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react'
 type I18nType = typeof import('i18next').default
 const realModule = jest.requireActual('react-i18next')
 
-const withTranslation = (namespace: string) => (Component: React.ComponentType<any>): React.ComponentType<any> => {
+const withTranslation = () => (Component: React.ComponentType<any>): React.ComponentType<any> => {
```
maxammann, line 5: I would pefere to use `_namespace:string` to make clear that it is unused. That way it is clear that this parameter exist.

How does TS deal with this? Isn't the type of withTranslation changed in that case to `() -> Component` instead of `string -> Component`?

maxammann (approved):

native/src/sagas/loadCityContent.ts
```diff
@@ -177,7 +177,7 @@ export default function* loadCityContent(
     }
 
     const { featureFlags } = buildConfig()
-    const [categoriesMap, events, pois] = (yield all([
+    const [categoriesMap, events] = (yield all([
```
Taggotty, line 180: here i really prefer if you write `_pois`, as we in fact do load the pois and even cast the array type below

Taggotty (approved):

klinzo:
I am using the prefix `_unused` now, i.e. it becomes `_unusedNamespace`. I think this makes it clearer why there is an `_` at all. What do you think of this?

PR #321, opened 2021-06-09
IGAPP-648: Refactor disclaimer

Taggotty:


web-ts/src/routes/DisclaimerPage.tsx
```diff
@@ -10,20 +24,70 @@ type PropsType = {
...
+
+  if (!disclaimer) {
+    const error =
+      disclaimerError ||
+      new NotFoundError({
+        type: 'disclaimer',
+        id: pathname,
+        city: cityCode,
+        language: languageCode
+      })
```
klinzo, line 75: Normally this should not be needed but I guess our typing of `useLoadFromEndpoint` is not exact enough.
```suggestion
      new Error('Disclaimer should not be null!')
```

klinzo (approved):
Looking good :+1:

web-ts/src/routes/DisclaimerPage.tsx
```diff
@@ -10,20 +24,70 @@ type PropsType = {
   languageModel: LanguageModel
 } & RouteComponentProps<{ cityCode: string; languageCode: string }>
 
-const DisclaimerPage = ({ match, cityModel, location }: PropsType) => {
-  const { languageCode } = match.params
+const DisclaimerPage = (props: PropsType): ReactElement => {
+  const { match, cityModel, languages, location } = props
+  const { languageCode, cityCode } = match.params
+  const viewportSmall = false
```
SteffiStuffel, line 30: Not really into this viewportSmall-Stuff, but Steffen did changes in IGAPP-679. Should your changes consider this?

SteffiStuffel (approved):
Not tested.

web-ts/src/routes/DisclaimerPage.tsx
klinzo: Depends on who merges first I'd say :) I can fix all existing occurrences in my PR.

web-ts/src/routes/DisclaimerPage.tsx
Taggotty: You mean the `disclaimerError` should normally never be null? Why did you use nofFoundError for Categories and Events then?

klinzo: No, but normally only one of the following can be true at the same time:
* loading is true
* an error occurred
* data is present
So basically it should be enough to check that loading is false and no error is present to be sure that the data (i.e. the disclaimer) is defined. The types of `useLoadFromEndpoint` do not represent the above restrictions though, therefore an extra check whether `disclaimer` is defined here is necessary.

For categories and events the situation is a bit different because we have an array of categories and events and need to check whether a specific category/event is defined. But you are right, a `NotFoundError` is used there as well if (theoretically) categories/events would be null after all.

PR #324, opened 2021-06-10
IGAPP-652: Move news route

klinzo:


web-ts/src/components/NewsListItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+      ontext(data: string) {
+        decodedContent += data
+      }
+    },
+    { decodeEntities: true }
+  )
+  parser.write(content)
+  parser.end()
+
+  const readMoreLinkText = `${t('readMore')} >`
```
klinzo, line 71: This was previous a Link/a tag, but since it is wrapped in a link already that does not make sense.

translations/translations.json
```diff
@@ -2700,7 +2700,8 @@
       "local": "Lokal",
       "readMore": "Weiterlesen",
       "currentlyNoNews": "Aktuell gibt es keine Nachrichten.",
-      "currentlyNoNewsAlternative": "Aktuell gibt es keine Neuigkeiten."
+      "currentlyNoNewsAlternative": "Aktuell gibt es keine Neuigkeiten.",
+      "tuNews": "tünews INTERNATIONAL"
```
klinzo, line 2704: Used for aria-labels in `NewsTab`.

web-ts/src/components/NewsTabs.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const localNewsPath = createPath(LOCAL_NEWS_TYPE, { cityCode: city, languageCode: language })
+  const tunewsPath = createPath(TU_NEWS_TYPE, { cityCode: city, languageCode: language })
+
+  return (
+    <>
+      <StyledTabs>
+        {localNewsEnabled && (
+          <NewsTab active={type === LOCAL_NEWS_TYPE} type={LOCAL_NEWS_TYPE} destination={localNewsPath} t={t} />
+        )}
+        {tunewsEnabled && <NewsTab active={type === TU_NEWS_TYPE} type={TU_NEWS_TYPE} destination={tunewsPath} t={t} />}
```
klinzo, line 34: Maybe we should just not show any tabs at all if only local OR tünews are enabled.

web-ts/src/components/TuNewsList.tsx
```diff
@@ -46,12 +43,12 @@ class TunewsList extends React.PureComponent<PropsType> {
 
     return (
       <StyledList>
-        <InfiniteScroll loadMore={this.loadItems} hasMore={!isFetching && hasMore} loader={<LoadingSpinner key={0} />}>
```
klinzo, line 49: This looked just weird, imo better without a loading spinner. But feel free to check for yourself :)

web-ts/src/routes/TuNewsPage.tsx
```diff
@@ -1,35 +1,152 @@
...
+  }, [page, cityCode, languageCode, tuNews, setTuNews, setError, setLoading, hasMore])
+
+  useEffect(() => {
+    // Reset on language change
+    setTuNews([])
+    setError(null)
+    setLoading(false)
+    setHasMore(true)
+    setPage(DEFAULT_PAGE)
+  }, [languageCode, setTuNews, setError, setLoading, setHasMore, setPage])
```
Taggotty, line 72: There is no reason th add the setters here and above, as the functions should always stay the same. You only need to add state, props and other actually changing variables to hook dependencies. Or just trust the "react-hooks/exhaustive-deps" eslint rule :)
```suggestion
  }, [languageCode])
```

web-ts/src/routes/LocalNewsPage.tsx
```diff
@@ -12,24 +30,120 @@ type PropsType = {
...
+  const languageChangePaths = languages.map(({ code, name }) => ({
+    path: newsId ? null : createPath(LOCAL_NEWS_ROUTE, { cityCode, languageCode: code }),
+    name,
+    code
+  }))
+
+  const locationLayoutParams = {
+    cityModel,
+    viewportSmall,
+    feedbackTargetInformation: null,
```
Taggotty, line 74: no feedback?

web-ts/src/routes/LocalNewsPage.tsx
```diff
@@ -12,24 +30,120 @@ type PropsType = {
...
+        t={t}
+        formatter={formatter}
+        type={LOCAL_NEWS_TYPE}
+      />
+    )
+  }
+
+  // Language change is not possible between local news detail views because we don't know the id of other languages
+  const languageChangePaths = languages.map(({ code, name }) => ({
+    path: newsId ? null : createPath(LOCAL_NEWS_ROUTE, { cityCode, languageCode: code }),
```
Taggotty, line 66: Still wouldn't it be more intuitive to return to `createPath(LOCAL_NEWS_ROUTE, { cityCode, languageCode: code })`?

Taggotty (approved):

web-ts/src/routes/LocalNewsPage.tsx
klinzo: That would differ from the behavior we currently have in both web and native, so I think this should happen in a separate issue if we want it. Maybe you are right though, but not really sure about that...

web-ts/src/routes/LocalNewsPage.tsx
klinzo: That is just as it is in our legacy web app. There are also these two issues to add feedback for news:
https://issues.tuerantuer.org/browse/IGAPP-574
https://issues.tuerantuer.org/browse/IGAPP-575

web-ts/src/routes/TuNewsPage.tsx
klinzo: Hmmm... I was quite sure that the exhaustive-deps rule forced me to add these dependencies as well, but you are right, not needed...

maxammann: The documentation does not say that they do not change, so I'd just leave them: https://reactjs.org/docs/hooks-reference.html#usestate

maxammann (approved):

web-ts/src/routes/TuNewsPage.tsx
Taggotty: [From the docs](https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often): 

> ```javascript
> function Counter() {
>   const [count, setCount] = useState(0);
> 
>   useEffect(() => {
>     const id = setInterval(() => {
>       setCount(c => c + 1); // ✅ This doesn't depend on `count` variable outside    }, 1000);
>     return () => clearInterval(id);
>   }, []); // ✅ Our effect doesn't use any variables in the component scope
>   return <h1>{count}</h1>;
> }
> ```
> (The identity of the setCount function is guaranteed to be stable so it’s safe to omit.)

PR #325, opened 2021-06-10
IGAPP-653: Refactor offers

Taggotty:


web-ts/src/routes/OffersPage.tsx
```diff
@@ -11,25 +28,85 @@ type PropsType = {
...
-      viewportSmall={false}
-      feedbackTargetInformation={null}
-      languageChangePaths={null}
-      isLoading={false}
-      route={OFFERS_ROUTE}
-      languageCode={languageCode}
-      pathname={location.pathname}>
-      <div>OffersPage</div>
+    <LocationLayout isLoading={false} {...locationLayoutParams}>
+      <Tiles title={t('offers')} tiles={toTileModels(offers)} />
```
Taggotty, line 109: Am I missing something? Somehow the translation does not work here -.-

klinzo: Had the same problem already and was very puzzled. You are importing `useTranslation` from the mock instead of `react-i18next` :)

klinzo (approved):
Tested :+1:

web-ts/src/routes/OffersPage.tsx
klinzo: We could exclude mocks from IntelliJ auto import:
Settings > Code Style > Typescript > Imports > Do not import exactly from

SteffiStuffel (approved):
Tested on Chrome. Works fine :rocket:

maxammann (approved):

PR #326, opened 2021-06-11
IGAPP-633: Change linting rule module boundary type to error

Taggotty:

* Will lead to merge conflicts in probably all PRs currently open. (Which are easy to fix. Just select your changes and add the correct type afterwards) 
* Also fixes some easy to fix eslint warnings ([see here](https://github.com/Integreat/integreat-app/pull/326/commits/1287d98e6548d259251e5afe644d756e73fc73d0))
* Functional Components should always return `ReactElement` or - if necessary - `ReactElement | null`.
* The `render` function however always returns `ReactNode`


native/src/__mocks__/@react-native-community/netinfo.ts
```diff
@@ -30,23 +30,23 @@ export const addEventListener = (
...
-export const removeEventListener = () => {
+export const removeEventListener = (): void => {
   throw Error('Function is deprecated')
 }
-export const getConnectionInfo = () => {
+export const getConnectionInfo = (): void => {
   throw Error('Function is deprecated')
 }
-export const isConnectionExpensive = () => {
+export const isConnectionExpensive = (): void => {
```
maxammann, line 39: `void` could be omitted because it is not really interesting I think.

native/src/App.tsx
```diff
@@ -49,7 +49,7 @@ const linking: LinkingOptions = {
 const dataContainer: DataContainer = new DefaultDataContainer()
 const store: Store<StateType, StoreActionType> = createReduxStore(dataContainer)
 
-const App = () => {
+const App = (): ReactElement => {
```
maxammann, line 52: Maybe we should add an exception for react components. The return type is always clear here.

native/src/components/AnchorIcon.tsx
```diff
@@ -48,7 +48,7 @@ class AnchorIcon extends React.Component<PropsType> {
     }
   }
 
-  render() {
+  render(): ReactNode {
```
maxammann, line 51: Maybe we should add an exception for react components. The return type is always clear here.

native/src/components/AppStateListener.ts
```diff
@@ -1,9 +1,9 @@
-import { useEffect } from 'react'
+import { ReactElement, useEffect } from 'react'
 import { AppState } from 'react-native'
 import sendTrackingSignal from '../services/sendTrackingSignal'
 import { RESUME_SIGNAL_NAME, SUSPEND_SIGNAL_NAME } from 'api-client'
 
-const AppStateListener = () => {
+const AppStateListener = (): ReactElement | null => {
```
maxammann, line 6: Maybe we should add an exception for react components. The return type is always clear here.

klinzo (approved):
I am fine with both keep and removing type annotations for react components and void, but I think I would rather keep `void` annotations. Totally indifferent about react components.

native/src/__mocks__/@react-native-community/netinfo.ts
SteffiStuffel: Not sure about this, it is still an information... hmh

native/src/App.tsx
Taggotty: Not really possible for functional components: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
Imo it would be inconsistent to a return type for functional components, but not for react components, especially we are cuttently migrating the components anyway. But feel free to proof me wrong :)

native/src/__mocks__/@react-native-community/netinfo.ts
Taggotty: There is no easy solution to omit specific return types. (You can omit them for specific funtion names, arrow functions with const returns, ...)

SteffiStuffel (approved):

PR #328, opened 2021-06-16
IGAPP-654: Move pois

klinzo:


web-ts/src/components/__tests__/PoiListItem.spec.tsx
```diff
@@ -1,5 +1,3 @@
-// @flow
```
Taggotty, line 1: This would basically be almost no work to rewrite as without enzyme and snapshot. Would you mind? :)

Taggotty (approved):

Taggotty:
Where can I test pois again?

klinzo:
> Where can I test pois again?

TBH I don't know which cities have the pois flag set to true. As a workaround you can just remove the check in the `CityContentSwitcher` disabling the pois route if the flag is set to false.

Taggotty (approved):

maxammann (approved):

PR #330, opened 2021-06-16
IGAPP-685: Fix error for not existing categories and highlighting of navigation item for tunews

klinzo:


web-ts/src/components/LocationHeader.tsx
```diff
@@ -45,7 +43,7 @@ const LocationHeader = (props: PropsType): ReactElement => {
   const eventsPath = createPath(EVENTS_ROUTE, params)
   const offersPath = createPath(OFFERS_ROUTE, params)
   const poisPath = createPath(POIS_ROUTE, params)
-  const newsPath = createPath(pushNotificationsEnabled ? LOCAL_NEWS_TYPE : TU_NEWS_TYPE, params)
+  const newsPath = createPath(pushNotificationsEnabled ? LOCAL_NEWS_ROUTE : TU_NEWS_ROUTE, params)
```
Taggotty, line 46: is this the same as it was in the old web app? Looks a little bit unintuitive.

Taggotty (approved):

web-ts/src/components/LocationHeader.tsx
klinzo: Yes, its the same. If local news are not enabled (i.e. `pushNotificationsEnabled === false`) we link to tünews, otherwise to lcoal news.

maxammann (approved):

PR #334, opened 2021-06-21
IGAPP-629: Type sagas

Taggotty:


native/src/sagas/__tests__/watchFetchCategory.spec.ts
```diff
@@ -159,7 +159,7 @@ describe('watchFetchCategories', () => {
             categoriesMap: categories,
             resourceCache: resources,
             path: initialPath,
-            cityLanguages: languages,
+            cityLanguages: [],
```
klinzo, line 162: Why these changes?

native/src/sagas/fetchResourceCache.ts
```diff
@@ -27,21 +22,21 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
   )
 }
 
-function* watchOnProgress(): Generator<StrictEffect, void, number> {
+function* watchOnProgress() {
```
klinzo, line 25: Types for void are omitted?

native/src/sagas/loadCityContent.ts
```diff
@@ -169,15 +161,15 @@ export default function* loadCityContent(
...
+      const languageAvailable = yield* call(prepareLanguages, dataContainer, newCity, newLanguage, shouldUpdate)
 
       if (!languageAvailable) {
         return false
       }
     }
 
     const { featureFlags } = buildConfig()
-    const [categoriesMap, events, _unusedPois] = (yield all([
+    const [categoriesMap, events, _unusedPois] = (yield* all<any>([
```
klinzo, line 172: Do we have to use any here?

native/src/sagas/watchContentLanguageSwitch.ts
```diff
@@ -34,31 +36,29 @@ export function* switchContentLanguage(
         false
       )
     )
-    const [categories, resourceCache, events, pois] = (yield all([
+    const [categories, resourceCache, events, pois] = (yield* all<any>([
```
klinzo, line 39: Do we have to use any here?

native/src/sagas/watchFetchCategory.ts
```diff
@@ -21,45 +18,28 @@ import { LanguageResourceCacheStateType } from '../redux/StateType'
  * @param routeCity The key of the current route
  * @returns true if the fetch corresponds to a peek
  */
-function* isPeeking(routeCity: string): Generator<SelectEffect, void> {
-  yield select(state =>
+function* isPeeking(routeCity: string): SagaGenerator<boolean> {
+  return yield* select(state =>
```
klinzo, line 22: :thinking:

klinzo (approved):
Very very nice, well done!

Not tested

SteffiStuffel (approved):

native/src/sagas/watchFetchCategory.ts
Taggotty: before we actually did call isPeeking, but it would always return `undefined`, of course, as return was missing. That's also what you see in the tests. If a route isPeeking the languages are not fetched for example.

klinzo: Yes, was just surprised that we did not return it, so the logic was faulty all the time :see_no_evil:

native/src/sagas/fetchResourceCache.ts
Taggotty: no. types for non-exported functions can be omitted.

klinzo:
@Taggotty I like the .all with the object syntax. I think there were one or two more occurrences of this `any` array syntax, could you check please?

PR #335, opened 2021-06-21
IGAPP-686: Migrate app module

klinzo:

I deleted all already migrated or not anymore needed files and moved everything else to web-ts. The only really relevant change is the migration of `initSentry` and the use of it in `App`.

vscode settings will be adjusted here: https://issues.tuerantuer.org/browse/IGAPP-688
documenation will be adjusted here: https://issues.tuerantuer.org/browse/IGAPP-598


Taggotty (approved):
Renaming of web happens in another PR? Or do we even want to rename web-ts to web?

klinzo:
> Renaming of web happens in another PR? Or do we even want to rename web-ts to web?

I would rename it because the `-ts` suffix has no valuable information in the future. There is already an issue for it: https://issues.tuerantuer.org/browse/IGAPP-690

klinzo:
@maxammann I was not able to send an error to sentry, could you perhaps test this please?

web-ts/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "isolatedModules": true,
     "jsx": "react",
     "lib": ["dom", "dom.iterable", "ES5", "ES6"],
+    "module": "ES2020",
```
maxammann, line 10: This could be the reason why sentry is no longer working as sentry is loaded in a deferred way.

maxammann:
> @maxammann I was not able to send an error to sentry, could you perhaps test this please?

I think the migration to TS is responsible that sentry is no longer working. Did you test whether sentry works before this PR?

maxammann:
I'll test this soon.

maxammann:
Sentry still works for me. I did a release build and disabled UBlock origin. Then it works :)

maxammann (approved):

PR #336, opened 2021-06-21
IGAPP-625: Update gems

klinzo:

@maxammann do we need to test anything here? Since there are no major updates of gems used directly by us I think we should be fine.


maxammann:
@klinzo there is a update of fastlane. Every small update of fastlane can break stuff.
But there is no way to test it expect for trying it. The changelog does not mention the work "break" at least: https://github.com/fastlane/fastlane/releases

maxammann (approved):

SteffiStuffel (approved):
Anything that can be tested here? Just had a short look at it, seems resonable

klinzo:
> Anything that can be tested here? Just had a short look at it, seems resonable

As discussed above doesn't seem like it :/ We probably just have to try if fastlane works the next time as expected...

PR #337, opened 2021-06-21
IGAPP-619: Sanitize endpoints

klinzo:


api-client/package.json
```diff
@@ -17,13 +17,13 @@
...
   "dependencies": {
     "entities": "^2.0.0",
+    "isomorphic-dompurify": "^0.14.0",
     "js-md5": "^0.7.3",
     "lodash": "^4.17.19",
     "moment": "2.29.1",
     "moment-timezone": "^0.5",
     "normalize-path": "^3.0.0",
-    "react": "16.13.1",
-    "sanitize-html-react": "^1.13"
```
Taggotty, line 26: now you can also remove `src/@types/sanitize-html-react.ts`

api-client/src/mapping/mapCategoryJson.ts
```diff
@@ -3,18 +3,14 @@ import CategoryModel from '../models/CategoryModel'
...
   new CategoryModel({
     root: false,
     path: normalizePath(json.path),
     title: json.title,
-    content: sanitizeHtml(json.content, {
-      allowedSchemes: ['http', 'https', 'data', 'tel', 'mailto'],
-      allowedTags: false,
-      allowedAttributes: false
-    }),
+    content: sanitize(json.content),
```
Taggotty, line 13: was there a reason why we didn't support ftp before?

Taggotty (approved):

api-client/src/mapping/mapCategoryJson.ts
klinzo: @maxammann

api-client/package.json
klinzo: I think I did remove them, didn't I?

Taggotty: oh, yes you did. missed that.

api-client/package.json
```diff
@@ -17,13 +17,13 @@
   },
   "dependencies": {
     "entities": "^2.0.0",
+    "isomorphic-dompurify": "^0.14.0",
```
maxammann, line 20: Why aren't we using dompurify directly?

api-client/src/mapping/mapCategoryJson.ts
maxammann: I dont think so, just no usecase. FTP is also not supported in most browsers afaik.

maxammann (approved):

api-client/package.json
klinzo: So I understood @Taggotty wrong, she just said that maybe `dompurify` does not work on native but I thought she checked and it does not work. However, it does work, so we'll use `dompurify` directly.

PR #338, opened 2021-06-21
IGAPP-683: Unescape HTML entities

a-kud:

The title of tünews is not displayed correctly on both native and web if it contains html entities.  This pull request fixes this on both platforms by decoding it before rendering.

<details>
  <summary>Web Screenshots</summary>

  
image

image

</details>

<details>
  <summary>Native Screenshots</summary>

Screenshot_20210622-052703_IntegreatTestCms
Screenshot_20210622-052719_IntegreatTestCms

image
image

</details>


Znippet (approved):

api-client/src/models/TunewsModel.ts
```diff
@@ -18,7 +19,7 @@ class TunewsModel {
   }) {
     const { id, date, title, tags, content, eNewsNo } = params
     this._id = id
-    this._title = title
```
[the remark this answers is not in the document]
klinzo: So both in web and native we use [htmlparser2](https://www.npmjs.com/package/htmlparser2) alredy for the content of news (which may contain actual html and not just html entities, so it is really needed there). The decoding of the content has to happen in web and native to be able to render with or without html conditionally.

This is not the case for the title as it won't ever be rendered as html, so we can decide whether to decode it in the api-client or in both web and native. Imo it makes more sense in the api-client so we only have to do it once though.

DOMParser is probably not supported on native, is it? :thinking: We'd have to check this for `he` as well.

klinzo:
@a-kud although it is not specified in the issue we need to decode the title for local news as well, see https://chat.tuerantuer.org/digitalfabrik/pl/69zcitqdi7gqi8x567bcxyyg8w

api-client/src/models/TunewsModel.ts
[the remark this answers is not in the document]
a-kud: I need help with this TS error:
image

klinzo:
@a-kud let us know when this is ready for review again :) It would be nice if you could prefix your commit messages with `IGAPP-684: `.

Taggotty (approved):
Not tested.

klinzo (changes requested):
I think it would be good to move the `parseHtml` utility to the api-client and have the `htmlparser2` dependency there as well. Than you can just use that utility in the api-client, web and native without having to implement it multiple times.

native/package.json
```diff
@@ -97,6 +97,7 @@
     "@testing-library/jest-native": "^3.4.3",
     "@testing-library/react-native": "^7.1.0",
     "@types/hashids": "^1.0.32",
+    "@types/htmlparser2": "^3.10.2",
```
klinzo, line 100: I guess we can remove `htmlparser2` and its types from both the native and web package.json now, right?

api-client/src/utils/helpers.ts
```diff
@@ -0,0 +1,13 @@
+import { Parser, ParserOptions } from 'htmlparser2'
+
+export function parseHTML(html: string, ontext: (data: string) => void, options: ParserOptions = {}): void {
```
klinzo, line 3: Can't we just return the decoded/parsed html instead of using callbacks? Would be simpler to use then imo.
We are also mainly using anonymous arrow functions so I would prefer that here as well.

web-ts/src/routes/TuNewsDetailPage.tsx
```diff
@@ -18,6 +18,7 @@ import { FailureSwitcher } from '../components/FailureSwitcher'
 import Page from '../components/Page'
 import styled from 'styled-components'
 import TunewsIcon from '../assets/TunewsActiveLogo.png'
+import { parseHTML } from 'api-client/src/utils/helpers'
```
klinzo, line 21: You can also just import and export `parseHTML` in the `api-client/src/index` so that you can import it here from just `api-client` instead of the full path.

api-client/src/utils/helpers.ts
klinzo: Maybe we could even set `decodeEntities` to true since we use that most of the time and it probably does not hurt in the cases we don't use it yet? But not sure about that, what do you think?

maxammann (approved):

api-client/src/utils/helpers.ts
a-kud: I think that htmlparser2 expects a callback to work.

klinzo: Sure, htmplarser needs that. But you can refactor your custom `parseHTML` method to handle the usage of the callback internally, such that you just input html and get the parsed text as string as return value.

api-client/src/models/TunewsModel.ts
```diff
@@ -17,8 +18,11 @@ class TunewsModel {
     eNewsNo: string
   }) {
     const { id, date, title, tags, content, eNewsNo } = params
+
+    const parsedTitle = parseHTML(title, {})
```
klinzo, line 22: You could make the second param optional so you don't have to pass an empty object all the time.

api-client/src/utils/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { Parser, ParserOptions } from 'htmlparser2'
+
+export function parseHTML(html: string, options: ParserOptions): string {
```
klinzo, line 3:
```suggestion
export const parseHTML = (html: string, options?: ParserOptions): string => {
```

api-client/src/models/TunewsModel.ts
klinzo: But shouldn't we use `decodeEntities` here anyway?

native/src/components/NewsDetail.tsx
```diff
@@ -56,6 +56,11 @@ const NewsDetail = ({ theme, newsItem, language, navigateToLink }: PropsType): R
     },
     [navigateToLink, language]
   )
+
+  const decodedNewsItemTitle = parseHTML(newsItem.title, {
```
klinzo, line 60: Why do we need this here, you decoded it in the api-client already, didn't you? At least for Tünews. I think we should probably completely decode in the api-client or completely decode in web and native.

web/package.json
```diff
@@ -84,6 +84,8 @@
     "@types/webpack": "~4",
     "@types/webpack-bundle-analyzer": "^3.9.1",
     "@types/webpack-dev-server": "^3.11.3",
+    "@typescript-eslint/eslint-plugin": "^4.22.0",
+    "@typescript-eslint/parser": "^4.22.0",
```
klinzo, line 88: Why did you add those here? I think those are already part of the root package.json and should not be needed here.

web/package.json
```diff
@@ -97,6 +99,7 @@
     "file-loader": "^6.2.0",
     "html-loader": "1.3.2",
     "html-webpack-plugin": "^4.5.1",
+    "identity-obj-proxy": "^3.0.0",
```
klinzo, line 102: What is this needed for?

klinzo (changes requested):
Well done, nice work! Nearly finished :)

api-client/src/utils/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { Parser, ParserOptions } from 'htmlparser2'
+
+export const parseHTML = (html: string, options: ParserOptions = { decodeEntities: true }): string => {
```
maxammann, line 3: Should we rename this to `decodeHtmlEntities`? If you look at Tünews model it looks a little bit weird. We are just passing a string to parseHtml and get a string back.

api-client/src/models/TunewsModel.ts
```diff
@@ -17,11 +18,15 @@ class TunewsModel {
     eNewsNo: string
   }) {
     const { id, date, title, tags, content, eNewsNo } = params
+
+    const parsedTitle = parseHTML(title)
+    const parsedContent = parseHTML(content)
```
maxammann, line 23: I think this should happen in the corresponding Endpoint classes which create TunewsModel. It is not expected that a constructor is doing heavy work or even throw errors if the html is invalid.

Also this could influence tests which use this model and do not expect that this happends.

maxammann (approved):

api-client/src/utils/helpers.ts
klinzo: I do not agree here. We use this to parse the content of html which includes strings with no html elementes apart from html entities. Nevertheless, it is also used in some places to parse actual html content, so I'd vote to keep the `parseHTML` naming. However, since you already changed it I won't really insist.

klinzo: It is actually used to parse just strings from html foe the tünews.

klinzo (approved):
Working fine :) The only thing I'd do is rename to `parseHTML` but your choice! Nice work :+1:

api-client/src/utils/helpers.ts
a-kud: renamed to `parseHTMLEntities`

PR #339, opened 2021-06-22
IGAPP-680: Set rule noimplicitany true for translations

Taggotty:


translations/src/transformTranslations.ts
```diff
@@ -7,11 +7,14 @@ import { TransformedTranslationsType, TranslationsType } from './types'
...
 export default (translations: TranslationsType): TransformedTranslationsType =>
-  reduce(
+  reduce<TranslationsType, TransformedTranslationsType>(
     translations,
     (transformedTranslations, namespace, namespaceName) => {
       forEach(namespace, (language, languageTag) => {
-        transformedTranslations[languageTag] = { ...transformedTranslations[languageTag], [namespaceName]: language }
+        transformedTranslations[languageTag] = {
+          ...transformedTranslations[languageTag],
+          [namespaceName]: language as Record<string, string>
```
klinzo, line 16: Wouldn't it be enough to add the type to language above in `forEach` instead of typecasting it?

klinzo (approved):

translations/src/transformTranslations.ts
Taggotty: no, does not work (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#strict-function-types)

maxammann (approved):

PR #340, opened 2021-06-23
IGAPP-598: Adjust documentation for ts and add section for adding new languages

klinzo:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **IGAPP**.

PR #341, opened 2021-06-23
IGAPP-672: Redirects for Aschaffenburg

Taggotty:


web-ts/src/RootSwitcher.tsx
```diff
@@ -62,22 +63,26 @@ const RootSwitcher = ({ setContentLanguage }: PropsType): ReactElement => {
...
   }
+  const relevantCities = fixedCity ? cities.filter(city => city.code === fixedCity) : cities
 
   return (
     <Switch>
-      <Route exact path={RoutePatterns[LANDING_ROUTE]} render={props => <LandingPage cities={cities} {...props} />} />
+      <Redirect exact from='/' to={fixedCity ? cityContentPath : landingPath} />
+      <Redirect exact from={`/:cityCode`} to={cityContentPath} />
+      <Redirect exact from={`/${LANDING_ROUTE}`} to={landingPath} />
+      {fixedCity && <Redirect exact from={RoutePatterns[LANDING_ROUTE]} to={landingPath} />}
```
klinzo, line 73: What is this used/needed for? It seems a little bit weird :see_no_evil:

web-ts/src/RootSwitcher.tsx
```diff
@@ -62,22 +63,26 @@ const RootSwitcher = ({ setContentLanguage }: PropsType): ReactElement => {
       </Layout>
     )
   }
+  const relevantCities = fixedCity ? cities.filter(city => city.code === fixedCity) : cities
 
   return (
     <Switch>
-      <Route exact path={RoutePatterns[LANDING_ROUTE]} render={props => <LandingPage cities={cities} {...props} />} />
+      <Redirect exact from='/' to={fixedCity ? cityContentPath : landingPath} />
+      <Redirect exact from={`/:cityCode`} to={cityContentPath} />
```
klinzo, line 71: This redirect effectively prevents all urls with only a single url param from working (i.e. main-disclaimer, not-found) and also shadows the landing redirect. That's why I put those redirects last.

klinzo (changes requested):

web-ts/src/RootSwitcher.tsx
Taggotty: yeah. I seem to have mixed up something here ^^

api-client/src/endpoints/hooks/useLoadFromEndpoint.ts
```diff
@@ -39,7 +39,7 @@ type ReturnType<T> = {
 export const useLoadFromEndpoint = <T>(request: RequestType<T>): ReturnType<T> => {
   const [data, setData] = useState<T | null>(null)
   const [error, setError] = useState<Error | null>(null)
-  const [loading, setLoading] = useState<boolean>(false)
+  const [loading, setLoading] = useState<boolean>(true)
```
Taggotty, line 42: Initially this function will return `{null, false, null}`, probably as it is called before the first render, I think. To not show an error (before the first redirect) we can initially set the state to `loading = true` as each endpoint handles this state correctly

web-ts/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "lib": ["dom", "dom.iterable", "ES5", "ES6"],
     "noEmit": false,
     "outDir": "dist",
+    "sourceMap": true,
```
Taggotty, line 12: This allwos to debug the webapp in typescript directly. Should I write some documentation for this?

web-ts/src/CityContentSwitcher.tsx
```diff
@@ -50,7 +50,7 @@ const CityContentSwitcher = ({ cities, match, location }: PropsType): ReactEleme
   const languageModel = languages?.find(it => it.code === languageCode)
 
   if (!cityModel || !languageModel || !languages) {
-    if (loading) {
+    if (loading || (!languageModel && !loadingError)) {
```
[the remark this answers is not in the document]
Taggotty: see https://github.com/Integreat/integreat-app/pull/341#discussion_r658627675

web-ts/tsconfig.json
klinzo: I think that would be nice, yes!

web-ts/src/RootSwitcher.tsx
```diff
@@ -83,6 +81,9 @@ const RootSwitcher = ({ setContentLanguage }: PropsType): ReactElement => {
       />
       <Route exact path={RoutePatterns[NOT_FOUND_ROUTE]} component={NotFoundPage} />
       <Route path={cityContentPattern} render={props => <CityContentSwitcher cities={relevantCities} {...props} />} />
+      <Redirect exact from='/' to={landingPath} />
+      <Redirect exact from={`/:cityCode`} to={cityContentPath} />
```
klinzo, line 85: This still shadows the redirect below from `/landing` to `/landing/de` I guess.

klinzo (approved):

web-ts/tsconfig.json
klinzo: Maybe keep in mind that there is currently an open documentation PR to avoid merge conflicts: #340

web-ts/src/RootSwitcher.tsx
Taggotty: When I testsed each ordering the outcome is the same. Maybe the more concrete route takes precedence over the general route? Anyhow, I changed the order as it was before, as it makes this more clear and added tests

web-ts/src/__tests__/RootSwitcher.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+    })
+
+    it.each`
+      from              | to
+      ${'/'}            | ${'/augsburg/de'}
+      ${'/landing'}     | ${'/augsburg/de'}
+      ${'/augsburg'}    | ${'/augsburg/de'}
+      ${'/augsburg/de'} | ${'/augsburg/de'}
+      ${'/oldtown'}     | ${'/augsburg/de'}
+      ${'/oldtown/de'}  | ${'/oldtown/de'}
```
Taggotty, line 108: Those two cases are probably a little bit unintuitive. As a user I would probably expect to end up at a not-found route. But as those are also only edge cases I think we don't need any further handling for that.

klinzo (approved):

maxammann (approved):

web-ts/src/RootSwitcher.tsx
klinzo: Yes, the output is the same because `/landing` is treated as city content rotue with `landing` as the city code. Then it redirects to `/landing/${language}`, just the same way as the actual landing redirect does. You can test this by e.g. using the folowing:
```
       <Redirect exact from={`/:cityCode`} to='/asdf' /> 
```
Then `/landing` is redirected to `/asdf`.

PR #342, opened 2021-06-24
IGAPP-622: Remove unused libraries

Taggotty:


Taggotty:
When packages previously were only installed implicitly, they were now added explicitly.

native/jest.config.ts
```diff
@@ -24,7 +24,6 @@ export default {
         https://github.com/facebook/jest/issues/6127 */
   moduleNameMapper: {
     '.+\\.(css|styl|less|sass|scss|png|jpg|ttf|woff|woff2)$': 'jest-transform-stub',
-    '\\.(css|less)$': 'identity-obj-proxy'
```
klinzo, line 27: You can remove 'hashids' from `transformNodeModules` as well, maybe other libraries as well

klinzo (approved):
Nice :+1: How did you check which libraries are unused?

Taggotty:
> Nice  How did you check which libraries are unused?

[depcheck](https://www.npmjs.com/package/depcheck) and manual search for those where I wasn't sure

api-client/package.json
```diff
@@ -29,7 +29,6 @@
     "@types/jest": "^26.0.23",
     "@types/js-md5": "^0.4.2",
     "@types/lodash": "^4.14.168",
-    "@types/node": "^15.0.2",
```
maxammann, line 32: Are you sure this is not used in jest tests? Maybe it makes sense to keep these types as we can call node functions in jest tests.

native/package.json
```diff
@@ -36,22 +36,17 @@
     "ts:check": "tsc --build"
   },
   "dependencies": {
-    "@react-native-community/art": "^1.1.2",
```
maxammann, line 39: This could be a peer dependency. While it is not required to specify them, it is actually good. I'm not sure whether any of the dependencies below are peer dependencies.

maxammann (approved):

native/package.json
klinzo: I checked for all dependencies, none of them are peer dependencies or dependencies of other modules.

klinzo: events is actually needed for htmlparser2 though.

api-client/package.json
klinzo: These are installed globally for the whole project so they are redundant here.

PR #343, opened 2021-06-28
IGAPP-675: Add types to web-ts for NoImplicitAny rule

Znippet:

Added missing types to method parameters and variables. Until now there are issues with untyped imports, so I didn't apply the rule itself


web-ts/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -84,11 +89,13 @@ describe('PoisPage', () => {
...
-    mockUseLoadFromEndpointOnce(() => ({
+    const useLoadFromEndpointMock = (() => ({
       data: pois,
       loading: false,
-      error: null
-    }))
+      error: null,
+      refresh: () => null
+    })) as typeof useLoadFromEndpoint
+    mockUseLoadFromEndpointOnce(useLoadFromEndpointMock)
```
klinzo, line 98: Maybe we could move this to a helper method/create a global mock for this.

klinzo (approved):

maxammann (approved):

web-ts/src/routes/__tests__/PoisPage.spec.tsx
Znippet: done

web-ts/src/testing/mockUseLoadFromEndpoint.tsx
```diff
@@ -0,0 +1,29 @@
+import { useLoadFromEndpoint } from 'api-client'
```
klinzo, line 1: I think this makes more sense in the api-client as `useLoadFromEndpoint` is also imported from there.

web-ts/src/testing/mockUseLoadFromEndpoint.tsx
```diff
@@ -0,0 +1,29 @@
...
+const mockData = (data): typeof useLoadFromEndpoint => {
+  return (() => ({
+    data: data,
+    loading: false,
+    error: null,
+    refresh: () => null
+  })) as typeof useLoadFromEndpoint
+}
+
+export function mockUseLoadFromEndpointWitData<T>(data: T): void {
```
klinzo, line 13: Please use arrow funcs

PR #345, opened 2021-06-30
IGAPP-619: Revert IGAPP-619

klinzo:

### DIFF
--- a/api-client/package.json
+++ b/api-client/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "entities": "^2.0.0",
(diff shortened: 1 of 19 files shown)


maxammann (approved):

a-kud (approved):

Znippet (approved):

klinzo:
Seems like this does not fix anything as isomorphic-dompurify does not work on native either. The reason behind this is probably that it internally uses jsdom which is written for node and therefore does not work on native. I guess I'll have to revert everything for now and we need to think about a different solution.

PR #349, opened 2021-07-05
IGAPP-697: Flipper not working with xcode v12.5

klinzo:


f1sh1918 (approved):

Taggotty (approved):

maxammann (changes requested):
I stopped auto-merge because I want to leave a comment here :)

We should document that we did this in some Jira task. All these changes are not supported by React-Native. Flipper should get updated with React-Native. That means our react-native project is now in an unsupported state.

While it maybe oke, there are also some weird changes in the project.pbxproj file, which should not happen. If these are reverted and we add a note to the next upgrade task then I'm fine with merging.

klinzo:
> I stopped auto-merge because I want to leave a comment here :)
> 
> We should document that we did this in some Jira task. All these changes are not supported by React-Native. Flipper should get updated with React-Native. That means our react-native project is now in an unsupported state.
> 
> While it maybe oke, there are also some weird changes in the project.pbxproj file, which should not happen. If these are reverted and we add a note to the next upgrade task then I'm fine with merging.

@maxammann where do you have the info from that flpper should only be updated together with react-native? The flipper getting started page states that react-native might ship with an outdated version and how to upgrade to the latest version: https://fbflipper.com/docs/getting-started/react-native/#using-the-latest-flipper-sdk

maxammann:
> > I stopped auto-merge because I want to leave a comment here :)
> > We should document that we did this in some Jira task. All these changes are not supported by React-Native. Flipper should get updated with React-Native. That means our react-native project is now in an unsupported state.
> > While it maybe oke, there are also some weird changes in the project.pbxproj file, which should not happen. If these are reverted and we add a note to the next upgrade task then I'm fine with merging.
> 
> @maxammann where do you have the info from that flpper should only be updated together with react-native? The flipper getting started page states that react-native might ship with an outdated version and how to upgrade to the latest version: https://fbflipper.com/docs/getting-started/react-native/#using-the-latest-flipper-sdk

Oh oke it seems like my knowledge is outdated then. In that case only the changes in the project file should be reverted if there is no reason for them.

maxammann (approved):

klinzo:
@maxammann tbh I am a little bit lost here. Without the `embed pods framework` build phase the build fails with the message that OpenSSL framework is not enabled, so seems like it is kind of needed. The `$(inherited)` prevent the following warning:
[!] The `Integreat [Release]` target overrides the `FRAMEWORK_SEARCH_PATHS` build setting defined in `Pods/Target Support Files/Pods-Integreat/Pods-Integreat.release.xcconfig'. This can lead to problems with the CocoaPods installation
I think we already had that some time ago and switched to inherited. Do you have any more information on whether this is fine or not?

maxammann (approved):
Oke then this is probably because of an upgrade of Pod or the updated dependencies changed this.

PR #350, opened 2021-07-05
IGAPP-619: Sanitize remote content on web

klinzo:


web/src/components/RemoteContent.tsx
```diff
@@ -59,68 +60,46 @@ const SandBox = styled.div<{ centered: boolean }>`
...
       }
     })
-  }
+  }, [html, handleClick, sandBoxRef])
 
-  componentDidMount(): void {
-    this.hijackATags()
+  const dangerouslySetInnerHTML = {
+    __html: Dompurify.sanitize(html)
   }
```
klinzo, line 100: The important change in this PR :)

klinzo:
Another try at sanitizing our html. Not sure how we should proceed with native and the react-native-webview.

web/src/components/RemoteContent.tsx
```diff
@@ -59,68 +60,46 @@ const SandBox = styled.div<{ centered: boolean }>`
...
 const HIJACK = new RegExp(buildConfig().internalLinksHijackPattern)
 
-class RemoteContent extends React.Component<PropsType> {
-  static defaultProps = {
-    centered: false
-  }
-
-  sandBoxRef: RefObject<HTMLDivElement>
+const RemoteContent = ({ html, onInternalLinkClick, centered = false }: PropsType): ReactElement => {
+  const sandBoxRef: RefObject<HTMLDivElement> = React.createRef<HTMLDivElement>()
```
Taggotty, line 71:
```suggestion
  const sandBoxRef = React.createRef<HTMLDivElement>()
```

web/src/components/RemoteContent.tsx
```diff
@@ -59,68 +60,46 @@ const SandBox = styled.div<{ centered: boolean }>`
...
-
-    if (target instanceof HTMLAnchorElement) {
-      const href = target.href
-      this.props.onInternalLinkClick(decodeURIComponent(new URL(decodeURIComponent(href)).pathname))
-    }
-  }
+  const handleClick = useCallback(
+    (event: MouseEvent): void => {
+      event.preventDefault ? event.preventDefault() : (event.returnValue = false)
+      const target: EventTarget | null = event.currentTarget
```
Taggotty, line 76:
```suggestion
      const target = event.currentTarget
```

web/src/components/RemoteContent.tsx
```diff
@@ -59,68 +60,46 @@ const SandBox = styled.div<{ centered: boolean }>`
...
+  )
 
-  hijackATags(): void {
-    if (!this.sandBoxRef.current) {
+  useEffect(() => {
+    if (!sandBoxRef.current) {
       return
     }
-    const collection: HTMLCollectionOf<HTMLAnchorElement> = this.sandBoxRef.current.getElementsByTagName('a')
+    const collection: HTMLCollectionOf<HTMLAnchorElement> = sandBoxRef.current.getElementsByTagName('a')
```
Taggotty, line 90:
```suggestion
    const collection = sandBoxRef.current.getElementsByTagName('a')
```

web/src/components/RemoteContent.tsx
Taggotty: So I guess this means we discarded the idea to sanitize the content in the api client? Or do I mix this up with the news?

Taggotty (approved):

api-client/src/endpoints/__tests__/createDisclaimerEndpoint.spec.ts
```diff
@@ -9,7 +10,7 @@ describe('disclaimer', () => {
     title: 'Feedback, Kontakt und mögliches Engagement',
     type: 'disclaimer',
     modified_gmt: '2017-06-12 12:27:57',
-    content: '<a href="javascript:IWantToBeRemoved();">Ich bleib aber da.</a>',
+    content: '<div>Some disclaimer test content :)</div>',
```
maxammann, line 13: I think removing the javascript makes the test invalid. I think this test is now only testing that the content is not changing.

api-client/src/endpoints/__tests__/createEventsEndpoint.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('events', () => {
     path: '/augsburg/de/events/asylpolitischer_fruehschoppen',
     title: 'Asylpolitischer Frühschoppen',
     excerpt: 'Asylpolitischer Fr&uuml;hschoppen',
-    content: '<a href="javascript:IWantToBeRemoved();">Ich bleib aber da.</a>',
+    content: '<div>Some event test content :)</div>',
```
maxammann, line 28: Same here, I think this test is not correct anymore.

web/src/components/__tests__/RemoteContent.spec.tsx
```diff
@@ -50,4 +50,22 @@ describe('RemoteContent', () => {
 
     expect(onInternalLinkClick).toHaveBeenCalledTimes(0)
   })
+
+  it('should sanitize the html', () => {
```
maxammann, line 54: Did you test whether this test really fails if you remove the dom purify sanitization?

maxammann (approved):
If I understood this PR correctly then we moved sanitization from the api-client to the rendering. This is good! Was the previous PR rejected?

What's missing is the same sanitization for the react native webview. But I think we have a task for that.

web/src/components/RemoteContent.tsx
klinzo: Yes, the problem was that neither dompurify nor isomorphic-dompurify work on native (isomorphic-dompurify only works in browser and node). Also imo it is better to sanitize where it is potentially dangerous, i.e. when setting it as inner html. That way, we make sure everything is really sanitized and safe.

klinzo:
> If I understood this PR correctly then we moved sanitization from the api-client to the rendering. This is good! Was the previous PR rejected?
> 
> What's missing is the same sanitization for the react native webview. But I think we have a task for that.

@maxammann no, it was merged but as neither dompurify nor isomorphic-dompurify did work with react-native I had to revert it: #345 

I'll create an issue for that :)

web/src/components/__tests__/RemoteContent.spec.tsx
klinzo: Adjusted it.

Taggotty (approved):

PR #351, opened 2021-07-05
IGAPP-693: Add urdu as language

klinzo:


klinzo:
Testable here: http://0.0.0.0:9000/testumgebung/ur

Taggotty:
> Testable here: http://0.0.0.0:9000/testumgebung/ur

Does not work for me...

translations/src/config.ts
```diff
@@ -109,6 +109,10 @@ class Config {
     },
     hu: {
       rtl: false
+    },
+    ur: {
+      rtl: true,
+      additionalFont: 'lateef'
```
maxammann, line 115: Lateef does not support all arabic characters:
https://software.sil.org/lateef/support/character-set-support/

while Urdu alphabet is:

Unicode range: U+0600 to U+06FF U+0750 to U+077F U+FB50 to U+FDFF  U+FE70 to U+FEFF
https://en.wikipedia.org/wiki/Urdu_alphabet



I'm not sure whether this is fine. Should we investigate more? @FritjofK

maxammann (approved):

translations/src/config.ts
maxammann: We could also create a separate issue and look for/create a tool to check whether our fonts really support all characters.

klinzo:
> > Testable here: http://0.0.0.0:9000/testumgebung/ur
> 
> Does not work for me...

@Taggotty sorry, I should have mentioned it is the malteser testumgebung, so https://malteapp.de/testumgebung/ur

translations/src/config.ts
FritjofK: Do we have a good way to investigate whether this font for example (https://software.sil.org/awami/) might be better than Lateef? 
I would investigate if we could. 

It is probably a good way to inform the Malteser to have an eye on their Urdu texts down the road to check if everything is displayed in a good way. Let me know if you think we want to include the Malteser in the decision making process here but I do not think that they would be of any help. I would inform them about our decision and have them counter our proposal if they want to.

maxammann: Lets discuss this here. Just fyi these discussions are all public :)

The character support from awami is also not complete. I think there is no font which supports all arabic symbols. So the quetsion is whether one of the two is enough. https://software.sil.org/awami/support/character-set-support/

maxammann: The difficult thing is that the system font will jump in if glyphs are missing. This is probably only detectable by native urdu speakers.

f1sh1918: Indeed difficult me neither dunno how prevent system fonts to be rendered. don't think thats even possible. the only thing which maybe could help is to define a fallback font in the font-faces which clearly distinguishes from the standard font

klinzo: @FritjofK we decided that we will keep lateef for now here and think about switching to noto sans completely, i.e. replacing open sans and lateef with it. We already use noto sans for chinese and georgian I think. Is that fine for you?

Taggotty (approved):

PR #352, opened 2021-07-05
IGAPP-698: Fix pdf by readding dependencies

klinzo:

### DIFF
--- a/native/docs/required-dependencies.md
+++ b/native/docs/required-dependencies.md
@@ -1,0 +1,8 @@
+# Required Dependencies
+
+There are several dependencies which seem to be unused but are actually required by other libraries and MUST not be removed.
+These are:
+
+- `@react-native-community/progress-bar-android`: Required by [react-native-pdf](https://github.com/wonday/react-native-pdf/issues/469#issuecomment-649426140)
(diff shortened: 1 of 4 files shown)


klinzo:
Required by react-native-pdf: https://github.com/wonday/react-native-pdf/issues/469#issuecomment-649426140
I also added docu so we know this in the future.

Pdf can be tested e.g. here: https://integreat.app/augsburg/de/wohnen/wohnung-mieten/mietemietvertrag

maxammann (approved):

SteffiStuffel (approved):

f1sh1918:
hm super seltsam, dass die nicht über die lib selbst aufgelöst werden oder nicht?
Die beiden libs werden ja nirgends separat importiert außerhalb von react-native-pdf?

f1sh1918 (approved):

klinzo:
@f1sh1918 right, we don't import them separately, that's also why we thought we can remove it. The reason why those are needed anyway is probably that they have native parts that need to be linked which apparently only happens if it is a main dependency in the package.json and not only a dependency of a dependency.

PR #353, opened 2021-07-05
IGAPP-699: Webview crashing the app

klinzo:

### DIFF
--- a/native/src/App.tsx
+++ b/native/src/App.tsx
@@ -21,9 +21,7 @@ import SnackbarContainer from './components/SnackbarContainer'
 import NetInfo from '@react-native-community/netinfo'
 import sendTrackingSignal from './services/sendTrackingSignal'
 import useSendOfflineJpalSignals from './hooks/useSendOfflineJpalSignals'
-import { enableScreens } from 'react-native-screens'
 
-enableScreens(true)
 NetInfo.configure({
   reachabilityUrl: 'https://cms.integreat-app.de/ping'
 })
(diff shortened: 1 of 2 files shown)


klinzo:
https://github.com/software-mansion/react-native-screens/issues/105#issuecomment-605369538
https://github.com/react-native-webview/react-native-webview/issues/575#issuecomment-635720390
https://github.com/react-navigation/react-navigation/issues/9067#issuecomment-728095564

klinzo:
we really need e2e tests :thinking:

klinzo:
App crashes only on production builds, with development builds and metro bundler nothing happens.
Please test this and make sure to run `yarn android:integreat:production` to do so.

native/src/App.tsx
```diff
@@ -21,9 +21,7 @@ import SnackbarContainer from './components/SnackbarContainer'
 import NetInfo from '@react-native-community/netinfo'
 import sendTrackingSignal from './services/sendTrackingSignal'
 import useSendOfflineJpalSignals from './hooks/useSendOfflineJpalSignals'
-import { enableScreens } from 'react-native-screens'
 
-enableScreens(true)
```
maxammann, line 26: I think it would be good tp explicitly state false here. Maybe screens gets turned on by default at some point in time.

maxammann (approved):

release-notes/unreleased/IGAPP-699-webview-crashing.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-699
+show_in_stores: true
+platforms:
+  - android
+en: Frequent app crashes fixed.
```
SteffiStuffel, line 5: Maybe we could express this more positive. Like "increased app stability"? When I would read this in a store I maybe would not download the app, assuming it is very buggy.

SteffiStuffel (approved):

PR #354, opened 2021-07-06
IGAPP-611: Refactor Feedback

SteffiStuffel:


web/src/components/Feedback.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const Container = styled.div`
+  display: flex;
+  width: 400px;
+  height: auto;
+  box-sizing: border-box;
+  flex-direction: column;
+  justify-content: space-between;
+  padding: 20px;
+  border-radius: 10px;
+  border-color: #585858;
```
klinzo, line 17: I think we should rather use a color from our theme.

web/src/components/Feedback.tsx
```diff
@@ -0,0 +1,87 @@
...
+  resize: none;
+`
+const TextContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`
+const OptionalText = styled.div`
+  font-weight: 400;
+`
```
klinzo, line 30: Isn't `400` the default anyway?

web/src/components/FeedbackContainer.tsx
```diff
@@ -0,0 +1,162 @@
...
+      language,
+      comment: commentWithMail,
+      alias,
+      query
+    }
+  }
+
+  const onFeedbackCommentChanged = (comment: string) => setComment(comment)
+
+  const onFeedbackContactMailChanged = (contactMail: string) => setContactMail(contactMail)
```
klinzo, line 107: I think you can just pass `setComment` and `setContactMail` to `Feedback`, no need to defined these functions.

web/src/components/FeedbackContainer.tsx
```diff
@@ -0,0 +1,162 @@
...
+    })
+  }
+
+  if (['IDLE', 'ERROR', 'SENDING'].includes(sendingStatus)) {
+    return (
+      <>
+        {isSearchFeedback && (
+          <IconTextContainer>
+            <FontAwesomeIcon icon={faFrown} size='4x' />
+            <Text>{t('nothingFound')}</Text>
```
klinzo, line 132: If there are search results some different text should be shown probably.

web/src/components/FeedbackContainer.tsx
```diff
@@ -0,0 +1,162 @@
...
+      setSendingStatus('SUCCESS')
+    }
+
+    request().catch(err => {
+      console.log(err)
+      setSendingStatus('ERROR')
+    })
+  }
+
+  if (['IDLE', 'ERROR', 'SENDING'].includes(sendingStatus)) {
```
klinzo, line 126: Switching the if would make it a little bit easier imo, i.e. checking for `sendingStatus === 'SUCCESS`.

web/src/components/FeedbackModal.tsx
```diff
@@ -28,80 +30,73 @@ const ModalContainer = styled.div`
...
+  return (
+    <FocusTrap>
+      <ModalContainer role='dialog' aria-modal>
+        <Overlay onClick={closeModal} />
+        <ModalContent>
+          <Header>
+            <div>{t('feedback')}</div>
+            <CloseButton aria-label={t('close')} onClick={closeModal}>
+              <FontAwesomeIcon icon={faTimes} />
+            </CloseButton>
```
klinzo, line 88: Why do we have a close button here and in `FeedbackContainer`?

web/src/components/FeedbackSearch.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ReactElement, useState } from 'react'
```
klinzo, line 1: Not sure about naming here, I think it is rather a `SearchFeedback`.

web/src/components/__tests__/__snapshots__/FeedbackSearch.spec.tsx.snap
```diff
@@ -11,13 +11,13 @@ exports[`SearchFeedback should render a FeedbackButton if results are found and
 
 exports[`SearchFeedback should render a NothingFoundFeedbackBox if no results are found 1`] = `
```
klinzo, line 12: The title/test seems to be outdated/wrong

web/src/components/Feedback.tsx
```diff
@@ -0,0 +1,87 @@
...
+      </TextContainer>
+      <CommentField rows={7} value={comment} onChange={event => onCommentChanged(event.target.value)} />
+
+      <TextContainer>
+        <Description>{t('contactMailAddress')}</Description>
+        <OptionalText>({t('optionalInfo')})</OptionalText>
+      </TextContainer>
+      <TextInput onChange={event => onContactMailChanged(event.target.value)} value={contactMail} />
+
+      {sendingStatus === 'ERROR' && <Description>{t('failedSendingFeedback')}</Description>}
```
f1sh1918, line 81: please move error to constants

web/src/components/Feedback.tsx
```diff
@@ -0,0 +1,87 @@
...
+  isSearchFeedback: boolean
+  comment: string
+  contactMail: string
+  onCommentChanged: (comment: string) => void
+  onContactMailChanged: (contactMail: string) => void
+  onSubmit: () => void
+  sendingStatus: SendingStatusType
+}
+
+export const Feedback = (props: PropsType): ReactElement => {
```
f1sh1918, line 52: remove export since using default export at the bottom

web/src/components/__tests__/FeedbackSearch.spec.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import FeedbackSearch from '../FeedbackSearch'
+import { shallow } from 'enzyme'
+import { fireEvent, render } from '@testing-library/react'
+import { ThemeProvider } from 'styled-components'
+import buildConfig from '../../constants/buildConfig'
+
+jest.mock('react-i18next')
+
+describe('SearchFeedback', () => {
```
klinzo, line 10: Rename if we keep the current naming

web/src/components/__tests__/FeedbackSearch.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    expect(
+      shallow(<FeedbackSearch cityCode={cityCode} languageCode={languageCode} query='ab' resultsFound />)
+    ).toMatchSnapshot()
+  })
+
+  it('should render neither a NothingFoundFeedbackBox nor a FeedbackButton', () => {
+    expect(
+      shallow(<FeedbackSearch cityCode={cityCode} languageCode={languageCode} query='' resultsFound />)
+    ).toMatchSnapshot()
+  })
```
klinzo, line 30: Could we replace this with testing library tests?

klinzo (changes requested):

klinzo:
Really nice work :tada:

web/src/components/FeedbackContainer.tsx
```diff
@@ -0,0 +1,162 @@
...
+  routeType: RouteType
+  isPositiveFeedback: boolean
+  isSearchFeedback: boolean
+  closeModal?: () => void
+  alias?: string
+  path?: string
+  query?: string
+}
+
+export type SendingStatusType = 'IDLE' | 'SUCCESS' | 'ERROR' | 'SENDING'
```
f1sh1918, line 55: i would prefer to use enum instead

web/src/components/FeedbackContainer.tsx
SteffiStuffel: no i think it is fine this way. If there are results we have the "Nothing found"-Button, and afterwards this is shown, and therefore the appropriate text.

web/src/components/FeedbackModal.tsx
SteffiStuffel: In FeedbackModal it is for the x in the Header, here it is for Closing the modal after sending the feedback.

web/src/components/FeedbackSearch.tsx
SteffiStuffel: Yes, but this way it is close to the other Feedback-stuff in the components folder. Maybe we can rename to `FeedbackForSearch` if you are really unhappy with this.

klinzo:
Compared to before I personally liked it better before but oc I am not a graphic designer. If you don't have a specific reason for these changes I think I'd wait until we have a designer eventually.

image
image

web/src/components/FeedbackModal.tsx
klinzo: Is this new?

web/src/components/FeedbackContainer.tsx
klinzo: I think this shows the wrong text here

web/src/components/FeedbackContainer.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    const request = async () => {
+      const apiUrl = cmsApiBaseUrl
+      const feedbackEndpoint = createFeedbackEndpoint(apiUrl)
+      await feedbackEndpoint.request(feedbackData)
+      setSendingStatus(SendingState.SUCCESS)
+    }
+
+    request().catch(err => {
+      console.log(err)
```
maxammann, line 122:
```suggestion
      console.error(err)
```

web/src/components/FeedbackSearch.tsx
maxammann: I think from the context it should be clear what it is, if there is FeedbackEvents for example then it makes sense.

web/src/components/__tests__/Feedback.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+    expect(getByText('feedback:send')).toBeDisabled()
+  })
+
+  it('button should be enabled for positive Feedback and no input', () => {
+    const { getByText } = render(
+      <ThemeProvider theme={buildConfig().lightTheme}>
+        <Feedback {...buildProps(true, '')} />
+      </ThemeProvider>
+    )
+    expect(getByText('feedback:send')).not.toBeDisabled()
```
maxammann, line 57: is there maybe a isEnabled? :D

tests look good!

web/src/components/__tests__/Feedback.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+    )
+    expect(getByDisplayValue('test@example.com')).toBeTruthy()
+    expect(queryByDisplayValue('new@example.com')).toBeFalsy()
+    expect(onContactMailChanged).not.toHaveBeenCalled()
+    fireEvent.change(getByDisplayValue('test@example.com'), {
+      target: {
+        value: 'new@example.com'
+      }
+    })
+    expect(onContactMailChanged).toHaveBeenCalledTimes(1)
```
maxammann, line 112: I thinkg `onContactMailChanged` is created only once for the whole testsuite. that means if you also call onContactMailChanged is some other test within this suite, then this test will fail.

I'm not 100% sure whether this is true, because jest is initializing tests is a weird way. But I think this test should be more isolated from the other tests by calling `jest.fn()` from within the `should call callback on contact mail changed` test.

web/src/components/__tests__/Feedback.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <ThemeProvider theme={buildConfig().lightTheme}>
+        <Feedback {...buildProps(false, 'comment')} />
+      </ThemeProvider>
+    )
+    expect(getByText('feedback:send')).not.toBeDisabled()
+  })
+
+  it('should display correct description for search', () => {
+    const { getByText } = render(
+      <ThemeProvider theme={buildConfig().lightTheme}>
```
maxammann, line 71: Should we maybe try to add a global mock for the `theme` similar to the `useTranslation` hook? I think we will probably add a `ThemeProvider` sever hundred times else.

maxammann (approved):

web/src/components/__tests__/__snapshots__/FeedbackSearch.spec.tsx.snap
Taggotty, line 12: this file can now be removed i guess.

web/src/components/Feedback.tsx
```diff
@@ -0,0 +1,85 @@
...
+  width: 400px;
+  height: auto;
+  box-sizing: border-box;
+  flex-direction: column;
+  justify-content: space-between;
+  padding: 20px;
+  border-radius: 10px;
+  border-color: ${props => props.theme.colors.textSecondaryColor};
+  font-size: ${props => props.theme.fonts.contentFontSize};
+`
```
Taggotty, line 19: please try to keep consistent spacings here

Taggotty (approved):

klinzo (approved):

klinzo:
I think you should open a release note for testing :)

web/src/components/__tests__/Feedback.spec.tsx
SteffiStuffel: yes, there is! Changed it :)

web/src/components/__tests__/Feedback.spec.tsx
SteffiStuffel: yeah sounds good to me.

SteffiStuffel:
> Compared to before I personally liked it better before but oc I am not a designer/expert. If you don't have a specific reason for these changes I think I'd wait until we have a designer eventually. There is lots of room for improvement here though, no question.
> 

I used the designs we have for the (native) feedback on zeplin.

web/src/components/FeedbackModal.tsx
SteffiStuffel: No it is not.

PR #356, opened 2021-07-07
IGAPP-578: Fix city names

f1sh1918:


native/src/components/Header.tsx
```diff
@@ -145,7 +147,12 @@ const Header = (props: PropsType): ReactElement => {
     }
 
     const description = routeCityModel.prefix ? ` (${routeCityModel.prefix})` : ''
-    return `${routeCityModel.sortingName}${description}`
+    const cityNameLength = routeCityModel.sortingName.length
+    const minus = '-'
+    return cityNameLength < dimensions.deviceWidth / dimensions.headerTextSize &&
+      routeCityModel.sortingName.indexOf(minus) !== -1
```
klinzo, line 153: I think this check can be omitted, if no minus is in the string nothing is replaced.

SteffiStuffel (approved):

klinzo (approved):
Could you write a release note as well?

Works nice :tada:

f1sh1918:
> Could you write a release note as well?
> 
> Works nice 🎉

done :)

native/src/components/Header.tsx
```diff
@@ -36,7 +36,7 @@ const Icon = styled.Image`
 const HeaderText = styled.Text`
   flex: 1;
   flex-direction: column;
-  font-size: ${props => props.width}px;
+  font-size: ${props => Math.min(props.width, dimensions.headerTextSize)}px;
```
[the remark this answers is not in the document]
f1sh1918: done

native/src/constants/dimensions.ts
```diff
@@ -1,17 +1,25 @@
...
 }
 const dimensions: DimensionsType = {
   headerHeight: 60,
   modalHeaderHeight: 40,
   categoryListItem: {
     iconSize: 40,
     margin: 10
-  }
+  },
+  fontScaling: 0.04,
```
maxammann, line 21: A comment would be nice here

native/src/constants/dimensions.ts
```diff
@@ -1,17 +1,25 @@
...
+  deviceWidth: Number(Dimensions.get('window').width),
```
maxammann, line 22: This can change when you rotate your device. I think there are simple implementations of hook around which allow automatic updating of this value.

native/src/services/forceNewLine.ts
```diff
@@ -0,0 +1 @@
+export const forceNewLine = (text: string, char: string): string => text.replace(char, `${char}\n`)
```
maxammann, line 1: maybe `forceNewlineAfterChar` would be a better name

native/src/components/Header.tsx
```diff
@@ -145,7 +147,10 @@ const Header = (props: PropsType): ReactElement => {
     }
 
     const description = routeCityModel.prefix ? ` (${routeCityModel.prefix})` : ''
-    return `${routeCityModel.sortingName}${description}`
+    const cityNameLength = routeCityModel.sortingName.length
+    return cityNameLength < dimensions.deviceWidth / dimensions.headerTextSize
```
maxammann, line 151: I'm not sure whether I understand this properly:

If the length of the sorting name is longer than the amount of characters we can fit on the screen, then we add a newline.

I see a few issues here:
* Sorting names can be anything, we do not have rules for them
* A character with fontsize 20 is not 20 pixel wide. This would be only an approximitation


Can we somehow let React Native handle the text wraping/breaking?

native/src/components/Header.tsx
```diff
@@ -172,7 +177,10 @@ const Header = (props: PropsType): ReactElement => {
             <Icon source={buildConfigAssets().appIcon} />
           )}
           {routeCityModel && (
-            <HeaderText allowFontScaling={false} theme={theme}>
+            <HeaderText
+              allowFontScaling={false}
+              theme={theme}
+              fontSize={dimensions.deviceWidth * dimensions.fontScaling}>
```
maxammann, line 183: What is this doing?

maxammann (changes requested):

native/src/components/Header.tsx
f1sh1918: Indeed its only an approximation. All i found for react native font scaling was using something like 'vw' and this can be achieved with device width and a scaling factor which is 5% by default. According to some spacing issues i decreased it to 4%
there are some libs doing simular things:
https://www.npmjs.com/package/react-native-responsive-fontsize

React Native deals by default with line breaking but not names including "-", thats why i'm replacing it to force breaking.

native/src/constants/dimensions.ts
f1sh1918: thx for info didn't know about that hook :) works fine

native/src/components/Header.tsx
f1sh1918: it calculates a responsive font size

klinzo:
@maxammann ping :)

native/src/components/Header.tsx
maxammann: K, yeah hacky but its probably the best we can hope for :D

PR #358, opened 2021-07-07
IGAPP-702: Enable screens

klinzo:

Another workaround still needed :/ But not a bad one I guess...

:exclamation: PLEASE TEST THIS WITH A PRODUCTION BUILD! :exclamation:


native/src/components/RemoteContent.tsx
```diff
@@ -37,20 +37,25 @@ type PropType = {
...
+  const defaultWebviewHeight = 1
+  const [webViewHeight, setWebViewHeight] = useState(defaultWebviewHeight)
+  const [webViewWidth, setWebViewWidth] = useState(defaultWebviewHeight)
+
   useEffect(() => {
-    if (webViewHeight !== 0) {
+    if (webViewHeight !== defaultWebviewHeight) {
       onLoad()
     }
   }, [onLoad, webViewHeight])
```
f1sh1918, line 49: why do we have to watch for changes of the onLoad fkt?

klinzo: If the onLoad func changes for whatever reason, you want to execute it again to apply those changes, In practice, this probably won't happen often though. There is also a recommended eslint rule forcing us to use exhaustive dependencies, i.e. specifying all dependencies of hooks.

maxammann: Its better to add whatever could change, than thinking about it :D

maxammann (approved):
Looks good, I'll test this after lunch

maxammann:
navigating to webviews works for me :) :+1:

Taggotty (approved):
Tested

f1sh1918 (approved):

PR #359, opened 2021-07-10
IGAPP-627: Report error to sentry if sending feedback fails

maxammann:


klinzo (approved):
Looking good. Hopefully this will not clash with #354 too much

native/src/components/FeedbackContainer.tsx
```diff
@@ -134,7 +137,10 @@ const FeedbackContainer = (props: PropsType): ReactElement => {
       }
     })
     request().catch(err => {
-      console.log(err)
+      console.error(err)
+      if (fromError(err) !== ErrorCode.NetworkConnectionFailed) {
+        Sentry.captureException(err)
```
Taggotty, line 142:
```suggestion
        reportError(err)
```

Taggotty (approved):

native/src/components/FeedbackContainer.tsx
maxammann: This was in native, in native I did not create a reportError function

PR #361, opened 2021-07-12
IGAPP-710: Open news on push notification click

klinzo:


native/src/App.tsx
```diff
@@ -36,17 +40,37 @@ const linking: LinkingOptions = {
...
+  subscribe: (listener: (url: string) => void) => {
+    const onReceiveURL = ({ url }: { url: string }) => listener(url)
+
+    Linking.addEventListener('url', onReceiveURL)
+
+    const unsubscribeNotification = messaging().onNotificationOpenedApp(() => {
+      console.log('on notification opened app')
+      listener(
+        urlFromRouteInformation({
+          cityCode: CITY_CODE_PLACEHOLDER,
```
klinzo, line 63: It is now possible to open deep links of the following style: `https://integreat.app/city_code_placeholder/language_code/placeholder/news/local` without specifying the city and language.

klinzo (commented):
Not really a nice PR, I would be glad about possible improvements/suggestions

native/src/App.tsx
Taggotty: But that doesn't really sound like a long term solution, right? What about people who switch cities. Do they receive push notifications for both cities or only the one which is currently selected?

klinzo: Upon switching cities the channel of the old city is unsubscribed, so they should only receive PNs of the currently selected city. And yes, there is also [this issue](https://github.com/Integreat/integreat-cms/issues/760) in the cms and [this issue](https://issues.tuerantuer.org/browse/IGAPP-263) in our project to change this. Still, I don't like my current solution either.

Taggotty (approved):

native/src/App.tsx
klinzo: and we don't have to merge this

SteffiStuffel: I had a look on this PR some time ago and did not know what to feedback you and I still feed the same way. Not sure what to do with this...

klinzo:
@maxammann do you have any better solutions for this problem?

native/src/App.tsx
maxammann: I think generally the requirement of opening a link within the current city/language without specifying it is fine. This can be implemented in a proper was I think.

The major design issue is that we never can be sure that the notification which is shown actually belongs to the current city. For example if I receive a PN, then switch the city, and then click on the Notification. What should happen?

If we allow to open other cities via a PN then we really require changes in the CMS. If we do not want to allow this then the current CMS implementation is enough.

native/src/App.tsx
```diff
@@ -36,17 +40,36 @@ const linking: LinkingOptions = {
...
+    ]
+  }),
+  subscribe: (listener: (url: string) => void) => {
+    const onReceiveURL = ({ url }: { url: string }) => listener(url)
+
+    Linking.addEventListener('url', onReceiveURL)
+
+    const unsubscribeNotification = messaging().onNotificationOpenedApp(() => {
+      listener(
+        urlFromRouteInformation({
```
maxammann, line 61: Maybe it would be cleaner to support links like: `https://integreat.app///placeholder/news/local`.
Empty string would evaluate also to `false`. Maybe that would be simpler.

native/src/App.tsx
```diff
@@ -36,17 +40,36 @@ const linking: LinkingOptions = {
...
-      ]
+      }
+    ]
+  }),
+  subscribe: (listener: (url: string) => void) => {
+    const onReceiveURL = ({ url }: { url: string }) => listener(url)
+
+    Linking.addEventListener('url', onReceiveURL)
+
+    const unsubscribeNotification = messaging().onNotificationOpenedApp(() => {
```
maxammann, line 59: We could also call this subscript funciton outside of the react-native navigation `subscribe`. 

That way we do not have to use placeholders. We can simply call `navigateToDeepLink` from the callback of onNotificationOpenedApp.

maxammann (approved):

native/src/App.tsx
klinzo: Yes, it is already planned that the cms delivers city code, language code and news id in the push notifications, but this will only happen with django.

native/src/App.tsx
klinzo: Tried it, makes everything even worse in the pathname parser as we don't know which are actual path segments. Imo it is also clearer with dedicated placeholders.

native/src/App.tsx
klinzo: Tried many stuff and invested way too much time for this, will keep it as it is. For `navigateToDeepLink` we need the `navigation` prop, so it is not possible in sagas. We also need city code and language code so, not really possible either.

SteffiStuffel (approved):

native/src/App.tsx
klinzo: Removed placeholders completely as we can just load city and language from async storage :facepalm:

PR #362, opened 2021-07-12
IGAPP-582: Change update check to before start of today

f1sh1918:


native/src/models/ContentLoadCriterion.ts
```diff
@@ -22,8 +21,8 @@ export class ContentLoadCriterion {
   }
 
   shouldUpdate(lastUpdate: Moment | null | undefined): boolean {
-    // If an update is forced or the last update was more than 24h ago it should update
-    return this._forceUpdate || !lastUpdate || lastUpdate.isBefore(moment.utc().subtract(MAX_CONTENT_AGE, 'hours'))
+    // If an update is forced or the last update was before the start of today
+    return this._forceUpdate || !lastUpdate || lastUpdate.isBefore(moment.utc().startOf('day'))
```
Taggotty, line 25: Could you add a test for this criterion? :)

SteffiStuffel (approved):

native/src/models/ContentLoadCriterion.ts
f1sh1918: @Taggotty 
added some unit tests :) let me know if that suits to requirements

native/src/models/__tests__/ContentLoadCriterion.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    false
+  )
+  it('should not update content if last update was now', () => {
+    const now = moment.utc()
+    expect(stateView.shouldUpdate(now)).toBeFalsy()
+  })
+  it('should update if content was updated 5 minutes before start of today', () => {
+    const beforeStartDay = moment.utc().startOf('day').subtract(5, 'minutes')
+    expect(stateView.shouldUpdate(beforeStartDay)).toBeTruthy()
+  })
```
klinzo, line 20: Maybe you could add one more test with the date set to 5 minutes after midnight?
I think you should also mock the current date, otherwise this test will fail if executed between 0:00 and 0:05. You could try the following to do so: https://github.com/facebook/jest/issues/2234#issuecomment-730037781

klinzo (approved):
Not tested.

Taggotty (approved):

native/src/models/__tests__/ContentLoadCriterion.spec.ts
f1sh1918: @klinzo 
updated tests and mocked current system time

PR #365, opened 2021-07-14
IGAPP-674: Cleanup services to utils

SteffiStuffel:


native/src/redux/reducers/__tests__/cityContentReducer.spec.ts
```diff
@@ -17,6 +17,8 @@ import { CATEGORIES_ROUTE, CategoriesMapModel, ErrorCode, EVENTS_ROUTE, Language
 import cityContentReducer from '../cityContentReducer'
 import { CityContentStateType } from '../../StateType'
 
+jest.mock('rn-fetch-blob')
+
```
klinzo, line 21: Why did you have to/why did you add those mocks?

native/src/utils/helpers.ts
```diff
@@ -0,0 +1,97 @@
...
+    // This way we only report JS crashes exactly when native crashes get reported.
+    return
+  }
+
+  Sentry.init({
+    dsn: 'https://3dfd3051678042b2b04cb5a6c2d869a4@sentry.tuerantuer.org/2'
+  })
+}
+
+export const normalizeSearchString = (str: string): string => normalizeStrings(str).toLowerCase()
```
klinzo, line 97: This will probably clash with #359

web/src/components/CategoryEntry.tsx
```diff
@@ -3,7 +3,7 @@ import { CategoryModel } from 'api-client'
 import iconPlaceholder from '../assets/IconPlaceholder.svg'
 import styled, { withTheme } from 'styled-components'
 import Highlighter from 'react-highlight-words'
-import normalizeSearchString from '../services/normalizeSearchString'
+import { normalizeSearchString } from '../utils/stringUtils'
```
klinzo, line 6: `string` instead of `stringUtils` would probably be enough as `utils`is already the dir. I guess we could also use helpers as in native.

klinzo: If we would use `helpers` you could also include stuff like `wrapDisplayName` there.

klinzo (approved):
How did you decide what to put into `helpers` and what to keep in a separate file?

native/src/redux/reducers/__tests__/cityContentReducer.spec.ts
Taggotty: I looked a little bit more into this. You need to mock it here as we are importing any dependency from `helpers.ts` in this module or any module imported in this module. In the `helpers.ts` at the beginning of the file we are importing `RNFetchBlob from 'rn-fetch-blob`. Withouth a mock this module will throw.
But now if we take a look at the `jest.setup.ts` we see that we actually do mock `rn-fetch-blob` there. 
However, I can tell this mock there will do nothing really. What happens in the `setup.jest.ts` is that we let jest generate a mock for `rn-fetch-blob`. But I think what we want is that jest uses our mock for the test files. To enable this behavior we can remove `jest.mock('rn-fetch-blob')` from the `jest.setup.ts` and add `jest.doMock('rn-fetch-blob')` below the call to walkDir, as we first have to unmock the automatic mock that jest creates for all our files. It is also important to call `doMock` instead of just `mock` as the `jest.mock` statements are all hoisted to the beginning of the file, meaning before we unmocked our `rn-fetch-blob` module.

After doing that you can basically remove all jest.mock('rn-fetch-blob') from the other files, as we then have a global mock for this module.

Taggotty (changes requested):
Please do apply the global mock for `rn-fetch-blob`. This will avoid lots of weird mocking, where nobody really knows why we are doing it.

native/src/redux/reducers/__tests__/cityContentReducer.spec.ts
SteffiStuffel: Thanks Sarah for checking this out! Makes it a lot nicer :)

web/src/components/CategoryEntry.tsx
SteffiStuffel: Yeah, i think both would be valid. I had a look at the existing stuff and thought it makes most sense this way. If your comment is just a idea and not a must-do I prefer keeping the existing solution.

Taggotty (approved):

PR #366, opened 2021-07-14
IGAPP-709: City name jumping

f1sh1918:


native/src/components/Header.tsx
```diff
@@ -159,6 +159,8 @@ const Header = (props: PropsType): ReactElement => {
       <Item title={title} accessibilityLabel={accessibilityLabel} iconName={iconName} show={show} onPress={onPress} />
     )
   }
+  // render placeholder while loading to spare space for icon
+  const renderIconPlaceholder = (title: string) => <Item style={{ opacity: 0 }} title={title} iconName='search' />
```
klinzo, line 163: Why don't you just add a property visible to the renderItem function above and set opacity to zero there if true?

klinzo:
Could you try to add a unit test for this?

native/src/components/Header.tsx
f1sh1918: yes i had this before but regarding to readability i found this approach nicer.
But i can change this

klinzo: Lets wait for other opinions here :)

f1sh1918: And i would have to adjust all the renderItem Fkt-Calls just for these 2 Cases and add a boolean value

klinzo: You could use true as default, then you only have to adjust those two calls

Taggotty: I think I prefer it in one function, if you can make it is clear that renderItem can return a placeholder e.g. by passing the availabilityFlags..

f1sh1918: inschallah :)

native/src/components/Header.tsx
```diff
@@ -151,12 +151,20 @@ const Header = (props: PropsType): ReactElement => {
   const renderItem = (
     title: string,
     show: 'never' | 'always',
-    onPress: () => void,
     accessibilityLabel: string,
```
klinzo, line 154: I think we could remove the accessibility label as it is always the same as the title. Furthermore we could just pass the key and call the `t` function here.

klinzo: We could also pass `show` as boolean and convert it to `never` or `always` here. Of course you don't have to do that, just if you want :)

native/src/components/Header.tsx
```diff
@@ -178,14 +186,19 @@ const Header = (props: PropsType): ReactElement => {
...
-          {renderItem(t('settings'), 'never', goToSettings, t('settings'), undefined)}
-          {routeCityModel && renderItem(t('disclaimer'), 'never', goToDisclaimer, t('disclaimer'), undefined)}
+          {renderItem(t('search'), 'always', t('search'), goToSearch, 'search', !peeking && categoriesAvailable)}
+          {renderItem(
+            t('changeLanguage'),
+            'always',
+            t('changeLanguage'),
+            goToLanguageChange,
+            'language',
+            !!(!peeking && goToLanguageChange)
```
klinzo, line 196:
```suggestion
            !peeking && !!goToLanguageChange
```

native/src/components/Header.tsx
```diff
@@ -178,14 +186,19 @@ const Header = (props: PropsType): ReactElement => {
...
+          )}
+          {showShare && renderItem(t('share'), 'never', t('share'), onShare, undefined)}
+          {showChangeLocation && renderItem(t('changeLocation'), 'never', t('changeLocation'), goToLanding, undefined)}
+          {renderItem(t('settings'), 'never', t('settings'), goToSettings, undefined)}
+          {routeCityModel && renderItem(t('disclaimer'), 'never', t('disclaimer'), goToDisclaimer, undefined)}
```
klinzo, line 201: Do we really need to pass `undefined` here?

f1sh1918: no :)

f1sh1918:
> Could you try to add a unit test for this?

added :)

native/src/components/Header.tsx
```diff
@@ -178,14 +194,18 @@ const Header = (props: PropsType): ReactElement => {
...
-          {renderItem(t('settings'), 'never', goToSettings, t('settings'), undefined)}
-          {routeCityModel && renderItem(t('disclaimer'), 'never', goToDisclaimer, t('disclaimer'), undefined)}
+          {renderItem(HeaderButtonTitle.Search, true, goToSearch, 'search', !peeking && categoriesAvailable)}
+          {renderItem(
+            HeaderButtonTitle.Language,
+            true,
+            goToLanguageChange,
+            'language',
+            !peeking && !!goToLanguageChange
+          )}
```
klinzo, line 204: Do these buttons appear at different times? Maybe we should just show/hide them together, i.e using the check `const visible = !peeking && categoriesAvailable && !!goToLanguageChange`. Not sure about that though, I have also never seen one button appear before another I think (at least not noticeable).

native/src/components/__tests__/MaterialHeaderButtons.spec.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { render } from '@testing-library/react-native'
+import MaterialHeaderButtons from '../MaterialHeaderButtons'
+import buildConfig from '../../constants/buildConfig'
+import { Item } from 'react-navigation-header-buttons'
+
+describe('Material Header Buttons', () => {
```
klinzo, line 7: Hmm so these tests are kind of redundant as you are just asserting that the items you are specifying in the test have the opacity you give them. Instead it would be better to add these tests to the header tests and just render the `Header` component with different props, then you can assert that the opacity of the item with a specific accessibility label is either 0 or 1.

klinzo (approved):

maxammann (approved):

native/src/components/Header.tsx
f1sh1918: as far is see it was not noticeable at all

native/src/components/__tests__/MaterialHeaderButtons.spec.tsx
f1sh1918: i added some proper test for the header component :)

native/src/components/__tests__/Header.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const buildProps = (peeking: boolean, categoriesAvailable: boolean, mode: 'float' | 'screen', goToLanguageChange) => {
+    return {
+      t,
+      theme: buildConfig().lightTheme,
+      peeking,
+      categoriesAvailable,
+      goToLanguageChange,
+      routeCityModel: city,
+      language,
```
Taggotty, line 31: language should be a string.

native/src/components/__tests__/Header.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  const t = key => key
+  const goToLanguageChange = jest.fn()
+  const dispatch = jest.fn()
+  const [city] = new CityModelBuilder(1).build()
+  const language = new LanguageModelBuilder(1).build()[0]
+  const navigation = createNavigationScreenPropMock()
+
+  const buildProps = (peeking: boolean, categoriesAvailable: boolean, mode: 'float' | 'screen', goToLanguageChange) => {
```
Taggotty, line 23: you can also just use one ts-ignore if you specify the return type of `buildProps`
```suggestion
  const buildProps = (peeking: boolean, categoriesAvailable: boolean, mode: 'float' | 'screen', goToLanguageChange): React.ComponentProps<typeof Header> => {
    // @ts-ignore StackHeaderProps not needed
```

Taggotty (changes requested):

Taggotty (approved):

PR #369, opened 2021-07-16
IGAPP-631: Styled-Components Native types

Taggotty:


native/src/components/EastereggImage.tsx
```diff
@@ -119,10 +119,15 @@ class EastereggImage extends React.Component<PropsType, StateType> {
   }
 
   render(): ReactNode {
+    const locationMarker = buildConfigAssets().locationMarker
+    if (!locationMarker) {
+      throw new Error("Location Marker is not available.")
+    }
```
Taggotty, line 125: Should only occur for Aschaffenburg. Not sure if this might break sth. though

native/src/components/IOSSafeAreaView.ts
```diff
@@ -1,8 +1,11 @@
...
 import withTheme from '../hocs/withTheme'
+import { ThemeType } from 'build-configs'
+import { ComponentType } from 'react'
 
 const IOSSafeAreaView = styled(SafeAreaView)`
   flex: 1;
 `
-export default withTheme(IOSSafeAreaView)
+
+export default withTheme(IOSSafeAreaView as (ComponentType<{theme:ThemeType}>))
```
Taggotty, line 11: Is withTheme even required here?

native/src/components/NavigationTile.tsx
```diff
@@ -92,7 +92,7 @@ class NavigationTile extends React.Component<PropsType> {
   render(): ReactNode {
     const { tile, theme, width } = this.props
     return (
-      <TileTouchable theme={theme} onPress={tile.onTilePress} width={width}>
+      <TileTouchable theme={theme} onPress={tile.onTilePress ?? undefined} width={width}>
```
Taggotty, line 95: This and other similar changes appear because most of the properties in React are either set or `undefined`, but `null` is not a valid value. If you have a better solution feel free to drop a comment.

native/src/components/SnackbarContainer.tsx
```diff
@@ -1,23 +1,23 @@
...
 import Snackbar from '../components/Snackbar'
 
-const Container = styled(Animated.View)`
+const Container = styled(View)`
   position: absolute;
   bottom: 0;
   left: 0;
   right: 0;
 `
+const AnimatedContainer = Animated.createAnimatedComponent(Container)
```
Taggotty, line 16: Related issue: https://github.com/styled-components/styled-components/issues/892
Solution: https://stackoverflow.com/q/41425178

native/src/components/IOSSafeAreaView.ts
klinzo: Does not look like it, no.

klinzo (commented):
Nice :tada: Working on linux as well?

api-client/src/models/OfferModel.ts
```diff
@@ -3,14 +3,14 @@ export default class OfferModel {
   _title: string
   _path: string
   _thumbnail: string
-  _postData: Map<string, string> | null
```
SteffiStuffel, line 6: Why is null better than undefined?

native/src/components/EastereggImage.tsx
```diff
@@ -119,10 +119,15 @@ class EastereggImage extends React.Component<PropsType, StateType> {
   }
 
   render(): ReactNode {
+    const locationMarker = buildConfigAssets().locationMarker
+    if (!locationMarker) {
```
SteffiStuffel, line 123: I think just adding locationMarker with the logical && operator to the jsx would be shorter and nicer.

SteffiStuffel (approved):

api-client/src/models/OfferModel.ts
Taggotty: Not quite sure what you are trying to say here ^^
I replaced `null` with `undefined`, as it can also be more convenient, as you can leave the parameter, instead of explicitly setting it to `null`. Furthermore `undefined` is more convenient for react, as there the properties are always set or `undefined`

SteffiStuffel: oke thank you. (i was somehow confused and thought you did it the other way round)

native/src/components/EastereggImage.tsx
```diff
@@ -29,75 +24,57 @@ const ApiUrlText = styled.Text`
...
-      )
+      .then(setApiUrlOverride)
       .catch(e => {
         console.error(e)
       })
-  }
+  }, [])
 
-  onImagePress = async (): Promise<void> => {
+  const onImagePress = useCallback(async () => {
```
f1sh1918, line 42: i'm just wondering why we need here a useCallback hook.
Just wanted to know your idea behind it in this use case

f1sh1918 (approved):

PR #370, opened 2021-07-21
IGAPP-704: added linter rule for no-console and suppress specific con…

f1sh1918:


Taggotty (approved):

.eslintrc.js
```diff
@@ -72,6 +72,7 @@ module.exports = {
       files: ['*.ts', '*.tsx'],
       excludedFiles: ['*.spec.{ts,tsx}', '**/__mocks__/*.ts'],
       rules: {
+        'no-console': 2,
```
klinzo, line 75:
```suggestion
        'no-console': 'error',
```

native/e2e/driver/driver.ts
```diff
@@ -24,13 +24,15 @@ const getConfig = (): ConfigType | null => {
   const configName: string | null | undefined = process.env.E2E_CONFIG
 
   if (!configName) {
+    // eslint-disable-next-line no-console
```
klinzo, line 27: You could also disable this rule for the whole file in this case I think.

native/src/components/EastereggImage.tsx
```diff
@@ -49,6 +49,7 @@ class EastereggImage extends React.Component<PropsType, StateType> {
         })
       )
       .catch(e => {
+        // eslint-disable-next-line no-console
         console.error(e)
```
klinzo, line 53: Would it be possible to exclude `console.error` from this rule?

native/src/components/FeedbackContainer.tsx
```diff
@@ -134,6 +134,7 @@ const FeedbackContainer = (props: PropsType): ReactElement => {
       }
     })
     request().catch(err => {
+      // eslint-disable-next-line no-console
       console.log(err)
```
klinzo, line 138:
```suggestion
      console.error(err)
```

klinzo (approved):

native/src/components/EastereggImage.tsx
f1sh1918: @klinzo 
    'no-console': ['error', { allow: ['error', 'warn'] }],
this should do the trick

PR #373, opened 2021-07-26
IGAPP-713: Translate header items

Taggotty:


klinzo (approved):
Could you add a test for this? I think we had this issue some time ago already. Maybe we should change the mock of the t function to always return something like `t_${key}` or similar.

klinzo (approved):

SteffiStuffel (approved):

f1sh1918 (approved):

PR #374, opened 2021-07-27
IGAPP-532: MapLibreWeb

f1sh1918:
Test Route:
[Testumgebung](http://0.0.0.0:9000/testumgebung/de/locations)


Taggotty (approved):
Tested on several browsers and devices on browserstack.

web/src/components/MapView.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { defaultViewportConfig, mapStyleUrl } from 'api-client/src/mapView'
+
+const MapView: React.FunctionComponent = () => {
+  const [viewport, setViewport] = useState({
+    ...defaultViewportConfig
+  })
+
+  return (
+    <div style={{ display: 'flex', justifyContent: 'center' }}>
+      <ReactMapGL {...viewport} onViewportChange={nextViewport => setViewport(nextViewport)} mapStyle={mapStyleUrl} />
```
klinzo, line 14:
```suggestion
      <ReactMapGL {...viewport} onViewportChange={setViewport} mapStyle={mapStyleUrl} />
```

web/src/components/MapView.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from 'react'
+import ReactMapGL from 'react-map-gl'
+import 'maplibre-gl/dist/maplibre-gl.css'
+
+import { defaultViewportConfig, mapStyleUrl } from 'api-client/src/mapView'
+
+const MapView: React.FunctionComponent = () => {
+  const [viewport, setViewport] = useState({
+    ...defaultViewportConfig
+  })
```
klinzo, line 10:
```suggestion
  const [viewport, setViewport] = useState(defaultViewportConfig)
```

api-client/src/mapView/index.ts
```diff
@@ -0,0 +1,17 @@
+interface MapViewViewport {
```
klinzo, line 1: I think we do not use interfaces anywhere, so my personal preference would be to use types here as well. Maybe we should add an linting rule for this.

klinzo (approved):

klinzo:
Tested on firefox and chrome.

api-client/src/mapView/index.ts
f1sh1918: we talked about personal preferences :D

klinzo: Thats why I explicitly said personal preference :D I just think we should use either types or interfaces and not both.

PR #375, opened 2021-07-27
IGAPP-700: Add User Agent for Native App

Taggotty:


native/src/constants/endpoint.ts
```diff
@@ -1,2 +1,7 @@
+import { Platform } from 'react-native'
+import NativeConstants from './NativeConstants'
+
 export const tunewsApiUrl = 'https://tunews.integreat-app.de'
 export const tunewsWebsiteUrl = 'https://tunewsinternational.com'
+
+export const userAgent = `Integreat Native App (${Platform.OS} ${NativeConstants.appVersion})`
```
klinzo, line 7: If this was discussed with/wished by @svenseeberg ignore this.
Our naming is normally `integreat app native`, at least in terms of mattermost, intellij project and repo docu so we could use that order here as well:
```suggestion
export const userAgent = `Integreat App Native (${Platform.OS} ${NativeConstants.appVersion})`
```
But I don't really mind, your choice :)

klinzo: But way more important: It should not be a hardcoded `Integreat` but instead use the build config name/app name.

klinzo (changes requested):
Seems to work fine, tested on android with integreat test cms build config:
`Integreat Native App (android null);`

native/src/App.tsx
```diff
@@ -51,6 +52,9 @@ const linking: LinkingOptions = {
 }
 const dataContainer: DataContainer = new DefaultDataContainer()
 const store: Store<StateType, StoreActionType> = createReduxStore(dataContainer)
+defaultRequestOptions.headers = {
+  'User-Agent': userAgent
+}
```
maxammann, line 57: I think a `Request.setUserAgent()` function would be cleaner and less error prone.

native/src/constants/endpoint.ts
maxammann: Usually a useragent is of the form `Name/Version (Kommentar)` https://de.wikipedia.org/wiki/User_Agent

```suggestion
export const userAgent = `Integreat/${NativeConstants.appVersion} (${Platform.OS})`
```

no need to add native imo as we know it from the OS.

maxammann (approved):

SteffiStuffel (approved):

native/src/constants/endpoint.ts
Taggotty: But it would also be android if we open the web app on android right, if we also set the user agent there. I'll ask @svenseeberg what he prefers.

klinzo (approved):

native/src/constants/endpoint.ts
```diff
@@ -1,2 +1,7 @@
+import { Platform } from 'react-native'
+import NativeConstants from './NativeConstants'
+
 export const tunewsApiUrl = 'https://tunews.integreat-app.de'
 export const tunewsWebsiteUrl = 'https://tunewsinternational.com'
+
+export const userAgent = `Integreat/${Platform.OS} ${NativeConstants.appVersion}`
```
klinzo, line 7:
```suggestion
export const userAgent = `${buildConfig().appName}/${Platform.OS} ${NativeConstants.appVersion}`
```

klinzo (changes requested):

native/src/constants/endpoint.ts
Taggotty: almost missed that. thx

klinzo (approved):

PR #377, opened 2021-07-28
IGAPP-715: Lighthouse SEO

f1sh1918:
Here are some articles to check about background of improvements:
[Preload Resources](https://web.dev/uses-rel-preload/?utm_source=lighthouse&utm_medium=devtools)
[Webpack Caching](https://webpack.js.org/guides/caching/)


klinzo (commented):
Sounds reasonable. @maxammann @Schedulaar do you have any input on the code splitting/caching topic with webpack?https://webpack.js.org/guides/caching/#extracting-boilerplate

web/src/index.ejs
```diff
@@ -34,15 +34,16 @@
...
         href="/fonts/open-sans/open-sans-v15-greek_cyrillic-ext_vietnamese_latin_cyrillic_greek-ext-700.woff2" as="font"
         type="font/woff2" crossorigin />
   <link rel="preload"
         href="/fonts/open-sans/open-sans-v15-greek_cyrillic-ext_vietnamese_latin_cyrillic_greek-ext-regular.woff2"
         as="font" type="font/woff2" crossorigin />
 
-  <link rel="stylesheet" href="/fonts/raleway/raleway.css" type="text/css" />
+   <link href='/fonts/open-sans/open-sans.css' rel='preload' as='style' />
+            <link href='/fonts/raleway/raleway.css' rel='preload' as='style' />
+            <link href='/fonts/varela-round/varela-round.css' rel='preload' as='style' />
```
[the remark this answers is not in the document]
klinzo: Sorry, my suggestion was wrong. It is still not aligned correctly :D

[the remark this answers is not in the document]
f1sh1918: i reformatted whole file, since it was a mess ;)

f1sh1918:
> Sounds reasonable. @maxammann @Schedulaar do you have any input on the code splitting/caching topic with webpack?https://webpack.js.org/guides/caching/#extracting-boilerplate

Just removed the vendor extraction part, since it's increasing the main package size (+500kb) and i'm not that sure how big the benefit is. Maybe we can make a separate ticket for webpack improvements and discuss them in the team ;)

Schedulaar:
Yeah, I think splitting away vendor (ie. the npm packages) has nearly no benefit (except for some very little caching opportunities when updating the main bundle but I think that's negligible)

Code splitting per sé seems good for improving page load but as far as I know it's hard to do well...
An option would be to figure out which components can be lazy loaded by react / webpack... Maybe it makes sense to lazy load components by route or for example lazy load the feedback dialog (and its dependencies) only once the user presses a feedback button...

But I don't have any experience on code splitting, I never implemented it...
Maybe the router also has some sort of capability for that out of the box?

For code-splitting in react, see https://reactjs.org/docs/code-splitting.html

f1sh1918:
> Yeah, I think splitting away vendor (ie. the npm packages) has nearly no benefit (except for some very little caching opportunities when updating the main bundle but I think that's negligible)
> 
> Code splitting per sé seems good for improving page load but as far as I know it's hard to do well...
> An option would be to figure out which components can be lazy loaded by react / webpack... Maybe it makes sense to lazy load components by route or for example lazy load the feedback dialog (and its dependencies) only once the user presses a feedback button...
> 
> But I don't have any experience on code splitting, I never implemented it...
> Maybe the router also has some sort of capability for that out of the box?
> 
> For code-splitting in react, see https://reactjs.org/docs/code-splitting.html

Indeed i agree.
So i removed the vendor splitting.

Well this Code Splitting approach with Suspense and Lazy imports is working well.
Even its an experimental feature, i see many people using it without big issues
I already implemented it :)

Using it for routing is a good option and also for components using big libraries

Schedulaar:
@f1sh1918 I experimented a little bit yesterday with it and created a branch for code-spliitting routes. See branch https://github.com/Integreat/integreat-app/tree/IGAPP-715-Split-Routes. It does in fact work quite well and it is fairly easy...
Especially for the white-label solution where many routes are not relevant for some apps (i.e. aschaffenburg, malte, ...) this could remove quite some unnecessary code (and hence reduce payload)

web/src/index.ejs
```diff
@@ -1,139 +1,146 @@
...
-      position: fixed;
-      z-index: 2;
-      text-align: center;
-      padding: 20px;
-      top: 0;
-      left: 0;
-      right: 0;
-    }
-  </style>
+
```
klinzo, line 11: What did happen here?

web/package.json
```diff
@@ -112,6 +112,7 @@
     "lolex": "^6.0.0",
     "moment-locales-webpack-plugin": "^1.2.0",
     "moment-timezone-data-webpack-plugin": "1.5.0",
+    "preload-webpack-plugin": "^3.0.0-beta.4",
```
maxammann, line 115: Is there a stable version we can use?

web/tools/webpack.config.ts
```diff
@@ -171,6 +173,10 @@ const createConfig = (
           config: buildConfig
         }
       }),
+      new PreloadWebpackPlugin({
```
maxammann, line 176: Which resources are preloaded by adding this plugin? Is there a benefit without current setup?

maxammann (approved):

web/tools/webpack.config.ts
maxammann: I would only add it if there is a realworld benefit because webpack configs get messy quickly

web/src/components/HeaderLogo.tsx
```diff
@@ -52,7 +51,7 @@ export const HeaderLogo = ({ link }: PropsType): ReactElement => {
   return (
     <LogoContainer>
       <Link to={link}>
-        <img src={src} alt={appName} />
+        <img src={src} alt={appName} width='100%' height='auto' />
```
Taggotty, line 54: whats this about?

web/package.json
f1sh1918: the fork you mentioned @Taggotty is working fine. we can switch to that package

web/src/components/HeaderLogo.tsx
f1sh1918: the problem is that the image gets rendered with 0px then browser requests resources and rerenders image with correct dimensions.
[Explicit Dimensions](https://dev.to/inhuofficial/quick-tips-how-to-fix-image-elements-do-not-have-explicit-width-and-height-in-page-speed-insights-lighthouse-3776)

web/src/index.ejs
f1sh1918: i reformatted the file, seems liek noone did that before. gonna mark the changes

web/src/index.ejs
```diff
@@ -1,139 +1,146 @@
...
-  <!-- Disable phone number auto detection in mobile safari in order to have consistent behavior -->
-  <!-- https://developers.google.com/web/fundamentals/native-hardware/click-to-call#disable_auto-detection_when_necessary -->
-  <meta name="format-detection" content="telephone=no">
-
-  <link rel="dns-prefetch" href="<%= config.cmsUrl %>" />
-  <link rel="preconnect" href="<%= config.cmsUrl %>" />
-
-  <link rel="preload" href="<%= config.cmsUrl %>/wp-json/extensions/v3/sites" as="fetch" crossorigin />
-
-  <link rel="stylesheet" href="/fonts/open-sans/open-sans.css" as="style" type="text/css" />
```
f1sh1918, line 37: Line 37-49: I added misplaced fonts from i18nProviider

web/tools/webpack.config.ts
f1sh1918: main and runtime JS-Files get preloaded. I mean its hard to measure to be honest.
Some blogs recommend doing that but i'm also fine with reverting it.

klinzo (approved):

klinzo:
So @f1sh1918 and I discussed the topic with preloading and splitting and chunks and came to the conclusion that we want to try this at a later time once we have a good estimate on our page loading times in the google search console. The page loading times in the google search console are always the average of the last month, so it would make sense to do this in a month (+ the time until #380 is merged). Then we could try this for another month and decide whether it has a measurable effect and we keep it or remove it again.

PR #381, opened 2021-07-29
IGAPP-675: Apply TS Rule no-implicit-any

Taggotty:


native/src/@types/normalize-strings.d.ts
```diff
@@ -0,0 +1,5 @@
+import 'normalize-strings'
```
Taggotty, line 1: package does not provide typings

native/src/@types/react-native-static-server.d.ts
```diff
@@ -0,0 +1,22 @@
+declare module 'react-native-static-server' {
```
Taggotty, line 1: there are typings for this package, but they weren't released for a year. This is the copy of their types https://github.com/futurepress/react-native-static-server/blob/master/index.d.ts

native/src/@types/untyped.d.ts
```diff
@@ -0,0 +1,3 @@
+// this file only declares untyped module definitions.
```
Taggotty, line 1: here we can add files that are not worth typing by hand, I guess

native/src/__tests__/NavigatorContainer.spec.tsx
```diff
@@ -28,15 +30,14 @@ jest.mock('../constants/NativeConstants')
 jest.mock('react-native-share')
 const mockStore = configureMockStore()
 
-class MockNavigator extends React.Component<void> {
-  render() {
-    return null
-  }
+const MockNavigator: React.FC = () => {
+  return null
```
Taggotty, line 34: Little bit hacky, but basically `React.FC` can also return null elements and is a ReactElement.

native/src/components/List.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { ReactNode } from 'react'
-import { isEmpty } from 'lodash/lang'
+import { isEmpty } from 'lodash'
```
Taggotty, line 3: regarding lodash. In react-native there is no tree shaking. Which basically it makes no difference wether you import from `lodash` or `lodash/lang` . Actually the bungle size gets even smaller when importing only from lodash, which is probably because the optimizer can do its job better or sth. https://github.com/facebook/metro/issues/227#issuecomment-583358386

native/src/components/LoadingSpinner.tsx
```diff
@@ -1,12 +1,7 @@
-import * as React from 'react'
-import { ReactElement } from 'react'
 import styled from 'styled-components/native'
-import { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet'
 
-type PropsType = {
-  style?: ViewStyleProp
-}
```
Taggotty, line 8: Did not work properly with `StyleProp<ViewStyle>`.

native/package.json
```diff
@@ -108,7 +109,7 @@
     "jest-transform-stub": "^2.0.0",
     "jetifier": "^1.6.5",
     "node-fetch": "^2.6.1",
-    "react-native-flipper": "^0.96.1",
+    "react-native-flipper": "^0.99.0",
```
klinzo, line 112: The flipper version in `Podfile` and `gradle.properties` should be adjusted accordingly. Probably someone has to run `bundle exec pod install` once on a mac.

native/src/components/__tests__/ProgressSpinner.spec.tsx
```diff
@@ -10,7 +10,7 @@ type DeepPartial<T> = {
 describe('ProgressSpinner', () => {
   let theme: DeepPartial<ThemeType>
 
-  const t = input => input
+  const t = (input: string) => input
```
klinzo, line 13: Maybe we should provide this as export in a testing utils file, maybe always mapping the key to `t_${key}` like in your last PR in the future to be able to assert that keys are actually translated. What do you think? You don't need to do this here of course though.

native/src/routes/CategoriesContainer.tsx
```diff
@@ -159,8 +159,7 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     }
   }
 
-  // @ts-ignore ts does not get that models and children cannot be undefined as it is already checked above
-  const stateView = new CategoriesRouteStateView(route.path, route.models, route.children)
+  const stateView = new CategoriesRouteStateView(route.path, models ?? {}, children ?? {})
```
klinzo, line 162: :+1:

native/src/routes/__tests__/OffersContainer.spec.tsx
```diff
@@ -80,30 +73,8 @@ describe('OffersContainer', () => {
...
-      <Provider store={store}>
-        <OffersContainer navigation={navigation} route={route} />
-      </Provider>
-    )
-    expect(queryByText('Offers')).toBeTruthy()
-    expect(queryByText('loading')).toBeTruthy()
-    expect(queryByText(errorText)).toBeFalsy()
-  })
-
   it('should display error without a loading spinner', () => {
```
klinzo, line 76:
```suggestion
  it('should display error', () => {
```

native/src/routes/__tests__/SprungbrettOfferContainer.spec.tsx
```diff
@@ -61,80 +57,44 @@ describe('SprungbrettOfferContainer', () => {
...
 
-  const mockUseLoadFromEndpointOnce = mock => {
-    mocked(useLoadFromEndpoint).mockImplementationOnce(mock)
-  }
-
   beforeEach(() => {
     jest.clearAllMocks()
   })
 
   it('should display offers without a Loading spinner', () => {
```
klinzo, line 64: Sorry, I was wrong. It is possible that useLoadFromEndpoint returns data (or an error) and loading true as the data/error is not instantly set to null when refreshing. I think this is fine to allow pull to refresh behavior and still show previously loaded data. So maybe it would be best to readd those tests.

tsconfig.base.json
```diff
@@ -5,7 +5,7 @@
     "lib": ["ES5", "ES6"],
     "moduleResolution": "Node",
     "noFallthroughCasesInSwitch": true,
-    "noImplicitAny": false,
+    "noImplicitAny": true,
```
klinzo, line 8: Do we still need this? What is missing?

web/src/routes/__tests__/SearchPage.spec.tsx
```diff
@@ -7,10 +7,10 @@ import { Route } from 'react-router-dom'
 import { ThemeProvider } from 'styled-components'
 import LanguageModelBuilder from '../../../../api-client/src/testing/LanguageModelBuilder'
 import buildConfig from '../../constants/buildConfig'
-import { mockUseLoadFromEndpointWitData } from '../../testing/mockUseLoadFromEndpoint'
 import { renderWithBrowserRouter } from '../../testing/render'
 import { createPath, RoutePatterns } from '../index'
 import SearchPage from '../SearchPage'
+import { mockUseLoadFromEndpointWitData } from '../../../../api-client/src/testing/mockUseLoadFromEndpoint'
```
klinzo, line 13:
```suggestion
import { mockUseLoadFromEndpointWitData } from 'api-client/src/testing/mockUseLoadFromEndpoint'
```

web/src/__tests__/RootSwitcher.spec.tsx
```diff
@@ -3,12 +3,14 @@ import { renderWithBrowserRouter } from '../testing/render'
...
-import { mocked } from 'ts-jest/utils'
-import { useLoadFromEndpoint } from 'api-client'
 import CityModelBuilder from '../../../api-client/src/testing/CityModelBuilder'
 import buildConfig from '../constants/buildConfig'
 import { ThemeProvider } from 'styled-components'
 import { RenderResult } from '@testing-library/react'
+import {
+  mockUseLoadFromEndpointOnceWitData,
+  mockUseLoadFromEndpointWitData
+} from '../../../api-client/src/testing/mockUseLoadFromEndpoint'
```
klinzo, line 13:
```suggestion
} from 'api-client/src/testing/mockUseLoadFromEndpoint'
```

native/src/utils/AppSettings.ts
```diff
@@ -36,22 +35,28 @@ class AppSettings {
   }
 
   loadSettings = async (): Promise<SettingsType> => {
-    const settingsKeys = Object.keys(defaultSettings)
+    const settingsKeys = Object.keys(defaultSettings) as [keyof SettingsType]
     const settingsArray = await this.asyncStorage.multiGet(settingsKeys)
-    return mapValues(fromPairs(settingsArray), (value, key) => {
+    const settings = fromPairs(settingsArray) as Record<keyof SettingsType, string | null>
+    return mapValues(settings, (value: string | null, key) => {
```
klinzo, line 41: This does not work right?
```suggestion
    return mapValues(settings, (value: string | null, key: SettingsType) => {
```

native/src/redux/reducers/morphContentLanguage.ts
```diff
@@ -43,8 +53,8 @@ const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, city: str
     return route
   }
 
-  const resultModels = {}
-  const resultChildren = {}
+  const resultModels = {} as Record<PathType, CategoryModel>
+  const resultChildren = {} as Record<PathType, ReadonlyArray<PathType>>
```
klinzo, line 57:
```suggestion
  const resultModels: Record<PathType, CategoryModel> = {}
  const resultChildren: Record<PathType, ReadonlyArray<PathType>> = {}
```

klinzo (commented):
Well done, nice PR :tada:

klinzo (approved):

native/src/components/__tests__/ProgressSpinner.spec.tsx
maxammann: I think it is fine to redo this in every test. Then the code which is executed is not split across files. In tests I kind of like repetition.

maxammann (approved):

native/src/components/__tests__/ProgressSpinner.spec.tsx
Taggotty: I also thought about that. But that would mean quite some refactoring for all tests, as all `getByText` calls would also need to change. And I am not sure whether that always makes sense. I will create an issue and we can then discuss this in our next grooming :)

native/src/__tests__/NavigatorContainer.spec.tsx
f1sh1918: I'm fine with that :)

native/package.json
f1sh1918: i'm gonna check that today and give you feedback

native/src/utils/AppSettings.ts
Taggotty: Nope. As map Values explicitly infers key as string

tsconfig.base.json
Taggotty: This one is in the root `tsconfig.base.json`. I removed the other ones to reduce redundancy. If we want to keep this strict typing we should keep `noImplicitAny` as true.

web/src/routes/__tests__/SearchPage.spec.tsx
klinzo: I also wondered about that already, not sure :/

native/package.json
klinzo: Reverted, will upgrade libs in separate PR

web/src/routes/__tests__/SearchPage.spec.tsx
f1sh1918: gonna be resolved in separate issue :)

PR #386, opened 2021-08-02
IGAPP-465: E2E web content search

SteffiStuffel:


e2e-tests/web/test/pageobjects/dashboard.page.ts
```diff
@@ -0,0 +1,16 @@
...
+class DashboardPage extends Page {
+  public language: string
+
+  public constructor(language = 'en') {
+    super()
+    this.language = language
+  }
+
+  get searchIcon() {
+    return $('//header/div/div[4]/div[1]/a')
```
SteffiStuffel, line 12: @Taggotty is this fine or is there a better way to get this icon. I tried to use Appium Image selector, but could not get this working.
https://webdriver.io/docs/selectors/#appium-image-selector

e2e-tests/web/test/utils/navigationUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+  await search.click()
+  await browser.keys(filter)
+
+  const filteredCity = await LandingPage.city(defaultCity)
+  await filteredCity.click()
+
+  const dashboardUrl = await browser.getUrl()
+  const parsedDashboardUrl = new URL(dashboardUrl)
+
+  expect(parsedDashboardUrl.pathname).toContain(dashboardPath)
```
klinzo, line 20: Why not
```suggestion
  expect(parsedDashboardUrl.pathname).toBe(dashboardPath)
```

e2e-tests/web/test/specs/navigateToSearchResults.e2e.ts
```diff
@@ -0,0 +1,29 @@
...
+  it('should open and search content', async () => {
+    const searchPath = Routes.search
+    await navigateToDashboard()
+
+    const searchIcon = await DashboardPage.searchIcon
+    await searchIcon.click()
+
+    const searchUrl = await browser.getUrl()
+    const parsedSearchUrl = new URL(searchUrl)
+    expect(parsedSearchUrl.pathname).toContain(searchPath)
```
klinzo, line 17: Same as below

e2e-tests/web/test/specs/navigateToSearchResults.e2e.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const searchBar = await SearchPage.search
+    expect(searchBar).toBeDefined()
+    await browser.keys(contentSearch)
+
+    const firstResult = await $('=Language')
+    const secondResult = await $('=Language courses')
+
+    expect(firstResult).toBeDefined()
+    expect(secondResult).toBeDefined()
```
klinzo, line 27: I think it would be nice if we could open a search result. Also you could assert that the url contains the query.

klinzo (approved):

e2e-tests/web/test/utils/navigationUtils.ts
SteffiStuffel: I just reused the code from the dashboardTest. But `toBe` also works.

e2e-tests/web/test/specs/navigateToSearchResults.e2e.ts
SteffiStuffel: I added opening a search result. 
I also thought about the query, but decided not to add this, as I assumed e2e-tests are for main user-stories. If the url contains the query is more something for a unit test, isn't it?

f1sh1918 (approved):

e2e-tests/web/test/pageobjects/dashboard.page.ts
Taggotty: it makes sense that the Appium Image selectors do not work, as you are writing web e2e tests. Appium is the native e2e backend. For web you can use this query: `//header//a[@aria-label='Suchen']`. For xPath queries open the browser navigate to the page where you need to query and you can test your xPath query with x('<your xPathQuery') -> https://devhints.io/xpath

Taggotty (approved):

e2e-tests/web/test/pageobjects/dashboard.page.ts
SteffiStuffel: Thanks for explaining :)
Fixed it and now it's a lot nicer :+1:

PR #390, opened 2021-08-06
IGAPP-729: Support .jpeg images

klinzo:

Not sure what to do about .gifs. Our image view does apparently not support gifs. I think there are three possibilities:
1. Keep the current behavior, i.e. just opening .gifs in the browser
2. Do nothing on click
3. Change to a different image view

I would not do 3. as the benefit is very little. What do you think?


native/src/routes/ImageViewModal.tsx
```diff
@@ -1,39 +1,40 @@
...
 import { View } from 'react-native'
 import { ImageViewer } from 'react-native-image-zoom-viewer'
 import { NavigationPropType, RoutePropType } from '../constants/NavigationTypes'
-import withTheme from '../hocs/withTheme'
-import { ThemeType } from 'build-configs'
 import { ImageViewModalRouteType } from 'api-client'
+import React, { ReactElement, useContext } from 'react'
+import { ThemeContext } from 'styled-components'
+
+export const SUPPORTED_IMAGE_FILE_TYPES = ['.jpg', '.jpeg', '.png']
```
f1sh1918, line 8: if the supported image file types gonna be use in several files (like above navigateToLink), it should be in a shared constants file

SteffiStuffel (approved):
Looks good. Created a new Test Image page.
https://webnext.integreat.app/testumgebung/de/bilder-test

Works fine. 

Also created most awesome gif ever :D

f1sh1918 (approved):

Taggotty:
> Not sure what to do about .gifs. Our image view does apparently not support gifs. I think there are three possibilities:
> 
> 1. Keep the current behavior, i.e. just opening .gifs in the browser
> 2. Do nothing on click
> 3. Change to a different image view

It actually should work. But you need to enable gifs on android https://stackoverflow.com/a/57026717

Taggotty (approved):

maxammann (approved):
About GIFs: Opening in Browser is nice to have. I would actually go as far as not supporting GIFs. With GIF support there also comes the question about .webm support and .mp4 support (most GIFs nowadays use modern formats).

Also encouraging GIFs would increase the download size. At least I would not cache GIFs offline.

PR #391, opened 2021-08-10
IGAPP-378: Remove RTL workaround

Taggotty:


maxammann (approved):

native/src/components/__tests__/NewsDetail.spec.tsx
```diff
@@ -46,7 +48,7 @@ describe('NewsDetail', () => {
     fireEvent.press(getByText('mailto:app@integreat-app.de'))
     expect(navigateToLink).toHaveBeenCalledWith('mailto:app@integreat-app.de', language, 'mailto:app@integreat-app.de')
     fireEvent.press(getByText('https://integreat.app'))
-    expect(navigateToLink).toHaveBeenCalledWith('https://integreat.app', language, 'https://integreat.app')
+    expect(navigateToLink).toHaveBeenCalledWith('https://integreat.app/', language, 'https://integreat.app/')
```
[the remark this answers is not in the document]
Taggotty: no. this is react-native-render-html stuff. But after some testing I found out that it only appends the slash to the domain and not the path:
https://integreat.app -> https://integreat.app/
https://integreat.app/testumgebung -> https://integreat.app/testumgebung

I guess they now normalize the path.

native/src/components/TimeStamp.tsx
```diff
@@ -14,36 +14,38 @@ const TimeStampText = styled.Text`
...
+  showText = true,
+  format = 'LL'
+}: PropsType): ReactElement => {
+  const { i18n, t } = useTranslation('common')
+  // only show day, month and year
+  const dateText = formatter.format(lastUpdate, {
+    format
+  })
+  return (
+    <DirectionContainer theme={theme} language={i18n.language}>
```
Taggotty, line 44: passing `i18n.language` will fix the alignment of Last Update, when system language is ltr and content language is rtl.

klinzo:
Where can we test this?

klinzo (approved):
Tested on android, works nicely :) Last update also fixed...

native/src/components/TimeStamp.tsx
klinzo: @Taggotty could you check the news list as well please? Seems like we have the same problem there on native

Taggotty: Hmmm. Cannot reproduce this. Does it only happen for local news?

klinzo: Don't know, then just ignore it for now and I will look into it. Thx for checking :)

PR #394, opened 2021-08-15
IGAPP-568: Upgrade async storage to react-native-async-storage version

klinzo:


f1sh1918 (approved):

.circleci/config.yml
```diff
@@ -139,15 +139,15 @@ commands:
         steps:
             - restore_cache:
                 keys:
-                    - 4-pods-{{ arch }}-{{ checksum "native/ios/Podfile.lock" }}
-                    - 4-pods-{{ arch }}-
+                    - 5-pods-{{ arch }}-{{ checksum "native/ios/Podfile.lock" }}
+                    - 5-pods-{{ arch }}-
```
Taggotty, line 143: What is the prefix actually for? 😅

Znippet (approved):

Taggotty (approved):

.circleci/config.yml
klinzo: To force completely reinstalling all pods. The key is different then and no old cache will be used.

PR #395, opened 2021-08-16
IGAPP-570: City bounding box

klinzo:


web/src/CityContentSwitcher.tsx
```diff
@@ -140,34 +140,63 @@ const CityContentSwitcher = ({ cities, match, location }: PropsType): ReactEleme
 
   const routes: ReactElement[] = []
   if (eventsEnabled) {
-    routes.push(<Route render={render(EVENTS_ROUTE, EventsPage)} path={RoutePatterns[EVENTS_ROUTE]} exact />)
+    routes.push(
+      <Route key={EVENTS_ROUTE} render={render(EVENTS_ROUTE, EventsPage)} path={RoutePatterns[EVENTS_ROUTE]} exact />
```
klinzo, line 144: Only added keys to all routes to fix react error.

f1sh1918 (approved):

api-client/src/Endpoint.ts
```diff
@@ -83,6 +84,7 @@ class Endpoint<P, T> {
       const fetchedData = this.mapResponse(json, params)
       return new Payload(false, url, fetchedData, null)
     } catch (e) {
+      console.error(e)
```
Taggotty, line 87: Is it best practice to log the error? Just asking :)

api-client/src/models/CityModel.ts
```diff
@@ -32,13 +27,8 @@ class CityModel {
...
     longitude: number | null
-    aliases: Record<
-      string,
-      {
-        longitude: number
-        latitude: number
-      }
-    > | null
+    aliases: Record<string, { longitude: number; latitude: number }> | null
+    boundingBox: { latitude: number; longitude: number }[] | null
```
Taggotty, line 31: Could you extract the bounding box type here? You seem to reuse `{ latitude: number; longitude: number }` quite often

Taggotty (approved):

api-client/src/models/CityModel.ts
f1sh1918: Another question is if we should not use the standard format for a bounding box refer to geoJSON
```js
/**
 * Bounding box
 * https://tools.ietf.org/html/rfc7946#section-5
 */
export type BBox = [number, number, number, number] | [number, number, number, number, number, number];
```

api-client/src/Endpoint.ts
klinzo: Not sure but I definitely find it useful. It is stripped in production anyway.

api-client/src/models/CityModel.ts
klinzo: @f1sh1918 sorry I don't have time to adjust this right now, but feel free to change this :)

api-client/src/endpoints/createCitiesEndpoint.ts
```diff
@@ -36,7 +37,10 @@ export default (baseUrl: string): Endpoint<void, Array<CityModel>> =>
               prefix: city.prefix,
               longitude: city.longitude,
               latitude: city.latitude,
-              aliases: city.aliases
+              aliases: city.aliases,
+              boundingBox: Array.isArray(city.bounding_box)
+                ? (city.bounding_box.reduce((acc, val) => acc.concat(val), []) as BBox)
```
f1sh1918, line 42: Maybe someone can take a short look how to properly type this. Couldn't find a proper solution instantly

api-client/src/endpoints/createCitiesEndpoint.ts
```diff
@@ -36,7 +37,10 @@ export default (baseUrl: string): Endpoint<void, Array<CityModel>> =>
               prefix: city.prefix,
               longitude: city.longitude,
               latitude: city.latitude,
-              aliases: city.aliases
+              aliases: city.aliases,
+              boundingBox: Array.isArray(city.bounding_box)
+                ? (city.bounding_box.reduce((minPoint, maxPoint) => [...minPoint, ...maxPoint]) as BBox)
```
f1sh1918, line 42: if someone finds a good way to type it, feel free :)

maxammann (approved):

Taggotty (approved):

klinzo:
@f1sh1918 if you want you can merge this. Added one minor optional suggestion :)

api-client/src/endpoints/createCitiesEndpoint.ts
f1sh1918: I think you mean [1] in the second case

klinzo: I do :grin:

klinzo: If you feel like the current solution is easier/more intuitive just keep it, just stumbled upon it. I also think you are working too late ^^

f1sh1918: no i thought omg why was your solution not my first idea...

PR #396, opened 2021-08-16
IGAPP-730: Report bad cms responses to sentry

klinzo:


web/src/components/FailureSwitcher.tsx
```diff
@@ -13,6 +22,12 @@ type PropsType = {
 const FailureSwitcher = ({ error }: PropsType): ReactElement => {
   const { t } = useTranslation('error')
 
+  useEffect(() => {
+    if (!(error instanceof NotFoundError) && !(error instanceof FetchError)) {
```
f1sh1918, line 26: why we can not use your created custom hook? useReportsError

f1sh1918:
```suggestion
   useReportError(error)
```

f1sh1918 (approved):

web/src/components/FailureSwitcher.tsx
klinzo: I only added that hook for native as it is only used once in web.

native/src/sagas/watchFetchCities.ts
```diff
@@ -16,6 +17,7 @@ export function* fetchCities(dataContainer: DataContainer, action: FetchCitiesAc
     yield* put(insert)
   } catch (e) {
     console.error(e)
+    reportError(e)
```
Znippet, line 20: As I for my part don't like the console.errors in the unit tests much perhaps we could use the new reportError hook for this also and make it configurable?

I had a quick look, seems there are mor places where we have a console.error and the new reportError Hook at the same time.

Locally (developing) we don't need the Sentry report and on the other hand the console.errors aren't used for Sentry supervision.

Only an idea - and if we do it probalby a new task would make sense

Znippet (approved):

native/src/sagas/watchFetchCities.ts
klinzo: Sure, could you create a task? We could also suppress console errors in unit tests by replacing it with a mock I guess.

PR #398, opened 2021-08-16
IGAPP-731: Added geolocation support for web

f1sh1918:

I just played a little bit around with geolocation ;-)
We can think about if it makes sense or just close this pr

You have to use this command to enable geolocation on localhost
```sh
yarn start:https
```
https://0.0.0.0:9000/testumgebung/de/locations


web/package.json
```diff
@@ -10,6 +10,7 @@
     "webpack-ts": "cross-env TS_NODE_PROJECT=\"tools/tsconfig.webpack.json\" webpack --config tools/webpack.config.ts",
     "webpack-dev-server-ts": "cross-env TS_NODE_PROJECT=\"tools/tsconfig.webpack.json\" webpack-dev-server --config tools/webpack.config.ts --progress",
     "start": "yarn run webpack-dev-server-ts --env.config_name=integreat-test-cms --env.dev_server",
+    "start:https": "yarn run webpack-dev-server-ts --https --env.config_name=integreat-test-cms --env.dev_server",
```
maxammann, line 13: I think it would be easier to run "start:malte --https".

To make it easy to remember this we could add the following text to the error message if enabling geolocation fails:

"To use geolocation in a development build you have to start the dev server with "--https" like "yarn start --https"

maxammann (approved):

web/package.json
f1sh1918: I agree. I just think we would have to add additional code which we don't need. I don't see any function in the API doc which handles this. We could add an useEffect and try to get the navigator.location and if that fails we throw an error, but i don't know if we should add this just to get error message

f1sh1918: i added some console.log and removed script

klinzo (approved):
Tested, nice!

PR #399, opened 2021-08-17
IGAPP-732: Dari error

klinzo:

See the issue for more details and how to test this.


f1sh1918 (approved):

f1sh1918:
Tested on IOS and android with persian dari

f1sh1918 (approved):

native/src/utils/NativeLanguageDetector.ts
```diff
@@ -1,17 +1,28 @@
 import { getLocales } from 'react-native-localize'
+import { config } from 'translations'
 
 export default {
```
Taggotty, line 4: Apparently you could also type this as `LanguageDetectorModule` now. Then you can also see that you can omit `async: false`, as there is a separate `LanguageDetectorAsyncModule type. But also feel free to keep it as it is :)

Taggotty (approved):
Tested on emulator.

PR #400, opened 2021-08-23
Igapp 740 poi from detail view

f1sh1918:

You can test Link from DetailPage:
http://0.0.0.0:9000/testumgebung/de/locations/cafe-tuer-an-tuer

or open with queryParams directly:
http://0.0.0.0:9000/testumgebung/de/locations?id=4

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -99,6 +99,7 @@
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^6.4.1",
     "cross-env": "^7.0.3",
+    "csstype": "^3.0.8",
     "css-loader": "5.1.1",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.6",
--- a/web/src/components/MapView.tsx
+++ b/web/src/components/MapView.tsx
@@ -34,6 +34,19 @@ const layerStyle: LayerProps = {
   paint: {}
 }
 
+const renderPopup = (coordinates: Position, description: string): ReactElement => {
+  return (
+    <StyledPopup
+      longitude={coordinates[0]}
+      latitude={coordinates[1]}
+      closeButton={false}
+      closeOnClick={false}
+      anchor='bottom'>
+      <div dangerouslySetInnerHTML={{ __html: description }} />
+    </StyledPopup>
+  )
+}
+
 const getPopUpDescription = (properties: GeoJsonProperties): string => {
   return `<div><strong>${properties?.title}</strong></div>`
 }
@@ -49,17 +62,15 @@ interface MapViewProps {
 const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): ReactElement => {
   const [viewport, setViewport] = useState(defaultViewportConfig)
   const [showPopup, togglePopup] = React.useState(false)
-  const [popUp, setPopup] = React.useState<{ coordinates: Position; description: string }>({
-    coordinates: [],
-    description: ''
-  })
+  const [popUp, setPopup] = React.useState<{ coordinates?: Position; description?: string }>()
+
   const { featureCollection } = props
   const query = new URLSearchParams(useLocation().search)
   const geoId = Number(query.get(mapParam))
 
   useEffect(() => {
     const item: Feature<Point, GeoJsonProperties> | undefined = featureCollection.features.find(
-      feature => feature?.properties?.id === geoId
+      feature => feature.properties?.id === geoId
     )
     if (item?.geometry?.coordinates) {
       const { geometry, properties } = item
@@ -73,23 +84,6 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
       togglePopup(true)
     }
   }, [featureCollection.features, geoId])
-
-  const renderPopup = (coordinates?: Position, description?: string): ReactElement | null => {
-    if (!coordinates || !description) {
-      return null
-    }
-
-    return (
-      <StyledPopup
-        longitude={coordinates[0]}
-        latitude={coordinates[1]}
-        closeButton={false}
-        closeOnClick={false}
-        anchor='bottom'>
-        <div dangerouslySetInnerHTML={{ __html: description }} />
-      </StyledPopup>
-    )
-  }
 
   const clickItem = (e: MapEvent) => {
     if (e.features) {
@@ -117,7 +111,7 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
         onClick={e => clickItem(e)}>
         <Source id='location-pois' type='geojson' data={featureCollection}>
           <Layer {...layerStyle} />
-          {showPopup && renderPopup(popUp.coordinates, popUp.description)}
+          {showPopup && popUp?.coordinates && popUp?.description && renderPopup(popUp.coordinates, popUp.description)}
         </Source>
       </ReactMapGL>
     </MapContainer>
--- a/web/src/components/MapView.tsx
+++ b/web/src/components/MapView.tsx
@@ -1,13 +1,19 @@
-import React, { ReactElement, useState } from 'react'
-import ReactMapGL, { Layer, LayerProps, Source } from 'react-map-gl'
+import React, { ReactElement, useState, useEffect } from 'react'
+import ReactMapGL, { Layer, LayerProps, MapEvent, Popup, Source } from 'react-map-gl'
 import styled from 'styled-components'
 import 'maplibre-gl/dist/maplibre-gl.css'
-import { defaultViewportConfig, mapConfig } from 'api-client'
-import { FeatureCollection } from 'geojson'
+import { defaultViewportConfig, detailZoom, mapConfig, mapParam } from 'api-client'
+import { FeatureCollection, Feature, Point, GeoJsonProperties, Position } from 'geojson'
+import { useLocation } from 'react-router-dom'
+import { Property } from 'csstype'
 
 const MapContainer = styled.div`
   display: flex;
   justify-content: center;
+`
+
+const StyledPopup = styled(Popup)`
+  width: 250px;
 `
 
 const textOffsetY = 1.25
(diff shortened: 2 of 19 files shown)


web/package.json
```diff
@@ -99,6 +99,7 @@
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^6.4.1",
     "cross-env": "^7.0.3",
+    "csstype": "^3.0.8",
```
f1sh1918, line 102: Added for proper checking css types outside of styled components

web/src/routes/PoisPage.tsx
```diff
@@ -107,15 +108,25 @@ const PoisPage = ({ match, cityModel, location, languages, history }: PropsType)
           title={title}
           formatter={formatter}
           onInternalLinkClick={history.push}>
-          {location.location && <PageDetail identifier={t('location')} information={location.location} />}
+          {location.location && (
+            <PageDetail
+              identifier={t('location')}
+              information={location.location}
+              mapLink={
+                featureLocation ? `${path.substring(0, path.lastIndexOf('/'))}?${mapParam}=${location.id}` : undefined
```
f1sh1918, line 116: Maybe there is a better way like a build-in js-fkt to do that
I think big libs like that we don't need:
https://www.npmjs.com/package/query-string

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,92 @@ const layerStyle: LayerProps = {
...
+      <div dangerouslySetInnerHTML={{ __html: description }} />
+    </StyledPopup>
+  )
+}
+
+const getPopUpDescription = (properties: GeoJsonProperties): string => {
+  return `<div><strong>${properties?.title}</strong></div>`
+}
+
+const getCursor = ({ isHovering, isDragging }: { isHovering: boolean; isDragging: boolean }): Property.Cursor => {
```
Taggotty, line 54: `Property.Cursor` is the same as `React.CSSProperties['cursor']`
I don't think we need the library here...

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,92 @@ const layerStyle: LayerProps = {
...
   return (
     <MapContainer>
-      <ReactMapGL {...viewport} onViewportChange={setViewport} mapStyle={mapConfig.styleJSON}>
+      <ReactMapGL
+        interactiveLayerIds={[layerStyle.id!]}
+        getCursor={getCursor}
+        {...viewport}
+        onViewportChange={setViewport}
+        mapStyle={mapConfig.styleJSON}
+        onClick={e => clickItem(e)}>
```
Taggotty, line 117:
```suggestion
        onClick={clickItem}>
```

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,92 @@ const layerStyle: LayerProps = {
...
+      features: FeatureCollection<Point, GeoJsonProperties>,
+      queryId: number
+    ): Feature<Point, GeoJsonProperties> | undefined => {
+      return featureCollection.features.find(feature => feature.properties?.id === queryId)
+    },
+    [featureCollection.features]
+  )
+
+  useEffect(() => {
+    const currentPoi: Feature<Point, GeoJsonProperties> | undefined = getCurrentPoi(featureCollection, queryId)
```
Taggotty, line 82: Explicit typing not needed here
```suggestion
    const currentPoi = getCurrentPoi(featureCollection, queryId)
```

Taggotty, line 82: I don't really see a reason to create a separate function for getting POIs here. As it would recteate the useEffect and the useCallback both when the features update and you also only use it once. See https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

web/src/routes/PoisPage.tsx
Taggotty: you cuold actually say: `${createPath(POIS_ROUTE, { cityCode, languageCode })}?${mapParam}=${location.id}`

Taggotty (changes requested):
I really prefer the usage of `Feature<Point>` or `FeatureCollection<Point>` over `Feature<Point, GeoJsonProperties>` and `FeatureCollection<Point, GeoJsonProperties>`. It's just much shorter and `GeoJsonProperties` are either way the default generic that is set.

Furthermore I would love if you could also "unclick" the popup, by clicking anywhere on the map.

f1sh1918:
> I really prefer the usage of `Feature<Point>` or `FeatureCollection<Point>` over `Feature<Point, GeoJsonProperties>` and `FeatureCollection<Point, GeoJsonProperties>`. It's just much shorter and `GeoJsonProperties` are either way the default generic that is set.
> 
> Furthermore I would love if you could also "unclick" the popup, by clicking anywhere on the map.

Hm the unclicking should work already, if you click somewhere on the map. Gonna check that

web/src/components/MapView.tsx
f1sh1918: well i think its not as good as the csstype lib since the typing is quite poor. CSSType offers way more exact types f.e. unions instead of strings. Since its a dev dependency it won't increase any performance or package size. We can discuss that tommorow :)

web/src/components/MapView.tsx
f1sh1918: yes you are right :) i mean i could use the fkt outside of the functional component and pass values but indeed its not reused anywhere (yet)

klinzo (changes requested):
Tested on firefox, works like a charm :tada:

klinzo:
I get the following error in the console:
> Error: The layer 'point' does not exist in the map's style and cannot be queried for features.

f1sh1918:
> I get the following error in the console:
> 
> > Error: The layer 'point' does not exist in the map's style and cannot be queried for features.
> 
> There also lots of warnings:
> 
> > Image "running_11" could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.
> 
> > WebGL warning: texImage: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.

1.I really couldn't figure out why sometimes the layer can not be found, maybe i create a check before.
2. Is because we need a proper sprite and a new style.json in general but i would like to put this onhold until we know more about design
3. i never faced this error/warning

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,82 @@ const layerStyle: LayerProps = {
...
+    }
+  }
 
   return (
     <MapContainer>
-      <ReactMapGL {...viewport} onViewportChange={setViewport} mapStyle={mapConfig.styleJSON}>
+      <ReactMapGL
+        interactiveLayerIds={[layerStyle.id!]}
+        getCursor={getCursor}
+        {...viewport}
```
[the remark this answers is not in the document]
f1sh1918: 1. Hm well we could make this. I even saw this in all the examples provided in the api doc used like that. A drawback would be that  you are at least updating long,lat and zoom together in the most cases and then you have to call setState three times. 
The only use case would be if we need to use the viewport object in a hook as a dependency.

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,82 @@ const layerStyle: LayerProps = {
...
 
+const renderPopup = (coordinates: Position, description: string): ReactElement => {
+  return (
+    <StyledPopup
+      longitude={coordinates[0]}
+      latitude={coordinates[1]}
+      closeButton={false}
+      closeOnClick={false}
+      anchor='bottom'>
+      <div dangerouslySetInnerHTML={{ __html: description }} />
```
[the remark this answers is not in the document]
maxammann: description should just be a plain string and not html I think

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,82 @@ const layerStyle: LayerProps = {
   paint: {}
 }
 
+const renderPopup = (coordinates: Position, description: string): ReactElement => {
```
maxammann, line 37: This is bascially a component, so naming should be: `MapPopup`

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,82 @@ const layerStyle: LayerProps = {
...
+      latitude={coordinates[1]}
+      closeButton={false}
+      closeOnClick={false}
+      anchor='bottom'>
+      <div dangerouslySetInnerHTML={{ __html: description }} />
+    </StyledPopup>
+  )
+}
+
+const getPopUpDescription = (properties: GeoJsonProperties): string => {
```
maxammann, line 50:
```suggestion
const PopupDescription = (properties: GeoJsonProperties): string => {
```

maxammann (approved):

web/src/routes/PoisPage.tsx
```diff
@@ -107,15 +108,27 @@ const PoisPage = ({ match, cityModel, location, languages, history }: PropsType)
...
           formatter={formatter}
           onInternalLinkClick={history.push}>
-          {location.location && <PageDetail identifier={t('location')} information={location.location} />}
+          {location.location && (
+            <PageDetail
+              identifier={t('location')}
+              information={location.location}
+              mapLink={
+                featureLocation
+                  ? `${createPath(POIS_ROUTE, { cityCode, languageCode })}?${mapParam}=${location.id}`
```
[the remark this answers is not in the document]
Taggotty: Yep. Could also be used in the `SearchPage` I think.

web/src/components/MapView.tsx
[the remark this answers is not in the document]
f1sh1918: True. Just reverted some fancy pictures included ;)

Taggotty (approved):

web/src/components/MapView.tsx
```diff
@@ -28,19 +34,82 @@ const layerStyle: LayerProps = {
...
   const { featureCollection } = props
+  const query = new URLSearchParams(useLocation().search)
+  const queryId = Number(query.get(mapParam))
+
+  useEffect(() => {
+    const currentPoi = featureCollection.features.find(feature => feature.properties?.id === queryId)
+    if (currentPoi?.geometry?.coordinates) {
+      const { geometry } = currentPoi
+      setViewport({
+        ...defaultViewportConfig,
```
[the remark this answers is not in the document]
f1sh1918: ´´´js
setState(prevState => ({
            ...prevState,
            fName: 'your updated value here'
         }));
´´´
i think this is the way to go

web/src/components/PageDetail.tsx
```diff
@@ -8,15 +8,21 @@ const Identifier = styled.span`
 type PropsType = {
   identifier: string
   information: string
+  mapLink?: string
```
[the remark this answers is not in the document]
f1sh1918: alright i also created missing "pois" part in translations since PoisPage is refering to it but it didn't exist

web/src/components/PageDetail.tsx
```diff
@@ -8,15 +8,21 @@ const Identifier = styled.span`
...
   render(): ReactNode {
-    const { identifier, information } = this.props
+    const { identifier, information, mapLink } = this.props
     return (
       <div>
         <Identifier>{identifier}: </Identifier>
         <span>{information}</span>
+        {mapLink && (
+          <p>
+            <a href={mapLink}>Karte</a>
```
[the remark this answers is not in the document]
f1sh1918: added translations

translations/translations.json
```diff
@@ -5311,5 +5311,15 @@
...
+  },
+  "pois": {
+    "de": {
+      "location": "Standort",
+      "map": "Karte"
+    },
+    "en": {
+      "location": "Location",
+      "map": "Map"
+    }
```
klinzo, line 5323: Only German is enough :) English should be translated by professionals as well
```suggestion
    }
```

web/src/components/MapPopup.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { coordinates, properties } = props
+  return (
+    <StyledPopup
+      longitude={coordinates[0]}
+      latitude={coordinates[1]}
+      closeButton={false}
+      closeOnClick={false}
+      anchor='bottom'>
+      <div>
+        <strong>{properties?.title}</strong>
```
klinzo, line 25: Wouldn't it make more sense to just not show a popup if there is no title? Not sure about that or rather whether it is possible at all that there is no title.

f1sh1918: i mean there is an extra popup ticket where we define whats shown there and what not. it was just to prove right thats its the right popup connected to the marker

PR #401, opened 2021-08-24
IGAPP-537: Move BBox web

f1sh1918:

you can double-check bounding box here, since maybe the points are not set perfectly but for a test its fine
https://geojson.io/#map=11/48.3489/10.7302

```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          10.7377731,
          48.2745231
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          10.9271413,
          48.4473535
        ]
      }
    }
  ]
}
```

### DIFF
--- a/web/src/components/MapView.tsx
+++ b/web/src/components/MapView.tsx
@@ -47,7 +47,7 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
   const [viewport, setViewport] = useState<MapViewViewport>(defaultViewportConfig)
 
   useEffect(() => {
-    setViewport(boundingBox ? moveViewToBBox(boundingBox, defaultViewportConfig) : defaultViewportConfig)
+    boundingBox && setViewport(moveViewToBBox(boundingBox, defaultViewportConfig))
   }, [boundingBox])
 
   return (
--- a/web/src/components/MapView.tsx
+++ b/web/src/components/MapView.tsx
@@ -2,7 +2,7 @@ import React, { ReactElement, useState } from 'react'
 import ReactMapGL, { GeolocateControl, Layer, LayerProps, Source } from 'react-map-gl'
 import styled from 'styled-components'
 import 'maplibre-gl/dist/maplibre-gl.css'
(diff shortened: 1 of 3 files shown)


web/src/components/MapView.tsx
```diff
@@ -30,11 +30,25 @@ const layerStyle: LayerProps = {
...
 }
 
 const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): ReactElement => {
-  const [viewport, setViewport] = useState(defaultViewportConfig)
-  const { featureCollection } = props
+  const { featureCollection, boundingBox } = props
+  const [viewport, setViewport] = useState<MapViewViewport>(defaultViewportConfig)
+
+  useEffect(() => {
+    setViewport(boundingBox ? moveViewToBBox(boundingBox, defaultViewportConfig) : defaultViewportConfig)
```
f1sh1918, line 50: @Taggotty 
and yes i love functions :)

klinzo: I think this will clash with #400 right? Not sure but I feel like we need to encapsulate some logic. I.e. have a component calculating the initial viewport and a child rendering the actual MapView and saving the current viewport. That way, no more `useEffects` should be necessary. But maybe I am just too tired and not thinking correctly anymore :D

f1sh1918: the mistake here is that i have to check if there is a bounding box and then Viewport if not do nothing.
You just have to think about the order doing stuff
1. query param
2. bounding box
3. defaultViewportConfig (which will be deprecated soon when we have all BBoxes)

i don't think it gonna get that complicated but maybe there is a better solution.
lets think about it tommorow

maxammann (approved):

Taggotty (approved):
Sooo checking for the bbox will be done in a separate PR?

klinzo (approved):
I like it :)

PR #403, opened 2021-08-25
IGAPP-746: set textColor to input field

f1sh1918:

Please test on native device, since for the the bug only occurs there.
Hopefully also works on ios native device.

Tested on: Android native device, Emulator(android+ios)

### DIFF
--- a/native/ios/Podfile.lock
+++ b/native/ios/Podfile.lock
@@ -695,9 +695,9 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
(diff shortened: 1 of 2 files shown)


native/ios/Podfile.lock
```diff
@@ -695,9 +695,9 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
+  DoubleConversion: cde416483dac037923206447da6e1454df403714
```
klinzo, line 698: Do you know why this file was changed?

klinzo (approved):
Tested on android, works as expected :+1:

klinzo:
AFAICT via browserstack: iOS seems to work fine before this fix already.

native/ios/Podfile.lock
f1sh1918: pod install creates a new checksum and a new pod version was installed indeed

Taggotty (approved):
Also tested on android. 👍🏼

PR #408, opened 2021-08-26
IGAPP-540: Display Pois on Map - Native

Taggotty:

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react'
 import styled from 'styled-components/native'
-import MapboxGL, { CameraSettings } from '@react-native-mapbox-gl/maps'
-import type { BBox } from 'geojson'
+import MapboxGL, { CameraSettings, SymbolLayerProps } from '@react-native-mapbox-gl/maps'
+import type { BBox, FeatureCollection } from 'geojson'
 import { defaultViewportConfig, mapConfig } from 'api-client'
 
 const MapContainer = styled.View`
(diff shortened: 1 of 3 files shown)


f1sh1918 (approved):

klinzo (approved):
Tested on Android :+1: 

Probably not related to your issue:
After switching the language in the pois route, I got the following error (among others):
> Mapbox error You're calling `queryRenderedFeatures` after the `MapView` was destroyed, were you invoking it after `onDestroy()`? {"level": "error", "message": "You're calling `queryRenderedFeatures` after the `MapView` was destroyed, were you invoking it after `onDestroy()`?", "tag": "Mbgl-NativeMapView"}

Same for methods `moveBy`, `getCameraValues` and `setGestureInProgress`. However, I could not reproduce it afterwards.

native/src/components/MapView.tsx
```diff
@@ -15,11 +15,28 @@ const StyledMap = styled(MapboxGL.MapView)`
...
+    iconAllowOverlap: true,
+    iconIgnorePlacement: true,
+    iconImage: ['get', 'symbol'],
+    textField: ['get', 'title'],
+    textFont: ['Roboto Regular'],
+    textOffset: [0, textOffsetY],
+    textAnchor: 'top',
+    textSize: 12
+  }
 }
```
klinzo, line 35: Just for my understanding: Is this something that will be replaced at a later point by the style spec served from our servers in this issue? Or will this stay? https://issues.tuerantuer.org/browse/IGAPP-535

maxammann: Could be, this is open for discussion I'd say. By moving this into the style.json we could reduce code duplication if we show the map in native and web.

maxammann (approved):

native/src/components/MapView.tsx
Taggotty: I'd say it is better to share what is the same. I'am not sure about the fonts though. As we use different fonts in native and web right?

Taggotty:
> Tested on Android 
> 
> Probably not related to your issue:
> After switching the language in the pois route, I got the following error (among others):
> 
> > Mapbox error You're calling `queryRenderedFeatures` after the `MapView` was destroyed, were you invoking it after `onDestroy()`? {"level": "error", "message": "You're calling `queryRenderedFeatures` after the `MapView` was destroyed, were you invoking it after `onDestroy()`?", "tag": "Mbgl-NativeMapView"}
> 
> Same for methods `moveBy`, `getCameraValues` and `setGestureInProgress`. However, I could not reproduce it afterwards.

Could not reproduce for now.

PR #411, opened 2021-08-31
IGAPP-749: Clean link component refactor

f1sh1918:

Please test Tiles, Links and Footer of the WebApp


web/src/components/CleanLink.tsx
```diff
@@ -1,9 +1,38 @@
...
   text-decoration: none;
 `
+const StyledCleanAnchor = styled.a`
+  color: inherit;
+  text-decoration: none;
+`
+
+interface CleanLinkProps {
+  to: string
+  children: ReactNode
```
klinzo, line 17: Maybe it makes sense to allow an array here as well, then we don't need to wrap everything in a fragment if using this component.

web/src/components/__tests__/CleanLink.spec.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import CleanLink from '../CleanLink'
+import { renderWithRouter } from '../../testing/render'
+import { render } from '@testing-library/react'
+
+describe('CleanLink', () => {
+  const linkInternal = '/augsburg/de/willkommen/willkommen-in-augsburg'
+  const externalLink = 'https://integreat.app/'
```
klinzo, line 8: Shouldn't this be an internal link to the root of integreat, i.e.to the path `/`?

web/src/utils/UrlCheck.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ *
+ * @param link to check whether internal or external using regex
+ */
+export const isExternalUrl = (link: string): boolean => {
+  const domainRe = /^(?:(?:https?|ftp):\/\/)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))(?::\d{2,5})?(?:\/\S*)?$/
+  return domainRe.test(link)
```
klinzo, line 7: What does this check for? That there is a host name? I don't really like this tbh. The behavior is very not clear and it does not work for internal links with a host name, e.g. https://integreat.app/augsburg. Instead I think something like the following makes sense:
```suggestion
const InternalLinkRegexp = new RegExp(buildConfig().internalLinksHijackPattern)

try {
  // Check whether link is a valid URL
  new URL(link)
  return !InternalLinkRegexp.test(link)
} catch (e) {
  // Link is not a valid URL and therefore just a pathname -> internal link
  return false
}
```

klinzo: But maybe I also don't get what that regexp is about :D

web/src/components/__tests__/ListItem.spec.tsx
```diff
@@ -1,18 +1,22 @@
...
+describe('ListItemSpec', () => {
+  const path = 'https://tuerantuer.org'
 
-  it('should render a Link and match snapshot', () => {
-    expect(
-      shallow(<ListItem thumbnail='thumbnail' title='first Event' path={path} isExternalUrl={false} />)
-    ).toMatchSnapshot()
+  it('should not render thumbnail in the ListItem', () => {
+    const { getByRole } = render(<ListItem title='first Event' path={path} />, {
+      wrapper: wrapWithTheme as FunctionComponent
```
klinzo, line 12: Why do we need the typecast? Do we need to do this everytime we use `wrapWithTheme`?

klinzo (changes requested):

web/src/components/__tests__/ListItem.spec.tsx
Taggotty: @f1sh1918 I would extract this in a separate PR and apply this to where neccessary/possible

web/src/testing/wrapWithTheme.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ReactElement } from 'react'
+import { ThemeProvider } from 'styled-components'
+import buildConfig from '../constants/buildConfig'
+
+const wrapWithTheme = ({ children }: { children: ReactElement }): ReactElement => (
+  <ThemeProvider theme={buildConfig().lightTheme}>{children}</ThemeProvider>
+)
+
+export default wrapWithTheme
```
Taggotty, line 9: you could write `wrapWithTheme as FunctionComponent`

web/src/components/CleanLink.tsx
```diff
@@ -1,9 +1,38 @@
...
+        <StyledCleanAnchor href={to} aria-label={ariaLabel} className={className} data-testid={'externalLink'}>
+          {children}
+        </StyledCleanAnchor>
+      ) : (
+        <StyledCleanLink to={to} data-testid={'internalLink'}>
+          {children}
+        </StyledCleanLink>
+      )}
+    </>
+  )
```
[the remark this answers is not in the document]
f1sh1918: hm are there any other opinions :D
Since this is a base component which won't be changed a lot (i guess) i think this is okay like this :)

Taggotty (approved):
Tested on Android

web/src/components/CleanLink.tsx
f1sh1918: not necessary

web/src/utils/UrlCheck.ts
f1sh1918: your solution is fine :) its integrated

web/src/components/__tests__/CleanLink.spec.tsx
f1sh1918: indeed it was a bad example ;)

web/src/utils/UrlCheck.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Function return if path is internal or external by checking if an URL can be created that requires an absolute path
+ * @param link provides absolute or relative path
+ */
+export const isExternalUrl = (link: string): boolean => {
+  const InternalLinkRegexp = new RegExp(buildConfig().internalLinksHijackPattern)
+  try {
+    // Check whether link is a valid URL
+    // eslint-disable-next-line no-new
+    new URL(link)
+    return !InternalLinkRegexp.test(link)
```
klinzo, line 13:
```suggestion
    return true

File: web/src/utils/UrlCheck.ts [hunk 8]
```diff
@@ -0,0 +1,18 @@
+import buildConfig from '../constants/buildConfig'
+
+/**
+ * Function return if path is internal or external by checking if an URL can be created that requires an absolute path
+ * @param link provides absolute or relative path
+ */
+export const isExternalUrl = (link: string): boolean => {
+  const InternalLinkRegexp = new RegExp(buildConfig().internalLinksHijackPattern)
```
klinzo, line 8:
```suggestion
```

klinzo (commented):
You were right, the router cannot handle this. Just remove this check and treat links to integreat.app with host as external urls.

web/src/utils/UrlCheck.ts
f1sh1918: @klinzo you can just do
```js
return !!new URL(link)
```

klinzo (approved):
Tested on firefox (some time ago already though)

PR #413, opened 2021-08-31
IGAPP-536: Center Screen around Current Position

Taggotty:


native/src/components/MapView.tsx
```diff
@@ -46,14 +50,28 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
     zoomLevel: defaultViewportConfig.zoom,
     bounds
   }
+
+  const updateLocationPermission = useCallback((locationPermission: boolean) => {
+    setLocationPermission(locationPermission)
+    setFollowUserLocation(false)
+    if (locationPermission) {
```
f1sh1918, line 57: i would prefer ternary expression shorthand if only one line of code follows
locationPermission && setFollowUserLocation(true)

native/src/components/MapView.tsx
```diff
@@ -46,14 +50,28 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
...
     <MapContainer>
       <StyledMap styleJSON={mapConfig.styleJSON} zoomEnabled>
+        <MapboxGL.UserLocation visible={locationPermission} />
         <MapboxGL.ShapeSource id='location-pois' shape={featureCollection}>
           <MapboxGL.SymbolLayer {...layerProps} />
         </MapboxGL.ShapeSource>
-        <MapboxGL.Camera defaultSettings={defaultSettings} />
+        <MapboxGL.Camera defaultSettings={defaultSettings}
+                         followUserLocation={followUserLocation}
+                         followZoomLevel={16}/>
```
f1sh1918, line 72: please use constant / create new one for zoom level in:
api-client/src/maps/

native/src/components/LocationButton.tsx
```diff
@@ -0,0 +1,58 @@
+import LocationIcon from '../assets/IconPlaceholder.svg'
```
klinzo, line 1: Perhaps rename `LocationButton` to `UserLocationButton` to avoid confusion with other usages of term location? Or something like `RequestLocationPermissionButton`?

native/src/components/MapView.tsx
```diff
@@ -37,6 +38,9 @@ const layerProps: SymbolLayerProps = {
 // Has to be set even if we use map libre
 MapboxGL.setAccessToken(mapConfig.accessToken)
 const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactElement => {
+  const [locationPermission, setLocationPermission] = useState(false)
+  const [followUserLocation, setFollowUserLocation] = useState(false)
```
klinzo, line 42:
```suggestion
  const [locationPermission, setLocationPermission] = useState<boolean>(false)
  const [followUserLocation, setFollowUserLocation] = useState<boolean>(false)
```

native/src/components/MapView.tsx
```diff
@@ -46,14 +50,28 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
     zoomLevel: defaultViewportConfig.zoom,
     bounds
   }
+
+  const updateLocationPermission = useCallback((locationPermission: boolean) => {
+    setLocationPermission(locationPermission)
+    setFollowUserLocation(false)
+    if (locationPermission) {
+      setFollowUserLocation(true)
+    }
```
klinzo, line 59:
```suggestion
    setFollowUserLocation(locationPermission)
```

klinzo: Isn't the `followUserLocation` always the same as `locationPermission`

native/src/components/MapView.tsx
klinzo: Perhaps `locationPermissionGranted` would be a better name

native/src/components/LocationButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+import SimpleImage from './SimpleImage'
+import styled from 'styled-components/native'
+import { TouchableOpacity, View } from 'react-native'
+import { checkLocationPermission, requestLocationPermission } from '../utils/LocationPermissionManager'
+import React, { ReactElement, useCallback, useEffect } from 'react'
+import { RESULTS } from 'react-native-permissions'
+
+const ICON_SIZE = 45
+
+const Circle = styled(View)`
```
klinzo, line 11: I think instead we could just reuse the current button from `CitySelector`, I think that icon is more intuitive and I think it makes sense to use the same design/look in both places.

klinzo (changes requested):
I think the behavior of this is a little bit off if the permission is granted but there is no current user location available. If I press the button, the map is just zoomed into to some random place. From other maps I am used to nothing happening instead of displaying a hint/toast if no location is available. Maybe we could use the snackbar for this?

Or is that something that will be done together with the rest of the permission handling in IGAPP-737?

Tested on Android, works if permission is granted and location is available.

Taggotty:
> I think the behavior of this is a little bit off if the permission is granted but there is no current user location available. If I press the button, the map is just zoomed into to some random place. From other maps I am used to nothing happening instead of displaying a hint/toast if no location is available. Maybe we could use the snackbar for this?


Did you test this on a real device? I had the same issue when I did not set the Location on the emulator.

native/src/components/MapView.tsx
Taggotty: I think not. Sometimes you want to zoom around the map, but the UserLocation should stay visible - thus locationPermission. And if you want to center the camera around the curent position again, you press the button, which will reset the followUserLocation variable.

native/src/components/LocationButton.tsx
Taggotty: Actually I only copy pasted this button 🙈

Taggotty: you mean the location marker? Currently there is only the refresh icon for the location in the `CitySelector`

klinzo: > > I think the behavior of this is a little bit off if the permission is granted but there is no current user location available. If I press the button, the map is just zoomed into to some random place. From other maps I am used to nothing happening instead of displaying a hint/toast if no location is available. Maybe we could use the snackbar for this?
> 
> Did you test this on a real device? I had the same issue when I did not set the Location on the emulator.

Yes, my phone. But it is sometimes a little bit weird as I have no google services. What happens for you if you just disable location on your phone (i.e. not the permission but the setting).

klinzo: Ups, you are right. Then it probably does not make sense...

native/src/components/MapView.tsx
klinzo: Yeah right, but with the current implementation it is always the same, right?

klinzo:
> > I think the behavior of this is a little bit off if the permission is granted but there is no current user location available. If I press the button, the map is just zoomed into to some random place. From other maps I am used to nothing happening instead of displaying a hint/toast if no location is available. Maybe we could use the snackbar for this?
> 
> Did you test this on a real device? I had the same issue when I did not set the Location on the emulator.

Yes, my phone. But it is sometimes a little bit weird as I have no google services. What happens for you if you just disable location on your phone (i.e. not the permission but the setting).

native/src/components/MapView.tsx
Taggotty: it may be. But to reset the camera you have to reset `followUserLocation`, which would lead to a flickering location indicator

klinzo: The suggestion still makes sense though I think :)

native/src/components/LocationButton.tsx
f1sh1918: i also suggested to check for standard button or pass the custom icon convenient to the api reference:
https://github.com/react-native-mapbox-gl/maps/blob/master/docs/UserLocation.md
this was just a reminder :)

SteffiStuffel:
Tested this on Android Emulator:
1) When setting a location and grant access, everything works fine
2) If setting a location and forbid location-permission in the settings, it get the error `Mapbox error: Faild to optain last location update`

Also in our current LocationPermissionManager we do never request LocationsPermissions for fixed cities. Is the app-feature also planned for Malte or Aschaffenburg?

native/src/components/MapView.tsx
```diff
@@ -46,14 +50,27 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
     zoomLevel: defaultViewportConfig.zoom,
     bounds
   }
+
+  const updateLocationPermission = useCallback((locationPermission: boolean) => {
+    setLocationPermissionGranted(locationPermission)
+    setFollowUserLocation(false)
+    locationPermission && setFollowUserLocation(true)
```
maxammann, line 57: I never used && as an alernative for an if. I think this is simpler:

```suggestion
setFollowUserLocation(locationPermission)
```

native/src/components/RequestLocationPermissionButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+import SimpleImage from './SimpleImage'
+import styled from 'styled-components/native'
+import { TouchableOpacity, View } from 'react-native'
+import { checkLocationPermission, requestLocationPermission } from '../utils/LocationPermissionManager'
+import React, { ReactElement, useCallback, useEffect } from 'react'
+import { RESULTS } from 'react-native-permissions'
+
+const ICON_SIZE = 45
+
+const Circle = styled(View)`
```
maxammann, line 11: We could use the FAB from react-native-elements: https://reactnativeelements.com/docs/fab/

maxammann (approved):

native/src/components/MapView.tsx
f1sh1918: short circuit evaluation (&& or ||) is a common thing. i think for one liners not a bad code styling but it can be confusing in some cases.
in this case the question is why we should set the FollowUserLocation to false and then depending on locationPermission to true. setting the state two time should be avoided.

native/src/components/LocationButton.tsx
Taggotty: I checked this one. This will replace the location indicator and not a location button. Same goes for the on press, which is triggered when tapping the userlocation

native/src/components/LocationButton.tsx
Taggotty: I decided to directly use the fitting material icons for this. If we do want custom buttons, we can still change this later.

native/src/components/MapView.tsx
Taggotty: yes. But with the previous setup it wouldn't have worked sadly.  I removed the separate LocationButtonComponent, and moved the permission logic to the mapview. Even though this blows up the map view, I think it becomes clearer, that `followUserLocation` is also changed by the mapview itself now and can influence the camera

native/src/components/LocationButton.tsx
Taggotty: I used the `FloatingActionButton` which @maxammann  recommended.

native/src/components/MapView.tsx
```diff
@@ -46,14 +50,27 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
     zoomLevel: defaultViewportConfig.zoom,
     bounds
   }
+
+  const updateLocationPermission = useCallback((locationPermission: boolean) => {
```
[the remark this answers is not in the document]
Taggotty: is refactored now.

native/src/components/MapView.tsx
Taggotty: Is removed now.

native/src/components/__tests__/Failure.spec.tsx
```diff
@@ -7,27 +7,27 @@ import buildConfig from '../../constants/buildConfig'
 import Failure from '../Failure'
 
 jest.mock('react-i18next')
+jest.mock('styled-components', () => ({
```
klinzo, line 10: Can't you just use `wrapWithTheme` instead of mocking now? This is okay for me as well though.

native/src/components/MapView.tsx
```diff
@@ -47,14 +56,58 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
...
     <MapContainer>
       <StyledMap styleJSON={mapConfig.styleJSON} zoomEnabled>
+        <MapboxGL.UserLocation visible={locationPermissionGranted} />
         <MapboxGL.ShapeSource id='location-pois' shape={featureCollection}>
           <MapboxGL.SymbolLayer {...layerProps} />
         </MapboxGL.ShapeSource>
-        <MapboxGL.Camera defaultSettings={defaultSettings} />
+        <MapboxGL.Camera
+          defaultSettings={defaultSettings}
+          followUserMode={'normal'}
```
klinzo, line 100:
```suggestion
          followUserMode='normal'
```
same below

native/src/components/MapView.tsx
```diff
@@ -47,14 +56,58 @@ const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactEle
...
   }
+
+  const onLocationPermissionRequest = useCallback((locationPermission: PermissionStatus | undefined) => {
+    if (locationPermission === RESULTS.GRANTED) {
+      setFollowUserLocation(true)
+      setLocationPermissionGranted(true)
+    } else {
+      setFollowUserLocation(false)
+      setLocationPermissionGranted(false)
+    }
```
klinzo, line 67:
```suggestion
    const permissionGranted = locationPermission === RESULTS.GRANTED
    setFollowUserLocation(permissionGranted)
    setLocationPermissionGranted(permissionGranted)
```
:smile:

klinzo (approved):
I like it :)

klinzo:
Issue with map jumping to random place if no location is available seems to be fixed as well :+1:

maxammann (approved):

PR #414, opened 2021-09-01
IGAPP-497: Snackbar on Failed Refresh

Taggotty:


Taggotty:
Not sure if that is what this issue wants, but it shows the snackbar if refresh should not be possible.

native/src/routes/Settings.tsx
```diff
@@ -100,14 +98,6 @@ const Settings = ({ navigation, t, languageCode, cityCode, theme }: PropsType):
 
   const ThemedItemSeparator = () => <ItemSeparator theme={theme} />
 
-  const showSnackbar = (key: string) =>
```
Taggotty, line 103: was never used

klinzo (changes requested):
Sadly this is not what the goal of the issue was. If the refresh fails because of no internet connection or for other reasons, the snackbar should be displayed and the old content is still displayed. Not sure how much effort to implement this is though.

Taggotty:
@klinzo thx for clarifying. Then i'll check out this case. The Payload provider really is confusing in this regard.

klinzo:
> @klinzo thx for clarifying. Then i'll check out this case. The Payload provider really is confusing in this regard.

Definitely! That's why there is also one issue to refactor it. Ugly af :vomiting_face: .

klinzo (approved):
Only approved to not block the merge later on once it is finished.

Taggotty:
Didn't get any further than that. Probably needs quite a bit of refactoring.

PR #416, opened 2021-09-02
IGAPP-726: Prettier plugin for import order

SteffiStuffel:

I added the rules so the output is the following:

everything from libraries
(newline)
... from api-client
(newline)
... from build-configs
(newline)
... from translations
(newline)
everything with relative path import

Inside of every block it is ordered alphabetically. 

I just added two files to the commit, so you can see the output, before merging I would apply the rules to every file.


.prettierrc
```diff
@@ -6,5 +6,7 @@
   "printWidth": 120,
   "jsxBracketSameLine": true,
   "jsxSingleQuote": true,
-  "endOfLine": "auto"
+  "endOfLine": "auto",
+  "importOrder": ["^([^(.|api\\-client|build\\-config)])", "api-client", "build-configs", "translations", "^./*"],
```
klinzo, line 10: Don't you miss a `translations` in the first group? Also `^./*`does not match `../my-file` but it seems to work anyway right :thinking: 
```suggestion
  "importOrder": ["^([^(.|api\\-client|build\\-configs)])", "api-client", "build-configs", "translations", "^./*"],
```

klinzo (approved):

.prettierrc
maxammann, line 10: I'm not sure whether we want to add these rules for manual sorting. I suspect that we are missing maybe some edgecases or will miss to adjust this regex as soon as we add a new internal dependency.

For simplicity I would just tread our dependencies as every other dependencies.

maxammann (approved):

.prettierrc
f1sh1918: i think we should differ between 3rd party library like node modules and our symlinks from the monorepo and local imports.

For me it would be okay to add a new regex rules part if we have a new internal lib. I don't think this happens too often.
I like the solution here. i would just avoid to make new lines except for local imports if that it possible

f1sh1918 (approved):

.prettierrc
SteffiStuffel: You are right. I fixed it.

SteffiStuffel: As there are different opinions about this (see discussions/votes on Mattermost), I decided to do it this way:

everything external dependencies
(newline)
our dependencies
(newline)
local imports

SteffiStuffel: If anyone is unhappy with this solution we should talk about this on Wednesday and discuss this together, but I think it does not really matter.

PR #417, opened 2021-09-04
IGAPP-99: Catch possible errors in ImageModal

maxammann:

I think we should remove the image view dependency as it is not updated anymore. I could add a pinch-to-zoom feature quite easily for images what do you think?


klinzo:
I totally agree with you, a pinch-to-zoom feature would be nice.

maxammann:
closed because I fucked up history while merging.

PR #422, opened 2021-09-07
IGAPP-743: Show poi detail

f1sh1918:

**Note:**
Please test on native device, too. I'm facing marker rendering problems (fade to black) only on emulator.
This happens when you click on a poi to get short information


native/src/components/MapView.tsx
```diff
@@ -12,49 +14,79 @@ const MapContainer = styled.View`
...
+  const [activeFeature, setActiveFeature] = useState<Feature<Point> | null>(currentFeature ?? null)
+  const [showPopup, setShowPopup] = useState<boolean>(false)
+  const layerProps: SymbolLayerProps = {
+    id: layerId,
+    style: {
+      symbolPlacement: 'point',
+      iconAllowOverlap: true,
+      iconIgnorePlacement: true,
+      iconImage: ['get', 'symbol'],
+      textField: ['case', ['==', ['get', 'id'], activeFeature?.properties?.id ?? -1], ['get', 'title'], ''],
```
f1sh1918, line 42: I didn't find a better approach than the buildIn expression to render the text field only for activePoi.
This may cause the fading in emulator. On native device everything works fine!

native/src/components/RemoteContent.tsx
```diff
@@ -91,26 +90,30 @@ const RemoteContent = (props: PropType): ReactElement => {
...
-        renderError={renderWebviewError}
-        onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
-        style={{
-          height: webViewHeight,
-          width: webViewWidth
-        }}
-      />
-    </StyledView>
+    <>
+      {!!content.length && (
```
f1sh1918, line 94: I fixed a unneeded rerendering when content.length === 0 --> was triggering the useEffect above

native/src/routes/Pois.tsx
```diff
@@ -118,7 +131,14 @@ const Pois = ({
...
           <>
             {location && (
-              <PageDetail identifier={t('location')} information={location} theme={theme} language={language} />
+              <PageDetail
+                identifier={t('location')}
+                information={location}
+                theme={theme}
+                language={language}
+                linkLabel={poi?.featureLocation && t('map')}
+                onLinkClick={navigateToMap(cityModel.code, language, String(poi.location.id))}
```
klinzo, line 140: A more general question/discussion: I think for UX it is better to have speaking ids in the urls, i.e. to use the poi slug instead of the id, e.g. `agentur-fuer-arbeit-augsburg` instead of `12`. We also did it like that whenever possible in other routes. What do you think?

native/src/routes/Pois.tsx
```diff
@@ -56,8 +59,9 @@ const Pois = ({
 }: PropsType): ReactElement => {
   const { t } = useTranslation('pois')
   const theme = useTheme()
+  const route = useRoute()
```
klinzo, line 62: If you just pass `route` as prop from `PoisContainer`, you don't need this or any type assertions/casts as everything is correctly typed there :)

native/src/routes/Pois.tsx
```diff
@@ -137,14 +157,22 @@ const Pois = ({
   const featureLocations = pois
     .map(poi => poi.featureLocation)
     .filter((feature): feature is Feature<Point> => !!feature)
+  const locId = (route?.params as RouteParams)?.locationId
```
klinzo, line 160: Can be greatly simplified if suggestion from above is applied

native/src/components/RemoteContent.tsx
klinzo: I'd really favor moving the condition out of the return and using `content.length === 0` as check.
```suggestion
  if (content.length === 0) {
    return null
  }
```

native/src/components/MapView.tsx
```diff
@@ -12,49 +14,84 @@ const MapContainer = styled.View`
...
-    iconIgnorePlacement: true,
-    iconImage: ['get', 'symbol'],
-    textField: ['get', 'title'],
-    textFont: ['Roboto Regular'],
-    textOffset: [0, textOffsetY],
-    textAnchor: 'top',
-    textSize: 12
-  }
-}
+const layerId = 'point'
```
klinzo, line 27:
```suggestion
const featureLayerId = 'point'
```

native/src/components/MapView.tsx
```diff
@@ -12,49 +14,84 @@ const MapContainer = styled.View`
...
 
 // Has to be set even if we use map libre
 MapboxGL.setAccessToken(mapConfig.accessToken)
-const MapView = ({ boundingBox, featureCollection }: MapViewPropsType): ReactElement => {
+const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewPropsType): ReactElement => {
+  const mapRef = React.useRef<MapboxGL.MapView | null>(null)
+  const cameraRef = React.useRef<MapboxGL.Camera | null>(null)
+  const [activeFeature, setActiveFeature] = useState<Feature<Point> | null>(currentFeature ?? null)
+  const [showPopup, setShowPopup] = useState<boolean>(!!currentFeature)
```
klinzo, line 35: Do we need this in the state? Can't we just set the active feature to null if the popup is closed?

native/src/components/MapView.tsx
```diff
@@ -12,49 +14,84 @@ const MapContainer = styled.View`
...
     sw: [boundingBox[0], boundingBox[1]]
   }
 
+  // if there is a current feature use the coordinates if not use bounding box
   const defaultSettings: CameraSettings = {
-    zoomLevel: defaultViewportConfig.zoom,
-    bounds
+    zoomLevel: activeFeature?.geometry.coordinates ? detailZoom : defaultViewportConfig.zoom,
+    centerCoordinate: activeFeature?.geometry.coordinates,
+    bounds: activeFeature?.geometry.coordinates ? undefined : bounds
```
klinzo, line 60: before:
``` js
  const coordinates = activeFeature?.geometry?.coordinates
```

```suggestion
    zoomLevel: coordinates ? detailZoom : defaultViewportConfig.zoom,
    centerCoordinate: coordinates,
    bounds: !coordinates && bounds
```

native/src/components/MapView.tsx
```diff
@@ -12,49 +14,84 @@ const MapContainer = styled.View`
...
   }
+
+  const onPress = useCallback(
+    async (e: Feature) => {
```
klinzo, line 64:
```suggestion
    async (pressedLocation: Feature) => {
```
or a similar speaking name please

native/src/components/MapView.tsx
klinzo: I think I would either not show a title at all or do it the other way round, i.e. only not show the title for the selected feature as the title is shown in the popup anyway and I don't see a reason why we need it twice.

native/src/components/MapView.tsx
```diff
@@ -12,49 +14,84 @@ const MapContainer = styled.View`
...
+          geometry: { coordinates }
+        } = features[0]
+        cameraRef?.current?.flyTo([coordinates[0], coordinates[1]])
+      } else {
+        setActiveFeature(null)
+        setShowPopup(false)
+      }
+    },
+    [cameraRef]
+  )
```
klinzo, line 84:
```suggestion
  const onPress = useCallback(async (e: Feature<Geometry, GeoJsonProperties>) => {
    if (!mapRef?.current || !cameraRef?.current || !e.properties) {
      return
    }
    const featureCollection = await mapRef.current.queryRenderedFeaturesAtPoint(
      [e.properties.screenPointX, e.properties?.screenPointY],
      undefined,
      [layerId]
    )
    const feature = featureCollection?.features?.find((it): it is Feature<Point> => it.geometry.type === 'Point')

    if (feature) {
      setActiveFeature(feature)
      const {
        geometry: { coordinates }
      } = feature
      cameraRef.current.flyTo(coordinates)
    } else {
      setActiveFeature(null)
    }
  }, []
```
I think we should avoid typecasts and nullish coalesence operator as much as possible.

klinzo: Or is there any reason why you just included `cameraRef` in dependencies?

klinzo (changes requested):
Works nicely already :tada: I assume text for the popup is added at a later point?

native/src/components/MapView.tsx
f1sh1918: no because the popup has to set true auto when redirecting from details page

klinzo: You sure? It is only ever updated together with activeFeature and its initial value also depends on it.

native/src/components/MapView.tsx
f1sh1918: cameraRef was suggested, i wouldn't have added it

native/src/components/MapView.tsx
klinzo: Just tried it, seems to work without it as well.

native/src/components/MapView.tsx
f1sh1918: no because !coordinates gonna get false

native/src/components/MapView.tsx
f1sh1918: ok then we keep it for now shown for all and we can discuss about this

native/src/components/MapView.tsx
f1sh1918: word.. :)

native/src/components/MapView.tsx
f1sh1918: Nullish coalescing operator is "??" or am i wrong?
What you mean is "Optional Chaining" that means checking for null | undefined of objects which became a ECMA standard. I see in your suggestion the advantage no question, but there are lots of cases where it is the recommend way to go

native/src/components/RemoteContent.tsx
f1sh1918: ìf (content.length)`
should be sufficient

f1sh1918:
> Works nicely already 🎉 I assume text for the popup is added at a later point?

Yes only the title for now. We need further information from cms like pictures
And we need the current location if we want to show there sth like a distance. So i think this pr is fat enough and we make a separate ticket like in the web project

native/src/components/MapView.tsx
klinzo: Yeah right, meant optional chaining, sorry :D

native/src/components/MapView.tsx
```diff
@@ -54,34 +52,33 @@ const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewProp
...
-      }
-    },
-    [cameraRef]
-  )
+  const onPress = useCallback(async (pressedLocation: Feature<Geometry, GeoJsonProperties>) => {
+    if (!mapRef?.current || !cameraRef?.current || !pressedLocation.properties) {
+      return
+    }
+    const featureCollection = await mapRef.current.queryRenderedFeaturesAtPoint(
+      [pressedLocation.properties.screenPointX, pressedLocation.properties?.screenPointY],
```
klinzo, line 67:
```suggestion
      [pressedLocation.properties.screenPointX, pressedLocation.properties.screenPointY],
```

native/src/components/MapView.tsx
```diff
@@ -24,23 +23,22 @@ type MapViewPropsType = {
...
 const textOffsetY = 1.25
-const layerId = 'point'
+const featureLayerId = 'point'
 
 // Has to be set even if we use map libre
 MapboxGL.setAccessToken(mapConfig.accessToken)
 const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewPropsType): ReactElement => {
   const mapRef = React.useRef<MapboxGL.MapView | null>(null)
   const cameraRef = React.useRef<MapboxGL.Camera | null>(null)
   const [activeFeature, setActiveFeature] = useState<Feature<Point> | null>(currentFeature ?? null)
```
klinzo, line 33: Only really minor and not sure about it: Perhaps `selectedFeature` would be better as we use that already most of the time, active is kind of weird I think

native/src/routes/Pois.tsx
```diff
@@ -157,9 +157,9 @@ const Pois = ({
   const featureLocations = pois
     .map(poi => poi.featureLocation)
     .filter((feature): feature is Feature<Point> => !!feature)
-  const locId = (route?.params as RouteParams)?.locationId
+
   const currentFeature: Feature<Point> | undefined = featureLocations.find(
-    feature => feature.properties?.id === Number(locId)
+    feature => feature.properties?.id === Number(route.params.selectedPoiId)
   )
```
klinzo: If you want you can remove the type annotation

klinzo (approved):

native/src/components/MapView.tsx
```diff
@@ -1,60 +1,94 @@
...
+  const coordinates = activeFeature?.geometry?.coordinates
   const defaultSettings: CameraSettings = {
-    zoomLevel: defaultViewportConfig.zoom,
-    bounds
+    zoomLevel: coordinates ? detailZoom : defaultViewportConfig.zoom,
+    centerCoordinate: coordinates,
+    bounds: coordinates ? undefined : bounds
   }
+
+  const onPress = useCallback(async (pressedLocation: Feature<Geometry, GeoJsonProperties>) => {
```
f1sh1918, line 62: @klinzo 
i remember what @Taggotty  said that <Geometry , GeoJsonProperties> are default so i guess we don't have to specify them here. Only to specify if we use a certain Geometry like a "Point" f.e.
I just overseen that in your suggestion

native/src/routes/Pois.tsx
f1sh1918: i mean i don't see the advantages to be honest for mobile? Maybe we can discuss this tomorrow after weekly

native/src/routes/Pois.tsx
f1sh1918: no the the ```properties``` are typed as following by 'geojson'
```[name:string]:any```
In this case ```id : number```
We could override the interface to get proper type checking

native/src/routes/Pois.tsx
klinzo: Question was mostly for web, but its relevant on mobile as well, e.g. for deep linking. Sadly I won't have time for that as I am already on vacation then :/

f1sh1918: I suggest to make a separate ticket for that

Znippet (approved):

PR #423, opened 2021-09-07
IGAPP-468: Web E2E-Test for location change

SteffiStuffel:


e2e-tests/web/test/specs/navigateToNewLocation.e2e.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const locationIcon = await DashboardPage.locationIcon
+    await locationIcon.click()
+
+    const landingUrl = await browser.getUrl()
+    const parsedLandingUrl = new URL(landingUrl)
+    expect(parsedLandingUrl.pathname).toContain(landingPath)
+
+    const filteredCity = await LandingPage.city('Stadt Augsburg')
+    expect(filteredCity).toBeDefined()
```
Znippet, line 20: only :art: : superficient, no need to test this, this will fail on next step with "trying to call click on an undefined object" anyways

Znippet (approved):

f1sh1918 (approved):

e2e-tests/web/test/specs/navigateToNewLocation.e2e.ts
SteffiStuffel: yeah, but on a failing test it is much nicer to get the feedback `filteredCity should be defined but is not`, instead of `undefined is not a object or something`

PR #424, opened 2021-09-07
IGAPP-757: Increase connectionRetryTimeout in E2E Tests

Taggotty:

As always this one cannot guarante, that there will be no more failing tests. But I ran both the ios and the android tests quite often to ensure that they do not time out in 1 of 5 runs, which is about 90% better what we currently have. As I just said increasing the timeout seems to be the most sufficient solution here, as I did try a few other things, too.

### DIFF
--- a/e2e-tests/native/capabilities.ts
+++ b/e2e-tests/native/capabilities.ts
@@ -10,6 +10,7 @@ const browserstackCaps = (
   const isCi = !!process.env.CI
   const prefix = isCi ? 'IG CI' : 'IG DEV'
   const app = process.env.E2E_BROWSERSTACK_APP
+
   return {
     'bstack:options': {
       buildName: `${prefix}: ${getGitBranch()}`,
@@ -17,8 +18,8 @@ const browserstackCaps = (
       projectName: 'integreat-app-native',
       debug: true,
       realMobile: isCi,
-      networkProfile: '4g-lte-good',
-      appiumVersion: '1.21.0'
+      appiumVersion: '1.21.0',
+      idleTimeout: 10000
     },
     ...config,
     'appium:app': app,
@@ -41,7 +42,7 @@ export default {
       'appium:deviceName': 'iPhone 8',
       'appium:automationName': 'XCUITest',
       // @ts-ignore unknown property
-      'appium:waitForQuiescence': true
+      'appium:waitForIdleTimeout': 10000
     },
     'ios'
   )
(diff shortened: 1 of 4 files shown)


e2e-tests/native/capabilities.ts
```diff
@@ -10,15 +10,16 @@ const browserstackCaps = (
...
   const app = process.env.E2E_BROWSERSTACK_APP
+
   return {
     'bstack:options': {
       buildName: `${prefix}: ${getGitBranch()}`,
       sessionName: `${platformName.toLowerCase()}: ${getGitHeadReference()}`,
       projectName: 'integreat-app-native',
       debug: true,
       realMobile: isCi,
-      networkProfile: '4g-lte-good',
```
Taggotty, line 20: was an experiment to get rid of timeouts in Android, which did not work, therefore I removed it.

e2e-tests/native/capabilities.ts
```diff
@@ -41,7 +42,7 @@ export default {
       'appium:deviceName': 'iPhone 8',
       'appium:automationName': 'XCUITest',
       // @ts-ignore unknown property
-      'appium:waitForQuiescence': true
+      'appium:waitForIdleTimeout': 10000
```
Taggotty, line 45: `waitForIdleTimeout` is very similar to `waitForQuiescence` as it waits for the timeout or idle by itself. I replaced those two as webdriver seems to translate unknown booleans to strings -.-

e2e-tests/native/wdio-browserstack.conf.ts
```diff
@@ -13,32 +15,31 @@ const getCapability = () => {
...
   user: process.env.E2E_BROWSERSTACK_USER,
   key: process.env.E2E_BROWSERSTACK_KEY,
 
   capabilities: [getCapability()],
 
   logLevel: 'info',
-  coloredLogs: true,
-  bail: 0,
+
+  bail: 1,
```
Taggotty, line 32: If now one test fails in a suite, the other tests wont be executed,

e2e-tests/native/wdio-browserstack.conf.ts
```diff
@@ -13,32 +15,31 @@ const getCapability = () => {
...
   baseUrl: 'http://localhost:9000',
   waitforTimeout: 100000,
   waitforInterval: 2000,
-  connectionRetryTimeout: 50000,
-  connectionRetryCount: 3,
-  services: [['browserstack']],
-  host: 'hub.browserstack.com',
+  connectionRetryTimeout: 120000,
+  connectionRetryCount: 2,
```
Taggotty, line 37: After a retry, the old session is not deleted or terminated immidiately, which can lead to a hand full of ghost sessions, when we retry to often. Decreasing the idle timeout (see Capabilities) and the connectionRetryCount helps cope with this.

SteffiStuffel (approved):

klinzo (approved):

PR #426, opened 2021-09-08
IGAPP-760-Tu News Preview Bug

f1sh1918:


web/src/utils/stringUtils.ts
```diff
@@ -1,15 +1,12 @@
...
-  if (words.length < numOfWordsAllowed) {
-    return text
+  const cleanText = text.replace(/\n/g, ' ')
+  if (cleanText.length < numOfCharsAllowed) {
+    return cleanText
   }
-
-  const truncatedText = words.splice(0, numOfWordsAllowed).join(' ')
-  return truncatedText + ellipsis
+  return `${cleanText.substring(0, cleanText.lastIndexOf(' ', numOfCharsAllowed))}${ellipsis}`
```
f1sh1918, line 9: It checks now the last empty space back from "numOfCharsAllowed"

Znippet (approved):

SteffiStuffel (commented):
Please add a release note, something like "improved preivew of news.."

And also I guess there shouldn't be a newline now. The ticket says: " the line break should be replaced by an empty space and not totally removed."

Image attached, I got a linebreak and not a space:
IGAPP760

maxammann (approved):

f1sh1918:
> Please add a release note, something like "improved preivew of news.."
> 
> And also I guess there shouldn't be a newline now. The ticket says: " the line break should be replaced by an empty space and not totally removed."
> 
> Image attached, I got a linebreak and not a space:
> IGAPP760

thx steffi for the hint :)
Well the "textTruncator" is not in use for native since there is a ```numberOfLines```property in the native Text Component which is in use. so there is no string adjustment at all
We could adjust it to be more convenient for both applications

Taggotty (approved):

PR #430, opened 2021-09-14
IGAPP-763: Poi short description native

f1sh1918:

@all
I thought about extending the GeoJsonProperties to have some of them required like:
```js
export type GeoPropertiesExtended =  {
  title: string
  path: string
  id: number
}
Feature<Point,GeoPropertiesExtended>
```
So we could avoid lots of null checks


Znippet (approved):

Taggotty (approved):
I think we kind of missed this, because it is not in the designs, but maybe we should also display the excerpt in the poi. But I think we can merge this for now and decide later what we do, maybe after some discussion.

native/src/components/MapView.tsx
```diff
@@ -69,7 +84,6 @@ const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewProp
 
   const onLocationPermissionRequest = useCallback((locationPermission: PermissionStatus | undefined) => {
     const permissionGranted = locationPermission === RESULTS.GRANTED
-    setFollowUserLocation(permissionGranted)
```
klinzo, line 72: Why did you remove this?

native/src/components/MapView.tsx
```diff
@@ -79,7 +93,8 @@ const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewProp
 
   const requestPermission = useCallback(() => {
     requestLocationPermission().then(onLocationPermissionRequest)
-  }, [onLocationPermissionRequest])
+    locationPermissionGranted && setFollowUserLocation(!followUserLocation)
```
klinzo, line 96: What is happening here? Can you add a comment please?

native/src/components/MapView.tsx
```diff
@@ -94,31 +109,46 @@ const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewProp
...
     <MapContainer>
-      <StyledMap styleJSON={mapConfig.styleJSON} zoomEnabled onPress={onPress} ref={mapRef}>
-        <MapboxGL.UserLocation visible={locationPermissionGranted} />
+      <StyledMap
+        styleJSON={mapConfig.styleJSON}
+        zoomEnabled
+        onPress={onPress}
+        ref={mapRef}
+        attributionEnabled={false}
+        logoEnabled={false}>
```
klinzo, line 150: Are we allowed to disable these? https://docs.mapbox.com/help/glossary/attribution/
> You must properly attribute and link Mapbox and OpenStreetMap when using the Mapbox Streets tileset.

native/src/components/MapView.tsx
```diff
@@ -23,7 +23,13 @@ const StyledMap = styled(MapboxGL.MapView)`
 type MapViewPropsType = {
   boundingBox: BBox
   featureCollection: FeatureCollection
-  currentFeature?: Feature<Point>
+  selectedFeature: Feature<Point> | null
+  setSelectedFeature: (feature: Feature<Point> | null) => void
+  navigateTo: (arg0: RouteInformationType) => void
```
klinzo, line 28:
```suggestion
  navigateTo: (routeInformation: RouteInformationType) => void
```

native/src/routes/Pois.tsx
```diff
@@ -41,6 +42,21 @@ export type PropsType = {
   route: RoutePropType<PoisRouteType>
 }
 
+// Calculate distance for all Feature Locations
+const prepareFeatureLocations = (pois: Array<PoiModel>, userLocation?: number[]): Feature<Point>[] => {
```
klinzo, line 46: No need for curly brackets

native/src/routes/Pois.tsx
```diff
@@ -41,6 +42,21 @@ export type PropsType = {
...
+  return pois
+    .map(poi => {
+      const featureLocation: Feature<Point> = poi.featureLocation as Feature<Point>
```
klinzo, line 49: I think we should not need a typecast here if you adjust the type of `poi.featureLocation` to `Feature<Point> | null`, the type annotation is also redundant.
```suggestion
      const featureLocation = poi.featureLocation
```

native/src/routes/Pois.tsx
```diff
@@ -60,6 +76,23 @@ const Pois = ({
 }: PropsType): ReactElement => {
   const { t } = useTranslation('pois')
   const theme = useTheme()
+  const [selectedFeature, setSelectedFeature] = useState<Feature<Point> | null>(null)
+  const [userLocation, setUserLocation] = useState<number[] | undefined>(undefined)
```
klinzo, line 80:
```suggestion
  const [userLocation, setUserLocation] = useState<number[] | null>(null)
```

native/src/components/MapView.tsx
```diff
@@ -94,31 +109,46 @@ const MapView = ({ boundingBox, featureCollection, currentFeature }: MapViewProp
...
-      } = feature
-      cameraRef.current.flyTo(coordinates)
-    } else {
-      setSelectedFeature(null)
-    }
-  }, [])
+  // set the user location coordinate once to calculate distance for all pois
+  const onLocationUpdate = (location: MapboxGL.Location): void => {
+    !userLocation && setUserLocation([location.coords.longitude, location.coords.latitude])
+  }
```
klinzo, line 115: Perhaps it would be better to only update the user location if the change in distance is smaller than `0.1` kilometers.

native/src/routes/Pois.tsx
```diff
@@ -60,6 +76,23 @@ const Pois = ({
...
+  const [featureLocations, setFeatureLocations] = useState<Feature<Point>[]>(prepareFeatureLocations(pois))
+
+  useEffect(() => {
+    if (!path) {
```
klinzo, line 84: What is this check needed for?

native/src/routes/Pois.tsx
```diff
@@ -60,6 +76,23 @@ const Pois = ({
...
+      const featureLocations = prepareFeatureLocations(pois, userLocation)
+      const selectedPoiId = Number(route.params.selectedPoiId)
+      if (selectedPoiId) {
+        const currentFeature: Feature<Point> | undefined = featureLocations.find(
```
klinzo, line 88:
```suggestion
        const currentFeature = featureLocations.find(
```

native/src/routes/Pois.tsx
```diff
@@ -60,6 +76,23 @@ const Pois = ({
...
+          feature => feature.properties?.id === Number(route.params.selectedPoiId)
+        )
+        currentFeature && setSelectedFeature(currentFeature)
```
klinzo, line 91: I think this should always be called, even if no feature is selected/found since otherwise we may show an outdated selection
```suggestion
        setSelectedFeature(currentFeature)
```

native/src/routes/Pois.tsx
```diff
@@ -154,14 +187,6 @@ const Pois = ({
     return <Failure code={fromError(error)} />
   }
 
-  const featureLocations = pois
-    .map(poi => poi.featureLocation)
-    .filter((feature): feature is Feature<Point> => !!feature)
-
-  const currentFeature: Feature<Point> | undefined = featureLocations.find(
-    feature => feature.properties?.id === Number(route.params.selectedPoiId)
-  )
```
klinzo, line 163: Why did you move this to the hook?

api-client/src/endpoints/createEventsEndpoint.ts
```diff
@@ -49,7 +49,9 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<EventModel>> =>
                 region: event.location.region,
                 country: event.location.country,
                 latitude: event.location.latitude,
-                longitude: event.location.longitude
+                longitude: event.location.longitude,
+                thumbnail: event.thumbnail,
+                path: normalizePath(event.path)
```
klinzo, line 54: I think we should not store the same properties twice in the same model. Really confusing and makes you think these are different things when they are not. They also have nothing to do with location.

api-client/src/endpoints/createPOIsEndpoint.ts
```diff
@@ -38,7 +38,9 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<PoiModel>> =>
               region: poi.location.region,
               country: poi.location.country,
               latitude: poi.location.latitude,
-              longitude: poi.location.longitude
+              longitude: poi.location.longitude,
+              thumbnail: poi.thumbnail,
+              path: normalizePath(poi.path)
```
klinzo, line 43: Same as above.

klinzo (changes requested):

native/src/routes/Pois.tsx
f1sh1918: because i set the current selected feature in the lifecycle hook? Why should i call it somewhere else?

api-client/src/endpoints/createEventsEndpoint.ts
f1sh1918: hm well i understand. But you could also say everything that should be shown on the map connected to the poi should be in the geoJSON. So i would have to touch in the frontend again the Feature to merge values

klinzo: Yes, but this is not the geoJSON but just a general model, isn't it? Perhaps you can just extend `convertToPoint` to take these as params and add them to the geoJSON. Did not really think thoroughly about this though.

native/package.json
```diff
@@ -51,6 +51,7 @@
     "@react-navigation/native": "^5.9.3",
     "@react-navigation/stack": "^5.14.3",
     "@sentry/react-native": "^2.3.0",
+    "@turf/distance": "^6.5.0",
```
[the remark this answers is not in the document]
f1sh1918: https://issues.tuerantuer.org/browse/IGAPP-777

its way more effort to refactor the getNearbyPlaces. So i created a separate ticket

native/src/components/MapPopup.tsx
```diff
@@ -20,13 +27,59 @@ const Popup = styled.View`
...
+`
+
+const MapPopup: React.FC<MapPopupProps> = ({
+  feature,
+  navigateTo,
+  cityCode,
+  language
+}: MapPopupProps): ReactElement => {
+  const { t } = useTranslation('pois')
+  return (
```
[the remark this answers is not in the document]
f1sh1918: yes

api-client/src/models/LocationModel.ts
```diff
@@ -46,6 +52,8 @@ class LocationModel {
     this._latitude = latitude
     this._longitude = longitude
     this._name = name
+    this._thumbnail = thumbnail
+    this._path = path
```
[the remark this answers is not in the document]
f1sh1918: not need when we pass it view convertToPoint

api-client/src/endpoints/createPOIsEndpoint.ts
f1sh1918: not need when we pass it view convertToPoint

native/src/components/MapView.tsx
f1sh1918: because it overrides the selectedFeature coordinates if you redirect from pageDetail. So everytime you enter mapView you follow automatically userLocation. This shouldn't be the case

native/src/components/MapView.tsx
f1sh1918: it only happens once at the moment so if there is no userLocation. There was no requirement that we should keep it up to date if the position changes, since we are not a navigation app ;)

native/src/components/MapView.tsx
f1sh1918: but we don't use mapbox but maplibre

native/src/routes/Pois.tsx
f1sh1918: we don't need the hook if we are on the details page (if path is available)

native/src/components/MapView.tsx
f1sh1918: done

native/src/components/MapView.tsx
klinzo: We still use OSM though. But yes, was just a question whether you know if we have to attribute those.

native/src/components/MapView.tsx
klinzo: I think it should not change `followUserLocation` to the opposite but instead always to true.

klinzo (approved):
Really nice work!

PR #431, opened 2021-09-14
IGAPP-766: Fix error when location disabled

Taggotty:

@f1sh1918 this is also a suggestion how to imporve the user location issue. Feel free to give feedback on that.


native/src/components/MapView.tsx
```diff
@@ -59,12 +58,10 @@ const MapView = ({
...
   cityCode,
-  setUserLocation,
-  userLocation
+  onRequestLocationPermission,
+  locationPermissionGranted
 }: MapViewPropsType): ReactElement => {
-  const [followUserLocation, setFollowUserLocation] = useState<boolean>(false)
-  const [locationPermissionGranted, setLocationPermissionGranted] = useState<boolean>(false)
-
+  const [followUserLocation, setFollowUserLocation] = useState<boolean>(locationPermissionGranted && !selectedFeature)
```
f1sh1918, line 64: shouldn't we go to the bounding box if nothing is selected and keep the value false in the beginning?

native/src/routes/Pois.tsx
```diff
@@ -76,9 +79,47 @@ const Pois = ({
...
+    const permissionGranted = locationPermission === RESULTS.GRANTED
+    if (permissionGranted) {
+      Geolocation.getCurrentPosition(
+        (position: GeolocationResponse) => {
+          setUserLocation([position.coords.longitude, position.coords.latitude])
+          setLocationPermissionGranted(true)
+        },
+        (_error: GeolocationError) => setLocationPermissionGranted(false),
+        {
+          enableHighAccuracy: true,
```
f1sh1918, line 98: Same values are used in Landing.tsx i would propose to move them to a shared constant file

native/src/routes/Pois.tsx
```diff
@@ -76,9 +79,47 @@ const Pois = ({
...
+          timeout: 50000,
+          maximumAge: 3600000
+        }
+      )
+    }
+  }, [])
+
+  useEffect(() => {
```
f1sh1918, line 106: can you add some comment to whats happening in this hook :)

f1sh1918 (approved):

native/src/components/MapView.tsx
Taggotty: hm. we did never discuss this, actually. I think it is more intuitive to center around current position if enabled. But I also see why you would display the bounding box instead. @f1sh1918 what is your opinion here?

f1sh1918: i think its fine to use location. to gain a good overview we should keep the zoom level same as it would use the bounding box

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+          setLocationState({
+            status: 'unavailable',
+            message: 'notAvailable'
+          })
+        }
+      }
+      const listener = SystemSetting.addLocationListener(onLocationChanged)
+      return () => {
+        listener.then(listener => listener && SystemSetting.removeListener(listener))
+      }
```
Taggotty, line 127: This updates the location state, whenever the user changes the location permission.

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+  status: 'unavailable'
+  message: 'noPermission' | 'notAvailable' | 'timeout' | 'loading'
+}
+
+type LocationStateType = SuccessfulLocationState | UnavailableLocationState
+
+export type LocationType = [number, number]
+
+const locationStateOnError = (error: GeolocationError): UnavailableLocationState => {
+  if (error.code === 1) {
```
Znippet, line 19: :art: only if possible: Do we have access to or can somhow use the associated constants instead of the numer, say PERMISSION_DENIED, POSITION_UNAVAILABLE, TIMEOUT?

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+  } else if (error.code === 2) {
+    return {
+      status: 'unavailable',
+      message: 'notAvailable'
+    }
+  } else {
+    return {
+      status: 'unavailable',
+      message: 'timeout'
+    }
```
Znippet, line 33: Is this the inteface? 
https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError/code

Another value then 3 is unexpcted. So perhaps better we use another if here (or switch case) with "3" (or TIMEOUT, see other comment). In the else we could put something like "status unknown" or event throw an error.

native/src/components/MapView.tsx
f1sh1918: ok i'm fine with that to only use bbox if we have no user location

native/src/hooks/useLocation.ts
Taggotty: okay. I found out they pass the error codes with the location error. Thanks for the hint.

native/src/hooks/useLocation.ts
Taggotty: Probably yes. But the docs you referenced is from web, therefore I'll use their interface

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+
+type SuccessfulLocationState = { status: 'ready' }
+type UnavailableLocationState = {
+  status: 'unavailable'
+  message: 'noPermission' | 'notAvailable' | 'timeout' | 'loading'
+}
+
+type LocationStateType = SuccessfulLocationState | UnavailableLocationState
+
+export type LocationType = [number, number]
```
f1sh1918, line 16: there is an interface you could use from geoJSON called Position
Optional to do ;)

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+    Geolocation.getCurrentPosition(
+      (position: GeolocationResponse) => {
+        setLocation([position.coords.longitude, position.coords.latitude])
+        setLocationState({ status: 'ready' })
+      },
+      (error: GeolocationError) => {
+        setLocationState(locationStateOnError(error))
+      },
+      {
+        enableHighAccuracy: true,
```
f1sh1918, line 60: its also used in Landing.tsx
I would recommend to move the settings to a shared constant file

f1sh1918:
I tested on android (native) and ios and it works except of one thing.
When i grant gps permission or i have already granted gps permission it does set the position to the userLocation
For example:
1. Go to POI Overview click on the gps FAB and grant the permission the position does not follow my location
2. When i go back to location overview and again to POIS the initial position is the bounding box

native/src/components/MapView.tsx
f1sh1918: here is the problem that its only set once and will not get true by changing ```locationPermissionGranted``` prop

native/src/components/MapView.tsx
```diff
@@ -82,38 +78,23 @@ const MapView = ({
...
-    checkLocationPermission().then(onLocationPermissionRequest)
-  }, [onLocationPermissionRequest])
-
-  const requestPermission = useCallback(() => {
-    requestLocationPermission().then(onLocationPermissionRequest)
-    locationPermissionGranted && setFollowUserLocation(!followUserLocation)
-  }, [followUserLocation, locationPermissionGranted, onLocationPermissionRequest])
+  const onRequestLocation = useCallback(() => {
+    onRequestLocationPermission()
+    setFollowUserLocation(locationPermissionGranted)
```
klinzo, line 83: Does this actually work? The prop `locationPermissionGranted` is not updated immediately, is it?

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+      if ((await requestLocationPermission()) === RESULTS.GRANTED) {
+        determineLocation()
+      } else {
+        setLocationState({
+          message: 'noPermission',
+          status: 'unavailable'
+        })
+      }
+    }
```
klinzo, line 108:
```suggestion
      if (locationPermissionStatus === RESULTS.GRANTED || (await requestLocationPermission()) === RESULTS.GRANTED) {
      determineLocation()
    } else {
      setLocationState({
        message: 'noPermission',
        status: 'unavailable'
      })
      if (locationPermissionStatus === RESULTS.BLOCKED) {
        await openSettings()
      }
    }
```

native/src/hooks/useLocation.ts
klinzo: Perhaps you could add a comment :)

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+      }
+      const listener = SystemSetting.addLocationListener(onLocationChanged)
+      return () => {
+        listener.then(listener => listener && SystemSetting.removeListener(listener))
+      }
+    }
+  })
+
+  const isReadyOrLoading =
+    locationState.status === 'ready' || (locationState.status === 'unavailable' && locationState.message === 'loading')
```
klinzo, line 132:
```suggestion
  const isReadyOrLoading =
    locationState.status === 'ready' || locationState.message === 'loading'
```

native/src/routes/Landing.tsx
```diff
@@ -24,161 +22,34 @@ const Wrapper = styled(View)`
...
-          cities={cities}
-          t={t}
-          location={location}
-          retryDetermineLocation={retryDetermineLocation}
-          navigateToDashboard={this.navigateToDashboard}
-        />
-      </Wrapper>
-    )
-  }
+const Landing = ({ cities, language, theme, navigateToDashboard, clearResourcesAndCache }: PropsType): ReactElement => {
```
klinzo, line 30: Very clean, nice!

native/src/routes/LandingContainer.tsx
```diff
@@ -68,7 +67,7 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
   }
 }
 
-const ThemedTranslatedLanding = withTranslation('landing')(withTheme<LandingPropsType>(Landing))
```
klinzo, line 71: You could move the theming inside `Landing` as well, right?

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
+import Geolocation, { GeolocationError, GeolocationResponse } from '@react-native-community/geolocation'
```
klinzo, line 1: I would maybe rename this hook to `useUserLocation` as location is used in many ways.

native/src/routes/Pois.tsx
```diff
@@ -77,20 +82,20 @@ const Pois = ({
   const { t } = useTranslation('pois')
   const theme = useTheme()
   const [selectedFeature, setSelectedFeature] = useState<Feature<Point> | null>(null)
-  const [userLocation, setUserLocation] = useState<number[] | undefined>(undefined)
   const [featureLocations, setFeatureLocations] = useState<Feature<Point>[]>(prepareFeatureLocations(pois))
+  const { location, requestAndDetermineLocation } = useLocation(path === null)
```
klinzo, line 86: Why do we only want to use a settings listener if the path is null?

klinzo (changes requested):

native/src/components/CitySelector.tsx
```diff
@@ -140,14 +140,16 @@ class CitySelector extends React.PureComponent<PropsType> {
         <CityGroupContainer>
           <CityGroup theme={theme}>{t('nearbyPlaces')}</CityGroup>
           <NearbyMessageContainer>
-            <NearbyMessage theme={theme}>{location ? t(location.message) : ''}</NearbyMessage>
+            <NearbyMessage theme={theme}>
+              {locationState.status !== 'ready' ? t(locationState.message) : ''}
+            </NearbyMessage>
             <RetryButtonContainer>
-              {retryDetermineLocation && (
+              {locationState.status === 'unavailable' && locationState.message === 'loading' && (
```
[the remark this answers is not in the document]
Taggotty: otherwise the user should have denied the permission for a location or the location service is disabled. But we could probably show a button if there is a timeout.

native/src/routes/Pois.tsx
Taggotty: if path is null we are in the pois detail view, where we definitely don't want to use the listener. But according to our designs we might want to show the distance there.

native/src/components/CitySelector.tsx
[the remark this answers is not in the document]
klinzo: I think showing a button always makes sense. If the permission is denied it should open the app settings to change that, if the location service is disabled maybe the location service settings.

native/src/hooks/useLocation.ts
```diff
@@ -0,0 +1,141 @@
...
+import { useCallback, useEffect, useState } from 'react'
+import { openSettings, RESULTS } from 'react-native-permissions'
+import SystemSetting from 'react-native-system-setting'
+
+import { checkLocationPermission, requestLocationPermission } from '../utils/LocationPermissionManager'
+
+type SuccessfulLocationState = { status: 'ready' }
```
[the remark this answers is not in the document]
Taggotty: this made problem with react useState and useCallback as if the location is not available we need to access the location state, but adding it to the dependencies is not possible due to typescript complaining a successfull location state does not have a message. And when only adding location the component will rerender the state and callback every time the location updates.

native/src/hooks/useLocation.ts
Taggotty: hmm. but what if the user changes his permission state while using the app? Then we'd assume he has still granted permission here 🤔

native/src/components/CitySelector.tsx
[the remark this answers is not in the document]
Taggotty: You are right. I checked how it was before and it should be the other way around. Nice catch 👍🏼

native/src/hooks/useLocation.ts
Taggotty: I prefer being more specific here, as the geojson type only defines the location as [number, number]

native/src/components/MapView.tsx
Taggotty: updated, as onRequestLocationPermission actually returns a promise

native/src/routes/Pois.tsx
Taggotty: is remove already.

Taggotty: is remove already.

native/src/hooks/useLocation.ts
klinzo: Isn't this an issue with your code as well?

native/src/hooks/useLocation.ts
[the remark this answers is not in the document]
klinzo: I don't really get what the problem is but I guess you tried and did not work so fine for me to keep it as it is :smile:

Taggotty:
> I tested on android (native) and ios and it works except of one thing.
> When i grant gps permission or i have already granted gps permission it does set the position to the userLocation
> For example:
> 
> 1. Go to POI Overview click on the gps FAB and grant the permission the position does not follow my location
> 2. When i go back to location overview and again to POIS the initial position is the bounding box.

I think I fixed it.

native/src/hooks/useLocation.ts
Taggotty: This part was just copy/pasted from Landing.tsx. What issue are you addressing?

klinzo: Hmm I thought you replied to this saying there is some issue with my code. but then nvm...

klinzo:
Is this issue already done with this PR: https://issues.tuerantuer.org/browse/IGAPP-737

Taggotty:
> Is this issue already done with this PR: https://issues.tuerantuer.org/browse/IGAPP-737

Yes, I think so as it is almost impossible to separate those two issues.

klinzo (approved):
Works perfectly! :tada: One minor thing: If clicking on the location button, the zoom level is not changed. I think we did that some time ago, but not sure?

I think once this is merged we can create a first test build for the other teams :muscle:

Taggotty:
> One minor thing: If clicking on the location button, the zoom level is not changed. I think we did that some time ago, but not sure?

Yes, I removed this one PR earlier, as it would disable  `followUserLocation` internally, which felt a little bit strange. But I think this is also acceptable.

PR #433, opened 2021-09-20
IGAPP-769: Stick to android browser 1.3.0

f1sh1918:


native/.idea/runConfigurations/Run_release_on_Android__integreat_.xml
```diff
@@ -1,12 +0,0 @@
-<component name="ProjectRunConfigurationManager">
```
Taggotty, line 1: Why did you remove this? Please readd.

native/android/build.gradle
```diff
@@ -2,10 +2,10 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "29.0.3"
+        buildToolsVersion = "31.0.3"
         minSdkVersion = 21
-        compileSdkVersion = 29
-        targetSdkVersion = 29
```
Taggotty, line 8: I'am not sure if we want to increase the target sdk version here. Did you find out why uses androidx.browser:browser:1.4.0-alpha01 instead of 1.3.0? Maybe setting it to use 1.3.0 will also fix the build.

Taggotty: sth like https://stackoverflow.com/a/69249047

Taggotty:
I think it would also be cool if you could do the browserlist update in this PR -> https://github.com/browserslist/browserslist#browsers-data-updating (I think you have to run the command from the project root)

native/android/build.gradle
Taggotty: I finally found the culprit [react-native-inapp-browser](https://github.com/proyecto26/react-native-inappbrowser) fetches the newest version of androidx.browser:browser. To fix this you can add `androidXBrowser = "1.3.0"` in the android/build.gradle

native/.idea/runConfigurations/Run_release_on_Android__integreat_.xml
f1sh1918: sry it was by accident ;)

Taggotty (approved):

native/android/app/build.gradle
```diff
@@ -79,12 +79,12 @@ import com.android.build.OutputFile
 
 project.ext.react = [
         extraPackagerArgs: (System.getenv('EXTRA_PACKAGER_ARGS') ?: "").split("\\s+"),
-        enableHermes: true,  // clean and rebuild if changing
-        entryFile: "index.ts",
+        enableHermes     : true,  // clean and rebuild if changing
```
maxammann, line 82: All formatting changes should be reverted in these files.

When we apply the diffs for react native upgrades it is beneficial if they only differ in important parts.

maxammann (approved):
Look good apart from the formatting

Taggotty (approved):

PR #437, opened 2021-09-22
IGAPP-747: Add navigation deep linking to details page

f1sh1918:

Just create the functionality with a basic button on detail page since we should create a separate task for the detail page to adjust it to the design proposal.

Tested on IOS and Android


Znippet (approved):

native/src/components/PageDetail.tsx
```diff
@@ -33,13 +39,25 @@ type PropsType = {
 
 const PageDetail: React.FC<PropsType> = (props: PropsType): ReactElement => {
   const { identifier, information, theme, language, onLinkClick, linkLabel } = props
+  const { t } = useTranslation('pois')
+
+  const navigationUrl = Platform.select({
+    ios: `maps:0,0?q=${information}`,
+    android: `geo:0,0?q=${information}`
+  })
```
klinzo, line 47: TBH I don't think this should be part of a generic PageDetail component but instead be moved to a separate component. Then you can also omit the wrapping in a fragment and checking for `linkLabel` and `onLinkClick` props below as those can be required in a separate component.

native/src/components/PageDetail.tsx
```diff
@@ -33,13 +39,25 @@ type PropsType = {
...
     <>
       <DetailContainer theme={theme} language={language}>
         <Identifier theme={theme}>{identifier}: </Identifier>
         {information}
       </DetailContainer>
       {linkLabel && onLinkClick && <Button title={linkLabel} onPress={() => onLinkClick()} />}
+      {linkLabel && (
+        <>
+          <Spacer />
+          <Button title={t('navigation')} onPress={() => navigationUrl && Linking.openURL(navigationUrl)} />
```
klinzo, line 58: Do we need to check for `navigationUrl` here?

klinzo (changes requested):

native/src/components/PageDetail.tsx
f1sh1918: hm so what do you prefer? passing the whole component? or sth like header and body part inside of the generic component.

native/src/components/PageDetail.tsx
f1sh1918: yes since the platform.select could return string | undefined

native/src/components/PageDetail.tsx
klinzo: Just split it in two components (one PageDetail as it was before and a separate one for the map deep linking and then just render both separately.

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,11 @@
+import { Platform } from 'react-native'
+
+/**
+ * Depending on the platform this function returns the correct deeplink for navigation app
+ * @param address: location address
+ */
+export const getNavigationDeepLinks = (address: string): string | undefined =>
+  Platform.select({
+    ios: `maps:0,0?q=${address}`,
+    android: `geo:0,0?q=${address}`
```
klinzo, line 10: Maybe we should specify long/lat as well: https://developers.google.com/maps/documentation/urls/android-intents#categorical-search

klinzo (approved):

klinzo:
Tested on android

SteffiStuffel (approved):
Tested on Android emulator.

native/src/utils/getNavigationDeepLinks.ts
f1sh1918: added coordinates --> worked on Maps.me

PR #438, opened 2021-09-22
IGAPP-583: Update translations and remove old and unused translations

f1sh1918:


klinzo (commented):
Not sure if we should merge this already. Things to consider:
* We did not really decide on what text to use here or gave it any deeper thoughts
* The new message currently does not really make sense for users as the location feature is still in development

Taggotty:
> Not sure if we should merge this already. Things to consider:
> 
> * We did not really decide on what text to use here or gave it any deeper thoughts
> * The new message currently does not really make sense for users as the location feature is still in development

So you'd postpone this issue until after we release POIs?

klinzo:
> So you'd postpone this issue until after we release POIs?

Not until after but until we release them. So probably together with enabling the feature flag.

maxammann (approved):

klinzo (changes requested):
See above, it does not make sense to merge this right now as long as we did not decide on the right texts/translations.

klinzo (approved):

PR #440, opened 2021-09-28
IGAPP-778: Add arrow-body-style rule

klinzo:

### DIFF
--- a/web/src/components/Selector.tsx
+++ b/web/src/components/Selector.tsx
@@ -84,40 +84,32 @@ const Selector = ({
   verticalLayout,
   closeDropDown,
   disabledItemTooltip
-}: PropsType): ReactElement => {
-  return (
-    <Wrapper vertical={verticalLayout}>
-      {items.map(item => {
-        if (item.href) {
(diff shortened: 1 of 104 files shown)


web/src/components/Selector.tsx
```diff
@@ -84,40 +84,32 @@ const Selector = ({
...
-            </Element>
-          )
-        }
-      })}
-    </Wrapper>
-  )
-}
+}: PropsType): ReactElement => (
+  <Wrapper vertical={verticalLayout}>
+    {items.map(item => {
```
f1sh1918, line 89: Well not your business but i would prefer to call a function in the template and not use all this logic within the template :/

f1sh1918 (approved):

web/src/components/Selector.tsx
klinzo: Agreed, but not in the scope of this PR

f1sh1918 (approved):

SteffiStuffel (approved):

PR #442, opened 2021-09-28
IGAPP-778: Add no-else-return rule

klinzo:

### DIFF
--- a/native/src/hocs/withPayloadProvider.tsx
+++ b/native/src/hocs/withPayloadProvider.tsx
@@ -146,31 +146,29 @@ const withPayloadProvider = <
             <Component {...componentProps} />
           </LayoutedScrollView>
         )
-      } else {
-        // Full screen loading spinner
-        return (
-          <LayoutedScrollView refreshControl={<RefreshControl refreshing={false} />}>
-            <ProgressContainer progress={props.progress} />
-          </LayoutedScrollView>
-        )
       }
-    } else {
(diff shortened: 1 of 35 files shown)


klinzo:
I split this in two commits:
* General no-else-return rule (0a0d391)
* no-else-return for else if (07c3c18)
If you don't like both I can revert it.

f1sh1918 (approved):

native/src/hocs/withPayloadProvider.tsx
```diff
@@ -119,58 +119,60 @@ const withPayloadProvider = <
...
       return (
-        <LayoutedScrollView refreshControl={<RefreshControl onRefresh={refreshIfPossible} refreshing={false} />}>
-          <Component {...componentProps} />
+        <LayoutedScrollView refreshControl={<RefreshControl refreshing={false} />}>
+          <ProgressContainer progress={props.progress} />
         </LayoutedScrollView>
       )
     }
+    const componentProps = { ...props.innerProps, dispatch: props.dispatch } as S
+    // props.status === 'success'
```
f1sh1918, line 162: rm comment if not needed

f1sh1918 (approved):

Taggotty (approved):

native/src/hocs/withPayloadProvider.tsx
klinzo: Out of scope and makes sense imo to keep.

PR #443, opened 2021-09-28
IGAPP-781: Throw FetchError if response.json() fails

klinzo:


klinzo:
This does not fix the issue (which is not fixable) but instead just throw the right error and therefore shows the right message to the user and prevents it from showing up in sentry. Seems like it is possible to let the response.json() call fail because of a missing internet connection.

Taggotty (commented):
Could you add tests for Endpoint request?

klinzo:
> Could you add tests for Endpoint request?

Already on it!

Taggotty (approved):

SteffiStuffel (approved):

PR #444, opened 2021-09-29
IGAPP-771: Geo json properties custom type

f1sh1918:


api-client/src/maps/index.ts
```diff
@@ -11,6 +11,18 @@ export type MapViewViewport = {
   zoom: number
 }
 
+/**
+ * Override existing GeoJsonProperties from types/geojson to be more precise
+ */
+export type GeoJsonPoiProperties = {
```
klinzo, line 17: Maybe we could create aliases for `Feature<Point, GeoJsonPoiProperties>` and `FeatureCollection<Point, GeoJsonPoiProperties` as well? We are using them really often and I think simpler types would be nice

klinzo (approved):

api-client/src/endpoints/createPOIsEndpoint.ts
```diff
@@ -20,29 +20,32 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<PoiModel>> =>
...
-              country: poi.location.country,
-              latitude: poi.location.latitude,
-              longitude: poi.location.longitude
-            }),
-            lastUpdate: moment.tz(poi.modified_gmt, 'GMT'),
-            hash: poi.hash
-          })
-        })
+        json
+          .filter(poi => poi.location.id && poi.location.name && poi.location.latitude && poi.location.longitude)
```
Taggotty, line 24: I think we still want to show locations without longitude and latitude in the Pois list

Taggotty (commented):
In POIListItem you'll need to change `poi.location` to `poi.location.location`
And `feature.properties` doesn't seem to be optional anymore. Therefore, replace all occurences of `feature.properties?` with `feature.properties`

api-client/src/endpoints/createPOIsEndpoint.ts
klinzo: Hmm not sure about that. I think every POI that is relevant should have a longitude/latitude.

f1sh1918: i'm with @klinzo the poi feature does not make any sense for me without displaying them on the map and have an address. i think with the cms plugin a poi that has an address also has long/lat

api-client/src/maps/index.ts
f1sh1918: is that what you have expected?

```js
// aliases for Features and FeatureCollections using custom GeoJsonProperties and Point
export type PoiFeature = Feature<Point, GeoJsonPoiProperties>
export type PoiFeatureCollection = FeatureCollection<Point, GeoJsonPoiProperties>
```

klinzo: Yep, would be nice!

f1sh1918:
> In POIListItem you'll need to change `poi.location` to `poi.location.location` And `feature.properties` doesn't seem to be optional anymore. Therefore, replace all occurences of `feature.properties?` with `feature.properties`

Updated pr so pls recheck

web/src/components/MapView.tsx
```diff
@@ -38,19 +37,19 @@ const geolocateControlStyle: React.CSSProperties = {
...
-  featureCollection: FeatureCollection<Point>
+  featureCollection: PoiFeatureCollection
 }
 
 const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): ReactElement => {
   const { featureCollection, bboxViewport } = props
   const [viewport, setViewport] = useState<MapViewViewport>(bboxViewport)
   const [showPopup, togglePopup] = React.useState<boolean>(false)
-  const [currentFeature, setCurrentFeature] = React.useState<Feature<Point> | null>(null)
+  const [currentFeature, setCurrentFeature] = React.useState<PoiFeature | null>(null)
```
Taggotty, line 47: Try staying consistent with `React.useState` and `useState`

Taggotty (approved):

web/src/components/MapPopup.tsx
```diff
@@ -1,15 +1,17 @@
-import { GeoJsonProperties, Position } from 'geojson'
+import { Position } from 'geojson'
 import React, { ReactElement } from 'react'
 import { Popup } from 'react-map-gl'
 import styled from 'styled-components'
 
+import { GeoJsonPoiProperties } from 'api-client/src'
```
klinzo, line 6:
```suggestion
import { GeoJsonPoiProperties } from 'api-client'
```

PR #446, opened 2021-09-29
IGAPP-778: Use airbnb config

klinzo:

This PR removes rules from our .eslintrc that are already part of the airbnb/standard config. Some rules appeared too strict for our project (for now at least). I will create one more PR with the two react rules set to off.

If you'd like any other rule that is currently disabled, let me know.


native/src/routes/Pois.tsx
```diff
@@ -96,7 +96,9 @@ const Pois = ({
         )
         setSelectedFeature(currentFeature ?? null)
       }
-      location && setFeatureLocations(featureLocations)
+      if (location) {
+        setFeatureLocations(featureLocations)
+      }
```
klinzo, line 101: @f1sh1918 @Taggotty is this if correct/necessary?

.eslintrc.js
```diff
@@ -21,33 +23,52 @@ module.exports = {
...
+    'class-methods-use-this': 'off',
+    'default-case': 'off',
+    'global-require': 'off',
+    'import/extensions': 'off',
+    'import/no-extraneous-dependencies': 'off',
+    'import/no-unresolved': 'off',
+    'import/prefer-default-export': 'off',
+    'lines-between-class-members': 'off',
+    'no-shadow': 'off',
+    'no-nested-ternary': 'off',
```
SteffiStuffel, line 40: Noooooo :sob: 

Seriously, I would think this could really make sense to have as a rule.

.eslintrc.js
```diff
@@ -21,33 +23,52 @@ module.exports = {
...
-    '**/lib-dist/'
+    '**/lib-dist/',
+    '.eslintrc.js'
   ],
   rules: {
-    'arrow-body-style': 'error',
-    'arrow-parens': ['error', 'as-needed'],
+    // Overly strict rules (for now)
+    'class-methods-use-this': 'off',
+    'default-case': 'off',
```
SteffiStuffel, line 32: Maybe this is also really helpful, as having no default can lead to unexpected results, when not thinking about the default case. Having this rule enforces to think about it.

.eslintrc.js
```diff
@@ -21,33 +23,52 @@ module.exports = {
...
+    'global-require': 'off',
+    'import/extensions': 'off',
+    'import/no-extraneous-dependencies': 'off',
+    'import/no-unresolved': 'off',
+    'import/prefer-default-export': 'off',
+    'lines-between-class-members': 'off',
+    'no-shadow': 'off',
+    'no-nested-ternary': 'off',
+    'no-underscore-dangle': 'off',
+    'no-use-before-define': 'off',
```
SteffiStuffel, line 42: Why not?

native/.idea/runConfigurations.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SteffiStuffel, line 1: Why these changes?

native/jest.setup.ts
```diff
@@ -25,7 +25,11 @@ const walkDir = (dir: string, callback: (filePath: string) => void): void => {
   fs.readdirSync(dir).forEach(f => {
     const filePath = path.join(dir, f)
     const isDirectory = fs.statSync(filePath).isDirectory()
-    isDirectory ? walkDir(filePath, callback) : callback(filePath)
+    if (isDirectory) {
```
SteffiStuffel, line 28: Why not ternery?

native/src/components/NavigationTiles.tsx
```diff
@@ -62,7 +62,7 @@ class NavigationTiles extends React.PureComponent<PropsType, StateType> {
     const contentSizeDiff = nativeEvent.contentSize.width - nativeEvent.layoutMeasurement.width
     this.setState({
       xPosition: nativeEvent.contentOffset.x,
-      contentSizeDiff: parseInt(contentSizeDiff.toFixed(0))
+      contentSizeDiff: parseInt(contentSizeDiff.toFixed(0), 10)
```
SteffiStuffel, line 65: Why is radix needed? It is optional.

SteffiStuffel (approved):

native/src/components/NavigationTiles.tsx
klinzo: Yes it is, but from the docs:
> An integer between 2 and 36 that represents the radix (the base in mathematical numeral systems) of the string. Be careful—this does not default to 10! If the radix value is not of the Number type it will be coerced to a Number.

We could also disable the rule oc but I think it makes sense (or does not hurt at least). The corresponding rule is https://eslint.org/docs/rules/radix
> By default, parseInt() will autodetect decimal and hexadecimal (via 0x prefix). Prior to ECMAScript 5, parseInt() also autodetected octal literals, which caused problems because many developers assumed a leading 0 would be ignored.

This confusion led to the suggestion that you always use the radix parameter to parseInt() to eliminate unintended consequences. So instead of doing this:

native/jest.setup.ts
klinzo: https://eslint.org/docs/rules/no-unused-expressions
> An unused expression which has no effect on the state of the program indicates a logic error.

> For example, n + 1; is not a syntax error, but it might be a typing mistake where a programmer meant an assignment statement n += 1; instead. Sometimes, such unused expressions may be eliminated by some build tools in production environment, which possibly breaks application logic.

It is possible to disable this for ternaries though.

native/.idea/runConfigurations.xml
klinzo: No idea, IntelliJ was quite persistent in re adding those even though I deleted them quite a few times :D

.eslintrc.js
klinzo: We already have `'@typescript-eslint/no-use-before-define': 'error'` which does the same with typescript support. This rule also throws the error all the time:
> 'React' was used before it was defined  no-use-before-define

.eslintrc.js
klinzo: We already have `'@typescript-eslint/no-use-before-define': 'error'` which does the same with typescript support. This rule also throws the error all the time:
> 'React' was used before it was defined  no-use-before-define

.eslintrc.js
klinzo: Will add https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.md which only enforces necessary switch statements. I don't think it makes sense being forced to add unnecessary defaults that are not possible to be executed.

.eslintrc.js
klinzo: Added to TODOs, will do in a separate PR. If you want you can do it as well :)

native/src/components/NavigationTiles.tsx
SteffiStuffel: yeah, oke, makes sense.

.eslintrc.js
klinzo: Done in #447

native/src/routes/Pois.tsx
Taggotty: Not sure. @f1sh1918 ?

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -89,8 +89,8 @@ const HeaderActionItemDropDown = (props: PropsType): ReactElement => {
 
   return (
     <Container ref={wrapperRef}>
-      <Tooltip text={text} flow={'down'} mediumViewportFlow={'left'}>
-        <button aria-label={text} onClick={toggleDropDown}>
+      <Tooltip text={text} flow='down' mediumViewportFlow='left'>
+        <button type='button' aria-label={text} onClick={toggleDropDown}>
```
Taggotty, line 93: why did you add `type='button'`?

Taggotty (approved):

PR #448, opened 2021-09-30
IGAPP-778: Enable noUncheckedIndexedAccess flag and add typescript parser rules

klinzo:

This is the last linting PR, promised! Or at least, the last linting PR for integreat-app :stuck_out_tongue: 

I was not sure about enabling the typescript flag [noUncheckedAccessIndex](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) since there were quite some false positive errors, but I think it really reduces possible error sources since one is forced to always check array elements or map values if they really exist. Let me know what you think!


build-configs/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../tsconfig.base.json",
+  "extends": "../tsconfig.json",
```
maxammann, line 2: Why this change?

web/src/components/Tooltip.tsx
```diff
@@ -208,18 +208,11 @@ const fixFlow = (element: Element | null, preferredFlow: FlowType, dimensions: V
...
-  }
-
   if (checker.check(element.getBoundingClientRect(), dimensions)) {
     return preferredFlow
   }
   const fallback = checker.fallbacks.find((fallbackFlow: FlowType) => {
     const fallbackChecker = spaceCheckers[fallbackFlow]
-    if (!fallbackChecker) {
-      throw new Error('Fallback not found')
-    }
```
maxammann, line 222: These checks should stay.

web/tools/webpack.config.ts
```diff
@@ -285,7 +285,7 @@ const createConfig = (
   }
 
   // Optimize the bundle in production mode
-  if (!devServer && optimize) {
```
maxammann, line 288: This changes the meaning of the code I suppose.

maxammann (approved):

web/src/components/Tooltip.tsx
klinzo: It is guaranteed by typescript that there are only valid keys used, so they should not be necessary.

web/tools/webpack.config.ts
klinzo: optimize is just a namespace/import from webpack, I don't really see a reason why we need to check for it.

build-configs/tsconfig.json
klinzo: To enable these rules with typescript support I had to set the parserOptions in .eslintrc.js to a ts config. To allow linting from the root dir as well as from the different workspaces the same name is needed. At least I don't know how to allow that otherwise.

SteffiStuffel (approved):

klinzo:
@maxammann could you have a look again please on my comments?

web/src/components/Tooltip.tsx
maxammann: spaceCheckers doesnt need to have every key present. The if check can be true if you change the config above.

PR #453, opened 2021-10-06
IGAPP-794: Remove weird background of news

klinzo:

### DIFF
--- a/release-notes/unreleased/IGAPP-794-remove-gradient-local-news.yml
+++ b/release-notes/unreleased/IGAPP-794-remove-gradient-local-news.yml
@@ -1,0 +1,5 @@
+issue_key: IGAPP-794
+show_in_stores: true
+platforms:
+  - web
+en: Removed weird color gradient from news.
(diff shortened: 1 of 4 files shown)


release-notes/unreleased/IGAPP-794-remove-gradient-local-news.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: IGAPP-794
+show_in_stores: true
+platforms:
+  - web
+en: Removed weird color gradient from news.
```
Taggotty, line 5: Why no German release note?

Taggotty (approved):
Tested in chrome. 
The TüNews list items don't have a separation line in between. Is that intentionally?

release-notes/unreleased/IGAPP-794-remove-gradient-local-news.yml
klinzo: There is no store for web so they are only ever posted/used in english.

PR #455, opened 2021-10-07
IGAPP-521: Show Snackbar if Opening Bad Link

sarahsporck:

As overview what happened: 
- Removed navigateToDeepLink from top level components, as it was passed down the whole tree  to Page and NativeHtml only. 
- Introduced a hook useNavigateToDeepLink, which does exactly what it sounds like
- Introduce hook useSnackbar and a 'SnackbarError' If you have better ideas on how to make `openExternalUrl`prettier, feel free to leave a note
- Refactored Categories to a functional component
- Extracted a single pois page into a separate component. (Was not really part of the task, but as I found out that opening the navigation url needed some fixup, i just split it
- and fixed the typing of some promise returning functions


SteffiStuffel:
I tested it, but can't get it working.

For a link without a valid city I get:
Bildschirmfoto zu 2021-10-12 14-48-45

For valid city but missing language I get:
Bildschirmfoto zu 2021-10-12 14-49-54

For valid city and valid language, I get:
Bildschirmfoto zu 2021-10-12 14-49-36

Not sure, if I tested what the ticket is about?

PR #457, opened 2021-10-08
IGAPP-770: Move clearRouteOnClose functionality into Hook

sarahsporck:


Taggotty:
> I think this breaks the app, routes are closed immediately after opening.

Adding dependencies to `useEffect` fixed this issue. This time I also tested with flipper Redux Debugger :)

PR #462, opened 2021-10-18
IGAPP-692: Automated store translations

steffenkleinle:


SteffiStuffel:
Maybe you can put the translations to google/drive -> H Templates and Translations

klinzo:
> Maybe you can put the translations to google/drive -> H Templates and Translations

Why? We never do that, everything is in the repo :see_no_evil:

PR #463, opened 2021-10-19
IGAPP-741: Short description for poi

ztefanie:

### DIFF
--- a/native/src/components/Page.tsx
+++ b/native/src/components/Page.tsx
@@ -6,6 +6,7 @@ import styled from 'styled-components/native'
 
 import { ThemeType } from 'build-configs'
 
+import dimensions from '../constants/dimensions'
 import DateFormatterContext from '../contexts/DateFormatterContext'
 import { PageResourceCacheEntryStateType, PageResourceCacheStateType } from '../redux/StateType'
 import { RESOURCE_CACHE_DIR_PATH } from '../utils/DatabaseConnector'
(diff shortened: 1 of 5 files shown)


SteffiStuffel:
> Hm i'm wondering how you get the distance on the screen. I couldn't find any method that calculates the distance Maybe i'm just blind ;) Here is the standard lib for calc distance if its needed http://turfjs.org/docs/#distance

There is a property `distance` existing on PoiFeature that i use, I assumed this will be correctly set in other tickets, as it does not make sense to calculate distance everytime, when it just can be passed.

f1sh1918:
> > Hm i'm wondering how you get the distance on the screen. I couldn't find any method that calculates the distance Maybe i'm just blind ;) Here is the standard lib for calc distance if its needed http://turfjs.org/docs/#distance
> 
> There is a property `distance` existing on PoiFeature that i use, I assumed this will be correctly set in other tickets, as it does not make sense to calculate distance everytime, when it just can be passed.

Yes the interface includes `distance` but it has to be calculated once you get the location of the user and shouldn't be updated. So it has to be added in the frontend. There was definitely not enough info in the ticket. We can create a separate ticket or you can add the functionality if you like

f1sh1918:
One small idea according to the thumbnail. For native we use a placeholder if there is no image, dunno if we want to do the same for web
`import EventPlaceholder1 from '../assets/EventPlaceholder1.jpg'`

PR #464, opened 2021-10-19
IGAPP-772: Poi details view

f1sh1918:
Tested on Android & Ios Native


native/src/components/MapView.tsx
```diff
@@ -111,30 +92,27 @@ const MapView = ({
...
+  const onPress = async (pressedLocation: Feature) => {
+    if (!mapRef.current || !cameraRef.current || !pressedLocation.properties) {
+      return
+    }
+    const featureCollection = await mapRef.current.queryRenderedFeaturesAtPoint(
+      [pressedLocation.properties.screenPointX, pressedLocation.properties.screenPointY],
+      undefined,
+      [featureLayerId]
+    )
+    const feature = featureCollection?.features.find((it): it is PoiFeature => it.geometry.type === 'Point')
```
maxammann, line 104: In the Ehrenamtskarte App we select the feature which is closest to the location of the tap. Also: The first icon in the featureCollection is the one with the highest z index.

Maybe this could help if features are close together.

maxammann (approved):
not yet tested

maxammann:
Works, but when I change the language while the map is visisble then I only see a white page.

native/src/components/MapView.tsx
f1sh1918: i only found in type declaration for the symbolLayer
` symbolZOrder?: 'auto' | 'viewport-y' | 'source' | Expression;`
added ` symbolZOrder: 'source',`

f1sh1918: the function retrieving features
`mapRef.current.queryRenderedFeaturesAtPoint` has some kind of tolerance but when you click too far away the it results in an empty response
We can discuss if we really want to have this, because if we can not deselect any feature we never can show the POI List, so i don't think this is a good idea

f1sh1918:
> Works, but when I change the language while the map is visisble then I only see a white page.

ok i'm gonna investigate in this..strange

native/src/components/PoiDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+  // TODO this has to be removed when we get proper images from CMS IGAPP-805
+  const thumbnail = feature.properties.thumbnail?.replace('-150x150', '') ?? EventPlaceholder1
+  const { location, address, postcode, town } = poi.location
+  const { distance, title } = feature.properties
+
+  // TODO refactor to use a get coordinates method IGAPP-806
+  const onNavigate = () => {
+    if (location && poi.featureLocation?.geometry.coordinates) {
+      const navigationUrl = getNavigationDeepLinks(location, poi.featureLocation.geometry.coordinates)
+      Linking.openURL(navigationUrl)
```
Taggotty, line 82: I think here you should catch the error and show a fancy snackbar, if the Link does not exist.
```suggestion
            Linking.openURL(navigationUrl).catch(() => showSnackbar(dispatch, 'No navigation application found.'))
```

Taggotty (changes requested):

Taggotty (approved):

PR #473, opened 2021-10-26
IGAPP-809: Languages json not existing

klinzo:

~~This PR probably does not fix anything and is rather meant to gather additional information. This bug should not be critical as I tried and afaics all errors logged are caught and handled correctly.~~

~~The error that the `languages.json does not exist` always occurs after another error during fetching, most of the time after the following error: `Failed to load the request for the languages endpoint. Body not allowed for GET or HEAD requests` (I also found [one occurrence](https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/264/events/25db970a57814cc89d608f53896a1686/?project=2) of the same error for the cities endpoint, so it should not be the fault of the language endpoint). Nearly always it also seems like `loadCityContent` (or `loadCities` for the one occurrence) are called and run twice.~~

~~The whole thing is really weird, first because of the iniital GET/Body request thingy, second because the following `languages.json does not exist` error only occurs after this error (and not e.g. after regular fetch errors) and third because sagas are run twice even though we are using `takeLatest` which should cancel any previous sagas. I suspect that the issue is somewhere in executing `loadLanguages` twice though.~~

See below :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -522,6 +522,9 @@ jobs:
                 command: yarn lint
                 name: Lint
             - run:
+                command: yarn workspace web stylelint
+                name: Stylelint
+            - run:
                 command: yarn workspaces run ts:check
                 name: TS check
             - unit_test:
--- a/api-client/src/Endpoint.ts
+++ b/api-client/src/Endpoint.ts
@@ -3,7 +3,7 @@ import { MapParamsToUrlType } from './MapParamsToUrlType'
 import { MapResponseType } from './MapResponseType'
 import Payload from './Payload'
 import FetchError from './errors/FetchError'
-import ResponseError, { RequestOptionsType } from './errors/ResponseError'
+import ResponseError from './errors/ResponseError'
 import { request as fetch } from './request'
 
 /**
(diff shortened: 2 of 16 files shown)


klinzo:
Finally found the issue :hear_no_evil: `merge´ also alters the first parameter so whenever a post request with body was made all following get requests had the same body set which lead to errors oc

.circleci/config.yml
```diff
@@ -517,6 +517,9 @@ jobs:
             - run:
                 command: yarn lint
                 name: Lint
+            - run:
+                command: yarn workspace web stylelint
+                name: Stylelint
```
SteffiStuffel, line 522: Is style-linting not available for native?

api-client/src/Endpoint.ts
```diff
@@ -61,7 +61,9 @@ class Endpoint<P, T> {
     const response = await fetch(url, requestOptions).catch((e: Error) => {
       throw new FetchError({
         endpointName: this.stateName,
-        innerError: e
+        innerError: e,
+        url,
```
SteffiStuffel, line 65: Can you explain why it is necessary to add this params or how they fix the error?

SteffiStuffel (approved):
Not tested, as I do not know how to reproduce.

api-client/src/Endpoint.ts
klinzo: Just adds more information to debug, not related to fixing the actual error.

.circleci/config.yml
klinzo: No idea, at least we don't use it on native.

Taggotty (approved):

PR #474, opened 2021-10-26
IGAPP-811: Add meta tags to prevent caching of index.html

f1sh1918:

This issue is connected to 
https://issues.tuerantuer.org/browse/IGAPP-735

Unfortunately i don't see a way to test this. We can just check the sentry events
https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/186/?project=2&referrer=jira_integration


Taggotty (approved):

maxammann:
@f1sh1918 I do not think this will fix it. We do not delete previous bundles from the server. This means even if you use an older index.html everything should work.

PR #475, opened 2021-10-26
IGAPP-812: Custom logging function

steffenkleinle:

### DIFF
--- a/api-client/src/endpoints/__tests__/createFeedbackEndpoint.spec.ts
+++ b/api-client/src/endpoints/__tests__/createFeedbackEndpoint.spec.ts
@@ -28,9 +28,9 @@ describe('feedback', () => {
         feedbackType: PAGE_FEEDBACK_TYPE,
         feedbackCategory: CONTENT_FEEDBACK_CATEGORY,
         isPositiveRating: true,
-        permalink: `/augsburg/de`
+        permalink: `/augsburg/de/willkommen`
       })
-    ).toEqual('https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/feedback/categories')
+    ).toEqual('https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/feedback')
   })
   it('should map the params to the body', () => {
     const formData = new FormData()


klinzo:
@maxammann what do you think of this?

build-configs/tools/manage.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
maxammann, line 1: I think we should disable this rule for the whole tools directory

native/src/redux/reducers/__tests__/morphContentLanguage.spec.ts
```diff
@@ -12,14 +12,23 @@ import {
...
 import { MorphContentLanguageActionType, PushCategoryActionType, PushEventActionType } from '../../StoreActionType'
 import createCityContent from '../createCityContent'
 import morphContentLanguage from '../morphContentLanguage'
 import pushCategory from '../pushCategory'
 import pushEvent from '../pushEvent'
 
+jest.mock('../../../utils/helpers', () => ({
+  forEachTreeNode: jest.requireActual('../../../utils/helpers').forEachTreeNode,
+  log: jest.fn()
+}))
```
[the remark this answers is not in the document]
maxammann: but this is no e2e test

native/src/sagas/__tests__/loadCityContent.spec.ts
```diff
@@ -29,6 +29,7 @@ import loadPois from '../loadPois'
 
 jest.mock('../../utils/helpers', () => ({
   ...jest.requireActual('../../utils/helpers'),
+  log: jest.fn(),
   reportError: jest.fn()
 }))
```
maxammann, line 34: I think we should mock this by default or not mock this at all. Why are we mocking log and reportError?

native/src/utils/helpers.ts
```diff
@@ -9,13 +9,75 @@ import { FetchError, NotFoundError } from 'api-client/src'
 import buildConfig from '../constants/buildConfig'
 import AppSettings from './AppSettings'
 
+const sentryEnabled = (): boolean => buildConfig().featureFlags.sentry
+const developerFriendly = (): boolean => buildConfig().featureFlags.developerFriendly
+
+export const initSentry = (): void => {
+  if (!sentryEnabled()) {
```
maxammann, line 16: Do we require mocking because of sentry? can we avoid mocking somehow?

maxammann (approved):

maxammann:
> @maxammann what do you think of this?

looks good!

native/src/sagas/__tests__/loadCityContent.spec.ts
klinzo: `reportError` needs to be mocked/spied to be able to assert on it. Removed mocking of `log`

native/src/utils/helpers.ts
klinzo: No, I just mock when I want to assert that the function was actually called.

PR #476, opened 2021-10-27
IGAPP-776: Export app settings instance instead of class

klinzo:


native/src/utils/AppSettings.ts
```diff
@@ -168,6 +168,6 @@ class AppSettings {
   }
 }
 
-export const appSettings = new AppSettings()
+const appSettings = new AppSettings()
 
-export default AppSettings
+export default appSettings
```
SteffiStuffel, line 173: Just merging the two lines into one should be fine.

SteffiStuffel (approved):

native/src/utils/AppSettings.ts
klinzo: I prefer having named constants instead of just exporting something as default :)

f1sh1918 (approved):

PR #479, opened 2021-10-29
IGAPP-765: Consistent display of news preview

Cmd190:
See https://issues.tuerantuer.org/secure/RapidBoard.jspa?rapidView=14&view=detail&selectedIssue=IGAPP-765. On native a line break is displayed after the author whereas on web there was no line break after the author.
Web now looks like this:

image


web/src/utils/stringUtils.ts
```diff
@@ -1,8 +1,8 @@
 import normalizeStrings from 'normalize-strings'
 
-export const textTruncator = (text: string, numOfCharsAllowed: number): string => {
+export const textTruncator = (text: string, numOfCharsAllowed: number, replaceNewLineByLineBreak?: boolean): string => {
```
Znippet, line 3: 🎨 replaceNewLineByLineBreak: boolean = false
So the type is always boolean, not boolean|null

Znippet (approved):

web/src/components/NewsListItem.tsx
```diff
@@ -61,7 +57,12 @@ const NewsListItem = ({ title, content, timestamp, formatter, t, type, link }: P
       <Link to={link}>
         <Description>
           <Title>{title}</Title>
-          <Body>{textTruncator(content, NUM_OF_CHARS_ALLOWED)}</Body>
+          <RemoteContent
```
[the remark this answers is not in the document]
klinzo: Actually I think we should not render the text as actual HTML here, otherwise formatting, links and so on will be applied. E.g. links do not make any sense as clicking on it just opens the news details page and not the link.

klinzo (changes requested):
I think your solution is a little bit of an overkill as we don't need to render full html but just want to display linebreaks. Have a look e.g. at the following article: https://darioghilardi.com/handling-newlines-with-react/
I think both 3) and 4) are fine solutions with 4) probably easier (I also tested it already). If you chose 3), it would probably best to create a custom component and use `<>` fragments. Then you just need a flag `removeNewLines` in `textTruncator`. A release note would be nice as well :)

web/src/utils/stringUtils.ts
Cmd190: Done

web/src/components/NewsListItem.tsx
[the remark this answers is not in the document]
Cmd190: @ klinzo 4) sounds like a nice and simple solution but I don't undestand how to apply it. The following does not work: 
```
const Body = styled.p`
  font-size: 16px;
  line-height: 1.38;
  white-space: pre-line;
`
```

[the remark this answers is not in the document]
Cmd190: Fixed. I took solution 4)

web/src/utils/stringUtils.ts
```diff
@@ -1,8 +1,9 @@
 import normalizeStrings from 'normalize-strings'
 
-export const textTruncator = (text: string, numOfCharsAllowed: number): string => {
+export const textTruncator = (text: string, numOfCharsAllowed: number, replaceLineBreaks = true): string => {
   const ellipsis = '...'
-  const cleanText = text.replace(/\n/g, ' ')
+  // const cleanText = text.replace(/\n/, replaceTextLineBreakByHtmlBreak ? '<br/>' : ' ')
```
klinzo, line 5:
```suggestion
```

web/src/utils/stringUtils.ts
```diff
@@ -1,8 +1,9 @@
 import normalizeStrings from 'normalize-strings'
 
-export const textTruncator = (text: string, numOfCharsAllowed: number): string => {
+export const textTruncator = (text: string, numOfCharsAllowed: number, replaceLineBreaks = true): string => {
```
klinzo, line 3: How do we handle this on native? Maybe we could move this to the api-client

Cmd190: <ListItemView language={language}>
              <Content numberOfLines={5} language={language}>
                {content}
              </Content>
 </ListItemView>

```
export const Content = styled.Text<{ language: string }>`
  font-size: 14px;
  letter-spacing: 0.5px;
  text-align: ${props => contentAlignment(props.language)};
`
```

Cmd190: TextTrunctator is only used in web.

Taggotty: in native we yould pass the number of lines which we sadly cannot pass in web

Taggotty (approved):
Lovely solution. Didn't know about this css attribute

klinzo (changes requested):
Tested, but this misses an important point from the issue:
> show a fixed amount of chars in web dependent on the viewport

You should either do this in this PR or create a separate PR/Issue.

maxammann (approved):

Cmd190:
> Tested, but this misses an important point from the issue:
> 
> > show a fixed amount of chars in web dependent on the viewport
> 
> You should either do this in this PR or create a separate PR/Issue.

I completely  oversaw this point. But we should dicuss this further. Imo the amount of chars should not be dependent from the viewport.

klinzo (approved):
Okay, lets merge this but not close the issue. Just open a discussion in mattermost or in our daily/team call.

PR #480, opened 2021-10-29
IGAPP-259: Add contrast for setting switches

Cmd190:
See https://issues.tuerantuer.org/browse/IGAPP-259. The contrast was improved by adding transparency to the trackColor. The trackColor is the same as described here https://app.zeplin.io/project/5f9040aed45a280985d019fb/screen/600863648dd7dea87cc759ff. 
The contrast is better than before but still a little weak. Improving it through setting the shadow is not possible for iOS. 
It nows look this:
Android: 
Screenshot_20211029-173731

iOS:

image


native/src/components/SettingsSwitch.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { ReactElement } from 'react'
+import { Switch } from 'react-native'
+
+import { ThemeType } from 'build-configs/ThemeType'
+
+type PropsType = {
+  theme: ThemeType
```
klinzo, line 7: You can remove this prop and just use the `useTheme` hook

build-configs/integreat/theme/colors.ts
```diff
@@ -1,8 +1,10 @@
 import { ColorsType, commonLightColors } from '../../common/theme/colors'
 
 const themeColor = '#fbda16'
+const themeColorLight = 'rgba(251, 218, 22, 0.5)'
```
klinzo, line 4: You could also use hex notation here: `#fbda1680`
See e.g. https://rgbacolorpicker.com/rgba-to-hex

Regarding the used colors: I am fine with this as this was proposed by a designer, but if you want we can also wait for additional expertise from our new designer starting on Tuesday :)

native/src/components/SettingsSwitch.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onPress?: () => void
```
klinzo, line 8: Is it possible/necessary that `onPress` is `undefined`?

klinzo (changes requested):
A short unit test for your new component asserting that the `onPress` function is actually called would be nice. If you don't now how to write one, let me know :)

Please also write a release note :writing_hand:

build-configs/integreat/theme/colors.ts
Cmd190: @klinzo It is for sure the best option to ask the new designer which colors we should chose here

native/src/components/SettingsSwitch.tsx
Cmd190: Indeed...it would even be an error it `onPress` is `undefindend`

Taggotty (approved):

Taggotty:
Tested on Android Emulator. Release note still missing :). And personally I still find the contrast relatively low. But maybe thats just by perception.

build-configs/integreat/theme/colors.ts
maxammann: I would say we merge this and create a new issue :)

maxammann (approved):

native/src/routes/JpalTracking.tsx
```diff
@@ -108,15 +109,11 @@ const JpalTracking = (props: PropsType) => {
...
-              true: props.theme.colors.themeColor,
-              false: props.theme.colors.backgroundAccentColor
-            }}
+          <SettingsSwitch
+            theme={theme}
             value={!!trackingEnabled}
-            onValueChange={toggleTrackingEnabled}
-            testID='switch'
+            onPress={toggleTrackingEnabled}
+            accessibilityLabel='switch'
```
klinzo, line 116: You should use `accessibilityRole` instead here

klinzo (approved):
Not tested again

PR #482, opened 2021-10-30
IGAPP-815: Store tarball instead of single files

klinzo:


Taggotty:
Do you have some more input for this?

SteffiStuffel (commented):
Is this just for the IGAPP-815 Task and will be reverted or do we always want to do this?

klinzo:
> Do you have some more input for this?

@Taggotty it is very tedious to download all files separately, by saving it as .tar.gz you just need to download one file and can just unpack it afterwards.

Taggotty (approved):

klinzo:
> Is this just for the IGAPP-815 Task and will be reverted or do we always want to do this?

@SteffiStuffel I would just keep this as I don't see a reason not to keep it. Makes sense imo

PR #487, opened 2021-11-04
IGAPP-825: Language change broken

klinzo:

### DIFF
--- a/native/src/hocs/withPayloadProvider.tsx
+++ b/native/src/hocs/withPayloadProvider.tsx
@@ -78,7 +78,7 @@ const withPayloadProvider = <
   T extends RoutesType
 >(
   refresh: (refreshProps: R, dispatch: Dispatch<StoreActionType>) => void,
-  clearRouteOnClose?: boolean,
+  clearRouteOnClose: boolean,
   noScrollView?: boolean
 ): ((Component: React.ComponentType<S>) => React.ComponentType<PropsType<S, R, T>>) => (
   Component: React.ComponentType<S>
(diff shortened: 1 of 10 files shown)


native/src/hooks/useClearRouteOnClose.ts
```diff
@@ -4,17 +4,23 @@ import { Dispatch } from 'redux'
 import { RoutePropType, RoutesType } from '../constants/NavigationTypes'
 import { StoreActionType } from '../redux/StoreActionType'
 
-const useClearRouteOnClose = (route: RoutePropType<RoutesType>, dispatch: Dispatch<StoreActionType>): void => {
+const useClearRouteOnClose = (
+  route: RoutePropType<RoutesType>,
+  dispatch: Dispatch<StoreActionType>,
+  clearRouteOnClose = false
```
maxammann, line 10: I think this should be a quired property to make clear that this is important. Not clearing by default is weird imo.

maxammann (approved):

native/src/hocs/withPayloadProvider.tsx
```diff
@@ -66,20 +67,28 @@ export type PropsType<
...
 const withPayloadProvider = <
-  S extends { dispatch: Dispatch<StoreActionType> },
+  S extends {
+    dispatch: Dispatch<StoreActionType>
+  },
   R extends Record<string, unknown>,
   T extends RoutesType
 >(
   refresh: (refreshProps: R, dispatch: Dispatch<StoreActionType>) => void,
+  clearRouteOnClose: boolean,
```
SteffiStuffel, line 81: Why is this needed now?

SteffiStuffel (approved):
Tested works fine

native/src/hocs/withPayloadProvider.tsx
klinzo: Because not all routes need to clear the route from redux on close. See the changes from #457, before a function `onRouteClose` was passed wherever necessary, but since the function is always the same (just dispatch an action) we don't need to pass it and a boolean is enough.

PR #497, opened 2021-11-08
IGAPP-755: Scrolling in e2e tests

Taggotty:


e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,178 @@
+import { RectReturn } from '@wdio/protocols/build/types'
```
Taggotty, line 1: Checkout the original https://github.com/webdriverio/appium-boilerplate/blob/main/tests/helpers/Gestures.ts

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
+// For reference https://github.com/webdriverio/appium-boilerplate/blob/main/tests/helpers/Gestures.ts
```
Taggotty, line 1: This file is mainly interesting because it is a lot better commentated an structured than the previous implementation and it provides this helper `checkIfDisplayedWithSwipeUp`

Taggotty:
Will wait for IGAPP-675

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
...
+      this.calculateXY(SWIPE_DIRECTION.right.start, percentage),
+      this.calculateXY(SWIPE_DIRECTION.right.end, percentage)
+    )
+  }
+
+  /**
+   * Swipe from coordinates (from) to the new coordinates (to). The given coordinates are
+   * percentages of the screen.
+   */
+  static async swipeOnPercentage(from: XY, to: XY): Promise<void> {
```
klinzo, line 111: Does this take percentages or coordinates? I think either the naming or the comment is wrong here. Perhaps it makes sense to just take a percentage and a `SWIPE_DIRECTION` as it is only used that way and calculate xy here

klinzo (approved):
Very well documented :muscle:

SteffiStuffel (approved):

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
+// For reference https://github.com/webdriverio/appium-boilerplate/blob/main/tests/helpers/Gestures.ts
+import { RectReturn } from '@wdio/protocols/build/types'
+import { ChainablePromiseElement } from 'webdriverio'
+
+/**
```
maxammann, line 5:
```suggestion
/*
```

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
...
+ * To make a Gesture methods more robust for multiple devices and also
+ * multiple screen sizes the advice is to work with percentages instead of
+ * actual coordinates. The percentages will calculate the position on the
+ * screen based on the SCREEN_SIZE which will be determined once if needed
+ * multiple times.
+ */
+
+let SCREEN_SIZE: RectReturn | null
```
maxammann, line 13: I think this should be a variable of Gesture class. Else you can not test on multiple devices with different screen sizes in parallel.

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
...
+    // Get the move to position on the screen for the swipe
+    const moveToScreenCoordinates = this.getDeviceScreenCoordinates(SCREEN_SIZE, to)
+
+    await this.swipe(pressOptions, moveToScreenCoordinates)
+  }
+
+  /**
+   * Swipe from coordinates (from) to the new coordinates (to). The given coordinates are in pixels.
+   */
+  static async swipe(from: XY, to: XY): Promise<void> {
```
maxammann, line 129: You could use JSDoc params to document the paramters: https://jsdoc.app/tags-param.html

e2e-tests/native/test/helpers/Gestures.ts
```diff
@@ -0,0 +1,179 @@
...
+    ])
+    // Add a pause, just to make sure the swipe is done
+    await driver.pause(WAIT_FOR_SWIPE_FINISHED)
+  }
+
+  /**
+   * Get the screen coordinates based on a device his screen size
+   */
+  private static getDeviceScreenCoordinates(screenSize: RectReturn, coordinates: XY): XY {
+    const hundredPercent = 100
```
maxammann, line 161:
```suggestion
    const HUNDRET_PERCENT = 100
```

e2e-tests/native/test/helpers/Selector.ts
```diff
@@ -10,6 +10,15 @@ export class Selector {
     return this
   }
 
+  public ByPartialText(text: string): Selector {
```
maxammann, line 13:
```suggestion
  public ByContainedText(text: string): Selector {
```

maxammann (approved):

e2e-tests/native/test/helpers/Gestures.ts
Taggotty: I can move it into the Gesture class, but as all methods are static the `SCREEN_SIZE` would also need to be static. The reasoning behind that is that we don't want to recompute `SCREEN_SIZE`. However I am not sure if it would really not be possible to tes on multiple devices, as I think wdio starts a new process for each capability passed.

e2e-tests/native/test/helpers/Gestures.ts
Taggotty: from and two are both coordinates relative to the screen width and height. Which makes those coordinate percentages

Znippet:
> Will wait for [IGAPP-675](https://github.com/Integreat/integreat-app/pull/500)

Merged into main

PR #498, opened 2021-11-08
IGAP-628: Add jpal translations and adjustments

klinzo:


native/src/routes/JpalTracking.tsx
```diff
@@ -30,43 +32,94 @@ const DescriptionContainer = styled.TouchableOpacity`
...
+          return
+        }
+        if (trackingCode?.length === TRACKING_CODE_LENGTH && trackingEnabled) {
+          // Tracking enabled and (hopefully) valid code entered, so we are good to leave the screen
+          return
+        }
+
+        e.preventDefault()
+
+        Alert.alert(t('trackingLeaveTitle'), t('trackingLeaveDescription', { appName: buildConfig().appName }), [
```
klinzo, line 96: Not beautiful to use an alert here but we need some buttons so the snackbar would not work.

translations/translations.json
```diff
@@ -5033,9 +5033,15 @@
       "tracking": "J-PAL Evaluation",
       "allowTracking": "Tracking zulassen",
       "trackingCode": "Tracking Code",
-      "trackingDescription": "Durch das Senden von Tracking-Daten Integreat langfristig verbessern.",
+      "trackingShortDescription": "Nehmen Sie an der J-PAL Evaluierungsstudie teil, um {{appName}} zu verbessern.",
```
klinzo, line 5036: Translations are from here if anyone wants to have a look, copied them manually: https://docs.google.com/document/d/1X7pfXlFT2M2E7zFVkMsul5rnzbANSusd/edit?usp=sharing&ouid=100822204933515470934&rtpof=true&sd=true

klinzo (commented):
Please test this :)

native/src/components/Link.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { ReactElement } from 'react'
```
f1sh1918, line 1: Just a general style issue. If i click on the link, background gets black. Looks a bit ugly ;)

native/src/routes/JpalTracking.tsx
```diff
@@ -119,15 +162,19 @@ const JpalTracking = ({ route, theme }: PropsType) => {
...
         <Input
+          error={invalid && trackingCode?.length !== TRACKING_CODE_LENGTH}
           value={trackingCode ?? ''}
           onChangeText={updateTrackingCode}
           theme={theme}
           editable={!!trackingEnabled}
           testID='input'
         />
+
+        {error && <ErrorText>{t(error)}</ErrorText>}
```
SteffiStuffel, line 173: This won't translate the error as the namespace is missing. Should be changed to `{error && <ErrorText>{t(`error:${error}`)}</ErrorText>}`

native/src/routes/JpalTracking.tsx
```diff
@@ -30,43 +32,94 @@ const DescriptionContainer = styled.TouchableOpacity`
...
+  }, [])
 
+  useEffect(
+    () =>
+      navigation.addListener('beforeRemove', e => {
+        if (!routeTrackingCode) {
+          // User probably just opened the settings out of interest, don't annoy em by opening an alert
+          return
+        }
+        if (trackingCode?.length === TRACKING_CODE_LENGTH && trackingEnabled) {
```
SteffiStuffel, line 89: So it you don't have a trackingCode in the url, it is fine to enter a TrackingCode with wrong lenght? Not sure if this is really wanted behaviour. Also would it make sense to give the user some feedback what the correct lenght would be?

native/src/routes/JpalTracking.tsx
```diff
@@ -30,43 +32,94 @@ const DescriptionContainer = styled.TouchableOpacity`
   justify-content: space-between;
   padding: 15px 0 5px;
 `
-const Input = styled(TextInput)`
+const Input = styled(TextInput)<{ editable: boolean; error: boolean }>`
   padding: 15px;
   border-width: 1px;
   border-color: ${props => (props.editable ? props.theme.colors.themeColor : props.theme.colors.textDisabledColor)};
+  ${props => (props.error ? 'border-color: red;' : '')};
```
SteffiStuffel, line 39: Not working. Border is not red, although an error occured. See attached screenshot.

SteffiStuffel: Oke, only working when entering the jpal-settings-screen with a deep-link.

native/src/routes/JpalTracking.tsx
SteffiStuffel: I get a wrong error message here:
Steps to reproduce:
1. Enter jpal-settings via deep link
2. keep switch in allowed-position but change tracking code to a code with wrong length
3. navigate back
4. Get the error that you should allow tracking, but it is allowed.

SteffiStuffel (commented):
Tested on Android Emulator. The error message is hidden behind the keyboard, which is really not good.
When the border is red, a user can at least imagine that something went wrong, but when the border is not red, it just does not save the changes and the user gets no feedback.

Also not sure if it is good, that is saving of a change (switch or textfield) fails only the error is displayed, but the toggle/text-field stays in the new-input which is not saved. So it is not possible to try again and also as a user i wouldn't be sure if the changes are now saved or not, i  would expect it to be resetted to the old value.

IGAPP-628
IGAPP-628-hidden-error

native/src/components/Link.tsx
klinzo: can you take a screenshot/video?

native/src/routes/JpalTracking.tsx
klinzo: It was included, see above. But not relevant anymore.

native/src/routes/JpalTracking.tsx
klinzo: See new changes

native/src/routes/JpalTracking.tsx
SteffiStuffel: Well, it was not working, I tested it.

klinzo: Was working for me :thinking:

klinzo:
Still working on adding tests, maybe will do that in a separate PR

SteffiStuffel (approved):
Tested, JpalSettings are not visible when not opening via deep-link, but is visable and correctly opened and set, when opening via deep-link.

Things i did not like:
- When leaving the JPAL-Tracking-Site with tracking disabled, when clicking on "Allow tracking"-Button, the switch is now colored in red. I would just expect it to be toggled and be now yellow and in enabled-status.
- When disabling tracking the color of the switch is depending on the fact if the quit-modal was shown before or not, this is not really intuitiv. Either it should always or never be red in disabled position
- Now we have the solution, that the tracking code cannot be edited, what really makes sense, but it would be appropriate to display it (if we want to display it all) just as a text and not as a text-input, especially because the textfield does not really look disabled, but also will never be enabled and its a bit confusing.

klinzo:
@SteffiStuffel could you have another look please? I figured we don't need to display the tracking code to the user at all. Tracking is now enabled on click on allow tracking.

f1sh1918 (commented):
Tested on android, ios emulator and ios native.
Unfortunately i couldn't get the deep link work on ios native (Iphone se 2020)
https://integreat.app/jpal/123456789
The app opens but no redirect to jpal

f1sh1918 (approved):

native/src/components/Link.tsx
f1sh1918: Simulator Screen Recording - iPhone 12 Pro Max - 2021-11-09 at 17 42 00

SteffiStuffel:
> Still working on adding tests, maybe will do that in a separate PR

Looks good.

native/src/routes/__tests__/JpalTracking.spec.tsx
```diff
@@ -1,39 +1,80 @@
...
 import appSettings from '../../utils/AppSettings'
 import JpalTracking from '../JpalTracking'
 
 jest.mock('react-i18next')
+jest.mock('react-native', () => ({
+  ...jest.requireActual('react-native'),
+  Alert: {
+    alert: jest.fn()
+  }
+}))
```
Taggotty, line 18: Pls remove. Not needed.

native/src/components/Link.tsx
Taggotty: Could not reproduce this on android emulator o.o

Taggotty:
teted on android emulator

native/src/components/Link.tsx
klinzo: Created https://issues.tuerantuer.org/browse/IGAPP-836

PR #500, opened 2021-11-08
IGAPP-675: E2E test for language change

Znippet:


e2e-tests/native/test/pageobjects/dashboard.page.ts
```diff
@@ -12,6 +13,21 @@ class DashboardPage extends Page {
...
+
+  get languageIcon() {
+    return $('~Change language')
+  }
+
+  get headerBack() {
+    return $('~Go back')
+  }
+
+  public async switchLanguageTo(language: string) {
```
SteffiStuffel, line 25: I think this should not be in the `.page`. Would make more sense to have it in the test.

SteffiStuffel (approved):

Taggotty:
Checkout the failing CI: 
iOS: https://app-automate.browserstack.com/builds/1887d102b75a0fe4271d40062b8a547ad89907f0/sessions/5e1fb207413c6439d7a39600df815019ee139f99?auth_token=dab544580f87cb9eb5b7a94cdf2cc93a5d5e51cf93129aedc3f6abb549a5bc27
android: https://app-automate.browserstack.com/builds/1887d102b75a0fe4271d40062b8a547ad89907f0/sessions/162dc240dd03ea74903c78567005c5173d36dd65?auth_token=7aa5aa6de8ae52ee991597ace24bd7417b7b1af3fd19cc8e2834d7910a4ac8e8
On Android it's probably enough to increase the correct timeout here. On iOS you'll have to checkout why the header-back button is not clicked. :)

klinzo (approved):
Could you please try whether this test would have failed before the fix in #487 ?

e2e-tests/native/test/pageobjects/dashboard.page.ts
Znippet: Fine for me also.
The thought behind this is to keep the tests as readable as possible and leave the implementation details inside the page object

So we could reuse the switchToLanguage also if we e.g. want to test a "date" component that should format the date for different languages (only example, no idea if we have that)

klinzo:
> Could you please try whether this test would have failed before the fix in #487 ?

@Znippet did you try?

PR #501, opened 2021-11-10
IGAPP-170: Show snackbar if sharing fails

Cmd190:
See title


native/src/components/TransparentHeader.tsx
```diff
@@ -63,30 +64,29 @@ class TransparentHeader extends React.PureComponent<PropsType> {
         message
       })
     } catch (e) {
-      // TODO Show snackbar
-      reportError(e)
+      const error = e as Error
```
SteffiStuffel, line 67: Why is this needed?

SteffiStuffel (approved):
Tested on Android Emulator. Works as expected.

native/src/components/TransparentHeader.tsx
```diff
@@ -63,30 +64,29 @@ class TransparentHeader extends React.PureComponent<PropsType> {
...
+      showSnackbar(error.message)
```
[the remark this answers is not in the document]
f1sh1918:
```suggestion
      showSnackbar(t('generalError')
```

Taggotty (changes requested):
Please also show the snackbar for the "normal" header component

native/src/components/TransparentHeader.tsx
Cmd190: If you do not cast to error explicitley you will get linting erros because  e is of type unknown and reportError() expects an Error as parameter

native/src/components/TransparentHeader.tsx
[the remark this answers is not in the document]
Cmd190: Your're absolutley right

native/src/components/TransparentHeader.tsx
klinzo: Should only be necessary starting with typescript 4.4

Taggotty (approved):

native/src/components/TransparentHeader.tsx
Cmd190: So let's leave it for future compatability

klinzo: Actually I'd prefer to remove it again and do this all at once as we probably need this in many places and can find a general solution then.

native/src/components/TransparentHeader.tsx
```diff
@@ -63,30 +66,27 @@ class TransparentHeader extends React.PureComponent<PropsType> {
         message
```
klinzo, line 66: Perhaps you could add a title here the same way as it is in `Header`

klinzo (commented):
Could you perhaps add tests for this?

native/src/components/TransparentHeader.tsx
Cmd190: ok, you've got the point

klinzo (approved):

PR #505, opened 2021-11-15
IGAPP-735: Added retry function for lazy imports

f1sh1918:

I created this as a draft, since i dunno if this the best approach. Caching policies didn't solve the issue.
All i read is that, since React 16.6 the lazy imports were provided as an experimental feature which may lead to 'Chunk load errors'
All i found so far to solve this issue  was to retry the import, so i added this retry function, please feel free to discuss


web/src/RootSwitcher.tsx
```diff
@@ -21,14 +21,15 @@ import buildConfig from './constants/buildConfig'
...
 type PropsType = {
   setContentLanguage: (languageCode: string) => void
 }
 
-const MainDisclaimerPage = React.lazy(() => import('./routes/MainDisclaimerPage'))
-const LandingPage = React.lazy(() => import('./routes/LandingPage'))
-const NotFoundPage = React.lazy(() => import('./routes/NotFoundPage'))
+const MainDisclaimerPage = React.lazy(() => retryImport(() => import('./routes/MainDisclaimerPage')))
+const LandingPage = React.lazy(() => retryImport(() => import('./routes/LandingPage')))
+const NotFoundPage = React.lazy(() => retryImport(() => import('./routes/NotFoundPage')))
```
klinzo, line 32: Why only for these three pages?

web/src/utils/retryImport.ts
```diff
@@ -0,0 +1,30 @@
+import { ComponentType } from 'react'
```
klinzo, line 1: Did you have a look at [this proposal](https://gist.github.com/raphael-leger/4d703dea6c845788ff9eb36142374bdb#file-lazywithretry-js) from [this article](https://raphael-leger.medium.com/react-webpack-chunkloaderror-loading-chunk-x-failed-ac385bd110e0)? The approach is a little bit different, he also says one should only try twice and throw an error if that does not help. 

He reloads the whole page why you are just retrying the import. I am no expert here, could you explain why you chose this solution? Reproducing/testing this is sadly very difficult :/

klinzo (changes requested):

web/src/RootSwitcher.tsx
f1sh1918: we only have these three lazy imports or did i miss sth. Other import should be included in the main chunk and should not fail

klinzo: Have a look at `CityContentSwitcher`

web/src/utils/retryImport.ts
f1sh1918: Well because this particular chunk failed to load during the lazy import so for me it makes sense to retry importing/loading this chunk again if it fails.
Well there are very recent articles about that case also here f.e.
https://medium.com/@botfather/react-loading-chunk-failed-error-88d0bb75b406
You can also find this solving approach in further blogs and git issues since this feature was released with React 16.6
And yes testing is difficult. I even could never reproduce this error.

And we could reduce the retry attempts. I think three can be fine, too.

web/src/RootSwitcher.tsx
f1sh1918: just added retry fkt to `CityContentSwitcher`

web/src/CityContentSwitcher.tsx
```diff
@@ -36,17 +36,18 @@ import {
...
-const EventsPage = React.lazy(() => import('./routes/EventsPage'))
-const CategoriesPage = React.lazy(() => import('./routes/CategoriesPage'))
-const LocalNewsPage = React.lazy(() => import('./routes/LocalNewsPage'))
-const SprungbrettOfferPage = React.lazy(() => import('./routes/SprungbrettOfferPage'))
-const PoisPage = React.lazy(() => import('./routes/PoisPage'))
-const SearchPage = React.lazy(() => import('./routes/SearchPage'))
-const DisclaimerPage = React.lazy(() => import('./routes/DisclaimerPage'))
+import retryImport from './utils/retryImport'
+
+const TuNewsDetailPage = React.lazy(() => retryImport(() => import('./routes/TuNewsDetailPage')))
```
klinzo, line 41: Could you perhaps refactor your `retryImport` to something like `lazyWithRetry`? I.e. move the `React.lazy` inside it such that you can write the following:
```suggestion
const TuNewsDetailPage = lazyWithRetry('./routes/TuNewsDetailPage')
```
or
```suggestion
const TuNewsDetailPage = lazyWithRetry(() => import('./routes/TuNewsDetailPage'))
```

klinzo (approved):

klinzo:
Tested on firefox

web/src/CityContentSwitcher.tsx
f1sh1918: good advice. done

SteffiStuffel (approved):

PR #509, opened 2021-11-16
IGAPP-834: Upgrade react navigation

klinzo:


native/src/Navigator.tsx
```diff
@@ -178,141 +178,37 @@ const Navigator = (props: PropsType): ReactElement | null => {
...
-      />
-      <Stack.Screen
-        name={JPAL_TRACKING_ROUTE}
-        component={JpalTracking}
-        options={{
-          header: transparentHeader
-        }}
-      />
+    <Stack.Navigator initialRouteName={initialRoute.name} screenOptions={{ ...transitionPreset, headerMode: 'screen' }}>
+      <Stack.Group screenOptions={{ header: () => null }}>
```
klinzo, line 182: These `Group`components just set `options` for all child screens which looks a lot nicer imo.

f1sh1918 (approved):
Tested on android emulator and ios native. works as expected :)

Taggotty (approved):
Tested on Android emulator

PR #515, opened 2021-11-22
IGAPP-839: Fixed missing overflow for p-tags in webview

f1sh1918:

Go to Kreis Höxter -> Asyl und Migration -> Information zum Asylverfahren
Switch language to FA

Text should be displayed correctly on ios & android (it was buggy on both systems!)

Tested on android emulator, ios native & emulator

### DIFF
--- a/release-notes/unreleased/IGAPP-839-Fix-Paragraph-Overflow-Native.yml
+++ b/release-notes/unreleased/IGAPP-839-Fix-Paragraph-Overflow-Native.yml
@@ -1,0 +1,7 @@
+issue_key: IGAPP-839
+show_in_stores: true
+platforms:
+  - ios
+  - android
+en: Fixed faulty display of paragraphs in webview.
+de: Fehlerhafte Darstellung von Paragraphen im Webview wurde behoben.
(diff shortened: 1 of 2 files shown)


Taggotty (approved):
Tested on Android device. Could also reproduce the original issue with my Sony XPeria XA1 with Android 8. Nice find 👍🏼

release-notes/unreleased/IGAPP-839-Fix-Paragraph-Overflow-Native.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-839
+show_in_stores: true
+platforms:
+  - ios
+  - android
+en: Fixed faulty display of paragraphs in webview.
+de: Fehlerhafte Darstellung von Paragraphen im Webview wurde behoben.
```
klinzo, line 7: These are implementation details that rather confuse users I think. Better to just use something generic like ` Fixed faulty display in certain cases.` or simlar

klinzo (approved):
Tested on android, nice work :tada: 
Yeah I guess I wasn't looking closely, I just saw that it is not a blank screen like Salua reported on Android, sorry :/

PR #518, opened 2021-11-26
IGAPP-850: iOS notification permission handling

klinzo:

Basically just gets rid of throwing an exception and uses return value instead to reset the shown settings. Sadly it is not possible to open apps notification settings directly. Maybe we could show some hint though before opening settings which should be discussed/designed here: https://issues.tuerantuer.org/browse/IGAPP-854


Taggotty (approved):
Uhhm. Tested on Android emulator ^^

translations/translations.json
```diff
@@ -2401,20 +2401,8 @@
...
       "offersDescription": "Jobsuche, Praktikasuche und mehr in der Umgebung",
-      "inquiryTitle": "Einstellungen",
-      "inquiryIntro": "Durch Klicken auf \"{{accept}}\" erlaube ich der {{appName}}-App, ...",
-      "sentryTitle": "App-Stabilität",
-      "sentryCondition": "... mit dem automatischen Senden von Absturzberichten {{appName}} zu verbessern",
-      "pushNewsTitle": "Push-Benachrichtigungen",
-      "pushNewsCondition": "... mir Push-Benachrichtigungen zu den neuesten lokalen Nachrichten zu senden",
       "skip": "Überspringen",
-      "next": "Weiter",
-      "cancel": "Abbrechen",
```
SteffiStuffel, line 2412: Do we really want to delete these translations, its really generic stuff we may need sometime in the future?

SteffiStuffel (approved):
Not tested as I cannot test ios-stuff right now.

translations/translations.json
klinzo: If we need them we can either translate them again or we look in the history and revert the commit. No need to keep them.

PR #519, opened 2021-11-27
IGAPP-782: Cannot store last update for unused city

klinzo:

The error `Cannot store last update for unused city` is probably just a symptom of a deeper error, since we just returned an empty object (`{}`) if anything went wrong during loading of the meta data of cities. 
Some probable candidate of real error (for at least some occurrences: `JSON Parse error: Unexpected end of input` (https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/353/?project=2&query=is%3Aunresolved&statsPeriod=14d)

Therefore I removed catching the error if something goes wrong during loading of the meta data of cities and added more debug information in some places. Once we have more information, we can look into this again.


native/src/utils/DatabaseConnector.ts
```diff
@@ -247,31 +249,26 @@ class DatabaseConnector {
...
-        log('An error occurred while loading cities from JSON', 'warning')
-        reportError(e)
-      }
+    if (!fileExists) {
+      return {}
     }
 
-    return {}
+    const citiesMetaJson = await this.readFile<MetaCitiesJsonType>(path)
+    return mapValues(citiesMetaJson, cityMeta => ({
```
SteffiStuffel, line 259: Why have you removed the `try ..catch`? I think it would be still nice to have this.

native/src/utils/DatabaseConnector.ts
```diff
@@ -695,14 +693,19 @@ class DatabaseConnector {
     return BlobUtil.fs.exists(path)
   }
 
-  async readFile(path: string): Promise<string> {
+  async readFile<T>(path: string): Promise<T> {
```
SteffiStuffel, line 696: What would happen if the Types do not match. 
E.g. `readFile<CityResourceCacheJsonType>(path)` does not return `CityResourceCacheJsonType`

Is this catched somewhere? Should it?

SteffiStuffel (approved):

native/src/utils/DatabaseConnector.ts
klinzo: No, otherwise operations that don't make sense if this fails will still be executed. This should be caught in some place higher up imo.

native/src/utils/DatabaseConnector.ts
klinzo: No, it is not caught anywhere. I am not sure if that is even possible to assert that. If so, it should be done in a separate issue though.

Taggotty: you could assert that if we had more information on the path. But not sure wether this is worth the work

Taggotty (approved):

native/src/utils/DatabaseConnector.ts
klinzo: A change I made though: If we detect that the data is corrupted, we delete the file now, otherwise we would never recover from it.

SteffiStuffel (approved):

PR #520, opened 2021-11-27
IGAPP-838: Failed to create parent directory

klinzo:

I think the error (https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/310/?statsPeriod=14d) occurs because of a race condition, as it only happens if both events AND pois are disabled. If that is the case, we previously tried to save an empty array to `events.json` and `pois.json` in the same directory. Since the directory does not yet exist, both `writeFile`s attempt to create the parent directory, the first one (events) succeeds while the second one (pois) does not succeed as the directory was not created (again).
The corresponding line in react-native-blob-util: https://github.com/RonRadtke/react-native-blob-util/blob/03aec7206ea46c32ddc3bf0fd76053eea2f364b0/android/src/main/java/com/ReactNativeBlobUtil/ReactNativeBlobUtilFS.java#L61
 mkdirs docs: https://www.tutorialspoint.com/java/io/file_mkdirs.htm
 > The method returns true if the directories was created, with all necessary parent directories; else false.

I'll create an issue and PR in react-native-blob-util. Since it does not make sense to store empty values anyway, I refactored this here as well: https://github.com/RonRadtke/react-native-blob-util/issues/84


SteffiStuffel (approved):

Taggotty (approved):

klinzo:
Will have to revert this as it breaks the language change (recognized thanks to e2e tests again :pray: ). So waiting for fix upstream.

klinzo:
Will be fixed by upgrading react-native-blob-util

PR #522, opened 2021-11-30
IGAPP-856: Skip checks and deliver apk and ipa to browserstack on main

klinzo:

* Tests (check and e2e tests) are not executed for commits on main anymore
* integreat-test-cms apk and ipa are uploaded to browserstack for commits on main

Workflow on feature branches:
image

Workflow on main:
image


.circleci/config.yml
```diff
@@ -916,39 +939,100 @@ workflows:
...
+                    branches:
+                        ignore:
+                            - main
+                name: build_e2e_ios
+                requires:
+                    - bump_version
+                    - check
+            - e2e_ios:
+                context:
+                    - mattermost
```
SteffiStuffel, line 1006: Mattermost context is not needed, as we don't notify in mattermost, when e2e fails on a feature branch, do we?

.circleci/src/workflows/commit.yml
```diff
@@ -5,39 +5,108 @@ jobs:
...
+            - main
+      name: build_e2e_android
+      build_config_name: integreat-e2e
       context:
         - mattermost
+        - tuerantuer-google-play
+        - tuerantuer-android
+      requires:
+        - bump_version
+        - check
```
SteffiStuffel, line 49: missing newline

.circleci/src/workflows/commit.yml
```diff
@@ -5,39 +5,108 @@ jobs:
         - mattermost
       delivery: none
 
+  - build_web:
```
SteffiStuffel, line 8: Would it maybe make sense to split the commit-workflow into to workflows (one for feautre branches and one for main) as they share almost nothing and we can get rid of the multiple filter-checks?

SteffiStuffel (approved):

.circleci/src/workflows/commit.yml
klinzo: Would be nice, yes. But AFAIK it is not possible to filter workflows by branch name. So it would be possible to split it in two files but we'd still need the filters :/

klinzo: Makes sense to split anyway though, you are right :+1:

.circleci/src/jobs/upload_browserstack.yml
```diff
@@ -0,0 +1,23 @@
+# Deliver the app to Browserstack.
+parameters:
+  artifact_name:
```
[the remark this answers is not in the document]
klinzo: Also saved as artifact

.circleci/src/workflows/commit.yml
klinzo: Maybe it is possible after all, will try :thinking:

klinzo: Seems to work, thanks for asking!

.circleci/src/workflows/commit.yml
klinzo: Not missing but by purpose to group jobs that belong together.

.circleci/src/jobs/upload_browserstack.yml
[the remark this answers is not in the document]
klinzo: Will rename nevertheless as name is not good

klinzo:
ipas and apks uploaded to browserstack with the new `upload_browserstack` job now contain the commit sha:
image

Taggotty (approved):

.circleci/src/jobs/upload_browserstack.yml
```diff
@@ -0,0 +1,26 @@
+# Deliver the app to Browserstack.
+parameters:
+  file_name:
+    type: string
+    description: the name of the file to upload
+docker:
+  - image: cimg/ruby:3.0.3
```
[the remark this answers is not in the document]
klinzo: For fastlane(which is used to upload to browserstack).

PR #523, opened 2021-11-30
IGAPP-862: Invalid language change paths for tünews

klinzo:


web/src/components/LocationLayout.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react'
+import React, { ReactElement, ReactNode, useState } from 'react'
```
klinzo, line 1: Only refactored to functional component

web/src/components/Selector.tsx
```diff
@@ -97,14 +97,14 @@ const Selector = ({
...
-            <BoldSpacer>{item.name}</BoldSpacer>
+            <BoldSpacer />
             {item.name}
           </Element>
         )
       }
       return (
         <Element as={Tooltip} key={item.code} text={disabledItemTooltip} flow='up' $enabled={false} $selected={false}>
-          <BoldSpacer>{item.name}</BoldSpacer>
+          <BoldSpacer />
```
SteffiStuffel, line 107: It think you should revert this. 
It was added to fix the jumping of the languages while hovering, because when they are hovered they get bold and therefore jump around.

web/src/components/LanguageFailure.tsx
```diff
@@ -13,31 +13,25 @@ const ChooseLanguage = styled.p`
...
-  }
+const LanguageFailure = ({ cityModel, languageCode, languageChangePaths }: PropsType): ReactElement => {
+  const { t } = useTranslation('error')
+  return (
+    <>
+      <Caption title={cityModel.name} />
+      <ChooseLanguage>{`${t('notFound.language')} ${t('chooseALanguage')}`}</ChooseLanguage>
+      <LanguageSelector
+        isHeaderActionItem={false}
+        pathname={null}
```
SteffiStuffel, line 29: Why ist pathname null now? Should somewhere be checked if it is null?

.gitignore
```diff
@@ -120,5 +120,6 @@ dist
 
 # IntelliJ
 **/workspace.xml
+runConfigurations.xml
```
SteffiStuffel, line 123: Ahh finally :)

web/src/components/LanguageFailure.tsx
```diff
@@ -13,31 +13,25 @@ const ChooseLanguage = styled.p`
...
-      </>
-    )
-  }
+const LanguageFailure = ({ cityModel, languageCode, languageChangePaths }: PropsType): ReactElement => {
+  const { t } = useTranslation('error')
+  return (
+    <>
+      <Caption title={cityModel.name} />
+      <ChooseLanguage>{`${t('notFound.language')} ${t('chooseALanguage')}`}</ChooseLanguage>
+      <LanguageSelector
```
SteffiStuffel, line 27: Would it make sense not only to disable the not-available languages, but to hide them. Think in the header it really makes sense to disable them, but in a list, which says 'select a option' it is a little uninutitive.

SteffiStuffel (approved):
Tested, works fine :tada:

web/src/components/LanguageFailure.tsx
klinzo: Good idea, I think I agree :+1: Other opinions?

Taggotty (approved):
Tested and works in chrome 👍🏼

PR #524, opened 2021-11-30
IGAPP-635: Refactor headers and sharing

klinzo:

This refactors our logic such that not the header components are responsible to calculate the shareUrl in a big switch/if mess but the different routes (and its containers).
The same thing will happen with the language change paths. That will allow us language change in non redux routes like news, offers, disclaimer, ...


native/src/constants/NavigationTypes.ts
```diff
@@ -70,42 +70,47 @@ export type RoutesType =
 type ShareUrlType = {
   shareUrl: string
 }
+
 type CityContentParamsType = {
   cityCode: string
   languageCode: string
 }
+type BasicParams = {
+  shareUrl?: string
```
klinzo, line 79: `languageChangePaths` should be added here as well.

native/src/components/News.tsx
```diff
@@ -81,7 +81,7 @@ const News = (props: PropsType): ReactElement => {
   }
 
   if (loading) {
-    return <LoadingSpinner />
+    return <LoadingSpinner testID='loadingSpinner' />
```
Taggotty, line 84: maybe we can set the `testID directly in loading spinner, as this could be useful for other tests

native/src/components/__tests__/News.spec.tsx
```diff
@@ -0,0 +1,156 @@
...
+      longitude: null,
+      aliases: null,
+      boundingBox: null
+    })
+    const props = { cityModel, language, selectNews, loadMore, refresh, selectedNewsType }
+    return render(
+      <ThemeProvider theme={buildConfig().lightTheme}>
+        <News {...props} loading={loading} data={data} newsId={newsId} error={error} loadingMore={loadingMore} />
+      </ThemeProvider>
+    )
```
Taggotty, line 93:
```suggestion
    return render(
        <News {...props} loading={loading} data={data} newsId={newsId} error={error} loadingMore={loadingMore} />, 
        { wrapper: wrapWithTheme }
    )
```

native/src/components/__tests__/News.spec.tsx
```diff
@@ -0,0 +1,156 @@
...
+  TuNewsType
+} from 'api-client'
+
+import buildConfig from '../../constants/buildConfig'
+import News from '../News'
+
+jest.mock('react-i18next')
+jest.mock('../../components/NativeHtml', () => ({ content }: { content: string }) => <Text>{content}</Text>)
+
+const news = [
```
Taggotty, line 23: This way you don't have to use the `!` below
```suggestion
const news: [TunewsModel, TunewsModel]
```

native/src/routes/__tests__/LocalNews.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+      latitude: null,
+      longitude: null,
+      aliases: null,
+      boundingBox: null
+    })
+    const props = { cityModel, language: 'de', selectNews }
+    return render(
+      <ThemeProvider theme={buildConfig().lightTheme}>
+        <LocalNews {...props} newsId={newsId} />
+      </ThemeProvider>
```
Taggotty, line 62:
```suggestion
    return render(
        <LocalNews {...props} newsId={newsId} />,
        { wrapper: wrapWithTheme }
```

native/src/routes/__tests__/NewsContainer.spec.tsx
```diff
@@ -0,0 +1,199 @@
...
+      changeUnavailableLanguage,
+      selectNews,
+      language,
+      newsId
+    }: {
+      selectNews: (newsId: string | null) => void
+      changeUnavailableLanguage: (newLanguage: string) => void
+      language: string
+      newsId: string | null | undefined
+    }) =>
```
Taggotty, line 34:
```suggestion
    ({
      changeUnavailableLanguage,
      selectNews,
      language,
      newsId
    }: ComponentProps<typeof TuNews>) =>
```

Taggotty (approved):

native/src/components/News.tsx
klinzo: Thought about that, yes. However, the loading spinner is currently just a simple styled component and we'd need to add a wrapper. I'd figured that that is not worth it. Open to discuss that though.

native/src/components/__tests__/News.spec.tsx
klinzo: Thanks :+1:

native/src/components/News.tsx
Taggotty: Hmm. If that's the case then it's probably okay. 👍🏼

SteffiStuffel (approved):
When trying to start this i get
"Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render."

Therefore did not /could not test it.

native/src/Navigator.tsx
```diff
@@ -55,11 +55,16 @@ import SprungbrettOfferContainer from './routes/SprungbrettOfferContainer'
 import appSettings from './utils/AppSettings'
 import { initSentry, log } from './utils/sentry'
 
-const transparentHeader = (headerProps: StackHeaderProps) => <TransparentHeader {...headerProps} />
+type HeaderProps = {
+  route: RoutePropType<RoutesType>
+  navigation: NavigationPropType<RoutesType>
+}
+
+const transparentHeader = (headerProps: StackHeaderProps) => <TransparentHeader {...(headerProps as HeaderProps)} />
```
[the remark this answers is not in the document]
klinzo: Of what kind? Not the case for me.
The problem is not whether we pass all props or only some, but the type of `navigation` and `route`.

SteffiStuffel (approved):
Tested sharing on different sites, content language and system language change. Works as expected

PR #526, opened 2021-12-06
IGAPP-755: Show NotFoundError for invalid Deep Links

Taggotty:


api-client/src/errors/NotFoundError.ts
```diff
@@ -1,6 +1,6 @@
 import { LocalNewsType, TuNewsType } from '../routes'
 
-type NotFoundType = 'category' | 'event' | 'poi' | 'offer' | 'disclaimer' | TuNewsType | LocalNewsType | 'route'
+type NotFoundType = 'city' | 'category' | 'event' | 'poi' | 'offer' | 'disclaimer' | TuNewsType | LocalNewsType | 'route'
```
SteffiStuffel, line 3: Why don't we have 'language' here also?

SteffiStuffel (commented):
Tested this for pois and got stuck in the loading screen, when trying to open a invalid poi, see attached video

https://user-images.githubusercontent.com/18463686/144980391-2be990aa-fa25-4fc9-bd7e-537d1a2b73d7.mp4

i.

api-client/src/errors/NotFoundError.ts
Taggotty: I guess because we treat `LanguageNotAvailable` differently.

klinzo:
Tested on Android, works fine for me (also POIs). I get the same error all the time (requested page could not be found), not like in web configured for the different routes. But I think that is fine, no need to make it complicated. Nice work!

klinzo (approved):

klinzo:
If you want you could add a release note like `Opening invalid links is not handled correctly` or similar

Taggotty:
> Tested on Android, works fine for me (also POIs). I get the same error all the time (requested page could not be found), not like in web configured for the different routes. But I think that is fine, no need to make it complicated. Nice work!

This is because of the Failure component, as it currently only distinguishes the error on base of the Error Code. I guess it's a design and translation decision what should be done.

SteffiStuffel (approved):
Tested again. Works fine also for invalid pois.

PR #527, opened 2021-12-08
IGAPP-852: Prettify E2E tests in CI

Taggotty:


e2e-tests/web/capabilities.ts
```diff
@@ -36,16 +38,11 @@ export const browserstackCapabilities = {
     browserName: 'Safari',
     browserVersion: '14.0'
   })
-}
+} as Record<typeof browsers[number], Capabilities>
```
klinzo, line 41: If you remove `edge` from `browsers` you can just add a type annotation instead of having to cast it. But probably it is intended, right?

e2e-tests/web/wdio-browserstack.conf.ts
```diff
@@ -0,0 +1,18 @@
+import { browsers, browserstackCapabilities } from './capabilities'
+import { config as defaultConfig } from './wdio.conf'
+
+const capabilities = browsers
+  .filter(browser => process.argv.includes(`--${browser}`))
```
klinzo, line 5: Shouldn't this rather be a find? Or is it possible that e2e tests are run in multiple browsers?

klinzo (approved):
Looks pretty good!

e2e-tests/web/wdio.conf.ts
```diff
@@ -1,23 +1,34 @@
...
+  const parsedCapabilies = browsers
+    .filter(browser => process.argv.includes(`--${browser}`))
+    .map(browser => ({ browserName: browser }))
+  return parsedCapabilies.length > 0 ? parsedCapabilies : [{ browserName: 'chrome' }]
+}
 
 export const config: Testrunner = {
   runner: 'local',
-  specs: ['./web/test/specs/**/*.ts'],
+  specs: ['./web/test/specs/**/navigateToSearchResults.e2e.ts'],
```
klinzo, line 18: Is this by change deliberate?

Taggotty: nope. will remove.

e2e-tests/web/wdio-browserstack.conf.ts
Taggotty: it is possible to run the e2e tests in multiple browsers. check out `yarn test:web --chrome --firefox`

SteffiStuffel (approved):

PR #530, opened 2021-12-08
IGAPP-806: Add get coordinates to LocationModel

Cmd190:
See https://issues.tuerantuer.org/browse/IGAPP-806


api-client/src/models/__tests__/LocationModel.spec.ts
```diff
@@ -1,126 +1,173 @@
...
-          name: 'test',
-          country: null,
-          region: null,
-          state: null,
-          address: null,
-          town: null,
-          postcode: null
-        }).location
-      ).toBe('test')
+    describe('location', () => {
```
Cmd190, line 5: Is this maybe a formatting Issue?

klinzo: Seems like it, yes. Running prettier should fix it.

api-client/src/models/LocationModel.ts
```diff
@@ -113,6 +114,13 @@ class LocationModel {
     )
   }
 
+  get coordinates(): Position | null {
```
klinzo, line 117: You should also remove the other definition of this method and use this new method :)

klinzo (changes requested):

api-client/src/models/LocationModel.ts
```diff
@@ -113,6 +114,13 @@ class LocationModel {
...
+    if (this.longitude == null || this.latitude == null) {
```
klinzo, line 118: Please also check for undefined and use strict equal checks (`===`)

Cmd190: done

api-client/src/models/LocationModel.ts
```diff
@@ -113,8 +115,20 @@ class LocationModel {
...
+  get coordinates(): Position | null {
+    if (
+      this.longitude === null ||
+      this.longitude === undefined ||
+      this.latitude === null ||
+      this.latitude === undefined
+    ) {
+      return null
+    }
+    return [Number(this.longitude), Number(this.latitude)]
```
klinzo, line 127:
```suggestion
    if (!this.longitude || !this.latitude) {
      return null
    }
    return [Number(this.longitude), Number(this.latitude)]
```

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,23 @@
-import type { Position } from 'geojson'
 import { Platform } from 'react-native'
 
+import { LocationModel } from 'api-client/src'
+
 /**
  * Depending on the platform this function returns the correct deeplink for navigation app
  * @param address{string}: location address
- * @param coordinates{number[]}: latitude,longitude
+ * @param location{number[]}: latitude,longitude
```
klinzo, line 8: This type is not correct.
I'd move all the checks here and just pass the whole `LocationModel` to `getNavigationDeepLinks`

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,23 @@
...
  */
-export const getNavigationDeepLinks = (address: string, coordinates: Position): string =>
-  Platform.select({
-    ios: `maps:${coordinates[1]},${coordinates[0]}?q=${address}`,
-    android: `geo:${coordinates[1]},${coordinates[0]}?q=${address}`
+export const getNavigationDeepLinks = (address: string, location: LocationModel): string | undefined => {
+  if (location.coordinates == null) {
+    return undefined
+  }
```
klinzo, line 13:
```suggestion
export const getNavigationDeepLink = (location: LocationModel): string | null => {
  if (location.location && location.coordinates) {
    return null
  }
```

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,23 @@
...
+
+  const x = location.coordinates[0]
+  const y = location.coordinates[1]
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
klinzo, line 18: Since this is just a warning you should not need the comment.

api-client/src/models/__tests__/LocationModel.spec.ts
```diff
@@ -124,3 +124,50 @@ describe('convertToPoint', () => {
     expect(location.convertToPoint(path, thumbnail, urlSlug)).toBeNull()
   })
 })
+describe('get coordinates', () => {
```
Taggotty, line 127: you should put those tests in the location model scope.

native/src/utils/getNavigationDeepLinks.ts
Taggotty:
```suggestion
export const getNavigationDeepLinks = (address: string, location: LocationModel): string | undefined => {
  if (!location.location && !location.coordinates) {
    return null
  }
```
Or am I missing something ^^

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,23 @@
...
+  return Platform.select({
+    ios: `maps:${y},${x}?q=${address}`,
+    android: `geo:${y},${x}?q=${address}`
   })!
```
Taggotty, line 22: You can also replace android with default. Thus, you don't have to add the non-null-assertion. Not sure, whether thats actually better style though.

Taggotty (approved):

native/src/utils/getNavigationDeepLinks.ts
klinzo: Nearly, it should be an OR :grimacing:

native/src/utils/getNavigationDeepLinks.ts
klinzo: Not sure, but I'd stick with explicit android here I guess

api-client/src/models/__tests__/LocationModel.spec.ts
Cmd190: Also the test under  describe('location', ?

native/src/utils/getNavigationDeepLinks.ts
Cmd190: @Taggotty @klinzo so which one should we use?

api-client/src/models/LocationModel.ts
```diff
@@ -113,8 +114,15 @@ class LocationModel {
...
+  get coordinates(): Position | null {
+    if (!this.longitude || !this.latitude) {
+      return null
+    }
+    return [Number(this.longitude), Number(this.latitude)]
+  }
+
   convertToPoint(path: string, thumbnail: string, urlSlug: string): PoiFeature | null {
-    if (this.longitude == null || this.latitude == null) {
+    if (this.coordinates == null) {
```
klinzo, line 125: Please use your coordinates getter below as well (line 133)

api-client/src/models/__tests__/LocationModel.spec.ts
```diff
@@ -1,126 +1,168 @@
...
-          name: 'test',
-          country: null,
-          region: null,
-          state: null,
-          address: null,
-          town: null,
-          postcode: null
-        }).location
-      ).toBe('test')
+        expect(
```
klinzo, line 25: Please revert, wrong indentation

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,22 @@
...
  * Depending on the platform this function returns the correct deeplink for navigation app
  * @param address{string}: location address
- * @param coordinates{number[]}: latitude,longitude
+ * @param location{LocationModel}: location model with coordinates
  */
-export const getNavigationDeepLinks = (address: string, coordinates: Position): string =>
-  Platform.select({
-    ios: `maps:${coordinates[1]},${coordinates[0]}?q=${address}`,
-    android: `geo:${coordinates[1]},${coordinates[0]}?q=${address}`
+export const getNavigationDeepLinks = (address: string, location: LocationModel): string | undefined => {
```
klinzo, line 10: Please remove param `address`, it is the same as `location.location`. We also don't use js doc so you can remove the comment above completely

native/src/utils/getNavigationDeepLinks.ts
```diff
@@ -1,13 +1,22 @@
...
+  if (!location.location && !location.coordinates) {
```
klinzo, line 11: Should be 
```suggestion
  if (!location.location || !location.coordinates) {
```

native/src/components/PoiDetails.tsx
```diff
@@ -75,11 +76,12 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
   const { location, address, postcode, town } = poi.location
   const { distance, title } = feature.properties
 
-  // TODO refactor to use a get coordinates method IGAPP-806
   const onNavigate = () => {
-    if (location && poi.featureLocation?.geometry.coordinates) {
-      const navigationUrl = getNavigationDeepLinks(location, poi.featureLocation.geometry.coordinates)
-      Linking.openURL(navigationUrl).catch(() => showSnackbar(t('navigationApplication')))
+    if (location) {
```
klinzo, line 80: Remove this check

Cmd190: If I the check  `if (location)` I get this error: Argument of type 'string | null | undefined' is not assignable to parameter of type 'string'.

klinzo: yes, but you don't need to pass that, see comment above. you already pass the full model, no need to pass one property again

Cmd190: oh you're right

api-client/src/models/__tests__/LocationModel.spec.ts
Cmd190: The commit which changed the indention was reverted. I took a look again and ran prettier:fix. If the indention ist still wrong maybe you should try to fix it?

native/src/components/PoiPage.tsx
```diff
@@ -72,10 +72,7 @@ const PoiPage = ({
   const { location } = poi.location
   const files = resourceCache[poi.path] || {}
 
-  const navigationUrl =
-    location && poi.featureLocation?.geometry.coordinates
-      ? getNavigationDeepLinks(location, poi.featureLocation.geometry.coordinates)
-      : null
+  const navigationUrl = location && getNavigationDeepLinks(poi.location)
```
klinzo, line 75: No need to check for `location` here
```suggestion
  const navigationUrl = getNavigationDeepLinks(poi.location)
```

klinzo (commented):
Are there no usages of this in `web`?

native/src/components/PoiPage.tsx
Cmd190: I only found usages of LocationModel within the api-client

klinzo (approved):
Well done :)

PR #534, opened 2021-12-22
IGAPP-833: Upgrade react router v6

klinzo:


api-client/src/routes/pathname.ts
```diff
@@ -0,0 +1,70 @@
+import {
```
klinzo, line 1: Moved to api-client from native. Nearly identical to previous solution.

web/package.json
```diff
@@ -114,7 +114,8 @@
     "moment-locales-webpack-plugin": "^1.2.0",
     "moment-timezone-data-webpack-plugin": "1.5.0",
     "raf": "^3.4.1",
-    "react-hot-loader": "^4.12.19",
```
klinzo, line 117: react-hot-loader is deprecated and did not work together with react-router-dom as it wraps every component in `ProxyFacade`

klinzo (commented):
The react-router api changed quite a bit:
* No more path-to-regexp syntax
    -> no more optional params, workaround: `<Route path=':poiId' element={null} />` as child)
    -> constructing pathname from route patterns not possible anymore -> usage of our own `pathnameFromRouteInformation`
* React elements instead of components as route content -> no props passed, instead usage of hooks necessary, but also cleaner syntax
* Route patterns/paths are now relative to parent

klinzo:
Please make sure to test all routes and navigate a little bit through the app.

SteffiStuffel (commented):
In my opinion this PR is too big, making it hard to review.
The changes to move and rename `pathnameFromRouteInformation` to api-client, could have been done nicely in a separate PR.

It would be really nice if we can have a discussion on the optimal / maximal size of PRs.

Why haven't you updated path-to-regexp as mentioned in the ticket?

klinzo:
Will open a new PR as this is quite messy. @SteffiStuffel feel free to open a discussion about PR size. I think it is perfectly fine to just request that a PR should be split up if it gets to big like you did here, so imo no discussion/hard boundaries needed.

klinzo:
Will open a new PR as this is quite messy.

PR #537, opened 2021-12-30
IGAPP-837: Update typescript to 4.4

SteffiStuffel:


tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "noUncheckedIndexedAccess": true,
     "resolveJsonModule": true,
     "strict": true,
-    "target": "ESNext"
+    "target": "ESNext",
+    "useDefineForClassFields": false
```
klinzo, line 13: Why is this change necessary?

web/src/utils/retryImport.ts
```diff
@@ -19,7 +19,7 @@ const retry = async <T>(
 ): Promise<{ default: ComponentType<T> }> => {
   try {
     return componentImport()
-  } catch (error) {
+  } catch (error: any) {
```
klinzo, line 22: I'd favor using `unknown` as type annotation, `any` is generally not a recommended type annotation. Same for other usages. We also have a linting rule for this (which is only a warning right now). Maybe we should change that.

web/src/utils/retryImport.ts
```diff
@@ -19,7 +19,7 @@ const retry = async <T>(
...
     await wait(interval)
     if (retriesLeft === 0) {
       throw new Error(error)
```
klinzo, line 25:
```suggestion
      throw error
```

native/src/utils/sendTrackingSignal.ts
```diff
@@ -18,7 +18,7 @@ export const sendRequest = async (signal: SignalType): Promise<void> => {
       await createTrackingEndpoint().request(signal)
     }
   } catch (e) {
-    if (fromError(e) === ErrorCode.NetworkConnectionFailed) {
+    if (e instanceof Error && fromError(e) === ErrorCode.NetworkConnectionFailed) {
```
klinzo, line 21: I think with your change to `fromError` this additional check should not be needed, right?

api-client/src/errors/__tests__/NotFoundError.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('NotFoundError', () => {
         city: 'augsburg',
         language: 'language'
       })
-    } catch (error) {
+    } catch (error: any) {
```
klinzo, line 23: Please use `unknown` instead.

api-client/src/endpoints/hooks/useLoadFromEndpoint.ts
```diff
@@ -23,7 +23,7 @@ export const loadFromEndpoint = async <T>(
       setError(null)
     }
   } catch (e) {
-    setError(e)
+    setError(e instanceof Error ? e : null)
```
klinzo, line 26: I don't think we should ignore errors if it is not an instance of type Error. Maybe just use
```suggestion
    setError(e instanceof Error ? e : new Error())
```

tsconfig.json
SteffiStuffel: `In 2021 the class fields feature was added into the JavaScript specification with behavior which differed from how TypeScript had implemented it. In preparation for this, in TypeScript 3.7, a flag was added (useDefineForClassFields) to migrate to emitted JavaScript to match the JavaScript standard behavior.

Now that the feature is in JavaScript we are changing the default to true for ES2022 and above, including ESNext.`

See more https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier

web/src/utils/retryImport.ts
SteffiStuffel: I had to add `any` explicitly because typescript changed it standard behavior from `any` to `unknown` leading to errors, as properties are not available on unknown, etc.

klinzo (approved):
not tested

f1sh1918 (approved):

PR #539, opened 2022-01-03
IGAPP-833: Upgrade react router to v6

klinzo:


klinzo:
The react-router api changed quite a bit:
* No more path-to-regexp syntax
    -> no more optional params, workaround: `<Route path=':poiId' element={null} />` as child)
    -> constructing pathname from route patterns not possible anymore -> usage of our own `pathnameFromRouteInformation`
* React elements instead of components as route content -> no props passed, instead usage of hooks necessary, but also cleaner syntax
* Route patterns/paths are now relative to parent

klinzo:
@SteffiStuffel regarding
> Why haven't you updated path-to-regexp as mentioned in the ticket?

`path-to-regexp` is no longer a dependency of `react-router`, therefore it makes no sense to do something like that here. The problem is now `webpack-dev-server` and some `appium` dependency.

web/src/CityContentSwitcher.tsx
```diff
@@ -44,18 +44,23 @@ const DisclaimerPage = lazyWithRetry(() => import('./routes/DisclaimerPage'))
 
 type PropsType = {
   cities: CityModel[]
-} & RouteComponentProps<{ cityCode: string; languageCode: string }>
+  languageCode: string
```
SteffiStuffel, line 47: isn't cityCode now missing?

web/src/CityContentSwitcher.tsx
```diff
@@ -106,7 +111,7 @@ const CityContentSwitcher = ({ cities, match, location }: PropsType): ReactEleme
     )
   }
 
-  const cityRouteProps: CityRouteProps = { cities, languages, cityModel, languageModel }
+  const cityRouteProps = { cities, languages, cityModel, languageModel, pathname, cityCode, languageCode }
```
SteffiStuffel, line 114: I know you are using implicit types a lot, but I actually think it is not the best way.

web/src/routes/PoisPage.tsx
```diff
@@ -44,14 +43,12 @@ const moveViewToBBox = (bBox: BBox, defaultVp: MapViewViewport): MapViewViewport
   ])
 }
 
-type PropsType = CityRouteProps & RouteProps<typeof POIS_ROUTE>
```
SteffiStuffel, line 47: Is `RouteProps<typeof XX_ROUTE>` no longer working with the update or why have you decided to remove/change it?

web/src/routes/CategoriesPage.tsx
```diff
@@ -50,12 +48,9 @@ const getBreadcrumb = (category: CategoryModel, cityName: string) => {
...
 
-type PropsType = CityRouteProps & RouteProps<typeof CATEGORIES_ROUTE>
-
-const CategoriesPage = ({ cityModel, match, location, languages }: PropsType): ReactElement => {
-  const previousPathname = useRef<string | null | undefined>(null)
-  const { cityCode, languageCode, categoryId } = match.params
-  const pathname = normalizePath(location.pathname)
+const CategoriesPage = ({ cityModel, pathname, languages, cityCode, languageCode }: CityRouteProps): ReactElement => {
+  const previousPathname = useRef<string | null>(null)
+  const categoryId = useParams()['*']
```
Taggotty, line 53: Not really intuitive. What does the `*` stand for? The placeholder?

SteffiStuffel (approved):
Tested web on firefox for aschaffenburg and integreat.

When going to integreat -> events -> select an event, i get an endless loading spinner

Taggotty (approved):

Taggotty:
tested in chrome. Routing looks pretty neat now 👍🏼

web/src/CityContentSwitcher.tsx
klinzo: No, we just have to extract it manually using `useParams`.

web/src/routes/PoisPage.tsx
klinzo: Yes, no longer working as this was also a path-to-regexp utility. Also, routes are now relative so `cityCode` and `languageCode` are not even included in `POIS_ROUTE` and so on.

web/src/routes/CategoriesPage.tsx
klinzo: It stands for a wildcard, i.e. matching any number of any path segments. See: https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-route-path-patterns
This is necessary for allowing category paths of arbitrary length.

klinzo:
@f1sh1918 could you please/test review this as well as I am a little bit afraid that I broke something here.

web/src/CityContentSwitcher.tsx
klinzo: Why do you think that? :thinking:

klinzo:
> Tested web on firefox for aschaffenburg and integreat.
> 
> When going to integreat -> events -> select an event, i get an endless loading spinner

@SteffiStuffel should be fixed, seems like the check wasn't necessary after all. Could you check that it works fine now and there is no flickering please?.

web/src/routes/EventsPage.tsx
```diff
@@ -29,17 +28,13 @@ import { cmsApiBaseUrl } from '../constants/urls'
...
-import { RouteProps } from './index'
 
-type PropsType = CityRouteProps & RouteProps<typeof EVENTS_ROUTE>
-
-const EventsPage = ({ cityModel, match, location, languages }: PropsType): ReactElement => {
-  const { cityCode, languageCode, eventId } = match.params
-  const pathname = normalizePath(location.pathname)
-  const history = useHistory()
+const EventsPage = ({ cityModel, languages, pathname, languageCode, cityCode }: CityRouteProps): ReactElement => {
+  const { eventId } = useParams()
```
Taggotty, line 33: why  is there no need for a wildcard here?

Taggotty (approved):

Taggotty:
Now also tested events and redirects 👍🏼

web/src/routes/EventsPage.tsx
klinzo: Because eventId is just one path segment (e.g. `event-1234`) while categoryId may have multiple segments, e.g. `willkommen/wilkommen-in-augsburg/erste-schritte`

web/src/CityContentSwitcher.tsx
SteffiStuffel: And why do we handle languageCode differently? Couldn't this also just be extracted using `useParams`?

web/src/CityContentSwitcher.tsx
SteffiStuffel: For example, if adding the type explicitly here, when changing the code later it is clear at the first view, what properties cityRouteProps would need, also if now removing a needed property the error would not show up at this line, but whenever we are using cityRouteProps. I think if the type is not really trivial the pros of explicit typing outweight the cons. At the bottom of this article there is a nice pro/con list:
https://www.schibsted.pl/blog/typescript-type-inference/
But in the end this is something we should agree on a consitent way to do, and nothing specific for this PR.

SteffiStuffel:
> 

Tested again, works now.

f1sh1918 (approved):

web/src/CityContentSwitcher.tsx
klinzo: Because we already extracted it somewhere else and there is no need to extract it again. But yes, would be possible.

PR #542, opened 2022-01-10
IGAPP-462: Upgrade react-native-webview

klinzo:


klinzo:
Please test opening links in the webview thoroughly but I could not reproduce the issue we previously had:
> Upgrade react-native-webview to the latest version. We had to downgrade because of this bug: https://github.com/react-native-webview/react-native-webview/issues/1741
This leads to links being opened in the webview AND the inapp browser.

May have been fixed with [v11.0](https://github.com/react-native-webview/react-native-webview/releases/tag/v11.0.0):
> The primary way this new behavior changes existing React Native WebView implementations on Android is that links that open in new tabs/windows (such as <a target="_blank">) will now prompt to open in the system browser, rather than re-using the current WebView.

f1sh1918 (approved):
Tested on ios & android

Taggotty (approved):
Tested on Android Emulator.

PR #544, opened 2022-01-11
IGAPP-836: Link background getting dark on click

f1sh1918:
To check easily use this terminal command when ios emulator is open:

`npx uri-scheme open integreat://integreat.app/jpal/12345678 --ios`


klinzo (approved):
Tested on iOS simulator :+1: 
PR names should be the same as those of the corresponding issue :)

Znippet (approved):

PR #546, opened 2022-01-12
IGAPP-261: Remove firebase for floss builds

klinzo:


klinzo:
@maxammann could I have a review from you here? https://github.com/Exodus-Privacy/exodus-standalone says that firebase is not included in the project anymore after building with `yarn android:integreat-floss:production`.

build-configs/integreat-floss/index.ts
```diff
@@ -0,0 +1,23 @@
...
+    ...integreatPlatformBuildConfigs.common.featureFlags,
+    floss: true
+  }
+}
+const androidIntegreatFlossBuildConfig: AndroidBuildConfigType = {
+  ...integreatPlatformBuildConfigs.android,
+  ...commonIntegreatFlossBuildConfig,
+  googleServices: null,
+  // TODO Use correct applicationId
+  applicationId: 'tuerantuer.app.integreat'
```
klinzo, line 16: https://issues.tuerantuer.org/browse/IGAPP-174

Taggotty:
How should we test this issue? Just build the flossed build config? Anything else?

klinzo:
> How should we test this issue? Just build the flossed build config? Anything else?

Yes, I'd say run `yarn android:integreat-floss:production` once and check that the app does not crash and there is no push notification setting and it would be nice if you could check that push notifications are still working on a non floss build, i.e. `yarn android:integreat:production` and send a push notification (or test on iOS of course).

f1sh1918 (changes requested):
Can not build on ios, android works fine
image

klinzo:
> Can not build on ios, android works fine Remove app from emulator/device and reinstall did not help

Seems like I have messed up with the import, could you try again please? Should work now :)

f1sh1918 (approved):
builds now also on ios :) nice1

klinzo:
> builds now also on ios :) nice1

@f1sh1918 did you test whether push notifications are still received?

native/android/app/build.gradle
```diff
@@ -226,14 +226,19 @@ android {
...
         variant.getCompileConfiguration().exclude group: 'com.mapbox.mapboxsdk', module: 'mapbox-android-accounts'
         variant.getRuntimeConfiguration().exclude group: 'com.mapbox.mapboxsdk', module: 'mapbox-android-accounts'
     }
 
+    // Exclude proprietary firebase for floss builds
+    if (buildConfig.featureFlags.floss) {
+        configurations {
+            all*.exclude group: 'com.google.firebase'
+        }
+    }
```
maxammann, line 240: Does this include runtime and compile configurations? Did you check after building whether the class fields of firebase are actually excluded?

maxammann: We probably also need to exclude `com.google.android.gms`. It brings for example `play-services-cloud-messaging`.

If we exclude whole GMS right away we possible break our location code. I'm not sure how to proceed here.

native/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "baseUrl": ".",
     "isolatedModules": true,
     "jsx": "react",
+    "module": "es2020",
```
maxammann, line 9: Why this change? For the dynamic imports?

If yes, then are we actually sure that React-Native/Hermes supports this properly? I guess we are because it works :D

maxammann (changes requested):
I'll request changes because we should discuss how we handle this. There are two possibilities:

* Add firebase and then exclude certain dependencies (Current solution)
* Avoido adding firebase dependency in the first place. Could be difficult because of RN autolinking.

native/tsconfig.json
klinzo: Yes, because of dynamic imports.

Hmm yes, I think so, everything should work :D Did not look specifically whether it is supported by react native/hermes though

maxammann:
Maybe this would be a cleaner alternative for React Native: https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library

In general all native dependencies do not make any RN app crash if you do not import anything from it. This is different to Flutter for example.
If we just say "do not autolink" then we basically say that we exclude this dependency (at least any proprietary parts of it).

native/android/app/build.gradle
klinzo: It should, yes. I checked with https://github.com/Exodus-Privacy/exodus-standalone and it says firebase is no longer included (also no google play services like `com.google.android.gms` or similar). How would I check this manually? But I guess we don't need to exclude anything else (at least for firebase).

klinzo:
> I'll request changes because we should discuss how we handle this. There are two possibilities:
> 
>     * Add firebase and then exclude certain dependencies (Current solution)
> 
>     * Avoido adding firebase dependency in the first place. Could be difficult because of RN autolinking.

I just followed your PR for the ehrenamtskarte. Would there be any advantages in not adding it in the first place? Not sure how that would be possible.

native/android/app/build.gradle
maxammann: That should be fine then. Did you check a release or debug build?
I used the decompiler of intellij previously. If you see something in kursiv, then it is not actually included.

maxammann (approved):

native/android/app/build.gradle
klinzo: I checked the release build. Ok, will try, thx.

klinzo:
> Maybe this would be a cleaner alternative for React Native: https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library
> 
> In general all native dependencies do not make any RN app crash if you do not import anything from it. This is different to Flutter for example. If we just say "do not autolink" then we basically say that we exclude this dependency (at least any proprietary parts of it).

Good idea and better solution! Done, thx :)

native/react-native.config.js
```diff
@@ -0,0 +1,18 @@
+// Exclude libraries from our builds by disabling autolinking
+// https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library
+const floss = JSON.parse(process.env.BUILD_CONFIG_FEATURE_FLAGS_FLOSS ?? 'false')
```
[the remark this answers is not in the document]
klinzo: Yes, correct, written in ts.

native/run
```diff
@@ -0,0 +1,28 @@
...
+  # Prepare common build config for packager
+  build_config=$(yarn workspace --silent build-configs --silent manage to-bash $build_config_name common)
+  # Start packager
+  cross-env $build_config react-native start --reset-cache
+else
+  # Prepare android build config
+  build_config=$(yarn workspace --silent build-configs --silent manage to-bash $build_config_name android)
+
+  # Extract application id from build config
+  [[ $build_config =~ .*BUILD_CONFIG_APPLICATION_ID=\"([^\"]+) ]]
```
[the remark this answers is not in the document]
klinzo: done

native/run
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+build_config_name=$1
+mode=$2
+
+if [ "$mode" == "packager" ]; then
+  # Prepare common build config for packager
+  build_config=$(yarn workspace --silent build-configs --silent manage to-bash $build_config_name common)
+  # Start packager
+  cross-env $build_config react-native start --reset-cache
```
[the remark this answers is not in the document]
klinzo: `cross-env` does not work that way. It just allows `key=value` pairs as input. The value is always wrapped in quotes.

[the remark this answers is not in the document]
klinzo: https://github.com/kentcdodds/cross-env#usage

PR #551, opened 2022-01-17
IGAPP-879: Remove logging statements

klinzo:


klinzo:
Testing:
* sending to sentry: (`yarn build`), running the app (`cd dist/integreat` and `python3 -m http.server`) and throwing an test error somewhere in the used files/routes. Then you can verify the error was actually sent to sentry
* logging for developer friendly builds: `yarn start` and look at console

web/src/utils/sentry.ts
```diff
@@ -1,52 +1,84 @@
...
+import { FetchError, NotFoundError } from 'api-client'
+
 import buildConfig from '../constants/buildConfig'
 
-const loadSentry = async (): Promise<typeof Sentry> =>
-  import(
-    /* webpackChunkName: "sentry" */
-    '@sentry/react'
-  )
+const sentryEnabled = (): boolean => buildConfig().featureFlags.sentry
```
SteffiStuffel, line 8: Why not just `const sentryEndable = buildConfig().featureFlags.sentry`?

web/tools/webpack.config.ts
```diff
@@ -87,7 +87,7 @@ const createConfig = (
   // eslint-disable-next-line no-console
   console.log('Used config: ', buildConfigName)
   // eslint-disable-next-line no-console
-  console.log('Configured as running in dev server: ', !devServer)
+  console.log('Configured as running in dev server: ', devServer)
```
SteffiStuffel, line 90: Why this change?

SteffiStuffel (approved):
Could not make it to send a error to sentry.

I added `reportError(new Error('Sentry test of IGAPP-879'))` in the code
did `yarn build` + `cd dist/integreat` + `python3 -m http.server`

web/tools/webpack.config.ts
klinzo: Because it was wrong before, it said `Configured as running in dev server: true` for production builds.

web/src/utils/sentry.ts
klinzo: Was that way before, not sure why it was done that way. However, it makes mocking the build config way easier in tests since you can change values in between test. If this is not a function, it keeps the first value (at least that happened to me on native lately).

web/src/routes/PoisPage.tsx
```diff
@@ -49,8 +49,6 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
   const formatter = useContext(DateFormatterContext)
   const { viewportSmall } = useWindowDimensions()
   const navigate = useNavigate()
-  // eslint-disable-next-line no-console
-  console.log('To use geolocation in a development build you have to start the dev server with\n "yarn start --https"')
```
f1sh1918, line 53: why did you rm this hint?

f1sh1918 (approved):

web/src/routes/PoisPage.tsx
klinzo: Will readd it if the build is developer friendly

PR #552, opened 2022-01-17
IGAPP-853: ChunkLoadError: Loading chunk xyz failed

klinzo:


klinzo:
You can test this by building the app (`yarn build`), running the app (`cd dist/integreat` and `python3 -m http.server`) and deleting one chunk from the corresponding dist folder.

Taggotty:
> 

Not sure whether I am testing this right. Deleting the chunk will result in 404, as the file does not exist anymore right? What's supposed to happen in production? Should the app refetch all recourses?

klinzo:
> Not sure whether I am testing this right. Deleting the chunk will result in 404, as the file does not exist anymore right? What's supposed to happen in production? Should the app refetch all recourses? Why do I still get an error in sentry? -> https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/449/?query=is%3Aunresolved

The request to load the deleted chunk results in a 404, yes. However, internally react lay just throws an error that the chunk could not be loaded. It is handled in the `lazyWithRetry` method by retrying the import and afterwards refreshing the page once. If that does not work, the error is thrown again (and sent to sentry), perhaps we could show an error in the UI instead? But not sure what to tell users there as there is not really anything else we can do to fix this.

Taggotty:
> 

What I mean is the following.
When I serve the server locally and delete the file there, then of course the chunk cannot be fetched, even when reloading, as the file still does not exist. Is that the reason why the error was reported to sentry?

And in production we would expect, that the reload will fetch all recources from the remote server (where they should exist),
which will then result in loading the page, right :) ?

klinzo:
> What I mean is the following. When I serve the server locally and delete the file there, then of course the chunk cannot be fetched, even when reloading, as the file still does not exist. Is that the reason why the error was reported to sentry?
> 
> And in production we would expect, that the reload will fetch all recources from the remote server (where they should exist), which will then result in loading the page, right :) ?
> 
> If it does not work, than that would mean, that our build is faulty or sth., as a chunk that is supposed to exist does not exist. I don't think we need to show an error in the UI then.

Good summary, yes. That is basically it!

It may be that there is an underlying issue with our app though which leads to the chunk loading error (even though the chunk is available). But we will see if we still receive errors in sentry after this PR.

SteffiStuffel (changes requested):
Not workting for me. Disabled my adblocker as this blocked sentry when I tried last time. Now requests to sentry are send (at least the init), but no error/logging is displayed there.
Sentry is enabled and Catch-block is reached, tested this by logging to the browser, but no error in sentry.

Also if i e.g. delete a chuck that is for tünews, i can use the app city-selection, categories, events and so on, but once I click to tünews, i got a 404 and navigating back or reloading or refresh with caches with `STRG+R` will not work, all pages beside `localhost:8000` wouldn't word. When going to `localhost:8000` redirecting to landing works fine, also navigating to categories and so on. So maybe this should be considered when fixing the error.

SteffiStuffel (commented):
deleted.

SteffiStuffel (approved):

klinzo:
> Not workting for me. Disabled my adblocker as this blocked sentry when I tried last time. Now requests to sentry are send (at least the init), but no error/logging is displayed there. Sentry is enabled and Catch-block is reached, tested this by logging to the browser, but no error in sentry.
> 
> Also if i e.g. delete a chuck that is for tünews, i can use the app city-selection, categories, events and so on, but once I click to tünews, i got a 404 and navigating back or reloading or refresh with caches with `STRG+R` will not work, all pages beside `localhost:8000` wouldn't work. When going to `localhost:8000` redirecting to landing works fine, also navigating to categories and so on. So maybe this should be considered when fixing the error.
> 
> 

Discussed in a call. Basically the problem is that the python http "server" we are using to test production builds only works for the entry point `localhost:8000`. Entering any other url or reloading results in a 404 since python server is not able to resolve the url. If you want to test that the error is still sent to sentry, you need to comment the reload logic.

f1sh1918 (approved):
tested. refresh works fine!

PR #553, opened 2022-01-17
IGAPP-881: Add eslint a11y styled components plugin

f1sh1918:
Since the plugin for eslint a11y styled component is more or less alpha, i just create this as a draft to discuss if it makes sense to implement it


.eslintrc.js
```diff
@@ -1,15 +1,16 @@
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'react', 'react-hooks', 'jest', 'prefer-arrow'],
+  plugins: ['@typescript-eslint', 'react', 'react-hooks', 'jest', 'prefer-arrow', 'jsx-a11y', 'styled-components-a11y'],
```
f1sh1918, line 4: gonna rm `jsx-a11y` reminder

.eslintrc.js
```diff
@@ -42,6 +43,16 @@ module.exports = {
     'react/jsx-filename-extension': 'off',
     'react/jsx-props-no-spreading': 'off',
     'react/require-default-props': 'off',
+    'styled-components-a11y/label-has-for': [
```
f1sh1918, line 46: force labels to have an id is important to connect them with input/textfields for screen reader

web/src/components/SearchInput.tsx
```diff
@@ -63,6 +63,7 @@ export class SearchInput extends React.PureComponent<PropsType> {
       <Spacer space={spaceSearch}>
         <Wrapper>
           <SearchIcon />
+          {/* eslint-disable-next-line styled-components-a11y/no-autofocus */}
```
f1sh1918, line 66: a11y doesn't recommend auto-focus at all ;)

web/src/components/Feedback.tsx
```diff
@@ -69,18 +74,20 @@ const Feedback = (props: PropsType): ReactElement => {
   return (
     <Container>
       <TextContainer>
-        <Description>{t(description)}</Description>
+        <Description htmlFor='comment'>{t(description)}</Description>
```
Znippet, line 77: 👍 Tested with and without this, agreed, we need that

web/src/components/MapPopup.tsx
```diff
@@ -56,7 +56,7 @@ const MapPopup: React.FC<MapPopupProps> = ({ properties }: MapPopupProps): React
   return (
     <CleanLink to={path}>
       <Popup>
-        {thumbnail !== 'null' ? <PopupThumbnail src={thumbnail} /> : <Placeholder />}
+        {thumbnail !== 'null' ? <PopupThumbnail src={thumbnail} alt='' /> : <Placeholder />}
```
Znippet, line 59: What is the sense of using the empty alt?

web/src/components/Selector.tsx
```diff
@@ -92,9 +92,13 @@ const Selector = ({
           <Element
             data-testid={item.code}
             key={item.code}
+            role='option'
             as={Link}
             to={item.href}
             onClick={closeDropDown}
+            onKeyPress={closeDropDown}
+            aria-selected={item.code === activeItemCode}
+            tabIndex={0}
```
Znippet, line 101: What happens if this tabIndes is not set? I would think it would work in the same way with the "natural" order

web/src/components/MapPopup.tsx
f1sh1918: In generell you should set alt tags for describing pictures. 
You can leave `alt=''` if the picture has only visual purpose 
Maybe in this case we should add a todo & disabled linting because we don't get an alt tag.

web/src/components/Selector.tsx
f1sh1918: i mean maybe it could work depending on browser and the convention is made for older browsers to ensure tab/focus works
But by default `<span>` `<div>` seem not to be focusable/tabable, so the rule forces you in this case for a span which was made "interactive" to set a tabIndex
https://bitsofco.de/how-and-when-to-use-the-tabindex-attribute/

web/src/components/SearchInput.tsx
klinzo: Perhaps you could add a comment why this is fine in our case?

web/src/components/MapPopup.tsx
klinzo: I think no alt tag is fine here since it not really has a purpose (the purpose is just adding a visual for the poi).

f1sh1918: yes but it's best practise to use empty alt tag

.eslintrc.js
klinzo: Please move this to the right place, these are all rules that are set to `off`. Is this rule disabled per default?

web/src/components/Feedback.tsx
```diff
@@ -35,7 +35,12 @@ const TextInput = styled.input.attrs({
   resize: none;
 `
 
-export const Description = styled.div`
+export const Description = styled.label`
+  padding: 10px 0 5px;
+  font-weight: 700;
+`
+
+export const ErrorSendingStatus = styled.div`
```
klinzo, line 43: You could just use the `as` prop for this: https://styled-components.com/docs/basics#extending-styles

web/src/components/Heading.tsx
```diff
@@ -20,7 +20,8 @@ const Heading = (): ReactElement => {
       throw new Error('This error was thrown for testing purposes.')
     }
   }
-
+  // disable eslint since it's a hidden fkt
```
klinzo, line 23:
```suggestion
  // Disable a11y linting since it is a hidden dev tool
```

web/src/components/Selector.tsx
klinzo: Nice read, didn't know about that :+1:

klinzo (approved):
Not tested

web/src/components/Feedback.tsx
f1sh1918: @klinzo thx for the hint. does this make any difference in this case?

klinzo:
This just handles web right? Is there something similar for native?

web/src/components/MapPopup.tsx
klinzo: Sorry, I meant an empty string as alt tag is fine here, you are right oc :)

f1sh1918:
> This just handles web right? Is there something similar for native?
In general yes, but i think not for styled components. At least i didn't find anything yet
https://www.npmjs.com/package/eslint-plugin-react-native-a11y

PR #554, opened 2022-01-18
IGAPP-796: Improve test performance

Taggotty:

for me the jest update saves up to 20 seconds on native and web. It's still incredibly slow, taking around 23s to startup and even start running the first test. Partially it is because of the typescript transformation (which was apparently even disabled in native). I'm still working on further improvements on another branch. But split it as it would become quite big.


SteffiStuffel (approved):

native/jest.setup.ts
```diff
@@ -2,7 +2,6 @@ import mockAsyncStorage from '@react-native-async-storage/async-storage/jest/asy
 import '@testing-library/jest-native/extend-expect'
 import fs from 'fs'
 import path from 'path'
-
```
klinzo, line 5: Revert, should be enforced by prettier anyway

native/src/routes/__tests__/Landing.spec.tsx
```diff
@@ -12,10 +11,12 @@ import wrapWithTheme from '../../testing/wrapWithTheme'
...
 import Landing from '../Landing'
 
+const mockTheme = buildConfig().lightTheme
+
 jest.mock('react-i18next')
 jest.mock('styled-components', () => ({
   ...jest.requireActual('styled-components'),
-  useTheme: () => buildConfig().lightTheme
+  useTheme: () => mockTheme
 }))
```
klinzo, line 20: Perhaps we could create a static mock for `styled-components`, lots of duplicated code. Oc no need to do this here.

klinzo (approved):
Nice work :)

f1sh1918 (approved):

native/src/routes/__tests__/Landing.spec.tsx
klinzo: https://issues.tuerantuer.org/browse/IGAPP-896

web/package.json
```diff
@@ -106,9 +105,8 @@
     "html-loader": "^3.0.1",
     "html-webpack-plugin": "^5.5.0",
     "img-loader": "^4.0.0",
-    "jest": "^27.4.7",
     "jest-fetch-mock": "^3.0.3",
-    "jest-mock":"^27.4.6",
+    "jest-mock": "^27.4.6",
```
klinzo, line 109: Could be moved to root package.json as well but fine as it currently is

PR #555, opened 2022-01-18
IGAPP-867: Anounce Nothing Found also in city search

Znippet:

Refactorings: 
- Moved "nothingFound" result from feedback component to CitySelector itself
- Reuse Failure component also for empty results 

Changes in the Failure component: 
   - When a page is not found the error message is now below the smiley, same as in search results
   - the smiley size is now also the same as in search results
   
image


SteffiStuffel:
Not sure if this is good design to have both elements, also i assume on small devices the suggest-city-footer will not be displayed without scrolling.

Znippet:
> Not sure if this is good design to have both elements, also i assume on small devices the suggest-city-footer will not be displayed without scrolling.

I think it looks good, the grey background separates the city-suggest part. Also the statement "nothing found." and the question "your region was not found? -> suggest it"  have different meanings. 

The scrolling is also necessary when a search result inside the city fails, the comments field is shown below it. We could add a media-css for smaller screens automatically hiding the Smiley.

klinzo:
> Not sure if this is good design to have both elements, also i assume on small devices the suggest-city-footer will not be displayed without scrolling.

Agreeing with @Znippet here. I also discussed this with @nikolahoff and she agrees that there should be a message `no results` if no results are found

translations/translations.json
```diff
@@ -4555,119 +4526,148 @@
   "search": {
     "de": {
       "searchPlaceholder": "Suche",
-      "pageTitle": "Suche"
+      "pageTitle": "Suche",
+      "nothingFound": "Es wurden leider keine passenden Ergebnisse gefunden."
```
klinzo, line 4530: I think this should go to the error namespace if it's only used in the failure component

web/src/components/CitySelector.tsx
```diff
@@ -52,19 +56,28 @@ const CitySelector = ({ cities, language, filterText, stickyTop = 0 }: PropsType
...
-      },
-      []
+    const failureProps = {
+      goToPath: null,
+      errorMessage: 'nothingFound'
+    }
+
+    return sorted.length === 0 ? (
+      <Failure {...failureProps} t={t} />
+    ) : (
```
klinzo, line 66: Optional: Imo it's more explicit and therefore better readable to use an regular if statement here

web/src/components/CitySelector.tsx
```diff
@@ -52,19 +56,28 @@ const CitySelector = ({ cities, language, filterText, stickyTop = 0 }: PropsType
...
-        )
-      },
-      []
+    const failureProps = {
+      goToPath: null,
+      errorMessage: 'nothingFound'
+    }
+
+    return sorted.length === 0 ? (
+      <Failure {...failureProps} t={t} />
```
klinzo, line 65: Spreading should be omitted if possible, makes the code less readable (especially if more code between `const failureProps =...` and this line is added. Also ts does not check for excess/unknown props if using the spread operator.

If you move the `nothingFound` translation to the error namespace, you also don't have to pass the `t` function

web/src/components/Failure.tsx
```diff
@@ -16,17 +16,18 @@ const Centered = styled.div`
 
 type PropsType = {
   errorMessage: string
-  goToPath?: string
+  goToPath?: string | null
```
klinzo, line 19: No need for `null` here, just don't pass a path as prop.

web/src/components/Failure.tsx
```diff
@@ -16,17 +16,18 @@ const Centered = styled.div`
...
   goToMessage?: string
   t: TFunction<'error'>
 }
 
+// screen reader test
```
klinzo, line 24: Should probably get removed?

web/src/components/CitySelector.tsx
```diff
@@ -52,19 +56,28 @@ const CitySelector = ({ cities, language, filterText, stickyTop = 0 }: PropsType
...
-            <CityListParent stickyTop={stickyTop}>{key}</CityListParent>
-            {cities.map(city => (
-              <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
-            ))}
-          </div>
-        )
-      },
-      []
+    const failureProps = {
+      goToPath: null,
```
klinzo, line 60: No need to pass `null` here, just don't pass anything at all.

web/src/components/Failure.tsx
```diff
@@ -16,17 +16,18 @@ const Centered = styled.div`
...
 const Failure = ({ errorMessage, goToPath = '/', goToMessage = 'goTo.start', t }: PropsType): ReactElement => (
   <Centered>
-    <div>{t(errorMessage)}</div>
     <div>
-      <FontAwesomeIcon icon={faFrown} size='5x' />
+      <FontAwesomeIcon icon={faFrown} size='4x' />
     </div>
+    <div role='alert'>{t(errorMessage)}</div>
```
klinzo, line 30: What is the reason for this design change? Did you discuss this with our designers?

web/src/components/FeedbackContainer.tsx
```diff
@@ -139,24 +139,16 @@ export const FeedbackContainer = (props: PropsType): ReactElement => {
...
-          onContactMailChanged={setContactMail}
-          onSubmit={handleSubmit}
-          sendingStatus={sendingStatus}
-          isPositiveFeedback={isPositiveFeedback}
-          isSearchFeedback={isSearchFeedback}
-          comment={comment}
-          contactMail={contactMail}
-        />
-      </>
+      <Feedback
```
klinzo, line 142: :+1:

web/src/routes/SearchPage.tsx
```diff
@@ -127,6 +128,11 @@ const SearchPage = ({ cityModel, languages, cityCode, languageCode, pathname }:
 
   const pageTitle = `${t('pageTitle')} - ${cityModel.name}`
 
+  const searchFailureProps = {
+    goToPath: null,
+    errorMessage: 'nothingFound'
+  }
```
klinzo, line 134: Same as for other usage of `Failure`

klinzo (changes requested):
Good find and nice refactoring :+1: Is there also an issue for native?

web/src/components/CitySelector.tsx
Znippet: This will result in "undefined", which does not work with the Failure Component, it will set the path to "/" :
Failure = ({ errorMessage, goToPath = '/', ....

klinzo: Ahh you're right. Then I'd remove the default value and set it explicitly wherever that was used.

web/src/components/Failure.tsx
Znippet: I reused the empty search result layout. It was mainly the same as the Failure layout.
We could do it the other way round also.
The only reason why I used the order of the empty search result was that having the text above the smiley looked strange (with url there is text above and below the smiley)

translations/translations.json
SteffiStuffel: Not really an error though. I think it's fine to have it here.

SteffiStuffel (approved):
Tested on Chrome big and small size, looks good :)

translations/translations.json
klinzo: If it's only used in the `Failure` component it should also be part of the namespace the failure component uses. Also, `search` is the namespace for the category search, not search in general. It also makes the code easier, so I definitely think it should be moved :)

Znippet: The translations in the error.notFound section translated by the FailureSwitcher component, which I don't use. So I can't leave out the explicit namespace without using the FailureSwitcher component.

I prefer using using only the namespace search in the search.
For the city search I agree that the "search" is the wrong content.
I used error - notFound.city now which is the existing message for a wrong spelled city in the url: Diese Stadt wurde leider nicht gefunden.

klinzo: You are absolutely right. However, I think `Es wurden leider keine passenden Ergebnisse gefunden` is the better text for the city search compared to `Diese Stadt wurde leider gefunden`. Reasons:
* We try to avoid the term `Region` as much as possible (`Stadt` should not be used in the future anymore at all in the UI)
* `Keine passenden Ergebnisse` is more general, the user may just have entered `as` which already gives no result but it is obviously not a city

So best just revert the last commit and use the `search` context again (or move the `const { t } = useTranslation('error')` into the `Failure` component and `nothingFound` to the error namespace).

klinzo (commented):
Tested on firefox :+1:  Could you change the translation back to `Keine passenden Ergebnisse gefunden`... like mentioned here https://github.com/digitalfabrik/integreat-app/pull/555#discussion_r792556752

native/src/routes/SearchModal.tsx
```diff
@@ -193,7 +193,7 @@ class SearchModal extends React.Component<PropsType, SearchStateType> {
         {filteredCategories.length === 0 && (
           <>
             <SadIconContainer source={SadIcon} />
-            <Heading theme={theme}>{t('search:nothingFound')}</Heading>
+            <Heading theme={theme}>{t('feedback:nothingFound')}</Heading>
```
klinzo, line 196: This should not be reverted I guess

klinzo (approved):

Znippet:
> Good find and nice refactoring 👍 Is there also an issue for native?

https://issues.tuerantuer.org/browse/IGAPP-398

> Good find and nice refactoring 👍 Is there also an issue for native?

Didn't find one, we should add one.
We thogh have a task for "no city found": https://issues.tuerantuer.org/browse/IGAPP-398

PR #561, opened 2022-01-25
IGAPP-868: Roles of ui elements not read by screen reader 

f1sh1918:
Some of the roles i already added within the a11y plugin tasks, because they were causing linting errors.
Here i adjusted some elements to semantic html elements to automatically set the inherit roles

Here is a nice overview if someone is interested in some background info
https://medium.com/swlh/semantic-html-for-better-accessibility-a54d0742033f


web/src/components/__tests__/__snapshots__/Header.spec.tsx.snap
```diff
@@ -22,25 +22,25 @@ exports[`Header should match snapshot with largeViewport 1`] = `
         link="/random_route"
```
klinzo, line 22: Could you perhaps remove the snapshot and write a new test for this?

f1sh1918: we can create tickets for this. as i also touched other components i would have to adjust more than just the header text which is in my opinion out of scope for this ticket

Taggotty (approved):
Tested in chrome. Nice job. But I guess there are probably a lot more places, where we could improve our accessibility.

web/src/components/__tests__/__snapshots__/Header.spec.tsx.snap
klinzo: I just meant the snapshot test, but of course this is never a must do, always a can do. So it is fine :)

klinzo (approved):
Not tested

PR #562, opened 2022-01-25
IGAPP-865: Alignment of rtl text for categories with icons 

f1sh1918:
Tested on ios


native/src/components/CategoryListItem.tsx
```diff
@@ -29,9 +30,9 @@ const DirectionContainer = styled.View<DirectionContainerPropsType>`
   display: flex;
   flex-direction: ${props => contentDirection(props.language)};
 `
-const CategoryEntryContainer = styled.View`
+const CategoryEntryContainer = styled(View)<{ language: string }>`
```
Taggotty, line 33: should be the same as
```suggestion
const CategoryEntryContainer = styled.View<{ language: string }>`
```

Taggotty (approved):
Tested on android emulator 👍🏼

klinzo (approved):
Tested on android. I could not reproduce the bug without your changes neither though.

PR #566, opened 2022-01-31
IGAPP-900: Adjustments for CityNotCooperating web and native

SteffiStuffel:

Moved the template and the icons to buildConfig, so it is not duplicated for native and web.


native/src/routes/CityNotCooperating.tsx
```diff
@@ -80,8 +79,10 @@ const TemplateText = styled.Text`
 
 const CityNotCooperating = (): ReactElement => {
   const { t } = useTranslation('cityNotCooperating')
+  const { cityNotCooperatingIcon } = buildConfigAssets()
   const [isCopied, setIsCopied] = useState<boolean>(false)
   const theme = useTheme()
+  const template: string = buildConfig().featureFlags.cityNotCooperatingTemplate!
 
   const copyToClipboard = () => {
     Clipboard.setString(template)
```
f1sh1918, line 88:
```suggestion
    if (template) {
      Clipboard.setString(template)
      setIsCopied(true)
    }
```
i think its safer to check here and avoid the not null assertion above

native/src/routes/CityNotCooperating.tsx
```diff
@@ -80,8 +79,10 @@ const TemplateText = styled.Text`
 
 const CityNotCooperating = (): ReactElement => {
   const { t } = useTranslation('cityNotCooperating')
+  const { cityNotCooperatingIcon } = buildConfigAssets()
   const [isCopied, setIsCopied] = useState<boolean>(false)
   const theme = useTheme()
+  const template: string = buildConfig().featureFlags.cityNotCooperatingTemplate!
```
f1sh1918, line 85:
```suggestion
  const template = buildConfig().featureFlags.cityNotCooperatingTemplate
```

web/src/routes/CityNotCooperatingPage.tsx
```diff
@@ -107,7 +108,8 @@ const CityNotCooperatingPage = ({ languageCode }: PropsType): ReactElement => {
         </HeadingContainer>
 
         <Text>{t('explanation')}</Text>
-        <Icon alt='' src={cityNotCooperationIcon} />
+        {}
```
f1sh1918, line 111: looks like `{}` added by mistake?

build-configs/integreat/index.ts
```diff
@@ -9,6 +9,13 @@ import mainImprint from './mainImprint'
 import { lightTheme } from './theme'
 
 const APPLE_APP_ID = '1072353915'
+const cityNotCooperatingTemplate =
```
klinzo, line 12: Could we move this to a separate file? I think it would also be nice if we did not have to duplicate it for integreat-test-cms or other build configs

native/src/routes/CityNotCooperating.tsx
klinzo: For this kind of assertion the route should only be enabled if the template is not null. I.e. use the following in the navigator:
```
        const cityNotCooperatingEnabled = template && icon
```
and then
```
        {cityNotCooperatingEnabled && <Stack.Screen name={CITY_NOT_COOPERATING_ROUTE} component={CityNotCooperating} />}
```
Then it is fine imo to use non null assertion all the time.

klinzo (commented):
The svg should not exist twice, instead symlinks should be used.

Also the route should be disabled if it is disabled in the build config (both web and native).

Please also address the last comment here: https://issues.tuerantuer.org/browse/PROMO-215

native/src/routes/CityNotCooperating.tsx
SteffiStuffel: only allowed the route when the template is set and there keep the not null assertion above.

SteffiStuffel:
> The svg should not exist twice, instead symlinks should be used. :heavy_check_mark: Done
> 
> Also the route should be disabled if it is disabled in the build config (both web and native). :heavy_check_mark: Done
> 
> Please also address the last comment here: https://issues.tuerantuer.org/browse/PROMO-215 :heavy_check_mark: Done

build-configs/integreat-test-cms/assets/city-not-cooperating.svg
```diff
@@ -0,0 +1 @@
+/home/steffi/Dokumente/integreat-app/build-configs/integreat/assets/city-not-cooperating.svg
```
Taggotty, line 1: should probably be relative

native/src/Navigator.tsx
```diff
@@ -93,6 +93,8 @@ const Navigator = (props: PropsType): ReactElement | null => {
   })
   const previousRouteKey = useRef<string | null | undefined>(null)
   const { fetchCities, fetchCategory, routeKey, routeName } = props
+  const cityCooperatingEnabled =
+    buildConfig().featureFlags.cityNotCooperatingTemplate && buildConfigAssets().cityNotCooperatingIcon
```
Taggotty, line 97: maybe it would make sense to extract this one into a function in e.g. the cityCooperatingFooter and to consistently always check on both properties. Sometimes you only check for the template and sometimes only for the icon, which could be error prune imo and also feels very complicated.

SteffiStuffel: you are right, this was not really consistent. I adjusted it. CityCooperatingPages are only available when template and icon are set correctly, the Footer as a reusable component checks itself if icon is available.

Taggotty (approved):

build-configs/integreat-e2e/index.ts
```diff
@@ -5,14 +5,9 @@ import {
   WebBuildConfigType
 } from '../BuildConfigType'
 import integreatPlatformBuildConfigs from '../integreat'
+import { template } from '../integreat/assets/cityNotCooperatingTemplate'
```
klinzo, line 8: Change export to default and
```suggestion
import cityNotCooperatingTemplate from '../integreat/assets/cityNotCooperatingTemplate'
```

build-configs/integreat-test-cms/index.ts
```diff
@@ -5,14 +5,9 @@ import {
   WebBuildConfigType
 } from '../BuildConfigType'
 import integreatPlatformBuildConfigs from '../integreat'
+import { template } from '../integreat/assets/cityNotCooperatingTemplate'
```
klinzo, line 8: Not sure if we shouldn't use a symlink here as well. Or you can just do the following below:
```
    featureFlags: {
      ...,
     cityNotCooperatingTemplate: integreatPlatformBuildConfigs.common.featureFlags.cityNotCooperatingTemplate
    }
```

native/src/Navigator.tsx
```diff
@@ -85,6 +85,9 @@ type InitialRouteType =
     }
 const Stack = createStackNavigator<RoutesParamsType>()
 
+export const cityNotCooperatingEnabled =
```
klinzo, line 88: I think this should be moved to `buildConfig.ts`

web/src/RootSwitcher.tsx
```diff
@@ -34,6 +34,9 @@ const MainDisclaimerPage = lazyWithRetry(() => import('./routes/MainDisclaimerPa
 const LandingPage = lazyWithRetry(() => import('./routes/LandingPage'))
 const NotFoundPage = lazyWithRetry(() => import('./routes/NotFoundPage'))
 
+export const cityNotCooperatingEnabled =
```
klinzo, line 37: Should be moved to `buildConfig.ts`

build-configs/integreat-test-cms/index.ts
SteffiStuffel: good enough as it is

build-configs/integreat/assets/cityNotCooperatingTemplate.ts
```diff
@@ -0,0 +1,9 @@
+export const template =
```
klinzo, line 1:
```suggestion
const template =
```

klinzo (changes requested):
Some issue with text direction on RTL languages (e.g. http://localhost:9000/recommend/ar):
image

Edit: Copying the text works fine though so it is not a really big issue but it would be nice if we could fix this.

build-configs/integreat/assets/cityNotCooperatingTemplate.ts
```diff
@@ -0,0 +1,9 @@
...
+  'Sehr geehrte Damen und Herren,\n\n' +
+  'Ich wohne seit einiger Zeit in PLATZHALTER und habe neulich versucht, meinen Ort in der Integreat-App zu finden. Die App stellt zahlreiche wichtige Informationen über den Alltag und die Region in verschiedenen Sprachen zur Verfügung. Das erleichtert vielen Neuzugezogenen die Eingewöhnung, was ich sehr hilfreich finde.\n\n' +
```
klinzo, line 3: Should be:
```suggestion
  'Ich wohne seit einiger Zeit in [Ort] und habe neulich versucht, meinen Ort in der Integreat-App zu finden. Die App stellt zahlreiche wichtige Informationen über den Alltag und die Region in verschiedenen Sprachen zur Verfügung. Das erleichtert vielen Neuzugezogenen die Eingewöhnung, was ich sehr hilfreich finde.\n\n' +
```

build-configs/integreat/assets/cityNotCooperatingTemplate.ts
SteffiStuffel: quite irrelevant...

klinzo (approved):
Tested on firefox and android

klinzo:
Tested malte on android and firefox again and it is not shown as expected. Good to merge! Well done :tada:

PR #567, opened 2022-01-31
IGAPP-873: Add back function for change location 

f1sh1918:
Tested on ios & android!


native/src/components/SettingsHeader.tsx
```diff
@@ -32,6 +32,8 @@ const HeaderText = styled.Text`
   flex-direction: column;
   text-align-vertical: center;
   height: 50px;
+  /* fix text-align-vertical for ios, line-height has to be added */
+  line-height: 50px;
```
klinzo, line 36: What is the problem without this? Can you add a screenshot please?

native/src/testing/createNavigationPropMock.ts
```diff
@@ -4,7 +4,7 @@ import { NavigationPropType, RoutesType } from '../constants/NavigationTypes'
 
 const createNavigationMock = <T extends RoutesType>(routeIndex = 0): NavigationPropType<T> => ({
   dispatch: jest.fn(),
-  canGoBack: jest.fn(),
+  canGoBack: jest.fn(() => routeIndex > 0 && true),
```
klinzo, line 7:
```suggestion
  canGoBack: jest.fn(() => routeIndex > 0),
```

klinzo (approved):
The transparent header is not really transparent right now. Not sure if we should change this (if yes, in a different issue oc). Tested on Android

klinzo (changes requested):
Removing the `CLEAR_CITY` action leads to opening of the last opened city after a restart of the app. Perhaps this is not so easy after all... Also the header should not be shown if there is no back button. Otherwise it leads to content vanishing behind the header

f1sh1918:
> Removing the `CLEAR_CITY` action leads to opening of the last opened city after a restart of the app. Perhaps this is not so easy after all... Also the header should not be shown if there is no back button. Otherwise it leads to content vanishing behind the header

Opening the last opened city after app restart is not the expected behavior?
Hiding the header the header if there is no GoBack fkt should not be the problem

native/src/components/SettingsHeader.tsx
f1sh1918: image

native/src/components/TransparentHeader.tsx
```diff
@@ -84,7 +84,7 @@ const TransparentHeader = ({ navigation, route }: PropsType): ReactElement => {
     <BoxShadow theme={theme}>
       <Horizontal>
         <HorizontalLeft>
-          <HeaderBackButton onPress={navigation.goBack} labelVisible={false} />
+          {navigation.canGoBack() && <HeaderBackButton onPress={navigation.goBack} labelVisible={false} />}
```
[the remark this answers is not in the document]
f1sh1918: well if the header really gonna be "transparent" it wouldn't matter.
f.e. the imprint page has additionally the burger menu and not only the back button
Maybe we should have a separate transparent header for "modals", even in pdf view the grey background looks fine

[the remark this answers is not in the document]
f1sh1918: Created separate ticket
https://issues.tuerantuer.org/browse/IGAPP-909

SteffiStuffel:
> > Removing the `CLEAR_CITY` action leads to opening of the last opened city after a restart of the app. Perhaps this is not so easy after all... Also the header should not be shown if there is no back button. Otherwise it leads to content vanishing behind the header
> 
> Opening the last opened city after app restart is not the expected behavior? Hiding the header the header if there is no GoBack fkt should not be the problem

It is (last != currect), but the problem now occuring is:
Opening city1 -> Change city -> Open city2 -> close app -> reopen app -> city1 is opend, but it should be city2.

release-notes/unreleased/IGAPP-873-add-back-function-for-change-location.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-873
+show_in_stores: true
+platforms:
+  - ios
+  - android
+en: Possibility to go back on location change added.
```
SteffiStuffel, line 6: `Possibility added to go ...` sounds less german^^

SteffiStuffel (commented):
Tested this, looks good, but the error Steffen mentioned occurs. Will have a deeper look / test again, when this is fixed.

klinzo:
I mean if you click on change location it shows the landing page and city selection (which was probably intended because I want to change the city). If I close the app now and I am back to the city I was previously, that is not really intuitive imo. This also leads to some other problems like not unsubscribing from push news, not clearing the settings/resource cache and so on. So probably the current solution (which was oc proposed by me) of keeping the previous city open in the background does not work and we may have to pass a separate action to execute on the back arrow press.

f1sh1918:
> I mean if you click on change location it shows the landing page and city selection (which was probably intended because I want to change the city). If I close the app now and I am back to the city I was previously, that is not really intuitive imo. This also leads to some other problems like not unsubscribing from push news, not clearing the settings/resource cache and so on. So probably the current solution (which was oc proposed by me) of keeping the route of the previous city open in the background does not work and we may have to pass a separate action to execute on the back arrow press.

I moved the `CLEAR_CITY` fkt to goToCategory if you click on a city on the Landing page instead of clearing instantly when opening the landing page. Please test again

klinzo (commented):
So the issue obviously still persists that if you go back to the landing page/city selection and kill the app on next startup we still start at the previously selected city. However, I guess this is fine. Maybe you can ask Nikola and in a public mattermost channel for opinions on this behavior.

native/src/navigation/createNavigate.ts
```diff
@@ -47,17 +46,11 @@ const createNavigate =
       })
 
       if (routeInformation.route === LANDING_ROUTE) {
-        navigateToLanding({
```
klinzo, line 50: I think these changes should be reverted as they don't matter for your logic anyway right? Not sure but I have a fear that this will break deep linking otherwise.

klinzo (approved):

native/src/navigation/createNavigate.ts
f1sh1918: how is deep linking connected to the landing page? but i can revert that sure

PR #570, opened 2022-02-01
IGAPP-911: Remove hocs

klinzo:


Taggotty (approved):
Don't forget to also close IGAPP-896 after merging this.

Tested on android  emulator

native/src/components/SettingItem.tsx
```diff
@@ -48,25 +33,35 @@ const Description = styled.Text`
   color: ${props => props.theme.colors.textSecondaryColor};
 `
 
+type PropType = {
+  title: string
+  description?: string
+  onPress: () => void
+  t: TFunction<'settings'>
```
SteffiStuffel, line 40: When du you sill pass `t` and when du you use `useTranslations` in the component directly?

native/src/components/__tests__/Header.spec.tsx
```diff
@@ -72,49 +72,45 @@ describe('Header', () => {
...
+    const { getByLabelText } = render(<Header {...buildProps(false, true, goToLanguageChange)} />)
     expect(getByLabelText(t('changeLanguage'))).toHaveStyle({ opacity: 1 })
     fireEvent.press(getByLabelText(t('changeLanguage')))
     await waitFor(() => expect(goToLanguageChange).toHaveBeenCalledTimes(1))
   })
 
   it('search header button should be disabled and invisible while loading', () => {
     const props = buildProps(true, true, goToLanguageChange)
-    const { getByLabelText } = render(<Header {...props} />, { wrapper: wrapWithTheme })
+    const { getByLabelText } = render(<Header {...props} />)
```
SteffiStuffel, line 91: Nice :tada: 

Would be nice to also do this for lunes, I will open a ticket.

native/src/components/__tests__/NativeHtml.spec.tsx
```diff
@@ -20,11 +18,7 @@ jest.mock('react-redux', () => ({
 jest.mock('@react-navigation/native', () => ({
   useNavigation: () => mockNavigation
 }))
-
-jest.mock('styled-components', () => ({
-  ...jest.requireActual('styled-components'),
-  useTheme: () => mockTheme
-}))
+jest.mock('styled-components')
```
SteffiStuffel, line 21: Why do we need to mock styled-components here, but do not need to mock it in e.g. Header.spec.ts? Both use `useTheme`?

SteffiStuffel (approved):
not tested.

native/src/components/SettingItem.tsx
klinzo: I think I don't use `useTranslations` directly, do I?

native/src/components/__tests__/NativeHtml.spec.tsx
klinzo: Because in `Header.tsx` the theme is only passed to styled components (which is redundant) and to a component which is mocked. Therefore it does not matter that is undefined.

native/src/components/SettingItem.tsx
SteffiStuffel: yes you do in a lot of components, therefore it is a litte inconsistent. Thats why i'm asking.

klinzo: ah okay, so more in general. In this specific component its not used though, so I have to pass it (or use the hook oc). Yeah we have no rule for that but it was also inconsistent before. I think this is not a problem though. My rule of thumb if I write code myself is to pass it if it is used in the parent component and use the hook if not used in the parent component or there are usages in different routes with different translation namespaces.

PR #571, opened 2022-02-02
IGAPP-250: Disable Firebase Analytics

Taggotty:


klinzo (approved):
Not tested (not sure if that is possible at all?)

native/ios/Integreat/Info.plist
```diff
@@ -24,8 +24,8 @@
 	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
-	<key>FIREBASE_ANALYTICS_COLLECTION_ENABLED</key>
```
maxammann, line 27: Are you sure it is `DISABLED`? https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios

I also verified this back then for iOS and the log did say that it is in fact disabled.

maxammann (approved):

native/ios/Integreat/Info.plist
Taggotty: but you are right it should be deactivated. Not sure how I missed that

PR #541, opened 2022-01-10
IGAPP-718: Add map translations

klinzo:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **IGAPP**.

PR #573, opened 2022-02-07
IGAPP-901: Poi list bottom action sheet mobile web

f1sh1918:


web/src/components/MapView.tsx
```diff
@@ -14,9 +15,12 @@ import {
   mapMarker
 } from 'api-client'
 
-import MapPopup from './MapPopup'
+import { getSnapPoints } from '../utils/getSnapPoints'
 
+// TODO refactor height
```
f1sh1918, line 20: i couldn't find a way to set the height: 100%. Some help required

native/src/hooks/useUserLocation.ts
```diff
@@ -3,17 +3,9 @@ import { useCallback, useEffect, useState } from 'react'
 import { openSettings, RESULTS } from 'react-native-permissions'
 import SystemSetting from 'react-native-system-setting'
 
-import { checkLocationPermission, requestLocationPermission } from '../utils/LocationPermissionManager'
-
-type SuccessfulLocationState = { status: 'ready' }
```
f1sh1918, line 8: moved reusable type stuff to api-client to reuse it

web/src/routes/PoisPage.tsx
```diff
@@ -51,17 +64,28 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
   const { data: pois, loading, error: poisError } = useLoadFromEndpoint(requestPois)
 
+  const [currentFeature, setCurrentFeature] = useState<PoiFeature | null>(null)
+  const [featureLocations, setFeatureLocations] = useState<PoiFeature[] | null>(null)
+
+  useEffect(() => {
+    if (pois && locationState.message !== 'loading') {
+      setFeatureLocations(prepareFeatureLocations(pois, userCoordinates))
+    }
+  }, [pois, locationState.message])
```
klinzo, line 83: I think we should set the `react-hooks/exhaustive-deps` linting rule to error. You are missing `userCoordinates`

klinzo, line 83: Perhaps we could move the whole poi loading/distance calculation in a separate hook? Something like `useFeatureLocations`? This file is really big. I could also do this in a separate issue.

web/src/routes/PoisPage.tsx
```diff
@@ -145,22 +169,36 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
+      <BottomActionSheet title={currentFeature ? '' : t('sheetTitle')} ref={sheetRef}>
+        {currentFeature ? (
+          <div>{currentFeature.properties.title}</div>
+        ) : (
+          sortedPois && (
+            <ListWrapper>
+              <List noItemsMessage={t('noPois')} items={sortedPois} renderItem={renderPoiListItem} borderless />
+            </ListWrapper>
+          )
+        )}
```
klinzo, line 199: Make title optional in the bottom sheet and use the following:
```suggestion
      <BottomActionSheet title={currentFeature?.properties.title} ref={sheetRef}>
        {sortedPois && !currentFeature && (
          <ListWrapper>
            <List noItemsMessage={t('noPois')} items={sortedPois} renderItem={renderPoiListItem} borderless />
          </ListWrapper>
        )}
```
Or any specific reason why you don't use the title for feature titles?

I think `Entdecke deine Stadt` is nothing that should be shown (it is also not shown in native and is removed from the translations in #541).

web/src/components/MapView.tsx
```diff
@@ -47,18 +51,18 @@ const geolocateControlStyle: React.CSSProperties = {
...
   bboxViewport: MapViewViewport
-  featureCollection: PoiFeatureCollection
+  featureCollection: PoiFeatureCollection | null
+  currentFeature: PoiFeature | null
+  setCurrentFeature: (feature: PoiFeature | null) => void
 }
 
-const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): ReactElement => {
-  const { featureCollection, bboxViewport } = props
+const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<BottomSheetRef>): ReactElement => {
```
klinzo, line 59: Why do we need these kind of `forwardRef` components? It should be enough to pass the ref as regular prop. Or even better: Can't we just pass a `onTouchMove` function prop and `setCurrentFeature` as
``` ts
  selectFeature = (feature: PoiFeature | null) => {
    setCurrentFeature(feature)
    // set snap point as desired
  }
```
This component should not have to handle snap points.

web/src/components/PoiListItem.tsx
```diff
@@ -1,17 +1,73 @@
...
+  &:first-child {
+    padding-top: 0;
+  }
+`
+
+const Thumbnail = styled.img`
+  width: 100px;
+  height: 100px;
+  flex-shrink: 0;
+  border: 1px solid transparent;
```
klinzo, line 24: What is a transparent border good for?

web/src/components/PoiListItem.tsx
```diff
@@ -1,17 +1,73 @@
...
+const PoiListItem = ({ properties }: PropsType): ReactElement => {
+  const { t } = useTranslation('pois')
+  const { path, thumbnail, title, distance } = properties
+  return (
+    <ListItemContainer>
+      <CleanLink to={path}>
+        <Thumbnail alt='' src={thumbnail || PoiPlaceholder} />
+        <Description>
+          <Title>{title}</Title>
+          {distance && <Distance>{`${distance} ${t('unit')} ${t('distanceText')}`}</Distance>}
```
klinzo, line 66: Could you use the translation from #541 already?
```suggestion
          {distance && <Distance>{t('distanceKilometre', { distance })}</Distance>}
```

klinzo (changes requested):
Tested on firefox. Looks really good already! The font looks a little bit weird imo but maybe just to me.

web/src/components/BottomActionSheet.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { ReactElement, ReactNode } from 'react'
+import { BottomSheet, BottomSheetRef } from 'react-spring-bottom-sheet'
+import 'react-spring-bottom-sheet/dist/style.css'
+import styled from 'styled-components'
+
+import '../styles/BottomActionSheet.css'
+import { getSnapPoints } from '../utils/getSnapPoints'
```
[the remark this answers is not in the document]
Taggotty: agreed.

web/src/components/MapView.tsx
```diff
@@ -88,22 +92,25 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
...
+        // @ts-ignore BottomSheetRef does not provide current
+        onTouchMove={() => ref?.current.snapTo(({ maxHeight }: { maxHeight: number }) => getSnapPoints(maxHeight)[0])}>
         {/* To use geolocation in a development build you have to start the dev server with "yarn start --https" */}
         <GeolocateControl
           style={geolocateControlStyle}
           positionOptions={{ enableHighAccuracy: true }}
           trackUserLocation
         />
-        <Source id='location-pois' type='geojson' data={featureCollection}>
+        <Source id='location-pois' type='geojson' data={featureCollection!}>
```
Taggotty, line 108: if `featureCollection` is null here this will render an error.
```suggestion
        <Source id='location-pois' type='geojson' data={featureCollection}>
```

web/src/components/MapView.tsx
```diff
@@ -47,18 +51,18 @@ const geolocateControlStyle: React.CSSProperties = {
 
 type MapViewProps = {
   bboxViewport: MapViewViewport
-  featureCollection: PoiFeatureCollection
+  featureCollection: PoiFeatureCollection | null
```
Taggotty, line 54: when is the `featureCollection` null? Couldn't it just be empty?

web/src/components/MapView.tsx
Taggotty: Not quite sure, what you mean right here.

SteffiStuffel: Same for me, seems like the map has 100% height for me.

SteffiStuffel:
I tested and i'm not sure about some acceptance criteras  in the issue:

> the action sheet only appears on mobile devices
For me it seems the action sheet also appears in non-mobile device (tested on chrome)

> Remove PopUp for mobile view
Which PopUp? This one? https://app.zeplin.io/project/5f9040aed45a280985d019fb/screen/6048d89c2f9bf0ba179f76f5
Does not appear on mobile nor on desktop.

> By selecting a poi on the map the poi details display (content of current detail page) within the bottom sheet
This is not yet implemented and also not part of the ticket, isn't it?

web/src/components/MapView.tsx
f1sh1918: you can try to set here `height: 100%` and check that nothing gets rendered i dunno how to set the height without calculating it here

web/src/styles/BottomActionSheet.css
```diff
@@ -0,0 +1,16 @@
+/* BottomActionSheet Overrides */
+/* stylelint-disable */
+[data-rsbs-header] {
+  box-shadow: none;
+}
+
+[data-rsbs-header]:before {
+  width: 28px;
+  height: 3px;
+  background-color: #585858;
```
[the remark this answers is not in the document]
f1sh1918: @klinzo 
you could only import a `variables.css` file and copy our colors and font stuff from our ts files

web/src/routes/PoisPage.tsx
f1sh1918: Feel free to open a ticket but i think it would help more to separate the routes detail/overview. I mean these few lines do not make things worse ;)

web/src/styles/BottomActionSheet.css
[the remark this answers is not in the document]
klinzo: not necessary, then this is fine :)

web/src/components/PoiListItem.tsx
f1sh1918: you can not get rounded corners for pictures without a border

web/src/components/MapView.tsx
klinzo: I think the comments are more about that the comment is not really explanatory and adds more confusion than it solves.
```suggestion
// Workaround since nothing is rendered if height is set to 100% 
```

f1sh1918:
> I tested and i'm not sure about some acceptance criteras in the issue:
> 
> > the action sheet only appears on mobile devices
> 
> For me it seems the action sheet also appears in non-mobile device (tested on chrome)
> 
> > Remove PopUp for mobile view
> 
> Which PopUp? This one? https://app.zeplin.io/project/5f9040aed45a280985d019fb/screen/6048d89c2f9bf0ba179f76f5 Does not appear on mobile nor on desktop.
> 
> > By selecting a poi on the map the poi details display (content of current detail page) within the bottom sheet
> 
> This is not yet implemented and also not part of the ticket, isn't it?

Yes you're right about that. Probably we should think about changing the scope. Since we have proper detail view (separate ticket) i can not show sth proper when selecting the poi.

Indeed i removed the popup since we don't need it atm. When we have a desktop design, we can also remove the component probably

I dunno if it makes sense to implement now the whole conditional rendering thing for desktop if we don't have a design yet. Because if we don't want just to hide stuff with css but to remove it from the DOM, we need some additional libs.

web/src/components/Layout.tsx
```diff
@@ -113,13 +116,16 @@ type PropsType = {
  */
 const Layout = ({ asideStickyTop = 0, footer, header, toolbar, modal, children }: PropsType): JSX.Element => {
   const modalVisible = !!modal
+  const hideSidebar = window.location.href.includes(POIS_ROUTE)
```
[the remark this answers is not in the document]
f1sh1918: ok we already have a separate ticket for adding toolbar into the bottomActionSheet

web/src/routes/PoisPage.tsx
f1sh1918: because i didn't implement the details view yet for the bottom sheet ;)
your change in line 190 makes sense. The rest not because i would like to keep the conditional rendering for the details view which comes up with another ticket

web/src/components/MapView.tsx
f1sh1918: very good hint. thx

web/src/components/PoiListItem.tsx
f1sh1918: hm i prefer that we add a todo. because your pr is still draft an no one can check if its really working in my pr

web/src/components/Layout.tsx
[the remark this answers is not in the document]
klinzo: I don't understand this comment :see_no_evil: Or rather, what one has to do with another.

web/src/components/MapView.tsx
f1sh1918: its not an array so it can't be empty. Indeed in the `PoisPage` the `featureLocations` can be empty if there are no pois. These array gonna be embedded in the collection before passing to MapsView

web/src/components/MapView.tsx
f1sh1918: resolved this be a null check!

web/src/utils/getSnapPoints.ts
```diff
@@ -0,0 +1,5 @@
+// get all snapPoints of the bottomActionSheet using the maxHeight
```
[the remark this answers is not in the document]
f1sh1918: removed this in last commit

f1sh1918:
> Tested on firefox. Looks really good already! The font looks a little bit weird imo but maybe just to me.

i forgot to add font-family pls check again

web/src/components/Layout.tsx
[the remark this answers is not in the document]
f1sh1918: I mean maybe we want to have at least the feedback bar somewhere on the pois page thats why we have a separate ticket. Just removing it and not replacing it could be not what we want in the end?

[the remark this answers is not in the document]
klinzo: but you did remove it before my comment already? just in a complicated way, right?

klinzo:
@f1sh1918 some hidden comments not yet addressed :)

web/src/components/Layout.tsx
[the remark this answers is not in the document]
f1sh1918: indeed

web/src/components/BottomActionSheet.tsx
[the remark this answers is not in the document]
f1sh1918: this is not the case since, we need also `getSnapPoint` for the ref in `PoisPage` to snap to a particular snapPoint

web/src/routes/PoisPage.tsx
f1sh1918: updated linting rules and missing deps in last commit

web/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -48,14 +63,15 @@ describe('PoisPage', () => {
...
   it('should render a list with all pois', () => {
-    mockUseLoadFromEndpointOnceWithData(pois)
-    const { getByText } = renderPois()
-    expect(getByText(poi0.title)).toBeTruthy()
-    expect(getByText(poi1.title)).toBeTruthy()
+    mockUseLoadFromEndpointWithData(pois)
+    const { getByText, debug } = renderPois()
+    debug()
+    expect(getByText(poi0.location.name)).toBeTruthy()
+    expect(getByText(poi1.location.name)).toBeTruthy()
```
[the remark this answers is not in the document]
f1sh1918: because we now render the information from PoiFeature instead of PoiModel in the list

web/src/hooks/useUserLocation.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react'
```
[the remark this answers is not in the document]
f1sh1918: check last commit

f1sh1918:
> @f1sh1918 some hidden comments not yet addressed :)

now everything should be resolved

web/src/routes/PoisPage.tsx
```diff
@@ -81,14 +83,20 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
     }
   })
 
+  const selectFeature = (feature: PoiFeature | null, snapPoint?: number) => {
+    setCurrentFeature(feature)
+    if (snapPoint !== undefined) {
+      sheetRef.current?.snapTo(({ maxHeight }) => getSnapPoints(maxHeight)[snapPoint]!)
+    }
+  }
```
klinzo, line 91: I find the snap point logic kind of hard to understand. When do you pass a snap point index? And when don't you? Sometimes you pass `0`, sometimes `1` and sometimes nothing.

f1sh1918: you only pass a snappoint if you want to change the current snappoint of the bottom action sheet

web/src/components/MapView.tsx
```diff
@@ -68,18 +69,16 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
...
       }
     }
-  }, [featureCollection, queryLocation])
+  }, [featureCollection, queryLocation, selectFeature])
 
   const clickItem = (e: MapEvent) => {
     if (e.features?.length) {
-      setCurrentFeature(e.features[0])
-      togglePopup(true)
+      selectFeature(e.features[0], 1)
```
SteffiStuffel, line 79: Ternary would be nicer `
  const clickItem = (e: MapEvent) => (e.features?.length ? selectFeature(e.features[0], 1) : selectFeature(null))`

web/src/components/MapView.tsx
```diff
@@ -88,9 +87,12 @@ const MapView: React.FunctionComponent<MapViewProps> = (props: MapViewProps): Re
       <ReactMapGL
         interactiveLayerIds={[layerStyle.id!]}
         {...viewport}
+        height='100%'
+        width='100%'
         onViewportChange={setViewport}
         mapStyle={mapConfig.styleJSON}
-        onClick={clickItem}>
+        onClick={clickItem}
```
SteffiStuffel, line 94: Mostly it is nicer to name it depending on what it does. e.g. selectFeature.

web/src/components/PoiListItem.tsx
SteffiStuffel: yes you can by adding border-radius?

web/src/components/__tests__/PoiListItem.spec.tsx
```diff
@@ -1,47 +1,29 @@
...
+    title: 'Sprachforum Internationale Deutschkurse',
+    id: 10,
+    symbol: 'marker_15',
+    path: '/testumgebung/de/locations/sprachforum-internationale-deutschkurse',
+    urlSlug: 'sprachforum-internationale-deutschkurse',
+    address: 'Neidhardstr. 15',
+    distance: '5422.0'
+  }
 
   it('should render and match snapshot', () => {
```
SteffiStuffel, line 21: not a real good name, but not from your pr.

web/tools/webpack.config.ts
```diff
@@ -253,7 +253,6 @@ const createConfig = (
         },
         {
           test: /\.css$/,
-          include: /node_modules/,
```
SteffiStuffel, line 256: Why removed this?

SteffiStuffel (commented):
Tested only on web, although something changed on native.

Still not sure, what should work now and what would be done in other PRs, but as this is nothing we release now, i'm fine with it.

web/src/components/BottomActionSheet.tsx
```diff
@@ -0,0 +1,37 @@
...
+  font-size: 1.25rem;
+  font-family: ${props => props.theme.fonts.web.contentFont};
+`
+
+type BottomActionSheetProps = {
+  title?: string
+  children: ReactNode
+}
+
+const BottomActionSheet = React.forwardRef(
```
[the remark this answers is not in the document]
f1sh1918: because the ref can not be passed properly and used in `PoisPage`
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

web/src/components/MapView.tsx
```diff
@@ -14,9 +14,10 @@ import {
   mapMarker
 } from 'api-client'
 
-import MapPopup from './MapPopup'
-
+// Workaround since nothing is rendered if height is set to 100%
 const MapContainer = styled.div`
+  ${`height: calc(100vh - 190px);`}
```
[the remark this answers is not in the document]
f1sh1918: updated comment above!

web/src/components/MapView.tsx
```diff
@@ -48,13 +49,13 @@ const geolocateControlStyle: React.CSSProperties = {
 type MapViewProps = {
   bboxViewport: MapViewViewport
   featureCollection: PoiFeatureCollection
+  currentFeature: PoiFeature | null
```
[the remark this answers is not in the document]
f1sh1918: hm dunno where this change comes from. current version ist `selectedFeature`

web/src/components/MapView.tsx
f1sh1918: i splitted up some functions like `changeSnapPoint` and had to `updateQueryParams` so it can not be done in tenary expression. sorry for the meantime update. i recognized some updated comments after approval from @klinzo

web/src/components/MapView.tsx
f1sh1918: used `onSelectFeature`

web/tools/webpack.config.ts
f1sh1918: crashes css files. don't see any case why this is needed. everything works as expected without it

web/src/components/__tests__/PoiListItem.spec.tsx
f1sh1918: rejected change

api-client/src/utils/geoJson.ts
```diff
@@ -1,6 +1,22 @@
...
+import { LocationType, PoiModel } from '../index'
 import { PoiFeature, PoiFeatureCollection } from '../maps'
 
 export const embedInCollection = (features: PoiFeature[]): PoiFeatureCollection => ({
   type: 'FeatureCollection',
   features
 })
+
+// Calculate distance for all Feature Locations
+export const prepareFeatureLocations = (pois: Array<PoiModel>, userLocation?: LocationType | null): PoiFeature[] =>
```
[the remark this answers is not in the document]
f1sh1918: if you check line 20. we return only the `poi.featureLocation` of each item. So in any case we have to map through it, as far as i can see

SteffiStuffel (approved):

PR #582, opened 2022-02-08
IGAPP-885: Add commands for linting and testing only changed files

f1sh1918:

Since git-diff with eslint does not work with current eslint version, i added a plugin and a extended config for only checking changed files


klinzo (approved):
Not tested

Taggotty (approved):
tested for api-client 👍🏼

SteffiStuffel:
> According to the issue this should also be added to the lunes project.

Created a new ticket for lunes.

PR #584, opened 2022-02-09
IGAPP-xxx: Move release notes to years directory

klinzo:


klinzo:
Moves all release notes from to 2020 and 2021 to corresponding directories. We don't have to merge this but I think it makes the release notes a little bit tidier.

Taggotty (approved):
Makes sense imo. Looks quite messy without more structure 🤷🏼‍♀️

SteffiStuffel (approved):
I also think this makes sense.

PR #586, opened 2022-02-14
IGAPP-912: Fix Linting Warnings

Taggotty:

Most warnings left now come from array indexing with the non-null-assertiion. 
I was also wondering why we don't have a `node_modules` folder and a `yarn.lock` in the tools folder. Any reason for this?


klinzo (changes requested):
yarn.lock should be reverted. There is a yarn.lock in the tools dir: https://github.com/digitalfabrik/integreat-app/blob/main/tools/yarn.lock
The node_modules are not synced to git, so the folder is only created once you run `yarn install` in the tools folder

yarn.lock
```diff
@@ -3548,6 +3548,134 @@
     "@nodelib/fs.scandir" "2.1.4"
     fastq "^1.6.0"
 
+"@octokit/auth-app@^2.4.5":
```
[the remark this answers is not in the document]
Taggotty: will revert. not sure what happened here.

klinzo (approved):

native/src/routes/ExternalOfferContainer.tsx
```diff
@@ -10,7 +10,7 @@ type PropsType = {
   navigation: NavigationPropType<ExternalOfferRouteType>
 }
 
-const ExternalOfferContainer = ({ route }: PropsType): ReactElement => {
+const ExternalOfferContainer = ({ route, navigation: _navigation }: PropsType): ReactElement => {
```
SteffiStuffel, line 13: Why not just removing `navigation` from PropsType?

SteffiStuffel (approved):

native/src/routes/ExternalOfferContainer.tsx
Taggotty: to keep the information, that navigation is available here, but we do not use it

PR #587, opened 2022-02-14
IGAPP-865: Alignment of RTL text on 2nd level

klinzo:

Hopefully this does not break the search layout again. Only testable on iOS :)


f1sh1918 (approved):
tested on ios

Taggotty (approved):
Not tested.

PR #589, opened 2022-02-16
IGAPP-883: Only log firebase errors

klinzo:

### DIFF
--- a/native/src/utils/PushNotificationsManager.ts
+++ b/native/src/utils/PushNotificationsManager.ts
@@ -9,7 +9,7 @@ import buildConfig from '../constants/buildConfig'
 import navigateToDeepLink from '../navigation/navigateToDeepLink'
 import urlFromRouteInformation from '../navigation/url'
 import appSettings from './AppSettings'
-import { log, reportError } from './sentry'
+import { log } from './sentry'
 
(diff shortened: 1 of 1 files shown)


klinzo:
No need to report these kind of errors to sentry.

native/src/utils/PushNotificationsManager.ts
```diff
@@ -43,10 +43,10 @@ export const unsubscribeNews = async (city: string, language: string): Promise<v
   try {
     const messaging = await importFirebaseMessaging()
     await messaging().unsubscribeFromTopic(topic)
+    log(`Unsubscribed from ${topic} topic!`)
   } catch (e) {
-    reportError(e)
+    log(`Failed to unsubscribe from ${topic} topic: ${e instanceof Error ? e.message : 'Unknown Error'}`)
```
Taggotty, line 48: But isn't it also a firebase error if unsubscribing fails?

SteffiStuffel (approved):
Why only a Draft?

klinzo:
> Why only a Draft?

@SteffiStuffel because this basically suppresses sending all errors of firebase to sentry. Not sure what we can do apart from that though. Perhaps I'll try upgrading.

klinzo:
Won't do, we'll just ignore this in sentry.

PR #590, opened 2022-02-17
IGAPP-928: Upgrade react native to v0.67.2

klinzo:


klinzo:
I had to switch react-native-static-server library because it was not compatible with react native 0.67. New one is a fork: https://github.com/birdofpreyru/react-native-static-server

f1sh1918 (approved):
Tested on android and ios. No issues so far

SteffiStuffel (approved):

PR #591, opened 2022-02-17
IGAPP-906: Add support to automatically wrap text with soft hyphens

klinzo:


klinzo:
Sadly this only works in certain cases on android, see https://github.com/facebook/react-native/issues/31878#issuecomment-1042944003
Since it does not break anything and works really nice on iOS and web I'd merge this anyway.

Testable here: https://integreat.app/testumgebung/de/eineseitemiteinemsehrsehrsehrlangemtitelweilwirinderappleidereinenbughaben

Taggotty (approved):
Tested on Android device. As expected: Works sometimes :)

Taggotty:
Why the changes in native though?

klinzo:
> Why the changes in native though?

What do you mean by this? We have to support `&shy;` in native as well, even though it only works sometimes on android. Apart from that I only moved the html parse logic to the api-client as we need it there anyway.

Taggotty:
> > Why the changes in native though?
> 
> What do you mean by this? We have to support `&shy;` in native as well, even though it only works sometimes on android. Apart from that I only moved the html parse logic to the api-client as we need it there anyway.

Ah. I was confused about the changes in native. But now I see, that you just moved the parsing to the api-client. Much cleaner now.

api-client/src/endpoints/createLocalNewsEndpoint.ts
```diff
@@ -19,16 +18,14 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<LocalNewsModel>> =>
...
-            timestamp: moment.tz(localNews.timestamp, 'GMT'),
-            title: decodedTitle,
-            message: decodedMessage
-          })
-        })
+        json.map(
+          (localNews: JsonLocalNewsType) =>
+            new LocalNewsModel({
+              id: localNews.id,
+              timestamp: moment.tz(localNews.timestamp, 'GMT'),
```
f1sh1918, line 25: since 'GMT' is used quite offen in several files i would prefer using a const here

klinzo: Since I did not change anything here and moment will be removed anyway (https://issues.tuerantuer.org/browse/IGAPP-281) I will keep it as it is.

f1sh1918 (approved):

PR #597, opened 2022-02-23
IGAPP-597: Migrate tools to typescript

klinzo:

Fixes failing builds in #586. Hopefully not too many builds will fail in our next releases.


Taggotty (approved):

Taggotty:
Not tested.

SteffiStuffel (approved):
Tested a few commands.

sentry-release should als be converted, why have you skipped it?

klinzo:
> Tested a few commands.
> 
> sentry-release should als be converted, why have you skipped it?

Because this was only done to fix the build errors in #586 and converting sentry-release was not required. Since we initially decided not to convert tools to typescript at all I think that is fine. The least possibly breaking changes the better.

PR #601, opened 2022-02-28
IGAPP-904: Support multitasking on iOS

klinzo:


klinzo:
Obviously only testable on iOS. This adds support for iOS multitasking on tablets (https://support.apple.com/en-us/HT207582). Seems to be well supported and looks quite good.

f1sh1918:
A short description how to test would be nice!

f1sh1918 (approved):
tested on ios

PR #605, opened 2022-03-01
IGAPP-934: Poi list view for desktop

f1sh1918:


web/src/components/FeedbackModal.tsx
```diff
@@ -27,7 +27,7 @@ const ModalContainer = styled.div`
   right: 0;
   bottom: 0;
   left: 0;
-  z-index: 11;
+  z-index: 111;
```
f1sh1918, line 30: Modal should always be on top. in this case over the `BottomActionSheet`

web/src/components/Footer.tsx
```diff
@@ -13,6 +13,7 @@ const FooterContainer = styled.footer`
   justify-content: center;
   padding: 15px 5px;
   background-color: ${props => props.theme.colors.backgroundAccentColor};
+  box-shadow: 0 2px 3px 3px rgba(0, 0, 0, 0.1);
```
f1sh1918, line 16: added box shadow to look similar to the `Header` box shadow and to separate from the PoiList Toolbar

web/src/components/Layout.tsx
```diff
@@ -28,7 +28,7 @@ const RichLayout = styled.div`
   }
 `
 
-const Body = styled.div`
+const Body = styled.div<{ fullWidth: boolean }>`
```
f1sh1918, line 31: added to make layout fullwidth for Pois Page

web/src/components/Toolbar.tsx
```diff
@@ -36,18 +35,28 @@ const Headline = styled.h5`
...
+  iconDirection?: 'row' | 'column'
 }
 
-const Toolbar = ({ children, className, t, viewportSmall }: PropsType): ReactElement => (
-  <ToolbarContainer className={className}>
-    {viewportSmall && <Headline>{t('isThisSiteUseful')}</Headline>}
-    {children}
-  </ToolbarContainer>
-)
+const Toolbar: React.FC<ToolbarProps> = ({
```
f1sh1918, line 45: refactored to functional component

web/src/constants/dimensions.ts
```diff
@@ -14,8 +17,11 @@ const dimensions: DimensionsType = {
   minMaxWidth: 'screen and (min-width: 1100px)',
   maxWidth: 1100,
   toolbarWidth: 125,
+  toolbarHeight: 73,
   headerHeightLarge: 90,
-  headerHeightSmall: 70
+  headerHeightSmall: 70,
+  footerHeight: 50,
+  navigationMenuHeight: 90
```
f1sh1918, line 24: heights are needed to calculate the `main`container for the pois page

Taggotty (approved):

web/src/components/FeedbackModal.tsx
SteffiStuffel: 111 is not a good value though.

If it needs to be bigger than 110, use 120, so we still have the possibility to put a element between, if needed in the future. If 100 works, use this.

web/src/components/Footer.tsx
SteffiStuffel: Looking at the designs the footer should not look like this on the maps page, but I guess it is not part of this ticket.

web/src/components/LocationToolbar.tsx
```diff
@@ -7,14 +7,15 @@ type PropsType = {
   openFeedbackModal: (rating: FeedbackRatingType) => void
   children?: ReactNode
   viewportSmall: boolean
+  iconDirection?: 'row' | 'column'
 }
 
 class LocationToolbar extends React.PureComponent<PropsType> {
   render(): ReactNode {
-    const { viewportSmall, children, openFeedbackModal } = this.props
+    const { viewportSmall, children, openFeedbackModal, iconDirection = 'column' } = this.props
```
SteffiStuffel, line 15: no need to set the default value here, as it is set to "column" in `Toolbar.tsx` anyway if not set. I think it is better not to set the default twice.

web/src/components/PoisDesktop.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { ReactElement } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+
+import { PoiFeature } from 'api-client'
+
+import dimensions from '../constants/dimensions'
+
+const ListViewWrapper = styled.div<{ panelHeights: number }>`
+  min-width: 370px;
```
SteffiStuffel, line 10: I think 370 is to big. 150 is also enough otherwise if the display is very small, but still counts as a desktop, there is almost no card visiable but only the sidebar.

web/src/components/PoisDesktop.tsx
```diff
@@ -0,0 +1,66 @@
...
+  justify-content: center;
+  background-color: ${props => props.theme.colors.backgroundAccentColor};
+  box-shadow: 1px 0 4px 0 rgba(0, 0, 0, 0.2);
+`
+
+const ListHeader = styled.div`
+  padding-top: 32px;
+  padding-bottom: 20px;
+  text-align: center;
+  font-size: 18px;
```
SteffiStuffel, line 27: somehow frustrating there is no appropriate fontSize in the theme for this.

web/src/hooks/useWindowDimensions.ts
```diff
@@ -1,6 +1,6 @@
 import { useState, useEffect } from 'react'
 
-const VIEWPORT_SMALL_THRESHOLD = 750
+const VIEWPORT_SMALL_THRESHOLD = 768
```
SteffiStuffel, line 3: Why is this such a weird number?

SteffiStuffel (commented):
Tested on Chrome:
- Scrollbar looks shitty for me, but I think it is just the standart scrollbar of the browser, but in the "middle" of the screen it seems to big/clumsy.
- "Haben Sie Feedback" text is missing, it is there for mobile view and also in the designs. Can't the component rendering the feedback be shared so we always have consitent behaviour?
- Some details from the design are not implementet correctly, e.g. the separation lines are to thick / dark.
- distance is not displayed for me, but I also did not allow location permission (as i wasn't asked), so this may be fine
- For better responsiveness it would be nice it the left part would be a little responsive (scale down, so the layout for the smalles-desktop-devices still looks good)

Can you also add few tests?

klinzo:
> Tested on Chrome:
> 
>     * Scrollbar looks shitty for me, but I think it is just the standart scrollbar of the browser, but in the "middle" of the screen it seems to big/clumsy.
> 
>     * "Haben Sie Feedback" text is missing, it is there for mobile view and also in the designs. Can't the component rendering the feedback be shared so we always have consitent behaviour?
> 
>     * Some details from the design are not implementet correctly, e.g. the separation lines are to thick / dark.
> 
>     * distance is not displayed for me, but I also did not allow location permission (as i wasn't asked), so this may be fine
> 
Probably you need to start using `yarn start --https`

web/src/components/Toolbar.tsx
klinzo: Was a functional component before already? :D

web/src/routes/PoisPage.tsx
```diff
@@ -154,32 +171,52 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
+          <PoisDesktop
+            currentFeature={currentFeature}
+            toolbar={toolbar(setFeedbackModalRating)}
+            panelHeights={panelHeights}
+            sortedPois={sortedPois}
+            mapView={mapView}
+            poiList={poiList}
+          />
         )}
-        {/* TODO add feedback toolbar IGAPP-914 */}
```
klinzo, line 181: Next time better separate the changes and create two PRs.

web/src/components/PoisDesktop.tsx
```diff
@@ -0,0 +1,66 @@
...
+  mapView,
+  toolbar
+}: PoisDesktopProps): ReactElement => {
+  const { t } = useTranslation('pois')
+  return (
+    <>
+      <div>
+        <ListViewWrapper panelHeights={panelHeights}>
+          <ListHeader>{currentFeature?.properties.title || t('listTitle')}</ListHeader>
+          {!currentFeature && sortedPois.length > 0 && poiList}
```
klinzo, line 57: I think it is quite weird that you pass `currentFeature` and `sortedPois` only to decide whether to render the list or not.
The check for the amount of sorted pois is definitely not needed as we probably want the message "no pois" to show:
```suggestion
          {!currentFeature && poiList}
```
The check for `currentFeature` can probably be removed once you remove the details view/move it to the sidebar, right?

web/src/routes/PoisPage.tsx
```diff
@@ -154,32 +171,52 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
-      <BottomActionSheet title={currentFeature?.properties.title || t('listTitle')} ref={sheetRef}>
-        {sortedPois.length > 0 && !currentFeature && (
-          <ListWrapper>
-            <List noItemsMessage={t('noPois')} items={sortedPois} renderItem={renderPoiListItem} borderless />
-          </ListWrapper>
+      <PoisPageWrapper panelHeights={panelHeights}>
+        {viewportSmall ? (
+          <PoisMobile
+            currentFeature={currentFeature}
+            toolbar={toolbar(setFeedbackModalRating)}
```
klinzo, line 202:
```suggestion
            toolbar={toolbar}
```

web/src/routes/PoisPage.tsx
```diff
@@ -154,32 +171,52 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
+      selectFeature={selectFeature}
+      changeSnapPoint={changeSnapPoint}
+      featureCollection={embedInCollection(sortedPois)}
+      bboxViewport={moveViewToBBox(cityModel.boundingBox, defaultViewportConfig)}
+      currentFeature={currentFeature}
+    />
+  )
+
+  const poiList = <List noItemsMessage={t('noPois')} items={sortedPois} renderItem={renderPoiListItem} borderless />
+  // To get the map fullheight its reduced by header, footer, navMenu
```
klinzo, line 192: Personal opinion: I think this comment does not really add any necessary information

f1sh1918: check last commit

web/src/hooks/useWindowDimensions.ts
f1sh1918: this is a standard viewport width used in many frameworks. 750 is weird ;) f.e. here under medium
https://getbootstrap.com/docs/4.0/layout/grid/#grid-options

web/src/components/PoisDesktop.tsx
f1sh1918: yea we have to refactor some typo stuff definitly but even think if we want to use css clamp for better responsiveness
https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()

web/src/components/PoisDesktop.tsx
f1sh1918: there are always edge cases in responsive design. Here is now a pic of 770px. Since we have no tablet design, we can either start showing desktop mode for large desktops `>992` and use the bottom action bar instead before. but i can not decrease the size for the panel even more

image

web/src/components/Toolbar.tsx
f1sh1918: oh sorry your are right :D just removed HOC

web/src/components/Footer.tsx
f1sh1918: yes we have to decide what happens with the footer, maybe it gonna be a small overlay on the map view. But yes separate ticket

web/src/components/PoisDesktop.tsx
f1sh1918: no, because we then pass here the details view and render it
`{currentFeature ? poiDetails : poiList}`

web/src/components/PoisMobile.tsx
```diff
@@ -0,0 +1,34 @@
...
+  (
+    { mapView, currentFeature, toolbar, sortedPois, poiList }: PoisMobileProps,
+    ref: React.Ref<BottomSheetRef>
+  ): ReactElement => {
+    const { t } = useTranslation('pois')
+    return (
+      <>
+        {mapView}
+        <BottomActionSheet title={currentFeature?.properties.title || t('listTitle')} toolbar={toolbar} ref={ref}>
+          {sortedPois.length > 0 && !currentFeature && poiList}
```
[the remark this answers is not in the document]
f1sh1918: check comment above

f1sh1918:
> Tested on Chrome:

> * "Haben Sie Feedback" text is missing, it is there for mobile view and also in the designs. Can't the component rendering the feedback be shared so we always have consitent behaviour?

It gonna be changed in the designs, since its never displayed it on desktop view
> 
> * Scrollbar looks shitty for me, but I think it is just the standart scrollbar of the browser, but in the "middle" of the screen it seems to big/clumsy.

This quite difficult since there are `scrollbar-kits` but not supported by all browsers like firefox. If we have fixed footer container i can not avoid this
> * Some details from the design are not implementet correctly, e.g. the separation lines are to thick / dark.
The used colors is not in design, gonna talk to nikola

> * distance is not displayed for me, but I also did not allow location permission (as i wasn't asked), so this may be fine
Hm should be browser preferences it def. works even you have to `yarn start --https`
Even the distance calc gonna be fixed in another pr, so value is atm wrong

> * For better responsiveness it would be nice it the left part would be a little responsive (scale down, so the layout for the smalles-desktop-devices still looks good)
> 
I added clamp() for responsive padding, fonts, images, check it out
> Can you also add few tests?
I wasn't sure what to test here because there are not real "new" components with logic, so if there are any suggestions. But i also gonna think about it once more!

web/src/components/LocationLayout.tsx
```diff
@@ -46,6 +46,7 @@ const LocationLayout = (props: PropsType): ReactElement => {
 
   return (
     <Layout
+      fullWidth={route === POIS_ROUTE}
```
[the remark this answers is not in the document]
f1sh1918: check last commit

web/src/components/PoisDesktop.tsx
SteffiStuffel: Not really a edge case. I just think it would be better / more responsive design if the sidebar width is more responsive. Any argument against changing the min-width to a smaller value?

klinzo (approved):
More things I encountered:
* No pois are shown until location request has been accepted or denied. This makes the layout/ui jump since the pois are only shown afterwards and the sidebar is smaller if only a nothing found message is displayed.
* Navigating back/clicking on `Karte` after selecting on a poi does nothing (except changing the url)

The design/responsiveness is really well done though.

f1sh1918:
> 

@klinzo which browser did you test?
I tested on chrome, firefox and mobile phone and deselection works for me (clicking on the map)

klinzo:
> > 
> 
> @klinzo which browser did you test? I tested on chrome, firefox and mobile phone and deselection works for me (clicking on the map)

I meant clicking on the header navigation item "Karte", sorry. Clicking on the map only works if there is no poi in the press area. If you click on another poi, nothing happens.

f1sh1918:
> > > 
> > 
> > 
> > @klinzo which browser did you test? I tested on chrome, firefox and mobile phone and deselection works for me (clicking on the map)
> 
> I meant clicking on the header navigation item "Karte", sorry. Clicking on the map only works if there is no poi in the press area. If you click on another poi, nothing happens.

ok pls create a bug ticket for "Karte" if you click on another poi this poi gets selected so not sure what else should happen

PR #607, opened 2022-03-02
IGAPP-939: Wrong calculation of poi distance

f1sh1918:

Since there are different implementations of Coords even for mapbox and there is no clear documentation of long,lat we keep it as it is. Just important that we do it everywhere same way `coords[long,lat]`
 https://docs.mapbox.com/help/glossary/lat-lon/#coordinate-format-handling
   


web/src/components/MapView.tsx
```diff
@@ -94,6 +94,7 @@ const MapView = (props: MapViewProps): ReactElement => {
   return (
     <MapContainer>
       <ReactMapGL
+        reuseMaps
```
f1sh1918, line 97: https://visgl.github.io/react-map-gl/docs/get-started/tips-and-tricks#minimize-cost-from-frequent-re-mounting

web/src/hooks/__tests__/useFeatureLocations.spec.tsx
```diff
@@ -45,7 +45,8 @@ describe('useFeatureLocations', () => {
   }
 
   it('should correctly receive featureLocation properties', () => {
-    const { getByText } = render(<MockComponent />)
+    const { getByText, debug } = render(<MockComponent />)
+    debug()
```
klinzo, line 49:
```suggestion
    const { getByText } = render(<MockComponent />)
```

web/src/hooks/useUserLocation.ts
```diff
@@ -35,7 +35,7 @@ export const useUserLocation = (): UserLocationType => {
         // eslint-disable-next-line -- check needed for tests
         if (coords) {
           const { latitude, longitude } = coords
-          setUserCoordinates([latitude, longitude])
+          setUserCoordinates([longitude, latitude])
```
klinzo, line 38: I think it should be the other way round (at least that is the default, right?) At least [latitude, longitude] seems to be the default, e.g. on wikipedia.

klinzo (changes requested):
Tested on Android, works fine

web/src/hooks/useUserLocation.ts
f1sh1918: true...
https://docs.mapbox.com/mapbox-gl-js/api/geography/
Then our poi coordinates are also wrong

web/src/hooks/__tests__/useFeatureLocations.spec.tsx
f1sh1918: maybe we could eslint `debug()` ;)

api-client/src/endpoints/__tests__/createCitiesEndpoint.spec.ts
```diff
@@ -74,7 +74,7 @@ describe('cities', () => {
             latitude: 48.3705449
           }
         },
-        boundingBox: [10.7880103, 48.447238, 11.0174493, 48.297834]
+        boundingBox: [48.447238, 10.7880103, 48.297834, 11.0174493]
```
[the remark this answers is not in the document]
Taggotty: I think it would make sense to change it in the django endpoint, as you normally always start with the latitude, when writing coordinates. https://stackoverflow.com/questions/7309121/preferred-order-of-writing-latitude-longitude-tuples-in-gis-services

klinzo (approved):

api-client/src/endpoints/__tests__/createCitiesEndpoint.spec.ts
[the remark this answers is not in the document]
f1sh1918: since everything is now done in cords[long,lat] we can also keep it. there is no clear rule for that. the documentations differ. Check my comment for mapbox f.e. i don't see that big value in changing everything. We just have to keep it everything there same

Taggotty (approved):
Tested on Android Emulator

PR #623, opened 2022-03-13
IGAPP-950: Add shelter contact form

klinzo:


f1sh1918 (approved):
Tested and received a call immediately. nice work!
I'm not sure about the form validation. maybe we should improve it a little bit. If not this could block the whole contact process

klinzo:
> Tested and received a call immediately. nice work! I'm not sure about the form validation. maybe we should improve it a little bit. If not this could block the whole contact process

Yes, sounds good. Will create an issue for this. Going to merge this now as this is quite important.

PR #624, opened 2022-03-13
IGAPP-951: Add filtering to shelter offers

f1sh1918:

Added minBeds & pets filter
http://localhost:9000/augsburg/de/offers/wohnraum


api-client/src/endpoints/createShelterEndpoint.ts
```diff
@@ -6,14 +6,38 @@ import { JsonShelterType } from '../types'
 const SHELTER_URL = 'wohnraum.tuerantuer.org/wp-json/accommodations'
 
 export const SHELTER_ENDPOINT_NAME = 'shelter'
-type Params = { type: 'detail'; id: string; cityCode: string } | { type: 'list'; page: number; cityCode: string }
+type Params =
+  | { type: 'detail'; id: string; cityCode: string }
+  | { type: 'list'; page: number; cityCode: string; filter?: FilterProps }
```
klinzo, line 11: Does this have to be optional? I think logic would be a little bit easier if it is not and it should always be possible to set filters doing nothing as default. I think it is also never null/undefined.

api-client/src/endpoints/createShelterEndpoint.ts
```diff
@@ -6,14 +6,38 @@ import { JsonShelterType } from '../types'
...
+    .join('')
 
 export default (): Endpoint<Params, ShelterModel[]> =>
   new EndpointBuilder<Params, ShelterModel[]>(SHELTER_ENDPOINT_NAME)
     .withParamsToUrlMapper(params => {
       const baseUrl = `https://${params.cityCode}.${SHELTER_URL}`
       if (params.type === 'list') {
-        return `${baseUrl}/list?page=${params.page}`
+        const filterParams = params.filter ? `${getFilterQueryParams(params.filter)}` : ''
+        return `${baseUrl}/list?page=${params.page}${filterParams}`
```
klinzo, line 40: Optional oc but seems a lot simpler to me.
```suggestion
        const { beds, pets } = params.filter
        let filterParams = beds ? `&min_beds=${beds}` : ''
        if (pets) {
          filterParams = `${filterParams}&pets=${pets}`
        }
        return `${baseUrl}/list?page=${params.page}${filterParams}`
```

translations/translations.json
```diff
@@ -6126,7 +6126,9 @@
       "shelterType": "Art des Wohnraums",
       "startDate": "frühestens verfügbar",
       "availableBeds": "verfügbare Betten",
-      "duration": "mögliche Aufenthaltsdauer"
+      "duration": "mögliche Aufenthaltsdauer",
+      "facetBeds": "min. Anzahl Betten",
```
klinzo, line 6130: What is a facet? :see_no_evil: I think `Anzahl Betten` is enough, you also do that for english already.

translations/translations.json
```diff
@@ -6126,7 +6126,9 @@
...
+      "facetPets": "Haustiere erlaubt"
```
klinzo, line 6131: We already have that translation. Doesn't hurt to have it twice though

web/src/components/FacetInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+  flex-wrap: wrap;
+`
+
+const StyledInput = styled.input`
+  font-size: 14px;
+  margin-left: -1px;
+  border-radius: 0.25rem;
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+  background-color: #fff;
```
klinzo, line 15: Maybe we could use colors from theme as much as possible.

web/src/components/FacetInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+const FacetInput: React.FC<FacetInputProps> = ({
+  value,
+  icon,
+  altTag,
+  updateSearchFilter,
+  name,
+  placeholder
+}: FacetInputProps): ReactElement => (
+  <FilterFacet>
+    <FilterIcon>
```
klinzo, line 50: No tooltip here?

web/src/components/InfiniteScrollList.tsx
```diff
@@ -62,14 +63,14 @@ const InfiniteScrollList = <T,>({
...
-    if (resetOnLanguageChange) {
+    if (resetOnLanguageChange || filterValues) {
       setData([])
       setError(null)
       setLoading(false)
       setHasMore(true)
       setPage(defaultPage)
     }
-  }, [languageCode, resetOnLanguageChange, defaultPage])
+  }, [languageCode, resetOnLanguageChange, defaultPage, filterValues])
```
klinzo, line 73: Does this work? I think we could just use the dependency on `[loadData]`, that function should change on language change/filter value change already.

klinzo (commented):
Everything is optional.

translations/translations.json
f1sh1918: facet is the regular name for a filter

web/src/components/FacetInput.tsx
f1sh1918: i think the placeholder says enough

web/src/components/InfiniteScrollList.tsx
f1sh1918: you have to reset the data first and then the load gets triggered auto (on filter changed)

web/src/routes/ShelterPage.tsx
```diff
@@ -20,10 +22,17 @@ const ShelterPage = ({ cityModel, cityCode, languageCode, pathname, languages }:
   const { shelterId } = useParams()
   const { viewportSmall } = useWindowDimensions()
   const { t } = useTranslation('shelter')
+  const [filter, setFilter] = useState<FilterProps>({ beds: null, pets: null })
```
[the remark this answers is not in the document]
f1sh1918: i think if we get more filter, an filter object is the better way than to set for each one a new variable

api-client/src/endpoints/createShelterEndpoint.ts
f1sh1918: if we get more filters, you want to add every filter here? Don't think thats an improvement to be honest.

klinzo: yep, would do it that way, we will never have more then five filters. but your way is fine so just keep it.

web/src/routes/ShelterPage.tsx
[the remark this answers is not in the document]
klinzo: yep, would do it that way, we will never have more then five filters. but your way is fine so just keep it.

klinzo, line 25: This default should be fine, then you don't need to check whether params are defined in the endpoint. At least it makes no sense to search for less than one bed.
```suggestion
  const [filter, setFilter] = useState<FilterProps>({ beds: '1', pets: '0' })
```

api-client/src/endpoints/createShelterEndpoint.ts
f1sh1918: thats true, check last commit

web/src/components/InfiniteScrollList.tsx
klinzo, line 73: This works and is way more future proof/easier to understand:
``` typescript
  useEffect(() => () => {
    setData([])
    setError(null)
    setLoading(false)
    setHasMore(true)
    setPage(defaultPage)
  }, [loadPage, defaultPage])
```

klinzo (approved):
Tested on firefox. Nice job :)

This still looks kind of active, maybe you can ask nikola for some better color scheme/highligting?
image

web/src/components/InfiniteScrollList.tsx
klinzo: see suggestion below

translations/translations.json
klinzo: Never heard that name, but fine. I'd just have called it filter.

web/src/components/InfiniteScrollList.tsx
f1sh1918: fixed

web/src/routes/ShelterPage.tsx
f1sh1918: i would like to keep the placeholder in the input field --> rejected

translations/translations.json
f1sh1918: you can also rename it when we have checked ukrainian translations if you prefer 'filter'. i'm fine with that

klinzo: no its fine, just had never heard that. but has no benefit so lets just keep it

Taggotty (approved):
Tested with chrome. Would be nice if you could add a test for filtering though. Especially as currently there aren't any shelters created in the cms. Or am I missing something 🙈

f1sh1918:
> Tested with chrome. Would be nice if you could add a test for filtering though. Especially as currently there aren't any shelters created in the cms. Or am I missing something 🙈

we gonna create the tiles later. use link above. we won't add test, since this is only temp implementation

klinzo:
> > Tested with chrome. Would be nice if you could add a test for filtering though. Especially as currently there aren't any shelters created in the cms. Or am I missing something see_no_evil
> 
> we gonna create the tiles later. use link above. we won't add test, since this is only temp implementation

Wohnraum offer just not activated in the test cms yet. Should be available shortly though after Sven has activated it. Link above should still work.

klinzo:
> Tested with chrome. Would be nice if you could add a test for filtering though. Especially as currently there aren't any shelters created in the cms. Or am I missing something see_no_evil

@Taggotty offer tile should now also be available in test cms.

PR #628, opened 2022-03-15
IGAPP-931: Fix layout of aside when language change is opened

SteffiStuffel:


klinzo:
If @f1sh1918 didn't find another solution I think we should remove the sticky positioning.

SteffiStuffel:
Also we can make the margin-top bigger, so it is enough for expanded language menu.

klinzo:
> Also we can make the margin-top bigger, so it is enough for expanded language menu.

That sounds like a good solution!

klinzo (approved):
Good idea :+1: Tested on firefox

f1sh1918 (approved):

PR #633, opened 2022-03-21
IGAPP-963: Fix passing of build config to android builds and metro

klinzo:


klinzo:
@Cmd190 `replaceAll` should be available with a node version > v15, please check what version you have installed. Node v15 is required for the project anyway.

@Taggotty any reason why you split it in two commands? This does not work as the envs are not persisted across steps.
Also from the docs:
> The rule of thumb is: if you want to pass to cross-env a command that contains special shell characters that you want interpreted, then use cross-env-shell. Otherwise stick to cross-env.

SteffiStuffel (commented):
Can you please explain, what was the problem here and how you fixed it. I do not really understand whats going on here, therefore I can not really review.

Taggotty:
> @Cmd190 `replaceAll` should be available with a node version > v15, please check what version you have installed. Node v15 is required for the project anyway.
> 
> @Taggotty any reason why you split it in two commands? This does not work as the envs are not persisted across steps. Also from the docs:
> 
> > The rule of thumb is: if you want to pass to cross-env a command that contains special shell characters that you want interpreted, then use cross-env-shell. Otherwise stick to cross-env.

Yes in windows, cross-env was not able to spawn the follow up command. After splitting it in two it worked. Therefore, I didn't think more about it. But the first one must have had an impact, as it did not work with only the second. I'll test this one later this day on a windows PC, but I don't really get my hopes up, as I think this will not work. I am definitely fine with it, but @Cmd190 might not be please with this solution :D

klinzo:
> Can you please explain, what was the problem here and how you fixed it. I do not really understand whats going on here, therefore I can not really review.

So the problem was that android builds (and the metro packager) did not receive the build config we supplied. Therefore always the default (integreat for yarn android and integreat test cms for metro) was used. This leads to building integreat when you want to bulid malte, aschaffenburg... The reason why the build config was not supplied is that cross-env does not persist the envs across multiple commands. Therefore I merged the two commands (cross-env and the actual command) together.

Taggotty (approved):
Tested on windows and seems to work fine now.
Maybe the problem was that the second command was not prefixed with yarn? 

Previous implementation:
```typescript
    const command = `react-native run-android --no-packager --appId ${applicationId} ${productionFlag}`
    execSync(`yarn cross-env ${buildConfig} ${command}`, { stdio: 'inherit' })
```

klinzo:
> Tested on windows and seems to work fine now. Maybe the problem was that the second command was not prefixed with yarn?
> 
> Previous implementation:
> 
> ```ts
>     const command = `react-native run-android --no-packager --appId ${applicationId} ${productionFlag}`
>     execSync(`yarn cross-env ${buildConfig} ${command}`, { stdio: 'inherit' })
> ```

Nice! Perhaps the missing yarn was the reason yes. Hopefully all cases handled now.

SteffiStuffel (approved):

PR #640, opened 2022-03-30
IGAPP-966: Fix jira releases

klinzo:


tools/jira-release.ts
```diff
@@ -0,0 +1,138 @@
+import { program } from 'commander'
```
klinzo, line 1: Converted to typescript and converted manual oauth get/post/put with jira api calls.

klinzo:
I already tested this, should work.

.circleci/config.yml
```diff
@@ -490,8 +490,9 @@ jobs:
                         name: Bump git version
                         working_directory: tools
                     - run:
-                        command: tools/create-jira-release create-release ${NEW_VERSION_NAME} --private-key ${JIRA_PRIVATE_KEY} --access-token ${JIRA_OAUTH_TOKEN} --consumer-key ${JIRA_CONSUMER_KEY} --project-name ${CIRCLE_PROJECT_REPONAME}
+                        command: yarn jira-release create ${NEW_VERSION_NAME} --private-key ${JIRA_PRIVATE_KEY} --access-token ${JIRA_OAUTH_TOKEN} --consumer-key ${JIRA_CONSUMER_KEY} --project-name ${CIRCLE_PROJECT_REPONAME}
```
SteffiStuffel, line 493: create-jira-release would be a better naming.

SteffiStuffel (approved):
Looks good.

.circleci/config.yml
klinzo: Imo not otherwise you have `create-jira-release create` which is duplicated.

SteffiStuffel: you are right. overlooked it.

f1sh1918 (approved):

PR #644, opened 2022-04-01
IGAPP-960: Deliver hidden files and apple-app-site-association

klinzo:


build-configs/BuildConfigType.ts
```diff
@@ -10,14 +10,16 @@ import { ThemeType } from './ThemeType'
 // If you change this make sure you are not navigating to the landing screen upon closing the intro slides.
```
klinzo, line 10: Mostly just moved comments to the right place

klinzo:
Testable here test.integreat.app/apple-app-site-association and here https://test.integreat.app/.well-known/assetlinks.json

SteffiStuffel (approved):
Do not really get what you have done here in detail, but looked good.

klinzo:
> Do not really get what you have done here in detail, but looked good.

@SteffiStuffel two things:
1. With our previous delivery method using sftp hidden files and directories (i.e. those starting with a dot `.`) where not delivered/served. Therefore e.g. `.well-known/assetlinks.json` was not available and therefore we got the errors in google ads that our deep links are not correctly configured (see https://issues.tuerantuer.org/browse/IGAPP-857).
2. Previously apple-app-site-association was not part of our repo but hardcoded on the server side. This was imo bad for documentation/responsibility purposes and was also broken because of some changes there. So now the apple-app-site-association is also created and delivered from this repo together with the rest of the delivery.

PR #645, opened 2022-04-04
IGAPP-974: Unify viewport constants

Schedulaar:


web/src/constants/dimensions.ts
```diff
@@ -12,8 +13,9 @@ export type DimensionsType = {
 }
 
 const dimensions: DimensionsType = {
-  smallViewport: '(max-width: 750px)',
-  mediumViewport: '(min-width: 750px) and (max-width: 1100px)',
+  maxWidthViewportSmall: 768,
+  smallViewport: '(max-width: 768px',
```
klinzo, line 17:
```suggestion
  smallViewport: '(max-width: 768px)',
```

klinzo (commented):
:tada: Layout is currently completely broken, see suggestion below. With that applied everything  should work fine.

Could you perhaps add a release note? Like this: https://github.com/digitalfabrik/integreat-app/blob/main/release-notes/unreleased/IGAPP-927-use-image-alignment.yml

klinzo (changes requested):
see above

klinzo (approved):

f1sh1918 (approved):

PR #646, opened 2022-04-05
IGAPP-902: Refactor and fix tests

klinzo:


klinzo:
- [ ] Test native location handling/nearby cities

web/src/routes/PoisPage.tsx
```diff
@@ -55,39 +53,57 @@ const moveViewToBBox = (bBox: BBox, defaultVp: MapViewMercatorViewport): MapView
...
-  const [currentFeature, setCurrentFeature] = useState<PoiFeature | null>(null)
-  const mapRef = useRef<MapRef>(null)
+  const { t } = useTranslation('pois')
 
-  const flyToPoi = (coordinates: Position) => {
-    if (coordinates[0] && coordinates[1] && mapRef.current) {
+  const flyTo = useCallback((coordinates: Position) => {
+    if (!mapRef.current) {
+      // The map ref is not available on initial rendering, therefore retry after a short timeout
+      setTimeout(() => flyTo(coordinates), 1)
```
f1sh1918, line 72:
```suggestion
      setTimeout(() => flyTo(coordinates), 0)
```
is sufficient for the retry

web/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -1,31 +1,23 @@
...
-jest.mock('react-i18next', () => ({
-  useTranslation: () => ({
-    t: (key: string) => key
-  })
-}))
-
-// @ts-expect-error -- ignore readOnly var
-navigator.geolocation = mockGeolocationSuccess
+jest.mock('react-i18next')
+jest.mock('../../utils/getUserLocation', () => async () => ({ status: 'ready', coordinates: [10.8, 48.3] }))
```
f1sh1918, line 19: seems like this mock does not work, as the distance wasn't rendered if you `debug()`

web/src/routes/PoisPage.tsx
```diff
@@ -184,28 +181,29 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
     )
   }
 
+  const switchFeature = (step: number) => {
+    const featureIndex = featureLocations.findIndex(
+      (poi: PoiFeature) => poi.properties.urlSlug === currentFeature?.properties.urlSlug
+    )
+    const updatedIndex = updateFeatureIndex(step, featureLocations.length, featureIndex)
```
sarahsporck, line 188: Could you also update the `updateFeatureIndex` method? I think it is clearer if the step is typed literally as it can just be -1 or 1 otherwise the method does not work as expected. Also I'd suggest naming it nextFeatureIndex, as update inclines that something is changed in the method. :)

```typescript
    const nextFeatureIndex = (step: -1 | 1, arrayLength: number, currentIndex: number) { /* ... */ }
```

native/src/hooks/useUserLocation.ts
```diff
@@ -115,11 +110,8 @@ const useUserLocation = (useSettingsListener = false): LocationInformationType =
     return () => undefined
   })
 
-  const isReadyOrLoading = locationState.status === 'ready' || locationState.message === 'loading'
```
sarahsporck, line 118: This one was very sure responsible to avoid the location button in the corner to behave strangely. When you remove this and try to request the location it will always go back into the state of noPermission and then blink between searching and available. I think this should either be readded or you propose a better solution :)

sarahsporck (commented):
Not sure what made this happen, but the location button is blinking wildly in the app when:
1. Location Permission is granted
2. Move the map to a new feature location
3. Try to center the map at the current location by pressing the button in the corner

This issue does not always happen, but almost consistently when I move once and try to reset for the first time. Also I only tested this on the Emulator. Maybe this doesn't happen on a real device.

web/src/routes/PoisPage.tsx
```diff
@@ -78,27 +78,22 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
...
-    if (currentFeature) {
-      const coordinates = currentFeature.geometry.coordinates
-      flyTo(coordinates)
+    const currFeature =
+      featureLocations?.find((feature: PoiFeature) => feature.properties.urlSlug === selectedFeatureSlug) ?? null
+    if (currFeature) {
+      flyTo(currFeature.geometry.coordinates)
     }
-  }, [currentFeature, flyTo])
+    setCurrentFeature(currFeature)
```
klinzo, line 95: Why not just all this in `selectFeature`? You have the feature already there.

f1sh1918: because this `queryParams` hook has to be set with a timeout then to not intercept the setCurrentFeature behavior. So i separated the queryParams update from the currentFeature update.

native/src/hooks/useUserLocation.ts
f1sh1918: my gps position does not work for integreat app on real android device :(
Maybe someone else could test this. On Emulator i couldn't see an issue though after refactoring

web/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -51,30 +42,30 @@ describe('PoisPage', () => {
...
+      data: null,
+      loading: false,
+      error: new Error('Something went wrong'),
+      refresh: jest.fn()
+    }))
+    const { getByText } = renderPois()
     expect(getByText('error:unknownError')).toBeTruthy()
   })
-
-  it('should render poi details page when list item was clicked', () => {
```
sarahsporck, line 71: I reevaluated this test, to see whether we could somehow make it work. Therby I found, that flyTo doesn't work for me in web in edge. Furthermore, I now know that this test fails, as the MapRef is always null. But when trying to mock/spy it you'll break the searchparams. I think somehow this is a really tricky logic and we maybe need to refactor this on once more at a later point. For now I'd suggest to maybe revert the additional work that was put into this PR. Although I know you put quite some work into this.

native/src/hooks/useUserLocation.ts
klinzo: @sarahsporck please check again, I changed the implementation a little bit and tested this on native. For me everything works fine now.

klinzo:
@sarahsporck @f1sh1918 @SteffiStuffel I think I fixed all issues now here. If this works for you, you can approve both this PR and #464 . Then we can merge this one and then the other one in main.

web/src/components/MapView.tsx
```diff
@@ -71,9 +71,15 @@ const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<MapRef>):
 
   const { viewportSmall } = useWindowDimensions()
 
-  const onDeselectFeature = useCallback(() => {
-    selectFeature(null)
-  }, [selectFeature])
+  const onDeselectFeature = useCallback(
+    e => {
+      // Currently selected feature should not be deselected if the user clicks on the controls like zoom or user location
+      if (!e.target.className.includes('mapboxgl-ctrl-icon')) {
```
klinzo, line 77: Any better solutions are welcome

web/src/routes/PoisPage.tsx
```diff
@@ -55,39 +54,54 @@ const moveViewToBBox = (bBox: BBox, defaultVp: MapViewMercatorViewport): MapView
...
+    }
+    if (feature) {
+      queryParams.set(locationName, feature.properties.urlSlug)
+    } else {
+      queryParams.delete(locationName)
     }
+    setQueryParams(queryParams)
   }
 
+  const updateMapRef = useCallback(node => {
```
f1sh1918, line 82: please add comment

web/src/components/MapView.tsx
```diff
@@ -86,26 +71,27 @@ const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<MapRef>):
...
 
-  const onDeselectFeature = useCallback(() => {
-    setQueryLocation(null)
-    selectFeature(null)
-    updateQueryParams()
-  }, [selectFeature, setQueryLocation])
+  const onDeselectFeature = useCallback(
+    e => {
+      // Currently selected feature should not be deselected if the user clicks on the controls like zoom or user location
+      if (!e.target.className.includes('mapboxgl-ctrl-icon')) {
```
f1sh1918, line 77:
```suggestion
      if (e.target.className.includes('mapboxgl-canvas')) {
```
i think this is slightly better, because if we place f.e. another button or input field on the map, we would have to add classes to your includes fkt

web/src/routes/PoisPage.tsx
klinzo: will do in #464

web/src/components/MapView.tsx
klinzo: Good idea, will do in #464

PR #649, opened 2022-04-06
IGAPP-965: Ensure that inAppBrowser was closed before

f1sh1918:

Issue is really hard and random to reproduce. Couldn't reproduce after the fix. But i'm not 100% sure if it does the trick

Open IOS Emulator/ Device
1. GoTo Testumgebung --> Corona-Virus -> Information zum Corona Virus
2. open link --> swipe from left to right
3. open next link

### DIFF
--- a/release-notes/unreleased/IGAPP-965-inappbrowser-already-opened-error.yml
+++ b/release-notes/unreleased/IGAPP-965-inappbrowser-already-opened-error.yml
@@ -1,0 +1,6 @@
+issue_key: IGAPP-965
+show_in_stores: true
+platforms:
+  - ios
+en: Fix multiple inappbrowser opened.
+de: Fehler mit mehreren gleichzeitig geöffneten InAppBrowsern behoben.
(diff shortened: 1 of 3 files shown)


release-notes/unreleased/IGAPP-965-inappbrowser-already-opened-error.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-965
+show_in_stores: true
+platforms:
+  - ios
+en: Fix multiple inappbrowser opened.
+de: Fehler mit mehreren gleichzeitig geöffneten InAppBrowsern behoben.
```
klinzo, line 6: I think this error was not visible to users, was it? If so I'd change it to 
```suggestion
show_in_stores: false
platforms:
  - ios
en: Fix multiple inappbrowser opened.
```
and add a `not-testable` label to the JIRA issue.

klinzo (approved):

PR #650, opened 2022-04-08
IGAPP-394: Add proper a11y label for kebab menu

f1sh1918:

Check bug description in[ IGAPP-394](https://issues.tuerantuer.org/browse/IGAPP-394) for testing on android real device


SteffiStuffel (approved):
(opional): Add release note?
not tested.

translations/translations.json
```diff
@@ -349,6 +350,7 @@
       "loading": "Φόρτωση ..."
     },
     "en": {
+      "moreOptions": "More Options",
```
klinzo, line 353: Only german translations should be added
```suggestion
```

translations/translations.json
```diff
@@ -313,6 +313,7 @@
   },
   "common": {
     "de": {
+      "moreOptions": "Weitere Optionen",
```
klinzo, line 316: Did you have a look what accessibility label other apps are setting? Not sure if this is a good one.

f1sh1918: Mattermost has "Mehr Optionen" if you prefer that :D

klinzo: Both is fine :) Mehr is perhaps a little bit simpler language but your choice :)

klinzo (approved):

PR #651, opened 2022-04-08
IGAPP-971: Titles that are to long are not devided by new line

f1sh1918:

Test scenario:

1. Go to Testumgebung --> Bilder --> Scroll to "HierstehteinebesonderslangeÜberschrift"
2. Headline now should be broken into multiple lines
3. It breaks with the same behavior as in web (just to check same behavior) https://webnext.integreat.app/testumgebung/de/bilder


klinzo (approved):
Tested on android

SteffiStuffel (approved):
Tested. `&8203;` is not working, but not sure if this was the plan. I guess not.

klinzo:
> Tested. `&8203;` is not working, but not sure if this was the plan. I guess not.

It was not because only the corresponding html tag would work and we don't want cities to directly work with html tags, see https://chat.tuerantuer.org/digitalfabrik/pl/oc8usf59ft8nbpqsrfrqrfnanr

f1sh1918:
> Tested. `&8203;` is not working, but not sure if this was the plan. I guess not.

no this was not the issue in this ticket

PR #653, opened 2022-04-11
IGAPP-978: Cant click on some languages in language selection 

f1sh1918:

Just provided a quick fix to make language selection usable, maybe we find a better long-term solution

Check for testing here 
https://issues.tuerantuer.org/browse/IGAPP-978

Update: check last comment for test


SteffiStuffel:
Display is not correct, when removing min height of 80px.
small

web/src/components/Layout.tsx
```diff
@@ -80,7 +80,7 @@ const Aside = styled.aside<{ asideStickyTop: number }>`
   display: inline-block;
   position: sticky;
   width: ${dimensions.toolbarWidth}px;
-  padding-top: max(35%, 80px);
+  top: 35%;
```
klinzo, line 83: This is a duplicate delcaration as the following is set above:
```
top: ${props => props.asideStickyTop}px;
```

release-notes/unreleased/IGAPP-978-cant-click-on-some-languages-in-language-selection.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-978
+show_in_stores: true
+platforms:
+  - web
+en: Fix language selection.
+de: Fehler bei der Sprachauswahl behoben.
```
klinzo, line 6: Oc it does not hurt but there is no need to add german release notes for web

klinzo (approved):
With this fixed position we could remove the sticky altogether, couldn't we?

I can't reproduce the issue of @SteffiStuffel, tested on web

web/src/components/Layout.tsx
f1sh1918: yea i updated the pr due to we have to ensure to have at least a top positioninf of (headerHeightMin+LanguageSelectionHeight)

f1sh1918:
the question is if we want to calculate the height of the language selection bar. this could be a lot of effort.

> Display is not correct, when removing min height of 80px. small

yes i updated the top value to a the minHeight of languageSelectionBar + HeaderSmall.
I know the languageSelection bar can differ in height depending on items and screen size, but thats really hard to calc. so i think we should be fine with 2 rows

klinzo:
Still not working completely: 
image

f1sh1918:
I updated the pr to set positioning dynamically depending on selectorPanelHeight
Test: http://localhost:9000/deutschlandvorlage/de
Check video ;)

Uploading igapp_978.mov…

web/src/components/Layout.tsx
```diff
@@ -117,24 +117,25 @@ type PropsType = {
...
+const Layout = ({ footer, header, toolbar, modal, children, fullWidth = false }: PropsType): JSX.Element => {
   const modalVisible = !!modal
+  const { width } = useWindowDimensions()
+  const [languageSelectorHeight, setLanguageSelectorHeight] = useState<number>(0)
+
+  useEffect(() => {
+    const panelHeight = document.getElementById('languageSelector')?.clientHeight
+    if (panelHeight) {
+      setLanguageSelectorHeight(panelHeight)
+    }
```
klinzo, line 129: Shouldn't this rather be:
```suggestion
    setLanguageSelectorHeight(panelHeight ?? 0)
```

klinzo (approved):
Tested on firefox, really nice solution now!

PR #657, opened 2022-04-14
IGAPP-961: Poi details view web mobile

f1sh1918:

http://localhost:9000/testumgebung/de/locations?name=contact-sozialkaufhaus

1. Check back button on map
2. Test Collapsible (DetailView --> Information Section)
3. Test opening external link with mobile device should open maps app


web/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,21 @@
+import { LocationModel } from 'api-client'
+
+export const getNavigationDeepLinks = (location: LocationModel, fallback: string): string => {
+  if (!location.location || !location.coordinates) {
+    return fallback
+  }
```
klinzo, line 6: I think we should always use `geo` or `maps` uri scheme, even if there are no coordinates/location. If there is no location, it should be easy (just don't pass a query). If there are no coordinates, we can just pass `0,0` to search around the current location (https://en.wikipedia.org/wiki/Geo_URI_scheme#Unofficial_extensions). Not sure about iOS though, the `geo` scheme is not supported there, is it?

web/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,21 @@
+import { LocationModel } from 'api-client'
+
+export const getNavigationDeepLinks = (location: LocationModel, fallback: string): string => {
```
klinzo, line 3: Could we somehow pass `isIos` and `isAndroid` as params and move the functionality to the api client? We implement the same on native, don't we?

web/src/components/PoiDetails.tsx
```diff
@@ -119,34 +146,43 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({ feature, poi, selectFeature }:
...
+          {!viewportSmall && <Marker src={iconMarker} alt='' />}
+          <AddressContent>
+            <span>{location.address}</span>
+            <span>
+              {location.postcode} {location.town}
+            </span>
+          </AddressContent>
+        </AddressContentWrapper>
+        <LinkContainer>
+          <CleanLink to={getNavigationDeepLinks(location, `https://maps.google.com?q=${title}`)} newTab>
```
klinzo, line 176: I think the fallback link should include the full address (and coordinates?) as well if possible and be constructed in `getNavigationLinks` as well. Otherwise I think there may be e.g. too many matches for `Arbeitsagentur` or similar.

klinzo (changes requested):
Nice work, tested on firefox.
Things I noticed/may be changed:
* When selecting a poi on the map, I think we may open the bottom sheet to the top, otherwise the only new information instantly shown is the thumbnail and (if existing) the distance. Not sure if that makes sense and is easily doable though.
* If the bottom sheet is completely closed, I think it is weird that `In deiner Nähe` is still displayed, this is not the case on native as well, is it? Imo instead the bottom sheet title should not be shown

web/src/utils/getNavigationDeepLinks.ts
f1sh1918: No react native has some interna function like Plattform.select

web/src/components/MapView.tsx
```diff
@@ -74,7 +97,7 @@ const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<MapRef>):
   const onDeselectFeature = useCallback(
     e => {
       // Currently selected feature should not be deselected if the user clicks on the controls like zoom or user location
-      if (e.target.className.includes('mapboxgl-canvas')) {
+      if (e.target.classList.toString().includes('mapboxgl-canvas')) {
```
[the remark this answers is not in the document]
f1sh1918: Your Implementation doesn't work for All HTML elements like svgs

web/src/utils/getNavigationDeepLinks.ts
f1sh1918: `geo` is a newer scheme that is not yet supported by ios, so we have ti use different schemes.

The fallback is also used for desktop devices and we could use desktop maps apps for macOS/Linux/Windows (schema) but thats not very common.
I disagree with your solution referring to the link you posted, because the coordinates may lead to the correct address but not to the correct POI and people will be confused if another description will be presented.
I guess most people creating pois using sth like google maps to get address and names, therefore `location.location` should work.
You can test it by `return `geo:0,0?q=${lat},${long}`  for `isAndroid` and open `bfz gGmbh`
GoogleMaps: Show correct address with wrong description (several places at this address)
MapsMe: works with both schema solutions

web/src/constants/dimensions.ts
```diff
@@ -16,6 +17,7 @@ const dimensions: DimensionsType = {
   maxWidthViewportSmall: 768,
   smallViewport: '(max-width: 768px)',
   mediumViewport: '(min-width: 768px) and (max-width: 1100px)',
+  largeViewport: '(min-width: 768px)',
```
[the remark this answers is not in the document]
f1sh1918: indeed. i adjusted `min-width: 769px` and renamed variable

f1sh1918:
> 



> Nice work, tested on firefox. Things I noticed/may be changed:
> 
> * When selecting a poi on the map, I think we may open the bottom sheet to the top, otherwise the only new information instantly shown is the thumbnail and (if existing) the distance. Not sure if that makes sense and is easily doable though.
> * If the bottom sheet is completely closed, I think it is weird that `In deiner Nähe` is still displayed, this is not the case on native as well, is it? Imo instead the bottom sheet title should not be shown

1. the question is where is the top? also the header and navigation or only the map? If you cover the whole map you can not find the `backButton`on the map
2. i reduced the minSnappoint just to show the handler

web/src/utils/getNavigationDeepLinks.ts
f1sh1918: pois without coordinates and address gonna be filter in IGAPP-985, so i keep the current implementation

web/src/utils/getNavigationDeepLinks.ts
klinzo: you don't have to use it though, there is also `Platform.OS`: https://reactnative.dev/docs/platform#os
so you could just do the following:
``` ts
getNavigatonDeepLinks(location, fallback, Platform.OS === 'android', Platform.OS === 'ios')
```

web/src/components/PoiDetails.tsx
f1sh1918: i added the full location and coordinates for google maps desktop

web/src/constants/dimensions.ts
```diff
@@ -16,8 +16,8 @@ export type DimensionsType = {
 const dimensions: DimensionsType = {
   maxWidthViewportSmall: 768,
   smallViewport: '(max-width: 768px)',
-  mediumViewport: '(min-width: 768px) and (max-width: 1100px)',
-  largeViewport: '(min-width: 768px)',
+  mediumViewport: '(min-width: 769px) and (max-width: 1100px)',
+  mediumLargeViewport: '(min-width: 769px)',
   minMaxWidth: 'screen and (min-width: 1100px)',
   maxWidth: 1100,
```
klinzo, line 22: Could you adjust those two values as well please to `1101`?

web/src/utils/getSnapPoints.ts
```diff
@@ -1,4 +1,9 @@
-const min = 0.1
+const min = 0.05
 const mid = 0.35
 const max = 0.65
-export const getSnapPoints = (maxHeight: number): number[] => [maxHeight * min, maxHeight * mid, maxHeight * max]
+export const getSnapPoints = (maxHeight: number): number[] => [
```
klinzo, line 4: Maybe you can eliminate some non null checks with this:
```suggestion
export const getSnapPoints = (maxHeight: number): [number, number, number, number] => [
```

klinzo (approved):

web/src/components/MapView.tsx
```diff
@@ -140,7 +140,7 @@ const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<MapRef>):
         mapStyle={mapConfig.styleJSON}
         onClick={onSelectFeature}
         onTouchMove={() => changeSnapPoint(0)}>
-        {currentFeature && (
+        {currentFeature && viewportSmall && (
```
klinzo, line 143: Shouldn't we show this on medium screens as well (i.e. tablets)?

web/src/utils/getNavigationDeepLinks.ts
f1sh1918: but we are on web here?

klinzo: Created https://issues.tuerantuer.org/browse/IGAPP-986

web/src/components/MapView.tsx
```diff
@@ -88,7 +111,7 @@ const MapView = React.forwardRef((props: MapViewProps, ref: React.Ref<MapRef>):
       const feature = event.features && event.features[0]
       if (feature) {
         selectFeature(feature)
-        changeSnapPoint(1)
+        changeSnapPoint(2)
```
sarahsporck, line 114: Why did you change to snap point 2? Now the user cannot see the selected feature on the map anymore, as it is covered by the `bottomActionSheet`.

web/src/components/PoiDetails.tsx
```diff
@@ -119,34 +146,43 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({ feature, poi, selectFeature }:
...
+          <CleanLink to={getNavigationDeepLinks(location, title)} newTab>
+            {!viewportSmall && <LinkLabel>{t('detailsMapLink')}</LinkLabel>}
+            <Marker src={iconExternalLink} alt='' />
+          </CleanLink>
+        </LinkContainer>
+      </DetailSection>
+      <Spacer borderColor={theme.colors.poiBorderColor} />
+      <Collapsible title={t('detailsInformation')} initialCollapsed>
+        <RemoteContent html={content} onInternalLinkClick={navigate} smallText />
+      </Collapsible>
```
sarahsporck, line 185: I think it makes sense to hide the information collapsible if there are no information
```suggestion
      {content.length > 0 &&
      <>
        <Spacer borderColor={theme.colors.poiBorderColor} />
        <Collapsible title={t('detailsInformation')} initialCollapsed>
          <RemoteContent html={content} onInternalLinkClick={navigate} smallText />
        </Collapsible>
      </>
      }
```

web/src/constants/dimensions.ts
sarahsporck: could you help me why we would change it to 1101?

sarahsporck (approved):

web/src/components/MapView.tsx
klinzo: This was requested by me. I think this makes sense since this only happens if you select a poi on the map. And if you do that, you already saw where the poi is on the map. Nothing was changed for selecting a poi in the list/opening it via query param.

web/src/constants/dimensions.ts
klinzo: Because otherwise for a `1100` px width both medium and large screen media queries match which leads to undefined behavior

PR #658, opened 2022-04-18
IGAPP-979: Remove poi details route and other adjustments

klinzo:


native/src/components/BottomActionsSheet.tsx
```diff
@@ -22,6 +23,17 @@ const BottomActionsSheet: React.FC<BottomActionsSheetProps> = ({
 }: BottomActionsSheetProps): ReactElement | null => {
   const renderHandle = useCallback(props => <BottomSheetHandler title={title} {...props} />, [title])
 
+  const scrollRef = useRef<ScrollView>(null)
+
+  useEffect(() => {
```
SteffiStuffel, line 28: When does this happen?

native/src/components/PoiDetails.tsx
```diff
@@ -6,74 +6,66 @@ import styled from 'styled-components/native'
...
+  padding: 0 18px;
 `
 
-const TitleContainer = styled.View`
-  flex-direction: row;
-  border-style: solid;
-  border-color: ${props => props.theme.colors.textDisabledColor};
-  border-bottom-width: 1px;
+const StyledText = styled.Text`
+  font-size: 15px;
```
SteffiStuffel, line 39: Font sizes from theme did not match?

native/src/components/PoiDetails.tsx
```diff
@@ -6,74 +6,66 @@ import styled from 'styled-components/native'
...
 
-const PoiDetails: React.FC<PoiDetailsProps> = ({
-  poi,
-  feature,
-  detailPage,
-  navigateToPois,
-  language
-}: PoiDetailsProps): ReactElement => {
+const ContentWrapper = styled.View`
+  padding-right: 32px;
```
SteffiStuffel, line 55: Why so much magic number css in this file?

native/src/components/PoiDetails.tsx
```diff
@@ -89,32 +81,29 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
 
   return (
     <PoiDetailsContainer>
+      <StyledText>{title}</StyledText>
+      {distance && <StyledDistance>{t('distanceKilometre', { distance })}</StyledDistance>}
```
SteffiStuffel, line 85: Distance shows wrong values for me, not sure if related to this PR. Set emulator location to augsburg, but distances are over 9000km.

native/src/components/PoiDetails.tsx
```diff
@@ -89,32 +81,29 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
...
-          onPress={onNavigate}
-          icon='map'
-          onLongPress={location ? copyToClipboard(location) : undefined}
-          language={language}>
+      <HorizontalLine />
+      <PoiDetailItem onPress={onNavigate} icon={<ExternalLink source={ExternalLinkIcon} />} language={language}>
+        <TextWrapper onPress={location ? copyToClipboard(location) : undefined}>
           <Text>{address}</Text>
           <Text>
             {postcode} {town}
```
SteffiStuffel, line 92: For me there is one whitespace before postcode to much, not sure if intended.

SteffiStuffel (approved):
Tested on android, only did a rough review.

Mentioned few things I saw, but as Jira is down maybe some of it is not part of this ticket.

native/src/components/BottomActionsSheet.tsx
f1sh1918: When the title changed so when the poi changed in Detail view.

native/src/components/PoiDetails.tsx
f1sh1918: As far as I know we don't have spacing classes. What should I do?

SteffiStuffel: at least using more dynamic css would be good (depending on device size)

klinzo: I think we don't do this in integreat anywhere yet, do we? So no need to start it here.

native/src/components/PoiDetails.tsx
klinzo: Perhaps we mixed longitude and latitude again

klinzo: You sure? For me it works fine:

native/src/components/PoiDetails.tsx
f1sh1918: Yes indeed we should create some spacing classes with wp f.e. like in lunes. but since we don't have that yet, i would wait for design team and refactor all paddings when we have them

f1sh1918 (approved):

PR #661, opened 2022-04-26
IGAPP-991: Jumping safari bottombar when using poi bottomactionsheet

f1sh1918:

Only testable on ios real device with latest safari browser

Scroll up and down. BottomActionSheet shouldn't flicker nor move


web/src/components/__tests__/__snapshots__/Layout.spec.tsx.snap
```diff
@@ -9,6 +9,7 @@ exports[`Layout should render all components 1`] = `
       <header />
```
klinzo, line 9: Ughhh ugly snapshot :face_with_thermometer: We should just remove those, nobody is looking at them anyway...

web/src/components/Layout.tsx
```diff
@@ -110,14 +119,23 @@ type PropsType = {
...
  */
-const Layout = ({ footer, header, toolbar, modal, children, fullWidth = false }: PropsType): JSX.Element => {
+const Layout = ({
+  footer,
+  header,
+  toolbar,
+  modal,
+  children,
+  fullWidth = false,
+  scrollable = true
```
klinzo, line 137: Since setting this to false only for specific conditions this should be named differently, e.g. something like `disableScrollingSafari` or similar

klinzo (approved):
Not tested as I don't have a real iOS device here anymore. I trust you though/it is enough if tested by Martin later on.

web/src/components/Layout.tsx
```diff
@@ -29,14 +29,23 @@ const RichLayout = styled.div`
   }
 `
 
-const Body = styled.div<{ fullWidth: boolean }>`
+const Body = styled.div<{ fullWidth: boolean; disableScrollingSafari: boolean }>`
```
SteffiStuffel, line 32: hmh scrolling is disabled not only on Safari but on every browser, therefore I think the naming is not really matching :thinking:

SteffiStuffel (approved):
Cannot test, as my iPhone is to old to support latest Safari

web/src/components/Layout.tsx
f1sh1918: the style only applies in the styled component for safari in this case. we had a conversation before and i changed it from `disabledScrolling` to be more specific

PR #663, opened 2022-04-27
IGAPP-994: Implement rtl for pois route

f1sh1918:

http://localhost:9000/testumgebung/ar/locations

1. Test MapView Controls for RTL
2. Desktop: Check Panel: Icons & content in POI Detail View, ListItems in PoiList
3. Mobile: Check BottomActionSheet PoiDetails and PoiList


klinzo (commented):
~~I think lots of these adjustments should not be necessary as most of the things seems to be aligned perfectly well without the direction property (with the exception of the icons, i.e. back arrow, previous/next arrows, external link icon). Or am I missing something here?~~

web/src/components/PoiPanelNavigation.tsx
```diff
@@ -21,19 +23,28 @@ const Label = styled.span`
...
   width: 16px;
   height: 14px;
   flex-shrink: 0;
   padding: 0 8px;
   object-fit: contain;
   align-self: center;
+  ${props =>
+    props.direction === 'rtl' &&
+    css`
+      transform: rotate(180deg);
```
klinzo, line 36: Why not the following as well?
```suggestion
      transform: scaleX(-1);
```

web/src/components/PoiDetails.tsx
```diff
@@ -21,20 +22,32 @@ const DetailsContainer = styled.div`
...
   width: 16px;
   height: 14px;
   flex-shrink: 0;
   padding: 0 8px;
   object-fit: contain;
   align-self: center;
+
+  ${props =>
+    props.direction === 'rtl' &&
+    css`
```
klinzo, line 35: Maybe we could create a helper for this as this is used multiple times?

klinzo (approved):
Tested on firefox

SteffiStuffel:
Can not test currently, as there are now pois for no language

klinzo:
> Can not test currently, as there are now pois for no language

@SteffiStuffel you can use the production cms by entering `window.localStorage.setItem('api-url', 'https://cms.integreat-app.de')` in the console to test the pois.

sarahsporck:
Currently not testable as the cityModel doesn't have a bounding box. Maybe we should set a default bounding box around Germany if there is none.

SteffiStuffel (approved):
Looks good :tada: 
tested on chrome

PR #670, opened 2022-05-06
IGAPP-941: split regex and don't allow . at the end of the string

sarahsporck:

Tested here https://regex101.com/r/ZfeAMj/1
If we also don't want to allow trailing , . / @ ... it is enough to set \b at the end of the regex. Opinions?
I also split the regex for a better overview. If you disagree with splitting it I can combine it again and keep the split part as a comment.

### DIFF
--- a/api-client/src/replaceLinks.ts
+++ b/api-client/src/replaceLinks.ts
@@ -1,6 +1,10 @@
 // https://stackoverflow.com/a/3809435
-const regex =
-  /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|[-;:&=+$,\w]+@[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/g
+const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/g
+const baseUrl =  /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/g
+const path = /(?:\/(?:[+~%/.\w-]*[+~%/\w-]+)?)?/g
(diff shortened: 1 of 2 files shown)


api-client/src/replaceLinks.ts
```diff
@@ -1,6 +1,11 @@
 // https://stackoverflow.com/a/3809435
-const regex =
-  /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|[-;:&=+$,\w]+@[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/g
+const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/
+const baseUrl =  /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
```
klinzo, line 3:
```suggestion
const hostname =  /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
```

klinzo (approved):
Nice :+1: I like the splitting. I also think it makes sense to match word boundaries (\b) to avoid matching of other unwanted trailing characters.

sarahsporck:
> Nice :+1: I like the splitting. I also think it makes sense to match word boundaries (\b) to avoid matching of other unwanted trailing characters.

But keep in mond this includes the trailing slash, which was the main reason why I didn't use it right away.

SteffiStuffel (approved):
Tested on web: Dot at the end of the link is still displayed as part of the link, but works fine and opens url without link: http://localhost:9000/testumgebung/de/bilder

klinzo:
> But keep in mond this includes the trailing slash, which was the main reason why I didn't use it right away.

Okay yes, missed that. Then we should not do that, you are right. But maybe it makes sense to not match at least commas `,`.

PR #675, opened 2022-05-24
IGAPP-1016: Adjust structure of bounding box

f1sh1918:

BoundingBox of germany should be loaded as fallback


steffenkleinle (approved):

SteffiStuffel (approved):

steffenkleinle:
Closed since these api changes are not backwards compatible and therefore will break old app installations. This can be done in a possible api v4.

PR #676, opened 2022-05-26
IGAPP-980: City filter for shelter information

DaRuOrg:
- added city filter for shelter
- edited FacetInput.tsx for optional type
- added translations for place/zip -> tbd

https://issues.tuerantuer.org/browse/IGAPP-980


SteffiStuffel (commented):
I think the InputField should be hidden until it is working / supported by cms.

Prettier fails for ShelterFilterBar. Please run `yarn prettier --write .` to fix it.

I think we should - when this PR is approved - commit it to a feature branch of digitalfabrik /integrea-app repositiory, so we can run the pipeline or what do you @steffenkleinle suggest/prefer?

steffenkleinle:
> I think the InputField should be hidden until it is working / supported by cms.

:+1: 

> Prettier fails for ShelterFilterBar. Please run `yarn prettier --write .` to fix it.

Alternatively you can just use `yarn prettier:fix` or use the pre-commit hook/intellij setting.

> I think we should - when this PR is approved - commit it to a feature branch of digitalfabrik /integrea-app repositiory, so we can run the pipeline or what do you @steffenkleinle suggest/prefer?

Agreed :+1:

web/src/components/ShelterFilterBar.tsx
```diff
@@ -41,6 +42,15 @@ const ShelterFilterBar: React.FC<ShelterFilterBarProps> = ({
         updateSearchFilter={updateSearchFilter}
         name='beds'
       />
+      <FacetInput
+        value={filter.city ?? ''}
+        icon={cityIcon}
+        altTag={t('facetCity')}
+        placeholder={t('facetCity')}
+        updateSearchFilter={updateSearchFilter}
+        type='string'
```
steffenkleinle, line 51: I don't think `string` is a valid input type, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#input_types. I think the correct one would be `text`. To avoid this kind of mistakes you should add a more specific type to the `type` prop which only accepts valid input types.

steffenkleinle (changes requested):

steffenkleinle:
@DaRuOrg thank you for your first PR :tada: Looking quite good already only some minor things that should be fixed :)

web/src/components/FacetInput.tsx
```diff
@@ -54,7 +56,7 @@ const FacetInput: React.FC<FacetInputProps> = ({
       name={name}
       value={value}
       placeholder={placeholder}
-      type='number'
+      type={type || 'number'}
```
f1sh1918, line 59: i would suggest to set the `type` as required and pass it also for the "number" input field. Then you don't need the `||` operator

f1sh1918 (commented):
Nice work just :)

f1sh1918 (commented):
@DaRuOrg  can you point your PR on this branch
https://github.com/digitalfabrik/integreat-app/tree/IGAPP-980-CityFilter instead of main.

I added a comment about broken mobile mode here
https://github.com/digitalfabrik/integreat-app/pull/682#pullrequestreview-992096058

steffenkleinle:
Closing in favor of #682

PR #688, opened 2022-06-07
IGAPP-1004:  Hide disable previous next poi buttons if there is only one poi

f1sh1918:

http://localhost:9000/testumgebung/de/locations?name=berlin

### DIFF
--- a/web/src/components/PoisDesktop.tsx
+++ b/web/src/components/PoisDesktop.tsx
@@ -46,6 +46,7 @@ type PoisDesktopProps = {
   switchFeature: (step: 1 | -1) => void
   selectFeature: (feature: PoiFeature | null) => void
   direction: UiDirectionType
+  showNavigation: boolean
 }
 
 const PoisDesktop: React.FC<PoisDesktopProps> = ({
@@ -57,7 +58,8 @@ const PoisDesktop: React.FC<PoisDesktopProps> = ({
   poi,
   switchFeature,
   selectFeature,
-  direction
+  direction,
+  showNavigation
 }: PoisDesktopProps): ReactElement => {
   const { t } = useTranslation('pois')
 
@@ -72,7 +74,7 @@ const PoisDesktop: React.FC<PoisDesktopProps> = ({
             poiList
           )}
         </ListViewWrapper>
-        {currentFeature ? (
+        {currentFeature && showNavigation ? (
           <PoiPanelNavigation switchFeature={switchFeature} direction={direction} />
         ) : (
           <ToolbarContainer>{toolbar}</ToolbarContainer>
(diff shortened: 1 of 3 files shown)


SteffiStuffel (approved):
Cannot test, as there is always a next location :thinking: 

In the provided link http://localhost:9000/testumgebung/de/locations?name=berlin
there is a poi right next to it, and also with the next-button i get navigated to a poi in augsburg. Maybe my location is set wrong, so feel free to merge if it is working for you.

web/src/components/PoisDesktop.tsx
```diff
@@ -46,6 +46,7 @@ type PoisDesktopProps = {
   switchFeature: (step: 1 | -1) => void
   selectFeature: (feature: PoiFeature | null) => void
   direction: UiDirectionType
+  showNavigation: boolean
```
steffenkleinle, line 49: This term is really ambiguous but I don't really know what would be better. Could be avoided if you just passed the number of pois and did the check here. I think it would also make more sense from a logical perspective as the `PoisPage` should not be responsible to decide whether buttons should be shown on desktop or not. Currently I would just wonder when it would happen that we don't want to show the buttons.

web/src/components/PoiPanelNavigation.tsx
```diff
@@ -48,7 +48,7 @@ const PoiPanelNavigation: React.FC<PoiPanelNavigationProps> = ({
 }: PoiPanelNavigationProps): ReactElement => {
   const { t } = useTranslation('pois')
   return (
-    <NavigationContainer>
+    <NavigationContainer data-testid='poi-panel-navigation'>
```
steffenkleinle, line 51: I think you should rather add accessibility labels to the buttons and use those to test for this.

steffenkleinle (changes requested):

steffenkleinle:
Tested on firefox, works fine :)

This can easily be tested by adding the following above this line https://github.com/digitalfabrik/integreat-app/blob/1540e4a0a4b1c2e9857e599e2c59dcee53cb559d/api-client/src/endpoints/createPOIsEndpoint.ts#L24
```
          .slice(0, 1)
```
`

web/src/components/PoisDesktop.tsx
steffenkleinle: You can just pass the amount of features instead of the boolean. I am fine with keeping the boolean but it should have a self explanatory name. In the context of a map `navigation` just means something different for me.

web/src/components/PoiPanelNavigation.tsx
f1sh1918: you are right!

steffenkleinle: `poi-panel-navigation` does not really make sense as aria label, does it? :see_no_evil: I think you should add something like `Previous` and `Next` as aria label to the buttons on not the container div.

web/src/components/PoisDesktop.tsx
f1sh1918: you are right, renamed to `showFeatureSwitch`

steffenkleinle (approved):

PR #690, opened 2022-06-08
IGAPP-1023: Footer overlay for map view desktop

f1sh1918:

I adjusted the styles to the MapView Info on the right to be consistent. (confirmed by nikola)


Cmd190 (approved):
Tested on Windows .Looks way better than before

SteffiStuffel (commented):
This seems very confusing and complicated.

There is a prop overlay in Footer.tsx and LocationFooter.tsx -> makes sense 
There is a prop showFooter in LocationLayout.tsx, where we use LocationFooter -> makes sense

But why is there also a <LocationFooter> usage in PoisPage? So why is it needed in PoisPage and LocationLayout?
And why is it set in PoisPage and also passed to MapView?

Edit: Tested on chrome

f1sh1918:
> This seems very confusing and complicated.
> 
> There is a prop overlay in Footer.tsx and LocationFooter.tsx -> makes sense There is a prop showFooter in LocationLayout.tsx, where we use LocationFooter -> makes sense
> 
> But why is there also a usage in PoisPage? So why is it needed in PoisPage and LocationLayout? And why is it set in PoisPage and also passed to MapView?
> 
> Edit: Tested on chrome

@SteffiStuffel i'm not 100% sure if i understood you right.
We pass the footer to the MapView because we want to render it in this MapView Container with the overlay prop. 
Additionally we hide the normal footer.
I also thought about just styling footer as an overlay and positioning it absolute but then i would have to calculate the `left` positioning to not overlap with the side panel.

So what would be your proposal to simplify it?

release-notes/unreleased/IGAPP-1023-footer-overlay-for-map-view-desktop.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1023
+show_in_stores: true
+platforms:
+  - web
+en: Footer is now display as an overlay within the map.
+de: Der Footer wird nun als Overlay innerhalb der Karte angezeigt.
```
steffenkleinle, line 6: JFYI: The german release note is never shown/used for web only or not shown in the store release notes.

web/src/components/Footer.tsx
```diff
@@ -34,27 +40,23 @@ const FooterContainer = styled.footer`
...
-        {children}
-        {Footer.getVersion()}
-      </FooterContainer>
+      <span>
+        {__VERSION_NAME__}+{__COMMIT_SHA__}
+      </span>
     )
   }
+  return null
 }
```
steffenkleinle, line 53: Why did you extract this? Please inline :)

web/src/components/__tests__/__snapshots__/LocationFooter.spec.tsx.snap
```diff
@@ -1,7 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
steffenkleinle, line 1: It would be really nice if you could replace this snapshot test with a unit test. We said we want to gradually move away from those and replace them if they need to be adjusted.

web/src/routes/PoisPage.tsx
```diff
@@ -208,12 +221,12 @@ const PoisPage = ({ cityCode, languageCode, cityModel, pathname, languages }: Ci
       bboxViewport={moveViewToBBox(cityModel.boundingBox, defaultMercatorViewportConfig)}
       currentFeature={currentFeature}
       direction={direction}
+      overlayFooter={overlayFooter}
```
steffenkleinle, line 224: Why is the footer passed as prop and not just defined in `MapsView`?

steffenkleinle (commented):
On mobile the footer on the map is still shown but hidden behind the bottom sheet. What are the plans here?

steffenkleinle:
Tested on firefox, good work :) I think the current approach is fine

web/src/routes/PoisPage.tsx
f1sh1918: because i would have to pass `cityCode` and `language` as additional props

release-notes/unreleased/IGAPP-1023-footer-overlay-for-map-view-desktop.yml
f1sh1918: alright thx for info. removed

f1sh1918:
> On mobile the footer on the map is still shown but hidden behind the bottom sheet. What are the plans here?

there are two separate tickets. we need an overflowMenu where the footer infos gonna be included. One of these tickets is in NextUp

web/src/components/Footer.tsx
f1sh1918: moved inline

web/src/components/__tests__/__snapshots__/LocationFooter.spec.tsx.snap
f1sh1918: well the question is here, why we need here a test at all. We pass just children to this component and i think to test this has no value for me.
To test css properties for the overlay is not a good approach.
The only thing we could test is the version info which is only for development.
My suggestion is to remove the test to be honest

steffenkleinle: Fine as well. But I just think whenever a snapshot test is updated instead it should either be removed or replaced, the choice is up to you :)

web/src/routes/PoisPage.tsx
steffenkleinle: Would be the better solution imo as long as there are no plans to use the MapsView with a different footer. We only use react element props when we want the components to be used multiple similar ways.

web/src/components/__tests__/__snapshots__/LocationFooter.spec.tsx.snap
f1sh1918: rm test

steffenkleinle (approved):

PR #691, opened 2022-06-14
IGAPP-1022: and 1007: Fix navigation slider

SteffiStuffel:

### DIFF
--- a/native/src/components/AnchorIcon.tsx
+++ b/native/src/components/AnchorIcon.tsx
@@ -1,71 +1,42 @@
-import React, { ReactNode } from 'react'
+import React, { ReactElement } from 'react'
 import { ScrollView } from 'react-native'
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons'
 import styled from 'styled-components/native'
 
 import { isRTL } from '../constants/contentDirection'
 
-const anchorWidth = 30
-const widthThreshold = 10
-const Icon = styled(MaterialIcon)<{ width: number }>`
+const Icon = styled(MaterialIcon)`
   font-size: 30px;
-  width: ${props => props.width}px;
 `
+
 type PropsType = {
-  navigationItemWidth: number
   name: string
-  _scrollView: React.ElementRef<typeof ScrollView> | null | undefined
-  language: string
-  xPosition: number
-  contentSizeDiff: number
+  scrollViewRef: React.ElementRef<typeof ScrollView> | null
+  isLeftAnchor: boolean
 }
 
-class AnchorIcon extends React.Component<PropsType> {
-  onAnchorPress = (): void => {
-    const { navigationItemWidth, _scrollView, xPosition, contentSizeDiff } = this.props
-    const didReachLastItem = parseInt(xPosition.toFixed(0), 10) + widthThreshold > contentSizeDiff
-
-    // when everything is at starting point and xPosition is zero
-    if (_scrollView) {
-      if (!xPosition) {
-        _scrollView.scrollTo({
-          y: 0,
-          x: navigationItemWidth,
-          animated: true
-        }) // When we reach the last element
-      } else if (didReachLastItem) {
-        _scrollView.scrollTo({
-          y: 0,
-          x: -1,
-          animated: true
-        })
-      } else {
-        _scrollView.scrollTo({
-          y: 0,
-          x: xPosition + navigationItemWidth,
-          animated: true
-        })
-      }
+const AnchorIcon = ({ name, scrollViewRef, isLeftAnchor }: PropsType): ReactElement => {
+  const onAnchorPress = (): void => {
+    if (isLeftAnchor) {
+      scrollViewRef?.scrollTo({ x: 0, y: 0, animated: true })
+    } else {
+      scrollViewRef?.scrollToEnd({ animated: true })
     }
   }
 
-  render(): ReactNode {
-    const { name } = this.props
-    return (
-      <Icon
-        name={name}
-        style={{
-          transform: [
-            {
-              scaleX: isRTL() ? -1 : 1
-            }
-          ]
-        }}
-        onPress={this.onAnchorPress}
-        width={anchorWidth}
-      />
-    )
-  }
+  return (
+    <Icon
+      name={name}
+      style={{
+        transform: [
+          {
+            scaleX: isRTL() ? -1 : 1
+          }
+        ]
+      }}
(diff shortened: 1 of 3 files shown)


f1sh1918 (approved):
nice works. couldn't reproduce any issue on ios. please add a release note

Cmd190 (approved):
Tested on Android. The errors don't occur

steffenkleinle (changes requested):
Please add a release note.

Really nice refactoring and good work. However, one thing I think is not that good:
When I scrolled to the start/end of the list, the arrows are shown and clickable but do nothing. I think that is not really good. Possible solutions I'd like more:
1. don't show left arrow if already at the start and don't show right error if already at the end
2. right button click when already at the end goes back to the start and left button click when already at the start goes to the end

Another problem (which may be not that important since we have max. 4? tiles right now):
Clicking on the right button may skip tiles if there are more than the total number of tiles is more than twice as big as the number of tiles shown on the screen at once. It would therefore perhaps be better if a click on a button only scrolls to the next/previous tile.

steffenkleinle:
Looks good now. However, scrolling by swiping breaks the logic, the same button always stays disabled. If you swipe to the end, the left button is still disabled when it should be the right one that is disabled. When there is one item hidden on the left side and one on the right side, both buttons should be enabled.

SteffiStuffel:
> Looks good now. However, scrolling by swiping breaks the logic, the same button always stays disabled. If you swipe to the end, the left button is still disabled when it should be the right one that is disabled. When there is one item hidden on the left side and one on the right side, both buttons should be enabled.

Good catch. Fixed it.

native/src/components/AnchorIcon.tsx
```diff
@@ -5,22 +5,27 @@ import styled from 'styled-components/native'
...
+  updatePosition: (position: 'start' | 'end') => void
+  disabled: boolean
 }
 
-const AnchorIcon = ({ name, scrollViewRef, isLeftAnchor }: PropsType): ReactElement => {
+const AnchorIcon = ({ name, scrollViewRef, isLeftAnchor, updatePosition, disabled }: PropsType): ReactElement => {
   const onAnchorPress = (): void => {
     if (isLeftAnchor) {
       scrollViewRef?.scrollTo({ x: 0, y: 0, animated: true })
+      updatePosition('start')
```
steffenkleinle, line 25: as far as I can tell these calls are not needed, `scrollTo` triggers the `handleScroll` method already, therefore it should be possible to remove this function completely
```suggestion
```

steffenkleinle (approved):
Tested on android. I really like it, nice :tada:

native/src/components/AnchorIcon.tsx
SteffiStuffel: removed.

native/src/components/NavigationTiles.tsx
```diff
@@ -92,8 +88,7 @@ const NavigationTilesWithScrollableView = ({ tiles, theme }: PropsType): ReactEl
           name='keyboard-arrow-right'
           isLeftAnchor={false}
           scrollViewRef={scrollViewRef.current}
-          updatePosition={updatePosition}
-          disabled={currentPosition >= navigationItemWidth}
+          disabled={currentPosition >= lastItemAutoScroll * navigationItemWidth}
```
steffenkleinle, line 91: Instead you could probably just use `onMomentumScrollEnd` to listen for the current position of the scroll view, i.e. remove `onScroll` and pass `handleScroll` to `onMomentumScrollEnd`
``` ts
     /**
     * Fires when scroll view has finished moving
     */
    onMomentumScrollEnd?: ((event: NativeSyntheticEvent<NativeScrollEvent>) => void) | undefined;
```

SteffiStuffel: This does not work for clicking the arrows, as this is no momentum scrolling. I added it nevertheless, as disabling is working smoother with it.

steffenkleinle (changes requested):
Tested your current solution and sadly the right arrow is not behaving correctly as it is only clickable if not scrolled to the right at all.
I found a solution though:
``` ts
  const [percentageScrolled, setPercentageScrolled] = useState<number>(0)

  const handleScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
    const { contentSize, contentOffset, layoutMeasurement } = event.nativeEvent
    setPercentageScrolled(contentOffset.x / (contentSize.width - layoutMeasurement.width))
  }
```

Then you can use the following check for the left arrow
``` ts
          disabled={percentageScrolled === 0}
```
and this one for the right arrow
``` ts
          disabled={percentageScrolled === 1}
```

Edit: See #712

SteffiStuffel:
> > > Tested your current solution and sadly the right arrow is not behaving correctly as it is only clickable if not scrolled to the right at all. I found a solution though:
> > 
> > 
> > Can you please share, what is not working correctly? For me it is working just fine, see attached video.
> > scroll.mp4
> 
> This is just an issue if there are two or more items hidden, to test this just add one more navigation tile and swipe one item to the right. The solution I proposed in #712 should be more robust (especially since your threshold is just a guess).

As this is never the case, and there are no plans to add more tiles, rather there are plans to remove offers-tab. So my solution is just fine, no need to add extra code for no use case. There is no threshold that is guessed.

steffenkleinle:
I definitely see your concerns but let me please try to explain my reasoning for requesting changes and the proposal:
> As this is never the case and there are no plans to add more tiles

There are *currently* no plans to add more tiles, doesn't mean that there won't be more tiles in the future. Code should cover edge cases and not only work in one particular case when there are exactly four tiles. It may happen that the navigation tiles have less space and therefore less tiles can be shown or it may happen that we want to reuse this component for other use cases which has less space or more tiles. It also leaves out the possibility of smaller devices or ones with a high zoom level (e.g. for accessibility reasons). While some of these are more likely than others, I don't think it would hurt to make it more error proof.

> no need to add extra code for no use case

I don't think my proposal is relevantly more code (it even removes one line but of course, LOC are not very meaningful). There is one additional calculation but also the removal of one magic value constant (`const lastItemAutoScroll = 0.8`) and one callback function. In my personal opinion it is easier to understand but that is of course just a personal preference.

> There is no threshold that is guessed

Sorry, guessed was the wrong term but `lastItemAutoScroll` is at least a magic value that can be avoided with an easy calculation. Where does the value come from?

steffenkleinle (approved):

PR #692, opened 2022-06-14
IGAPP-993: Add multilingual screenshots

SteffiStuffel:

Farsi is not possible for ios. So i did not add these screenshots.

Not sure about naming of the ios files. Can someone please check?
Did not make any changes in the stores.

Language codes for google differ from docs: https://support.google.com/googleplay/android-developer/table/4419860?hl=de#query=ru
and store settings. For example Russian is once "ru" and once "ru-RU". Not sure which one to use, used the one from the store settings.


SteffiStuffel:

Farsi is not possible neither for android nor for ios. So i did not add these screenshots.

Not sure about naming of the ios files. Can someone please check?
Did not make any changes in the stores.

Cmd190 (approved):
I only looked at a few but I think this is fine

f1sh1918 (approved):
just checked ios with that table also couldn't find farsi
https://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html
Really difficult to find good knowledge sources
For me it looks fine, but lets wait for max/steffen

steffenkleinle (changes requested):
Please use consistent upper/lowercase naming. I'd also stick to our internal naming (dashboard, events, offers, ...).
Yes, there is no persian/farsi for apple, you can also have a look at our meta data maps here https://github.com/digitalfabrik/integreat-app/blob/1540e4a0a4b1c2e9857e599e2c59dcee53cb559d/tools/manage-metadata.ts#L218, relevant resources about available languages are linked there as well.

From https://docs.fastlane.tools/actions/deliver/#tips:
> [Starting March 20, 2019 Apple's App Store](https://developer.apple.com/news/?id=03202019a) requires 12.9-inch iPad Pro (3rd generation) [screenshots](https://help.apple.com/app-store-connect/#/devd274dd925) additionally to the iPad Pro 2nd generation screenshots. As fastlane historically uses the screenshot dimensions to determine the "display family" of a screenshot, this poses a problem as both use the same dimensions and are recognized as the same device family.

> To solve this a screenshot of a 12.9-inch iPad Pro (3rd generation) must contain either the string iPad Pro (12.9-inch) (3rd generation), IPAD_PRO_3GEN_129, or ipadPro129 (Apple's internal naming of the display family for the 3rd generation iPad Pro) in its filename to be assigned the correct display family and to be uploaded to the correct screenshot slot in your app's metadata.

So since only the dimensions are relevant for iOS (with the exception of 12.9 iPad Pro), so a similar naming as for android would be nice, but the screenshots should definitely be prefixed with `01` to `05` to avoid random ordering of the screenshots.
Maybe it is possible to use the same screenshots for both iPads on iOS.

steffenkleinle:
Looking good apart from the naming issues

SteffiStuffel:
Before starting to do this, i would:

1) Rename folders to the language codes in the code in the `manage-metadata.ts` file.
2) Copy iPad screenshots and rename one of them, so they have a string `IPAD_PRO_3GEN_129` in the filename. Will keep the name of the other iPad screenshots. I think duplicates are needed here.

I would not rename all ios files so the ordering number is at the beginning, because they have the number, which leads to the correct ordering and it would be a lot of effort for almost no benefit.

Is this fine for you and the changes you requested or am I missing something. Would really appreciate not to get change request and need to rename everything later again.

steffenkleinle:
I did all the renaming in #705, please have a look and let me know if that works for you :)

steffenkleinle (approved):

PR #693, opened 2022-06-14
IAGPP-1009: Fix address clipboard copy

SteffiStuffel:


Cmd190 (approved):
Tested on android, works fine

PR #694, opened 2022-06-21
IGAPP-986: Move getNavigationDeepLink to api-client

SteffiStuffel:


sarahsporck (approved):
Would you mind adding a small test for getNavigationDeepLinks for each Platform one maybe :)

SteffiStuffel:
Please have a look at both commits, these are two different versions:

Version 1: navigator.userAgent is called in /web/.. and then depending on the result `getNavigationDeepLink` is called with 'web' or 'android. 

Version 2: navigator.userAgent is called in /api-client. So it is easier to call the method and better design, as when using the method you don't have to think/know about the fact you have to check for isAndroid first, but we need to adjust our testing setup. Not sure if this is worth the effort.

Let me know which version you prefer.

SteffiStuffel:
> Would you mind adding a small test for getNavigationDeepLinks for each Platform one maybe :)

Added for android and ios, but couldn't get it working for web, as `navigator` is not defined.

f1sh1918:
> > Would you mind adding a small test for getNavigationDeepLinks for each Platform one maybe :)
> 
> Added for android and ios, but couldn't get it working for web, as `navigator` is not defined.

@SteffiStuffel should i investigate into this or are you still working on this?

api-client/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,29 @@
+import LocationModel from '../models/LocationModel'
+
+const getNavigationDeepLinks = (
```
steffenkleinle, line 3:
```suggestion
const getNavigationDeepLink = (
```
Optional and personal preference:
I think the term `navigation` is really misleading/ambivalent, I'd rather use something like
```suggestion
const getExternalMapsLink = (
```

api-client/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,29 @@
...
+  location: LocationModel,
+  platform: 'web' | 'android' | 'ios' | string
```
steffenkleinle, line 5:
```suggestion
  platform: 'web' | 'android' | 'ios'
```

api-client/src/utils/getNavigationDeepLinks.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  const long = location.coordinates[0]
+  const lat = location.coordinates[1]
+  const android = `geo:${lat},${long}?q=${location.location}`
+
+  switch (platform) {
+    case 'web': {
+      const baseUrl = `https://maps.google.com?q=`
+      const isAndroid = /Android/i.test(navigator.userAgent)
+      return isAndroid ? android : `${baseUrl}${location.location},${lat},${long}`
```
steffenkleinle, line 18: Wouldn't it be way easier if his check happened in the place where this code is called in web and then just pass `android` or `web` depending on the result?
I also think that would be the way I'd expect to use it. If I pass `web` as platform I'd not expect that there is internally a check for the user agent and I may get an android deep link instead of an url.

steffenkleinle, line 18: Why is no iOS link returned if the user agent matches an iOS device but instead just the gmaps url anyway?

api-client/src/utils/getNavigationDeepLinks.ts
steffenkleinle: Please make sure that the naming is always the same, currently it sometimes is plural and sometimes singular (e.g. in the tests).

steffenkleinle (changes requested):

api-client/src/utils/getNavigationDeepLinks.ts
steffenkleinle: fine for me :)

steffenkleinle:
test name is still `getExternalDeepLink`

api-client/src/utils/getNavigationDeepLinks.ts
SteffiStuffel: Fine for me, will change.

SteffiStuffel: As in my first comment I opened discussion for this solution: https://github.com/digitalfabrik/integreat-app/pull/694#issuecomment-1162834448

steffenkleinle (approved):

PR #695, opened 2022-06-21
IGAPP-986

SteffiStuffel:

@sarahsporck


sarahsporck:
So there are multiple possible fixes for the failing native tests. The most obvious one being to add lib: ["dom"] to the native `tsconfig.json`. When looking at https://github.com/tsconfig/bases this could actually make sense, although I think it seems a little bit overkill, as we'd just need the dom types for this specific case right now.
Another solution would be to create a testing tsconfig whith the lib setting adjusted which extends the general native tsconfig. This also seems a little bit unnecessary imo. And if you wonder it is not possible right now to set specific ts options for ts-jest. Read more about this here https://github.com/kulshekhar/ts-jest/issues/2869.
And last but not least you can just add 
```typescript
// @ts-expect-error
const isAndroid = /Android/i.test(navigator.userAgent)
```
which will disable all typechecking for this line. Also not optimal, but currently probably the simplest solution.

PR #696, opened 2022-06-21
IGAPP-1006: pois do not center

f1sh1918:

Click on poi list item and minimize Bottom Action Sheet to check that poi centers now on the map

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -1,6 +1,6 @@
 import MapboxGL, { CameraSettings, MapboxGLEvent, SymbolLayerProps } from '@react-native-mapbox-gl/maps'
 import type { BBox, Feature } from 'geojson'
-import React, { ReactElement, useCallback, useState } from 'react'
+import React, { MutableRefObject, ReactElement, useCallback, useState } from 'react'
 import { FAB } from 'react-native-elements'
 import { useTheme } from 'styled-components'
 import styled from 'styled-components/native'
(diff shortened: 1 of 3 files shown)


release-notes/unreleased/IGAPP-1006-pois-do-not-center.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-1006
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: Poi centers on the map by clicking on a location in the list.
+de: Poi wird auf der Karte zentriert, wenn ein Ort aus der Liste ausgewählt wird.
```
steffenkleinle, line 7:
```suggestion
en: The map is now centered to the selected location.
de: Die Karte wird nun auf den ausgewählten Ort zentriert.
```

native/src/components/MapView.tsx
```diff
@@ -37,126 +37,126 @@ const featureLayerId = 'point'
...
-        ['==', ['get', 'id'], selectedFeature?.properties.id ?? -1],
-        mapMarker.symbolActive,
-        ['get', 'symbol']
-      ],
-      textField: ['get', 'title'],
-      textFont: ['Roboto Regular'],
-      textOffset: [0, textOffsetY],
-      textAnchor: 'top',
-      textSize: 12
+const MapView = React.forwardRef(
```
steffenkleinle, line 40: :/ Seems like I have to read up on these forward refs because currently I don't get what it is doing and why it is needed...

steffenkleinle, line 40: Am I correct that the only thing changed here is the wrapping with `forwardRef`? Why is it needed to call `cameraRef.flyTo` in the parent component?

steffenkleinle (commented):
Tested on anroid :+1: 
While it centers around the selected poi, the zoom is not changed. E.g. in the testumgebung it still shows nearly the whole of germany after selecting a poi, I think this is different in web. Any reason for not implementing this here?

native/src/components/MapView.tsx
f1sh1918: React forwardRef is a method that allows parent components pass down (i.e., “forward”) refs to their children.

f1sh1918: Well we want to use the `selectPoiFeature` in the MapView and ListItem. This method uses the cameraRef that has to be set in MapView(Camera) to gain control to the Camera Component. 
So i think its the easiest way to pass the `selectedPoiFeature` fkt from POIS to the child component

f1sh1918:
> Tested on anroid 👍 While it centers around the selected poi, the zoom is not changed. E.g. in the testumgebung it still shows nearly the whole of germany after selecting a poi, I think this is different in web. Any reason for not implementing this here?

You are right. I added the zoom in my last commit. Please check again

SteffiStuffel (approved):
Works fine on android.

PR #700, opened 2022-06-24
IGAPP-1033: Show contact information for pois

steffenkleinle:


SteffiStuffel (approved):
Nice PR. Looks good. 

Why have you moved icons from "shelter" subfolder?

Tested on Android emulator, looks nice.

steffenkleinle:
> Why have you moved icons from "shelter" subfolder?

Because previously those where only used in web and are now used in native as well. Apart from that they are now not exclusively used for the shelter offer but also for pois, so I thought it makes no sense anymore to keep those in a shelter subfolder. Sooner or later it makes sense anyway to use the same icons on native and web (see https://issues.tuerantuer.org/browse/IGAPP-921).

f1sh1918 (changes requested):
Pressable Text is not aligning center to the icon

image

I know it looks like in the design. But i think it should be center aligned vertically

native/src/components/PoiDetails.tsx
```diff
@@ -81,19 +91,59 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({ poi, feature, language }: PoiDe
     }
   }
 
+  const contactInformationCollapsibleItem = (
+    <CollapsibleItem initExpanded headerText={t('contactInformation')} language={language}>
+      <ContentWrapper>
+        {website && (
+          <Row>
```
f1sh1918, line 98: wouldn't it make sense to use a reusable component here?

native/src/components/PoiDetailRow.tsx
```diff
@@ -0,0 +1,36 @@
...
+  flex-direction: row;
+  padding-vertical: 3px;
+`
+
+const Icon = styled.Image`
+  margin-horizontal: 5px;
+`
+
+const StyledPressable = styled.Pressable`
+  alignself: center;
```
f1sh1918, line 17: `align-self: center;`

native/src/components/PoiDetails.tsx
```diff
@@ -55,7 +58,14 @@ const ContentWrapper = styled.View`
   padding-right: 32px;
```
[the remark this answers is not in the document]
f1sh1918: you can set `padding-left` if its rtl. or just create a bug ticket, since its not your issue

f1sh1918 (changes requested):
please adjust align attrib

native/src/components/PoiDetailRow.tsx
steffenkleinle: no idea how that happened... good catch, thx

native/src/components/PoiDetails.tsx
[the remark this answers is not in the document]
steffenkleinle: https://issues.tuerantuer.org/browse/IGAPP-1038

f1sh1918 (approved):

PR #701, opened 2022-06-25
IGAPP-985: Adjust poi location model and require coordinates

steffenkleinle:


api-client/src/endpoints/createPOIsEndpoint.ts
```diff
@@ -15,39 +15,36 @@ type ParamsType = {
 export default (baseUrl: string): Endpoint<ParamsType, Array<PoiModel>> =>
   new EndpointBuilder<ParamsType, Array<PoiModel>>(POIS_ENDPOINT_NAME)
     .withParamsToUrlMapper(
-      (params: ParamsType): string => `${baseUrl}/${params.city}/${params.language}/wp-json/extensions/v3/locations`
+      (params: ParamsType): string =>
+        `${baseUrl}/${params.city}/${params.language}/wp-json/extensions/v3/locations?on_map=1`
```
f1sh1918, line 19: what does the param stand for `on_map` ?

api-client/src/models/LocationModel.ts
```diff
@@ -1,51 +1,40 @@
-import { Position } from 'geojson'
-
-class LocationModel {
+class LocationModel<T> {
```
f1sh1918, line 1: do we expect different types than number for long/lat or why do we need a generic here. Maybe i overseen sth

f1sh1918 (approved):
Tested on android works fine

api-client/src/models/LocationModel.ts
steffenkleinle: long/lat may be null for the locations of events. therefore the generic should only ever either be `number` or `number | null`. To avoid type checks for pois I introduced this generic such that it is guaranteed that pois always have long/lat defined while it is not guaranteed for events.

api-client/src/endpoints/createPOIsEndpoint.ts
steffenkleinle: only pois that should be shown on the map (we don't need anything else). Only for these pois it is guaranteed that long/lat are not null.

SteffiStuffel (approved):
Tested on chrome. Not sure if in the overview of events, we really want to display the address that prominent.

event

steffenkleinle:
> Tested on chrome. Not sure if in the overview of events, we really want to display the address that prominent.
> 
> event

This is just as it was before. Discussion about this topic is taking place here: https://chat.tuerantuer.org/digitalfabrik/pl/owc96caitjnrbmu8sgj38zgkpr

PR #704, opened 2022-06-27
IGAPP-1030:  Remove query_all_packages permission

f1sh1918:

Since the in-app-browser lib doesn't release frequently, we can not expect to get the package permission removed in time.
Therefore i created a patch file to remove the permission in the android manifest.
I'm gonna create an additional ticket to remove this patch when the library was updated

Testing:
- `yarn android`
-check `android/app/build/intermediates/merged_manifest/AndroidManifest.xml` for the `used-permission query_all_packages` it should be removed now
- check if in-app-browser is still working

Tested on android real device

### DIFF
--- a/native/patches/react-native-inappbrowser-reborn+3.6.3.patch
+++ b/native/patches/react-native-inappbrowser-reborn+3.6.3.patch
@@ -1,0 +1,3717 @@
+diff --git a/node_modules/react-native-inappbrowser-reborn/android/build/.transforms/bbb1612691222624429793ba28463d88/results.bin b/node_modules/react-native-inappbrowser-reborn/android/build/.transforms/bbb1612691222624429793ba28463d88/results.bin
+new file mode 100644
+index 0000000..0d259dd
+--- /dev/null
++++ b/node_modules/react-native-inappbrowser-reborn/android/build/.transforms/bbb1612691222624429793ba28463d88/results.bin
+@@ -0,0 +1 @@
++o/classes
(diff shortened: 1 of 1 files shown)


steffenkleinle (commented):
I still have QUERY_ALL_PACKAGES in AndroidManifest, even after deleting build folder and running yarn install. What am I missing?

SteffiStuffel:
> I still have QUERY_ALL_PACKAGES in AndroidManifest, even after deleting build folder and running yarn install. What am I missing? It is in the subfolder debug though (android/app/build/intermediates/merged_manifest/**debug**/AndroidManifest.xml)

Same for me. Query all packages is still there.

f1sh1918:
@steffenkleinle  & @SteffiStuffel 
i had to add some packages for yarn to execute the patch and a postinstall script. Now it should work fine.
I also added instruction for removing this in the follow up ticket
https://www.npmjs.com/package/patch-package
I'm not 100% sure about this
```js
You can use --dev if you don't need to run yarn in production, e.g. if you're making a web frontend.
```
Is it fine for a native app too? Dunno if its okay to keep them in dev deps

steffenkleinle:
> @steffenkleinle & @SteffiStuffel i had to add some packages for yarn to execute the patch and a postinstall script. Now it should work fine. I also added instruction for removing this in the follow up ticket https://www.npmjs.com/package/patch-package I'm not 100% sure about this
> 
> ```js
> You can use --dev if you don't need to run yarn in production, e.g. if you're making a web frontend.
> ```
> 
> Is it fine for a native app too? Dunno if its okay to keep them in dev deps

Installing it as dev dependency should definitely be enough since it is only needed for the postinstall script

steffenkleinle:
Just tested this with a production build and the permission is not included and the inappbrowser still works :+1:

steffenkleinle:
I guess there will not be an update in time. Therefore lets merge it :)

PR #705, opened 2022-06-28
IGAPP-993: Renaming and german and ipad pro129 screenshots

steffenkleinle:


steffenkleinle:
@SteffiStuffel any reason why the german android screenshots were not in images/phoneScreenshots?

PR #708, opened 2022-06-28
IGAPP-1033 Show Contact Info in POIs

LeandraH:
It starts with the keyword **IGAPP-1033**.

### DIFF
--- a/web/src/components/PoiDetails.tsx
+++ b/web/src/components/PoiDetails.tsx
@@ -19,6 +19,7 @@ import CleanLink from './CleanLink'
 import Collapsible from './Collapsible'
 import RemoteContent from './RemoteContent'
 import Spacer from './Spacer'
+import ContactItem from './ContactItem'
 
 const DetailsContainer = styled.div`
   font-family: ${props => props.theme.fonts.web.contentFont};
@@ -141,17 +142,6 @@ const DetailSection = styled.div`
   }
 `
 
-const ContactContainer = styled.div`
-  display: flex;
-  align-items: center;
-  padding-top: 4px;
-  font-size: clamp(0.55rem, 1.6vh, ${props => props.theme.fonts.hintFontSize});
-
-  @media ${dimensions.smallViewport} {
-    gap: 8px;
-  }
-`
-
 type PoiDetailsProps = {
   feature: PoiFeature
   poi: PoiModel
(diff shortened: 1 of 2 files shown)


web/src/components/PoiDetails.tsx
```diff
@@ -202,6 +216,39 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
...
+        <>
+          <Spacer borderColor={theme.colors.poiBorderColor} />
+          <Collapsible title={t('contactInformation')} initialCollapsed direction={direction}>
+            <>
+              {website && (
+                <ContactContainer>
+                  <Marker src={iconWebsite} alt={t('website')} />
+                  <a href={website}>
+                    {website}
+                  </a>
```
steffenkleinle, line 229: Our `CleanLink` component should be used here instead of plain `a` tags. Also, the marker should be included in the link. If you want, you can adjust this here and for the other two usages, otherwise we'll just do that!

steffenkleinle (commented):
Thank you for your first PR :tada: Great work :+1:

Tested on firefox, works perfectly!

web/src/components/PoiDetails.tsx
```diff
@@ -202,6 +216,39 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
           </LinkContainer>
         )}
       </DetailSection>
+      {(website || phoneNumber || email) && (
+        <>
+          <Spacer borderColor={theme.colors.poiBorderColor} />
+          <Collapsible title={t('contactInformation')} initialCollapsed direction={direction}>
+            <>
+              {website && (
+                <ContactContainer>
```
f1sh1918, line 225: Could be extracted to component like `ContactItem`

web/src/components/PoiDetails.tsx
```diff
@@ -202,6 +216,39 @@ const PoiDetails: React.FC<PoiDetailsProps> = ({
...
+                  <Marker src={iconWebsite} alt={t('website')} />
+                  <a href={website}>
```
f1sh1918, line 227: Since it was not defined clearly in the requirements, its just optional. But i think the whole item including icon should be clickable

f1sh1918 (commented):
Tested on Chrome works fine. Nice works. Comments are just minor optional issues

web/src/components/PoiDetails.tsx
LeandraH: I thought so too but then I realised that the icon is not clickable in the app version. I think if we change it here, we should also change it in the app?

LeandraH:
I extracted the lines into a new component called `ContactItem` and used the `CleanLink` component. I did decide against making the marker part of the link because it isn't in the app version, and I think that should be consistent. We could change it in both places, though :)

steffenkleinle:
> I extracted the lines into a new component called `ContactItem` and used the `CleanLink` component. I did decide against making the marker part of the link because it isn't in the app version, and I think that should be consistent. We could change it in both places, though :)

Thanks again! You are completely right, I missed this in my PR for native so my bad. Will create a separate issue for this :)

steffenkleinle:
Closing in favor of #711

PR #714, opened 2022-07-04
IGAPP-263: Deep link to news item on push notification click

steffenkleinle:


native/src/utils/PushNotificationsManager.ts
```diff
@@ -65,26 +73,28 @@ export const subscribeNews = async (city: string, language: string): Promise<voi
...
+  })
+}
+
 export const quitAppStatePushNotificationListener = async (
   dispatch: Dispatch,
   navigation: NavigationPropType<RoutesType>
 ): Promise<void> => {
   const messaging = await importFirebaseMessaging()
-  const message = await messaging().getInitialNotification()
+  const message = (await messaging().getInitialNotification()) as Message | null
```
SteffiStuffel, line 92: So `message` is null, when data has the wrong format?

native/src/utils/PushNotificationsManager.ts
```diff
@@ -8,9 +8,17 @@ import { NavigationPropType, RoutesType } from '../constants/NavigationTypes'
...
 import navigateToDeepLink from '../navigation/navigateToDeepLink'
 import urlFromRouteInformation from '../navigation/url'
-import appSettings from './AppSettings'
 import { log, reportError } from './sentry'
 
+type Message = FirebaseMessagingTypes.RemoteMessage & {
+  data: {
+    city_code: string
+    language_code: string
+    group: string
```
SteffiStuffel, line 17: group is `news` or `tünews`? Should we maybe make this type more strict?

release-notes/unreleased/IGAPP-263-open-news-item.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-263
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: Clicking on push notifications now opens the correct news.
+de: Beim Anklicken von Push Notifications wird nun die richtige Nachricht geöffnet.
```
SteffiStuffel, line 7: (minor/optional) both english and german sound a little like before this updated, the wrong news was opened.

SteffiStuffel (approved):
Cannot test, as current main, which fixes the app, is not yet merged into this branch.

native/src/utils/PushNotificationsManager.ts
steffenkleinle: `group` is something like `testumgbung-de-news` and not used. Will therefore remove this to avoid confusion.

PR #715, opened 2022-07-04
IGAPP-1049: Remove hash property

steffenkleinle:


SteffiStuffel (approved):
Looks good. Please write into the ticket for testing, whole app should be checked.

PR #719, opened 2022-07-06
IGAPP-1051: Add react-native-url-polyfill to fix broken poi urls

steffenkleinle:


SteffiStuffel (approved):
Could reproduce the error on main, but not on this branch.

PR #724, opened 2022-07-12
IGAPP-892: Remove large icon of parent category

charludo:


native/src/components/CategoryList.tsx
```diff
@@ -54,8 +46,7 @@ const CategoryList = ({
   listContent,
   query,
   onItemPress,
-  language,
-  thumbnail
```
steffenkleinle, line 58: If a prop is not used anymore at all it should also be removed from the prop type and not be passed to the component anymore.

steffenkleinle (commented):
Tested on android :+1:

steffenkleinle:
Minor: Please include the issue key in the commit message, e.g. `IGAPP-892: Remove category icon in list view` or similar.

PR #726, opened 2022-07-12
IGAPP-1047: Include icon in POI links

charludo:


native/src/components/PoiDetailRow.tsx
```diff
@@ -26,9 +26,11 @@ type Props = {
 
 const PoiDetailRow = ({ externalUrl, text, accessibilityLabel, icon }: Props): ReactElement => (
   <Row>
-    <Icon source={icon} accessibilityLabel={accessibilityLabel} />
     <StyledPressable onPress={() => openExternalUrl(externalUrl)}>
```
steffenkleinle, line 29: Would it be possible to remove one of the `Row` or `StyledPressable` components now and apply all the styling there?

steffenkleinle (changes requested):
Tested and works fine on firefox :+1: 
On native the styling is now a bit off, maybe you can try to get the same styling as before? Tip: No additional styling is needed, you just need to apply it to different components that are already used.

web/src/components/ContactItem.tsx
```diff
@@ -39,10 +39,12 @@ const ContactItem: React.FC<ContactItemProps> = ({
...
-  <ContactContainer>
-    <Marker src={iconSrc} alt={iconAlt} />
-    <CleanLink to={link}>{content}</CleanLink>
-  </ContactContainer>
+  <CleanLink to={link}>
+    <ContactContainer>
+      <Marker src={iconSrc} alt={iconAlt} />
+      {content}
+    </ContactContainer>
+  </CleanLink>
```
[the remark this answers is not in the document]
charludo: Good point. I moved the corresponding styles to the `StyledCleanLink` component, since it's not used anywhere else. However, the theme is not passed as a prop to `StyledCleanLink` in tests, for a reason unknown to me. (Result in the browser looks to be OK though!)

Would appreciate some feedback!

web/src/components/CleanLink.tsx
```diff
@@ -1,13 +1,21 @@
...
   color: inherit;
   text-decoration: none;
   display: flex;
+  align-items: center;
+  padding-top: 4px;
+  font-size: clamp(0.55rem, 1.6vh, ${props => props.theme.fonts.hintFontSize});
+
+  @media ${dimensions.smallViewport} {
+    gap: 8px;
+  }
```
steffenkleinle, line 18: This should not be applied here because it would change the styles for all usages. Instead you can do ```
``` ts
const Link = styled(CleanLink)`
... styles
```

native/src/components/PoiDetailRow.tsx
```diff
@@ -4,17 +4,14 @@ import styled from 'styled-components/native'
...
-  padding-vertical: 3px;
-`
-
 const Icon = styled.Image`
   margin-horizontal: 5px;
 `
 
 const StyledPressable = styled.Pressable`
-  align-self: center;
+  align-self: flex-start;
```
steffenkleinle, line 12: Not tested if this works as well but you could remove this and add `align-self: center` as style to the Text component.

web/src/components/CleanLink.tsx
charludo: True - sorry about that. Moved the styles back to their own component. The issue I previously mentioned about the theme prop not being accessible was also resolved by this.

web/src/components/ContactItem.tsx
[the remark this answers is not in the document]
charludo: Resolved, see below

steffenkleinle (approved):
Tested on android and firefox

PR #728, opened 2022-07-12
IGAPP-1008: Upgrade dependencies

steffenkleinle:


web/stylelint.config.js
```diff
@@ -1,107 +1,11 @@
 /* eslint-disable no-magic-numbers */
 
 module.exports = {
-  processors: ['stylelint-processor-styled-components'],
-  plugins: ['stylelint-order'],
-  extends: ['stylelint-config-styled-components'],
-  syntax: 'scss',
+  extends: ['stylelint-config-recommended'],
+  customSyntax: 'postcss-jsx',
   rules: {
```
steffenkleinle, line 6: @f1sh1918 any specific rules you'd like to keep? I figured it makes more sense to extend from a recommended rule config than to define all rules ourselves.

.circleci/src/jobs/check.yml
```diff
@@ -25,8 +25,12 @@ steps:
...
   - unit_test:
       directory: api-client
-  - unit_test:
-      directory: native
+  # TODO IGAPP-1070: Use unit_test command again with coverage
+  - run:
+      name: Unit Tests with Coverage
+      command: yarn workspace native run test:ci --maxWorkers ${TOTAL_CPUS}
+  - store_test_results:
+      path: << parameters.directory >>/reports/unit-test
```
steffenkleinle, line 33: I was not able to get jest to work with the `--coverage` flag. Seems like it produces invalid code somehow.

web/src/components/CityNotCooperatingFooter.tsx
```diff
@@ -39,11 +39,11 @@ type PropsType = {
   languageCode: string
 }
 
-const CityNotCooperatingFooter = ({ languageCode }: PropsType): ReactElement => {
+const CityNotCooperatingFooter = ({ languageCode }: PropsType): ReactElement | null => {
   const { t } = useTranslation('landing')
 
   if (!buildConfig().featureFlags.cityNotCooperating) {
-    return <></>
+    return null
```
SteffiStuffel, line 46: Why do you prefer "null" over "<>"?

web/src/components/Footer.tsx
```diff
@@ -12,18 +12,14 @@ const FooterContainer = styled.footer<{ overlay: boolean }>`
...
+  padding: ${props => (props.overlay ? '0 0 0 10px' : '15px 5px')};
   background-color: ${props => (props.overlay ? `rgba(255, 255, 255, 0.5)` : props.theme.colors.backgroundAccentColor)};
   box-shadow: 0 2px 3px 3px rgba(0, 0, 0, 0.1);
 
-  ${props =>
-    props.overlay &&
-    css`
-      color: rgba(0, 0, 0, 0.75);
-    `}
+  ${props => (props.overlay ? 'color: rgba(0, 0, 0, 0.75);' : '')}
```
SteffiStuffel, line 19: This will result in invalid css for no overlay. Not a problem as it will be ignored, but not a good solution though.

web/src/components/RemoteContent.tsx
```diff
@@ -10,12 +10,8 @@ const SandBox = styled.div<{ centered: boolean; smallText: boolean }>`
     props.smallText ? `clamp(0.55rem, 1.6vh, ${props.theme.fonts.hintFontSize})` : props.theme.fonts.contentFontSize};
   line-height: ${props => props.theme.fonts.contentLineHeight};
 
-  ${props =>
-    props.centered &&
-    css`
-      text-align: center;
-      list-style-position: inside;
-    `}
+  ${props => (props.centered ? 'text-align: center;' : '')}
```
SteffiStuffel, line 13: Same as above, I don't think it is a good approach to set a attribute to an invalide value

web/src/components/ToolbarItem.tsx
```diff
@@ -8,7 +7,7 @@ import Tooltip from './Tooltip'
 
 type PropsType = {
   href: string
-  icon: IconDefinition
+  icon: FontAwesomeIconProps['icon']
```
SteffiStuffel, line 10: No icon types available any more?

SteffiStuffel (commented):
Tested on android, seems to work fine.

Web did not work for me:
igapp1008

web/src/components/ToolbarItem.tsx
steffenkleinle: Yes :/

native/src/utils/__tests__/sentry.spec.ts
```diff
@@ -106,7 +106,7 @@ describe('log', () => {
     expect(Sentry.addBreadcrumb).toHaveBeenCalledTimes(1)
     expect(Sentry.addBreadcrumb).toHaveBeenCalledWith({
       message: 'test error log',
-      level: Sentry.Severity.Error
+      level: 'error'
```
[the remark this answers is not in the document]
steffenkleinle: See https://github.com/getsentry/sentry-javascript/blob/7.0.0/MIGRATION.md#severity-severitylevel-and-severitylevels.
Sentry.Severity enum does not exist anymore, instead just plain strings should be used. Typescript makes sure that only valid values are passed. Therefore it is even more robust than before since it was possible to use invalid severity levels.

native/src/utils/sentry.ts
```diff
@@ -21,30 +22,26 @@ export const initSentry = (): void => {
...
     switch (level) {
-      case Sentry.Severity.Fatal:
-      case Sentry.Severity.Critical:
-      case Sentry.Severity.Error:
+      case 'fatal':
+      case 'error':
         console.error(message)
         break
-      case Sentry.Severity.Warning:
+      case 'warning':
```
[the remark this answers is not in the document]
steffenkleinle: See above

web/src/components/CityNotCooperatingFooter.tsx
steffenkleinle: Changed it because of this rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md
It is now a part of one of our extended eslint configs. I think we should stick to those rules most of the time if there is no important reason not to do so. I also think it kind of makes sense to have this consistent and to show that nothing is rendered.

web/src/components/Footer.tsx
steffenkleinle: Why does this result in invalid css? :see_no_evil: It just resolves to `''` i.e. empty string.

web/src/components/RemoteContent.tsx
steffenkleinle: Same as above

web/stylelint.config.js
f1sh1918: it's fine for me to use recommended config. so far i found nothing to add

.circleci/src/jobs/check.yml
steffenkleinle: Edit: Still failed in the Ci, just disabled the tests now. Created IGAPP-1070 to reenable.

web/src/components/RemoteContent.tsx
f1sh1918: hm i also don't see the benefit in the refactoring. Do you think its better readable? 
Can not start webpack to the if it results in invalid css, check comment

web/src/routes/TuNewsDetailPage.tsx
```diff
@@ -42,8 +42,8 @@ const StyledTitle = styled.div`
   height: 100%;
   align-items: center;
   justify-content: center;
-  background-color: ${({ theme }) => theme.colors.tunewsThemeColor};
-  color: ${({ theme }) => theme.colors.backgroundColor};
+  background-color: ${props => props.theme.colors.tunewsThemeColor};
+  color: ${props => props.theme.colors.backgroundColor};
```
f1sh1918, line 46: i'm fine with having everything same style but the object destructuring is the proper way in this case? Even shorter

f1sh1918 (changes requested):
Tested on ios. Everything works fine. Really nice refactoring :)
Unfortunately i can not start web anymore due to an `htmlparser2` error
```js
ERROR in ../node_modules/htmlparser2/lib/Tokenizer.js 6:41-81
Module not found: Error: Package path ./lib/decode_codepoint is not exported from package /Users/afischer/projects/integreat-app/web/node_modules/entities (see exports field in /Users/afischer/projects/integreat-app/web/node_modules/entities/package.json)

ERROR in ../node_modules/htmlparser2/lib/Tokenizer.js 7:15-45
Module not found: Error: Package path ./lib/decode is not exported from package /Users/afischer/projects/integreat-app/web/node_modules/entities (see exports field in /Users/afischer/projects/integreat-app/web/node_modules/entities/package.json)
```
I deleted web node modules and reinstalled. I'm on the recommended node lts version (16.16.0)

steffenkleinle:
@SteffiStuffel @f1sh1918 I think I fixed it, at least it works for me now again. I just added `htmlparser2` as web dependency and not it seems to work. I guess I missed it because I upgraded htmlparser2 to v8.0.1, tested it on web where everything worked, tested on native and had to downgrade to v7.2.0 again because it didn't work and didn't test web again.

web/src/components/RemoteContent.tsx
steffenkleinle: I definitely think so, yes. CSS attributes are defined separately which makes it better readable (at least in my opinion). We also don't need this weird `css` helper. However, I did not change this because I wanted to but because it didn't work together with stylelint.

web/src/components/CityNotCooperatingFooter.tsx
steffenkleinle: @SteffiStuffel if you have no objections I'd say `null` is the way to go here.

web/src/routes/TuNewsDetailPage.tsx
steffenkleinle: Didn't work with stylelint, feel free to try though. But I don't see the benefit right now to change this everywhere.
Sure it is shorter, but why is one the proper way and the other not?

web/src/components/Footer.tsx
```diff
@@ -12,18 +12,14 @@ const FooterContainer = styled.footer<{ overlay: boolean }>`
...
   & > * {
     @mixin remove-a;
-    margin: ${props => (props.overlay ? 0 : `5px`)};
```
f1sh1918, line 26: remove the `@mixin remove-a` this results errors. I even dunno where this mixin should be its never defined (optional)

f1sh1918 (approved):
Tested on ios & chrome works fine :)

.circleci/src/jobs/check.yml
steffenkleinle: Edit: Still failed in the CI, just deleted the tests now. @sarahsporck had a look as well and sadly we both didn't manage to fix it.

web/src/components/CityNotCooperatingFooter.tsx
SteffiStuffel: Just wanted to know.

SteffiStuffel (approved):
Tested on web. Works fine now.

PR #729, opened 2022-07-13
IGAPP-1048: Show not found error instead of response error

charludo:


api-client/src/Endpoint.ts
```diff
@@ -80,6 +82,12 @@ class Endpoint<P, T> {
       })
     })
 
+    // separate check for NotFoundErrors: response is not ok, but ResponseError should not be thrown
```
steffenkleinle, line 85: I think the comment does not really add additional information so I'd just remove it.

release-notes/unreleased/IGAPP-1048-show-not-found-error.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1048
+show_in_stores: true
+platforms:
+  - web
+en: Show Not Found Error instead of Response Error on invalid endpoints
```
steffenkleinle, line 5: Perhaps:
```suggestion
en: Show correct error message for non-existing categories
```
Not Found Error, Response Error and endpoints are just implementation details and not really relevant for non-devs.

api-client/src/errors/NotFoundError.ts
```diff
@@ -30,11 +30,13 @@ class NotFoundError extends Error {
       Error.captureStackTrace(this, NotFoundError)
     }
 
+    // default values are provided for cases where parameters are unknown
+    // e.g. when throwing NotFoundError from an endpoint
```
steffenkleinle, line 34: Same as above, comments should only be necessary if workarounds, complicated logic or similar need to be explained

api-client/src/Endpoint.ts
```diff
@@ -1,8 +1,10 @@
+// import { NotFoundError } from '.'
```
steffenkleinle, line 1:
```suggestion
```

api-client/src/errors/NotFoundError.ts
```diff
@@ -30,11 +30,13 @@ class NotFoundError extends Error {
...
     this.name = 'NotFoundError'
-    this._type = params.type
     this._id = params.id
-    this._city = params.city
-    this._language = params.language
+    this._type = params.type
+    this._city = params.city ?? 'unknown city'
+    this._language = params.language ?? 'en'
```
steffenkleinle, line 39: So the best solution would probably be to just remove `city` and `language` from the not found error completely and pass it separately as props if necessary. However, this would make changes in a lot of place necessary and is not worth the effort right now.
Therefore I'd prefer just making the properties optional and not using any fallback. Then you'd just have to handle the case that there is no city or language separately in `FailureSwitcher` by adding that as a condition to the if clause but everything else should continue to work.

steffenkleinle (changes requested):

steffenkleinle:
Tested on firefox and works great, thanks for diving into our weird endpoint logic :smile:

PR #730, opened 2022-07-20
IGAPP-1018: Header sidebar

f1sh1918:
1. Open Kebab Menu in Web Mobile Mode (Right Corner)
2. Check if Action Items work in KebabMenu
3. Check backdrop close of the kebab menu
4. check especially in MapView if KebabMenu overlays the bottomAction sheet


web/src/styles/MapView.css
```diff
@@ -5,3 +5,14 @@
 .maplibregl-ctrl-group button.maplibregl-ctrl-compass {
   cursor: auto;
 }
+
```
f1sh1918, line 8: hiding map controls under sidepanel

web/src/styles/KebabMenu.css
```diff
@@ -0,0 +1,8 @@
+.kebab-menu {
```
f1sh1918, line 1: css is needed since i couldn't use `styled(Portal)` and set style there

web/src/styles/BottomActionSheet.css
```diff
@@ -11,10 +11,6 @@
   top: calc(18px + env(safe-area-inset-top));
 }
 
-[data-rsbs-overlay] {
```
f1sh1918, line 14: unneeded css

web/src/components/LocationHeader.tsx
```diff
@@ -15,7 +15,9 @@ import {
   SHELTER_ROUTE,
   SPRUNGBRETT_OFFER_ROUTE
 } from 'api-client'
+import { config } from 'translations'
 
+import searchIconMobile from '../assets/IconSearch.svg'
```
f1sh1918, line 20: the old icons (desktop) gonna be replaced in a separate ticket for header redesign

web/src/routes/TuNewsPage.tsx
```diff
@@ -31,7 +32,7 @@ const DEFAULT_COUNT = 10
 const TuNewsPage = ({ cityCode, languageCode, cityModel, languages }: CityRouteProps): ReactElement => {
   const formatter = useContext(DateFormatterContext)
   const { t } = useTranslation('news')
-  const viewportSmall = false
+  const { viewportSmall } = useWindowDimensions()
```
steffenkleinle, line 35: :laughing:

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -5,29 +5,52 @@ import languageIcon from '../assets/language-icon.svg'
...
 
+  const closeDropDown = (): ReactElement => (
+    <Selector
+      closeDropDown={closeDropDown}
+      verticalLayout={false}
+      items={selectorItems}
+      activeItemCode={activeItemCode}
+      disabledItemTooltip={t('noTranslation')}
+    />
+  )
```
steffenkleinle, line 34: Something looks wrong here: You define a react component which receives itself as `closeDropDown` prop? Is this an error or could you explain this?

Edit: Ok, I think it was bad code before but now you are not passing the `closeDropDown` anymore as prop. It probably should be if we want to keep the previous way:
```suggestion
  const Item = (closeDropDown: () => void): ReactElement => (
    <Selector
      closeDropDown={closeDropDown}
      verticalLayout={false}
      items={selectorItems}
      activeItemCode={activeItemCode}
      disabledItemTooltip={t('noTranslation')}
    />
  )
```

web/src/components/KebabActionItemLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <span>{text}</span>
+      </StyledLink>
+    ) : (
+      <StyledSpan aria-label={text} dir='auto' style={{ flex: 1 }} data-testid='kebab-action-item'>
+        <img alt='' src={iconSrc} width='24px' height='24px' />
+        <span>{text}</span>
+      </StyledSpan>
+    )}
+  </>
+)
```
steffenkleinle, line 39: Optional because of personal opinion: I find the following way easier to read:
```suggestion
const KebabActionItemLink = ({ href, text, iconSrc }: PropsType): ReactElement => {
    if (href) {
    return (
      <StyledLink to={href} aria-label={text} dir='auto' data-testid='kebab-action-item'>
        <img alt='' src={iconSrc} width='24px' height='24px' />
        <span>{text}</span>
      </StyledLink>
    )}
    return (
      <StyledSpan aria-label={text} dir='auto' style={{ flex: 1 }} data-testid='kebab-action-item'>
        <img alt='' src={iconSrc} width='24px' height='24px' />
        <span>{text}</span>
      </StyledSpan>
    )})
}
```

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+  font-family: ${props => props.theme.fonts.web.decorativeFont};
+  position: absolute;
+  top: 0;
+  width: 80vw;
+  height: 100vh;
+  background-color: ${props => props.theme.colors.backgroundColor};
+  box-shadow: -3px 3px 3px 0 rgba(0, 0, 0, 0.13);
+  -webkit-font-smoothing: antialiased;
+  /* to stop flickering of text in safari 
+   */
```
steffenkleinle, line 33: Is the comment in the right place or does it belong to line 31?

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+  height: 28px;
+`
+
+const Overlay = styled.div<{ checked: boolean }>`
+  position: absolute;
+  display: none;
+  width: 100%;
+  height: 100vh;
+  top: 0;
+  left: 0;
```
steffenkleinle, line 55: Is this RTL compatible?

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+  background-color: rgba(0, 0, 0, 0.5);
+  z-index: 30;
+  ${props => (props.checked ? `display: block` : `display:none`)}
```
steffenkleinle, line 58:
```suggestion
  ${props => (props.checked ? `display: block` : `display: none`)}
```
Optional, not sure what is better:
```suggestion
  display: ${props => (props.checked ? `block` : `none`)}
```

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+`
+
+const CheckboxLabel = styled.label`
+  display: flex;
+  justify-content: center;
+`
+
+const Heading = styled.div<{ direction: string }>`
+  display: flex;
+  ${props => (props.direction === 'rtl' ? `justify-content: flex-start;` : `justify-content: flex-end;`)}
```
steffenkleinle, line 72: Optional: Same as above

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+`
+
+const List = styled.div<{ direction: UiDirectionType; checked: boolean }>`
+  font-family: ${props => props.theme.fonts.web.decorativeFont};
+  position: absolute;
+  top: 0;
+  width: 80vw;
+  height: 100vh;
+  background-color: ${props => props.theme.colors.backgroundColor};
+  box-shadow: -3px 3px 3px 0 rgba(0, 0, 0, 0.13);
```
steffenkleinle, line 30: Lots of magic values here in this css, this is also reused further down. Probably fine but how did you compute these values?

web/src/components/LocationHeader.tsx
```diff
@@ -54,16 +57,39 @@ const LocationHeader = (props: PropsType): ReactElement => {
 
   const { t } = useTranslation('layout')
 
-  const getActionItems = (): Array<ReactNode> => [
-    <HeaderActionBarItemLink key='search' href={searchPath} text={t('search')} iconSrc={searchIcon} />,
-    ...(!buildConfig().featureFlags.fixedCity
-      ? [<HeaderActionBarItemLink key='location' href={landingPath} text={t('changeLocation')} iconSrc={landingIcon} />]
-      : []),
+  const getActionItems = (viewportSmall: boolean): Array<ReactNode> => {
```
steffenkleinle, line 60: Optional: Maybe you could refactor `getActionItems` and `getKebabItems` to not be functions as it is unnecessary, i.e. just
```ts
  const actionItems = viewportSmall ? [...] : [...]
  const kebabItems = [...]
```

web/src/components/Portal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children: ReactNode
+  className: string
+  element?: string
+  opened?: boolean
+}
+/** A portal creates a new DOM Node outside the regular DOM. You can use it f.e. to fix z-index problems */
+export const Portal = ({ children, className, element = 'reach-portal', opened = false }: PropsType): ReactPortal => {
+  const [container] = React.useState(() =>
+    // This will be executed only on the initial render
+    // https://reactjs.org/docs/hooks-reference.html#lazy-initial-state
```
steffenkleinle, line 14: Imo no need to explain react hooks here so I'd remove the comment.

web/src/components/Portal.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+type PropsType = {
+  children: ReactNode
+  className: string
+  element?: string
+  opened?: boolean
+}
+/** A portal creates a new DOM Node outside the regular DOM. You can use it f.e. to fix z-index problems */
+export const Portal = ({ children, className, element = 'reach-portal', opened = false }: PropsType): ReactPortal => {
+  const [container] = React.useState(() =>
```
steffenkleinle, line 12:
```suggestion
  const [container] = useState(() =>
```

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,121 @@
...
+        data-testid='kebab-menu-checkbox'
+        checked={checked}
+        id='trigger'
+        type='checkbox'
+        onClick={() => setChecked(!checked)}
+        onChange={e => setChecked(e.target.checked)}
+      />
+      <CheckboxLabel htmlFor='trigger'>
+        <Icon src={iconKebabMenu} alt='' />
+      </CheckboxLabel>
```
steffenkleinle, line 103: Why do you use this strange way to define a button? Or what is the reasoning behind this?

steffenkleinle (changes requested):
Tested on firefox. Terrific job, very nice! :iphone: 

Minor things:
* Language selector does not close anymore after selecting a language, see comment below regarding `closeDropDown` "component"
* Search icon not mirrored on RTL layouts (optional, wasn't mirrored before as well)
* Multiple buttons/items don't use mouse pointer but just the regular mouse cursor: open kebab menu button, close kebab menu button, change language in kebab menu (probably not that relevant on mobile)

web/src/components/KebabMenu.tsx
f1sh1918: moved above

web/src/components/KebabMenu.tsx
f1sh1918: sorry forgot after changing to portal implementation to refactor this. find latest commit

web/src/components/KebabMenu.tsx
f1sh1918: well its from penpot. we could define it somewhere in the theme.

web/src/components/KebabMenu.tsx
f1sh1918: yes because its not important where the overlay starts its always covering the complete screen in this case and the Menu is moving in from the particular side and just showing a part of the overlay

steffenkleinle: Thanks for explaining, did not recognize any problems in RTL layout as well :+1:

f1sh1918:
> Tested on firefox. Terrific job, very nice! 📱
> 
> Minor things:
> 
> * Language selector does not close anymore after selecting a language, see comment below regarding `closeDropDown` "component"
> * Search/language icon not mirrored on RTL layouts (optional, wasn't mirrored before as well)
> * Multiple buttons/items don't use mouse pointer but just the regular mouse cursor: open kebab menu button, close kebab menu button, change language in kebab menu (probably not that relevant on mobile)

1. should be fixed now
2. we should implement this during replacing the old icons by the new ones, since it never worked and we have to touch everything there. I'm gonna add some additional redesign tasks, since there is stuff missing
3. We don't need cursor on mobile so no need to add css styles you can not see

steffenkleinle (approved):

steffenkleinle (approved):
Not sure if it would make sense from an UI/UX perspective to close the sidebar if a language is selected.
Tested, well done, language selection looks quite nice now :)

SteffiStuffel (commented):
Layout is broken for desktop view:
Open http://localhost:9000/augsburg/de -> Click on language change -> Französisch

1018fr

web/src/components/KebabMenu.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+type KebabMenuProps = {
+  items: Array<ReactNode>
+  direction: UiDirectionType
+}
+
+const ToggleContainer = styled.div`
+  display: flex;
+  padding: 0 8px;
+  z-index: 32;
```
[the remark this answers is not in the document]
f1sh1918: well we have the app where we can use up to 1x, bottomActionSheet 2x, and Sidebar 3x with the components inside which may have different z-index. If there are any better suggestions please let me know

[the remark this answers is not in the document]
SteffiStuffel: Not sure what the used z-indeces are, but it is recommended to use 30, 40, 50 and so on, and when needing a value in between, it is best to use the "middel", so you always have options to put stuff between, because values are still available.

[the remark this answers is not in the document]
f1sh1918: well we use 10 1000, 2000 and so one. i've fine with sticking to 10x values and gonna change it

f1sh1918:
> Layout is broken for desktop view: Open http://localhost:9000/augsburg/de -> Click on language change -> Französisch
> 
> 1018fr

happens also on main for me :( 
Seems like its not connected to this and only happens on localhost for me not webnext

SteffiStuffel (approved):

michael-markl:
Just saw this PR. It's quite hard to get a sidebar right I think.
It seems with the current impl. there is a bug with the layout when the page is high enough to be scrollable:
You can open the sidebar when scrolled down. Then the side bar will open "at the top" of the page:

Screen Shot 2022-08-04 at 11 42 51
 
Also, even if it's opened at the top of the page, you can scroll down and get the same mixed up layout.

To 3.: I would disagree with @f1sh1918 . I think cursors should reflect the intended action no matter what you think the device looks like that sees that particular screen. (It's somewhat of an accessibility feature)

And another note (maybe just taste): I would expect "Change language" to be the first option and "Change location" to be the second one.

PR #732, opened 2022-07-21
IGAPP-1028: Poi markers pixelated

f1sh1918:
1. http://localhost:9000/testumgebung/de/locations
2. Zoom on the map and check if markers are still pixelated

Tested on web & ios

### DIFF
--- a/api-client/src/maps/index.ts
+++ b/api-client/src/maps/index.ts
@@ -84,7 +84,7 @@ export const animationDuration = 2000
 export const nameQueryParam = 'name'
 
 export const mapMarker: MarkerConfig = {
-  iconSize: 1.75,
-  symbol: 'marker_15',
-  symbolActive: 'marker_15_active'
+  iconSize: 0.75,
+  symbol: 'marker_40',
+  symbolActive: 'marker_40_active'
 }
--- a/release-notes/unreleased/IGAPP-1028-poi-markers-pixelated.yml
+++ b/release-notes/unreleased/IGAPP-1028-poi-markers-pixelated.yml
@@ -1,0 +1,6 @@
+issue_key: IGAPP-1028
+show_in_stores: true
+platforms:
+  - web
+de: Darstellung des Marker Icons auf der Karte wurde verbessert
+en: Display of the marker icon was improved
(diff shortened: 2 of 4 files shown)


Cmd190 (approved):
Looks fine.

release-notes/unreleased/IGAPP-1028-poi-markers-pixelated.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1028
+show_in_stores: true
+platforms:
+  - web
```
steffenkleinle, line 4: Is this really web only? I think the used icons are changed on native as well, right?

api-client/src/maps/index.ts
```diff
@@ -84,7 +84,7 @@ export const animationDuration = 2000
 export const nameQueryParam = 'name'
 
 export const mapMarker: MarkerConfig = {
-  iconSize: 1.75,
-  symbol: 'marker_15',
-  symbolActive: 'marker_15_active'
+  iconSize: 0.75,
```
steffenkleinle, line 87: Not sure and maybe check back with the design team but maybe the icons are a little bit small now?

steffenkleinle (approved):
Tested on firefox and android

api-client/src/maps/index.ts
f1sh1918: well the icons size should be increased by 10%. Are you talking about web or native?

steffenkleinle: Ok. isn't even that smaller than before. I was just confused by `iconSize: 1.75` and `iconSize: 0.75` and thought it looks quite small now. Your choice if something has to be done here.

f1sh1918: no i increased it a little bit to have slightly bigger markers than the other icons on the map. We increased the svgs from 15 to 40 and these svs were converted to spriteicons. so if you compare `15*1.75 and 40*0.75` you can get the size by pixels. I think if thats not fine we gonna get the design feedback during beta

release-notes/unreleased/IGAPP-1028-poi-markers-pixelated.yml
f1sh1918: gonna adjust it

PR #735, opened 2022-07-26
IGAPP-983: Wrong text alignment

steffenkleinle:


steffenkleinle:
Created some test german content in http://localhost:9000/testumgebung/ar if you search for `Deutsch` and `Willkommen`.

web/src/components/CategoryEntry.tsx
```diff
@@ -76,79 +75,64 @@ type PropsType = {
...
-        </StyledLink>
-      </SubCategory>
-    ))
-  }
-
-  getMatchedContent(numWordsSurrounding: number): ReactNode {
-    const { query, theme, contentWithoutHtml } = this.props
+const CategoryEntry = ({ category, contentWithoutHtml, subCategories, query }: PropsType): ReactElement => {
+  const theme = useTheme()
+  const textToHighlight = useMemo<string | null>(() => {
```
f1sh1918, line 83: nice :)

web/src/components/CategoryEntry.tsx
```diff
@@ -76,79 +75,64 @@ type PropsType = {
...
+    />
+  )
+
+  return (
+    <Row>
+      <StyledLink to={category.path}>
+        <CategoryThumbnail alt='' src={category.thumbnail || iconPlaceholder} />
+        <CategoryListItem dir='auto'>
+          {Title}
+          <div style={{ margin: '0 5px', fontSize: '12px' }} dir='auto'>
```
f1sh1918, line 128: please use styled component  and optional we should define a fontSize in rem in `fonts.ts` with 0.75rem and replace the 12px values.

f1sh1918 (approved):
Looks fine to me. Nice refactoring using memo 🥇

web/src/components/CategoryEntry.tsx
steffenkleinle: not done in this PR so won't do

web/src/components/EventListItem.tsx
```diff
@@ -28,10 +28,10 @@ const getEventPlaceholder = (path: string): string => {
 const EventListItem = ({ event, formatter }: PropsType): ReactElement => (
   <ListItem thumbnail={event.thumbnail || getEventPlaceholder(event.path)} title={event.title} path={event.path}>
     <div>
-      <div>{event.date.toFormattedString(formatter)}</div>
-      {event.location && <div>{event.location.fullAddress}</div>}
+      <div dir='auto'>{event.date.toFormattedString(formatter)}</div>
```
SteffiStuffel, line 31: Hmh, would it make sense to create a component for this, which we reuse?

web/src/components/ShelterInformation.tsx
```diff
@@ -24,6 +24,10 @@ import ShelterInformationSection from './ShelterInformationSection'
 import { StyledButton } from './TextButton'
 import Tooltip from './Tooltip'
 
+const FullWidth = styled.div`
+  flex: 1;
```
SteffiStuffel, line 28: so `flex: 1` in container (line 27) is not needed any longer.

SteffiStuffel (approved):
Tested looks good.

Only in events, not sure if this is the best solution (date and adress showing in arabic rtl, but values in german), not sure what a better handling would be.

steffenkleinle:
> Only in events, not sure if this is the best solution (date and adress showing in arabic rtl, but values in german), not sure what a better handling would be.

Date is only in English format because of a moment bug and we need to wait for the CMS to deliver ISO8601 date strings to fix this by replacing moment. However, I think this is the current (even though not really nice/optimal) way to display it.

release-notes/unreleased/IGAPP-983-wrong-text-alignment.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: IGAPP-983
+show_in_stores: false
+platforms:
+  - web
```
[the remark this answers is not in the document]
steffenkleinle: Created https://issues.tuerantuer.org/browse/IGAPP-1079 since I don't know how to solve this (apart from content rendered as html which I fixed in this PR). If you have any ideas how to solve it for native it would be appreciated.

web/src/components/ShelterInformation.tsx
steffenkleinle: Removed flex in line 32.

web/src/components/EventListItem.tsx
steffenkleinle: Thought about this as well but there are quite a few different html elements we set this property on so, so separate reusable components or casts would be necessary so I decided I'll just keep it this way. Let me know if you'd prefer reusable component(s).

SteffiStuffel: Fine for me. We can change it in the future with same effort anyways, if we change our opinion.

PR #736, opened 2022-07-27
IGAPP-1011: Bed filter for shelter returns duplicated results

steffenkleinle:


f1sh1918 (approved):
Even i can reproduce the console warning with children with duplicated key, its not shown duplicated in the list anymore. So really nice work. I was quite sure i tried the same solution before once but maybe i was confused because the console warning still existed and didn't check for the particular entries in the list

```js
Warning: Encountered two children with the same key, `1155`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

steffenkleinle:
> Even i can reproduce the console warning with children with duplicated key, its not shown duplicated in the list anymore. So really nice work. I was quite sure i tried the same solution before once but maybe i was confused because the console warning still existed and didn't check for the particular entries in the list

The warning is still shown? That's odd, will look into it.

SteffiStuffel (approved):
Please delete branch IGAPP-1011-bed-filter-for-shelter-returns-duplicated-results   


Tested, works fine. Don't get a warning.

steffenkleinle:
> Even i can reproduce the console warning with children with duplicated key, its not shown duplicated in the list anymore. So really nice work. I was quite sure i tried the same solution before once but maybe i was confused because the console warning still existed and didn't check for the particular entries in the list
> 
> ```js
> Warning: Encountered two children with the same key, `1155`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
> ```

Cannot reproduce, no warnings for me either. Perhaps you didn't clear the console?

PR #738, opened 2022-08-01
IGAPP-1066: Back navigation after redirects impossible

charludo:


f1sh1918 (approved):
Works fine on Chrome & Safari

steffenkleinle (approved):

PR #740, opened 2022-08-01
IGAPP-1065: Poi center gps position

f1sh1918:
1.Go to Map.
2. Click on the GPS position FAB to center your position
3. Click on a poi without changing the current camera position 
4. Click on GPS fab to center position
5. Click on a poi on the list of the bottomActionSheet


steffenkleinle (changes requested):
Works for clicking pois on the map but not for clicking pois in the list:
* Open map
* Click on FAB to go to current position (no zoom changed, should probably be changed as well?)
* click on poi in list
* nothing happens on the map

steffenkleinle (approved):

SteffiStuffel (approved):
Tested on Android.

PR #744, opened 2022-08-08
IGAPP-1081: Fix failing tools

steffenkleinle:

Seems like it always have to first be command, then options and only then the action. Otherwise the options are just lost. Only tested commands where this was the case last time :/


SteffiStuffel (approved):
Looks valid. Not tested

charludo (approved):

PR #745, opened 2022-08-12
IGAPP-1035: Reference open source licenses on web

Cmd190:
Click on Footer>openSourceLicenses to see the license page.


web/src/routes/LicensePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, [])
+
+  const renderItem = (item: License) => (
+    <LicenseItem name={item.name} license={item.licenses} version={item.version} onPress={item.licenseUrl} />
+  )
+
+  return (
+    <div>
+      <Caption title={t('openSourceLicenses')} />
+      <List items={licenses} renderItem={renderItem} noItemsMessage='TODO' />
```
Cmd190, line 54: Should we add an extra translation for "no licenses found" or should we just use an english text without translation? The licenses are english anyway

f1sh1918: i'm fine with english only.

web/src/components/LocationFooter.tsx
```diff
@@ -29,6 +30,7 @@ const LocationFooter: React.FC<PropsType> = ({ city, language, overlay = false }
       <CleanLink to={disclaimerPath}>{t('imprintAndContact')}</CleanLink>
       <CleanLink to={aboutUrl}>{t('settings:about', { appName: buildConfig().appName })}</CleanLink>
       <CleanLink to={privacyUrl}>{t('privacy')}</CleanLink>
+      <CleanLink to={RoutePatterns[LICENSE_ROUTE]}>{t('openSourceLicenses')}</CleanLink>
```
f1sh1918, line 33:
```suggestion
      <CleanLink to={RoutePatterns[LICENSE_ROUTE]}>{t('settings:openSourceLicenses')}</CleanLink>
```

web/src/routes/LicensePage.tsx
f1sh1918: i think we should keep consistent and add a translation

f1sh1918 (approved):
deleted

f1sh1918 (changes requested):
Some minor issue with i18n. Rest looks fine

.prettierignore
```diff
@@ -3,6 +3,7 @@ native/ios/**/*
 native/vendor/**/*
 
 native/assets/licenses_native.json
+web/assets/licenses.json
```
steffenkleinle, line 6: Please be consistent here. Either use `licenses_native.json` and `licenses_web.json` or `licenses.json` in both cases. I'd prefer `licenses.json` since the web/native suffix is really redundant.

web/package.json
```diff
@@ -56,11 +57,14 @@
     "focus-trap-react": "^9.0.2",
     "htmlparser2": "^7.2.0",
     "i18next": "^20.6.1",
+    "license-checker": "^25.0.1",
```
steffenkleinle, line 60: Why do we need this for web but not for native?

web/package.json
```diff
@@ -56,11 +57,14 @@
...
     "lodash": "^4.17.21",
     "mapbox-gl": "npm:empty-npm-package@1.0.0",
     "maplibre-gl": "^2.1.9",
     "moment": "2.29.4",
     "normalize.css": "^8.0.1",
+    "npm-license-crawler": "^0.2.1",
+    "path-to-regexp": "^1.7.0",
```
steffenkleinle, line 67: Why do we need this?

web/src/RootSwitcher.tsx
```diff
@@ -8,6 +8,7 @@ import {
   createCitiesEndpoint,
   JPAL_TRACKING_ROUTE,
   LANDING_ROUTE,
+  LICENSE_ROUTE,
```
steffenkleinle, line 11: Didn't recognize before but I think this should be `LICENSES_ROUTE` (plural), especially since its value is also `licenses`.

web/src/routes/LicensePage.tsx
steffenkleinle: It should never be shown since there will always be more than zero licenses. So I would be fine with just passing an empty string as well.

web/src/routes/LicensePage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <ListItem path={onPress} title={name}>
+      <LicenseContainer>
+        {`version: ${version}`}
+        <br />
+        {`license: ${license}`}
+      </LicenseContainer>
+    </ListItem>
+  )
+}
```
steffenkleinle, line 35: Should be moved to its own file or just inlined below.

web/src/routes/LicensePage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  license: string
+  onPress: string
+}
+
+const LicenseItem = (props: PropsType): ReactElement => {
+  const { name, version, license, onPress } = props
+  return (
+    <ListItem path={onPress} title={name}>
+      <LicenseContainer>
+        {`version: ${version}`}
```
steffenkleinle, line 29: Could you try to make this simpler?

web/src/routes/LicensePage.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type PropsType = {
+  name: string
+  version: string | undefined
+  license: string
+  onPress: string
+}
+
+const LicenseItem = (props: PropsType): ReactElement => {
+  const { name, version, license, onPress } = props
```
steffenkleinle, line 25: Optional: If you destructure `props` directly you could remove brackets/return statement.

web/src/routes/LicensePage.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <br />
+        {`license: ${license}`}
+      </LicenseContainer>
+    </ListItem>
+  )
+}
+
+const LicensePage = (): ReactElement => {
+  const { t } = useTranslation('settings')
+  const [licenses, setLicenses] = useState<License[]>([])
```
steffenkleinle, line 39:
```suggestion
  const [licenses, setLicenses] = useState<License[] | null>(null)
```

web/src/routes/LicensePage.tsx
Cmd190: good idea

web/src/routes/LicensePage.tsx
Cmd190: done

web/src/routes/LicensePage.tsx
Cmd190: done

web/src/routes/LicensePage.tsx
Cmd190: I tried

web/package.json
Cmd190: I dont't know where this dependencies came from. I removed them and it still works fine

.prettierignore
Cmd190: done

web/src/RootSwitcher.tsx
Cmd190: renamed

web/src/routes/LicensePage.tsx
Cmd190: Imo it is not necessary to move 20 loc in its own file when it's only used in this file

web/src/routes/LicensePage.tsx
```diff
@@ -21,26 +21,21 @@ type PropsType = {
...
-        {`license: ${license}`}
-      </LicenseContainer>
-    </ListItem>
-  )
-}
+const LicenseItem = ({ license, name, onPress, version }: PropsType): ReactElement => (
+  <ListItem path={onPress} title={name}>
+    <LicenseContainer>
+      <div>{`version: ${version}`}</div>
+      <div>{`license: ${license}`}</div>
```
steffenkleinle, line 28: The following should work as well (maybe divs/br tag are needed again):
```suggestion
      version: {version}
      license: ${license}
```

web/src/routes/LicensePage.tsx
```diff
@@ -21,26 +21,21 @@ type PropsType = {
...
+    </LicenseContainer>
+  </ListItem>
+)
 
 const LicensePage = (): ReactElement => {
```
steffenkleinle, line 33: Should be the following (same for file name):
```suggestion
const LicensesPage = (): ReactElement => {
```

steffenkleinle (changes requested):
Translation is missing, it should be available at least in German:
image

I think there should be translations for `version` and `license` as well.

I also think you shouldn't use the whole width of the page for the list (probably you just need to wrap it in `<Layout />` or something-

web/src/routes/LicensePage.tsx
Cmd190: Well, I think both ways are ok and therefore I leave it as it is

Cmd190:
> Translation is missing, it should be available at least in German: image
> 
> I think there should be translations for `version` and `license` as well.
> 
> I also think you shouldn't use the whole width of the page for the list (probably you just need to wrap it in `<Layout />` or something-



> 

The translation for open source licenses  is available in multiple languages though...
Imo it's not necessary to translate the content of the licenses page. Anyone who want to understand the content must speak english because the packages name aren't translated anyway.

native/assets/licenses_native.json
```diff
@@ -0,0 +1,320 @@
+{
```
steffenkleinle, line 1: Please remove

web/src/routes/LicensePage.tsx
steffenkleinle: @f1sh1918 @SteffiStuffel what do you think? Do we have/should we have a policy for this?

web/src/routes/LicensePage.tsx
steffenkleinle: Well, please try to keep stuff as simple as possible. It just does not make sense to wrap the whole string in a template string  and use one template in another one if it works otherwise as well.

steffenkleinle (changes requested):
Two more names that just be adjusted:
* `LicenseRouteType` -> `LicensesRouteType` (`api-client/src/routes/index.ts`)
* `LicenseRouteInformationType` -> `LicensesRouteInformationType` (`api-client/src/routes/RouteInformationTypes.ts`)

The page should also show the `GeneralHeader` and the `GeneralFooter`. Sorry, missed that before.

steffenkleinle:
> The translation for open source licenses is available in multiple languages though...

Yes, maybe, but the translations are not used, see my screenshot. 

> Imo it's not necessary to translate the content of the licenses page. Anyone who want to understand the content must speak english because the packages name aren't translated anyway.

Please translate it anyway, we don't use hard coded strings for the UI in our apps for various reasons. We want our apps to be as accessible to all users as possible. That includes that a user can always see and understand what a specific page is about (for example if clicked on a link by mistake) even if the specific content is not understood.

web/src/routes/LicensePage.tsx
Cmd190: I'll just fix it along with the other issues

Cmd190:
> Translation is missing, it should be available at least in German: image
> 
> I think there should be translations for `version` and `license` as well.
> 
> I also think you shouldn't use the whole width of the page for the list (probably you just need to wrap it in `<Layout />` or something-



> 

The translation for open source licenses  is available in multiple languages though...
Imo it's not necessary to translate the content of the licenses page. Anyone who want to understand the content must speak english because the packages name aren't translated anyway. 

> > The translation for open source licenses is available in multiple languages though...
> 
> Yes, maybe, but the translations are not used, see my screenshot. You can fix this by changing
> 
> ```tsx
>  <CleanLink to={RoutePatterns[LICENSES_ROUTE]}>{t('openSourceLicenses')}</CleanLink>
> ```
> 
> to
> 
> ```tsx
>  <CleanLink to={RoutePatterns[LICENSES_ROUTE]}>{t('settings:openSourceLicenses')}</CleanLink>
> ```
> 
> in `GeneralFooter.tsx`.
> 
> > Imo it's not necessary to translate the content of the licenses page. Anyone who want to understand the content must speak english because the packages name aren't translated anyway.
> 
> Please translate it anyway, we don't use hard coded strings for the UI in our apps for various reasons. We want our apps to be as accessible to all users as possible. That includes that a user can always see and understand what a specific page is about (for example if clicked on a link by mistake) even if the specific content is not understood.

ok, I get it now. And hardcoded ui strings are really ugly...

web/src/routes/LicensePage.tsx
Cmd190: I just extracted it into its own file

native/package.json
```diff
@@ -84,7 +84,7 @@
     "react-native-elements": "^3.4.2",
     "react-native-gesture-handler": "^2.5.0",
     "react-native-highlight-words": "^1.0.1",
-    "react-native-inappbrowser-reborn": "^3.7.0",
+    "react-native-inappbrowser-reborn": "^3.6.3",
```
steffenkleinle, line 87: Why the downgrade? Please revert

web/src/routes/LicensesPage.tsx
```diff
@@ -45,11 +26,26 @@ const LicensesPage = (): ReactElement => {
...
+    languageCode,
+    pathname,
+  }
+
   return (
-    <Layout>
-      <Caption title={t('openSourceLicenses')} />
-      <List items={licenses ?? []} renderItem={renderItem} noItemsMessage='' />
-    </Layout>
+    <LocationLayout isLoading={false} {...locationLayoutParams}>
```
steffenkleinle, line 45: This should not be a location layout route since the licenses are not specific to one city/language. This breaks navigation for the link in the general footer e.g. in the city selection:
image

api-client/src/routes/pathname.ts
```diff
@@ -35,7 +35,7 @@ export const cityContentPath = ({ cityCode, languageCode, route, path }: CityCon
   constructPathname([cityCode, languageCode, route, path])
 
 export const pathnameFromRouteInformation = (routeInformation: NonNullableRouteInformationType): string => {
-  if (routeInformation.route === JPAL_TRACKING_ROUTE || routeInformation.route === LICENSES_ROUTE) {
```
steffenkleinle: Revert

steffenkleinle (changes requested):

native/package.json
```diff
@@ -135,6 +135,8 @@
     "jetifier": "^2.0.0",
     "metro-react-native-babel-preset": "^0.66.2",
     "node-fetch": "^3.2.8",
+    "patch-package": "^6.4.7",
+    "postinstall-postinstall": "^2.1.0",
```
steffenkleinle, line 139:
```suggestion
```

native/package.json
f1sh1918: probably due to merging by accident. reverted

api-client/src/routes/RouteInformationTypes.ts
```diff
@@ -46,7 +46,7 @@ export type NewsRouteInformationType = ParamsType & {
 }
 export type SimpleCityContentFeatureType = ParamsType & {
   // Routes without customizable ids, e.g. '/augsburg/de/disclaimer/
-  route: DisclaimerRouteType | OffersRouteType | SprungbrettOfferRouteType | SearchRouteType
+  route: DisclaimerRouteType | OffersRouteType | SprungbrettOfferRouteType | SearchRouteType | LicensesRouteType
 }
```
steffenkleinle, line 50: This is not correct, licenses route is not a city content route. Please adjust.

translations/translations.json
```diff
@@ -4427,6 +4427,18 @@
...
   },
+  "licenses": {
+    "de": {
+      "version": "Version:",
+      "license": "Lizenz:",
+      "noLicensesMessage": "Keinen Lizenzinformationen verfügbar"
+    },
+    "en": {
+      "version": "version:",
+      "license": "license:",
```
steffenkleinle, line 4438: Since its the first word I'd use uppercase I think
```suggestion
      "version": "Version:",
      "license": "License:",
```

web/src/components/KebabMenu.tsx
```diff
@@ -76,9 +76,13 @@ const ToggleButton = styled.button`
   margin-top: 6px;
 `
 
-const KebabMenu = ({ items, direction }: KebabMenuProps): ReactElement => {
+const KebabMenu = ({ items, direction }: KebabMenuProps): ReactElement | null => {
   const [checked, setChecked] = useState<boolean>(false)
 
+  if (items.length === 0) {
+    return null
+  }
```
steffenkleinle, line 84: Why is this change necessary?

web/src/components/LocationFooter.tsx
```diff
@@ -23,12 +23,18 @@ const LocationFooter: React.FC<PropsType> = ({ city, language, overlay = false }
     cityCode: city,
     languageCode: language,
   })
+  const licensesPath = pathnameFromRouteInformation({
+    route: LICENSES_ROUTE,
+    cityCode: city,
+    languageCode: language,
```
steffenkleinle, line 29:
```suggestion
```

steffenkleinle (approved):
Tested :+1:

steffenkleinle:
When I run yarn on this branch the lockfile changes. Apparently it does not in the CI though so probably just a problem on my machine?

sarahsporck (approved):
Works nice, but currently the licenseUrl doesn't work for repos without a master branch (e.g. dompurify). Opened a [PR](https://github.com/mwittig/license-checker/pull/2).
But i think we can still merge this PR.
Honestly, I'm okay with either approach, but I was wondering why we create the license file beforehand. Was this a performance consideration or do we not now the licenses during runtime?

web/src/components/KebabMenu.tsx
f1sh1918: for the `General Header` we do not pass items, so the `KebabMenu` shouldn't be displayed. But maybe we have to remove that soon, because the footer gonna take place in the SidebarMenu.

web/src/components/LicenseItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+const LicenseContainer = styled.div`
+  padding: 5px 10px;
+  line-height: 120%;
+`
+
+type PropsType = {
+  name: string
+  version: string | undefined
+  license: string
+  onPress: string
```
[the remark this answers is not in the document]
f1sh1918: Definitely . didn't see this while refactoring ;)

api-client/src/routes/RouteInformationTypes.ts
f1sh1918: done

steffenkleinle (approved):

f1sh1918 (approved):

PR #746, opened 2022-08-15
IGAPP-1029: Chat bot widget for munich

SteffiStuffel:

Unfortunately the rasa react package is not working, therefore I just used the plain js script. Furthermore the css is a bit broken for mobile view, I tried to fix it, but I think this will be a lot of work, which is - in my opinion - too much for current stage of the feature. Also using our Icon and style can get a lot of work, so I just sticked to the rasa style, but used Integreat color.

The backend url is not set correctly now, as I do not know it, i asked in ~twa-chatbot channel.


web/src/components/LocationLayout.tsx
```diff
@@ -28,6 +31,8 @@ type PropsType = {
 
 const LocationLayout = (props: PropsType): ReactElement => {
   const [feedbackModalRating, setFeedbackModalRating] = useState<FeedbackRatingType | null>(null)
+  const theme = useTheme()
+  useScript('https://unpkg.com/@rasahq/rasa-chat')
```
f1sh1918, line 35: We should add a condition here, since for the moment we only want to load the script in devFriendlyMode & for Munich but the script will be loaded in any case even its not rendered in the DOM

f1sh1918 (changes requested):
The widget works fine so far but i recognized some issues.
1. When you switch to location 'Munich' Widget Button gets rendered. Switch to another location. Widget is still available
2. Since the useScript is not conditionally loaded the scripts will be downloaded in any case which cause ~250kb

web/src/components/LocationLayout.tsx
```diff
@@ -73,7 +78,17 @@ const LocationLayout = (props: PropsType): ReactElement => {
...
       toolbar={toolbar}>
-      {children}
+      <>
+        {children}
+        {buildConfig().featureFlags.developerFriendly && cityModel.name === 'München' && (
+          <div
+            id='rasa-chat-widget'
+            data-primary={theme.colors.themeColor}
+            data-primary-highlight={theme.colors.themeColor}
+            data-websocket-url='https://your-rasa-url-here/'
```
f1sh1918, line 88: seems not to be a working url? do we need to pass this?

web/src/hooks/useScript.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react'
+
+const useScript = (url: string): void => {
+  useEffect(() => {
+    const script = document.createElement('script')
+
+    script.src = url
+    script.async = true
+
```
f1sh1918, line 9: maybe its safer to also add here
`script.defer = true`

https://javascript.info/script-async-defer#defer

web/src/components/LocationLayout.tsx
SteffiStuffel: As i mentioned in the PR description, i do not know the correct url right now, but I asked and will update as soon as I get it.

web/src/hooks/useScript.ts
SteffiStuffel: Good point. Thanks for sharing!

SteffiStuffel:
> The widget works fine so far but i recognized some issues.
> 
> 1. When you switch to location 'Munich' Widget Button gets rendered. Switch to another location. Widget is still available
> 2. Since the useScript is not conditionally loaded the scripts will be downloaded in any case which causes ~250kb

Fixed it.

web/src/components/RasaWidget.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactElement, useEffect } from 'react'
```
steffenkleinle, line 1: I think I'd rather name this `ChatBotWidget` as the technology behind it is an implementation detail and not really relevant/may not be known when looking for the code.

web/src/components/LocationLayout.tsx
```diff
@@ -73,7 +75,10 @@ const LocationLayout = (props: PropsType): ReactElement => {
       footer={!isLoading && showFooter ? <LocationFooter city={cityModel.code} language={languageCode} /> : null}
       modal={feedbackModal}
       toolbar={toolbar}>
-      {children}
+      <>
+        {children}
+        {buildConfig().featureFlags.developerFriendly && cityModel.name === 'München' && <RasaWidget />}
```
steffenkleinle, line 80: Probably does not really matter but I'd prefer a check on `code` since it is unique and does not contain uppercase or special characters.
```suggestion
        {buildConfig().featureFlags.developerFriendly && cityModel.code === 'muenchen' && <RasaWidget />}
```

web/src/components/RasaWidget.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const script = document.createElement('script')
+
+    script.src = 'https://unpkg.com/@rasahq/rasa-chat'
+    script.async = false
+
+    document.body.appendChild(script)
+
+    return () => {
+      document.body.removeChild(script)
+      const button = document.getElementById('rasa-chat-widget-container')
```
steffenkleinle, line 17: Shouldn't this be (matching the id below):
```suggestion
      const button = document.getElementById('rasa-chat-widget')
```

web/src/components/RasaWidget.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useTheme } from 'styled-components'
+
+const RasaWidget = (): ReactElement => {
+  const theme = useTheme()
+
+  useEffect(() => {
+    const script = document.createElement('script')
+
+    script.src = 'https://unpkg.com/@rasahq/rasa-chat'
+    script.async = false
```
steffenkleinle, line 11: No async and no defer now?

web/src/components/RasaWidget.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactElement, useEffect } from 'react'
+import { useTheme } from 'styled-components'
+
+const RasaWidget = (): ReactElement => {
+  const theme = useTheme()
+
+  useEffect(() => {
```
steffenkleinle, line 7: Can you please explain why we need the `useEffect` and `documentcreateElement` instead of just rendering it normally:
``` tsx
<script defer ... />
```

web/src/hooks/useScript.ts
steffenkleinle: Why is async safer then defer?

steffenkleinle (approved):
Works fine and looks good as far as I can test it :)

web/src/components/RasaWidget.tsx
SteffiStuffel: This will load the script everytime the component is rendered.

web/src/components/RasaWidget.tsx
SteffiStuffel: It won't work as the script only takes the element with the id 'rasa-chat-widget', but does not render it directly. Have a look at the DOM and you will see, there is no element with this id.

web/src/hooks/useScript.ts
SteffiStuffel: Both defer and async does not make sense. 

Defer and Async both download the script while HTML rendering (not waiting with rendering while downloading)
Defer executes script after HTML rendering is complete. Async waits with HTML rendering while script executing. 

See a nice image for explanation here: https://stackoverflow.com/questions/10808109/script-tag-async-defer

I decided to use defer instead of  async, as we do not what this chat-bot to block/break anything.

web/src/components/RasaWidget.tsx
SteffiStuffel: Hmh, what was I doing there. :thinking: Will fix.

f1sh1918 (approved):
works nice now. great work :)

PR #748, opened 2022-08-16
IGAPP-889: Header redesign

f1sh1918:


web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -0,0 +1,104 @@
...
+      setShowArrowLeft(scrollPosition > 0)
+    }
+  }, [scrollContainer, scrollPosition])
+
+  return (
+    <Container>
+      <Button disabled={!showArrowLeft} onClick={() => scrollContainer.current?.scroll({ left: 0 })}>
+        <Arrow src={iconArrowBack} direction={direction} visible={showArrowLeft} alt='' />
+      </Button>
+      <ScrollContainer ref={scrollContainer} onScroll={(e: any) => setScrollPosition(Math.abs(e.target.scrollLeft))}>
```
f1sh1918, line 88: didn't find an eventType that supports scrollLeft. Any suggestions?

SteffiStuffel: You can use this: `onScroll={(e: React.UIEvent<HTMLElement>) => setScrollPosition(Math.abs(e.currentTarget.scrollLeft))}`

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { ReactElement, ReactNode, RefObject, useEffect, useState } from 'react'
```
SteffiStuffel, line 1: In which file was this scroll behavior implemented before? Do we have it duplicated now?

web/src/components/HeaderNavigationItem.tsx
```diff
@@ -58,22 +77,31 @@ const StyledLink = styled(Link)<{ $active: boolean }>`
       : ''}
 `
 
+const StyledText = styled.span<{ $active: boolean }>`
+  @media ${dimensions.smallViewport} {
+    ${props => (props.$active ? 'font-weight: 800' : '')}
```
SteffiStuffel, line 82: Where is the text set to bold, if active, for bigger screens? It makes much more sense to set it in the same file for both screen sizes.

SteffiStuffel (changes requested):
Scrolling the NavigationItems is not working for rtl languages.

Images are missing for malte and obdach.
889-malte
889-obdach



Should the heading for aschaffenburg really look like this? (e.g. centered looks weird)

web/src/components/BottomActionSheet.tsx
```diff
@@ -54,6 +54,7 @@ const BottomActionSheet = React.forwardRef(
         direction={direction}
         ref={ref}
         open
+        scrollLocking={false}
```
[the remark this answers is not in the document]
f1sh1918: this is a safari browser issue. fixed by this prop. You can check library for more info.
https://github.com/stipsan/react-spring-bottom-sheet#scrolllocking

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: there was only the standard scrolling of a div container via css `overflow-x: auto` no additional logic was implemented before for web

web/src/components/DirectionIcon.tsx
```diff
@@ -0,0 +1,12 @@
+import styled, { css } from 'styled-components'
+
+import { UiDirectionType } from 'translations/src'
+
+export const DirectionIcon = styled.img<{ direction?: UiDirectionType }>`
```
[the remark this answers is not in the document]
f1sh1918: nice hint thx

web/src/components/HeaderNavigationItem.tsx
f1sh1918: fixed

f1sh1918:
> Scrolling the NavigationItems is not working for rtl languages.
> 
> Images are missing for malte and obdach. 
> 
> Should the heading for aschaffenburg really look like this? (e.g. centered looks weird) 

1. Scrolling works for me here: http://localhost:9000/testumgebung/ar/events even with button and touch. (Firefox and chrome tested) How did you test?
2. fixed missing images
3. aligned headerText left

image

SteffiStuffel (approved):
Tested again. Scroll for arabic works fine.

web/src/components/HeaderNavigationItem.tsx
```diff
@@ -58,22 +77,29 @@ const StyledLink = styled(Link)<{ $active: boolean }>`
...
-  width: ${ICON_SIZE_LARGE}px;
   display: flex;
   justify-content: center;
   align-items: center;
-  box-shadow: 0 2px 5px -3px rgba(0, 0, 0, 0.7);
-  transition: box-shadow 0.2s, border 0.2s;
-  border: white 2px solid;
+
+  @media ${dimensions.mediumLargeViewport} {
+    background-color: white;
```
steffenkleinle, line 94: I think we should use a color from our theme.

web/src/components/HeaderNavigationItem.tsx
```diff
@@ -58,22 +77,29 @@ const StyledLink = styled(Link)<{ $active: boolean }>`
...
+    box-sizing: border-box;
+    border-radius: 100%;
+    height: ${ICON_SIZE_LARGE}px;
+    width: ${ICON_SIZE_LARGE}px;
+    box-shadow: 0 2px 5px -3px rgba(0, 0, 0, 0.7);
+    transition: box-shadow 0.2s, border 0.2s;
+    border: white 2px solid;
```
steffenkleinle, line 101: Same

web/src/components/IconWithUiDirection.tsx
```diff
@@ -0,0 +1,12 @@
+import styled, { css } from 'styled-components'
+
+import { UiDirectionType } from 'translations/src'
+
+export const IconWithUiDirection = styled.img<{ direction?: UiDirectionType }>`
+  ${props =>
+    props.direction === 'rtl' &&
+    css`
```
steffenkleinle, line 8: Any reason why we need this helper?

web/src/components/IconWithUiDirection.tsx
```diff
@@ -0,0 +1,12 @@
+import styled, { css } from 'styled-components'
+
+import { UiDirectionType } from 'translations/src'
+
+export const IconWithUiDirection = styled.img<{ direction?: UiDirectionType }>`
```
steffenkleinle, line 5: Remove named import

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -0,0 +1,106 @@
...
+const Arrow = styled.img<{ direction?: string; visible: boolean }>`
+  width: 12px;
+  height: 10px;
+  padding: 0 8px;
+  align-self: center;
+  opacity: 0;
+
+  ${props =>
+    props.direction === 'rtl' &&
+    css`
```
steffenkleinle, line 29: Same, why do we need these helpers? Again belowl.

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [showArrowRight, setShowArrowRight] = useState<boolean>(false)
+  const [showArrowLeft, setShowArrowLeft] = useState<boolean>(false)
+  const [scrollPosition, setScrollPosition] = useState(0)
+
+  useEffect(() => {
+    if (scrollContainer.current) {
+      setShowArrowRight(scrollPosition < getScrollableWidth(scrollContainer))
+      setShowArrowLeft(scrollPosition > 0)
+    }
+  }, [scrollContainer, scrollPosition])
```
steffenkleinle, line 81: Using a ref as dependency does nothing I think.

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return (
+    <Container>
+      <Button disabled={!showArrowLeft} onClick={() => scrollContainer.current?.scroll({ left: 0 })}>
```
steffenkleinle, line 85: Optional: I'd prefer not to use inline functions that much. Makes jsx clearer and you can directly see what happens on click if you use good variable names (e.g. `scrollToStart`).

Same for `onScroll` and other `onClick` method.

web/src/components/NavigationBarScrollContainer.tsx
steffenkleinle, line 81: I think this should not be done with a `useEffect`. If you need to keep `showArrowRight` and `showArrowLeft` the better solution would be just setting it directly in the callback handlers (i.e. in `onClick` and `onScroll`). However, I think that the state is not needed at all as the values can just be calculated directly from `scrollPosition`:
``` ts
const showArrowLeft = scrollContainer.current ? scrollPosition < getScrollableWidth(scrollContainer) : false
const showArrowRight = scrollContainer.current ? scrollPosition > 0 : false
```

web/src/components/NavigationBarScrollContainer.tsx
steffenkleinle, line 85: Shouldn't the button/arrow not be rendered at all if there is enough space for all items to fit?

steffenkleinle (commented):
image

I think the scroll bar/arrows should be hidden here as they are not needed.

f1sh1918:
> image
> 
> I think the scroll bar/arrows should be hidden here as they are not needed.

i tested this but then when you scoll the content jumps when they appear so i decided to use opacity

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: see answer to you comment below. already talked to nikola about this

web/src/components/IconWithUiDirection.tsx
f1sh1918: what do you mean with "helpers"?

steffenkleinle: I meant the `css` helper. Why do we need that one instead of just using a regular string?

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: i agree with you if the functions are not just one liners. I added a onScrollForward fkt. The rest is fine imo

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: removed useEffect

web/src/assets/Offers.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="23" height="20" viewBox="0 0 23 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
f1sh1918: moved

web/src/components/IconWithUiDirection.tsx
f1sh1918: `import { UiDirectionType } from 'translations'`

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: nice hint thx

web/src/components/IconWithUiDirection.tsx
steffenkleinle: What do you mean by that comment? I was referring to the following:
```suggestion
const IconWithUiDirection = styled.img<{ direction?: UiDirectionType }>`

export default IconWithUiDirection
```

steffenkleinle:
> > I think the scroll bar/arrows should be hidden here as they are not needed.
> @steffenkleinle
> i tested this but then when you scoll the content jumps when they appear so i decided to use opacity

I understand what you mean but I am just referring to the specific situation shown in my screenshot above. The buttons should not be rendered at all if all the navigation items fit on the screen. Which is the case on my device. If you don't want to fix this here I can create an issue.

steffenkleinle:
Apart from that everything looks to be working just fine. There are still some unaddressed comments though.

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: i adjusted this in the last commit

web/src/components/IconWithUiDirection.tsx
f1sh1918: fixed in last commit

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: removed helpers

f1sh1918:
> 

i resolved the issues with the css helpers and also added condition for notRendering the arrows at all. Could you pls test agains @steffenkleinle

steffenkleinle (commented):
If all items fit on the screen it works and looks really well now :tada: 

However, there is still one issue if the screen does not fit all items (added one item for testing purposes): If you scroll completely to the right, it jumps back to the start and the arrows disappear. If you start scrolling again, it is still in the beginning but both the left and the right arrow is shown.
https://user-images.githubusercontent.com/29095487/189674856-1d591a44-9440-4b73-b6c6-0a8594021c31.mp4

f1sh1918:
> If all items fit on the screen it works and looks really well now 🎉
> 
> However, there is still one issue if the screen does not fit all items (added one item for testing purposes): If you scroll completely to the right, it jumps back to the start and the arrows disappear. If you start scrolling again, it is still in the beginning but both the left and the right arrow is shown. https://user-images.githubusercontent.com/29095487/189674856-1d591a44-9440-4b73-b6c6-0a8594021c31.mp4

@steffenkleinle: good catch! i added a separate check for `showArrows` since its not depending on scrollPosition. should be fine now

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -33,7 +33,7 @@ const Button = styled.button`
   padding: 0;
 `
 
-const ScrollContainer = styled.div<{ showArrows: boolean }>`
+const ScrollContainer = styled.div<{ showArrows?: boolean }>`
```
steffenkleinle, line 36:
```suggestion
const ScrollContainer = styled.div<{ showArrows: boolean }>`
```

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -61,8 +61,10 @@ const getScrollableWidth = (scrollContainer: RefObject<HTMLDivElement>): number
 
 const NavigationBarScrollContainer = ({ children, direction, scrollContainerRef }: PropsType): ReactElement => {
   const [scrollPosition, setScrollPosition] = useState(0)
+
+  const showArrows = scrollContainerRef.current ? getScrollableWidth(scrollContainerRef) > 0 : false
   const showArrowLeft = scrollContainerRef.current ? scrollPosition > 0 : false
-  const showArrows = scrollContainerRef.current ? scrollPosition < getScrollableWidth(scrollContainerRef) : false
+  const showArrowRight = scrollContainerRef.current ? scrollPosition < getScrollableWidth(scrollContainerRef) : false
```
steffenkleinle, line 67: I would prefer to adjust the naming here. `showArrows` is a different `show` than `showArrowLeft` and `showArrowRight` (not rendering arrows vs just hiding the arrows). Maybe `scrollContainerNeeded` or something like that?

steffenkleinle (approved):

steffenkleinle:
Nice work!

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: you are right but i think it should be named like `showArrowContainer` because the scrollContainer is rendered in any case

steffenkleinle: Works for me, just different :)

PR #750, opened 2022-08-23
IGAPP-893: Improve breadcrumbs

f1sh1918:

1. Switch to mobile mode
2. Navigate through local information


web/src/components/Breadcrumbs.tsx
```diff
@@ -49,9 +65,15 @@ const Breadcrumbs = ({ direction, ancestorBreadcrumbs, currentBreadcrumb }: Prop
     <Wrapper direction={direction}>
       <JsonLdBreadcrumbs breadcrumbs={jsonLdBreadcrumbs} />
       <OrderedList direction={direction}>
-        {ancestorBreadcrumbs.map(breadcrumb => (
-          <Breadcrumb key={breadcrumb.title}>{breadcrumb.node}</Breadcrumb>
-        ))}
+        {ancestorBreadcrumbs.map((breadcrumb, index) =>
+          ancestorBreadcrumbs.length > 1 && index === 0 ? (
+            <StyledLink to={pathnameFromUrl(breadcrumb.link)}>
```
steffenkleinle, line 70: Now we are creating the url from the pathname in CategoriesPage and extracting the pathname again here. I think we should rather just pass the pathname and create the url when needed.

steffenkleinle (approved):
Tested! Really really good work :tada:

web/src/components/Breadcrumbs.tsx
f1sh1918: @steffenkleinle  i removed the fkt here and also the unneeded convertion in `CategoriesPage`. Why did we need this fkt at all? I think relative links should be fine here?

web/src/routes/CategoriesPage.tsx
```diff
@@ -31,15 +31,14 @@ import { cmsApiBaseUrl } from '../constants/urls'
...
-import { urlFromPath } from '../utils/stringUtils'
 
 const CATEGORY_NOT_FOUND_STATUS_CODE = 400
 
 const getBreadcrumb = (category: CategoryModel, cityName: string) => {
   const title = category.isRoot() ? cityName : category.title
   return new BreadcrumbModel({
     title,
-    link: urlFromPath(category.path),
+    link: category.path,
```
steffenkleinle, line 41: This perhaps breaks the json ld creation (see `JsonLdBreadcrumbs`. It is used there for the `item` property. So you need `urlFromPath` there I'd say. Optional: I would rename `link` to `pathname` to make it clearer that it is not a full url.

steffenkleinle (changes requested):

SteffiStuffel (commented):
I do not get the home icon displayed:
893-home-icon

And also the text is cut off:
893-cut

f1sh1918:
> I do not get the home icon displayed and the text is cut off: (should be "... employmenT") 893-cut

I cannot reproduce this issue. i use the minimum viewPort width we support in design. In this case both parts gonna be ellipsed which is fine. Icon is also working for me

image

web/src/routes/CategoriesPage.tsx
f1sh1918: are you sure this break? i don't see any difference:
with urlFromPath:
image
without:
image

steffenkleinle:
> > I do not get the home icon displayed and the text is cut off: (should be "... employmenT")
> I cannot reproduce this issue. i use the minimum viewPort width we support in design. In this case both parts gonna be ellipsed which is fine, because its way too much content and handled by "flex-shrink". Icon is also working for me Maybe fresh `yarn install` 

Agreeing with @f1sh1918 here, can't reproduce, home icon is shown and text is cut off with ellipsis with a reasonable length still visible. I think we are good here.

web/src/routes/CategoriesPage.tsx
steffenkleinle: For me it differs (using yarn start):
with:
` {"@context":"https://schema.org","@type":"BreadcrumbList","itemListElement":[{"@type":"ListItem","position":1,"name":"Testumgebung","item":"/testumgebung/en"},{"@type":"ListItem","position":2,"name":"Children, youth and family","item":"/testumgebung/en/children-youth-and-family"},{"@type":"ListItem","position":3,"name":"School for children from 6 to 18 years","item":"/testumgebung/en/children-youth-and-family/school-for-children-from-the-age-of-6-to-15"}]}`

without:
`{"@context":"https://schema.org","@type":"BreadcrumbList","itemListElement":[{"@type":"ListItem","position":1,"name":"Testumgebung","item":"http://localhost:9000/testumgebung/en"},{"@type":"ListItem","position":2,"name":"Children, youth and family","item":"http://localhost:9000/testumgebung/en/children-youth-and-family"},{"@type":"ListItem","position":3,"name":"School for children from 6 to 18 years","item":"http://localhost:9000/testumgebung/en/children-youth-and-family/school-for-children-from-the-age-of-6-to-15"}]}`

It would be nice if we could make the naming clear nevertheless (either `path` if it is a path or `url` if it is an url).

charludo (approved):
Feels a lot cleaner and more purposeful than the previous version.

Minor, but the only thing I dislike is the `>` symbol in front of the city name, especially when navigating back up from a level below.

For example, these breadcrumbs make intuitive sense:
image

But the breadcrumbs for the parent category do not, especially since this is the only time a breadcrumb starts with `>`:
image

IMO, the house symbol should always be shown to the left of the city name, and `>` should never be shown as the first symbol in a breadcrumb. The former also makes the abbreviation "city name"-> "house symbol" more sensical, while the latter makes it seem like there's a crumb missing to the left.

web/src/routes/CategoriesPage.tsx
f1sh1918: alright i added `item: urlFromPath(breadcrumb.link),` even i couldn't reproduce.
Well i renamed link to pathname. Didn't understand if there is more to do here or it was just a general hint (path/url)

web/src/models/BreadcrumbModel.ts
```diff
@@ -21,7 +21,7 @@ export default class BreadcrumbModel {
   }
 
   get link(): string {
```
steffenkleinle, line 23: Should be renamed as well I guess :)

steffenkleinle (approved):

steffenkleinle:
Not tested again

PR #751, opened 2022-08-23
IGAPP-1072: Show excerpt for search results for title match

SteffiStuffel:


f1sh1918 (changes requested):
Tested on web. I do not get an excerpt when title matching
image
I think we agreed on numberOfChars (TextTruncator) not NumberOfWords for the excerpt

Tested on ios. Excerpt was shown correctly for matching title :) just the textTruncator is missing

SteffiStuffel:
> Tested on web. I do not get an excerpt when title matching image I think we agreed on numberOfChars (TextTruncator) not NumberOfWords for the excerpt
> 
> Tested on ios. Excerpt was shown correctly for matching title :) just the textTruncator is missing

Oke this is really weird. I tried and I see the excerpt, checked for test and live cms. City of augsburg. Can you please share some more information?




About the textTruncator: It is only used for events and news currently. If we want to use it also for search results, we should open a ticket for it, and change it everywhere. Here I only added an excerpt for a special case and handled it the same way as the other cases handle it. So I do not see this in the scope of this ticket.

f1sh1918 (approved):
works fine now also for web

steffenkleinle:
> About the textTruncator: It is only used for events and news currently. If we want to use it also for search results, we should open a ticket for it, and change it everywhere. Here I only added an excerpt for a special case and handled it the same way as the other cases handle it. So I do not see this in the scope of this ticket.

It is part of the acceptance criteria of the issue to unify those cases. If you feel like this is too much for this ticket or you don't want to do it's fine but then you should change the issue description and create a new issue.

native/src/components/CategoryListItem.tsx
```diff
@@ -91,18 +92,27 @@ const CategoryListItem = ({
...
+      />
+    ) : (
+      <Text>
+        {new ContentMatcher().getContentAfterMatchIndex(
+          category.contentWithoutHtml ?? '',
+          0,
+          2 * NUM_WORDS_SURROUNDING_MATCH
+        )}
+      </Text>
+    )
```
steffenkleinle, line 115: Alternative solution:
``` tsx
  const excerpt =
    contentMatcher.getMatchedContent(query, category.contentWithoutHtml, NUM_WORDS_SURROUNDING_MATCH) ??
    contentMatcher.getContentAfterMatchIndex(category.contentWithoutHtml ?? '', 0, 2 * NUM_WORDS_SURROUNDING_MATCH)
  const content = query && (
    <Highlighter
      searchWords={[query]}
      sanitize={normalizeSearchString}
      textToHighlight={excerpt}
      autoEscape
      highlightStyle={{
        backgroundColor: theme.colors.backgroundColor,
        fontWeight: 'bold',
      }}
    />
```

steffenkleinle:
> Tested on web. I do not get an excerpt when title matching image which returns the result `Ausländerbehörde (Einwohneramt/Einwohnermeldeamt)` in `/testumgebung/de`.

web/src/components/CategoryEntry.tsx
```diff
@@ -117,6 +117,15 @@ const CategoryEntry = ({ category, contentWithoutHtml, subCategories, query }: P
       textToHighlight={textToHighlight}
       highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }}
     />
+  ) : (
+    query && (
+      <p>
+        {new ContentMatcher()
+          .getWords(parseHTML(category.content).slice(0))
```
steffenkleinle, line 124: Why do we need the `.slice(0)`?
~~Also isn't this the same as directly using `contentWithoutHtml`?~~
Edit: I see, we only pass `contentWithoutHtml` if it is a match by content. Imo we should make it uniform and either pass `contentWithoutHtml` for all categories or just parse it for every category here. Since we only need `contentWithoutHtml` for the search and not the regular category list I'd prefer the former.

web/src/components/CategoryEntry.tsx
```diff
@@ -108,7 +108,7 @@ const CategoryEntry = ({ category, contentWithoutHtml, subCategories, query }: P
     />
   )
 
-  const Content = textToHighlight && (
+  const Content = textToHighlight ? (
```
steffenkleinle, line 111: Alternative solution:
``` tsx
  const excerpt = useMemo<string | null>(() => {
    if (!contentWithoutHtml) {
      return null
    }
    const contentMatcher = new ContentMatcher()
    return (
      contentMatcher.getMatchedContent(query, contentWithoutHtml, NUM_WORDS_SURROUNDING_MATCH) ??
      contentMatcher
        .getWords(contentWithoutHtml)
        .slice(0, 2 * NUM_WORDS_SURROUNDING_MATCH)
        .join(' ')
    )
  }, [contentWithoutHtml, query])

...

  const Content = query && (
    <ContentMatchItem
      aria-label={excerpt}
      searchWords={[query]}
      autoEscape
      sanitize={normalizeSearchString}
      textToHighlight={excerpt}
      highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }}
    />
```

steffenkleinle:
Tested on android and web, works fine :+1:

native/src/components/CategoryList.tsx
```diff
@@ -14,6 +14,7 @@ export type CategoryListModelType = {
   thumbnail: string
   path: string
   contentWithoutHtml?: string
+  titleMatch?: boolean
```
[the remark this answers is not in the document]
SteffiStuffel: Unfortunately it is necessary. If it is a title match it should show the beginning of the content (not the sentence in the content where there is also a match - this is also the behavior for web). In web, we do not need this `titleMatch`-prop, as the SearchPage only sets the `contentWithoutHtml` prop if it is not a title match, so we can check for this prop in web, to decide if we show the beginning of the text or not. For native this is handled differently, therefore we need this additional prop, as the information is missing and needs to be passed.

web/src/components/CategoryEntry.tsx
SteffiStuffel: Done

SteffiStuffel:
Created a new ticket for refactoring.

Changed web: contentWithoutHtml moved to SearchPage -> added titleMatch props (similar to native)

steffenkleinle:
Imo if there is a match by title but the query is also found in the content, we should not show the start of the content in the excerpt but instead the content surrounding the match in the content.

So the behavior would be as following:
* Match in content (only content or content and title): Show excerpt surrounding match
* No match in content (only title): Show beginning of content

SteffiStuffel:
> 
> I didn't get that you were using (2) before, therefore sorry for not already addressing that in my first review. If you don't agree with that, I'd ask our UI/UX experts and create a separate issue and approve this one.

I would stick to the current solution, this is also as google handle search results often. If we want to discuss this with UI/UX, maybe we can do this in a separate ticket? let me know if you insist on this change, so I can adjust this.

steffenkleinle (approved):
No, that works :+1: Will open a discussion on mattermost (and depending on the result a separate issue) :)

PR #752, opened 2022-08-30
IGAPP-1055: Improve events and lists

steffenkleinle:

Especially the events list was quite bad performance wise and very laggy. Therefore I switched to the use of FlatLists instead of simple mapping and using PureComponents. This also fixes the problem that no thumbnails are shown in the events list or icons when navigating back to the home screen (https://issues.tuerantuer.org/browse/IGAPP-1056).


steffenkleinle:
Things not working yet:
* Event list items text goes outside of the screen
* Footer (e.g. FeedbackBox) is shown even if not the end of the list

PureComponents were the only way I could really improve the performance with. Therefore I'd just keep that.

@f1sh1918 is it possible to use the flat list in the bottom action sheet as well? For lots of pois the performance won't be good otherwise.

f1sh1918:
> Things not working yet:
> 
> * Event list items text goes outside of the screen
> * Footer (e.g. FeedbackBox) is shown even if not the end of the list
> 
> PureComponents were the only way I could really improve the performance with. Therefore I'd just keep that.
> 
> @f1sh1918 is it possible to use the flat list in the bottom action sheet as well? For lots of pois the performance won't be good otherwise.

1. Things not working: Are you gonna fix or in separate tickets?
2. I would really like to remove the PureComponents and use a React.memo instead and functional components. If you don't want to adjust that here, just create a separate ticket for that.
3. yes you can pass any children to the `BottomActionSheet`. Just create a separate Ticket for that, since its way too much in this pr

steffenkleinle:
>     1. Things not working: Are you gonna fix or in separate tickets?

I hoped one of you could possibly look into it since I don't have a lot of time at the moment. Has to be fixed here I think, releasing this to main is not good.

>     2. I would really like to remove the PureComponents and use a React.memo instead and functional components. If you don't want to adjust that here, just create a separate ticket for that.

I can try again but the performance was a lot worse when I tried it. But maybe I did something else that improved the performance.


>     3. yes you can pass any children to the `BottomActionSheet`. Just create a separate Ticket for that, since its way too much in this pr
Will do.

f1sh1918:
> > ```
> > 1. Things not working: Are you gonna fix or in separate tickets?
> > ```
> 
> I hoped one of you could possibly look into it since I don't have a lot of time at the moment. Has to be fixed here I think, releasing this to main is not good.
> 
> > ```
> > 2. I would really like to remove the PureComponents and use a React.memo instead and functional components. If you don't want to adjust that here, just create a separate ticket for that.
> > ```
> 
> Feel free to try in this PR or in a new ticket but the performance was a lot worse when I tried it. But maybe I did something else that improved the performance.
> 
> > ```
> > 3. yes you can pass any children to the `BottomActionSheet`. Just create a separate Ticket for that, since its way too much in this pr
> > ```
> 
> Will do.

@steffenkleinle  i don't understand what you mean with footer is shown even if you are not at the end of the list. Can not reproduce that.



I converted Pure components to memo and can't see any performance differences.
Maybe you can check current state of the branch or tell me where performance issues occur

steffenkleinle:
> @steffenkleinle i don't understand what you mean with footer is shown even if you are not at the end of the list. Can not reproduce that.

If you go to munichs events and scroll down really fast you see the footer.
> 
> I tried converting to memo and couldn't see any performance difference but recognized the warning to i kept PureComponents.

I don't really mind then if we take the warning or the PureComponents.

steffenkleinle:
@SteffiStuffel please review :)

f1sh1918 (approved):
looks good so far.
@steffenkleinle one question the picture issue will be fixed via cms event limit? Just tested on test-cms and still recognized this issue

steffenkleinle:
> looks good so far. @steffenkleinle one question the picture issue will be fixed via cms event limit? Just tested on test-cms and still recognized this issue

Which issue do you mean? Pictures not showing up for the events? I think this is just for the test cms because the pics do not exist in the cms anymore, its the same way for web and not reproducable with live cms. The problem described by martin in the issue though should be fixed by the performance improvements in this PR. Let me know if I am wrong herr though :)

native/src/components/SprungbrettListItem.tsx
```diff
@@ -1,24 +1,22 @@
...
-class SprungbrettListItem extends React.PureComponent<PropsType> {
-  render(): ReactNode {
-    const { language, job, openJobInBrowser, theme } = this.props
+// This should stay a PureComponent for performance reasons
+class SprungbrettListItem extends PureComponent<Props> {
+  render(): ReactElement {
+    const { language, job, openJobInBrowser } = this.props
     return (
-      <ListItem thumbnail={null} title={job.title} navigateTo={openJobInBrowser} theme={theme} language={language}>
+      <ListItem thumbnail={null} title={job.title} navigateTo={openJobInBrowser} language={language}>
```
SteffiStuffel, line 19: This is not working for me on Android emulator. When trying to open a sprungbrett offer i only get an error

native/src/routes/Licenses.tsx
```diff
@@ -61,14 +61,15 @@ const Licenses = (): ReactElement => {
...
+    const { licenses, name, licenseUrl, version } = item
+    const openLink = () => openExternalUrl(licenseUrl).catch(() => showSnackbar(t('error:unknownError')))
+    return <LicenseItem key={name} name={name} version={version ?? ''} license={licenses} onPress={openLink} />
   }
 
   return (
     <Layout>
       <Caption title={t('openSourceLicenses')} />
-      <FlatList data={licenses} renderItem={renderItem} keyExtractor={item => item.name} />
+      <FlatList data={licenses} renderItem={renderItem} />
```
SteffiStuffel, line 72: Routes are not opening here also. Just an error.

SteffiStuffel (approved):
Bug about not opening external links also occurs on other branch, so I guess it is because of my emulator. Maybe someone could check nevertheless this is only a problem I have.

native/src/components/SprungbrettListItem.tsx
SteffiStuffel: Ah yes you are right, i remember now. Thanks for info.

steffenkleinle:
> Bug about not opening external links also occurs on other branch, so I guess it is because of my emulator. Maybe someone could check nevertheless this is only a problem I have.

Just because of the inappbrowser bug not working on android emulators.

PR #758, opened 2022-09-05
IGAPP-499: Improve no internet connection error

charludo:


charludo:
Some new translations might be needed (see comment on the issue tracker)

web/src/components/FailureSwitcher.tsx
```diff
@@ -29,6 +29,7 @@ const FailureSwitcher = ({ error }: PropsType): ReactElement => {
   }, [error])
 
   const getFailureProps = (error: Error): { goToPath?: string; goToMessage?: string; errorMessage: string } => {
+    /* eslint-disable @typescript-eslint/no-unnecessary-condition */
     if (error instanceof NotFoundError && error.city !== undefined && error.language !== undefined) {
```
charludo, line 33: I really don't like disabling linter errors, but I have no clue why this one happens in the first place:

`error.city` and `error.language` *can* be undefined, as per `NotFoundError`'s type. Linter complains that there's no overlap between the first, and each of the last two conditionals.

If someone knows what's up, I'd love to remove the disable-line...

SteffiStuffel: You do not need this comment. Linting only fails when you run `yarn lint` in `/web`-subfolder, as the types are from the api-client. If you run linting in parent directory, everything works fine, pipeline should also succeed.

SteffiStuffel (approved):
Tested "no internet error" on android emulator. Looks very nice :tada:

web/src/components/FailureSwitcher.tsx
charludo: Ah, makes sense. Thanks.

steffenkleinle (approved):
Tested on android

PR #759, opened 2022-09-06
IGAPP-1073: yarn 2

SteffiStuffel:


.circleci/src/commands/restore_yarn_cache.yml
```diff
@@ -3,15 +3,12 @@ steps:
...
   - save_cache:
       name: Save Yarn Package Cache
-      key: 8-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}-{{ checksum "native/package.json" }}-{{ checksum "web/package.json" }}
+      key: 8-yarn-3-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}-{{ checksum "native/package.json" }}-{{ checksum "web/package.json" }}
       paths:
-        - ~/.cache/yarn
-        - node_modules
-        - native/node_modules
-        - web/node_modules
+        - .yarn/cache386
```
f1sh1918, line 14: @SteffiStuffel  why `cache386`

SteffiStuffel: This is a draft. I was trying around, if a invalid path to cache will cause the same or another error, as yarn2 is working fine lcoally but not in the pipeline. :-/

PR #761, opened 2022-09-07
IGAPP-1087: Fix PDF on Android

SteffiStuffel:

### DIFF
--- a/native/src/routes/PDFViewModal.tsx
+++ b/native/src/routes/PDFViewModal.tsx
@@ -38,6 +38,7 @@ const PDFViewModal = ({ route, navigation: _navigation }: PropsType): ReactEleme
         source={{
           uri: url,
         }}
+        trustAllCerts={false}
         onError={() => setError(true)}
       />
     </View>
--- a/release-notes/unreleased/IGAPP-1087-fix-pdf-on-android.yml
+++ b/release-notes/unreleased/IGAPP-1087-fix-pdf-on-android.yml
@@ -1,0 +1,6 @@
+issue_key: IGAPP-1087
+show_in_stores: true
+platforms:
+  - android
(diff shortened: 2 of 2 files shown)


native/src/routes/PDFViewModal.tsx
```diff
@@ -38,6 +38,7 @@ const PDFViewModal = ({ route, navigation: _navigation }: PropsType): ReactEleme
         source={{
           uri: url,
         }}
+        trustAllCerts={false}
```
steffenkleinle, line 41: Could you please explain this? Why does this help?

release-notes/unreleased/IGAPP-1087-fix-pdf-on-android.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1087
+show_in_stores: true
+platforms:
+  - android
```
steffenkleinle, line 4: Didn't happen on ios?

steffenkleinle (approved):
Tested on android.

release-notes/unreleased/IGAPP-1087-fix-pdf-on-android.yml
SteffiStuffel: No only an android issue

native/src/routes/PDFViewModal.tsx
SteffiStuffel: Some bug in react-native-pdf or react-native-fetch-blob. Should be default be false, but does not work correctly. See here:
https://github.com/wonday/react-native-pdf/issues/460

steffenkleinle: Thanks! According to the docs the default is true.

PR #764, opened 2022-09-17
IGAPP-1106: Bottom action sheet active poi

f1sh1918:

1. Select a poi on the map. Bottom Action Sheet should snap to 35% and marker should be centered


dkehne (approved):
35% is a good value to work with. Awesome job @f1sh1918 !

steffenkleinle (approved):
Not sure if it was that way before and whether it is a problem but some things I observed:
* (native) If the bottom sheet uses the 95% snappoint and you select a poi in the list, the map does not immediately center around the poi but only once the bottom sheet is minimized.Not the case for web.
* (native) If you move the map, close the bottom sheet and open it again it centers on the poi again. Not the case for we

Tested on Android and Firefox Android.

SteffiStuffel (approved):

steffenkleinle:
@f1sh1918 please take a look at my comment before merging. I think some behavior is not optimal but may be fixed in a separate issue (or not at all if you disagree).

f1sh1918:
> @f1sh1918 please take a look at my comment before merging. I think some behavior is not optimal but may be fixed in a separate issue (or not at all if you disagree).

Can you give me a more precise hint what you are talking about? which behavior? I can not find any issues

@steffenkleinle i fixed your previous annotations and increased the maxSnapPoint to 100% and removed borderRadius to look nicer. I think there is no use to spare out 5% because the backButton is on the navigationHeader

PR #765, opened 2022-09-20
IGAPP-1094: Add social media preview placeholder

SteffiStuffel:


steffenkleinle (commented):
Is this already set up correctly in the cms/server config? Or will this break existing functionality til then?

Not sure and maybe we should just keep it but it may make sense to remove the stuff in the Helmet that also sets meta data after rendering. I think we should discuss that first though. However, I think it would make sense to add a comment there that metadata is also set here or rather by the cms.

SteffiStuffel:
Will close until discussing with cms is finished

PR #766, opened 2022-09-20
IGAPP-1075: Adjust testing rendering function

SteffiStuffel:


web/package.json
```diff
@@ -98,6 +98,7 @@
     "@types/styled-components": "^5.1.25",
     "@types/testing-library__jest-dom": "^5.14.5",
     "@types/webpack-bundle-analyzer": "^4.4.1",
+    "@types/react-router-dom": "^5.3.3",
```
steffenkleinle, line 101: I think this should not be added again, afaik react-router is written in typescript since v6 (types are also indicating that since its still v5).

steffenkleinle (approved):
Not tested

web/package.json
SteffiStuffel: Good point. Thanks for the hint. Was also wondering about v5. Removed it.

f1sh1918 (approved):

PR #768, opened 2022-09-20
IGAPP-1108: Lock scrolling open sidebar 

f1sh1918:


web/src/components/KebabMenu.tsx
```diff
@@ -78,19 +79,28 @@ const ToggleButton = styled.button`
 
 const KebabMenu = ({ items, direction }: KebabMenuProps): ReactElement => {
   const [checked, setChecked] = useState<boolean>(false)
+  const { locked, setLocked } = useLockedBody(checked)
+
+  const onClick = () => {
+    if (!locked) {
+      window.scrollTo(0, 0)
```
SteffiStuffel, line 86: So when opening the the header, scroll gets reset. This may be annoying, when reading a text, check if it is available in an other language a user prefers, and the wants to go on reading.

I think it would be nicer if it does not get reset.

SteffiStuffel (approved):

web/src/components/KebabMenu.tsx
f1sh1918: but this won't work without scrolling because the sidebar is absolute positioned on the top of the page.

Well we could save the scroll position and if you just open and close the sidebar you can restore the former position but if you forward the page or switch the language it won't work. Different language may lead to different amount of chars

SteffiStuffel: Can the sidebar be sticky like the header? Seems most suitable solution for me, but not sure if possible with reasonable efford.

f1sh1918: @SteffiStuffel found a way to position the sidebar according to the scrollPosition. please test again

SteffiStuffel: Much smoother now, but if the language gets changed, i guess scroll position should be reset. So only if the sidebar is closed by using the x-icon, the scrollPosition should be same, if the content changes, e.g. language change, the scrollPostion should be resetted.

charludo (approved):
Works nicely!

web/src/components/KebabMenu.tsx
f1sh1918: @SteffiStuffel i had to update a lot to get this working properly. so now the scroll to top is implemented on every route change and the closeSidebar are implemented on change language.

SteffiStuffel (approved):

PR #771, opened 2022-09-28
IGAPP-1095: Poi clustering web

f1sh1918:

Munich is the city with the highest amount of pois. Best option to test here:
http://localhost:9000/muenchen/de/locations


web/src/constants/layers.ts
```diff
@@ -0,0 +1,63 @@
...
+  type: 'circle',
+  source: 'point',
+  filter: ['has', 'point_count'],
+  paint: {
+    'circle-color': [
+      'step',
+      ['get', 'point_count'],
+      `${theme.colors.themeColor}`,
+      groupCount,
+      `${theme.colors.themeColor}`,
```
steffenkleinle, line 25: Any specific need for the template string?
```suggestion
      theme.colors.themeColor,
      groupCount,
      theme.colors.themeColor,
```

steffenkleinle (approved):
Tested on firefox, looks nice :) There is a short time period in which both pois and clusters are shown. I think it is good enough as it currently is implemented though.

https://user-images.githubusercontent.com/29095487/193449739-5ae922d0-3be7-4dcd-9ed0-f311ebda033e.mp4

SteffiStuffel (approved):

PR #775, opened 2022-10-04
IGAPP-1064: Navigation keep scroll position

f1sh1918:

1. Go to mobile mode (select small device that needs scrolling of navigation items)
2. Select navigation item or http://localhost:9000/testumgebung/de/news/local


web/jest.setup.ts
```diff
@@ -52,3 +52,6 @@ Object.defineProperty(URL, 'createObjectURL', {
   writable: true,
   value: jest.fn(),
 })
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function -- scrollTo doesn't exist in JSDOM
+Element.prototype.scrollTo = () => {}
```
steffenkleinle, line 57:
```suggestion
Element.prototype.scrollTo = () => undefined
```

web/src/components/Header.tsx
```diff
@@ -140,8 +141,11 @@ export const Header = ({
           </ActionBar>
         </Row>
         {hasNavigationBar && (
-          <NavigationBarScrollContainer scrollContainerRef={scrollContainerRef} direction={direction}>
-            <NavigationBar>{navigationItems}</NavigationBar>
+          <NavigationBarScrollContainer
+            scrollContainerRef={scrollContainerRef}
+            direction={direction}
+            activeIndex={navigationItems.findIndex(el => el.props.active)}>
+            <NavigationBar id='navbar'>{navigationItems}</NavigationBar>
```
steffenkleinle, line 148:
```suggestion
            <NavigationBar id='navigation-bar'>{navigationItems}</NavigationBar>
```
or, depending on conventions
```suggestion
            <NavigationBar id='navigationBar'>{navigationItems}</NavigationBar>
```

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -59,8 +60,26 @@ const getScrollableWidth = (scrollContainer: RefObject<HTMLDivElement>): number
...
-const NavigationBarScrollContainer = ({ children, direction, scrollContainerRef }: PropsType): ReactElement => {
-  const [scrollPosition, setScrollPosition] = useState(0)
+const NavigationBarScrollContainer = ({
+  children,
+  direction,
+  scrollContainerRef,
+  activeIndex,
+}: PropsType): ReactElement => {
+  const getInitScrollPosition = useCallback((): number => {
+    const navBar = document.getElementById('navbar')
```
steffenkleinle, line 70: Personal opinion: I don't like these kind of abbreviations. Instead, I would prefer using the complete name, especially if the component is named differently.
```suggestion
    const navigationBar = document.getElementById('navbar')
```

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -59,8 +60,26 @@ const getScrollableWidth = (scrollContainer: RefObject<HTMLDivElement>): number
...
 
-const NavigationBarScrollContainer = ({ children, direction, scrollContainerRef }: PropsType): ReactElement => {
-  const [scrollPosition, setScrollPosition] = useState(0)
+const NavigationBarScrollContainer = ({
+  children,
+  direction,
+  scrollContainerRef,
+  activeIndex,
+}: PropsType): ReactElement => {
+  const getInitScrollPosition = useCallback((): number => {
```
steffenkleinle, line 69:
```suggestion
  const getInitialScrollPosition = useCallback((): number => {
```

steffenkleinle: Why does this have to be a callback? It's only used once during initial render.

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -59,8 +60,26 @@ const getScrollableWidth = (scrollContainer: RefObject<HTMLDivElement>): number
...
+    const navBar = document.getElementById('navbar')
+    if (!navBar) {
+      return 0
+    }
+    const navBarOffset = navBar.offsetLeft
+    const elementOffset = navBar.getElementsByTagName('div')[activeIndex]?.offsetLeft
+    return elementOffset ? navBarOffset + elementOffset : 0
+  }, [activeIndex])
+  const [scrollPosition, setScrollPosition] = useState<number>(activeIndex > 0 ? getInitScrollPosition() : 0)
```
steffenkleinle, line 78: This is never updated, is it? So there should be no need to save it in a state.

steffenkleinle (changes requested):
Not sure if this is because of this PR but scrolling the navigation bar does not work smoothly, instead you can always just scroll vertically very little until it stops.

Also, as @LeandraH pointed out, when selecting a navigation item, it jumps back to scroll position 0. Afterwards, the scroll position seems to be kept however.

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: > When testing in Firefox with a screen width of 372px, I can only see the very beginning of the selected icon (which is what I see at scroll 0).
> 
> image

okay looks like the solution is only working on chrome...have to check

f1sh1918:
> Not sure if this is because of this PR but scrolling the navigation bar does not work smoothly, instead you can always just scroll vertically very little until it stops.
> 
> Also, as @LeandraH pointed out, when selecting a navigation item, it jumps back to scroll position 0. Afterwards, the scroll position seems to be kept however.

gonna check this with firefox. for chrome it reloads the page and scrolls smoothly to the correct position. Maybe we should think about giving each item the same width and calculate the position instead of using the offset which seems to work not very well tbh @steffenkleinle

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: forgot to readjust since i used it in a useEffect hook before. thx

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: @steffenkleinle i adjusted some things and added a custom hook to deal with ref changes. Should work now also on forefox

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -51,16 +52,34 @@ const ScrollContainer = styled.div<{ showArrowContainer: boolean }>`
...
+  const [scrollPosition, setScrollPosition] = useState<number>(0)
+  const { toggle, refCallback, ref: scrollContainerRef } = useRefWithCallback()
 
-const NavigationBarScrollContainer = ({ children, direction, scrollContainerRef }: PropsType): ReactElement => {
-  const [scrollPosition, setScrollPosition] = useState(0)
+  useEffect(() => {
+    if (activeIndex > 0 && scrollContainerRef.current) {
+      scrollContainerRef.current.scrollTo({ left: getInitialScrollPosition(activeIndex) })
+    }
+  }, [activeIndex, scrollContainerRef, toggle])
```
f1sh1918, line 82: this is used instead of unsafe `ref.current` check`useRefWithCallback` hook

LeandraH (approved):
Very nice! Also works in Firefox now :)

web/src/components/NavigationBarScrollContainer.tsx
steffenkleinle: Nice idea. I think this could be improved a little further though, there is no need for the toggle or the useEffect, instead we can just pass the callback to the hook and execute it directly there. Furthermore, I did some minor adjustments and renamed the hook to `useCallbackRef` in accordance with the react docs.
See #780. Feel free to reject it completely or to just apply some changes.

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -51,16 +52,34 @@ const ScrollContainer = styled.div<{ showArrowContainer: boolean }>`
...
+const getScrollableWidth = <T extends HTMLSpanElement | HTMLDivElement | HTMLParagraphElement>(
+  scrollContainer: RefObject<T>
+): number => {
   if (scrollContainer.current) {
     const { scrollWidth, clientWidth } = scrollContainer.current
     return scrollWidth - clientWidth
   }
   return 0
 }
+const getInitialScrollPosition = (activeIndex: number): number => {
```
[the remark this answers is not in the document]
f1sh1918: fixed in last commit

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -51,16 +52,34 @@ const ScrollContainer = styled.div<{ showArrowContainer: boolean }>`
...
+  const navigationBar = document.getElementById('navigation-bar')
+  if (!navigationBar) {
+    return 0
+  }
+  const navigationBarOffset = navigationBar.offsetLeft
+  const elementOffset = navigationBar.getElementsByTagName('div')[activeIndex]?.offsetLeft
```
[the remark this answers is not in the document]
f1sh1918: we can not use id but `getElementsByClassName` so i added a className `navigation-item`

web/src/components/HeaderNavigationItem.tsx
```diff
@@ -129,7 +129,7 @@ export type HeaderNavigationItemProps = {
 }
 
 const HeaderNavigationItem = ({ active, text, href, icon, direction }: HeaderNavigationItemProps): ReactElement => (
-  <Container direction={direction}>
+  <Container direction={direction} className='navigation-item'>
```
steffenkleinle, line 132: Optional: Would prefer 
```suggestion
  <Container direction={direction} className='header-navigation-item'>
```

web/src/components/NavigationBarScrollContainer.tsx
```diff
@@ -53,22 +53,22 @@ const ScrollContainer = styled.div<{ showArrowContainer: boolean }>`
...
+const getActiveItemScrollPosition = (activeIndex: number): number => {
   const navigationBar = document.getElementById('navigation-bar')
   if (!navigationBar || activeIndex === 0) {
     return 0
   }
   const navigationBarOffset = navigationBar.offsetLeft
-  const elementOffset = navigationBar.getElementsByTagName('div')[activeIndex]?.offsetLeft
-  return elementOffset ? navigationBarOffset + elementOffset : 0
+  const navigationItem = navigationBar.getElementsByClassName('navigation-item')[activeIndex] as HTMLElement | null
+  return navigationItem?.offsetLeft ? navigationItem.offsetLeft - navigationBarOffset : 0
```
steffenkleinle, line 63: Optional:
``` ts
  const navigationBar = document.getElementById('navigation-bar')
  const navigationItem = navigationBar?.getElementsByClassName('navigation-item')[activeIndex]
  if (!navigationBar || !(navigationItem instanceof HTMLElement)) {
    return 0
  }
  return navigationItem.offsetLeft - navigationBar.offsetLeft
```

steffenkleinle: At least the check for `activeIndex === 0` is not needed anymore and should be removed thanks to your adjustments in calculating the scroll position.

steffenkleinle (approved):
Tested on firefox, nice work!

web/src/components/HeaderNavigationItem.tsx
f1sh1918: fixed in last commit

web/src/components/NavigationBarScrollContainer.tsx
f1sh1918: fixed

steffenkleinle: Check for `navigationItem?.offsetLeft` is not necessary. Not sure why our linitng does not mark that though

PR #776, opened 2022-10-04
IGAPP-1113: Map link in pdf

f1sh1918:

Kreis Olpe -> Beratungsstellen und Hilfsangebote -> Beratungsstellen Integration und Migration -> open PDF -> klick on GMaps Icon

Tested on android & ios

### DIFF
--- a/native/src/routes/PDFViewModal.tsx
+++ b/native/src/routes/PDFViewModal.tsx
@@ -45,7 +45,7 @@ const PDFViewModal = ({ route, navigation: _navigation }: PropsType): ReactEleme
         }}
         trustAllCerts={false}
         onError={() => setError(true)}
-        onPressLink={url => openExternalUrl(url).catch(() => showSnackbar(t('urlError')))}
+        onPressLink={url => openExternalUrl(url).catch(() => showSnackbar(t('unknownError')))}
       />
     </View>
   )
--- a/native/src/routes/PDFViewModal.tsx
+++ b/native/src/routes/PDFViewModal.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import { View } from 'react-native'
 import Pdf from 'react-native-pdf'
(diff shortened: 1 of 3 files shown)


LeandraH (approved):
Looks good, tested in Firefox

native/src/routes/PDFViewModal.tsx
```diff
@@ -40,9 +45,9 @@ const PDFViewModal = ({ route, navigation: _navigation }: PropsType): ReactEleme
         }}
         trustAllCerts={false}
         onError={() => setError(true)}
+        onPressLink={uri => openExternalUrl(uri).catch(() => showSnackbar(t('linkError')))}
```
steffenkleinle, line 48: We use `url` for naming everywhere. Even though this may not be completely correct, I would prefer that here.
```suggestion
        onPressLink={url => openExternalUrl(url).catch(() => showSnackbar(t('urlError')))}
```

sarahsporck (commented):
Does this issue only appear on real devices? On the emulator everything worked fine for me...

translations/translations.json
```diff
@@ -1134,6 +1134,7 @@
       "networkRequestFailed": "The request could not be processed.",
       "responseMappingFailed": "An error occurred while processing the server's response.",
       "unknownError": "An unknown error occurred.",
+      "linkError": "Link could not be opened",
```
steffenkleinle, line 1137: Please add a german translation if you want to use a custom translation. However, I would just use `unknownError` instead.
```suggestion
      "linkError": "Link could not be opened.",
```

steffenkleinle (commented):
Tested on android :+1: 

@sarahsporck at least on real device opening links from pdfs does not work for me without this.

translations/translations.json
f1sh1918: i take the unknow error and remove this line

f1sh1918:
> Does this issue only appear on real devices? On the emulator everything worked fine for me...

hm didn't check on emulator before tbh. but on real devices it didn't work.

PR #777, opened 2022-10-05
IGAPP-1120: Fix for Jumping Chat Widget on Pages with a Scrollbar

LeandraH:

The issue is https://issues.tuerantuer.org/browse/IGAPP-1120

To test:
1. Go to http://localhost:9000/muenchen/de
2. Clicking back and forth between Veranstaltungen and Lokale Informationen, the chat widget should not jump anymore when the scrollbar appears


web/src/components/ChatBotWidget.tsx
```diff
@@ -1,5 +1,11 @@
 import React, { ReactElement, useEffect } from 'react'
-import { useTheme } from 'styled-components'
+import styled, { useTheme } from 'styled-components'
+
+const ChatContainer = styled.div`
+  > div {
+    right: calc(100% - 100vw + 33px);
```
f1sh1918, line 6: where does this 33px come from?

LeandraH: It's a magic number consisting of the average 17px that a scroll bar usually has as width + the original 16px that the box had as padding.

web/src/components/ChatBotWidget.tsx
```diff
@@ -1,5 +1,11 @@
 import React, { ReactElement, useEffect } from 'react'
-import { useTheme } from 'styled-components'
+import styled, { useTheme } from 'styled-components'
+
+const ChatContainer = styled.div`
+  > div {
```
f1sh1918, line 5: i think this should be wrapped in a media query, since its only needed for tablet/desktop. for mobile it just replaces the icon
```js
  @media ${dimensions.mediumLargeViewport} {
    > div {
      right: calc(100% - 100vw + 33px);
    }
  }
  ```

f1sh1918 (commented):
Tested on firefox and chrome. just small adjustments. nice work ! :)

sarahsporck (approved):
Nice! Tested on Firefox and Chrome.

sarahsporck (approved):

PR #778, opened 2022-10-05
IGAPP-1126: Event title and overflow

sarahsporck:

### DIFF
--- a/web/src/components/EventListItem.tsx
+++ b/web/src/components/EventListItem.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react'
+import styled from 'styled-components'
 
 import { DateFormatter, EventModel } from 'api-client'
 
@@ -9,6 +10,10 @@ import { textTruncator } from '../utils/stringUtils'
 import ListItem from './ListItem'
 
 export const NUM_OF_CHARS_ALLOWED = 110
+
+const Content = styled.div`
+  overflow-wrap: anywhere;
+`
 
 type PropsType = {
   event: EventModel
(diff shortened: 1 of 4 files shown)


web/src/components/EventListItem.tsx
```diff
@@ -10,6 +11,10 @@ import ListItem from './ListItem'
 
 export const NUM_OF_CHARS_ALLOWED = 110
 
+const Content = styled.div`
+  overflow-wrap: anywhere;
```
sarahsporck, line 15: `break-word` did not wrap `veeeeeeeeeeeeeeeeeeeery long` text.

web/src/components/SprungbrettListItem.tsx
```diff
@@ -1,23 +1,23 @@
...
-        <div dir='auto'>{job.location}</div>
-      </ListItem>
-    )
-  }
-}
+const SprungbrettListItem = ({ job }: PropsType): ReactElement => (
+  <ListItem title={job.title} path={job.url}>
+    <Content dir='auto'>{job.location}</Content>
+  </ListItem>
+)
```
f1sh1918, line 21: Any reason to not keep `PureComponent` f.e. with
`export default memo(SprungbrettListItem)`
hard to measure performance but memo may increase the performance slightly

f1sh1918 (approved):
Tested on chrome & firefox. Works fine

web/src/components/SprungbrettListItem.tsx
steffenkleinle: I think on web we don't really have performance problems, especially for Sprungbrett since the list items are very simple. So I think we should be good without.

steffenkleinle (approved):
Tested on firefox

PR #779, opened 2022-10-07
IGAPP-1116: Prevent no network error

charludo:

Suggestion for how to handle attempts to change language while offline.


steffenkleinle:
Closing since this is currently not possible. Once #802 is done this can be fixed.

PR #782, opened 2022-10-10
IGAPP-1122 Hide Chat on Location Route

LeandraH:

To test:
1. Go to http://localhost:9000/muenchen/de/locations
2. See that there is no chatbot icon
3. http://localhost:9000/muenchen/de/news/local
4. See that there is a chatbot icon


steffenkleinle (approved):

Not 100% sure but I think I would prefer an optional prop `showChatbot = true` instead of the if check in location layout. I think the power of deciding what to show should be part of the routes responsibilities and not the layouts. What do you think? Other opinions @sarahsporck @SteffiStuffel @f1sh1918?

release-notes/unreleased/IGAPP-1122-hide-chatbot-on-maps.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: IGAPP-1122
+show_in_stores: true
+platforms:
+  - web
+en: Hide the Icon for the Chatbot on the Locations Route
```
f1sh1918, line 5:
```suggestion
en: Hide the icon for the chatbot on the locations route
```

f1sh1918 (approved):
Tested on chrome :)

LeandraH:
> Tested on firefox, works as expected.
> 
> Not 100% sure but I think I would prefer an optional prop `showChatbot = true` instead of the if check in location layout. I think the power of deciding what to show should be part of the routes responsibilities and not the layouts. What do you think? Other opinions @sarahsporck @SteffiStuffel @f1sh1918?

Hmmm, good question, I'm more inclined to leave it in the layout because the reason for the chatbot being hidden is purely visual, and if we change the design in the future, we can probably also add the chatbot back in. But it's not a hill that I'm willing to die on :D

sarahsporck:
> > Tested on firefox, works as expected.
> > Not 100% sure but I think I would prefer an optional prop `showChatbot = true` instead of the if check in location layout. I think the power of deciding what to show should be part of the routes responsibilities and not the layouts. What do you think? Other opinions @sarahsporck @SteffiStuffel @f1sh1918?
> 
> Hmmm, good question, I'm more inclined to leave it in the layout because the reason for the chatbot being hidden is purely visual, and if we change the design in the future, we can probably also add the chatbot back in. But it's not a hill that I'm willing to die on :D

I think I agree with Leandra with this. Her proposed solution is very simple and not very intrusive. Modifying every route for this seems a little bit overkill when there is no reason to do that right now or in the near future. and if a reason arises at some point, this can be change very easily. No reason to create more work fo this imo. :)

PR #783, opened 2022-10-10
IGAPP-1096: Poi clustering native

f1sh1918:

Testing: Most poi markers can be found in "Landeshauptstadt München"


native/src/constants/layers.ts
```diff
@@ -0,0 +1,54 @@
...
+  fontSizeLarge,
+  fontSizeSmall,
+  groupCount,
+  mapMarker,
+  PoiFeature,
+  textOffsetY,
+} from 'api-client'
+import { ThemeType } from 'build-configs/ThemeType'
+
+export const clusterLayer = (theme: ThemeType): CircleLayerProps => ({
```
steffenkleinle, line 15: Styling differs between web and native, right? No way to make this more uniform/reuse stuff?

steffenkleinle (approved):
Tested on android, looks good :+1:

native/src/constants/layers.ts
f1sh1918: Well the libs using different interfaces for the layers. But the result should be quite same. I first thought that we can unify the layers but thats not possible unfortunately

sarahsporck (approved):
You were really fast on this one :)
Tested on the android emulator 👌🏼.

PR #787, opened 2022-10-11
IGAPP-992: Add a Footer to the Sidebar

LeandraH:

To test, open the app in a mobile browser and click on the kebab icon.


web/src/components/KebabMenu.tsx
```diff
@@ -110,10 +114,23 @@ const KebabMenu = ({ items, direction, show, setShow }: KebabMenuProps): ReactEl
...
+- Get items
+  - Datenschutz
+  - Ueber Integreat
+  - Impressum und Kontakt
+
+
+TODO:
+- Style them
+- Put them in a separate file
+*/
```
steffenkleinle, line 134: Is this done?

web/src/components/KebabFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { aboutUrls, privacyUrls } = buildConfig()
+  const { t } = useTranslation('layout')
+  const aboutUrl = aboutUrls[language] || aboutUrls.default
+  const privacyUrl = privacyUrls[language] || privacyUrls.default
+
+  return (
+    <Footer>
+      <StyledLink to={privacyUrl}>{t('privacy')}</StyledLink>
+      <StyledLink to={aboutUrl}>{t('settings:about', { appName: buildConfig().appName })}</StyledLink>
+      <StyledLink to={RoutePatterns[MAIN_DISCLAIMER_ROUTE]}>{t('imprintAndContact')}</StyledLink>
```
steffenkleinle, line 44: You are missing the open source licenses. Also, for city content routes, the disclaimer of the city instead of the main disclaimer should be shown. See the `LocationFooter` for reference. Please also keep that ordering.

web/src/components/KebabFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  &:last-child {
+    border-bottom: none;
+  }
+`
+
+type PropsType = {
+  language: string
+}
+
+const KebabFooter = ({ language }: PropsType): ReactElement => {
```
steffenkleinle, line 34: I think we should not implement everything twice. Instead it would be nice if we could reuse the `LocationFooter`, for example by just passing a prop `inSidebar` or similar and applying the necessary styling based on that.

steffenkleinle (changes requested):
Good work already!

Two remarks:
1. The footer should definitely not be shown twice on mobile now as it currently is (once in the sidebar and once at the end of the page). The one on the end of the page should be removed.
2. I think this time it should really be handled somewhere in the `LocationLayout` where to show the footer (bottom of page or sidebar) to avoid exactly case 1. Otherwise, there are multiple pages responsible for that. As a solution it may work to just add a prop `Footer` to the `KebabMenu` that is passed through from the `LocationLayout`.
3. I don't really like the black separation lines but that is probably a personal opinion, don't have any ideas how to make it better. I think we should check with UI/UX before though since we don't have a design for this, right? Maybe also @f1sh1918 has some ideas.

@f1sh1918 @SteffiStuffel @sarahsporck do you think we should rename all the `LoationXYZ` stuff to `CityContentXYZ`? e.g. rename `LocationHeader` to `CityContentHeader`? Because right now we are quite inconsistent here.

LeandraH:
Sorry, my bad, the design is here: https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Design-Tickets-Jira (it does come with black lines)

web/src/components/KebabMenu.tsx
LeandraH: Wooooops

steffenkleinle:
> Sorry, my bad, the design is here: https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Design-Tickets-Jira (it does come with black lines)

Oh :see_no_evil: Definitely not your bad, just missed it in the issue! Sorry for that :cry: Then this is totally fine :)

web/src/components/KebabFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  border-bottom: 1px solid ${props => props.theme.colors.textColor};
+  &:first-child {
+    margin-top: -10px;
+  }
+  &:last-child {
+    border-bottom: none;
+  }
+`
+
+type PropsType = {
```
f1sh1918, line 30: pls rename to `KebabFooterProps`

f1sh1918 (dismissed):
Links for About, Datenschutz etc. do not work correctly, pls fix.
`href="/testumgebung/de/https://integreat-app.de/about/"`

According the design of the borders, i'm gonna open a thread in ui/ux channel probably my fault with the missing review. I would suggest to use the `textDecorationColor: '#c7c7c7'`
Lets wait for the ux team response

Nice refactoring :)

web/src/components/CityContentFooter.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import { DISCLAIMER_ROUTE, LICENSES_ROUTE, pathnameFromRouteInformation } from 'api-client'
+
+import buildConfig from '../constants/buildConfig'
+import CleanLink from './CleanLink'
+import Footer from './Footer'
+
+const SidebarFooterContainer = styled.div`
+  width: 100%;
+  margin: -10px 12px 0 23px;
```
f1sh1918, line 13: this looks like magic :)
i think instead of that margin i would use a padding. Even the spacings don't look equal

image
```suggestion
  padding: 0 24px; 
```

web/src/components/CityContentFooter.tsx
```diff
@@ -0,0 +1,77 @@
...
+  > * {
+    color: ${props => props.theme.colors.textColor};
+    height: 48px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding-right: 5.5px; /* the text is centered, the lines between are not */
```
f1sh1918, line 20: if you remove the margin above you should also be able to remove that padding

image

web/src/components/CityContentFooter.tsx
```diff
@@ -0,0 +1,77 @@
...
+  })
+  const licensesPath = pathnameFromRouteInformation({
+    route: LICENSES_ROUTE,
+  })
+
+  if (inSidebar) {
+    return (
+      <Footer>
+        <SidebarFooterContainer>
+          <CleanLink to={disclaimerPath}>{t('imprintAndContact')}</CleanLink>
```
f1sh1918, line 58: just assign these clean links to a const like
```js
const links = <>
<CleanLinks>....
</>
```
to reduce redundant code

web/src/components/CityContentFooter.tsx
```diff
@@ -0,0 +1,77 @@
...
+import buildConfig from '../constants/buildConfig'
+import CleanLink from './CleanLink'
+import Footer from './Footer'
+
+const SidebarFooterContainer = styled.div`
+  width: 100%;
+  margin: -10px 12px 0 23px;
+  > * {
+    color: ${props => props.theme.colors.textColor};
+    height: 48px;
```
f1sh1918, line 16:
```suggestion
    padding: 16px 0;
```
setting fixed heights reduces the flexibility of a layout. I prefer using a padding here

web/src/components/KebabMenu.tsx
```diff
@@ -83,7 +82,7 @@ const ToggleButton = styled.button`
   margin-top: 6px;
 `
 
-const KebabMenu = ({ items, direction, show, setShow, language }: KebabMenuProps): ReactElement | null => {
+const KebabMenu = ({ items, direction, show, setShow, footer }: KebabMenuProps): ReactElement | null => {
```
steffenkleinle, line 85: We don't really have a convention for this yet but I'd follow the react one to use uppercase names for react nodes/elements.
```suggestion
const KebabMenu = ({ items, direction, show, setShow, Footer }: KebabMenuProps): ReactElement | null => {
```

web/src/components/Header.tsx
```diff
@@ -19,8 +20,8 @@ type PropsType = {
   viewportSmall: boolean
   cityName?: string
   direction: UiDirectionType
-  showSidebar?: boolean
-  setShowSidebar?: (show: boolean) => void
+  isSidebarOpen?: boolean
```
steffenkleinle, line 23: :+1:

web/src/components/CityContentFooter.tsx
```diff
@@ -1,19 +1,43 @@
...
+
+type CityContentFooterProps = {
   city: string
   language: string
   overlay?: boolean
+  inSidebar?: boolean
 }
 
-const LocationFooter: React.FC<PropsType> = ({ city, language, overlay = false }: PropsType): ReactElement => {
+const CityContentFooter: React.FC<CityContentFooterProps> = ({
```
steffenkleinle, line 35: Optional, unneeded type:
```suggestion
const CityContentFooter = ({
```

web/src/components/CityContentFooter.tsx
```diff
@@ -1,19 +1,43 @@
...
+  city,
+  language,
+  overlay = false,
+  inSidebar = false,
```
steffenkleinle, line 39: Optional:
What happens if overlay and inSidebar are set to true at the same time? `overlay` is just ignored. Therefore perhaps one property `mode: 'normal' | 'overlay' | 'sidebar' = 'normal'` would perhaps be better here.

web/src/components/CityContentFooter.tsx
```diff
@@ -27,14 +51,24 @@ const LocationFooter: React.FC<PropsType> = ({ city, language, overlay = false }
     route: LICENSES_ROUTE,
   })
 
-  return (
-    <Footer overlay={overlay}>
+  const links = (
```
steffenkleinle, line 54:
```suggestion
  const Links = (
```

web/src/components/CityContentLayout.tsx
```diff
@@ -63,15 +64,21 @@ const LocationLayout = (props: PropsType): ReactElement => {
...
           viewportSmall={viewportSmall}
           languageCode={languageCode}
           route={route}
+          isSidebarOpen={isSidebarOpen}
+          setIsSidebarOpen={setIsSidebarOpen}
         />
       }
-      footer={!isLoading && showFooter ? <LocationFooter city={cityModel.code} language={languageCode} /> : null}
+      footer={
+        !isLoading && showFooter && !isSidebarOpen ? (
```
steffenkleinle, line 78: I think this should not be dependend on whether the sidebar is currently opened or closed. Imo there should not be a footer shown here at all if we are on a small viewport and have the sidebar.

LeandraH: Hmmm, I feel like the imprint and privacy should always be easily seen and available, not just after clicking on an icon that you first have to scroll up to find, that's not very intuitive to me.

steffenkleinle: Did not really think about that but according to the issue it should be moved and therefore not displayed here for city content. Therefore I'd propose to check back with the rest of the team, especially UI/UX to discuss what to do. I am personally fine with both. But I think there was a reason to move it to the sidebar in the first place.

steffenkleinle (approved):

steffenkleinle (approved):
Not tested again

web/src/components/CityContentLayout.tsx
LeandraH: Sorry, forgot to comment, yes, the footer is now not shown on the main screen on small screens, it's only in the sidebar there.

f1sh1918 (approved):
Tested on chrome. Looks fine now :)

PR #788, opened 2022-10-12
IGAPP-1107: Sort events if start dates equal

steffenkleinle:

Happens for me for example for all occurrences of `Frauengruppe` and `Frauen lernen Bildungs- und Erziehungseinrichtungen kennen` in `/muenchen/de/events` using the test cms (native: Frauengruppe before Frauen lernen..., web: the other way round). I think even if this is not reproducible on all platforms (e.g. @f1sh1918 failed to reproduce it) it makes sense to have a deterministic sorting for the case that start dates are equal.


SteffiStuffel (approved):

LeandraH (approved):
Tested in Firefox, looks good

f1sh1918 (approved):

PR #789, opened 2022-10-13
IGAPP-1135: Wrong height map route 

f1sh1918:

Go to map route, page should be fullscreen without scrollbar Y


steffenkleinle (approved):
Not tested

sarahsporck (approved):
Tested in chrome. Nice find!

PR #790, opened 2022-10-14
IGAPP-1117: Readded FocusTrap to feedback

Cmd190:
Added the FocusTrap to the feedback dialog without locking the focus after closing the dialog.

Note: As the options already states, it the focus lock is removed through a mouse click. Therefore the user can not navigate wiith tab immeadiately after closing the feedback dialog. Is this a problem? 


sarahsporck (approved):
Tested on chrome and firefox.

f1sh1918 (approved):
Tested on chrome. nice find!

Cmd190:
I don't know why the iOS e2e tests failed. Can someone just rerun them please?

PR #792, opened 2022-10-17
IGAPP-1089: Add site_name meta tag

sarahsporck:

For reference: https://ogp.me/#optional

Did also not find out where "von Integreat-App" is comming from, but I guess that's generated automatically by google. In duckduckgo or ecosie this text isn't displayed.
For site_name reference - you can check out most pages in the web and see how the site name influences what is displayed in the search. Generally, the `site_name` is just appended to the pages.
So you would see `Regionen - integreat.app` instead of only `Regionen`


steffenkleinle (approved):
Aren't og tags only for facebook?

sarahsporck:
Did need a few days to find a few more infos on this: https://developers.google.com/search/docs/appearance/site-names

PR #795, opened 2022-10-19
IGAPP-636: Refactor events

steffenkleinle:
### Idea

This PR is a first attempt to remove redux, sagas, routeMapping and withPayloadProvider from our native app and instead just use hooks to directly access the data container. For now this is only a POC for the events route but looks very promising for now. Let me know what you think, feedback is very much appreciated!

### Plan:
* Replace redux store and route mappings with
  * direct access to the data container for cms data
  * react context for stuff like selected city code and selected language code
  * navigation params for stuff like route city code, route language code and route id (e.g. event slug, news id, ...)
* Replace sagas with hooks loading the data directly in the route containers
* Replace withPayloadProvider with component to handle language unavailable/errors/loading/refreshing/...
* Use slug instead of cityContentPath for identification of events/pois (stays for categories)
* Use another branch then main for working
* Remove peeking (opening content in other city/language then currently selected) and instead just open web app in inappbrowser

### Pros:
* Removal of old legacy code (sagas, redux, withPayloadProvider, routeMapping)
* Reduction of code overall
* Increase of flexibility, e.g. some issues that were not solvable in the past should be possible afterwards (e.g. https://issues.tuerantuer.org/browse/IGAPP-919, https://issues.tuerantuer.org/browse/IGAPP-899, https://issues.tuerantuer.org/browse/IGAPP-497, https://issues.tuerantuer.org/browse/IGAPP-122)
* Possibility to move downloading of resources (images, pdfs, ...) to the background
* Straightforward and understandable responsibilities and code

### Cons:
* Work
* Things may break
* You tell me :)

### Whats already working:
* Events in general
* Offline functionality
* Refreshing
* Share url
* Language change
*  Selection of specific events
* Use of resource cache

### Not working yet:
* Force refresh
* Language selection for unavailable languages


steffenkleinle:
@f1sh1918 @SteffiStuffel @sarahsporck let me know what you think. Please be honest and mention things you like/don't like/think we may run into problems. If you like this and think it may work, let me know then I'll continue working on it :)

f1sh1918:
> @f1sh1918 @SteffiStuffel @sarahsporck let me know what you think. Please be honest and mention things you like/don't like/think we may run into problems. If you like this and think it may work, let me know then I'll continue working on it :)

Well everything that makes redux saga obsolete is welcoming to me. I've had kind of problems to understand whats happening behind the scenes to be honest. Your approach now looks clean and readable. I also recognized some speed improvements.

SteffiStuffel:
Really good idea to remove the sagas and the complicated routing stuff and thanks for advancing this.

Some thoughts on your suggestions:

> react context for stuff like selected city code and selected language code

I think react context does not persist after closing the app.


Replacing sagas with hook sounds really good


> Replace withPayloadProvider with component to handle language unavailable/errors/loading/refreshing/...

Not sure about this. I think this is a really good usecase for a hoc. We use a component for this, as you suggested, in Lunes, and I think it is easier to just surround a lot of components with a hoc, than adding an additional component to every jsx. But using a component will make it easier to show parts of the content, e.g. a heading that is already known and so on. But I think sticking to a HOC here makes a lot of sense. 


> Use another branch then main for working

What do you mean? Introducting a develop branch or just a normal feature branch as we always do?


> Removal of old legacy code (sagas, redux, withPayloadProvider, routeMapping)

Sounds sooo good :) 


So in total, i will approve everything, which will remove the sagas and improve the whole navigation topic :D

steffenkleinle:
> > react context for stuff like selected city code and selected language code
> 
> I think react context does not persist after closing the app.

Yes, did not really make clear what I wanted to do here I guess. Just meant react context instead of redux state, it will still be persisted in the async storage.

> > Replace withPayloadProvider with component to handle language unavailable/errors/loading/refreshing/...
> 
> Not sure about this. I think this is a really good usecase for a hoc. We use a component for this, as you suggested, in Lunes, and I think it is easier to just surround a lot of components with a hoc, than adding an additional component to every jsx. But using a component will make it easier to show parts of the content, e.g. a heading that is already known and so on. But I think sticking to a HOC here makes a lot of sense.

Thought about that as well but I think we should use a component instead of a hook. If I look at the way the hoc currently looks like it is very complicated (the way the props are structured, the typings, ...) while it is really easy to wrap a component in another component:
``` tsx
    <LoadingErrorHandler {...response}>
      {data && <Events ... />}
    </LoadingErrorHandler>
```

Pros of component:
* No complicated typings
* No handling of nested component props necessary
* No outdated/not really common syntax and pattern (hocs)
* Easy to use

Pros of hoc:
* No check `data && ` necessary

What else do you think is easier for hocs?
 
> > Use another branch then main for working
> 
> What do you mean? Introducting a develop branch or just a normal feature branch as we always do?

Yes, kind of. A new branch alongside main that we do all the refactoring in. This will take some more PRs and work and will quite sure break some stuff (which we may only be able to fix in the end). Therefore I'd rather split this up in multiple PRs then to do everything in one big PR. We just can rebase it regularly on main and not work on navigation topics till then and we should be fine.

native/src/routes/EventsContainer.tsx
```diff
@@ -1,232 +1,107 @@
...
+        />
+      ),
+    })
+  }, [route, navigation, cityCode, languageCode, data, currentEvent])
+
+  const onLanguageChange = useCallback(
+    (newLanguage: string) => {
+      if (currentEvent) {
+        const newSlug = currentEvent.availableLanguageSlugs[newLanguage]
+        // TODO IGAPP-636: Handle language not available?
```
steffenkleinle, line 70: I think it may already also be fine to just show the event list again if there is no translation in the new language?

sarahsporck:
Logic is definitely minimalized. Like that one. Could not really check for performance improvements in native as jcenter is down currently.

The hooks definitely feel more efficient. 
Personally I dislike that most of the LanguageUnavailable logic cannot reused. But as of a first draft I think that's fine. 

> Use slug instead of cityContentPath for identification of events/pois (stays for categories)

This one will be very nice for the new feedback endpoints in Django :)

I feel you're definitely on the right track here.

native/src/routes/EventsContainer.tsx
steffenkleinle: @sarahsporck @f1sh1918 @SteffiStuffel opinions on this? What should we do if a site that is opened in the background is not available in the new language? Just going back to the list of events/home page for categories is good enough I think. Other opinions?

steffenkleinle:
> The hooks definitely feel more efficient. Personally I dislike that most of the LanguageUnavailable logic cannot reused. But as of a first draft I think that's fine.

@sarahsporck can you elaborate what language unavailable logic you mean? I think in this branch there is none yet at all, right? Feel free to check #800 to see how I did it there. Suggestions/recommendations are of course appreciated.

PR #796, opened 2022-10-20
IGAPP-1128: Replaced deprecated keyCode event property

Cmd190:
KeyboardEvent.keyCode was replaced.


LeandraH (approved):
Nice!

web/src/hooks/useOnClickOutside.ts
```diff
@@ -12,11 +12,11 @@ const useOnClickOutside = (ref: { current: HTMLElement | null }, callback: () =>
     }
 
     const handleKeyPress = (event: KeyboardEvent) => {
-      const KEY_ENTER = 13
-      const KEY_ESCAPE = 27
-      if (event.keyCode === KEY_ESCAPE) {
+      const KEY_ENTER = 'Enter'
```
SteffiStuffel, line 15: I think now these constants are not needed any longer, as the naming is already clear.

steffenkleinle (commented):
Tested on firefox. Minor since it probably only affects a rare minority of users like me: I switched Escape and Caps Lock on my keyboard but I still have to use the same keys here, i.e. `Escape` to close instead of `Caps Lock`. Before it worked fine for me to close with `Caps Lock`. Maybe you could investigate if there is a way to fix this?

web/src/hooks/useOnClickOutside.ts
Cmd190: done

Cmd190:
> Tested on firefox. Minor since it probably only affects a rare minority of users like me: I switched Escape and Caps Lock on my keyboard but I still have to use the same keys here, i.e. `Escape` to close instead of `Caps Lock`. Before it worked fine for me to close with `Caps Lock`. Maybe you could investigate if there is a way to fix this?

Yes this was because it checked for the code send from the keyboard and not the actual value. I changed to the `key` event which should do the jobt

steffenkleinle (approved):
Good stuff!

PR #798, opened 2022-10-25
IGAPP-1073: Yarn2 update

SteffiStuffel:


sarahsporck:
I've also shortly looked into this issue and realised that yarn2 currently does not yet fully support our monorepo setup. For example I was not able to call the tests from a subproject, even though jest was installed in the root project. Did you also encounter this issue? I did have an issue reffering to this problem, but can't find it anymore. But if you do encounter the issue maybe https://yarnpkg.com/getting-started/qa#how-to-share-scripts-between-workspaces can help.

SteffiStuffel:
> I've also shortly looked into this issue and realised that yarn2 currently does not yet fully support our monorepo setup. For example I was not able to call the tests from a subproject, even though jest was installed in the root project. Did you also encounter this issue? I did have an issue reffering to this problem, but can't find it anymore. But if you do encounter the issue maybe https://yarnpkg.com/getting-started/qa#how-to-share-scripts-between-workspaces can help.

Thank you for looking into this.
You can call a script container a `:` from everywhere, also from the subprojects. Script sharing is not a problem I see here. Locally it works fine for me. Something with the caching is not working. [Here](https://github.com/yarnpkg/berry/discussions/2621) it is written that you can either cache only `./yarn`, but than have to call `yarn install` on every job OR cache `./yarn` and `node_modules` and only need to call `yarn install` once. But on Lunes project we only cache `./yarn` and do not call `yarn install` on every job. Locally everything works fine for me.

PR #800, opened 2022-10-27
IGAPP-636: Refactor categories

steffenkleinle:
Done:
* Remove dashboard and include it in categories (routes + containers)
* Refactor categories container to use hooks
* Handle language not available
* Remove use of categories state view


native/src/routes/LandingContainer.tsx
```diff
@@ -67,20 +67,11 @@ const mapStateToProps = (state: StateType, ownProps: OwnProps): StateProps => {
...
-        languageCode,
-      }),
-      forceRefresh: false,
-      resetNavigation: true,
-    })
+  const { changeCityCode } = useContext(AppContext)
+
+  const navigateToDashboard = (cityCode: string) => {
+    changeCityCode(cityCode)
+    navigation.reset({ index: 0, routes: [{ name: CATEGORIES_ROUTE, params: {} }] })
```
sarahsporck, line 74: In general, does this also work if you deep link into the app from cold start?

sarahsporck:
I get an error when starting the app:

native/src/routes/LandingContainer.tsx
steffenkleinle: Probably not yet but its on my list and there is a TODO in `navigateToDeepLink`

steffenkleinle:
> I get an error when starting the app: 

Can not reproduce :/ Which screen is that? Just the home screen of a city? Do you have the latest version of the branch?

sarahsporck:
> Can not reproduce :/ Which screen is that? Just the home screen of a city? Do you have the latest version of the branch?

works now :D

api-client/src/utils/index.ts
```diff
@@ -1,7 +1 @@
-export const getSlugFromPath = (path: string): string => {
-  const slug = path.split('/').pop()
-  if (!slug) {
-    throw new Error('Empty string is not a valid path')
-  }
-  return slug
-}
+export const getSlugFromPath = (path: string): string => path.split('/').pop()!
```
sarahsporck, line 1: why do we always have a slug?

native/src/components/LanguageNotAvailablePage.tsx
```diff
@@ -5,37 +5,54 @@ import styled from 'styled-components/native'
...
     ({ code, name }) =>
       new SelectorItemModel({
         code,
         name,
         enabled: true,
-        onPress: () => changeLanguage(code),
+        onPress: () => {
+          changeLanguageCode(code)
+          if (changeLanguage) {
+            changeLanguage(code)
```
SteffiStuffel, line 40: Why do we need both: the changeLanguageCode-function and the changeLanguage-function?

sarahsporck (changes requested):
Nice work! The Categories are definitely much cleaner now.
There are a few things that need to addressed though:
- While navigating the categories the Header always disappers (and reloads?). This doesn't really feel intuitive
- Search does not work
- When changing the language the error screen is shortly visible

native/src/hooks/useUpdateParamsOnLanguageChange.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react'
+
+type UseOnLanguageChangeProps = {
+  languageCode: string
+  onLanguageChange: (newLanguageCode: string, previousLanguageCode: string) => void
+}
+
+const useOnLanguageChange = ({ languageCode, onLanguageChange }: UseOnLanguageChangeProps): void => {
```
SteffiStuffel, line 8: Should this really be named useOnLangugeChange?

What is the difference between this "useUpdateParamsOnLanguageChange" and "useOnLanguageChange"?
Which Params are updated? When do I use which hook?

SteffiStuffel (approved):
Not tested, looks good.

native/src/components/LanguageNotAvailablePage.tsx
steffenkleinle: One is legacy for routes that still use the redux store and will be removed later once the redux store gets kicked out completely.

api-client/src/utils/index.ts
steffenkleinle: Split always returns an array of length >= 1, even for an empty string this results in an empty string, therefore it will never be undefined.

native/src/routes/LandingContainer.tsx
steffenkleinle: Added a todo here: https://github.com/digitalfabrik/integreat-app/pull/802

steffenkleinle:
>     * While navigating the categories the Header always disappers (and reloads?). This doesn't really feel intuitive

Added a TODO (https://github.com/digitalfabrik/integreat-app/pull/802) and would take a look at it separately if that is okay for you :)

>     * Search does not work

Should be fixed now :)

>     * When changing the language the error screen is shortly visible

Should be fixed now :)

sarahsporck (approved):
tested in web

PR #804, opened 2022-11-03
IGAPP-1136: Fix overlapping POIs

LeandraH:

POIs that are very close to each other or at the same address currently only display the first one (I believe in alphabetical order), so other POIs disappear. Here, overlapping ones have the names displayed in different locations. It's still not perfect, e.g. you need to zoom in really far for them to not be clustered anymore (but with a maximum cluster zoom, it displays only one POI again when the cluster disappears).

image

There is a [design ticket](https://issues.tuerantuer.org/browse/IGAPP-1146) for coming up with a better solution but for now, this at least doesn't hide POIs anymore.


steffenkleinle (approved):
LGTM, but I think we should wait for a review from @f1sh1918.

release-notes/unreleased/IGAPP-1136-fix-overlapping-pois-hiding-each-other.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: IGAPP-1136
+show_in_stores: false
+platforms:
+  - web
+  - android
+  - ios
+en: Fix overlapping POIs hiding each other
```
f1sh1918, line 7: pls add german translation for play store

steffenkleinle: Only necessary if `show_in_stores` is set to true. But I think it should be set to true so I agree with @f1sh1918 .

f1sh1918 (changes requested):
Nice work, i just found two small bugs

1. When you zoom close enough to see the pois with the same address and the select one, it zooms out and shows them as a cluster. (tested on iOS)
Either we could adjust the `detailZoom` (to 16) or somehow check these Pois and give them a special flag and use a different zoom for them.
2. The marker labels overlaps the clusteringCountLayer. check my comments to fix this

release-notes/unreleased/IGAPP-1136-fix-overlapping-pois-hiding-each-other.yml
f1sh1918: oh sorry didn't see that.

api-client/src/testing/PoiModelBuilder.ts
```diff
@@ -25,8 +25,8 @@ const pois = [
       address: 'Test Address 1',
       town: 'Test Town',
       postcode: '12345',
-      latitude: 29.979848,
       longitude: 31.133859,
+      latitude: 29.979848,
       name: 'Test Title',
```
f1sh1918, line 30: pls reverted if change not needed

api-client/src/utils/__tests__/geoJson.spec.ts
```diff
@@ -44,31 +45,67 @@ describe('geoJson', () => {
...
   })
 
   describe('prepareFeatureLocations', () => {
     it('should sort feature location by distance ascending', () => {
-      const poiFeature1 = prepareFeatureLocation(poi1, userLocation)
-      const poiFeature2 = prepareFeatureLocation(poi2, userLocation)
-      const poiFeature3 = prepareFeatureLocation(poi3, userLocation)
+      const poiFeature1 = prepareFeatureLocation(poi1, userLocation, [])
+      const poiFeature2 = prepareFeatureLocation(poi2, userLocation, [])
+      console.log(poiFeature2)
```
f1sh1918, line 99: rm clog

api-client/src/utils/geoJson.ts
```diff
@@ -8,8 +9,26 @@ export const embedInCollection = (features: PoiFeature[]): PoiFeatureCollection
...
+  userLocation: LocationType | null,
+  coordinateList: Position[]
+): PoiFeature => {
   const { featureLocation } = poi
+  const { coordinates } = featureLocation.geometry
+
+  const maxDistanceForOverlap = 0.0003
+  const longitude = coordinates[0] as number
+  const latitude = coordinates[1] as number
+  const overlappingCoordinates = coordinateList.filter(
```
f1sh1918, line 23: its easier and cleaner to use turfs distance function and then set maybe
```js
 // 5 meters
  const maxDistanceForOverlap = 0.05
```
```js
const overlappingCoordinates = coordinateList.filter(coord => distance(coord, coordinates) < maxDistanceForOverlap)
```

f1sh1918 (commented):
Added some comments to simplify the calculation

api-client/src/maps/index.ts
```diff
@@ -29,6 +29,7 @@ export type GeoJsonPoiProperties = {
   address?: string
   distance?: string
   thumbnail?: string
+  closeToOtherPoi?: boolean
```
f1sh1918, line 32: maybe i overseen it but is there any reason to make it optional? I think its `false` by default

api-client/src/testing/PoiModelBuilder.ts
LeandraH: I think it makes it easier to understand to have them in this order because that's the same order as in the coordinates.

api-client/src/maps/index.ts
LeandraH: Good point

f1sh1918 (approved):
Tested on chrome & ios. Nice work :)

PR #811, opened 2022-11-09
IGAPP-636: Refactor landing

steffenkleinle:


sarahsporck (commented):
Good work!

The EasterEggImage throws an error: `[Error: Failed to delete '/data/user/0/app.integreat.test/files']`

steffenkleinle:
> Good work!
> 
> The EasterEggImage throws an error: `[Error: Failed to delete '/data/user/0/app.integreat.test/files']`

Should be fixed now, could you please test again? Not sure what caused it in the first place though. Looks not that nice when switching the cms url since the list takes a while to load but since this is only a dev tool I think that should be fine.

LeandraH (approved):
Looks good, tested on iOS simulator

steffenkleinle:
@sarahsporck ping

PR #813, opened 2022-11-10
IGAPP-1157: Fix handle bar poi detail

f1sh1918:

Happened due to refactoring
https://github.com/digitalfabrik/integreat-app/pull/793

Pls test web mobile that the handleBar is shown when a poi was selected.
No release notes due to fixing before it occurs in next version :)

http://localhost:9000/testumgebung/de/locations/cafe-t%C3%BCr-an-t%C3%BCr

### DIFF
--- a/web/src/components/BottomActionSheet.tsx
+++ b/web/src/components/BottomActionSheet.tsx
@@ -56,7 +56,7 @@ const BottomActionSheet = React.forwardRef(
         open
         scrollLocking={false}
         blocking={false}
-        header={!!title && <Title>{title}</Title>}
+        header={!title ? <Title>{title}</Title> : null}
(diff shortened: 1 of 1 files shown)


web/src/components/BottomActionSheet.tsx
```diff
@@ -56,7 +56,7 @@ const BottomActionSheet = React.forwardRef(
         open
         scrollLocking={false}
         blocking={false}
-        header={!!title && <Title>{title}</Title>}
+        header={!title ? <Title>{title}</Title> : null}
```
steffenkleinle, line 59: I think this should be
```suggestion
        header={title ? <Title>{title}</Title> : null}
```

steffenkleinle (approved):

PR #816, opened 2022-11-15
IGAPP-1145: Locally deploy rasa widget

SteffiStuffel:


web/src/components/ChatBotWidget.tsx
```diff
@@ -18,7 +19,7 @@ const ChatBotWidget = (): ReactElement => {
   useEffect(() => {
     const script = document.createElement('script')
 
-    script.src = 'https://unpkg.com/@rasahq/rasa-chat'
+    script.src = `/rasa-widget/chatbot.js`
```
SteffiStuffel, line 22: Widget is released unter Apache 2.0 licence, so copying it to our MIT licenced project should be fine.

web/www/rasa-widget/chatbot.js
```diff
@@ -0,0 +1,2 @@
+/*! For license information please see widget.js.LICENSE.txt */
```
SteffiStuffel, line 1: It is just the same file as from unpkg, but changed the font, as it was a google font.

f1sh1918 (commented):
Implementation looks fine, but i get an invalid certificate error

image

LeandraH (approved):
Looks good, tested in Firefox.

web/src/components/ChatBotWidget.tsx
```diff
@@ -18,7 +18,7 @@ const ChatBotWidget = (): ReactElement => {
...
```
steffenkleinle, line 21:
```suggestion
    script.src = '/rasa-widget/chatbot.js'
```

web/www/rasa-widget/chatbot.js
steffenkleinle: Does anyone whats a good practice to keep scripts like this updated? Should we just have a reminder in like 6 months to check if something changed?

steffenkleinle (approved):
Tested on firefox, looks good

web/www/rasa-widget/chatbot.js
LeandraH: Good question, I don't know one and a quick Google doesn't turn up an obvious answer either. A reminder every six months to check all copied scripts (maybe with a growing check list of files to check?) seems like a decent solution for now.

LeandraH:
We should check if the errors @f1sh1918 got are fixed after the Salt PR is merged.

SteffiStuffel:
> Implementation looks fine, but i get an invalid certificate error
> 

Cannot reproduce. I do not get this error. Also this seems to be an error of the certificate twb created, so we cannot change this anyways. But it seems they have fixed this already.

PR #822, opened 2022-12-01
IGAPP-636: Fix news

steffenkleinle:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **IGAPP**.

PR #802, opened 2022-11-01
IGAPP-636: Refactor native app to hooks

steffenkleinle:
### Idea

This PR is a first attempt to remove redux, sagas, routeMapping and withPayloadProvider from our native app and instead just use hooks to directly access the data container. For now this is only a POC for the events route but looks very promising for now. Let me know what you think, feedback is very much appreciated!

### Plan:
* Replace redux store and route mappings with
  * direct access to the data container for cms data
  * react context for stuff like selected city code and selected language code
  * navigation params for stuff like route city code, route language code and route id (e.g. event slug, news id, ...)
* Replace sagas with hooks loading the data directly in the route containers
* Replace withPayloadProvider with component to handle language unavailable/errors/loading/refreshing/...
* Use slug instead of cityContentPath for identification of events/pois (stays for categories)
* Use another branch then main for working
* Remove peeking (opening content in other city/language then currently selected) and instead just open web app in inappbrowser

### Pros:
* Removal of old legacy code (sagas, redux, withPayloadProvider, routeMapping)
* Reduction of code overall
* Increase of flexibility, e.g. some issues that were not solvable in the past should be possible afterwards (e.g. https://issues.tuerantuer.org/browse/IGAPP-919, https://issues.tuerantuer.org/browse/IGAPP-899, https://issues.tuerantuer.org/browse/IGAPP-497, https://issues.tuerantuer.org/browse/IGAPP-122)
* Possibility to move downloading of resources (images, pdfs, ...) to the background
* Straightforward and understandable responsibilities and code

### Cons:
* Work
* Things may break
* You tell me :)

### TODOs:
- [x] Refactor events #795 
- [x] Refactor categories #800 
- [x] Refactor landing #811 
- [x] Refactor pois and navigation #814 
- [x] Fix news #822 
- [x] Fix flickering header #823 
- [x] Deep linking #830 
- [x] Snackbar #826 
- [x] Fetch resource cache #844 
- [x] Last usages of city content #834
- [x] Push news subscription #834
- [x] Force refresh/last updates/refreshing failed handling #833
- [x] Remove sagas/redux/useCities/useSetShareUrl/peeking #845 
- [x] Fix navigation bar (See https://github.com/digitalfabrik/integreat-app/pull/823#pullrequestreview-1208079328): #850 
- [x] Tests #847

PR #823, opened 2022-12-04
IGAPP-636: Fix and refactor header

steffenkleinle:


steffenkleinle:
This should be the last big PR with refactoring on the different routes. All routes that load data async from the cms (except news, did that separately in #822) should now use the new LoadingErrorHandler component to handle loading and errors and use hooks to
* load data
* set the header (including available languages and share url) and
* handle language change

Other things that happend:
* Refactored header and fixed flickering (mentioned by @sarahsporck in #https://github.com/digitalfabrik/integreat-app/pull/800#pullrequestreview-1178996986)
* Added hook `useOnBackNavigation` to overwrite default back navigation behavior
* ErrorCodes are now mostly used in the native app (Failure only used that before already)

f1sh1918 (changes requested):
Really awesome refactoring and nice backNavigation hook.

I guess news route gonna be fixed in separate pr?
I found one issue. When you change your location and try to use the forward arrow in the navigation header, its not working. If you once scroll everything is fine and you can use the arrows

https://user-images.githubusercontent.com/37902063/206142166-f1750767-b30b-4214-abc7-ef7e68503555.mp4

native/src/hooks/useLoadDisclaimer.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { determineApiUrl } from '../utils/helpers'
+import useLoadCityContent, { CityContentReturn } from './useLoadCityContent'
+
+type UseLoadDisclaimerProps = {
+  cityCode: string
+  languageCode: string
+}
+
+const useLoadDisclaimer = ({
```
SteffiStuffel, line 13: We have a lot of duplication in this loading hooks, e.g. useLoadDisclaimers and useLoadOffers are almost the same. I think these hook can nicely be refactored with using template method pattern.

native/src/hooks/useOnBackNavigation.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const useOnBackNavigation = (onBackNavigation?: (action: NavigationAction) => void): void => {
+  const navigation = useNavigation()
+
+  useEffect(
+    () =>
+      navigation.addListener('beforeRemove', event => {
+        if (onBackNavigation) {
+          event.preventDefault()
+          onBackNavigation(event.data.action)
```
SteffiStuffel, line 12: Can you please explain what is happening here? I don't get it.

native/src/routes/LoadingErrorHandler.tsx
```diff
@@ -13,9 +14,10 @@ const LOADING_TIMEOUT = 800
 
 type LoadingErrorHandlerProps = {
   children?: ReactNode
-  error: Error | null
+  error: Error | ErrorCode | null
```
SteffiStuffel, line 17: Hmh, does this make sense? Why do we even have this ErrorCode class instead of creating own custom error class?

native/src/routes/OffersContainer.tsx
```diff
@@ -37,83 +23,51 @@ type OffersContainerProps = {
...
+  useHeader({ navigation, route, availableLanguages, data, shareUrl })
 
-  const request = useCallback(async () => {
-    const apiUrl = await determineApiUrl()
-    return createOffersEndpoint(apiUrl).request({
-      city: cityCode,
+  const navigateToOffer = (tile: TileModel) => {
+    const { title, path, isExternalUrl, postData } = tile
+    if (isExternalUrl && postData) {
+      // HTTP POST is neither supported by the InAppBrowser nor by Linking, therefore we have to open it in a webview
```
SteffiStuffel, line 36: By moving `isExternalUrl && postData` condition to a own variable with a meaningful name, we can get rid of this comment.

native/src/routes/OffersContainer.tsx
```diff
@@ -37,83 +23,51 @@ type OffersContainerProps = {
...
+      navigation.push(EXTERNAL_OFFER_ROUTE, {
+        url: path,
+        shareUrl: path,
+        postData,
+      })
+    } else if (isExternalUrl) {
+      openExternalUrl(path).catch(error => showSnackbar(error.message))
+    } else if (data?.offers.find(offer => offer.title === title)?.alias === SPRUNGBRETT_OFFER_ROUTE) {
```
SteffiStuffel, line 44: What happens if none of these three if-cases are true? Should an else case which reports an error be added?

native/src/routes/OffersContainer.tsx
```diff
@@ -37,83 +23,51 @@ type OffersContainerProps = {
...
+      navigateTo({ route: SPRUNGBRETT_OFFER_ROUTE, cityCode, languageCode })
+    }
+  }
+  const navigateToFeedback = (isPositiveFeedback: boolean) => {
```
SteffiStuffel, line 48: Newline before function please.

native/src/routes/OffersContainer.tsx
```diff
@@ -37,83 +23,51 @@ type OffersContainerProps = {
...
-        language: languageCode,
-      })
-    return (
-      <LayoutedScrollView refreshControl={<RefreshControl onRefresh={refresh} refreshing={loading} />}>
-        <Failure code={fromError(error)} tryAgain={refresh} />
-      </LayoutedScrollView>
-    )
   }
 
+  const error = data?.city && !data.city.offersEnabled ? ErrorCode.PageNotFound : response.error
```
SteffiStuffel, line 57: `const error...` should be moved above the function.

native/src/routes/SprungbrettOfferContainer.tsx
```diff
@@ -31,48 +17,14 @@ type SprungbrettOfferContainerProps = {
 }
 
 const SprungbrettOfferContainer = ({ route, navigation }: SprungbrettOfferContainerProps): ReactElement => {
-  const cities = useCities()
-  const [title, setTitle] = useState<string>('')
   const { cityCode, languageCode } = useCityAppContext()
   const alias = SPRUNGBRETT_OFFER_ROUTE
```
SteffiStuffel, line 21: (optinal) as this was not part of your PR, but saving this well named constant to a local variable does not really make sense.

native/src/routes/SprungbrettOfferContainer.tsx
```diff
@@ -84,40 +36,20 @@ const SprungbrettOfferContainer = ({ route, navigation }: SprungbrettOfferContai
...
-      <LayoutedScrollView refreshControl={refreshControl}>
-        <Failure code={fromError(error)} tryAgain={refresh} />
-      </LayoutedScrollView>
-    )
-  }
-
-  if (!jobs) {
-    return <LayoutedScrollView refreshControl={refreshControl} />
-  }
+  const error = data?.city && !data.city.offersEnabled ? ErrorCode.PageNotFound : response.error
```
SteffiStuffel, line 39: Error should be moved above the function.

native/src/hooks/useLoadDisclaimer.ts
steffenkleinle: Okay, will try. What is template method pattern, do you have an example?

native/src/routes/LoadingErrorHandler.tsx
steffenkleinle: In my opinion it does but not 100% sure. Imo handling of just an error code enum is so much more easier then creating a new error and initializing it for every error. I just figured since we don't ever use the additional information our error classes provide but instead we always just did `fromError(error)` which returns an error code we might as well use just the error code directly.

native/src/routes/OffersContainer.tsx
steffenkleinle: Do you have a good suggestion for a name?

steffenkleinle:
> I guess news route gonna be fixed in separate pr?

> except news, did that separately in https://github.com/digitalfabrik/integreat-app/pull/822

native/src/hooks/useLoadDisclaimer.ts
SteffiStuffel: See: https://en.wikipedia.org/wiki/Template_method_pattern

native/src/routes/OffersContainer.tsx
SteffiStuffel: Maybe canOnlyBeHandledByWebview

native/src/hooks/useLoadCityContent.ts
```diff
@@ -76,7 +75,14 @@ const useLoadCityContent = <T>({ cityCode, languageCode, load }: UseLoadCityCont
         }
       : null
 
-  return { ...getError(), loading, refresh, data }
+  return { error: getError(), loading, refresh, data }
 }
 
+// Simple utility helper to just load cities and languages
+export const useSimpleLoadCityContent = (params: {
```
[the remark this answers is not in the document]
steffenkleinle: Will be in #822 since tunews is not using normal fetch logic but endless scroll/pagination fetch logic and therefore we still need cities and languages but not a `load` function. However, I might rewrite this such that we only need one export.

steffenkleinle:
> I guess news route gonna be fixed in separate pr? I found one issue. When you change your location and try to use the forward arrow in the navigation header, its not working. If you once scroll everything is fine and you can use the arrows

Added a todo in #802.

native/src/routes/LoadingErrorHandler.tsx
steffenkleinle: @SteffiStuffel can you just open a ticket yourself please? Looking forward to see your proposal :)

native/src/routes/OffersContainer.tsx
steffenkleinle: Hmm sorry but I feel like the comment still provides additional information since HTTP Post is the problem here. Will keep it like it was before.

native/src/routes/OffersContainer.tsx
steffenkleinle: Right now not possible since the only internal extra on native is the sprungbrett offer and `isExternalUrl` is set to `      isExternalUrl: offer.alias !== SPRUNGBRETT_OFFER_ROUTE,
`

native/src/hooks/useLoadDisclaimer.ts
steffenkleinle: done

PR #824, opened 2022-12-05
IGAPP-1089: Add Json-Ld for WebSite

sarahsporck:

### DIFF
--- a/web/src/RootSwitcher.tsx
+++ b/web/src/RootSwitcher.tsx
@@ -29,6 +29,7 @@ import { cityContentPattern, RoutePatterns } from './routes'
 import CityNotCooperatingPage from './routes/CityNotCooperatingPage'
 import JpalTrackingPage from './routes/JpalTrackingPage'
 import lazyWithRetry from './utils/retryImport'
+import JsonLdWebSite from './components/JsonLdWebSite'
 
 type RootSwitcherProps = {
   setContentLanguage: (languageCode: string) => void
@@ -86,6 +87,7 @@ const RootSwitcher = ({ setContentLanguage }: RootSwitcherProps): ReactElement =
 
   return (
     <Suspense fallback={<LoadingSpinner />}>
+      <JsonLdWebSite />
       <Routes>
         <Route path={RoutePatterns[LANDING_ROUTE]} element={<LandingPage {...props} />} />
         <Route path={RoutePatterns[MAIN_DISCLAIMER_ROUTE]} element={<MainDisclaimerPage {...props} />} />
--- a/web/src/components/JsonLdWebSite.tsx
+++ b/web/src/components/JsonLdWebSite.tsx
@@ -1,0 +1,25 @@
+import React, { ReactElement } from 'react'
+import { Helmet } from 'react-helmet'
+import { WebSite, WithContext } from 'schema-dts'
+
+import buildConfig from '../constants/buildConfig'
+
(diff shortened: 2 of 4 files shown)


web/src/RootSwitcher.tsx
```diff
@@ -86,6 +87,7 @@ const RootSwitcher = ({ setContentLanguage }: RootSwitcherProps): ReactElement =
 
   return (
     <Suspense fallback={<LoadingSpinner />}>
+      <JsonLdWebSite />
```
steffenkleinle, line 90: I would rather put it in the Helmet.tsx file, don't think RootSwitcher is the best place for that (at least I would not expect it here).

web/src/components/JsonLdWebSite.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import buildConfig from '../constants/buildConfig'
+
+const JsonLdBreadcrumbs = (): ReactElement => {
+  const config = buildConfig()
+
+  const jsonLd: WithContext<WebSite> = {
+    '@context': 'https://schema.org',
+    '@type': 'WebSite',
+    name: config.hostName,
```
steffenkleinle, line 13: Shouldn't the name rather be `config.appName`?

web/src/components/JsonLdWebSite.tsx
```diff
@@ -0,0 +1,25 @@
...
+    url: window.location.origin,
+    version: __VERSION_NAME__,
```
steffenkleinle, line 15: How can for example martin find the version? Just by going to the inspector and looking for it there? Or is there an easier way? Is that okay for him?

web/src/components/JsonLdWebSite.tsx
sarahsporck: See ticket description.

web/src/components/JsonLdWebSite.tsx
sarahsporck: The Inspecor solution is Martin-Approved ✅

web/src/components/JsonLdWebSite.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Helmet } from 'react-helmet'
+import { WebSite, WithContext } from 'schema-dts'
+
+import buildConfig from '../constants/buildConfig'
+
+const JsonLdBreadcrumbs = (): ReactElement => {
+  const config = buildConfig()
+
+  const jsonLd: WithContext<WebSite> = {
+    '@context': 'https://schema.org',
```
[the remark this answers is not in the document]
sarahsporck: checkout https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data

web/src/RootSwitcher.tsx
steffenkleinle: Okay, makes sense. I think `App` would be the better place then. Please also add a comment with this reasoning why it has to stay there.

web/src/components/JsonLdWebSite.tsx
steffenkleinle: Missed that, thanks. Where does `Local & Multilingual Information |` come from?

sarahsporck: I guess that's supposed to be the title of a site. But not sure of which site. I guess we could use this as title for the landing page as "Regionen" does not supply much information

steffenkleinle (approved):

PR #826, opened 2022-12-06
IGAPP-636: Fix snackbar and remove redux store

steffenkleinle:


steffenkleinle:
Require cycle already fixed here: https://github.com/digitalfabrik/integreat-app/pull/823/commits/73276d9c60b5bf3e35feb2b94c6a838960163883 in #823

native/src/components/PoiDetails.tsx
```diff
@@ -70,12 +70,12 @@ const PoiDetails = ({ poi, feature, language }: PoiDetailsProps): ReactElement =
...
   const openExternalMaps = () => {
     const externalMapsUrl = getExternalMapsLink(poi.location, Platform.OS)
-    openExternalUrl(externalMapsUrl).catch(() => showSnackbar(t('error:noSuitableAppInstalled')))
+    openExternalUrl(externalMapsUrl).catch(() => showSnackbar({ text: t('error:noSuitableAppInstalled') }))
   }
 
   const copyLocationToClipboard = (): void => {
     Clipboard.setString(`${address}, ${postcode} ${town}`)
-    showSnackbar(t('addressCopied'))
+    showSnackbar({ text: t('addressCopied') })
```
f1sh1918, line 78: why do we need the `SnackbarType` that includes `text:string` instead of just passing a `string`

f1sh1918 (approved):
works fine! i think you could remove all the sagas staff (or in a separate PR)

steffenkleinle:
> works fine! i think you could remove all the sagas stuff (or in a separate PR)

Not yet possible bc
1. Types are imported by a lot of files from the sagas
2. Fetch resource cache logic is not yet migrated

Will be done in a separate PR.

native/src/components/PoiDetails.tsx
steffenkleinle: Makes it more flexible, in the future we quite sure will need to pass additional info to the snackbar (buttons and onClick methods for example). E.g. for this issue: https://issues.tuerantuer.org/browse/IGAPP-328

Since I was already refactoring I figured I could already change the types/usages here.

steffenkleinle:
@f1sh1918 sorry for the force push, didnt change anything so no need to review again.

LeandraH (approved):
Looks good, tested on iOS and Android

PR #832, opened 2022-12-15
IGAPP-1100: Opening hours native

f1sh1918:

Go to testumgebung -> check Tür an Tür Cafe, WWK Arena & Olympiastadium for some example of opening hours, closed, temp closed


native/src/components/OpeningHours.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </OpeningLabel>
+    </TitleContainer>
+  )
+
+  if (!openingHours) {
+    if (isTemporarilyClosed) {
+      return <TitleContainer language={language}>{openingHoursTitle}</TitleContainer>
+    }
+    return null
+  }
```
steffenkleinle, line 75:
```suggestion
  if (isTemporarilyClosed) {
    return <TitleContainer language={language}>{openingHoursTitle}</TitleContainer>
  }
  if (!openingHours) {
    return null
  }
```

native/src/components/PoiDetails.tsx
```diff
@@ -123,9 +124,18 @@ const PoiDetails = ({ poi, feature, language }: PoiDetailsProps): ReactElement =
...
         </>
       )}
+      <>
+        <OpeningHours
+          language={language}
+          openingHours={openingHours}
+          isCurrentlyOpen={isCurrentlyOpen}
+          isTemporarilyClosed={temporarilyClosed}
+        />
+        {((openingHours && openingHours.length > 0) || temporarilyClosed) && <HorizontalLine />}
```
steffenkleinle, line 134: Optional: Does it make sense to move the horizontal line to the collapsible item/OpeningHours components. Basically we render the horizontal line whenever we render a collapsible item or some other information, right?

steffenkleinle (commented):
> Go to testumgebung -> check Tür an Tür Cafe, WWK Arena & Olympiastadium for some example of opening hours, closed, temp closed

Only opening hours for TaT available, perhaps data got deleted? Always add to live cms (or maybe we should talk to cms team about this issue).
If I don't refresh the data, the app crashes:


Also, while it is perfectly aligned, for my eye it looks a little off (i.e. opening hours not aligned to the day, probably just bc of the last entry with two times). Not sure if that is just me and if we could/should do anything about it, just wanted to mention it.

native/src/components/OpeningHours.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <CollapsibleItem headerContent={openingHoursTitle} language={language} initExpanded>
+      <Content>
+        {openingHours.map((entry, index) => (
+          <OpeningEntry
+            key={`${weekdays[index]!}-OpeningEntry`}
+            weekday={weekdays[index]!}
```
SteffiStuffel, line 82: Weekdays should also be translated.

SteffiStuffel (approved):
Tested on Android emulator. Looks good.

native/src/components/OpeningEntry.tsx
```diff
@@ -0,0 +1,68 @@
...
+  flex-direction: ${props => contentDirection(props.language)};
+  justify-content: space-between;
+  padding: ${props => (contentDirection(props.language) === 'row' ? '4px 36px 4px 0' : '4px 26px 4px 0')};
+`
+const Timeslot = styled.View`
+  display: flex;
+  flex-direction: column;
+`
+
+const TimeSlotEntry = styled.Text<{ isCurrentDay: boolean; notFirstChild?: boolean }>`
```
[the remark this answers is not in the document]
f1sh1918: not sure if its better readable because i set explicity style for `notFirstChild` so i would have to negate in the styled component
` ${props => !props.firstChild && 'margin-top: 8px'};`
Don't think its nicer tbh.

native/src/components/PoiDetails.tsx
f1sh1918: but the downside is you have to replicate it, but okay

f1sh1918:
> > Go to testumgebung -> check Tür an Tür Cafe, WWK Arena & Olympiastadium for some example of opening hours, closed, temp closed
> 
> Only opening hours for TaT available, perhaps data got deleted? Always add to live cms (or maybe we should talk to cms team about this issue). If I don't refresh the data, the app crashes: 
> 
> Also, while it is perfectly aligned, for my eye it looks a little off (i.e. opening hours not aligned to the day, probably just bc of the last entry with two times). Not sure if that is just me and if we could/should do anything about it, just wanted to mention it. 

@steffenkleinle 
added data to both systems. Should work now. The alignment was required by design team. i can't see that big issue tbh.

native/src/components/OpeningHours.tsx
f1sh1918: okay then we have to add them in `translations` i guess, since we need to keep the german weekday order starting with monday (since we get it like that from cms)

steffenkleinle (approved):

PR #833, opened 2022-12-19
IGAPP-636: Actually load whole city content

steffenkleinle:


api-client/src/endpoints/loadSprungbrettJobs.ts
```diff
@@ -0,0 +1,55 @@
+import NotFoundError from '../errors/NotFoundError'
```
steffenkleinle, line 1: Just moved from `useLoadSprungbrettJobs` for reusing in web

native/src/hooks/useLoadExtraCityContent.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback } from 'react'
```
steffenkleinle, line 1: As the comment says this is just for easier handling if city content (i.e. offline available content) and something else needs to be loaded at the same time. Otherwise both responses (data, loading, error,  refresh) would need to be checked. Used for disclaimer, offers, sprungbrett offers.

native/src/hooks/useLoadCityContent.ts
```diff
@@ -1,49 +1,99 @@
-import { useCallback } from 'react'
-
-import { CityModel, Endpoint, ErrorCode, LanguageModel, ReturnType, useLoadAsync } from 'api-client'
+import moment from 'moment'
```
steffenkleinle, line 1: This is sadly a little hard to review, let me know if you have ideas for improvements here. Basically the changes is just that we now always load all the offline available content here now at once here (languages, categories, events, pois) instead of heaving multiple hooks that do it separately. Only if everything succeeds, the whole loading of the hook succeeds. This is basically the behavior that was used before in our `loadCityContent` saga. This is also necessary at that point because we only store one last update time stamp.

sarahsporck (commented):
Only roughly reviewed the code as there are a lot of changes.
Tested in Firefox and the web app is incredibly fast now! Nice 👍🏼 

In mobile I get cannot store last update for unused city. Cannot find city 'augsburg' in metadata.

LeandraH (approved):
Very cool! Tested in Firefox, iOS and Android, feels quite snappy to me too

PR #834, opened 2022-12-19
IGAPP-636: Subscribe to push notifications and update last usage

steffenkleinle:

### DIFF
--- a/native/src/contexts/AppContextProvider.tsx
+++ b/native/src/contexts/AppContextProvider.tsx
@@ -2,6 +2,7 @@ import React, { createContext, ReactElement, useCallback, useEffect, useMemo, us
 
 import { useLoadAsync } from 'api-client'
 
+import useSubscribePushNotifications from '../hooks/useSubscribePushNotifications'
 import LoadingErrorHandler from '../routes/LoadingErrorHandler'
 import appSettings from '../utils/AppSettings'
 import { dataContainer } from '../utils/DefaultDataContainer'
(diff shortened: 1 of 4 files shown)


native/src/contexts/AppContextProvider.tsx
```diff
@@ -28,15 +30,26 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
...
   const settingsResponse = useLoadAsync(loadSettings)
+  const allowPushNotifications = settingsResponse.data?.allowPushNotifications ?? null
+
+  useEffect(() => {
+    if (cityCode) {
+      dataContainer.storeLastUsage(cityCode).catch(reportError)
+    }
+  }, [cityCode])
+
+  useSubscribePushNotifications({ cityCode, languageCode, allowPushNotifications })
```
steffenkleinle, line 52: Not sure if this is the right place for this, let me know if you think it should be moved somewhere else.

sarahsporck (commented):
PN didnt really work. Did you test this?

steffenkleinle:
> PN didnt really work. Did you test this?

TBH not sure anymore, planned to at least haha, will check

steffenkleinle:
@sarahsporck can not reproduce, works for fine for me (at least with a newly installed app).

SteffiStuffel:
Could receive a PN, but get this error:

SteffiStuffel (approved):

steffenkleinle:
> Could receive a PN, but get this error:
> 
> The 'navigation' object hasn't been initialized yet. This might happen if you don't have a navigator mounted, or if the navigator hasn't finished mounting. See https://reactnavigation.org/docs/navigating-without-navigation-prop#handling-initialization for more details. ERROR [TypeError: undefined is not a function]

This issue is already known and existed before the refactoring, would therefore ignore it for now: 
https://issues.tuerantuer.org/browse/IGAPP-1085

steffenkleinle:
@sarahsporck seems that the app has to be reinstalled for it to work. We just have to check for production upgrades if that is the case as well which would not be good.

sarahsporck:
Ok will try again. With newly installed app.

sarahsporck (approved):
Tested again. I didn't have the app in background 🤦🏼 Works fine 👍🏼

PR #846, opened 2023-01-18
IGAPP-1153: Use url slug in poi test

sarahsporck:


web/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -70,7 +81,7 @@ describe('PoisPage', () => {
     expect(getByText('error:unknownError')).toBeTruthy()
   })
   // TODO IGAPP-1153: fix tests for Pois with URL Slug instead of just skipping
```
LeandraH, line 83:
```suggestion
 
```
I think that's done :)

LeandraH (approved):
Nice!

PR #847, opened 2023-01-18
IGAPP-636: Fix tests and checks

steffenkleinle:

### DIFF
--- a/native/src/__tests__/Navigator.spec.tsx
+++ b/native/src/__tests__/Navigator.spec.tsx
@@ -108,6 +108,7 @@ jest.mock('../utils/PushNotificationsManager', () => ({
   pushNotificationsSupported: jest.fn(() => true),
   quitAppStatePushNotificationListener: jest.fn(),
 }))
+jest.mock('../utils/FetcherModule')
 
 const cityCode = 'augsburg'
 const languageCode = 'de'


native/src/routes/__tests__/TuNews.spec.tsx
```diff
@@ -38,78 +48,77 @@ const news: [TunewsModel, TunewsModel] = [
...
+  it('should show language selector if language not available', async () => {
+    mocked(useLoadTuNews).mockImplementation(() => ({
+      ...tuNewsResponse,
+      availableLanguages: languages.slice(1),
+    }))
 
-    const { getByText, queryByText } = renderNews({ language: 'es' })
-    expect(getByText('languageNotAvailable')).toBeTruthy()
+    const { getByText, queryByText } = renderNews()
+    await waitFor(() => expect(getByText('languageNotAvailable')).toBeTruthy())
```
sarahsporck, line 120: findByText already wraps waitFor internally and should therefore also work here
```suggestion
    expect(await findByText('languageNotAvailable')).toBeTruthy()
```

sarahsporck (approved):
👍🏼 In `useLoadAsync.spec.tsx` you need to clear all mocks in before each. Everything else seems fine. Good job :)

SteffiStuffel (approved):
Looks good.

Wouldn't it be better to keep `react/jsx-no-constructed-context-values` rule active, but add useMemo everywhere. As this is a quite important feature, I would prefer the tests to check the production code as precise as possible.

steffenkleinle:
> Looks good.
> 
> Wouldn't it be better to keep `react/jsx-no-constructed-context-values` rule active, but add useMemo everywhere. As this is a quite important feature, I would prefer the tests to check the production code as precise as possible.

Readded rule and fixed that in generally by moving the constructed object out of the rendering function. Was not possible in only one case where I disabled the rule manually. Using useMemo is usually not possible bc it is not a hook/react component.

PR #850, opened 2023-01-24
IGAPP-1200: Fix navigation bar arrows

steffenkleinle:


steffenkleinle:
Fixes unresponsive behavior of arrows on first render and coloring of right arrow if scrolled to end.

native/src/components/NavigationTiles.tsx
```diff
@@ -52,16 +53,15 @@ const NavigationTiles = ({ tiles, theme }: NavigationTilesProps): ReactElement =
     setPercentageScrolled(contentOffset.x / (contentSize.width - layoutMeasurement.width))
   }
 
+  const scrolledToStart = percentageScrolled === 0
+  const scrolledToEnd = percentageScrolled >= scrolledToEndThreshold
+
+  const scrollToStart = () => scrollViewRef.current?.scrollTo({ x: 0, y: 0, animated: true })
+  const scrollToEnd = () => scrollViewRef.current?.scrollToEnd({ animated: true })
```
LeandraH, line 60: I guess this is fine as long as we don't get more icons there. Which I think is not planned?

LeandraH (approved):
Looks good, tested on Android emulator

native/src/components/NavigationTiles.tsx
steffenkleinle: Agreed and yes, don't think more icons are planned :+1:

f1sh1918 (commented):
The related issue is fixed so far but when using horizontal scrolling by touch the arrows were not disabled if they are reaching the end or start

https://user-images.githubusercontent.com/37902063/214334140-8955466c-03bb-42e0-b84a-c007e6af5c05.mp4

steffenkleinle:
> The related issue is fixed so far but when using horizontal scrolling by touch the arrows were not disabled if they are reaching the end or start, f.e. scrolling to "Map" should disabled right arrow
> Simulator.Screen.Recording.-.iPhone.SE.2nd.generation.-.2023-01-24.at.16.18.32.mp4

Can't reproduce, can you play around with the thresholds a bit? What is the percentage scrolled that is left?

f1sh1918 (approved):
Bug only happens on ios if the horizontal scrolling is strong. Dunno if we need a bug ticket for that

PR #853, opened 2023-01-24
IGAPP-328: Foreground push notifications

steffenkleinle:


steffenkleinle:
Only implemented the logic part, the design part should be done in https://issues.tuerantuer.org/browse/IGAPP-1204.
To test this, uncomment the corresponding showSnackbar block in `useForegroundPushNotificationListener` in `PushNotificationManager` (look for the TODO).

f1sh1918 (approved):
Works fine :)
I just thought that we could increase maybe the `SHOW_DURATION` i think if you are using the app and immediately the foreground notification come up you can be too slow.
Maybe we can keep `5000` as default and pass a optional value to override it
https://github.com/digitalfabrik/integreat-app/pull/859 is just a proposal

native/src/utils/PushNotificationsManager.ts
```diff
@@ -71,16 +77,47 @@ export const subscribeNews = async (city: string, language: string): Promise<voi
...
+            // TODO IGAPP-1024: Uncomment and improve snackbar
+            log(JSON.stringify(message))
+            // showSnackbar({
+            //   text: message.notification.title,
+            //   positiveAction: {
+            //     onPress: () => navigate(NEWS_ROUTE, routeInformationFromMessage(message)),
+            //     label: 'Show',
+            //   },
+            // })
+          }, WAITING_TIME_FOR_CMS)
```
sarahsporck, line 113: So this does not work without the setTimeout? I thought on message only triggers when there already is a message.

sarahsporck (approved):
Tested on Android

native/src/utils/PushNotificationsManager.ts
steffenkleinle: There is already a message, yes. But its not delivered by the cms via the local news api. I will maybe ask them about that.

steffenkleinle: So without the timeout the app just says "page not found" since there is no local news yet with that id if you click on the snackbar fast

steffenkleinle:
> Works fine :) I just thought that we could increase maybe the `SHOW_DURATION` i think if you are using the app and immediately the foreground notification come up you can be too slow. Maybe we can keep `5000` as default and pass a optional value to override it #859 is just a proposal

Agreed, see my comment in your PR. I will create a separate proposal and merge this one already since it won't be enabled on production anyway so far.

f1sh1918:
> > Works fine :) I just thought that we could increase maybe the `SHOW_DURATION` i think if you are using the app and immediately the foreground notification come up you can be too slow. Maybe we can keep `5000` as default and pass a optional value to override it #859 is just a proposal
> 
> Agreed, see my comment in your PR. I will create a separate proposal and merge this one already since it won't be enabled on production anyway so far.

Works fine, tested on android. I'm fine with merging this pr

PR #854, opened 2023-01-24
IGAPP-1085: News not opened upon click on push notifications

steffenkleinle:


f1sh1918 (approved):
Tested on ios real device

PR #856, opened 2023-01-25
IGAPP-1203: Fix timeslots

f1sh1918:
Please test if you can open location with timeslots f.e. on Android and iOS
Tested on both devices

https://integreat.app/testumgebung/de/locations/cafe-t%C3%BCr-an-t%C3%BCr

### DIFF
--- a/native/src/utils/DatabaseConnector.ts
+++ b/native/src/utils/DatabaseConnector.ts
@@ -114,7 +114,7 @@ type ContentPoiJsonType = {
   excerpt: string
   location: LocationJsonType<number>
   lastUpdate: string
-  openingHours: OpeningHoursModel[] | null
+  openingHours: { allDay: boolean; closed: boolean; timeSlots: { start: string; end: string }[] }[] | null
   temporarilyClosed: boolean
 }
 type CityCodeType = string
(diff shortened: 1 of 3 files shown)


SteffiStuffel (approved):
Note tested as I currently can not start the app. Code looks good.

steffenkleinle (changes requested):
The api response includes the timeslots, I guess this is because timeslots are not saved/loaded correctly from offline files. I think the problem may be that you are just saving the timeslots like this:
``` js
        openingHours: poi.openingHours,
```
Since `poi.openingHours` is a class instance, it gets saved like following (mind the `_` prefix):
``` json
  [{"_allDay":true,"_closed":false,"_timeSlots":[]}, ...]
```
When loading, this obviously fails bc there is no mapping to a new class instance (and e.g. `allDay` is undefined while `_allDay` is defined).

Therefore, when saving stuff permantently to files, you should always destructure it property by property, i.e. something like:
``` js
        openingHours: poi.openingHours.map(hours => ({
          allDay: hours.allDay,
          closed: hours.closed,
          timeslots: hours.timeslots.map(timeslot => ({
            start: timeslot.start,
            end: timeslot.end
          })
        }),
```

Also you always set `temporarilyClosed: true` in `loadPois` in `DatabaseConnector`

Edit: Also the JSON type should include all properties that are saved explicitly.

steffenkleinle (approved):
Tested on android

PR #859, opened 2023-01-25
IGAPP-328: foreground push notification custom duration

f1sh1918:
As i mentioned in the other PR. I think for foreground push notifications the `SHOW_DURATION` is too short, so here my proposal


f1sh1918:
@steffenkleinle ok perfect. this was more or les just a short draft. thx for your proposal

PR #864, opened 2023-01-30
IGAPP-1209: App crashes openingHours

f1sh1918:
- Added explicit moment locale import to fix minified import error

Reference: https://github.com/moment/moment/issues/5789

**Note:** can only be tested with release apk since error is only occurring when minifying js with hermes
(Enabling "JS Minify" in Dev Settings did not work for me to reproduce the error)

`cd android`
`./gradlew assembleRelease`
or [releaseApk](https://output.circle-artifacts.com/output/job/6732a316-7f57-4fb5-bb81-07589e139004/artifacts/0/native/android/integreat-e2e.apk)

Go to map -> click on a poi list item with OpeningHours f.e."Olympiastadion"

### DIFF
--- a/api-client/src/models/PoiModel.ts
+++ b/api-client/src/models/PoiModel.ts
@@ -1,5 +1,5 @@
 import moment, { Moment } from 'moment'
-// Fix for this issue https://github.com/moment/moment/issues/5789
+// Fix for minifying js issue with hermes using moment().locale https://github.com/moment/moment/issues/5789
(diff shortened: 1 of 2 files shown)


SteffiStuffel (approved):
Tested, looks good.

But something with the calculation is not working properly on language change, I will open a ticket for it.

sarahsporck (approved):
Tested 👍🏼

PR #865, opened 2023-01-31
IGAPP-1211: Wrong temporarily closed cache

f1sh1918:
Testumgebung -> Map -> check poi that is not marked as temp closed in cms f.e. "Tür an Tür Cafe"


SteffiStuffel (approved):
Not tested.

sarahsporck (approved):
Only executed. Not specifically tested.

PR #866, opened 2023-01-31
IGAPP-1214: App crashes with no city context

f1sh1918:
Since Aschaffenburg has no location selected because it's fixed to 'hallo' (cityCode) we have to provide it in AppContext (Refactoring issue)

**Testing**

- Build and Start Aschaffenburg App
- City "hallo Aschaffenburg" should be preselected


LeandraH (approved):
Nice, tested on emulated iOS

steffenkleinle (commented):
good find!

PR #867, opened 2023-01-31
IGAPP-1198: Upgrade e2e dependencies

sarahsporck:

- updated dependencies (of course)
- removed expect-webdriverio and therefore all expect...toExist
    - wdio will still fail when an element cannot be found
- refactored swiping motion with native implementation


SteffiStuffel (approved):
Looks good. Thank you :)

LeandraH (approved):
Nice!

PR #871, opened 2023-02-01
IGAPP-1178: Poi categories web

f1sh1918:
Show Poi Categories on web


LeandraH (approved):
Nice, tested in Firefox

release-notes/unreleased/IGAPP-1178-poi-categories-web.yaml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1178
+show_in_stores: true
+platforms:
+  - web
+en: Categories of Location will be shown.
```
SteffiStuffel, line 5: of **l**ocation**s**

web/src/components/PoiListItem.tsx
```diff
@@ -38,6 +38,11 @@ const Distance = styled.div`
   font-size: clamp(0.55rem, 1.6vh, ${props => props.theme.fonts.hintFontSize});
 `
 
+const Category = styled.div`
+  font-size: clamp(0.55rem, 1.6vh, ${props => props.theme.fonts.hintFontSize});
```
SteffiStuffel, line 42: Why such an complicated way to define the font-size? Can't this be simplified?
Also as this same computation with magic number is duplicated, please move it to one location and import it in the other.

SteffiStuffel (commented):
Tested on Chrome, looks good, but I think size of the icons should be adjusted. When a poi is selected the marker-icon is smaller than the previous category icon, but when a poi is selected it should be more eye-catching:
not-selected
selected

web/src/components/PoiListItem.tsx
f1sh1918: because its the way how to use different paddings and font-size using the viewport width (CSS3 feature). If we use equal paddings and font-size it will look odd in different desktop screen sizes. 
I will define the sizes and reuse them

f1sh1918:
@SteffiStuffel is that fine for you?
image

PR #874, opened 2023-02-06
IGAPP-1223: Fix testumgebung

SteffiStuffel:

As deleting the cache worked for my real devices, I guess this PR fixes the issue, but I am not really sure, as I do not get this error on the emulator, but there was a warning and an error in the console, which disappeared with these changes. Hopefully someone of you has a good idea how to test this, otherwise we have to try.


f1sh1918:
> 
> As deleting the cache worked for my real devices, I guess this PR fixes the issue, but I am not really sure, as I do not get this error on the emulator, but there was a warning and an error in the console, which disappeared with these changes. Hopefully someone of you has a good idea how to test this, otherwise we have to try.

@SteffiStuffel 
I think this issue is related to saving metaData with the same location name like "testumgebung" for test-cms if you have already saved data for the "testumgebung" in production cms.
I'm facing this error `'cannot store last update for unused city'` but i cannot reproduce Leandra's issue, so not 100% sure if there is a relation. Even the bug report is not helping tbh, so maybe a clear bug report could help reproducing the error.

Would be good if leandra can test it, because it can not reproduce and i'm not sure if we should merge it without more investigation. Even because i think it only happens if you switch cms and won't happen for users

SteffiStuffel:
Will add better cache-naming for test-cms and prod-cms cities and reopen this, when I am finished

LeandraH:
It's unfortunately not fixed with this. I attached a video but I'm not sure how helpful that is either. In the console, I'm getting a warning `Sending `progress` with no listeners registered.`

https://user-images.githubusercontent.com/18513943/217550369-a409ff74-4348-4808-95ca-8f04aa52c22a.mp4

f1sh1918:
created new pr with proper branch naming

PR #875, opened 2023-02-08
IGAPP-1218: investigate loading times - fonts

f1sh1918:
This is just a small part of the optimizations but can be merged in my opinion
- added `display:swap` to fonts to use system font fallback when font is not loaded
- small a11y fixes


web/www/fonts/noto-sans-sc/noto-sans-sc.css
```diff
@@ -1,29 +1,29 @@
...
-    url('./noto-sans-sc-v12-latin_chinese-simplified-regular.woff') format('woff'), /* Modern Browsers */
-    /* TTF is not available for SC */
-    /* url('./noto-sans-sc-v12-latin_chinese-simplified-regular.ttf') format('truetype'), */ /* Safari, Android, iOS */
-    url('./noto-sans-sc-v12-latin_chinese-simplified-regular.svg#NotoSansSC') format('svg'); /* Legacy iOS */
+  font-family: 'Noto Sans SC';
+  font-style: normal;
+  font-weight: 400;
+  src: url('./noto-sans-sc-v12-latin_chinese-simplified-regular.eot'); /* IE9 Compat Modes */
+  src: local(''), url('./noto-sans-sc-v12-latin_chinese-simplified-regular.eot?#iefix') format('embedded-opentype'),
+    /* IE6-IE8 */ url('./noto-sans-sc-v12-latin_chinese-simplified-regular.woff2') format('woff2'),
```
sarahsporck, line 8: did you use a different formatter here? or is this change intended?

sarahsporck (approved):
Tested in safari.

web/www/fonts/noto-sans-sc/noto-sans-sc.css
f1sh1918: Looks like. I think we do not Format css with prettier?

PR #876, opened 2023-02-08
IGAPP-1227: dynamically wait for localhost in e2e tests

sarahsporck:

### DIFF
--- a/e2e-tests/web/wdio.conf.ts
+++ b/e2e-tests/web/wdio.conf.ts
@@ -1,6 +1,6 @@
 import { Capabilities } from '@wdio/types/build/Capabilities'
-
 import { browsers, ciCapabilities } from './capabilities'
+import waitForLocalhost from './waitForLocalhost'
 
 const getCapabilities = (): Array<Capabilities> => {
   if (process.env.CI) {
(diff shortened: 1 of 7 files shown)


LeandraH (approved):
Looks like a reasonable attempt, fingers crossed

e2e-tests/web/wdio.conf.ts
```diff
@@ -22,27 +23,29 @@ export const config: WebdriverIO.Config = {
...
 
   jasmineOpts: {
-    defaultTimeoutInterval: 300000,
+    defaultTimeoutInterval: 300_000,
   },
 
   onPrepare: async (): Promise<void> => {
     if (process.env.CI) {
-      const startupDelay = 20000
+      const startupDelay = 10_000
```
sarahsporck, line 39: a delay is still required for web pack to actually start. otherwise the tests will fail with econrefused.

f1sh1918:
hm pipeline still looks the same, even this seem to be a reasonable change :)

f1sh1918:
@sarahsporck 
i added this pr to fix pipeline
https://github.com/digitalfabrik/integreat-app/pull/878

f1sh1918 (approved):

sarahsporck:
Hmm. Okay. I guess downgrading is also reasonable. I'll investigate further why webpack cannot resolve the version, as it clearly is then related to node.

f1sh1918:
> Hmm. Okay. I guess downgrading is also reasonable. I'll investigate further why webpack cannot resolve the version, as it clearly is then related to node.

If we put more effort into it, I prefer upgrading instead of downgrading :)

PR #879, opened 2023-02-13
IGAPP-1229: Times of events will be exported correctly

f1sh1918:
Fixed endDate and added correct timezones to startDate & endDate

Pls export an event and import to a calendar und check times, title, description
http://localhost:9000/testumgebung/de/events/zweite-testveranstaltung


LeandraH (approved):
Nice, thanks for the quick fix! Tested in Firefox

SteffiStuffel (approved):
Not tested.

PR #881, opened 2023-02-18
IGAPP-1195: Use poi metadescription from cms

sarahsporck:


LeandraH (approved):
Nice, tested in Firefox.

Does this mean we can remove the excerpt or do we need to keep that for backward compatibility?

api-client/src/endpoints/createPOIsEndpoint.ts
```diff
@@ -31,6 +31,7 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<PoiModel>> =>
               thumbnail: poi.thumbnail,
               availableLanguages: mapAvailableLanguages(poi.available_languages),
               excerpt: poi.excerpt,
+              metaDescription: poi.meta_description ? poi.meta_description : null,
```
SteffiStuffel, line 34: You can just write `metaDescription: poi.meta_description ?? null`

SteffiStuffel (approved):
Tested in Chrome.

api-client/src/endpoints/createPOIsEndpoint.ts
sarahsporck: Does not work as the meta description is an empty string if it is not set and `"" ?? null` resolves to ""

sarahsporck:
> Nice, tested in Firefox.
> 
> Does this mean we can remove the excerpt or do we need to keep that for backward compatibility?

excerpt != meta_description.

> Nice, tested in Firefox.
> 
> Does this mean we can remove the excerpt or do we need to keep that for backward compatibility?

Excerpt seems to be different from the meta description.

PR #887, opened 2023-02-23
IGAPP-1222: Back navigation bottom sheet fullscreen

f1sh1918:
1. http://localhost:9000/testumgebung/de/locations/cafe-t%C3%BCr-an-t%C3%BCr
2. open mobile mode
3. snap bottom sheet to fullscreen
4. click back navigation


web/src/components/PoiDetails.tsx
```diff
@@ -169,7 +169,7 @@ const DetailSection = styled.div`
   display: flex;
   flex-direction: column;
 
-  @media ${dimensions.smallViewport} {
+  @media screen and ${dimensions.smallViewport} {
```
f1sh1918, line 172: just fixed some warnings

sarahsporck (approved):
Tested in Safari. ✅ 

This is mostly personal preference I guess, but I think we can keep always keep the back arrow in the sheet. It feels weird that it disappears after moving it back down...

SteffiStuffel (commented):
Without an animation (fade in the header) it looks bad.

LeandraH (approved):
I agree with Steffi, an animation would be nicer but I think it's also okay like this.

f1sh1918:
> Without an animation (fade in the header) it looks bad.

@SteffiStuffel i added a basic fadeIn animation for appearing. Gonna create a  ticket to improve animation also for fading out and disappearing the header div. This is not that easy.
https://issues.tuerantuer.org/browse/IGAPP-1254

PR #888, opened 2023-03-01
IGAPP-1168: Add caching for tile image

SteffiStuffel:

- caching will not work until cms PR is merged: https://github.com/digitalfabrik/integreat-cms/pull/2099


timoludwig (commented):
Tiles are the representation of first level categories, right?
Then this PR is not blocked by https://github.com/digitalfabrik/integreat-cms/pull/2099, because the icons and media images are directly served by the Apache server and not by Django. So the caching needs to be configured in the [Apache config on salt](https://git.tuerantuer.org/DF/salt/src/branch/master/states/integreat-cms/files/apache2.conf).

timoludwig:
> So if I understand you correctly, locally my settings are different and therefore django takes care of the caching, so my changes lead into the desired results, but this will not work with our apache server, as there the caching is not handled by django for these images?

Indeed, in the local dev environment there is no apache server running, so Django serves the static files. This is however only a convenience method to make developing easier (see [docs](https://docs.djangoproject.com/en/4.1/howto/static-files/#serving-static-files-during-development)), in production the apache server intercepts these routes and directly serves the media files instead of forwarding the request to Django.

SteffiStuffel:
> > So if I understand you correctly, locally my settings are different and therefore django takes care of the caching, so my changes lead into the desired results, but this will not work with our apache server, as there the caching is not handled by django for these images?
> 
> Indeed, in the local dev environment there is no apache server running, so Django serves the static files. This is however only a convenience method to make developing easier (see [docs](https://docs.djangoproject.com/en/4.1/howto/static-files/#serving-static-files-during-development)), in production the apache server intercepts these routes and directly serves the media files instead of forwarding the request to Django.

Oke thank you, opened a PR in gitea and closed the cms PR.

web/src/components/Tile.tsx
```diff
@@ -63,25 +65,38 @@ const TileContainer = styled.div`
...
+
+  const fetchImageWithCaching = (): void => {
+    request(tile.thumbnail, {})
+      .then(res => res.blob())
+      .then(blob => {
+        if (imageRef.current) {
+          imageRef.current.src = URL.createObjectURL(blob)
+        }
+      })
+  }
```
steffenkleinle, line 80: How does this add caching?  Do you have some docs for that?

LeandraH (approved):
Nice! Tested in Firefox, where the icons indeed are shown as being cached :)

steffenkleinle (approved):
Tested on firefox. We only want/need this for the tiles? Maybe we could introduce a custom image component which uses this as built-in caching which we use everywhere?

f1sh1918:
@SteffiStuffel is there a reason to keep this pr open? for me steffens comment would be nice, but we could do it in a separate pr

PR #892, opened 2023-03-01
IGAPP-1172: Restore scroll position poi deselection

f1sh1918:
Note: On native the restore of the `listScrollPosition` only works if BottomActionSheet is fullscreen, because you basically can not scroll if you are not at highest `snapPosition(fullscreen)`

We could create an issue for this. Last one was closed due to missing template
https://github.com/gorhom/react-native-bottom-sheet/issues/917

Tested on Chrome(web), android(native) & ios(Simulator)


native/src/routes/Pois.tsx
```diff
@@ -59,29 +60,47 @@ type PoisProps = {
...
+
+  const scrollTo = (position: number) => {
+    setTimeout(() => {
+      if (scrollRef.current) {
+        scrollRef.current.scrollTo({
+          y: position,
+          animated: false,
+        })
+      }
+    }, RESTORE_TIMEOUT)
```
f1sh1918, line 88: couldn't get it work without timeout, event when i trigger it in the useEffect after `setCamera` something seems not to be ready

web/src/components/BottomActionSheet.tsx
```diff
@@ -45,13 +47,17 @@ const BottomActionSheet = React.forwardRef(
   ): ReactElement => {
     const theme = useTheme()
     const listRef = useRef<HTMLDivElement>(null)
+    const previousPath = useLocation().state?.from?.pathname
```
f1sh1918, line 50: or maybe `previousFeaturePath`

web/src/components/MapView.tsx
```diff
@@ -91,41 +91,33 @@ const MapView = forwardRef((props: MapViewProps, ref: React.Ref<MapRef>): ReactE
   const [cursor, setCursor] = useState<MapCursorType>('auto')
   const theme = useTheme()
   const navigate = useNavigate()
+  const location = useLocation()
 
   const { viewportSmall } = useWindowDimensions()
 
-  const onDeselectFeature = useCallback(
```
f1sh1918, line 97: this seem not to be needed anymore, since the MapControl do not affect a deselection maybe due to lib version update, so i added a regular `onDeselect`  in `onSelectFeature`

LeandraH (approved):
Nice! Tested on Firefox and simulated iOS. One small thing: on iOS, I am returned to the actual scroll point I was at whereas in the browser, I am returned to a different scroll point where the selected POI is at the top of the view. I don't think it's an important difference though.

f1sh1918:
> Nice! Tested on Firefox and simulated iOS. One small thing: on iOS, I am returned to the actual scroll point I was at whereas in the browser, I am returned to a different scroll point where the selected POI is at the top of the view. I don't think it's an important difference though.

Yes i used two different technologies for web & native.
Web -> scrollTo elementID which is super nice and only possible in DOM
native: scrollTo scrollPosition, since we would have to track `onLayout` all the coordinates of all `PoiListItems` which is very resource intense in some case like munich (150+ pois)

sarahsporck (approved):
Tested in safari.
When you select the poi via the map, the scroll position will "falesly" be restored to the position of the selected poi. But that sth. I am really not mad about.

Tested on android emulator. 
When deselecting the poi via the map the scroll position is not restored. I think this could be easily fixed, right?

f1sh1918:
> Tested in safari. When you select the poi via the map, the scroll position will "falesly" be restored to the position of the selected poi. But that sth. I am really not mad about.
> 
> Tested on android emulator. When deselecting the poi via the map the scroll position is not restored. I think this could be easily fixed, right?

@sarahsporck you can not restore the scroll position when selecting on the map because you basically didn't scroll to it on the list so how should you know the scroll position. I think the restore of the position should only happen if you select it  on the list. So i'm gonna change that in the web app i think.

sarahsporck:
> 1. I think the restore of the position should only happen if you selected it on the list even for web when you scroll by elementId. So i adjusted it to only scroll if you selected the poi from the list.

Jep. As stated, in web it really doesn't bother me. It is just a small inconsistency.
And in native 2. seems to prevent restoring either way, so that's also fine then.

> 2. As I mentioned in the description the scrolling inside of the bottom Sheet is only enabled at it's highest snappoint which is fullscreen (in our case) , so if you snap it down to a lower point and deselect the poi scrolling does not work. I could investigate some time to patch the library or create an issue (at least)

Ahh okay. I missed that comment. Either way, nice work. Thank you for investigating that. 👍🏼

PR #895, opened 2023-03-08
IGAPP-1224: Fix search result highlights

sarahsporck:


api-client/src/utils/getExcerpt.ts
```diff
@@ -37,27 +37,34 @@ type ExcerptOptions = {
 }
 
 const getExcerpt = (text: string, { query, maxChars, replaceLineBreaks }: ExcerptOptions): string => {
-  const match = query ? normalizeString(text).match(normalizeString(query)) : []
```
sarahsporck, line 40: Before it tried to also match regex strings, meaning writing `?` would break the search

LeandraH: Ouch, good catch!

LeandraH (approved):
Very nice, tested in Firefox and iOS emulator

f1sh1918:
Nice find! Just a small question to "Umlaute"
Should this match?
image

sarahsporck:
> Also "u" matches "ü" i know this is not related to your pr. Just a general question :)

Jep. This should match. That's the reason why we normalize all our strings in the search.

f1sh1918 (approved):

PR #900, opened 2023-03-16
IGAPP-1248: New icons for language, location, and search

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/IGAPP-1248 .
NB: The language icon changes directions if the font is written from right to left, but only if the system language is also right to left.


assets/icons/location.svg
sarahsporck, line 1: somehow this icon feels too small compared to the other ones. Did you use the correct svg from figma?:
```svg
<svg width="14" height="20" viewBox="0 0 14 20" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path d="M7 0C3.13 0 0 3.13 0 7C0 12.25 7 20 7 20C7 20 14 12.25 14 7C14 3.13 10.87 0 7 0ZM2 7C2 4.24 4.24 2 7 2C9.76 2 12 4.24 12 7C12 9.88 9.12 14.19 7 16.88C4.92 14.21 2 9.85 2 7Z" fill="black"/>
  <path d="M7 9.5C8.38071 9.5 9.5 8.38071 9.5 7C9.5 5.61929 8.38071 4.5 7 4.5C5.61929 4.5 4.5 5.61929 4.5 7C4.5 8.38071 5.61929 9.5 7 9.5Z" fill="black"/>
</svg>
```

assets/icons/search.svg
sarahsporck, line 1: same as with location. I think this should be right:
```svg
<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path d="M12.755 11.255H11.965L11.685 10.985C12.665 9.845 13.255 8.365 13.255 6.755C13.255 3.165 10.345 0.255001 6.755 0.255001C3.165 0.255001 0.255 3.165 0.255 6.755C0.255 10.345 3.165 13.255 6.755 13.255C8.365 13.255 9.845 12.665 10.985 11.685L11.255 11.965V12.755L16.255 17.745L17.745 16.255L12.755 11.255ZM6.755 11.255C4.265 11.255 2.255 9.245 2.255 6.755C2.255 4.265 4.265 2.255 6.755 2.255C9.245 2.255 11.255 4.265 11.255 6.755C11.255 9.245 9.245 11.255 6.755 11.255Z" fill="black"/>
</svg>
```

web/src/components/CityContentHeader.tsx
```diff
@@ -64,31 +61,27 @@ const CityContentHeader = ({
 
   const { t } = useTranslation('layout')
 
+  const SearchButton = (
+    <HeaderActionBarItemLink
```
sarahsporck, line 65: When navigating to the search I see that the search icon in the search directly is also inverted when changing language. So I think we should either also invert the header icon or not revert the search icon next to the search query:
Bildschirm­foto 2023-03-16 um 11 34 12

sarahsporck (commented):
Tested on Android emulator.
-> Language icon is not inverted for RTL languages in native.

Tested in Safari.
-> Language icon sometimes jumps a little bit, e.g.

web/src/components/CityContentHeader.tsx
LeandraH: Nice catch, we are already doing that in main but that's very much not in the design. I'll remove the inversion of the search symbol.

LeandraH:
> Tested on Android emulator. -> Language icon is not inverted for RTL languages in native.

It is inverted on my Android emulator (though only when the system language is changed, the app language isn't enough). Can you tell me more about which virtual device etc you are using?
image

sarahsporck:
> It is inverted on my Android emulator (though only when the system language is changed, the app language isn't enough). 

Oh you're totally right here. I missed that I had to change the system language, too. 👍🏼

sarahsporck (approved):
Very nice! Tested again :)

In web I get the feeling that the icons are very prominent and could be a little smaller (e.g. 0.6 * headerHeightLarge). Maybe @f1sh1918. Or ask @hauf-toni directly.

Large:
Bildschirm­foto 2023-03-27 um 10 34 46

Small:
Bildschirm­foto 2023-03-27 um 10 32 29

LeandraH:
Toni agrees, they could be a bit smaller :)

f1sh1918:
> Toni agrees, they could be a bit smaller :)
i think the strokes are quite big and i think we should use a lighter color like this gray we use for the navigation items when they aren't hovered. It feels like they are slaying you :D

hauf-toni:
> > Toni agrees, they could be a bit smaller :)
> 
> i think the strokes are quite big and i think we should use a lighter color like this gray we use for the navigation items when they aren't hovered. It feels like they are slaying you :D

I wouldn't recommend using the same grey as the navigation items since the colors displays them as inactive which is not fitting for the search/language/location icons. however, choosing a more subtle black (e.g. #232323) is even preferred from an ui/ux perspective.

I agree with matching the strokes to the other icons 👍

hauf-toni:
@LeandraH has pointed out the inconsistency of the line width, thank you for that, an updated version of the design proposal for the icons plus a new icon for the category "map" can be found [here in figma](https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Jira?node-id=1025%3A3191&t=I2iVwiXxGlM0TrrW-1)

LeandraH:
Toni and I also figured out that the category icons were different between the native and the web version and decided to make those match.

SteffiStuffel (approved):
Nice work :+1: 

Looks good, tested on web and android

PR #901, opened 2023-03-20
IGAPP-1703: Upgrade yarn

sarahsporck:

The main change of yarn v2 and v3 is that dependencies are now not shared with child workspaces. As such we have to use global scripts (that use the global binary) to invoke global binaries, such as prettier or lint. Global scripts must include a `:`.
As a result jest and typescript are also not shared.

To improve jest usage I used the jest project config. Which over all improves the jest performance in ci too :)


sarahsporck:
For now rejected until yarn improves on ram in ci

PR #904, opened 2023-03-22
Add `zh` as fallback language code for `zh-CN`

timoludwig:

### DIFF
--- a/translations/src/config.ts
+++ b/translations/src/config.ts
@@ -139,6 +139,7 @@ class Config {
     fa_pr: ['pes'],
     'de-si': ['de'],
     sr: ['sr-Cyrl'],
+    zh: ['zh-CN'],
     'zh-hans': ['zh-CN'],
     // Slugs from the CMS are (and have to be) lowercase
     'sr-cyrl': ['sr-Cyrl'],


timoludwig:
Hmm, I wasn't expecting side effects by this change, in my mind this would just result in "if the backend sends `zh` as language slug, use `zh-CN` as UI language"...

But since simplified Chinese is only used in Aschaffenburg at the moment, I don't think it's worth looking into the failing tests, we will keep the slug as it is in the cms.

steffenkleinle:
@timoludwig do you want me to look into this?

timoludwig:
@steffenkleinle if it's not too much effort, sure :+1:
The benefit would be that Aschaffenburg could use DeepL to create machine translations for simplified Chinese.

steffenkleinle:
Sounds good, will have a look!

steffenkleinle:
Okay, I had a look at it and tbh I don't get what i18next is doing there and what is happening.... adding a new fallback should not break an existing other fallback. I also don't get why `ku` is resolved to `ku` even though there is a fallback to `kmr` while `zh-hans` is resolved to its fallback `zh-CN` in the first place. That adding a fallback from `zh` to `zh-CN` changes the output of `zh-hans` to `zh` makes it even weirder...

PR #905, opened 2023-03-29
IGAPP-919: Add the title of the previous route to the header

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/IGAPP-919

### DIFF
--- a/native/src/components/Header.tsx
+++ b/native/src/components/Header.tsx
@@ -177,7 +177,8 @@ const Header = ({
       return ''
     }
     const routes = navigation.getState().routes
-    if (routes.length <= 2) {
+    const levelsDisplayingCityName = 2
+    if (routes.length <= levelsDisplayingCityName) {
       return cityDisplayName(city)
     }
     const previousPath = (routes[routes.length - 2]?.params as { path: string }).path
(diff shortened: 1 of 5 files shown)


native/src/components/Header.tsx
```diff
@@ -167,13 +169,28 @@ const Header = ({
...
 
+  const { cityCode, languageCode } = useCityAppContext()
+  const { data } = useLoadCityContent({ cityCode, languageCode })
+
+  const getHeaderText = (): string => {
+    if (!city) {
+      return ''
+    }
+    const routes = navigation.getState().routes
+    if (routes.length <= 2) {
```
SteffiStuffel, line 180: Can you move the "2" to a variable with a naming that explains why this is two?

SteffiStuffel (approved):
Looks good. Tested on android emulator.

f1sh1918 (changes requested):
1. Not sure if this connected to this pr but when i click on "Veranstaltungen" and choose an Item, i get an Error and white screen



2.It seems like the navigation tile is only working for Sites (categories) when i go to News and choose one there is still the city name displayed (instead of "Nachrichten") and click on it leads me to News Overview. This should also work for "Veranstaltungen" which i couldn't test due to the error

PR #907, opened 2023-04-03
IGAPP-1257: Add missing focus

f1sh1918:
Just use tab to navigate through the page

### DIFF
--- a/release-notes/unreleased/IGAPP-1257-Add-missing-focus.yml
+++ b/release-notes/unreleased/IGAPP-1257-Add-missing-focus.yml
@@ -1,4 +1,4 @@
-issue_key: IGAPP-1172
+issue_key: IGAPP-1257
 show_in_stores: false
 platforms:
   - web
--- a/release-notes/unreleased/IGAPP-1257-Add-missing-focus.yml
+++ b/release-notes/unreleased/IGAPP-1257-Add-missing-focus.yml
@@ -1,0 +1,5 @@
+issue_key: IGAPP-1172
+show_in_stores: false
+platforms:
+  - web
+de: Fehlender Fokus für Bedienelemente wurde hinzugefügt.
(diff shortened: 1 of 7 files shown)


LeandraH (approved):
Tested in Firefox, works. I see now what the report meant about how hard it is to navigate our page with only a keyboard :/

release-notes/unreleased/IGAPP-1257-Add-missing-focus.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: IGAPP-1172
```
sarahsporck, line 1:
```suggestion
issue_key: IGAPP-1257
```

sarahsporck (approved):
Tested in Safari.
I guess you agreed on the default highlight color? In the issue another color was adressed.

f1sh1918:
> 

@sarahsporck the textSecondary color was addressed in the issue and should be used here? What do you mean by default color the greenish one?

sarahsporck:
> > 
> 
> @sarahsporck the textSecondary color was addressed in the issue and should be used here! What do you mean by default color the greenish one?

For me a blieish color is used, so I assumed it is the default one o.o

f1sh1918:
> 

image
this is #585858

LeandraH:
Huh. This is what it looks like on my machine (in Firefox), and the only thing I can focus on while tabbing is the language selector.
image

f1sh1918:
> Huh. This is what it looks like on my machine (in Firefox), and the only thing I can focus on while tabbing is the language selector. image

I fixed this by last commit

**Enable tap navigation in browser**

mac & firefox
https://stackoverflow.com/a/64151235

safari
https://osxdaily.com/2011/08/14/enable-tab-key-navigation-safari/

LeandraH:
True, I see full on gray frames now, and did have to enable the tabbing in the system settings, thanks for the hint :)

PR #910, opened 2023-04-12
IGAPP-1230: Map unscrollable

f1sh1918:
When many pois are placed at the same location and the user scrolls very close, `react-maps-gl` facing an rendering issue which is still not solved.

I think restricting `maxZoom: 21` should be fine since no one needs to zoom closer

### DIFF
--- a/api-client/src/maps/index.ts
+++ b/api-client/src/maps/index.ts
@@ -9,6 +9,7 @@ export type MapViewViewport = {
   latitude: number
   longitude: number
   zoom: number
+  maxZoom?: number
 }
 
 // type is used to calculate the initial boundingBox using height and width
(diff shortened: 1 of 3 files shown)


LeandraH (approved):
Looks reasonable though I couldn't reproduce the original issue :(

f1sh1918:
> Looks reasonable though I couldn't reproduce the original issue :(

check the video in the jira ticket. At least i could reproduce that

sarahsporck (approved):
Tested in Safari. Is fixed now. Good work :)
@LeandraH you can test this by setting `api-url` to https://cms.integreat-app.de in local storage.
Or simply run `localStorage.setItem('api-client', 'https://cms.integreat-app.de')` in the web console.

PR #911, opened 2023-04-13
IGAPP-1256: Remove circle icons

sarahsporck:

In native I removed the SimpleCategoryType, as imo it only introduced more complications. Instead, I directly passed the category to the corresponding components. Also, the CategoryListItem is not abused for search anymore.


native/src/components/SearchListItem.tsx
```diff
@@ -0,0 +1,102 @@
...
+        fontWeight: 'bold',
+      }}
+    />
+  )
+  return (
+    <FlexStyledLink onPress={() => onItemPress(category)} underlayColor={theme.colors.backgroundAccentColor}>
+      <DirectionContainer language={language}>
+        <SearchEntryContainer>
+          <TitleDirectionContainer language={language}>
+            {!!thumbnail && <CategoryThumbnail source={getCachedThumbnail(category, resourceCache)} />}
```
f1sh1918, line 92: i think we shouldn't reuse the thumbnail here since the margin left should be removed. All items should align left
image
While searching it's getting worser
image

f1sh1918 (approved):
Really nice refactoring!
Just one small margin issue for native in the search item

LeandraH (approved):
Nice! Tested in Firefox and simulated iOS

native/src/components/SearchListItem.tsx
sarahsporck: I fixed this by using only right/left margin as needed. Looks much better now. Thank you for the feedback.

web/src/components/__tests__/CategoryListItem.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+    expect(getAllByRole('img')[0]!).toHaveProperty('src', category.thumbnail)
+
+    expect(getByText(childCategory.title)).toBeTruthy()
+    expect(getByText(childCategory.title).closest('a')).toHaveProperty('href', `http://localhost${childCategory.path}`)
+    expect(getAllByRole('img')[1]!).toHaveProperty('src', childCategory.thumbnail)
+
+    const regex = /.+/
+    const texts = queryAllByText(regex)
+    // Only category.title and childCategory.title, nothing split up because of highlighting
+    expect(texts).toHaveLength(2)
```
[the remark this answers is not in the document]
sarahsporck: Ah. this is still here, because originally the CategoryListItem was also a SearchListItem. I'll adjust this test.

PR #913, opened 2023-04-19
IGAPP-1279: Fix Symmetry

SteffiStuffel:


LeandraH (approved):
Thanks for the quick fix, tested in Firefox

sarahsporck (approved):
Tested in Safari ✅

f1sh1918 (commented):
really strange looks like styled components can not use this `screen and` syntax which is in general valid css syntax. Please also remove the other `screen and` occurrences which also lead to errors. thx for the quick fixing

f1sh1918 (approved):

f1sh1918 (commented):
@SteffiStuffel oh dear i found the issue
`dimension.ts`

`minMaxWidth: 'screen and (min-width: 1101px)',` -> we should not add here `screen and` because we may want to use a different media type for some cases with this dimension and its error prone if it differs from other dimensions.
After removing that the valid css syntax is working and you can revert you change

LeandraH:
Good catch, thanks for fixing!

PR #920, opened 2023-05-06
IGAPP-1273: Prevent sentry errors from failing the whole pipeline

steffenkleinle:


.circleci/src/jobs/sentry_android.yml
```diff
@@ -0,0 +1,21 @@
...
+shell: /bin/bash -eo pipefail
+steps:
+  - skip_job
+  - checkout
+  - prepare_workspace
+  - restore_environment_variables
+  # A react-native app reports its version as: application_id@version+build_number
+  - run:
+      name: 'Sentry Upload: integreat'
+      command: tools/sentry-release tuerantuer.app.integreat "${NEW_VERSION_NAME}" ~/attached_workspace/sourcemaps/integreat --version-code "${NEW_VERSION_CODE}"
```
steffenkleinle, line 14: I hardcoded all the build config names, bundle ids and application ids, not 100% convinced of it but I think it makes sense. Otherwise we always have to restore the whole yarn cache to evaluate it from the build config and I thought that is not worth it.

steffenkleinle:
Not tested.

The only way I saw to make sentry not fail the complete pipeline/block the other jobs was to move it to a separate job no other job is depend on. I also moved the sentry uploads of the sourcemaps of the different build configs of ios and android to one job because of job overhead/configuration lines in the workflow yml files. Let me know if you would do it differently.

sarahsporck (approved):
I think this chang will make handling setry outages much easier. 
However, the build will still fail, if e.g. sentry is not reachable, right? 
Not really what the issue description refers to :). Did you think of a follow up PR? :)

f1sh1918 (approved):

PR #921, opened 2023-05-08
IGAPP-1284: Fix transparent language dropdown

steffenkleinle:


web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -35,6 +34,7 @@ export const Container = styled.div`
     object-fit: contain;
     width: 100%;
     height: 100%;
+    opacity: 0.85;
```
f1sh1918, line 37: can you give me a hint why this is needed?

f1sh1918 (approved):
transparency is fixed. thx for that. couldn't find the need for this opacity you set

sarahsporck (approved):
And here I was and thought this was intended ^^
Tested in Firefox.

web/src/components/HeaderActionItemDropDown.tsx
steffenkleinle: just to make the language change icon look more grey and less black I guess, should be fixed in https://issues.tuerantuer.org/browse/IGAPP-1287

PR #922, opened 2023-05-09
IGAPP-1289: Use correct cursor

sarahsporck:


f1sh1918 (approved):
Well it works but i even think this is not needed, since in small viewports (smartphones) there is no cursor.

steffenkleinle (approved):
Not tested

PR #923, opened 2023-05-10
IGAPP-1089: Set title and WebSite jsonLd statically

steffenkleinle:


web/src/index.ejs
```diff
@@ -3,13 +3,21 @@
...
+            "url" : "https://<%= config.hostName %>"
+        }
+    </script>
 
     <!-- Nginx Server Side Include template for dynamic social media previews -->
     <!--# if expr="$render_title = yes" -->
     <!--# include virtual="/proxy/socialmeta/$request_uri" -->
     <!--# else -->
     <!-- Default tags for a prettier social media preview, will be overwritten in Helmet.tsx for most pages. -->
-    <title><%= config.appName %></title>
```
steffenkleinle, line 12: @svenseeberg what is the background behind this whole if/else comments and so on? Is it okay to move the `<title>` tag out of it?

steffenkleinle:
My guess why your solution @sarahsporck did not work: Apparently google does not wait until the loading of the js part is finished, therefore its also never seeing the dynamic json ld script (perhaps the url is always redirected to `/landing/<language>`?). Maybe by making the json ld and the title statically, google might pick it up. Just a vague guess though but probably worth a try.

LeandraH (approved):
Worth a try 🤞

web/src/index.ejs
sarahsporck: For [certain user agents](https://git.tuerantuer.org/DF/salt/src/commit/2418c8a97ffd416ddf0ce2d6dfb38bb97b2976e7/states/webapp/files/nginx.conf#L14) our server renders the [content of the page](https://git.tuerantuer.org/Integreat/social-media-meta-tags/src/branch/master/meta.php)
To test this you can manually set your User Agent to e.g. "Twitterbot 1" and test this. Although I see that it is currently not working properly -.-
Therefore having both jsonld and the title could break this behavior. Not sure though.

steffenkleinle:
Closed in favor of #937

PR #925, opened 2023-05-10
IGAPP-1265: Use webview for all remote html content

steffenkleinle:


steffenkleinle:
This PR removes the NativeHtml (using react-native-render-html) component in favor of the Page/RemoteContent component (using react-native-webview). This removes a lot of additional complexity, e.g. having to maintain two ways to render and style html and allows everything being used in leaf categories to be used in categories with children, poi descriptions, ... For example svgs from uris did not really work with react-native-render-html.

The downside, however, is decreased loading speed. In my opinion the app is still snappy enough though and if this would be to slow, it would even more so to slow on leaf categories where we were using the webview all the time before.

Please test the PR (for real performance testing preferably in a production build `yarn android:integreat-test-cms:production`) and make up your own opinion though.

TODO:
- [ ] Remove react-native-render-html if approved

native/src/components/RemoteContent.tsx
```diff
@@ -26,22 +25,22 @@ export const renderWebviewError = (
...
-  cacheDirectory: ParsedCacheDictionaryType
+  cacheDictionary: ParsedCacheDictionaryType
   language: string
   resourceCacheUrl: string
   onLinkPress: (url: string) => void
   onLoad: () => void
+  webViewWidth: number
 }
 
+// If the app crashes without an error message while using RemoteContent, consider wrapping it in a ScrollView or setting a manual height
```
LeandraH, line 36: Is that a WIP comment or is it supposed to stay in? Hasn't crashed in my testing so far, btw.

native/src/components/Page.tsx
```diff
@@ -1,80 +1,105 @@
...
-  title: string
+  title?: string
   content: string
-  navigateToFeedback?: (positive: boolean) => void
-  files: PageResourceCacheStateType
-  children?: React.ReactNode
+  BeforeContent?: ReactNode
+  AfterContent?: ReactNode
   language: string
-  resourceCacheUrl: string
```
LeandraH, line 29: Nice!

native/src/routes/TuNews.tsx
```diff
@@ -17,7 +17,7 @@ type TuNewsProps = {
   navigateToNews: (newsId: string) => void
 }
 
-const TuNewsNews = ({ route, navigation, data, navigateToNews }: TuNewsProps): ReactElement => {
```
LeandraH, line 20: 🤪

LeandraH (approved):
Very cool! Testing on my iPhone, I didn't notice any difference 💪

native/src/components/RemoteContent.tsx
steffenkleinle: Its more for new usages of RemoteContent, I think I wrapped all current usages. Took me quite a while to figure out.

f1sh1918:
I think we should discuss that in a bigger round. This is a compatibility vs performance question.
Since Native components are ~5 times faster than webview we should consider users with older devices or pages with lots of cpu & gpu extensive content.
For devs this approach is way better to handle and probably has a higher compatibility with other html components if we have to support more in the future.

sarahsporck (approved):
Tested in simulated ios. I didn't feel any performance degression. So I think this proposal is a good change 👍🏼

steffenkleinle:
As discussed on the conference, we'll do this. Please review again and especially test this carefully. Thanks :)

SteffiStuffel (approved):
Please adjust the Jira Ticket, so Martin knows what to test.

Tested on Android emulator, looks good. Really nice work :tada:

SteffiStuffel:
Maybe also renaming the branch would be nice

LeandraH (approved):
Re-tested on my iPhone, still feels pretty snappy :)

PR #926, opened 2023-05-11
IGAPP-XXX: Fix for XCode

LeandraH:
This pull request fixes an issue where some XCode versions throw the error "'value' is unavailable: introduced in iOS 12.0" when building. More here: https://github.com/facebook/react-native/issues/34106

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -75,7 +75,7 @@
     "query-string": "^7.1.1",
     "react": "18.2.0",
     "react-i18next": "^11.18.0",
-    "react-native": "0.70.6",
+    "react-native": "0.70.9",
     "react-native-blob-util": "^0.17.1",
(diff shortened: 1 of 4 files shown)


native/package.json
```diff
@@ -75,7 +75,7 @@
     "query-string": "^7.1.1",
     "react": "18.2.0",
     "react-i18next": "^11.18.0",
-    "react-native": "0.70.6",
+    "react-native": "0.70.9",
```
steffenkleinle, line 78: Any specific reason why we upgrade react native? Normally this should just be done in a real issue. Did you check possibly necessary manual upgrading steps?

steffenkleinle (approved):

f1sh1918 (approved):
Tested on ios

PR #927, opened 2023-05-15
IGAPP-1293: Fix back button flickering

steffenkleinle:

### DIFF
--- a/native/src/Navigator.tsx
+++ b/native/src/Navigator.tsx
@@ -72,9 +72,9 @@ type HeaderProps = {
 
 const transparentHeader = (headerProps: StackHeaderProps) => <TransparentHeader {...(headerProps as HeaderProps)} />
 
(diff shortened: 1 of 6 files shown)


LeandraH (approved):
Looks good, I couldn't reproduce the original issue in emulated iOS or Android but it also didn't break anything for me 😄

steffenkleinle:
> I even couldn't reproduce the issue neither on android nor ios. For me using this works as expected:
> 
> ```js
>   const previousRoute = navigation.getState().routes[navigation.getState().routes.length - 2]
>   const canGoBack = navigation.canGoBack()   
> ```

See attached video in https://issues.tuerantuer.org/browse/IGAPP-1283, you can clearly see the header text changing there. Same for back button on news, it quickly disappears bc the `useHeader` call with `isHome={false}` only kicks in after data is loaded.

f1sh1918 (approved):

PR #928, opened 2023-05-15
IGAPP-1231: Offline cache invalidation and migration routine

steffenkleinle:


native/src/components/I18nProvider.tsx
```diff
@@ -46,7 +46,6 @@ const I18nProvider = ({ children }: I18nProviderProps): ReactElement | null => {
           /* Escaping is not needed for react apps:
              https://github.com/i18next/react-i18next/issues/277 */
         },
-        debug: buildConfig().featureFlags.developerFriendly,
```
steffenkleinle, line 49: Just disables logging the whole i18n object every app start. If you would like to keep it let me know :) Just annoyed me during debugging this bc I did not see the logs.

native/src/routes/CategoriesContainer.tsx
```diff
@@ -71,16 +69,8 @@ const CategoriesContainer = ({ navigation, route }: CategoriesContainerProps): R
   const error =
     data?.categories && !category && previousLanguageCode === languageCode ? ErrorCode.PageNotFound : response.error
 
-  // Workaround clear cache on refresh if city content can't be loaded.
-  // TODO IGAPP-1231: Proper cache invalidation for version updates
-  const clearResourcesAndCache = useCallback(() => {
-    dataContainer.clearInMemoryCache()
-    dataContainer.clearOfflineCache().catch(reportError)
```
steffenkleinle, line 78: This should not be called without good reason, at least not during every refresh.

native/src/components/I18nProvider.tsx
LeandraH: Good call

LeandraH (approved):
I can't reproduce the original issue anymore either on main with the workaround commented out or on this branch, so yay 😭

steffenkleinle:
> I can't reproduce the original issue anymore either on main with the workaround commented out or on this branch, so yay sob

Yeah I guess for the original issue you would need some outdated offline cache thats long gone by now.

f1sh1918:
> > I can't reproduce the original issue anymore either on main with the workaround commented out or on this branch, so yay sob
> 
> Yeah I guess for the original issue you would need some outdated offline cache thats long gone by now.

You can just use this revision https://github.com/digitalfabrik/integreat-app/commit/d84703890f50229db9709536ce94767d6e8a1183

native/src/components/I18nProvider.tsx
sarahsporck: like it as this also makes the tests much cleaner 😄

native/src/routes/CategoriesContainer.tsx
steffenkleinle: No, so far it was only called for switching between the cms after tapping landing 10 times to make sure we use the correct content/resources. In the landing page this is still kinda fine but in a city content page it breaks stuff bc it also deletes the metadata like last updates/last usages.

PR #929, opened 2023-05-15
IGAPP-1291: Disable chatbot for munich

steffenkleinle:


LeandraH (approved):

f1sh1918 (approved):
Looks fine. Not tested
Should we keep the `<ChatBotWidget>` itself since its not used anymore?

steffenkleinle:
> Looks fine. Not tested Should we keep the `<ChatBotWidget>` itself since its not used anymore?

AFAIK there are plans to reintroduce the chatbot with different technology so for now I would keep it.

PR #930, opened 2023-05-15
IGAPP-1286: Fix text alignment of local news

steffenkleinle:


LeandraH (commented):
It hasn't been fixed in this one single push notification in Arabic in Testumgebung, but maybe that's because of the Latin letters in there?

image

steffenkleinle:
> > It hasn't been fixed in this one single push notification in Arabic in Testumgebung, but maybe that's because of the Latin letters in there?
> Hmmm, in theory LTR and RTL chars should be counted and the direction with more chars wins. So not sure what is happening here.

I just tested it and I can't reproduce it, are you sure you checked out the right branch? For me its right aligned.

LeandraH (approved):
You're right, sorry about that, must have gotten confused about the branches

f1sh1918 (approved):
Works fine on android

PR #932, opened 2023-05-16
IGAPP-1294: Internal links broken ios

f1sh1918:
fix opening internal links (which are handled by opening in the browser) by encoding the url

### DIFF
--- a/native/src/utils/openExternalUrl.ts
+++ b/native/src/utils/openExternalUrl.ts
@@ -9,8 +9,8 @@ import buildConfig from '../constants/buildConfig'
 import sendTrackingSignal from './sendTrackingSignal'
 import { reportError } from './sentry'
 
-const openExternalUrl = async (url: string, showSnackbar: (snackbar: SnackbarType) => void): Promise<void> => {
-  const encodedUrl = encodeURI(url)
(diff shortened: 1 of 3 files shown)


native/src/utils/openExternalUrl.ts
```diff
@@ -10,32 +10,33 @@ import sendTrackingSignal from './sendTrackingSignal'
 import { reportError } from './sentry'
 
 const openExternalUrl = async (url: string, showSnackbar: (snackbar: SnackbarType) => void): Promise<void> => {
-  const { protocol } = new URL(url)
+  const encodedUrl = encodeURI(url)
+  const { protocol } = new URL(encodedUrl)
```
steffenkleinle, line 14: Personal preference: I always like it a little bit more to name variables that should not be used some way that make that clear, for example something like `unsafeUrl` or `rawUrl` or similar... Otherwise its pretty easy to just type `url` and use that further down, while only `encodedUrl` should be used which would introduce bugs again.

steffenkleinle (approved):
Perhaps you could add a test :)

Not tested but looks good

LeandraH (approved):
Nice, tested on emulated iOS

f1sh1918:
> Perhaps you could add a test :)
> 
> Not tested but looks good

you are right. added a test for ugly raw links :)

steffenkleinle (approved):

PR #933, opened 2023-05-17
IGAPP-1296: Fix city not visible after golive

steffenkleinle:


LeandraH (commented):
Hmmm, I just tested it in cms-test by taking Landkreis Amberg-Sulzbach active and that worked, and then taking Aschaffenburg active and that didn't work. Both on simulated iOS.

native/src/hooks/useLoadCities.ts
```diff
@@ -1,24 +1,43 @@
...
+      showSnackbar({ text: fromError(e) })
+    }
+  }
+  return cachedData
 }
 
+const useLoadCities = (params: Params): ReturnType<CityModel[]> =>
+  // Normally using params as dependency triggers infinite re-renders
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useLoadAsync(useCallback(forceUpdate => loadWithCache({ ...params, forceUpdate }), []))
```
sarahsporck, line 41: This should only be true if the params are not stateful. 
When I tested using only `useLoadCities(showSnackbar)` there were no rerenders anymore. Aparrently `{ showSnackbar } creates a new object which is not stateful and therefore triggers those re-renders. Please change this
```suggestion
const useLoadCities = (showSnackbar: (snackbar: SnackbarType) => void): ReturnType<CityModel[]> =>
  useLoadAsync(useCallback(forceUpdate => loadWithCache({ showSnackbar, forceUpdate }), [showSnackbar]))
```

native/src/hooks/useLoadCities.ts
```diff
@@ -1,24 +1,43 @@
...
+type Params = {
+  forceUpdate?: boolean
+  showSnackbar: (snackbar: SnackbarType) => void
+}
+
+const loadWithCache = async ({ showSnackbar, forceUpdate = false }: Params): Promise<CityModel[] | null> => {
+  const cachedData = (await dataContainer.citiesAvailable()) ? await dataContainer.getCities() : null
+
+  if (!forceUpdate && cachedData) {
+    return cachedData
```
sarahsporck, line 18: We never actually invoke this with `forceUpdate=true` anywhere. This means if a user never pulls to refresh in Landing no new content will be loaded? Or does this just work because we invalidate our cache every time we update our app?

sarahsporck (changes requested):

native/src/hooks/useLoadCities.ts
steffenkleinle: Good catch, guess I just copied it haha :)

native/src/hooks/useLoadCities.ts
steffenkleinle: @sarahsporck however, this leads to the showing of a snackbar on the landing page if offline (even if the user did not manually pull to refresh). Any ideas on how to circumvent this? Or is that even okay behavior?

LeandraH:
I'm now getting the new city showing up but an error on actually trying to open it :/

https://github.com/digitalfabrik/integreat-app/assets/18513943/abc200b9-6b9a-4701-b231-7106a1302d9d

LeandraH:
Yup, sorry, my bad. The URL is https://cms-test.integreat-app.de/aschaffenburg/de/wp-json/extensions/v3/pages/ and gives me an error 500. Note to self, don't test on Aschaffenburg 😓

LeandraH (approved):

native/src/hooks/useLoadCities.ts
sarahsporck: I think this is okay. To circumvent we'd have to throw the error and add the showSnackbar in place instead of passing it. Too much work 🙈

steffenkleinle: Agreed :+1:

sarahsporck (approved):
Tested on simulated iphone

PR #939, opened 2023-05-25
IGAPP-1307: Use languages property instead of endpoint

steffenkleinle:


sarahsporck (approved):
Tested in firefox and android emulator. Already feels smoother, but maybe that's me :)

PR #940, opened 2023-05-26
IGAPP-1290: Better test performance

sarahsporck:

I decided against swc/jest as we encountered some issues with the framework in the entitlementcard project. Regardless, swc/jest is still a very young project. Give it a few more zears and it might become a real alternative. 

- updated test dependencies
- disabled typechecking during test runs (we already do typechecking in the ci and locally we use static typechecking I guess) { reduces test runs by around 50% compared to before }
- add missing i18-next mocks


steffenkleinle (approved):

LeandraH (approved):
In a very unscientific test, this sped up the tests by about 15%. Yay!

PR #941, opened 2023-05-26
IGAPP-1302: Fix spacing after external links

sarahsporck:

The width and height in the svg messed with the actual height and with of the icon. Removing this fixed this. 

Additional changes:
- removed css properties with no effect
- normalized svg path to width and height 18x18 to make it more readable (found a [nice tool](https://www.npmjs.com/package/pathfit) for that btw)


LeandraH (approved):
Nice! Tested in emulated iOS and Android.

steffenkleinle (approved):
Looks good but also looked okay to me before :) Tested on android

PR #942, opened 2023-05-30
IGAPP-1217: Cut loading times

steffenkleinle:


steffenkleinle:
This cuts the loading times by using the single city endpoint for city content routes (IGAPP-1267) and executing requests simultaneously (IGAPP-1217).

web/src/CityContentSwitcher.tsx
```diff
@@ -42,27 +44,23 @@ const SearchPage = lazyWithRetry(() => import('./routes/SearchPage'))
...
-const CityContentSwitcher = ({ cities, languageCode }: CityContentSwitcherProps): ReactElement => {
+const CityContentSwitcher = ({ languageCode }: CityContentSwitcherProps): ReactElement => {
   const cityCode = useParams().cityCode!
+  const { data: city, loading } = useLoadFromEndpoint(createCityEndpoint, cmsApiBaseUrl, { city: cityCode })
   const pathname = normalizePath(useLocation().pathname)
   const { viewportSmall } = useWindowDimensions()
-  const cityModel = cities.find(it => it.code === cityCode)
 
-  if (!cityModel) {
+  if (!city && !loading) {
```
steffenkleinle, line 63: This is the important change that renders the routes and therefore loads the other requests without waiting for the loading of the city.

web/src/CityContentSwitcher.tsx
```diff
@@ -94,21 +92,19 @@ const CityContentSwitcher = ({ cities, languageCode }: CityContentSwitcherProps)
...
-  const { eventsEnabled, offersEnabled } = cityModel
-  const localNewsEnabled = buildConfig().featureFlags.newsStream && cityModel.localNewsEnabled
-  const tuNewsEnabled = buildConfig().featureFlags.newsStream && cityModel.tunewsEnabled
-  const poisEnabled = buildConfig().featureFlags.pois && cityModel.poisEnabled
+
+  const eventsEnabled = !city || city.eventsEnabled
+  const offersEnabled = !city || city.offersEnabled
+  const localNewsEnabled = buildConfig().featureFlags.newsStream && (!city || city.localNewsEnabled)
+  const tuNewsEnabled = buildConfig().featureFlags.newsStream && (!city || city.tunewsEnabled)
+  const poisEnabled = buildConfig().featureFlags.pois && (!city || city.poisEnabled)
```
steffenkleinle, line 105: If the city is not available yet nothing is rendered in the routes anyway, therefore render the route until we know if its enabled.

steffenkleinle:
@SteffiStuffel we can probably close the epic once this two PRs are merged right? https://issues.tuerantuer.org/browse/IGAPP-1216 Or anything else left to do?

web/src/CityContentSwitcher.tsx
```diff
@@ -127,9 +123,13 @@ const CityContentSwitcher = ({ cities, languageCode }: CityContentSwitcherProps)
...
-            </CityContentLayout>
+            city ? (
+              <CityContentLayout {...suspenseLayoutProps} route={route} city={city}>
+                <LoadingSpinner />
+              </CityContentLayout>
+            ) : (
+              <Layout />
+            )
           }>
           <Component {...cityRouteProps} />
```
sarahsporck, line 134: so we could also change the `CityRouteProps` to 
```ts
type CityRouteProps {
  city: CityModel
  pathname: string
  cityCode: string
  languageCode: string
}
```
and then we can write sth like: 
```suggestion
          city ? <Component city={city} pathname={pathname} cityCode={cityCode} languageCode={languageCode} /> : null
```
and get rid of all the `if (!city)` in the routes right?

steffenkleinle: Nope, actually we can't, thats part of the whole PR. By rendering the componentns/routes even if the city is not available, we trigger the loading of the routes data (i.e. events, categories, ...). If we only render the component if the city is available, it will be subsequent requests, not simultaneous ones.

LeandraH (approved):
Nice! In five tests of this branch vs main with Lighthouse (on a sketchy train internet connection) I got an improvement of 6%.

api-client/src/index.ts
```diff
@@ -55,6 +55,7 @@ export {
   SHELTER_CONTACT_ENDPOINT_NAME,
 } from './endpoints/createShelterContactEndpoint'
 export { default as createCitiesEndpoint, CITIES_ENDPOINT_NAME } from './endpoints/createCitiesEndpoint'
+export { default as createCityEndpoint, CITY_ENDPOINT_NAME } from './endpoints/createCityEndpoint'
```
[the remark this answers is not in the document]
steffenkleinle: Hmmm not that sure actually, just historically grown. Since all the things in the api-client should only be used via the exports from index.ts, it should not matter (so also changing it if there is a reason to change it would be okay).

web/src/index.ejs
```diff
@@ -57,8 +57,6 @@
     <link rel="dns-prefetch" href="<%= config.cmsUrl %>" />
     <link rel="preconnect" href="<%= config.cmsUrl %>" />
 
-    <link rel="preload" href="<%= config.cmsUrl %>/wp-json/extensions/v3/sites" as="fetch" crossorigin />
```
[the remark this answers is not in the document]
steffenkleinle: The sites endpoint is only loaded for the landing page anymore, so I figured it does not make sense to always preload it.

web/src/routes/CategoriesPage.tsx
```diff
@@ -87,13 +87,17 @@ const CategoriesPage = ({ cityModel, pathname, languages, cityCode, languageCode
   }, [cityCode, languageCode, pathname, categoryId])
   const { data: parents, loading: parentsLoading, error: parentsError } = useLoadAsync(requestParents)
 
+  if (!city) {
```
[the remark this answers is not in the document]
steffenkleinle: I don't think so, no. If `parentsLoading` is true, we should just display a loading spinner. If `city` is null, the loading logic is handled in `CityContentSwitcher`. I admit, this is all a little confusing, but it was the only way I could think of to solve the problem of executing the requests simultaneously.

SteffiStuffel:
> Nice! In five tests of this branch vs main with Lighthouse (on a sketchy train internet connection) I got an improvement of 6%.

hmh, i do not get any improvement.
On this branch loading time of city of Augsburg is: 769ms
on main it is: 715ms
Details: WLAN, disable all caching, reload page for several times and calculate the average.

SteffiStuffel:
> @SteffiStuffel we can probably close the epic once this two PRs are merged right? https://issues.tuerantuer.org/browse/IGAPP-1216 Or anything else left to do?

Sure we can close it, when we are happy with the loading times for now.

steffenkleinle:
> > Nice! In five tests of this branch vs main with Lighthouse (on a sketchy train internet connection) I got an improvement of 6%.
> 
> hmh, i do not get any improvement. On this branch loading time of city of Augsburg is: 769ms on main it is: 715ms Details: WLAN, disable all caching, reload page for several times and calculate the average.

TBH I think its not really possible to measure the effect in localhost. I tried both lighthouse and @SteffiStuffel's method to look at the loading times in firefox and it seems to be super dependent on the time executed and it fluctuates quite some even with throttling enabled. Without merging main into my feature branch, I could on average verify @SteffiStuffel's result with an average of 662ms for main and 690ms for testumgebung/de, however, with an updated feature branch it was suddenly 687ms on main and 650ms on the feature branch...

What I can see, however, is that on the feature branch the second request (the children endpoint) is executed always around 1.1s while on the main branch its only around 1.5s. In (my simple) theory my feature branch should therefore be 400ms faster, not sure why it is not showing.

Anyway, my conclusion would be to try this out, look at the times in google search console and if it gets worse/doesn't show an improvement we could revert it. Except @sarahsporck or @f1sh1918 have some other ideas.

sarahsporck (commented):
I'm fine with testing it out and see the result. While testing whether this pr actually reduces simultanieous requests I found that these changes lead to more requests allover than the current main branch.
At least for the user story I tested.

For the same actions (navigating to a city, navigating to a leaf category, changing location) this pr makes more requests compared to the main branch. This makes me unsure whether this actually improves loading times. Any opinions?
this pr:
Bildschirmfoto 2023-06-22 um 10 13 50
main:
Bildschirmfoto 2023-06-22 um 10 18 23

steffenkleinle:
> I'm fine with testing it out and see the result. While testing whether this pr actually reduces simultanieous requests I found that these changes lead to more requests allover than the current main branch. At least for the user story I tested.
> 
> For the same actions (navigating to a city, navigating to a leaf category, changing location) this pr makes more requests compared to the main branch. This makes me unsure whether this actually improves loading times. Any opinions?

The main reason why we want to cut the loading times is SEO and that google is rating our loading times as too big. This would not be affected by the use case you tested as google is always just indexing single pages without changing between pages/cities and inside one city, the amount of requests stays exactly the same.

The additional requests in your screenshot are:
- lkaugsburg in the start (not an additional request, on main the sites endpoint is loaded instead which is not on your screenshot)
- sites after going to city change
- aichach-friedberg after selecting that city

Therefore, the only additional requests are related to changing a city which is, at least in my opinion, not a normal or even frequent use case for normal users (except selecting the wrong city) or SEO.

Conclusion: In my opinion, we should merge this if we expect or hope for a loading time improvement while staying inside one city and should ignore the use case of changing cities.

sarahsporck (approved):
I guess I'm happy with you're explanation. Then let's go and see if there is an improvement :)

PR #945, opened 2023-06-05
IGAPP-122: Send user to landing page if current city becomes inactive

steffenkleinle:


native/src/Navigator.tsx
```diff
@@ -112,63 +113,61 @@ const Navigator = (): ReactElement | null => {
...
+      initSentry()
+    }
+
+    if (!storageVersion) {
+      appSettings.setVersion(ASYNC_STORAGE_VERSION).catch(reportError)
+    }
+
+    if (storageVersion !== ASYNC_STORAGE_VERSION) {
+      // start a migration routine
+    }
```
LeandraH, line 137: Is this a todo for the future? Then we can maybe add a `TODO` to make it easier to find? Or can it be removed?

native/src/Navigator.tsx
```diff
@@ -112,63 +113,61 @@ const Navigator = (): ReactElement | null => {
...
+      return
+    }
+    if (buildConfig().featureFlags.introSlides && !settings.introShown) {
+      setInitialRoute({ name: INTRO_ROUTE })
+    } else if (!cityCode) {
+      setInitialRoute({ name: LANDING_ROUTE })
+    } else if (cities?.find(it => it.code === cityCode)) {
+      setInitialRoute({ name: CATEGORIES_ROUTE })
+    } else if (cities) {
+      // City is not available anymore
```
LeandraH, line 155: Thanks, I think that would have been really hard to understand :)

LeandraH (commented):
It looks reasonable but I can't get the functionality to work on the emulated iPhone I'm using. I archived Aichach-Friedberg and I can still change the language in there. I also exited the app on the phone and re-opened, I pulled down to refresh, I even restarted it from XCode, nothing has sent me back to the selection screen so far :(

native/src/Navigator.tsx
steffenkleinle: Its not a todo, more like an instruction if our async storage/app settings schema changes at some point where to put migration routines. But yeah, I think we can remove it.

steffenkleinle:
> It looks reasonable but I can't get the functionality to work on the emulated iPhone I'm using. I archived Aichach-Friedberg and I can still change the language in there. I also exited the app on the phone and re-opened, I pulled down to refresh, I even restarted it from XCode, nothing has sent me back to the selection screen so far :(

To test this, you can just add a `const cities = data.filter(it => it.code !== 'testumgebung')` in the navigator which just simulates the cms not delivering the city with the api anymore.

native/src/__tests__/Navigator.spec.tsx
```diff
@@ -129,39 +139,50 @@ describe('Navigator', () => {
...
+    mockSettings({ introShown: true })
     mocked(quitAppStatePushNotificationListener).mockImplementation(async navigate =>
       navigate('https://integreat.app/augsbug/de/news/local/1234')
     )
-    await appSettings.setIntroShown()
-    const { findByText } = renderNavigator()
+    const { getByText } = renderNavigator()
 
-    await findByText('Redirect')
+    waitForExpect(() => expect(getByText('Redirect')).toBeTruthy())
```
sarahsporck, line 185: findByText and waitForExpect should do the same: https://github.com/testing-library/dom-testing-library/blob/d09b3c25b07486c17603317f4c5a0320c188cbed/src/query-helpers.ts#LL120C1-L120C1 If this is making a difference it's because you are missing thi await

sarahsporck (approved):
Tested on Android emulator.

f1sh1918 (approved):
Tested on iOS. Works fine :)

steffenkleinle:
I will recheck this once I am back so pls dont merge :)

native/src/__tests__/Navigator.spec.tsx
steffenkleinle: Thanks for the hint :+1:

steffenkleinle:
> It looks reasonable but I can't get the functionality to work on the emulated iPhone I'm using. I archived Aichach-Friedberg and I can still change the language in there. I also exited the app on the phone and re-opened, I pulled down to refresh, I even restarted it from XCode, nothing has sent me back to the selection screen so far :(

I could reproduce this (partly). If you don't close the app, the hook in Navigator will never be called and the user is therefore not sent back to the city selection. Instead, I now improved the error displayed if the city does not exist anymore and added the possibility to manually navigate to the city selection (since clearing the city code while for example the categories screen is still opened leads to errors being thrown). On the next app start, the city should automatically be cleared now together with a shown snackbar (if the user did not manually select a new city beforehand).

LeandraH (approved):
Awesome! Got the two described error messages!

PR #947, opened 2023-06-12
IGAPP-1308: Remove ios image caching

f1sh1918:
I removed the cache control `reload` since its not needed in our case and was used to fix a bug with local images on ios which were not working and was already fixed. So we can fallback to default caching strategy.
You can find more information here
https://reactnative.dev/docs/images#cache-control-ios-only

### DIFF
--- a/release-notes/unreleased/IGAPP-1308-remove-ios-image-caching.yml
+++ b/release-notes/unreleased/IGAPP-1308-remove-ios-image-caching.yml
@@ -1,6 +1,0 @@
-issue_key: IGAPP-1308
-show_in_stores: false
-platforms:
-  - ios
-en: Remove ios image caching
-de: Entferne Bildercaching für ios.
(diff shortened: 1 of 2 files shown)


LeandraH (commented):
Looks good, tested on a real iPhone :)

steffenkleinle (approved):
Not tested

release-notes/unreleased/IGAPP-1308-remove-ios-image-caching.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: IGAPP-1308
```
steffenkleinle, line 1: Not sure if release notes for issues like this makes sense. There is not really anything to test for martin, nothing interesting for users and also not relevant for us devs (since the ticket/PR/commit message give enough info). Our guideline was so far to just add release notes if relevant for users, testers or the other teams.

PR #949, opened 2023-06-13
IGAPP-1130: Native Event Export

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/IGAPP-1130 .

I used an [external library](https://github.com/wmcmahan/react-native-calendar-events) in the end, and opened a snackbar with a button to the calendar on success.

It doesn't handle recurring events yet because we haven't made a decision about how to export those yet.


native/src/routes/Events.tsx
```diff
@@ -81,8 +92,68 @@ const Events = ({
...
+                },
+              },
+            })
+          })
+          .catch(() => {
+            showSnackbar({ text: 'generalError' })
+          })
+      }
+
+      const handlePress = async (): Promise<void> => {
```
f1sh1918, line 127: please find an proper naming like `checkCalendars` or sth suitable

native/src/routes/Events.tsx
```diff
@@ -81,8 +92,68 @@ const Events = ({
...
+        }
+      }
       return (
         <LayoutedScrollView refreshControl={<RefreshControl onRefresh={refresh} refreshing={false} />}>
+          <CalendarChoice
+            closeOverlay={() => setShowCalendarChoiceOverlay(false)}
+            overlayVisible={showCalendarChoiceOverlay}
+            chooseCalendar={(id: string) => {
+              setShowCalendarChoiceOverlay(false)
+              saveEventToCalendar(id)
```
f1sh1918, line 153: promise return is missing

native/src/routes/Events.tsx
```diff
@@ -81,8 +92,68 @@ const Events = ({
...
+        } else if (editableCalendars.length === 1) {
+          saveEventToCalendar(editableCalendars[0]?.id)
+        } else {
+          setCalendars(editableCalendars)
+          setShowCalendarChoiceOverlay(true)
+        }
+      }
       return (
         <LayoutedScrollView refreshControl={<RefreshControl onRefresh={refresh} refreshing={false} />}>
+          <CalendarChoice
```
f1sh1918, line 148: i think you should check here if `calendars.length > 1` and render it conditionally
then you can make calendars non optional in the `CalendarChoiceProps`

native/src/components/CalendarChoice.tsx
```diff
@@ -0,0 +1,58 @@
...
+  align-items: center;
+  margin: 5px 20px;
+`
+
+const ButtonText = styled.Text`
+  font-family: ${props => props.theme.fonts.native.decorativeFontRegular};
+`
+
+type CalendarChoiceProps = {
+  calendars?: Calendar[]
```
f1sh1918, line 31: shouldn't be optional.

native/src/routes/Events.tsx
```diff
@@ -81,8 +92,68 @@ const Events = ({
...
+        if (permission === 'undetermined') {
+          permission = await RNCalendarEvents.requestPermissions()
+        }
+        if (permission === 'denied' || permission === 'restricted') {
+          showSnackbar({ text: 'noCalendarPermission' })
+          return
+        }
+        const editableCalendars = (await RNCalendarEvents.findCalendars()).filter(cal => cal.allowsModifications)
+        if (editableCalendars.length === 0) {
+          showSnackbar({ text: 'generalError' })
```
f1sh1918, line 138: i think you should throw an error message like. "No calendars available", not just a general error

f1sh1918 (commented):
Tested on android and ios. Works fine. :)
I think in the CalendarChoice should have a back button. especially for ios (i couldn't test the modal there because i don't have multiple calendars on simulator)

If the calendar permission is denied once it should be possible to re-request the permission somehow imo. Maybe show next to the error a button to re-request it.

f1sh1918 (commented):
One small thing. The "schließen" Button is not very prominent. Why not using arrow back in the left corner with an modal header like in the feedback modal to have the same look and feel

sarahsporck (approved):
Taseted on Android and simulated ios. Nice result 👍🏼

native/src/components/CalendarChoice.tsx
```diff
@@ -0,0 +1,71 @@
...
+  closeOverlay,
+}: CalendarChoiceProps): ReactElement => {
+  const { t } = useTranslation('events')
+  return (
+    <Modal visible={overlayVisible} onRequestClose={closeOverlay} animationType='fade'>
+      <StyledView>
+        <TitleContainer>
+          <HeaderBackButton onPress={closeOverlay} label={t('back')} />
+          <Title>{t('chooseCalendar')}</Title>
+        </TitleContainer>
```
steffenkleinle, line 60: I think we should use the same header style (same buttons, shadows, header text style...) as for our other headers, the following would work if you export the components from `TransparentHeader`:
```suggestion
      <BoxShadow>
        <Horizontal>
          <HorizontalLeft>
            <HeaderBackButton onPress={closeOverlay} labelVisible={false} tintColor={theme.colors.textColor} />
          </HorizontalLeft>
        </Horizontal>
      </BoxShadow>
      <StyledView>
        <Title>{t('chooseCalendar')}</Title>
```
However, since we use and define the same elements two times already (Header, TransparentHeader), it might also make sense to create a new component `HeaderBox`, which takes a `HeaderLeft` component and optional text and header buttons as props which can then be used in TransparentHeader, Header and here.

steffenkleinle (changes requested):
Three things I realized:
* The `Add to Calendar` button flickers quickly bc its rendered before the page content is rendered. Using `AfterContent` instead of `pageFooter` should fix that, you just have to apply additional styling to avoid the button spanning the whole width.
* The text of the system status bar on android is white and therefore not readable, see screenshot.
* The header of the modal is not stuck to the top, see screenshot.


I am not 100% convinced that we should use a modal for this (we could also use a new route like for ChangeLanguageModal) but it seems to be the easiest solution for now. However, it should correspond to the layout/design of our other screens, mainly the header.

I tested this already and afaics it works great, nicely done!

native/src/components/CalendarChoice.tsx
```diff
@@ -53,17 +48,22 @@ const CalendarChoice = ({
...
-        ))}
-      </StyledView>
+      <Container>
+        <Header>
+          <HeaderBox goBack={closeOverlay} text={t('chooseCalendar')} textCentered />
+        </Header>
+        <ButtonContainer
+          contentContainerStyle={{
+            flex: 1,
+            justifyContent: 'center',
```
steffenkleinle, line 58: Can't we use styled components for this as well? Best keep it consistent

native/src/components/CalendarChoice.tsx
```diff
@@ -1,28 +1,23 @@
-import { HeaderBackButton } from '@react-navigation/elements'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Modal, TouchableOpacity } from 'react-native'
 import { Calendar } from 'react-native-calendar-events'
 import styled from 'styled-components/native'
 
-const StyledView = styled.View`
+import dimensions from '../constants/dimensions'
```
steffenkleinle, line 7: Sorry for not mentioning this earlier but I think we should somehow make clear that this is a full screen modal/overlay and therefore name it `CalendarChoiceModal` or `CalendarChoiceOverlay`. Since react-native seems to use the term modal and we do in lunes as well, I would prefer the first choice. I would then also be consistent and avoid the term overlay (e.g. in ExportEventButton).

native/src/components/CalendarChoice.tsx
```diff
@@ -53,17 +48,22 @@ const CalendarChoice = ({
...
-        </TitleContainer>
-        {calendars.map(cal => (
-          <CalendarButton key={cal.id} onPress={() => chooseCalendar(cal.id)} color={cal.color}>
-            <ButtonText>{cal.title}</ButtonText>
-          </CalendarButton>
-        ))}
-      </StyledView>
+      <Container>
+        <Header>
+          <HeaderBox goBack={closeOverlay} text={t('chooseCalendar')} textCentered />
```
steffenkleinle, line 53: As mentioned before I think we should stay consistent with the header text and not put some user action request there when normally it always shows the title of the previous route. Therefore please move the text to a separate component directly above the calendar list.

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const saveEventToCalendar = async (calendarId: string | undefined): Promise<void> => {
+    await RNCalendarEvents.saveEvent(event.title, {
+      startDate: event.date.startDate.toISOString(),
+      endDate: event.date.endDate.toISOString(),
+      allDay: event.date.allDay,
+      location: event.location?.fullAddress,
+      description: event.excerpt,
+      calendarId,
+    })
+      .then(id => {
```
steffenkleinle, line 36: Please stay consistent and just use async/await

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+          positiveAction: {
+            label: t('goToCalendar'),
+            onPress: () => {
+              if (Platform.OS === 'ios') {
+                const appleRefDate = moment('Jan 1 2001', 'MMM DD YYYY')
+                const secondsSinceRefDate = event.date.startDate.diff(appleRefDate, 'seconds')
+                Linking.openURL(`calshow:${secondsSinceRefDate}`)
+              } else if (Platform.OS === 'android') {
+                Linking.openURL(`content://com.android.calendar/events/${id}`)
+              }
```
steffenkleinle, line 49: Please extract to a separate method and perhaps also add some comment or reference about how this is handled on iOS.

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+          chooseCalendar={async (id: string) => {
+            setShowCalendarChoiceOverlay(false)
+            await saveEventToCalendar(id).catch(() => {
+              showSnackbar({ text: 'generalError' })
+            })
+          }}
+          calendars={calendars}
+        />
+      )}
+      <Button
```
steffenkleinle, line 100: Optional since this was used in a few places before already: We should probably create a custom `TextButton` component at some time since we use the same code/styling in quite a few places. Margin is the only property specific to the use here which could be applied with styled-components as well.

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  return (
+    <>
+      {!!calendars?.length && (
+        <CalendarChoice
+          closeOverlay={() => setShowCalendarChoiceOverlay(false)}
+          overlayVisible={showCalendarChoiceOverlay}
+          chooseCalendar={async (id: string) => {
```
steffenkleinle, line 91: Please extract function for readability

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+      .catch(() => {
+        showSnackbar({ text: 'generalError' })
+      })
+  }
+
+  const checkCalendarsAndSaveEvent = async (): Promise<void> => {
+    let permission = await RNCalendarEvents.checkPermissions()
+    if (permission === 'undetermined') {
+      permission = await RNCalendarEvents.requestPermissions()
+    }
```
steffenkleinle, line 63: Optional: Normally we don't really like `let`s:
```suggestion
    const checkedPermission = await RNCalendarEvents.checkPermissions()
    const permission = checkedPermission === 'undetermined' ? await RNCalendarEvents.requestPermissions() : checkedPermission
```

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
+  const { t } = useTranslation('events')
+  const theme = useTheme()
+  const showSnackbar = useSnackbar()
+
+  const [eventExported, setEventExported] = useState<boolean>(false)
+  const [showCalendarChoiceOverlay, setShowCalendarChoiceOverlay] = useState<boolean>(false)
+  const [calendars, setCalendars] = useState<Calendar[]>()
+
+  const saveEventToCalendar = async (calendarId: string | undefined): Promise<void> => {
```
steffenkleinle, line 27: We should stick to either save or export I think:
```suggestion
  const exportEvent = async (calendarId: string | undefined): Promise<void> => {
```

native/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+            },
+          },
+        })
+      })
+      .catch(() => {
+        showSnackbar({ text: 'generalError' })
+      })
+  }
+
+  const checkCalendarsAndSaveEvent = async (): Promise<void> => {
```
steffenkleinle, line 59:
```suggestion
  const checkCalendarsAndExportEvent = async (): Promise<void> => {
```

steffenkleinle (changes requested):
I like how smooth this all works together! Now I find integreat events all the time in my calendars though after testing haha :D

Sadly the text in the status bar on android is white and therefore not readable. Not sure why this happens.

native/src/components/CalendarChoice.tsx
LeandraH: Apparently, we can! Just need the `attrs` in there, thanks for the hint!

LeandraH:
> Sadly the text in the status bar on android is white and therefore not readable. Not sure why this happens.

Sorry, missed that earlier! I can't reproduce that, are we talking about the "Choose a calendar" one? Or is that another one? Which device to you use?
image

native/src/components/ExportEventButton.tsx
LeandraH: Great idea, I'll make a ticket for that!

steffenkleinle (commented):
Seems like that the status bar prop `barStyle` just gets ignored if you render a modal (at least on my device). I have a custom rom installed so in theory that could also be the problem. I checked both light and dark mode and same issue either way.
Maybe someone could test this with a real android device as well?

Another note:
In landscape mode it doesn't look very good :see_no_evil: The react native docs also say that as default on ios only portrait works.
Screenshot_20230703-131048_IntegreatTestCms

steffenkleinle (approved):
Not tested again. Would be cool if someone else could test on another real android device. If not reproducible, feel free to merge.

LeandraH:
We figured out that it is an issue in some Android devices that the StatusBar gets a light content when a Modal is opened (https://github.com/facebook/react-native/issues/34350). There might be a solution in RN 0.71, will do that update and then check again.

PR #951, opened 2023-06-15
IGAPP-1315: Adjust toolbar

f1sh1918:
Please check:
- the design where the toolbar should be shown and which items
- check that the items are working

Note: 
- Feedback adjustments and social media sharing will be done in different tasks
- Toolbar is missing in designs for local news. for tunews list (infinite scrolling) there is no toolbar.


web/src/components/NewsTab.tsx
```diff
@@ -27,6 +26,9 @@ const StyledTab = styled(Link)<{ $active: boolean }>`
   font-size: 18px;
   font-weight: 700;
   text-decoration: none;
+  &:not(:last-child) {
```
f1sh1918, line 29: this fixes the alignment issue for the news tabs on mobile view

web/src/components/Layout.tsx
```diff
@@ -156,9 +149,10 @@ const Layout = ({
   disableScrollingSafari = false,
 }: LayoutProps): JSX.Element => {
   const modalVisible = !!modal
-  const { width } = useWindowDimensions()
+  const { width, viewportSmall } = useWindowDimensions()
   const [languageSelectorHeight, setLanguageSelectorHeight] = useState<number>(0)
 
+  // TODO fix wrong panel height when switching from local to tunews
```
f1sh1918, line 155: i really couldn't figure out why the `languageSelector` is getting null only on this route. Maybe some can help me out :)

web/src/components/CityContentToolbar.tsx
```diff
@@ -1,31 +1,31 @@
...
-          openFeedbackModal={openFeedbackModal}
-          viewportSmall={viewportSmall}
-        />
-      </Toolbar>
-    )
-  }
+  return (
+    <Toolbar iconDirection={iconDirection} hasDivider={hasDivider}>
+      {children}
+      <ToolbarItem icon={CopyIcon} text={t('copyLink')} onClick={copyToClipboard} />
```
sarahsporck, line 25: I guess sharing will be done in a separate ticket?

web/src/components/Layout.tsx
sarahsporck: The reason for this is that the language change path is null while tu news are loading. Therefore the language selector is not rendered. Therefore, you should use the defaultLanguageChangePath while loading

web/src/routes/TuNewsPage.tsx
```diff
@@ -74,13 +76,18 @@ const TuNewsPage = ({ cityCode, languageCode, cityModel, languages }: CityRouteP
     )
   }
 
+  const toolbar = (openFeedback: (rating: FeedbackRatingType) => void) => (
```
sarahsporck, line 79: A little weird to put a toolbar below the TuNews List in mobile as it is growing indefinitely. I'll never reach the damn copy button. Maybe we omit it for the small viewport?

web/src/components/Layout.tsx
steffenkleinle: @sarahsporck is right, as long as there is `languageChangePaths` are null, no language selector is rendered and therefore the height is undefined. Once the language selector is rendered, no more rerender/useEffect of the Layout is triggered and therefore the height is never updated. I created a fix in https://github.com/digitalfabrik/integreat-app/pull/955. Automatic retriggering the useEffect would probably be better but not sure how that could work.

steffenkleinle (changes requested):
Observations:
* For example on the [offers route](http://localhost:9000/testumgebung/de/offers) or [categories route with little content](http://localhost:9000/testumgebung/de/test-erste-ebene-aschaffenburg) the toolbar doesn't leave enough space to the top for the language dropdown, see screenshot below. I guess that this is because the toolbar is bound by the height of the content and can't expand further down for contents with small height. Making the content take as much space as available should fix it.
* On middle sized screens (full width is used, no left and right padding around content) the toolbar makes the content jump, maybe we should use a placeholder component (or the empty toolbar without items) here? E.g. reproducible by setting a width of 1000px and switching between the news tabs
* On mobile the toolbar is pushed down by the content once loaded which makes the whole toolbar flickering. It should probably only displayed once the content is loaded on mobile.




I love the design and your PR, well done! :heart_eyes:

web/src/components/Toolbar.tsx
```diff
@@ -21,42 +18,45 @@ const ToolbarContainer = styled.div<{ direction: 'row' | 'column'; hasPadding: b
...
 `
 
-const Headline = styled.h5`
-  width: 100vw;
-  margin: 0;
-  text-align: center;
-  font-size: 90%;
+const Divider = styled.hr`
+  margin: 12px 24px;
+  background-color: ${props => props.theme.colors.poiBorderColor};
```
steffenkleinle, line 34: I think we should change the naming of the color `poiBorderColor` to something more generic

web/src/routes/TuNewsPage.tsx
f1sh1918: yes thats currently the agreement. i can disabled it for mobile viewport

translations/translations.json
```diff
@@ -2,6 +2,7 @@
   "categories": {
     "de": {
       "createPdf": "PDF erstellen",
+      "copyLink": "Link kopieren",
```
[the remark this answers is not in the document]
f1sh1918: for the key i agree. non tech users don't understand "URL kopieren"

web/src/routes/OffersPage.tsx
```diff
@@ -64,10 +64,6 @@ const OffersPage = ({ city, cityCode, languageCode }: CityRouteProps): ReactElem
     return null
   }
 
-  const toolbar = (openFeedback: (rating: FeedbackRatingType) => void) => (
```
steffenkleinle, line 63: Why did you remove the toolbar here?

steffenkleinle (approved):
Sorry to nitpick but I found a few more small issues:
* The toolbar is quickly rendered very much on top before being pushed down and therefore quickly flickers behind the "local information" icon. Observable by switching back and forth between news and local information tabs and looking at the local information icon (or between the news tabs).
* I can still reproduce the toolbar being pushed down on mobile after the content was loaded.
* The bottom border color of the last item in the pois list shadows the divider of the toolbar on desktop which looks a little weird, see screenshot. Maybe we can also just use the same border color for the pois list?
image

Feel free to ignore any or all of those issues and/or create an issue for them, it looks quite amazing!

f1sh1918:
> Sorry to nitpick but I found a few more small issues:
> 
> * The toolbar is quickly rendered very much on top on desktop before being pushed down and therefore quickly flickers behind the "local information" icon. Observable by switching back and forth between news and local information tabs  (or between the local and tünews news tabs) and looking at the local information icon.
> * I can still reproduce the toolbar being pushed down on mobile after the content was loaded.
> * The bottom border color of the last item in the pois list shadows the divider of the toolbar on desktop which looks a little weird, see screenshot. Maybe we can also just use the same border color for the pois list?
>   image
> 
> Feel free to ignore any or all of those issues and/or create an issue for them, it looks quite amazing!

@steffenkleinle i fixed all the issues
1. by using `useLayoutEffect`
2. checking `isLoading` for mobile
3. just relies that you have to restart `yarn start` to use updated theme colors

steffenkleinle (approved):
Can confirm, all issues fixed! Amazing :heart_eyes:

sarahsporck (commented):
Nicely done. 👍🏼 
I did sth. similar like the copy button [in a backend ticket](https://github.com/digitalfabrik/integreat-cms/pull/1779), and we agreed, that it's best to have some feedback, when clicking the button. We could replace the copy icon with a ✔️ for three seconds when the link was copied. For a reference, check out the search feedback in the backend (you can copy the query there). Or checkout how github does this, whe you clone a repo you can click a copy button to copy the repo url.

If you feel like this needs some more discussion let me know. :)

web/src/components/FacetToggle.tsx
```diff
@@ -37,7 +37,7 @@ const FacetToggle: React.FC<FacetToggleProps> = ({
 
   return (
     <FilterFacet>
-      <Tooltip text={tooltip} flow='up'>
+      <Tooltip text={tooltip} flow='up' trigger='hover'>
```
steffenkleinle, line 40: Could we just use `trigger='hover' as default to avoid those changes? Apparently its also definitely the default and most common use case for our tooltips.

translations/translations.json
```diff
@@ -3,6 +3,7 @@
     "de": {
       "createPdf": "PDF erstellen",
       "copyUrl": "Link kopieren",
+      "copyUrlTooltip": "Kopiert!",
```
steffenkleinle, line 6:
```suggestion
      "copied": "Kopiert!",
```

steffenkleinle: Optional: We also have one `Kopiert` translation already, maybe we can reuse that one.

steffenkleinle (approved):

web/src/components/CityContentToolbar.tsx
```diff
@@ -20,7 +20,7 @@ const COPY_TIMEOUT = 3000
 const CityContentToolbar = (props: CityContentToolbarProps) => {
   const { children, openFeedbackModal, iconDirection, hasFeedbackOption = true, hasDivider } = props
   const [linkCopied, setLinkCopied] = useState<boolean>(false)
-  const { t } = useTranslation('categories')
+  const { t } = useTranslation(['categories', 'cityNotCooperating'])
```
steffenkleinle, line 23: Optional: This is actually not necessary, you could keep it like before and then still use `t('copyUrl')` and `t('cityNotCooperating:textCopied')`.
You just need to add the namespace in t keys for unspecified namespaces.

steffenkleinle (approved):
Amazing job :toolbox:

web/src/components/CityContentToolbar.tsx
f1sh1918: thx for the hint. didn't know. adjusted

sarahsporck (approved):
Great! I really like the result 👍🏼

PR #952, opened 2023-06-15
IGAPP-1316: Adjust feedback modal

f1sh1918:
Please test neutral feedback on Testumgebung.
Check styling

Not sure if we need a separate ticket for native?


LeandraH (commented):
Works nicely! I don't love the scrolling in a big viewport though: the background scrolls, even if my cursor is in the modal (in Firefox).

build-configs/common/theme/fonts.ts
```diff
@@ -36,7 +36,7 @@ export const commonFonts = {
   contentFontSize: '0.95rem',
   contentLineHeight: 1.4,
   standardParagraphMargin: '0.75rem',
-  subTitleFontSize: '1.8rem',
```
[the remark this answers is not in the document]
f1sh1918: resized the font-size to 1.125rem (18px) that we need and replaced all 18px occurrences in web.

f1sh1918:
> Works nicely! I don't love the scrolling in a big viewport though: the background scrolls, even if my cursor is in the modal (in Firefox).

Added a general scroll lock to the `FeedbackModal`

LeandraH (approved):
Nice!

web/src/components/FeedbackContainer.tsx
```diff
@@ -111,6 +101,9 @@ export const FeedbackContainer = (props: FeedbackContainerProps): ReactElement =
       const feedbackEndpoint = createFeedbackEndpoint(cmsApiBaseUrl)
       await feedbackEndpoint.request(feedbackData)
       setSendingStatus(SendingState.SUCCESS)
+      if (onSubmitted) {
+        onSubmitted(true)
+      }
```
steffenkleinle, line 106: Why is the check for `onSubmitted` necessary?
```suggestion
      onSubmitted(true)
```

web/src/components/FeedbackContainer.tsx
```diff
@@ -41,11 +30,11 @@ type FeedbackContainerProps = {
   cityCode: string
   language: string
   routeType: RouteType
-  isPositiveFeedback: boolean
   isSearchFeedback: boolean
   closeModal?: () => void
   query?: string
   slug?: string
+  onSubmitted?: (isSubmitted: boolean) => void
```
steffenkleinle, line 37: Same naming issue, I think we should just either rename it to `setIsSubmitted`, or probably even better, change to 
```suggestion
  onSubmit?: () => void
```

web/src/components/Feedback.tsx
```diff
@@ -88,7 +153,13 @@ const Feedback = (props: FeedbackProps): ReactElement => {
       {sendingStatus === SendingState.ERROR && (
         <ErrorSendingStatus role='alert'>{t('failedSendingFeedback')}</ErrorSendingStatus>
       )}
-      <TextButton disabled={!isPositiveFeedback && !comment} onClick={onSubmit} text={t('send')} />
+      {sendFeedbackDisabled && (
+        <NoteContainer>
+          <img src={NoteIcon} alt='' />
+          <NoteText>{t('note')}</NoteText>
+        </NoteContainer>
+      )}
```
steffenkleinle, line 161: Could we just hide the note instead of unrendering it completely? That way, the height of the modal/position of the send button wouldn't jump.

web/src/components/FeedbackModal.tsx
```diff
@@ -69,30 +72,42 @@ type FeedbackModalProps = {
...
+  useScrollToTop()
+  useLockedBody(visible)
 
   return (
     <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
-      <ModalContainer role='dialog' aria-modal>
+      <ModalContainer
+        role='dialog'
+        aria-modal
+        topPosition={viewportSmall ? dimensions.headerHeightSmall : 0}
```
steffenkleinle, line 93: Shouldn't we really do it fullscreen and not show the header? I think otherwise it doesn't really look like a modal/jumps a bit.

steffenkleinle (changes requested):
Nicely done! Tested on firefox (desktop + mobile). On mobile, the padding/size seems off, see screenshot:



Whats left furthermore in my opinion is mostly minor naming issues.
Otherwise I was thinking, if we still need to render the feedback in the Layout or if we can just render it in the toolbar item. Then we could perhaps get rid of all those `openFeedbackModal` handling in the different routes. If you think that would be feasible and make sense, let me know than I would create an issue for it.

steffenkleinle:
> Not sure if we need a separate ticket for native?

At some point we should definitely adjust the feedback there as well (e.g. the icons). Actually, maybe it makes sense to move the Feedback there to the overflow menu in the header. But I guess discussion is needed here and a design would be nice. Opening an issue already still seems like a good idea :)

steffenkleinle:
> Not sure if we need a separate ticket for native?

Created an issue: https://issues.tuerantuer.org/browse/IGAPP-1330

f1sh1918:
> Nicely done! Tested on firefox (desktop + mobile). On mobile, the padding/size seems off, see screenshot: 
> 
> Whats left furthermore in my opinion is mostly minor naming issues. Otherwise I was thinking, whether we still need to render the feedback in the Layout or if we can just render it in the toolbar item. Then we could perhaps get rid of all those `openFeedbackModal` handling in the different routes. If you think that would be feasible and make sense, let me know than I would create an issue for it.

Yes i think it make more sense to add this in the toolbar. Feel free to create a ticket

web/src/components/FeedbackModal.tsx
f1sh1918: i prefer showing the header on mobile as shown in design

web/src/components/Feedback.tsx
f1sh1918: just set the opacity to 0 or 1

web/src/components/CityContentLayout.tsx
```diff
@@ -27,9 +26,13 @@ type CityContentLayoutProps = {
 }
 
 const CityContentLayout = (props: CityContentLayoutProps): ReactElement => {
-  const [feedbackModalRating, setFeedbackModalRating] = useState<FeedbackRatingType | null>(null)
+  const [openFeedbackModal, setOpenFeedbackModal] = useState<boolean>(false)
   const { viewportSmall } = useWindowDimensions()
 
+  const onCloseFeedbackModal = () => {
```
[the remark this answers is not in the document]
f1sh1918: just passed function directly to the component
`closeModal={() => setOpenFeedbackModal(false)}`

web/src/components/Feedback.tsx
```diff
@@ -40,13 +45,48 @@ export const ErrorSendingStatus = styled.div`
...
+const NoteContainer = styled.div`
+  display: flex;
+  margin-top: 8px;
+  background-color: ${props => props.theme.colors.themeColor};
+  padding: 8px;
+`
+
+const NoteText = styled(StyledSmallViewTip)`
+  margin-block-start: 0;
+  margin-left: 12px !important;
```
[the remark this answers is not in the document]
f1sh1918: i can not override the margin of this component `StyledSmallViewTip` no idea why this doesn't work here for styled components

web/src/components/FeedbackContainer.tsx
steffenkleinle: Must have looked wrong first time, I thought `onSubmit` is always truthy

steffenkleinle:
> > Otherwise I was thinking, whether we still need to render the feedback in the Layout or if we can just render it in the toolbar item. Then we could perhaps get rid of all those `openFeedbackModal` handling in the different routes. If you think that would be feasible and make sense, let me know than I would create an issue for it.
> 
> Yes i think it make more sense to add this in the toolbar. Feel free to create a ticket

https://issues.tuerantuer.org/browse/IGAPP-1339

PR #956, opened 2023-06-23
IGAPP-1234: Proposal

steffenkleinle:

### DIFF
--- a/web/src/components/PoisDesktop.tsx
+++ b/web/src/components/PoisDesktop.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import React, { ReactElement, useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { GeolocateControl, NavigationControl } from 'react-map-gl'
 import styled from 'styled-components'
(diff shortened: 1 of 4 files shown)


steffenkleinle:
I still struggled a little to understand the different responsibilities between PoisPage, PoisMobile/Desktop and the two hooks. While trying to understand, I played around a bit and found that my proposal may help with that a bit by removing one of the two hooks (imo the one for PoisPage is not necessary, we only need to know the current poi) and moving a little more shared to the now merged usePoiFeatures hook that is only used in PoisMobile and PoisDesktop.

Feel free to close this without any reason if you are not happy with it :)

web/src/components/PoisDesktop.tsx
```diff
@@ -141,8 +138,8 @@ const PoisDesktop = ({
             <ListHeader>{t('listTitle')}</ListHeader>
           )}
 
-          {currentPoi && currentPoiFeature ? (
-            <PoiDetails poi={currentPoi} feature={currentPoiFeature} direction={direction} t={t} />
+          {currentPoi ? (
+            <PoiDetails poi={currentPoi} feature={currentPoi.feature} direction={direction} t={t} />
```
steffenkleinle, line 142: @sarahsporck is this the same? Or do we have to select the current feature from `currentFeatureOnMap`?

PR #958, opened 2023-06-27
IGAPP-1310: SVG transformer

sarahsporck:

To test:
- every svg/image looks and scales the same
- web didn't change (especially the icons)


steffenkleinle (approved):
Nice! Tested on android and firefox

f1sh1918 (changes requested):
Nice that you found a way to get it working :)

Some icons are not loading for me in chrome mobile. Please check all the removed pngs if they are used anywhere in web app for the several projects!
image

image

Native App on ios looks fine

sarahsporck:
> Some icons are not loading for me in chrome mobile. Please check all the removed pngs if they are used anywhere in web app for the several projects! 

Very nice find. I didn't that this would impact web mobile. I set the correct image url. Now it works.

native/metro.config.js
```diff
@@ -1,52 +1,64 @@
...
+  {
+    resolver: {
+      assetExts: defaultAssetExts.filter(ext => ext !== 'svg'),
+      sourceExts: [...defaultSourceExts, 'svg'],
+      extraNodeModules: new Proxy(
+        {},
+        {
+          get: (target, name) => {
+            if (name === 'build-config-name') {
+              const buildConfigName = process.env.BUILD_CONFIG_NAME || 'integreat-test-cms'
```
f1sh1918, line 33: why only 'integreat-test-cms' and not also malte?

native/src/routes/CityNotCooperating.tsx
```diff
@@ -83,12 +77,13 @@ const CityNotCooperating = (): ReactElement => {
...
   }
 
+  const CityNotCooperatingIcon = buildConfigAssets().CityNotCooperatingIcon!
   return (
     <Container>
       <Heading>{t('callToAction')}</Heading>
 
       <Description>{t('explanation')}</Description>
-      <Icon source={buildConfigAssets().cityNotCooperatingIcon!} />
+      <CityNotCooperatingIcon width='50%' style={{ alignSelf: 'center' }} />
```
f1sh1918, line 86: say no to inline style :) please used styled component

f1sh1918 (approved):
Very nice implementation!!! Nice work :)

native/metro.config.js
steffenkleinle: This is a fallback if the env variable is not set I think and the test cms is just the default here I think

sarahsporck: Think so, too. This was just copy/pasted :)

steffenkleinle:
@sarahsporck I just realized some of the icon sizes are a little off, e.g. the icon in the Failure component and the loading spinner are too small compared to before

PR #960, opened 2023-06-28
IGAPP-1328: Toolbar not sticky

f1sh1918:
can only be reproduced on webnext


steffenkleinle (approved):
not tested

sarahsporck (approved):
Tested. Actually, why does the Aside have a z-index of 10? is there Any situation where this actually makes sense?

f1sh1918:
> Tested. Actually, why does the Aside have a z-index of 10? is there Any situation where this actually makes sense?

no clue. you can create a ticket and remove it. will merge this urgent fix now :)

PR #964, opened 2023-07-04
IGAPP-1329: Text button component

steffenkleinle:

### DIFF
--- a/native/src/components/Feedback.tsx
+++ b/native/src/components/Feedback.tsx
@@ -1,5 +1,4 @@
-import * as React from 'react'
-import { ReactElement } from 'react'
+import React, { ReactElement } from 'react'
 import { TFunction } from 'react-i18next'
 import { ActivityIndicator, ScrollView, Text, TextInput } from 'react-native'
 import Icon from 'react-native-vector-icons/MaterialIcons'
@@ -44,6 +43,10 @@ const Description = styled(ThemedText)`
 `
 const HappyIconContainer = styled(HappyIcon)`
   margin: 100px auto 10px;
+`
+
+const ButtonContainer = styled.View`
+  padding: 16px 0;
 `
 
 export type FeedbackProps = {
@@ -93,12 +96,15 @@ const Feedback = (props: FeedbackProps): ReactElement => {
             value={contactMail}
           />
           {sendingStatus === 'failed' && <Description theme={theme}>{t('failedSendingFeedback')}</Description>}
-          <PrimaryTextButton
-            Icon={<Icon name='send' size={15} color='black' />}
-            disabled={!isPositiveFeedback && !comment}
-            onPress={onSubmit}
-            text={t('send')}
-          />
+          <ButtonContainer>
+            <PrimaryTextButton
+              Icon={<Icon name='send' size={15} color='black' />}
+              disabled={!isPositiveFeedback && !comment}
+              onPress={onSubmit}
+              text={t('send')}
+              padding={false}
+            />
+          </ButtonContainer>
         </>
       )
     }


LeandraH (approved):
Nice, tested in emulated iOS

native/src/components/PrimaryTextButton.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { ReactElement } from 'react'
+import { Button } from 'react-native-elements'
+import { IconNode } from 'react-native-elements/dist/icons/Icon'
+import styled, { useTheme } from 'styled-components/native'
+
+const Container = styled.View<{ $padding: boolean }>`
+  padding: ${props => (props.$padding ? '16px' : '0')};
+`
```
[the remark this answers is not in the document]
steffenkleinle: We could, but in general we try to avoid inline styles in favor of styled components

native/src/components/Feedback.tsx
```diff
@@ -93,18 +93,11 @@ const Feedback = (props: FeedbackProps): ReactElement => {
...
-              backgroundColor: theme.colors.themeColor,
-              marginTop: 15,
-            }}
+          <PrimaryTextButton
+            Icon={<Icon name='send' size={15} color='black' />}
             disabled={!isPositiveFeedback && !comment}
             onPress={onSubmit}
-            title={t('send')}
+            text={t('send')}
           />
```
sarahsporck, line 101: Here the button should use the whole width.

native/src/components/PrimaryTextButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+        title={text}
+        disabled={disabled}
+        titleStyle={{
+          color: theme.colors.textColor,
+          fontFamily: theme.fonts.native.contentFontBold,
+          fontSize: 14,
+        }}
+        buttonStyle={{
+          backgroundColor: theme.colors.themeColor,
+          borderRadius: 8,
```
sarahsporck, line 41: This changes the borderRadius for some buttons compared to before. Although I see that you probably tried to align this with the new feedback design right?

sarahsporck (approved):
Tested on simulated ios and android emulator.
Just one small comment :)

native/src/components/PrimaryTextButton.tsx
steffenkleinle: Yes, both the new feedback design and the design for the poi filter (https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Jira?node-id=961%3A3409&t=HoTJnEmjdDKZVa3n-1) have the rounded buttons and I figured it makes sense to make it consistent. You think we shouldn't?

sarahsporck: No, it is fine as it is. Although, I personally prefer the corners. :)

native/src/components/PrimaryTextButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const Container = styled.View<{ $padding: boolean }>`
+  padding: ${props => (props.$padding ? '16px' : '0')};
+`
+
+type PrimaryTextButtonProps = {
+  onPress: () => void
+  text: string
+  disabled?: boolean
+  Icon?: IconNode
```
f1sh1918, line 14: i'm really not a big fan to be bound to icons from a specific library for a base component. What if want to have a specific svg icon here in future?

native/src/components/PrimaryTextButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  onPress,
+  text,
+  padding = true,
+  disabled = false,
+  Icon,
+}: PrimaryTextButtonProps): ReactElement => {
+  const theme = useTheme()
+
+  return (
+    <Container $padding={padding}>
```
f1sh1918, line 28: why don't you use the `containerStyle` that the library offers and pass a style instead of padding. That makes the component more flexible. Even you can define a default style.

native/src/components/PrimaryTextButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type PrimaryTextButtonProps = {
+  onPress: () => void
+  text: string
+  disabled?: boolean
+  Icon?: IconNode
+  padding?: boolean
+}
+
+const PrimaryTextButton = ({
```
f1sh1918, line 18: I think its better to call it TextButton and pass a `variant` that can be primary, secondary, whatever.
I think you should talk with Toni about the design system. Now we refactor and later we refactor again when different text button variants comes up. I think this should take place in a `component/base` folder

native/src/components/SlideButton.tsx
```diff
@@ -7,12 +7,11 @@ type SlideButtonProps = {
   label: string
   onPress: () => void | Promise<void>
   theme: ThemeType
-  highlighted?: boolean
 }
 
-const SlideButton = ({ label, onPress, theme, highlighted }: SlideButtonProps): ReactElement => (
+const SlideButton = ({ label, onPress, theme }: SlideButtonProps): ReactElement => (
```
f1sh1918, line 12: i think there is no need for a separate component when just some styles change. This just should be another `variant` of a `<TextButton/>`

f1sh1918 (commented):
I think its a good start for a refactoring but we should do way more of restructuring when we build "base components" I think we should have a `TextButton` component with different variants like other component libraries do f.e. MUI or React Bootstrap.
That makes our components more flexible and we are prepared for a design system.

PR #966, opened 2023-07-05
IGAPP-1331: Update RN to 0.71

LeandraH:


native/src/components/OpeningHours.tsx
```diff
@@ -62,7 +62,7 @@ const OpeningHours = ({
   const openingHoursTitle = (
     <TitleContainer language={language}>
       <Text style={{ fontWeight: 'bold', alignSelf: 'center' }}>{t('openingHours')}</Text>
-      <OpeningLabel isOpened={isCurrentlyOpen} direction={contentDirection(language)}>
+      <OpeningLabel isOpened={isCurrentlyOpen} contentDirection={contentDirection(language)}>
```
LeandraH, line 65: Apparently, a `Text` can throw when it has a property `direction`. https://stackoverflow.com/questions/60479653/invalid-ygdirection-row-should-be-one-of-inherit-ltr-rtl-react-native

native/android/.idea/compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="1.8" />
+    <bytecodeTargetLevel target="11" />
```
steffenkleinle, line 4: Please revert if there isn't a specific reason to change this file
```suggestion
    <bytecodeTargetLevel target="1.8" />
```

steffenkleinle (approved):
Seems like https://github.com/digitalfabrik/integreat-app/pull/949#pullrequestreview-1510988641 is fixed with this, nice job! :tada: 

Tested on Android, seems to work just fine :)

native/package.json
```diff
@@ -86,13 +86,13 @@
...
-    "react-native-safe-area-context": "^4.5.0",
-    "react-native-screens": "^3.18.0",
+    "react-native-safe-area-context": "^4.5.1",
+    "react-native-screens": "^3.19.0",
     "react-native-svg": "^13.9.0",
     "react-native-system-setting": "^1.7.6",
     "react-native-url-polyfill": "^1.3.0",
     "react-native-vector-icons": "^9.2.0",
-    "react-native-webview": "^11.22.7",
+    "react-native-webview": "^12.0.2",
```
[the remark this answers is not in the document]
LeandraH: Honestly, because the upgrade helper suggested it. I guess I'll check if things also work with the latest version.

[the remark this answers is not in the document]
steffenkleinle: No, thats okay :) We can upgrade it separately.

native/android/app/src/main/java/tuerantuer/app/integreat/ReactNativeFlipper.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * <p>This source code is licensed under the MIT license found in the LICENSE file in the root
+ * directory of this source tree.
+ */
+package com.rndiffapp;
```
sarahsporck, line 7:
```suggestion
package tuerantuer.app.integreat;
```

sarahsporck (commented):
On android: When navigating to testumgebung/sprache and scrolling to the bottom of the screen the app crashes.

ios is fine. Nice. Lot's of unnecessary code removed.

LeandraH:
@sarahsporck I can't reproduce the crashing on Android but I also changed a few more things, so that may be why. Can you check if you can still reproduce it?

sarahsporck:
The crash still occurs. But I also found the solution for this.
https://github.com/react-native-webview/react-native-webview/issues/811
Setting opacity 0.99 on the webview seems to be the best solution honestly, as it does not have any impact on performance.

LeandraH:
@sarahsporck I updated Webview to the latest version since that package seems to be where the bug is happening, can you test again? (Sorry to keep bugging you)

sarahsporck:
> @sarahsporck I updated Webview to the latest version since that package seems to be where the bug is happening, can you test again? (Sorry to keep bugging you, and on your birthday no less! Happy birthday?)

Crash still happens :/.

LeandraH:
@sarahsporck I put the fix back in, can you check (hopefully) one last time?

sarahsporck (approved):
Again tested in Android. Seems fixed now. Thank you :)

PR #967, opened 2023-07-05
IGAPP-1332: Remove js-md5 and replace with md5

steffenkleinle:


sarahsporck (commented):
Any guidelines on how to test this? :)

steffenkleinle:
> Any guidelines on how to test this? :)

You can do a production build in web and then search for `eval("require('crypto')")` and if you don't find anything we should be good.

sarahsporck (approved):
so when looking at the yarn.lock the `crypto` library should not be bundled.

LeandraH (approved):
Looks reasonable, and I didn't get any errors in Firefox or emulated iOS :)

PR #971, opened 2023-07-12
IGAPP-1330: Adjust feedback native

f1sh1918:
The feedback can now be found in the overflow menu.
Please test properly if the correct feedback was sent to the cms!


native/src/components/Header.tsx
```diff
@@ -137,6 +151,9 @@ const Header = ({
           ? [renderOverflowItem(HeaderButtonTitle.Location, () => navigation.navigate(LANDING_ROUTE))]
           : []),
         renderOverflowItem(HeaderButtonTitle.Settings, () => navigation.navigate(SETTINGS_ROUTE)),
+        ...(route.name !== NEWS_ROUTE
+          ? [renderOverflowItem(HeaderButtonTitle.Feedback, () => navigateToFeedback())]
```
steffenkleinle, line 155:
```suggestion
          ? [renderOverflowItem(HeaderButtonTitle.Feedback, navigateToFeedback)]
```

native/src/components/Header.tsx
```diff
@@ -124,6 +127,17 @@ const Header = ({
     }
   }
 
+  const navigateToFeedback = () => {
+    if (cityCode) {
+      createNavigateToFeedbackModal(navigation)({
+        routeType: route.name as RouteType,
+        language: languageCode,
+        cityCode,
+        slug: (route.params as RoutesParamsType[PoisRouteType]).slug,
```
steffenkleinle, line 136: Why as `PoisRouteType`? This might e.g. break feedback for the root category

native/src/components/FeedbackButtons.tsx
```diff
@@ -0,0 +1,77 @@
...
+  align-items: center;
+  border-radius: 18px;
+  width: 100px;
+  height: 80px;
+  justify-content: center;
+`
+const CardShadow = styled.View<{ $active: boolean | null }>`
+  background-color: ${props => (props.$active ? `${props.theme.colors.themeColor}` : 'white')};
+  border-radius: 18px;
+  shadow-color: #171717;
```
steffenkleinle, line 27: Do we actually need a new color for this?

native/src/routes/SprungbrettOfferContainer.tsx
```diff
@@ -36,24 +29,13 @@ const SprungbrettOfferContainer = ({ route, navigation }: SprungbrettOfferContai
   const shareUrl = urlFromRouteInformation({ route: SPRUNGBRETT_OFFER_ROUTE, languageCode, cityCode })
   useHeader({ navigation, route, availableLanguages, data, shareUrl })
 
-  const navigateToFeedback = (isPositiveFeedback: boolean) => {
-    createNavigateToFeedbackModal(navigation)({
-      routeType: OFFERS_ROUTE,
-      cityCode,
-      slug: SPRUNGBRETT_OFFER_ROUTE,
```
steffenkleinle, line 43: I think this also breaks with the changes

steffenkleinle (changes requested):
Tested on Android, looking good!
The only problem is the wrong addressing of the feedback/crashing on a few routes:
- For offers, sprungbrett, disclaimer the app crashes
- For ALL categories the feedback is interpreted as region feedback

Perhaps we should also remove the auto focus on the comment field, not that sure though.

native/src/components/FeedbackButtons.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const FeedbackTouchableOpacity = styled(TouchableOpacity)`
+  align-items: center;
+  border-radius: 18px;
+  width: 100px;
+  height: 80px;
+  justify-content: center;
+`
+const CardShadow = styled.View<{ $active: boolean | null }>`
+  background-color: ${props => (props.$active ? `${props.theme.colors.themeColor}` : 'white')};
```
steffenkleinle, line 25: Please use `backgroundColor` instead of `white`

native/src/routes/SprungbrettOfferContainer.tsx
f1sh1918: added this to the header to define correct slug for this route

native/src/components/FeedbackButtons.tsx
f1sh1918: use background color

native/src/components/Header.tsx
f1sh1918: i added a function to define to correct slug for each route

native/src/components/Header.tsx
```diff
@@ -122,13 +142,42 @@ const Header = ({
...
+        return (route.params as RoutesParamsType[PoisRouteType]).slug
+
+      case CATEGORIES_ROUTE:
+        return getCategorySlug((route.params as RoutesParamsType[CategoriesRouteType]).path)
+
+      case SPRUNGBRETT_OFFER_ROUTE:
+        return SPRUNGBRETT_OFFER_ROUTE
+
+      case DISCLAIMER_ROUTE:
+        return (data as UseLoadExtraCityContentData<PageModel>).extra.slug
```
steffenkleinle, line 167: I think this should actually always be `disclaimer` so you should be able to just change this to
```suggestion
        return DISCLAIMER_ROUTE
```

native/src/hooks/useHeader.tsx
```diff
@@ -29,6 +29,7 @@ const useHeader = <T extends RoutesType>({
           showItems={!!data}
           availableLanguages={availableLanguages}
           shareUrl={shareUrl}
+          data={data ?? undefined}
```
steffenkleinle, line 32: I think this can be reverted

native/src/hooks/useLoadExtraCityContent.ts
```diff
@@ -22,7 +22,7 @@ type Load<T> =
 
 type UseLoadExtraCityContentParams<T> = Params & Load<T>
 
-type UseLoadExtraCityContentData<T> = CityContentData & { extra: T }
+export type UseLoadExtraCityContentData<T> = CityContentData & { extra: T }
```
steffenkleinle, line 25: I think this can be reverted

steffenkleinle (approved):
Not tested again

web/src/components/Feedback.tsx
LeandraH, line 1: Should this file be in this PR?

native/src/components/FeedbackButtons.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const FeedbackButtons = ({ isPositiveFeedback, setIsPositiveFeedback }: FeedbackButtonsProps): ReactElement => {
+  const { t } = useTranslation('feedback')
+  return (
+    <Container>
+      <CardShadow $active={isPositiveFeedback}>
+        <FeedbackTouchableOpacity
+          onPress={() => setIsPositiveFeedback(isPositiveFeedback !== true ? true : null)}
+          testID='feedback-positive-rating'>
```
LeandraH, line 60: Do we need the testID if we have text on the button?

LeandraH (approved):
Nice, tested in simulated iOS

PR #973, opened 2023-07-13
IGAPP-1323: List items fullwidth

f1sh1918:

http://localhost:9000/testumgebung/de/events


sarahsporck (changes requested):
Nice! Can you also adjuste the POIListItems? :)

web/src/components/ListItem.tsx
```diff
@@ -34,6 +34,10 @@ const Title = styled.div`
   font-weight: 700;
 `
 
+const FullWidthLink = styled(StyledCleanLink)`
```
steffenkleinle, line 37: Why are you not using the default export component?

f1sh1918: because we can use the standard styles from the styled clean link. If not i would have to duplicate the style code

f1sh1918:
> Nice! Can you also adjuste the POIListItems? :)

good catch updated pr

sarahsporck (approved):
Tested in Firefox. Works like charm :)

web/src/components/ListItem.tsx
steffenkleinle: Yeah I got that but why the named export `StyledCleanLink` and not the deafult export `CleanLink` which internally uses the `StyledCleanLink`?

steffenkleinle (approved):

PR #974, opened 2023-07-13
IGAPP-1344: Ios url schemes

f1sh1918:
Ios Simulator: go to testumgebung -> map -> tür an tür cafe > click on telephone or mail


sarahsporck (approved):
I think we could also add https to the url schemes, as there could be issues with other browsers than safari. Can you test this shortly? https://developer.apple.com/forums/thread/660241

steffenkleinle (approved):

steffenkleinle:
Not tested

f1sh1918:
I'm on vacation pls feel free to add and merge @sarahsporck

PR #981, opened 2023-07-19
IGAPP-1255: Disable e2e tests and builds on feature branches

SteffiStuffel:


.circleci/src/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,45 @@
+when:
+  or:
+    - and:
+        - << pipeline.parameters.api_triggered >>
+        - equal: [<< pipeline.parameters.workflow_type >>, e2e_tests]
+    - and:
+        - equal: [main, << pipeline.git.branch >>]
+        - not: << pipeline.parameters.api_triggered >>
```
LeandraH, line 8: Just to clarify because I'm confused by CircleCI's syntax, this means: run if someone triggers the e2e tests via API or if we're on the branch main and it hasn't been triggered by the API?

If that's right, then I'm not sure why we wouldn't want to run the tests on main for a manual trigger?

LeandraH (approved):
Looks good

docs/cicd.md
```diff
@@ -26,6 +26,7 @@ Several workflows exist for different purposes:
 | native_production_delivery | script                 | :heavy_check_mark: | production      | :x:          | :heavy_check_mark: | :x:                |
 | web_beta_delivery          | script                 | :heavy_check_mark: | :x:             | beta         | :heavy_check_mark: | :x:                |
 | web_production_delivery    | script                 | :heavy_check_mark: | :x:             | production   | :heavy_check_mark: | :x:                |
+| e2e_tests                  | commits on main        | :x:                | :x:             | :x:          | :x:                | :x:                |
```
sarahsporck, line 29: Did you test triggering this manually? The command would be like this, right?
`cd tools && yarn trigger-pipeline trigger e2e_tests --api-token <api-token>`?

steffenkleinle (commented):
Please also add `e2e-tests` as possibility to the trigger-workflow script

.circleci/src/workflows/e2e_tests.yml
SteffiStuffel: It would run on main for a manual trigger, as the first condition (first half of the or) is true then.

docs/cicd.md
SteffiStuffel: Yes I triggered it manually, see here. I did it by just calling the circleci api, as there is no need for a script, but as Steffen also wishes not to use the api, but have a script, I will add one...

See workflow here:
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/8704/workflows/ba0944f6-b798-4c21-9dcc-99ac2ce0a0d9

.circleci/src/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  - bump_version:
+      prepare_delivery: false
+  - build_web:
+      name: build_ig_test_cms_web
+      build_config_name: integreat-test-cms
+      requires:
+        - bump_version
```
sarahsporck, line 16: this is actually not required for the web e2e tests and can removed here.

sarahsporck (approved):
I left anothe small comment, besides that I am a little excited for this.

.circleci/src/workflows/e2e_tests.yml
SteffiStuffel: It is needed as it is somehow our "setup task" perparing the environment variables.

PR #982, opened 2023-07-25
IGAPP-1354: Remove empty space at top of small screens

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/IGAPP-1354

### DIFF
--- a/web/src/components/Layout.tsx
+++ b/web/src/components/Layout.tsx
@@ -47,6 +47,7 @@ const Body = styled.div<{ fullWidth: boolean; disableScrollingSafari: boolean }>
   width: 100%;
   box-sizing: border-box;
   margin: 0 auto;
+  flex-grow: 1;
   background-color: ${props => props.theme.colors.backgroundColor};
   word-wrap: break-word;
   /* Fix jumping iOS Safari Toolbar by prevent scrolling on body */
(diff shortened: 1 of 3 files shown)


web/src/components/Layout.tsx
```diff
@@ -69,11 +69,6 @@ const Body = styled.div<{ fullWidth: boolean; disableScrollingSafari: boolean }>
         padding-left: calc((100vw - ${dimensions.maxWidth}px) / 2);
       }
     `};
-
-  @media screen and ${dimensions.smallViewport} {
-    display: flex;
-    flex-direction: column-reverse;
```
LeandraH, line 75: I have no idea why we used that `column-reverse` since it means that content starts being displayed from the end of the space (and with the first elements at that end of the space but that is fine here since we only have one element inside).

steffenkleinle: Maybe it got broken when I removed the feedback modal from here... wouldn't know why though.

SteffiStuffel (changes requested):
Unfortunatly this does not work, see the attached video:


https://github.com/digitalfabrik/integreat-app/assets/18463686/12650c7e-ed02-451d-b21b-634c97af4ae1

f1sh1918 (approved):
Works nice! Just added a commit to fix missing toolbar divider for news detail

image

steffenkleinle (approved):
Not tested but looks good 👍

SteffiStuffel (commented):
Not tested again

PR #984, opened 2023-07-26
IGAPP-1168: Minimize css

SteffiStuffel:


sarahsporck (approved):
Nice find! Tested in firefox.

f1sh1918 (changes requested):
How did you check the bundle sizes and the effect of this plugin @SteffiStuffel ? 
Since we don't use much css and the libraries are minified that we import, i don't think this plugin is worth to use it in our case.
When i tried to start a build i even get an error with `bundle size limit reached` and then i checked the parsed bundle sizes, resulting this. 

Before
image


After
image

SteffiStuffel:
@f1sh1918 Good find, thanks for mentioning. I accidentally overwrote the default Terser config. I readded it.

f1sh1918 (commented):
@SteffiStuffel 

i just created bundles with main and your branch and i don't get any errors.
But i really don't see any reason to add an additional plugin that doesn't bring any value in our case. The bundle sizes are the same at least what i could figure out and we are just adding an additional dependency that could fail during dependency updates.

If we are planning to use more css in the future what i don't see we could think about adding this plugin

Stats

your branch

stat:4.75MB
parsed: 2.49MB
gzipped:737.02KB

main
stat:4.75MB
parsed: 2.49MB
gzipped:737.02KB

If i didn't check properly and you can find some improvements according the bundle sizes please let me know

SteffiStuffel:
Will close as @f1sh1918 does not want to include this plugin

steffenkleinle:
@SteffiStuffel if you see a reason to include this plugin, please say so. Otherwise I think the objection of @f1sh1918 is totally valid, no need to include another plugin/dependency if there is no measurable benefit of doing so.

Also, can the branch be deleted?

SteffiStuffel:
> @SteffiStuffel if you see a reason to include this plugin, please say so, could obviously be that we are missing something. Otherwise I think the objection of @f1sh1918 is totally valid, no need to include another plugin/dependency if there is no measurable benefit of doing so.
> 
> Also, can the branch be deleted?

There is a (small) improvement of this plugin, as we have some css (of the fonts), which will be compressed by this plugin. It can be, that this also imporves some ratings, (maybe of google) as we follow best practices when compressing css, but this is just a guess.

steffenkleinle:
> > @SteffiStuffel if you see a reason to include this plugin, please say so, could obviously be that we are missing something. Otherwise I think the objection of @f1sh1918 is totally valid, no need to include another plugin/dependency if there is no measurable benefit of doing so.
> > Also, can the branch be deleted?
> 
> There is a (small) improvement of this plugin, as we have some css (of the fonts), which will be compressed by this plugin. It can be, that this also imporves some ratings, (maybe of google) as we follow best practices when compressing css, but this is just a guess.

According to @f1sh1918 calculation its not measurable, do you have some different numbers/calculations where the improvement shows?

PR #986, opened 2023-07-26
IGAPP-1340: Adjust colors and icons

sarahsporck:

@f1sh1918 I get an issue with an icon that cannot be fetched from the cms when i start the dev server. Can you look into this until I can log into the cms myself? :)

Testable with `yarn start:obdach`


f1sh1918:
> 
> @f1sh1918 I get an issue with an icon that cannot be fetched from the cms when i start the dev server. Can you look into this until I can log into the cms myself? :)
> 
> Testable with `yarn start:obdach`

I guess this relies to some new imageCache maybe. You can activate CORS for local development in this case @sarahsporck

web/src/components/HeaderLogo.tsx
```diff
@@ -18,7 +18,7 @@ const LogoContainer = styled.div`
   display: flex;
   justify-content: start;
   align-items: center;
-  flex: 0 1 220px;
+  flex: initial;
```
f1sh1918, line 21: tbh couldn't see that this is needed! I think you can remove it

f1sh1918 (approved):
Works fine. just one small comment

LeandraH (approved):
Looks good, tested in Firefox

steffenkleinle (approved):

web/src/components/HeaderLogo.tsx
sarahsporck: Without this it looks like this for me on firefox: 
Bildschirmfoto 2023-07-27 um 10 35 52
And with `flex: initial` it looks like this:
Bildschirmfoto 2023-07-27 um 10 35 52

PR #989, opened 2023-07-27
IGAPP-1324: Fix for push notifications

SteffiStuffel:

Explanations: Problem why Push Notifications (PN) are not working is the workflow:
1) Push Notifications are currently disabled in the settings.
2) When enabling PN in the settings, we first run the "changeAction" and if this is successful, we save the changed value to the Settings in the storage.
While running changeAction the PNManager checks if the pushNotifications are allowed and if not it will not subscribe.

There are different solutions for this:
1) Change the whole logic, but I think it the logic is good, as it really makes sense, to only persist the settings if they can be applied properly.
2) Skip the checking if the PN are allowed when subscribing in general, but I think this is not a good idea, as we use this function in differnt places
3) The solution I decided is the best: Add a param to the subscribe function, if the check should be skipped in a specific call.


LeandraH (approved):
Great idea! Not tested

steffenkleinle (commented):
Good find. However, I am a little skeptical if that is the reason for the unreceived push notifications described in the issues, the descriptions don't really match:
1) All of the people that did not receive PNs at some point would have had to first not allow PNs and then enable them manually at a later point
2) This fix should only fix a bug where users NEVER get PNs and not sometimes right?
Maybe I am also wrong though.

SteffiStuffel:
> Good find. However, I am a little skeptical if that is the reason for the unreceived push notifications described in the issues, the descriptions don't really match:
> 
> 1. All of the people that did not receive PNs at some point would have had to first not allow PNs and then enable them manually at a later point
> 2. This fix should only fix a bug where users NEVER get PNs and not sometimes right?
>    Maybe I am also wrong though.

You are right, I cannot find the cause or reproduce this sometimes, sometimes not. I suggest to merge this ticket, as it fixes PN errors and open a new investigate ticket, which we ask Martin to do?

steffenkleinle (approved):
LGTM

SteffiStuffel:
New ticket is here https://issues.tuerantuer.org/browse/IGAPP-1375

PR #950, opened 2023-06-13
IGAPP-281: Replace moment.js with luxon

lunars97:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **IGAPP**.

PR #992, opened 2023-07-31
IGAPP-1358: Fix date in recurring allday events

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/IGAPP-1358

### DIFF
--- a/api-client/src/endpoints/createEventsEndpoint.ts
+++ b/api-client/src/endpoints/createEventsEndpoint.ts
@@ -91,7 +91,16 @@ const getEndDate = (event: JsonEventType, startDate: Date): Date => {
   return moment(startDate).add(durationInDays, 'days').toDate()
 }
 
-const dateToString = (date: Date): string => date.toISOString().split('T')[0]!
+const leftPad = (number: number): string => {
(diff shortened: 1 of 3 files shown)


f1sh1918 (approved):
Hm seems like there is no usable date function. so this is a nice helper fkt

api-client/src/endpoints/createEventsEndpoint.ts
```diff
@@ -91,7 +91,16 @@ const getEndDate = (event: JsonEventType, startDate: Date): Date => {
...
+const leftPad = (number: number): string => {
+  const numberAsString = number.toString()
+  if (numberAsString.length === 1) {
+    return `0${number}`
+  }
+  return numberAsString
+}
+
+export const dateToString = (date: Date): string =>
+  `${date.getFullYear()}-${leftPad(date.getMonth() + 1)}-${leftPad(date.getDate())}`
```
steffenkleinle, line 103: I think this will become redundant with #950. However, nice fix for now :)

steffenkleinle (approved):
Event doesn't exist anymore but seems to do the trick. Good job :)

PR #1000, opened 2023-08-02
IGAPP-1349: Fix linting complaints

LeandraH:
This pull request belongs to https://issues.tuerantuer.org/browse/IGAPP-1349

It just fixes some linting complaints in the web folder that started showing up after the last RN update.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -109,6 +109,7 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'error',
     '@typescript-eslint/prefer-ts-expect-error': 'error',
     '@typescript-eslint/switch-exhaustiveness-check': 'error',
+    '@typescript-eslint/no-non-null-assertion': 'off',
 
     'jest/consistent-test-it': 'error',
     'jest/no-alias-methods': 'error',
--- a/api-client/src/endpoints/__tests__/createPOIsEndpoint.spec.ts
+++ b/api-client/src/endpoints/__tests__/createPOIsEndpoint.spec.ts
@@ -93,7 +93,7 @@ describe('pois', () => {
   }
   it('should map params to url', () => {
     expect(pois.mapParamsToUrl(params)).toBe(
-      'https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/locations/?on_map=1'
+      'https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/locations/?on_map=1',
     )
   })
   const json = [poi1, poi2, poi3, poi4]
--- a/web/stylelint.config.js
+++ b/web/stylelint.config.js
@@ -7,5 +7,6 @@ module.exports = {
     'function-no-unknown': null,
     'at-rule-no-unknown': null,
     'no-extra-semicolons': null,
+    'media-query-no-invalid': null,
   },
 }
--- a/web/stylelint.config_css.js
+++ b/web/stylelint.config_css.js
@@ -1,7 +1,6 @@
 module.exports = {
   extends: ['stylelint-config-standard'],
   rules: {
-    indentation: 2,
     'unit-allowed-list': ['em', 'rem', 'px', '%'],
     'string-quotes': null,
   },
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
   "semi": false,
   "singleQuote": true,
   "arrowParens": "avoid",
--- a/api-client/src/Endpoint.ts
+++ b/api-client/src/Endpoint.ts
@@ -26,7 +26,7 @@ class Endpoint<P, T> {
     mapParamsToBody: MapParamsToBodyType<P> | null | undefined,
     mapResponse: MapResponseType<P, T>,
     responseOverride?: T | null,
-    errorOverride?: Error | null
+    errorOverride?: Error | null,
   ) {
     this.mapParamsToUrl = mapParamsToUrl
     this.mapParamsToBody = mapParamsToBody
--- a/api-client/src/EndpointBuilder.ts
+++ b/api-client/src/EndpointBuilder.ts
@@ -91,7 +91,7 @@ class EndpointBuilder<P, T> {
       this.paramsToBodyMapper,
       this.mapper,
       this.responseOverride,
-      this.errorOverride
+      this.errorOverride,
     )
   }
 }
--- a/api-client/src/Payload.ts
+++ b/api-client/src/Payload.ts
@@ -10,7 +10,7 @@ class Payload<T> {
     requestUrl: string | null | undefined = null,
     data: T | null | undefined = null,
     error: Error | null | undefined = null,
-    fetchDate: number = Date.now()
+    fetchDate: number = Date.now(),
   ) {
     this._isFetching = isFetching
     this._fetchDate = fetchDate
--- a/api-client/src/__tests__/Endpoint.spec.ts
+++ b/api-client/src/__tests__/Endpoint.spec.ts
@@ -147,7 +147,7 @@ describe('Endpoint', () => {
     mockedFetch.mockRejectedValue(error)
     const endpoint = new Endpoint('endpoint', defaultMapParamsToUrl, null, defaultJsonMapper)
     await expect(endpoint.request(params)).rejects.toThrow(
-      new FetchError({ endpointName: 'endpoint', innerError: error, url, requestOptions: { method: 'GET' } })
+      new FetchError({ endpointName: 'endpoint', innerError: error, url, requestOptions: { method: 'GET' } }),
     )
   })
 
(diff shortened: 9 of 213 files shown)


web/stylelint.config_css.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
   extends: ['stylelint-config-standard'],
   rules: {
-    indentation: 2,
```
LeandraH, line 4: The indentation is done automatically by prettier.

steffenkleinle (approved):
Looks good

.eslintrc.js
```diff
@@ -109,6 +109,7 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'error',
     '@typescript-eslint/prefer-ts-expect-error': 'error',
     '@typescript-eslint/switch-exhaustiveness-check': 'error',
+    '@typescript-eslint/no-non-null-assertion': 'off',
```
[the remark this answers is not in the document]
LeandraH: Because the linter was complaining about some 40 no-non-null assertions

api-client/src/endpoints/__tests__/createPOIsEndpoint.spec.ts
```diff
@@ -93,7 +93,7 @@ describe('pois', () => {
   }
   it('should map params to url', () => {
     expect(pois.mapParamsToUrl(params)).toBe(
-      'https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/locations/?on_map=1'
+      'https://integreat-api-url.de/augsburg/de/wp-json/extensions/v3/locations/?on_map=1',
```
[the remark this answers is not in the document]
LeandraH: Yes, sorry, that's the main change from the latest linting version, adding a trailing comma absolutely everywhere. I do also in general prefer a trailing comma because it makes changes cleaner, though I don't think most of these cases will ever get an additional argument.

.eslintrc.js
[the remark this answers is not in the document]
sarahsporck: Hmm. Personally I'd prefer to set this to error and add a comment to explicitly disable the rule for the corresponding line, as non-null-assertions **are** in fact bad practice in general.

web/stylelint.config.js
```diff
@@ -7,5 +7,6 @@ module.exports = {
     'function-no-unknown': null,
     'at-rule-no-unknown': null,
     'no-extra-semicolons': null,
+    'media-query-no-invalid': null,
```
sarahsporck, line 10: Nice 👍🏼

steffenkleinle (changes requested):
@LeandraH sorry but I don't like your changes. Changing `!` to `as string` or similar does exactly the same, a typecast that may or may not work. However, it makes the code harder to read. Instead, please just comment out the rule for the next line specifically or find other fixes. If you don't want to do it here, we can also do it in a separate ticket if you just revert to `warning`.

web/src/hooks/useCityContentParams.ts
```diff
@@ -2,7 +2,7 @@ import { useParams } from 'react-router-dom'
 
 const useCityContentParams = (): { cityCode: string; languageCode: string } => {
   const params = useParams()
-  return { cityCode: params.cityCode!, languageCode: params.languageCode! }
+  return { cityCode: params.cityCode as string, languageCode: params.languageCode as string }
```
steffenkleinle, line 5: Example:
```suggestion
  // This hook should only be used in city content routes where it is guaranteed that both city and language code are set
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return { cityCode: params.cityCode as string, languageCode: params.languageCode as string }
```

sarahsporck: I also prefer this solution. If you use the typecast it becomes intransparent, that there is a possibility that the type is not defined. @steffenkleinle I think you messed up the suggestion ^^ (as you still have the typecast)

```suggestion
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return { cityCode: params.cityCode as string, languageCode: params.languageCode! }
```

steffenkleinle: True 🙈 You also still have one typecast though haha @sarahsporck

LeandraH: 😂 I am working my way through the typecasts and will fix them, no need to fix the example further :D

steffenkleinle (approved):

build-configs/tools/manage.ts
```diff
@@ -6,6 +6,8 @@ import fs from 'fs'
 import loadBuildConfig, { PlatformType } from '../index'
 
 const loadBuildConfigAsKeyValue = (buildConfigName: string, platform: PlatformType, spaces = true, quotes = false) => {
+  // If no buildConfig is found, that is handled in the the loadBuildConfig function
```
SteffiStuffel, line 9: This comment should be removed, as 
1) it is not really true, if no buildConfig is found an error is thrown, that is not really "handled"
2) if loadBuildConfig is changed somewhere in the future this comment will get outdated as no one can know an adjustment of this comment at this file is needed.

native/src/components/CityNotCooperatingFooter.tsx
```diff
@@ -38,6 +38,8 @@ const CityNotCooperatingFooter = ({
   if (!buildConfig().featureFlags.cityNotCooperating) {
     return null
   }
+  // This footer is only displayed in the Integreat configuration, and that has the CityNotCooperatingIcon
```
SteffiStuffel, line 41: Please remove this comment, can easily become outdated, I think best solution here would be to add to the check above (line 38), that also the icon needs to be present.

native/src/components/OpeningHours.tsx
```diff
@@ -82,7 +82,9 @@ const OpeningHours = ({
         <Content>
           {openingHours.map((entry, index) => (
             <OpeningEntry
-              key={`${weekdays[index]!}-OpeningEntry`}
+              key={`${weekdays[index]}-OpeningEntry`}
+              // always supplied by Moment
```
SteffiStuffel, line 86: This comment will be out dated very soon, as we are about to remove Moment, also it is only supplied if the index is correct and not "always". Please remove the comment.

native/src/routes/CityNotCooperating.tsx
```diff
@@ -77,6 +77,8 @@ const CityNotCooperating = (): ReactElement => {
   const { t } = useTranslation('cityNotCooperating')
   const [isCopied, setIsCopied] = useState<boolean>(false)
   const theme = useTheme()
+  // Only displayed in the Integreat configuration, which has the template
```
SteffiStuffel, line 80: That is only true as long as we do not add this footer als e.g. for malte, the reason why this can be assumed not null is that it will be only rendered in the Navigator, if cityNotCooperating flag is set. Please also remove this comment.

web/src/components/OpeningHours.tsx
```diff
@@ -83,7 +83,9 @@ const OpeningHours = ({
       <Content>
         {openingHours.map((entry, index) => (
           <OpeningEntry
-            key={`${weekdays[index]!}-OpeningEntry`}
+            key={`${weekdays[index]}-OpeningEntry`}
+            // always supplied by Moment
```
SteffiStuffel, line 87: delete or adjust comment, see above

web/src/routes/CityNotCooperatingPage.tsx
```diff
@@ -81,6 +81,8 @@ type CityNotCooperatingPageProps = {
 const CityNotCooperatingPage = ({ languageCode }: CityNotCooperatingPageProps): ReactElement => {
   const { t } = useTranslation('cityNotCooperating')
   const [isCopied, setIsCopied] = useState<boolean>(false)
+  // Only displayed in the Integreat configuration, which has the template
```
SteffiStuffel, line 84: see comment above

SteffiStuffel (changes requested):
Thanks for your PR :+1: 

Unfortunately I am not happy with adding the "no-non-null-assertion", as we use these often in the code. I think we should either
1) add the rule AND make these assertions unnecessary by adding checks and throwing errors if the values are null. This would make the code more robust/stable but also adding more code where it is not "needed".
2) not add the rule and keep it as it is, this will keep the code shorter

But adding the rule, just to add a lot of "ignore this rule" comments seems to be the worst of both, as it does not make the code more stable, but longer / more unclear.

And for future refactoring PRs, it is way nicer to review if big changes like adding the trainling commas would be done in a seperate PR.

.eslintrc.js
```diff
@@ -109,6 +109,7 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'error',
     '@typescript-eslint/prefer-ts-expect-error': 'error',
     '@typescript-eslint/switch-exhaustiveness-check': 'error',
+    '@typescript-eslint/no-non-null-assertion': 'error',
```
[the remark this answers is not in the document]
LeandraH: Two reasons, one is that it was annoying me to get a pile of warnings whenever I run `yarn lint`, second is that I think it leads to better code :)

SteffiStuffel (approved):

PR #1003, opened 2023-08-03
IGAPP-1329: Add text button component

steffenkleinle:


steffenkleinle:
I tried to unify the styles of the different usages of our button and our designs which will obviously lead to some (minor) changes. I think consistency is important here. If you are not happy with the common style, feel free to propose different font size, padding, ...

native/src/components/__tests__/Failure.spec.tsx
```diff
@@ -13,21 +13,21 @@ describe('Failure', () => {
     render(<Failure code={code} buttonAction={tryAgain} />)
 
   it('should render a retry button if tryAgain is passed', () => {
-    const { getByTestId } = renderFailure(() => undefined)
-    expect(getByTestId('button-tryAgain')).toBeTruthy()
+    const { getByText } = renderFailure(() => undefined)
```
LeandraH, line 16: ❤️

web/src/components/FeedbackSearch.tsx
LeandraH, line 1: The padding of the tiles here is a little off.
image

LeandraH (approved):
Looks good, I think the buttons feel more coherent now. Tested in emulated iOS and Firefox.

native/src/components/TextButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+      return (
+        <TileButton active={props.active} onPress={onPress} style={style}>
+          {props.Icon}
+          <TileButtonText>{text}</TileButtonText>
+        </TileButton>
+      )
+
+    case 'overlay':
+      return (
+        <OverlayButton onPress={onPress} style={style}>
```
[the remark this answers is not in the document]
steffenkleinle: Hmm I think it will only be used in overlays (at least at the moment we don't have another use case for it). Therefore I find the name quite fitting. At least I like the wording `IconButton` or `TwoIconButton` or similar not at all bc it doesn't say any helpful information, e.g. tile buttons also have an icon. Therefore I would perhaps only rename once we figure out what the purpose of this button is?

Other opinions?

web/src/components/FeedbackSearch.tsx
steffenkleinle: Fixed, good catch! :+1:

native/src/components/TextButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+  | {
+      type: 'overlay'
+      Icon: ReactElement
+      closeButton?: boolean
+    }
+  | {
+      type: 'tile'
+      Icon: ReactElement
+      active: boolean
+    }
```
SteffiStuffel, line 84: Some thoughts on this:
1) For every type of button the option "disabled" would make sense, although not used yet for other types than "primary"
2) overlay buttons have this option "closeButton", I see this is used for the pois details, but this is somehow only Poi-logic and should not belong to this general component, something like "IconLeft" and "IconRight" would be more generic.

native/src/components/TextButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+type TextButtonProps = {
+  text: string
+  onPress: () => Promise<void> | void
+  style?: StyleProp<ViewStyle>
+} & ButtonSpecificProps
+
+const TextButton = ({ text, onPress, style, ...props }: TextButtonProps): ReactElement => {
```
SteffiStuffel, line 92: As this component does not share any code, this should be splitted to three components:
1) TextButton (for primary and clear)
2) OverlayButton
3) TileButton

Then all of them can have one matching type, avoiding these "or"-types and the switch-case. 
Then we can easily use at every point in the code
`<OverlayButton ... `
instead of
`<TextButton
          type='overlay'`
and auto complete feature of the IDE can do its job with adding needed.

These components can go to the `components/base` folder

SteffiStuffel (changes requested):
Why is the base branche IGAPP-1173?

native/src/components/TextButton.tsx
steffenkleinle: See #964, that was exactly what was critized there and I agree with the reviewers there. Therefore I now created one component.

native/src/components/TextButton.tsx
steffenkleinle: 1. Might make sense to have all kind of buttons disabled but I don't see the point in adding (atm) unnecessary code and testing it. In general our policy is to remove any unused code and logic, therefore as long as we don't need it I would also not invest time in it (and maintain it later).
2. I am okay with changing it, I just don't think we will use IconRight different then for close buttons. I don't really like the naming Right/Left bc of RTL layouts. But I guess there is no way around that. Other opinions?

native/src/components/TextButton.tsx
SteffiStuffel: Well you got two approvals for this PR. 

@f1sh1918 whats your opinion on my comment above? or also @LeandraH and @sarahsporck What do you think is the better way to do this?

native/src/components/TextButton.tsx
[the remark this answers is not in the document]
LeandraH: Well, it's used in one very specific overlay, and from the name it's not clear that those are buttons for the filters, so I guess I'm somewhat unhappy about that naming in general. I'm not sure why these buttons should only be used in overlays but since we don't have those yet, I'd call this a premature optimization.

I think for a component like this, which is meant to be reused in many places, we should use naming for the layout, not for the use cases.

native/src/components/TextButton.tsx
LeandraH: My first inclination is that a single button is more likely to be used than three buttons. At the same time, I guess it might not be clear that the tiles are text buttons, and if someone re-implements those, that would be very annoying. So really, I don't have a clear preference.

I also think (hope?) that this button will be an interim component until we have a design system, at which point it might be a clearer decision point.

steffenkleinle:
> Why is the base branche IGAPP-1173?

Because I introduced both tiles and overlay buttons there. ATM merging that PR so that the base branch is main :)

native/src/components/TextButton.tsx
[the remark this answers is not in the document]
LeandraH: Sounds great! Good point, it's really more the shape that's special. Let's go for elliptic, round sounds to like it should always be completely round.

steffenkleinle:
TODOs:
- [] Use custom pressable instead of touchable opacity

native/src/components/TextButton.tsx
sarahsporck: In general I prefer smaller and expendable components. Therefore, having one component manage multiple use cases makes me feel uneasy already. In my understanding there should be separate components for separate use cases. Not that this solution is not expandable, but I think it can bloat up very quickly. I'm also curious to hear @f1sh1918 oppinion on this.

f1sh1918: sorry for the late response. I agree with steffi. I think these variants should only differ in some styling differences to keep in simple an clear. I think like "primary/secondary/clear (text only)" and if needed pass a `iconStart` / `iconEnd` prop.
What i criticized on #964 should be solved with including the slide button into the `TextButton`
The other button types differ too much in their behavior and should be defined in separate components

native/src/components/TextButton.tsx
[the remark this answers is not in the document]
steffenkleinle: Renamed to chip button as that seems to be the name: https://mui.com/material-ui/react-chip/

native/src/components/TextButton.tsx
steffenkleinle: Thanks for the discussion! I like separate components more as well :)

SteffiStuffel (approved):
Lokks good, not tested again.

PR #1004, opened 2023-08-04
IGAPP-1366: Unify onPress behavior

steffenkleinle:


native/src/components/CategoryListItem.tsx
```diff
@@ -65,37 +64,33 @@ const CategoryListItem = ({
...
-          subCategory={subCategory}
-          resourceCache={resourceCache[subCategory.path] ?? {}}
-          onItemPress={onItemPress}
-          language={language}
-        />
-      ))}
-    </>
-  )
-}
+}: CategoryListItemProps): ReactElement => (
```
steffenkleinle, line 67: No changes here instead of indentation and missing destructuring

native/src/components/SubCategoryListItem.tsx
```diff
@@ -41,21 +40,15 @@ const SubCategoryListItem = ({
...
-      <SubCategoryTitleContainer language={language}>
-        {!!subCategory.thumbnail && (
-          <CategoryThumbnail language={language} source={getCachedThumbnail(subCategory, resourceCache)} />
-        )}
-        <SubCategoryTitle>{subCategory.title}</SubCategoryTitle>
-      </SubCategoryTitleContainer>
-    </FlexStyledLink>
-  )
-}
+}: SubCategoryListItemProps): ReactElement => (
```
steffenkleinle, line 43: No changes except indentation

steffenkleinle:
For poi detail components this is done in #1005

native/src/components/CategoryListItem.tsx
```diff
@@ -8,10 +8,9 @@ import dimensions from '../constants/dimensions'
 import { LanguageResourceCacheStateType } from '../utils/DataContainer'
 import { getCachedThumbnail } from './Categories'
 import SimpleImage from './SimpleImage'
-import StyledLink from './StyledLink'
 import SubCategoryListItem from './SubCategoryListItem'
 
-const FlexStyledLink = styled(StyledLink)`
+const FlexStyledLink = styled.TouchableOpacity`
   display: flex;
   flex-direction: column;
```
LeandraH, line 15:
```suggestion
  flex-direction: column;
  margin: 0 20px;
```

native/src/components/SubCategoryListItem.tsx
```diff
@@ -16,10 +15,10 @@ const SubCategoryTitleContainer = styled.View<{ language: string }>`
   flex-direction: ${props => contentDirection(props.language)};
 `
 
-const FlexStyledLink = styled.TouchableHighlight<{ language: string }>`
+const FlexStyledLink = styled.TouchableOpacity<{ language: string }>`
   display: flex;
   flex-direction: ${props => contentDirection(props.language)};
-  margin: 0 20px 0 60px;
+  margin: 0 0 0 24px;
```
LeandraH, line 21: The 60px here is necessary to keep the visual hierarchy.
margin 0 0 0 24px:
image

margin 0 20px 0 60px: 
image

LeandraH (approved):
Nice, just some small styling remarks :)

Tested in emulated iOS

native/src/components/CategoryListItem.tsx
LeandraH: Without the margin, the links stretch the entire screen.
image

SteffiStuffel (changes requested):
I disagree that we should go back from `Pressable` to `TouchableOpacity`, as Pressable is the preferred and future-proof way to handle clicks, as mentioned in the React Native docs.

sarahsporck:
I agree with @SteffiStuffel here. `Pressable` should be the way to go here.

steffenkleinle:
@SteffiStuffel @sarahsporck fixed, I now use a custom implementation of Pressable. I can adjust the pressed styling if you want.

native/src/components/CategoryListItem.tsx
steffenkleinle: I can't reproduce this, for me it looks like this:
< src="https://github.com/digitalfabrik/integreat-app/assets/29095487/6539025b-488d-4ba5-a762-8ae1021af435" width="400px" />

native/src/components/SubCategoryListItem.tsx
steffenkleinle: I also can't reproduce this, for me it looks like this (I adjusted the margins, so there is not that much indentation anymore):

LeandraH (approved):
Nice, tested in emulated iOS

native/src/components/CollapsibleItem.tsx
```diff
@@ -67,7 +67,7 @@ const CollapsibleItem = ({
       <Collapse
         isExpanded={isExpanded}
         onToggle={() => setIsExpanded(!isExpanded)}
-        touchableOpacityProps={{ activeOpacity: 1 }}>
+        PressableProps={{ activeOpacity: 1 }}>
```
SteffiStuffel, line 70: Where does this come from?

native/src/components/EastereggImage.tsx
```diff
@@ -17,6 +17,10 @@ const ApiUrlText = styled.Text`
   color: red;
 `
 
+const FullOpacityPressable = styled(Pressable)`
```
SteffiStuffel, line 20: Normally we name these styled props just "styledPressable", why have you decided to do it different here?

SteffiStuffel (commented):
1) Please rename the branch
2) There are some usages of pressable/touchable Buttons that do not use this new "Pressable" component, is this on purpose, e.g. SlideFooter-Buttons, CityNotCooperatingFooter, ExportEvent-Button....

native/src/components/EastereggImage.tsx
steffenkleinle: I thought styling in this case only has one sole purpose, i.e. preventing showing a feedback to the user in order not to indicate that it is pressable, so I thought naming this differently might add additional information. However, I guess that `FullOpacityPressable` doesn't really achieve that so I will just rename to `StyledPressable`.

native/src/components/CollapsibleItem.tsx
steffenkleinle: Good catch, just did search/replace all.

steffenkleinle:
> 1. Please rename the branch

@SteffiStuffel thought about this but then decided not to bc this closes the PR. Why do you think that is the better choice? Bc of the naming of the merge commit? I am totally okay with renaming :)
 
> 2. There are some usages of pressable/touchable Buttons that do not use this new "Pressable" component, is this on purpose, e.g. SlideFooter-Buttons, CityNotCooperatingFooter, ExportEvent-Button....

That are all usages of react-native-elements button component. Also good catch but will be done in #1003 as I refactored those already there (have to change from touchable opacity to pressable as well though).

SteffiStuffel:
> > 1. Please rename the branch
> 
> @SteffiStuffel thought about this but then decided not to bc this closes the PR. Why do you think that is the better choice? Bc of the naming of the merge commit? I am totally okay with renaming :)
> 

I thought this can cause confusion when later looking at the git history and the name of the branch is the opposite of what is done, but not that important though,

PR #1008, opened 2023-08-09
IGAPP-1254: Add transition for back navigation header on pois detail …

SteffiStuffel:
…view

Resolves: #2185


LeandraH (approved):
Very nice, except when I actually use the button to go back, it disappears without animation. Which I think is okay.

I would probably also speed up the animation a little, to maybe 1s but your call.

steffenkleinle (approved):

SteffiStuffel:
Changed it to 1s.

PR #1010, opened 2023-08-10
IGAPP-1366: Unify onPress behavior by using Pressable everywhere

steffenkleinle:

In favor of #1004, sorry you have to review again :/ Replacing react-native-elements button component will be done in #1003.


sarahsporck (approved):
Much better :) Tested on ios simulator.

PR #1011, opened 2023-08-10
IGAPP-1373: Normalize SVGs

SteffiStuffel:


steffenkleinle (approved):
Two icons are somehow off/two big here: 
image
http://localhost:9000/lkmuenchen/ar/offers/wohnraum
Maybe you could fix those?
Otherwise tested on firefox and looking good.

sarahsporck (approved):
Also tested on firefox. Nicely done! The wrongly sized icons @steffenkleinle mentioend seem to already have the wrong size in the live system. Still, it would be nice if you could fix them :)

SteffiStuffel:
Thanks for your reviews, the wrong size of the two icons is not related to the svgs, but due to a bug in our Tooltip-Component, I opend a ticket for this here: https://issues.tuerantuer.org/browse/IGAPP-1374

PR #1012, opened 2023-08-11
IGAPP-1371: Make poi feedback modal fullscreen

steffenkleinle:


web/src/components/Modal.tsx
```diff
@@ -21,9 +23,10 @@ const Overlay = styled.div`
   background-color: ${props => props.theme.colors.textSecondaryColor};
   opacity: 0.9;
 `
-const ModalContainer = styled.div<{ topPosition: number }>`
+
+const ModalContainer = styled.div`
   position: fixed;
-  top: ${props => props.topPosition}px;
+  top: 0;
```
steffenkleinle, line 29: I made the modal fullscreen now to avoid flickering if opening feedback after scrolling down (look at the header): 

https://github.com/digitalfabrik/integreat-app/assets/29095487/aeed619c-c20d-4405-b0a7-2671efbd7c84

web/src/components/Modal.tsx
```diff
@@ -44,71 +48,81 @@ const ModalContent = styled.div`
...
-const FeedbackModal = (props: FeedbackModalProps): ReactElement => {
+const Modal = ({ title, closeModal, children, direction, wrapInPortal = false }: ModalProps): ReactElement => {
   const { viewportSmall } = useWindowDimensions()
-  const { closeModal, topPosition = viewportSmall ? dimensions.headerHeightSmall : 0, ...otherProps } = props
-  const [isSubmitted, setIsSubmitted] = useState<boolean>(false)
-  const { t } = useTranslation('feedback')
-  const headline = isSubmitted ? `${t('thanksHeadline')}` : `${t('headline')}`
+  const { t } = useTranslation('common')
   useScrollToTop()
   useLockedBody(true)
```
steffenkleinle, line 85: @f1sh1918 not sure what this does, can you explain? Do we still need this?

LeandraH (approved):
Nice! I get a different font in the feedback in the BottomActionSheet than usual though, and only there.
image

steffenkleinle:
> Nice! I get a different font in the feedback in the BottomActionSheet than usual though, and only there

Fixed, good catch :)

web/src/components/SharingPopup.tsx
```diff
@@ -208,14 +209,14 @@ const SharingPopup = ({ shareUrl, title, flow, direction, isWithinPortal }: Shar
     <SharingPopupContainer>
       {shareOptionsVisible && (
         <>
-          {isWithinPortal && (
+          {portalNeeded && (
```
sarahsporck, line 212: Actually when do we need the portal? Why do we need the backdrop, when the poifeedback is fullscreen for the bottom action sheet?

sarahsporck (approved):
Tested in web. 👍🏼 Works fine.

web/src/components/SharingPopup.tsx
steffenkleinle: The backdrop is for the sharing popup, not the feedback. I dont know what exactly the reason is but the bottom action sheet is doing some weird stuff that breaks some other styling...

sarahsporck: Ah. You're right. Well then you're good to go :)

PR #1013, opened 2023-08-11
IGAPP-1377: Fix footer hiding zoom buttons

steffenkleinle:


web/src/components/MapView.tsx
```diff
@@ -60,6 +60,7 @@ const StyledIcon = styled(FontAwesomeIcon)<{ direction: string }>`
 const FooterContainer = styled.div`
   position: absolute;
   bottom: 0;
+  max-width: calc(100% - 50px);
```
sarahsporck, line 63: Not sure, whether I like the hard coded 50px, but also don't hava a much better idea :). So in my regards you can keep it.

sarahsporck (approved):
Tested in Firefox.

web/src/components/MapView.tsx
steffenkleinle: Me neither but as we don't even set the width of the buttons I figured its okayish...

LeandraH: Agreed, not the greatest but it's okay.

LeandraH (approved):
Looks good, tested in Firefox

PR #2302, opened 2023-08-14
Add issue and PR templates

sarahsporck:


.github/ISSUE_TEMPLATE/technical-task.md
```diff
@@ -0,0 +1,19 @@
+---
+name: Technical Task \U0001F4A1
```
SteffiStuffel, line 2: I think the name "technical Task" is a bit confusing, as this is a feature request isn't it? Or are we missing the feature request template?

sarahsporck: Feature === Technical Task. I copied this from the entitlementcard template tbh. I'm also fine with feature. Should we also rename the label (Task -> Feature) then?

SteffiStuffel: Oke, for the technical task sounds like a version update or a refactoring, but I am fine with this, if this is the correct term for a feature

LeandraH: I guess the feature request is then handled by the user story, and the technical task is actually more of a version update thing?

LeandraH (approved):

.github/ISSUE_TEMPLATE/technical-task.md
steffenkleinle: I also think the term `Technical Task` is a little misleading. What is the difference between a `User Story` and a `Technical Task` then if a `Technical Task` is a feature? I have the same definition in my head as @SteffiStuffel that a technical task is more refactoring/version updates/... that doesn't add a new "feature" or behavior

SteffiStuffel: Difference between User Story and Feature is normally that a User Story is bigger than a Feature. You need multiple features to achieve a User Story and multiple User Stories to achieve an epic. 
A User Story is also not necessarily in a good size for a development ticket.

steffenkleinle (approved):

sarahsporck:
We decided to go for the term `task` instead of `technical task`.

PR #2305, opened 2023-08-14
IGAPP-1350: Add offset to poi active symbol

### LINKED ISSUE #2275: IGAPP-1350: Poi Active Symbol feels offset
# sarahsporck - 18.7.2023, 12:08:34

As the active marker has the exact same position as the category symbol it feels as if it was offset. From a visual perspective the pointy thing of the marker should point to the coordinates.
This should be achievable by adding a icon offset when there is an active symbol. Or by changing the symbol so that it the pointy thing is located in the y-center in the actual svg.

**Environment**: -
**Linked issues:**

SteffiStuffel:

Resolves #2275

### DIFF
--- a/api-client/src/maps/index.ts
+++ b/api-client/src/maps/index.ts
@@ -37,6 +37,7 @@ export type MarkerConfig = {
   defaultSymbol: string
   symbolActive: string
   iconSize: number
+  offsetY?: number
 }
 
 type OpenStreetMapCopyrightType = {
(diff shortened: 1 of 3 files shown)


sarahsporck (approved):
Love it! Nice! Tested in Firefox, Android emulator and ios simulator.

steffenkleinle (approved):
Tested on android and web, way better :heart_eyes:

Issue #2314, opened 2023-08-14
App crashing because of invalid DateTime


steffenkleinle:
### Describe the Bug

Upon changing cities, the app crashes because of invalid date time.

### Steps to Reproduce

1. Click on change location
2. Select Kreis Bergstraße
3. See message:  ERROR  [Error: Invalid DateTime: unparsable: the input "undefined" can't be parsed as ISO 8601]

### Expected Behavior

It should be possible to switch cities and languages as much as you want without the app crashing.

### Actual Behavior

The app crashes if opening a new city.

### Additional Information

Probably introduced with #950.

PR #2317, opened 2023-08-14
2314: Remove unused and sometimes missing lastUpdate property

steffenkleinle:
### Short description

Remove the unused property lastUpdate of fetch results which was sometimes undefined and therefore crashing the app.

### Proposed changes

- Unused property lastUpdate sometimes returned by NativeFetcherModule is now ignored
- Unused property lastUpdate of resource cache was removed and is not saved to database connector anymore
- Removing the property and not paring it as DateTime avoids the app crashing

### Side effects

- There is no more lastUpdate timestamp for resource cache entries

### Resolved issues

Fixes: #2314 

---


sarahsporck (approved):
Nice! Tested on Android emulator and ios simulator.

Issue #2312, opened 2023-08-14
Replace font-awesome with custom icons


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
We only use font-awesome for a few icons anymore. This is bad in terms of consistency (own custom icons vs font-awesome icons) and also in terms of bundle size/maintainability/carbon footprint.

**Describe the solution you'd like**
font-awesome should be removed and all used icons replaced with custom icons made by our designers. Probably they can already be downloaded from figma.

**Describe alternatives you've considered**
None

**Additional context**
Relates to #2290, #1882, #2109.

steffenkleinle:
Fixed in #2329

PR #2329, opened 2023-08-21
1182: Move and unify assets

steffenkleinle:
### Short description

This PR unifies the icons used in web and native, removes usage of icon libraries and aligns the styling.

### Proposed changes

- All used icons are moved to assets/icons
- Icons from react-native-vector-icons and font-awesome are replaced with own icons
- A new Icon components are introduced both for web and native

### Side effects

- The icons may have changed a little (different paddings, sizes, looks)
- A few icons are now black and the color will be fixed in a separate PR: #2378 

### Resolved issues

Fixes: #1882, #2312 

### Note

**It is not necessary to test this PR on its own! Just test #2378 as I will merge them both together and otherwise you have to test all icons twice!**

---


web/src/assets/index.ts
```diff
@@ -1,76 +1,98 @@
...
+import MenuIcon from '../../../assets/icons/menu.svg'
 import NewsIcon from '../../../assets/icons/news.svg'
 import NoteIcon from '../../../assets/icons/note.svg'
 import OffersIcon from '../../../assets/icons/offers.svg'
 import PdfIcon from '../../../assets/icons/pdf.svg'
 import PetIcon from '../../../assets/icons/pet.svg'
 import PhoneIcon from '../../../assets/icons/phone.svg'
 import POIsIcon from '../../../assets/icons/pois.svg'
-import PositiveFeedbackIcon from '../../../assets/icons/positive-feedback.svg'
+import SadSmileyIcon from '../../../assets/icons/sad-smiley.svg'
```
SteffiStuffel, line 38: Just a thought: We are somehow inconsistent with the icon namings, sometimes we name them depending on their content (what can you see, e.g. SadSmileIcon) and sometimes on their usage (e.g. SearchIcon).

native/src/components/CustomHeaderButtons.tsx
```diff
@@ -1,36 +1,27 @@
...
-  const { name } = props
-  if (name === 'search') {
-    return <SearchIcon {...props} />
-  }
-  if (name === 'language') {
-    return <LanguageIcon scaleX={isRTL() ? -1 : 1} {...props} />
-  }
-  return <MaterialIcon {...props} name={name} />
-}
+const HeaderIcon = ({ name, ...props }: { name: string; style: StyleProp<SVGImageElement> }): ReactElement => (
```
SteffiStuffel, line 15: Does this really what it should do? Now LanguageIcon is somehow the fallback. I think it would be more stable code, if we have a proper fallback.

SteffiStuffel (approved):
Not tested but built, web size decreased from 2.58MB to 2.49MB :tada:

web/src/assets/index.ts
steffenkleinle: Do you have any preference? Someone else?

native/src/components/SettingItem.tsx
```diff
@@ -69,7 +69,6 @@ const SettingItem = (props: SettingItemProps): ReactElement => {
               }}>
               <Badge status={value ? 'success' : 'error'} />
               <Text> {value ? t('enabled') : t('disabled')}</Text>
-              <Icon name='chevron-right' />
```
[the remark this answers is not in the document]
steffenkleinle: It looked like a little overkill/the settings item quite crowded. I can definitely readd it though.

native/src/components/CustomHeaderButtons.tsx
steffenkleinle: Hmmm. In general true, just not sure what to use as a fallback and whether it actually makes sense to just introduce a new fallback icon that is then never used. I think it should be quite easy to spot if a new header icon would be added that the wrong icon is rendered. Open for suggestions here though.

steffenkleinle: We could also add a throw Error block if the name isn't either language or search

web/src/assets/index.ts
SteffiStuffel: No preference, but consistency would be nice, currently we mostly use naming on their usage (semantic). Pros and cons are for semantic naming are:
pros: - easy to change a icon, if the design is updated (no renaming needed)
cons: - use icon for multiple use cases will break the naming convention, e.g. when we use the sad smile for giving negative feedback and also for no search results
As the pro will not happen that often and a renaming is not really hard as we only import the file in the index file, i have a tendency to use naming on the content (syntactical), this is also how we do this in Lunes.

But as most icons are named semantically, maybe it is to much overhead for this ticket? If you do not want to do it here open a ticket or let me know and i will open one.

native/src/components/CustomHeaderButtons.tsx
SteffiStuffel: Yes I think this would be a better solution :+1:

web/src/components/HeaderActionItemLink.tsx
```diff
@@ -35,22 +33,21 @@ type HeaderActionItemLinkProps = {
...
  * Designed to work with Header. In the ActionBar you can display icons as link or dropDown involving actions like
  * 'Change language', 'Change location' and similar items.
  */
-const HeaderActionItemLink = ({ href, text, iconSrc, direction }: HeaderActionItemLinkProps): ReactElement => (
+const HeaderActionItemLink = ({ href, text, iconSrc }: HeaderActionItemLinkProps): ReactElement => (
   <Tooltip text={text} flow='down' smallViewportFlow='left'>
     {href ? (
       <StyledLink to={href} aria-label={text}>
-        <IconWithUiDirection alt='' src={iconSrc} direction={direction} />
+        <Icon src={iconSrc} />
```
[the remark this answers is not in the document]
SteffiStuffel: ah i see, thanks for explanation. Was confused because previously it was "WithUIDirection".

web/src/assets/index.ts
steffenkleinle: Lets do it in another PR. I am also in favor of syntactically.

PR #2375, opened 2023-08-22
2247: Move geocontrol with bottom action sheet

sarahsporck:
### Short description

The geocontrol icon should move together with the `bottomActionSheet`. The previous behavior had it jumping, each time you moved the sheet into a new position

### Proposed changes

- Extract geocontroljsx into geocontrol object and move it into the map
- Create a way to add a new "Position" for the map (where it can put controls into)
- create handles and a mapViewRef 

### Side effects

Hopefully not.

### Resolved issues

Fixes: #2247 

---


web/src/components/PoisMobile.tsx
```diff
@@ -163,24 +177,14 @@ const PoisMobile = ({
...
-          position='bottom-right'
-        />
       </MapView>
       <BottomActionSheet
         title={!currentFeatureOnMap ? t('listTitle') : undefined}
         toolbar={toolbar}
         ref={sheetRef}
         setBottomActionSheetHeight={setBottomActionSheetHeight}
         direction={direction}>
+        <div ref={geolocatePosition} style={{ position: 'absolute', right: 10, top: -40 }} />
```
steffenkleinle, line 187: Please use a styled-component. Also, this seems like it won't look correct for RTL

web/src/components/PoisDesktop.tsx
```diff
@@ -142,6 +144,16 @@ const PoisDesktop = ({
     }
   }, [currentFeatureOnMap, currentPoi, scrollOffset])
 
+  useEffect(() => {
+    if (mapViewRef) {
+      const geolocate = new GeolocateControl({
+        positionOptions: { enableHighAccuracy: true },
+        trackUserLocation: true,
+      })
+      mapViewRef.setGeocontrol('bottom-right', geolocate)
```
steffenkleinle, line 153: RTL

steffenkleinle (commented):
I am not really sure whether I like it that the geocontrol moves the whole height with the bottom action sheet, imo that looks a little weird. For comparison, I think it is solved quite nicely in google maps where the geocontrol is only visible in the smallest bottom sheet snappoint and slowly disappears behind the bottom action sheet if pulling up the bottom action sheet. Do you think something like that would be feasible?

Otherwise, I would also be okay with just displaying the control on one fixed point (above smallest or middle snappoint) and not changing the position at all.

However, obviously I am also okay with your solution. Nice work :+1:

web/src/components/PoisDesktop.tsx
sarahsporck: This is how it was implemented previously ^^. Should we really mirror the map control icons for rtl languages?

steffenkleinle: True, didn't see that, then feel free to ignore. I think we shouldnt mirror the map control icons (no effect anyway since symmetric) but yes, I think we should move them to the left side. Especially bc the zoom buttons already are on the left side

sarahsporck:
> 



> I am not really sure whether I like it that the geocontrol moves the whole height with the bottom action sheet, imo that looks a little weird. For comparison, I think it is solved quite nicely in google maps where the geocontrol is only visible in the smallest bottom sheet snappoint and slowly disappears behind the bottom action sheet if pulling up the bottom action sheet. Do you think something like that would be feasible?

I do agree with you, but the previous solution
I think this is very unintuitive, as the bottom action sheet is initialized at the middle snappoint and users might not move it, e.g. when the screen has a sufficient size.
 
> Otherwise, I would also be okay with just displaying the control on one fixed point (above smallest or middle snappoint) and not changing the position at all.

Hmm. not sure if I like it that the geolocate symbol is stuck in the middle on the map...

steffenkleinle (approved):

LeandraH (commented):
Unfortunately, I found two problems:
1. In Firefox with a small viewport, if I drag the BottomSheet all the way up (so that the geolocate button disappears) and then drag it back down, the button doesn't reappear. It only reappears when I drag the BottomSheet all the way to the bottom of the screen.
2. In emulated iOS on Safari, the geolocate button is just missing.

web/src/components/PoisMobile.tsx
```diff
@@ -163,24 +191,18 @@ const PoisMobile = ({
...
       <BottomActionSheet
         title={!currentFeatureOnMap ? t('listTitle') : undefined}
         toolbar={toolbar}
         ref={sheetRef}
         setBottomActionSheetHeight={setBottomActionSheetHeight}
         direction={direction}>
+        <GeoLocateContainer
+          id='geolocate'
+          ref={geolocatePosition}
+          isLastSnapPoint={bottomActionSheetHeight >= getSnapPoints(height)[2]}
```
f1sh1918, line 204:
```suggestion
          isLastSnapPoint={bottomActionSheetHeight > getSnapPoints(height)[2]}
```
I'm fine if we don't want to show the geocontrol on the last two snapPoints (as you implemented) but then the wording is misleading. So either rename the variable if the last two snapPoints are affected or change it to my suggestion

web/src/components/PoisMobile.tsx
```diff
@@ -65,6 +65,20 @@ const StyledIcon = styled(FontAwesomeIcon)<{ direction: string }>`
     `};
 `
 
+const GeoLocateContainer = styled.div<{ isLastSnapPoint: boolean }>`
+  position: absolute;
+  right: 10px;
+  top: -40px;
```
f1sh1918, line 71: i think you should use a const for the `40` and reuse it in the transform transform translate

f1sh1918 (approved):
Nicely done :) just some minor comments

sarahsporck:
> In Firefox with a small viewport, if I drag the BottomSheet all the way up (so that the geolocate button disappears) and then drag it back down, the button doesn't reappear. It only reappears when I drag the BottomSheet all the way to the bottom of the screen.

hmm. cannot reproduce. could you test this again? :)

> In emulated iOS on Safari, the geolocate button is just missing.

this I can also not reproduce... 🤷🏼‍♀️

LeandraH:
Ooops, sorry, I seem to have been a bit impatient with the animation, the icon does show back up in Firefox. But not in mobile Safari.

simulator_screenshot_762EC044-1C18-4BAC-A08E-612D8CF714E6

LeandraH (approved):
Very nice solution!

web/src/components/PoisDesktop.tsx
```diff
@@ -181,7 +194,6 @@ const PoisDesktop = ({
         languageCode={languageCode}>
         <NavigationControl showCompass={false} position={direction === 'rtl' ? 'bottom-left' : 'bottom-right'} />
         {/* To use geolocation in a development build you have to start the dev server with "yarn start --https" */}
```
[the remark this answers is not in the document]
steffenkleinle: For me it works without. We could then also remove the logging of that message in PoisPage (which is annoying me quite a while already xD).

web/src/utils/getSnapPoints.ts
```diff
@@ -8,7 +8,7 @@ const mapIconsHeight = 60
 const getLargeSnapPoint = (maxHeight: number) =>
   maxHeight - dimensions.headerHeightLarge - dimensions.navigationMenuHeight - mapIconsHeight
 
-export const getSnapPoints = (maxHeight: number): number[] => [
+export const getSnapPoints = (maxHeight: number): [number, number, number, number] => [
```
steffenkleinle, line 11: Could we now remove the non-null assertions for getSnapPoins?

web/src/components/PoisMobile.tsx
```diff
@@ -158,24 +188,14 @@ const PoisMobile = ({
...
-        />
       </MapView>
       <BottomActionSheet
         title={!currentFeatureOnMap ? t('listTitle') : undefined}
         toolbar={toolbar}
         ref={sheetRef}
         setBottomActionSheetHeight={setBottomActionSheetHeight}
-        direction={direction}>
+        direction={direction}
+        sibling={<GeoLocateContainer id='geolocate' direction={direction} ref={geolocatePosition} height={height} />}>
```
steffenkleinle, line 198: Can we just use one spelling? I found at least `GeoLocate`, `geolocate` and `geolocator`. I think we should stick to the one used by the library, so `Geolocate`/`geolocate`.

steffenkleinle (approved):
Love it, perfect! :heart: Tested on firefox :)

web/src/utils/getSnapPoints.ts
sarahsporck: sadly not all of them, as `react-map-gl` uses `number[]` as a type.

Issue #2382, opened 2023-08-23
Old and expired events shown in the App


vthejay:
### Describe the Bug

Events that are old and have expired are shown in the App.
### Steps to Reproduce

1. Go to 'Kreis Olpe'
2. Click on 'Events'
3. See old and expired events.

### Expected Behavior

Only actaual and future events should be shown, not the expired and old events
### Actual Behavior

Expired and old events are shown in the App
### Additional Information

(https://integreat.app/kreisolpe/de/events)
<details>
  <summary>Traceback</summary>

```

```

</details>

PR #2383, opened 2023-08-23
2382: Hide expired events

SteffiStuffel:
### Resolved issues

Fixes: #2382 

---

### DIFF
--- a/api-client/src/endpoints/createEventsEndpoint.ts
+++ b/api-client/src/endpoints/createEventsEndpoint.ts
@@ -133,11 +133,10 @@ const createRecurringEvents = (event: JsonEventType): JsonEventType[] => {
     event: {
       ...event.event,
       start: date.toISOString(),
-      start_date: dateToString(date),
       end: getEndDate(event, date).toISOString(),
-      end_date: dateToString(getEndDate(event, date)),
     },
   }))
+  //console.log(events[0].event)
   return events.slice(0, MAX_NUMBERS_OF_RECURRING_EVENTS)
 }
 
--- a/api-client/src/endpoints/createEventsEndpoint.ts
+++ b/api-client/src/endpoints/createEventsEndpoint.ts
@@ -87,8 +87,15 @@ const mapJsonToEvent = (event: JsonEventType): EventModel => {
 const getEndDate = (event: JsonEventType, startDate: Date): Date => {
   const start = DateTime.fromISO(event.event.start)
   const end = DateTime.fromISO(event.event.end)
-  const durationInDays = end.diff(start, 'days').days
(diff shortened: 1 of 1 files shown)


api-client/src/endpoints/createEventsEndpoint.ts
```diff
@@ -125,7 +132,9 @@ const createRecurringEvents = (event: JsonEventType): JsonEventType[] => {
     })),
     event: {
       ...event.event,
+      start: date.toISOString(),
```
LeandraH, line 135: Do we really need `start` and `start_date` both?

api-client/src/endpoints/createEventsEndpoint.ts
```diff
@@ -125,7 +132,9 @@ const createRecurringEvents = (event: JsonEventType): JsonEventType[] => {
...
       start_date: dateToString(date),
+      end: getEndDate(event, date).toISOString(),
```
LeandraH, line 137: Same for `end` and `end_date`?

LeandraH (approved):
Nice, tested in Firefox

api-client/src/endpoints/createEventsEndpoint.ts
SteffiStuffel: We do not really "need" both, but as this function updates these values for recurring events, i think it makes sense and the code more robust if both values are set correctly.

LeandraH: Maybe add a comment explaining that? I definitely see myself at some point attempting to remove one of these 😅

f1sh1918: Sorry also to ask. I didn't get the purpose yet for it. Where do we want to use `start_date`. Even these attributes are not listed in `JsonEventType[]`

steffenkleinle: Same, we shouldn't need both :see_no_evil:

api-client/src/endpoints/createEventsEndpoint.ts
```diff
@@ -125,10 +132,11 @@ const createRecurringEvents = (event: JsonEventType): JsonEventType[] => {
     })),
     event: {
       ...event.event,
-      start_date: dateToString(date),
-      end_date: dateToString(getEndDate(event, date)),
+      start: date.toISOString(),
+      end: getEndDate(event, date).toISOString(),
     },
   }))
+  //console.log(events[0].event)
```
steffenkleinle, line 139:
```suggestion
```

steffenkleinle (commented):
Tested on firefox

Issue #2380, opened 2023-08-22
Poi header title not translated


steffenkleinle:
### Describe the Bug

The title if the last route was the map is `locations` and not translated.

### Steps to Reproduce

1. Go to testumgebung/de/locations
2. Go to feedback
3. Look at header title `locations`

### Expected Behavior

The header title should be translated and be `Map`

### Actual Behavior

The header title is not translated and is `locations`.

### Additional Information

Check if all other cases are handled.

PR #2390, opened 2023-08-24
2380: Poi header title not translated

f1sh1918:
### Short description

We are using the route names as keys for the translations of the header title except for the pois route that has `locations` as route name but `pois` as key.

### Proposed changes

Unfortunately we use pois and locations during the project. Since we can not change the route 'locations' because all existing links would be invalid, i had to adjust the key in translations for layout section.

Fixes: #2380


LeandraH (approved):
Nice! Tested in emulated iOS

release-notes/unreleased/2380-poi-header-title-not-translated.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2380
+show_in_stores: false
+platforms:
+  - native
```
steffenkleinle, line 4: This is not a valid platform!
```suggestion
  - android
  - ios
```

steffenkleinle (commented):
I think we should probably use the same key `locations` everywhere then and get rid of `pois` key, otherwise it will just be confusing...

steffenkleinle (approved):

Issue #2377, opened 2023-08-22
POI filter "Other" not working


Docmale:
### Describe the Bug

POI filter "Other" is not working  as other filters do. Is this a mixture of category "Other" and no category set?
1. tapping "Other" dosn't highlite the button but sets the filter, tapping again dosn't clears the filter
2. All POIs with no category set are shown as "Other" but are not found by the filter "Other"

### Steps to Reproduce

1. Go to "Testumgebung" - Map
2. Click on 'Other' than 'Orte anzeigen'
3. and so on

### Expected Behavior

Either introduce category "No categorie" or set all POIs wit no category set as "Other"

### Actual Behavior

see description

sarahsporck:
@Docmale could you add the correct label here (web/native)? :) And if it is native and android or ios specific there are also labels for the corresponing platform

Docmale:
Havn't found any possibility to filter on web.

sarahsporck:
Just wanted to enforce good patterns when creating issues :).

Docmale:
POIs without a category are now shown up as "other"

PR #2395, opened 2023-08-28
2377: Fix other category

steffenkleinle:
### Short description

This PR fixes the selection of the `Other` category in the poi filters.

### Proposed changes

- Use correct check instead of `===`

### Resolved issues

Fixes: #2377

---


f1sh1918 (approved):
tested on ios

Issue #2398, opened 2023-08-29
Text in native cut on right screen edge


osmers:
### Describe the Bug

The text in Malte and Integreat is cut at the right edge of the screen. 

### Steps to Reproduce

1. Go to any region on native
2. Navigate to a page with text
3. See text cut

### Expected Behavior

The text is wrapped at the end of the line/screen and fits nicely on it.

### Actual Behavior

It gets cut and users cannot read the full text

### Additional Information

as an example
![image](https://github.com/digitalfabrik/integreat-app/assets/64901227/edde7ce0-4335-45ff-8e4c-42cd99e0db75)

PR #2399, opened 2023-08-29
2398: Fix cut off categories content

ztefanie:
### Short description
Fix the bug

### Proposed changes

- remove width of RemoteContent 

Other option would be to change padding of Page to margin, but I think this does make more sense.

### Side effects

hopefully none

### Resolved issues

Fixes: #2398

---


f1sh1918 (commented):
Thx for the fast fix :)
Tested on ios works fine.

I wonder whether why this was needed. maybe for old webview version

Could you please also remove the unneeded code like  webViewWidth props and the unneeded functions and props on page.tsx level

f1sh1918 (approved):

steffenkleinle:
I'll going to check in a few minutes whether I find out why we needed that. Pls dont merge before.

Issue #2387, opened 2023-08-24
Remove jest warning regarding update to SimpleImage


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Since we changed to the svg transformer the native images in our tests aren't mocked anymore. Therefore there are a bunch of errors although these errors probably don't have any impact on the result this should be fixed.

<details>
<summary>Stacktrace</summary>
  console.error
    Warning: An update to SimpleImage inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at SimpleImage (/Users/digitalfabrik/projects/integreat-app/native/src/components/SimpleImage.tsx:28:3)
        at StyledNativeComponent (/Users/digitalfabrik/projects/integreat-app/native/node_modules/styled-components/native/dist/styled-components.native.cjs.js:5949:29)
        at Styled(SimpleImage)
        at View
        at Component (/Users/digitalfabrik/projects/integreat-app/native/node_modules/react-native/jest/mockComponent.js:28:18)
        at StyledNativeComponent (/Users/digitalfabrik/projects/integreat-app/native/node_modules/styled-components/native/dist/styled-components.native.cjs.js:5949:29)
        at Styled(View)
        at PoiDetails (/Users/digitalfabrik/projects/integreat-app/native/src/components/PoiDetails.tsx:54:23)
        at useContext (/Users/digitalfabrik/projects/integreat-app/native/node_modules/styled-components/src/models/ThemeProvider.js:48:22)
        at wrapWithTheme (/Users/digitalfabrik/projects/integreat-app/native/src/testing/wrapWithTheme.tsx:6:26)
</details>

**Describe the solution you'd like**
Mock SimpleImage where the error occurs or wrap in act where needed. 

**Describe alternatives you've considered**
Mock svgs in tests. But I'm unsure whether this actually fixes the issue.

**Additional context**
N/A

steffenkleinle:
I think we only use simpleimage for non-svgs? Not sure though. Possibly related to #1882 and #2378.

sarahsporck:
Maybe. I'll take a look at this, as soon I've got some time on my hands. Don't like it if tests are 🔴red

PR #2400, opened 2023-08-29
2387: Remove jest errors

sarahsporck:
2387: add mock for react-native-svg

### Short description

Fix errors in tests.

### Proposed changes

- Add mock for for SVG Uri
- Add missing react-i18next mock

### Resolved issues

Fixes: #2387


native/src/components/SimpleImage.tsx
```diff
@@ -30,12 +30,14 @@ const SimpleImage = ({
   resizeMode = 'contain',
   specifyAspectRatio = false,
 }: SimpleImageProps): ReactElement => {
-  const [aspectRatio, setAspectRatio] = useState<number | undefined>(undefined)
-
-  useEffect(() => {
+  const aspectRatio = useMemo(() => {
```
sarahsporck, line 33: `useMemo` is resolved before the first render while `useEffect` is only run after the first render. This leads to weird test behavior. Also useMemo seems to be the right choice here, as we just assign a value.

steffenkleinle (approved):
Looks good :+1: Tested on android

steffenkleinle:
@sarahsporck do you think we could not log error messages from console.error in the console for the tests? Like sometimes we are testing what happens if an error is thrown and that clutters the logs. Not related to the issue though.

sarahsporck:
> @sarahsporck do you think we could not log error messages from console.error in the console for the tests? Like sometimes we are testing what happens if an error is thrown and that clutters the logs. Not related to the issue though.

What a throwback to one of my first tickets at digitalfabrik -> https://github.com/digitalfabrik/integreat-webapp/pull/366
Not sure why we removed the workaround completely though. I also don't see a reason to not override `console.error` for tests. Feel free to create a ticket :)

PR #2393, opened 2023-08-28
2115: Add poi filters web

steffenkleinle:
### Short description

This PR adds the poi filters for web mobile and desktop.

### Proposed changes

- Add an overlay to the MapView containing the filter buttons
- Show the poi filters as full screen modal on mobile and in the side panel on web

### Side effects

- Hopefully none

### Resolved issues

Fixes: #2115 

---

PR #2408, opened 2023-08-29
2115: Ensure unique checkbox id

steffenkleinle:
### Short description

Ensure that no checkbox/label id is used twice not just assuming the label will be unique.

### Proposed changes

- Pass an id instead of just using the label as id.

### Additional infos

Related to #2393.

---


LeandraH (approved):
Nice!

f1sh1918 (approved):
tested on chrome

PR #2418, opened 2023-08-31
2246: Query params in deep links

LeandraH:
### Short description

We have not been using query parameters in deep links, this implements them for search parameters and finishes implementing them for multipois.

### Proposed changes

This parses the query parameter into an object in the api-client and passes those into the search and pois routes. They are then used in native for opening the correct multipoi or showing the search key that was used.

Also, deep linking in iOS only worked correctly if the app wasn't open in the background, it would pop the app back into the foreground but not change to the screen of deep link. That is now fixed.

This can be tested on iOS using the command `npx uri-scheme open https://integreat.app/testumgebung/de/search?query=test --ios`. I have not been able to test locally on Android but the deep links work there in production so I assume it should also work with some extra parameters.

On a side note, we currently don't actually have the option of sharing from the search screen in native since the search bar takes over the header. I think that's fine but makes the use case of this PR a little rarer.

### Side effects

This also passes the parameters to web but I have not seen any problems there.

### Resolved issues

Fixes: #2246 

---


api-client/src/routes/InternalPathnameParser.ts
```diff
@@ -57,6 +59,19 @@ class InternalPathnameParser {
...
+    const queryObject: Record<string, string> = {}
+    queryItems?.forEach(query => {
+      const [key, value] = query.split('=')
+      if (!key || !value) {
+        return
+      }
+      queryObject[key] = value
+    })
+    return queryObject
+  }
```
sarahsporck, line 73:
```suggestion
  query = (): URLSearchParams | undefined => {
    return this._query ? new URLSearchParams(this._query) : undefined
  }
```

native/ios/Integreat/AppDelegate.mm
```diff
@@ -63,6 +64,22 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
...
+}
+
+- (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull NSUserActivity *)userActivity
+ restorationHandler:(nonnull void (^)(NSArray<id<UIUserActivityRestoring>> * _Nullable))restorationHandler
+{
+ return [RCTLinkingManager application:application
+                  continueUserActivity:userActivity
+                    restorationHandler:restorationHandler];
+}
+
```
sarahsporck, line 82: Did deep linking not work on ios beforehand? whats this change about?

api-client/src/routes/InternalPathnameParser.ts
sarahsporck: This way you also don't have to remove the `?` at the beginning.

native/ios/Integreat/AppDelegate.mm
LeandraH: Correct, in iOS it only worked if the app wasn't open in the background. So, two cases:
1. The app is fully closed, you tap on a deep link somewhere, the app opens and sends you to the correct link.
2. The app is open in the background, you tap on a deep link somewhere, the app opens on the page that you were last on and the deep link is otherwise ignored.

native/src/components/Header.tsx
```diff
@@ -179,7 +179,11 @@ const Header = ({
 
   const visible = showItems && !!goToLanguageChange
   const items = [
-    renderItem(HeaderButtonTitle.Search, 'search', visible, () => navigation.navigate(SEARCH_ROUTE)),
+    renderItem(HeaderButtonTitle.Search, 'search', visible, () =>
+      navigation.navigate(SEARCH_ROUTE, {
+        searchText: '',
```
SteffiStuffel, line 184: I think it is cleaner and less error prone, to set the default in SearchModalContainer, as it is allowed the searchText is null or undefined the component should handle these cases.
`          initialSearchText={route.params.searchText ?? ''}`

native/src/routes/SearchModal.tsx
```diff
@@ -53,8 +54,9 @@ const SearchModal = ({
   cityCode,
   closeModal,
   t,
+  initialSearchText = '',
 }: SearchModalProps): ReactElement => {
-  const [query, setQuery] = useState<string>('')
+  const [query, setQuery] = useState<string>(initialSearchText ?? '')
```
SteffiStuffel, line 59: Setting the default twice is unnecessary. Also if you apply my changes above, by setting the `initalSearchText` in the SearchModalContainer, you do not need to do it here and can adjust the type of the prop to `initialSearchText: string` as neigher `undefined` nor `null` is possible

SteffiStuffel (approved):
Nice :tada: 
Tested on Android. Works fine  :+1:

native/ios/Integreat/AppDelegate.mm
sarahsporck: Interesting. Thank you for the information.

sarahsporck (approved):
Tested on ios simulator with: 
`xcrun simctl openurl booted 'integreat://integreat.app/testumgebung/de/search?query=abc'`. 
Only works when you deep link in the same language. Otherwise the browser is openend. But I guess that is alright :)

LeandraH:
I think the language thing is on purpose, since we also only send to the landing screen if someone has a different municipality open. But I guess that could be a little confusing? I'll open an issue for discussing it.

PR #2421, opened 2023-09-04
2293: ease scrolling in bottom action sheet

sarahsporck:
### Short description

When the bottom action sheet is full screen scrolling should now work as intended. I couldn't figure out how to improve the behavior when it is not full screen.

### Proposed changes

- Deactivated panning gestures on the content, when the bottom action sheet is full screen.

### Side effects

- I didn't test the behavior on ios, but it should not be affected

### Resolved issues

Fixes: #2293

---


SteffiStuffel (approved):
Tested on Android emulator.

f1sh1918 (approved):
nice find. couldn't find any issue neither android nor ios

PR #2422, opened 2023-09-04
2014: Update i18next

LeandraH:
### Short description

Updating `i18next` and `react-i18next` to the latest versions.

### Proposed changes

- The type TFunction is now exported by `i18next` instead of `react-i18next`.
- We can now use the new pluralisation from `i18next`, hence the small changes in `ShelterInformation`

### Side effects

- We are polyfilling the new Intl.Pluralization rules in native.
-

### Resolved issues

Fixes: #2014 

---


web/src/components/__tests__/Failure.spec.tsx
```diff
@@ -7,7 +8,7 @@ jest.mock('react-i18next')
 jest.mock('react-inlinesvg')
 
 describe('Failure', () => {
-  const mockTranslate = (key: string) => key
+  const mockTranslate = ((key: string) => key) as TFunction
```
f1sh1918, line 11: for me it works without casting

f1sh1918 (approved):
Looks fine.
I just couldn't figure out why you need the casting `as TFunction` for me i had no issues removing it everywhere

web/src/components/__tests__/Failure.spec.tsx
LeandraH: I needed it in `yarn ts:check`, did you run that one?

native/src/routes/__tests__/SearchModal.spec.tsx
```diff
@@ -25,7 +26,7 @@ describe('SearchModal', () => {
   })
   const dummy = jest.fn()
 
-  const t = (key: string) => key
+  const t = ((key: string) => key) as TFunction
```
steffenkleinle, line 29: Optional: Otherwise we could also just not pass tFunction as prop but use the hook instead

release-notes/unreleased/2014-update-i18next.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2014
+show_in_stores: false
+platforms:
+  - web
+  - android
+  - ios
+en: Updating dependencies
```
steffenkleinle, line 7: Maybe we should discuss release notes again. I think a release note like this does not add any benefit as there is no user facing changes/nothing testable :see_no_evil:

web/src/components/Failure.tsx
```diff
@@ -23,7 +23,7 @@ type FailureProps = {
   errorMessage: string
   goToPath?: string
   goToMessage?: string
-  t: TFunction<'error'>
+  t: TFunction
```
steffenkleinle, line 26: I think this is a little dangerous, I think we should just not pass it as a prop but instead use the hook again.

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -11,7 +11,7 @@ import Selector from './Selector'
 type HeaderLanguageSelectorItemProps = {
   selectorItems: Array<SelectorItemModel>
   activeItemCode: string
-  t: TFunction<'layout'>
+  t: TFunction
```
steffenkleinle, line 14: Same as above.

release-notes/unreleased/2014-update-i18next.yml
LeandraH: I'm okay with not adding one 🤷

steffenkleinle (approved):

Issue #2331, opened 2023-08-21
Remove react-native-elements and react-native-vector-icons


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
Since we are now using our own button and icons components, react-native-elements and react-native-vector-icons libraries are mostly redundant and not used anymore. Therefore we could remove those (react-native-vector-icons is required by react-native-elements).

**Describe the solution you'd like**
Replace all left usages of react-native-elements with custom components and remove the two libraries.

**Describe alternatives you've considered**
None.

**Additional context**
Related: #1882, #2312, #2329.

PR #2432, opened 2023-09-07
2331: Remove rn-elements and rn-vector-icons

LeandraH:
### Short description

This PR removes the two packages `react-native-elements` and `react-native-vector-icons`, and adds a round `IconButton` which I already needed twice in this PR.

### Proposed changes

There were four places where we still used `react-native-elements`:

- The icon on the map to center on the current location: added the specific icons we use and used those with the new `IconButton`
- The icon to request the current location on the screen where you can choose the municipality: replaced their button with our `IconButton`
- The badges in the settings screen: replaced those with round `Views`
- The `ThemedSearchBar`: wrote our own with a `TextInput` from `react-native`

### Side effects

- Should be nothing

### Resolved issues

Fixes: #2331 

---


native/src/components/base/IconButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  height: ${props => props.size}px;
+  width: ${props => props.size}px;
+  align-items: center;
+  justify-content: center;
+  border-radius: ${props => props.size / 2}px;
+  background-color: ${props => props.theme.colors.backgroundColor};
+`
+
+type IconButtonProps = {
+  accessibilityLabel?: string
```
SteffiStuffel, line 17: I think accessibilityLabel should not be optional.

SteffiStuffel (approved):
Tested on Android emulator. Looks good, but

native/src/assets/index.ts
```diff
@@ -40,6 +43,9 @@ export {
   EventThumbnailPlaceholder2,
   EventThumbnailPlaceholder3,
   ExternalLinkIcon,
+  GPSFixed,
+  GPSNotFixed,
+  GPSOff,
```
steffenkleinle, line 48: So far we never use this naming afaik, how about
```suggestion
  LocationFixedIcon,
  LocationUnfixedIcon,
  LocationDisabledIcon
```
At least it should all be called `Icon`.

native/src/components/MapView.tsx
```diff
@@ -16,22 +16,29 @@ import {
...
+
 const StyledMap = styled(MapLibreGL.MapView)`
   width: 100%;
 `
 
-const StyledFAB = styled(FAB)<{ position: number | string }>`
-  align-items: flex-end;
+const StyledIcon = styled(IconButton)<{ position: number | string }>`
+  position: absolute;
+  right: 0;
```
steffenkleinle, line 38: Looks like this might break on RTL layouts

native/src/components/SettingItem.tsx
```diff
@@ -13,25 +12,36 @@ const PadView = styled.View`
...
 const Title = styled.Text<{ bigTitle: boolean }>`
   color: ${props => props.theme.colors.textColor};
   ${props => (props.bigTitle ? 'font-size: 18px;' : '')}
 `
+
 const Description = styled.Text`
   color: ${props => props.theme.colors.textSecondaryColor};
 `
 
+const RightContentContainer = styled.View`
```
steffenkleinle, line 31:
```suggestion
const FlexEndContainer = styled.View`
```

native/src/components/SettingItem.tsx
```diff
@@ -13,25 +12,36 @@ const PadView = styled.View`
...
+  justify-content: center;
+  align-items: flex-end;
+  padding: 0 5px;
+`
+
+const Badge = styled.View<{ enabled: boolean }>`
+  width: 8px;
+  height: 8px;
+  border-radius: 4px;
+  background-color: ${props => (props.enabled ? 'limegreen' : 'red')};
```
steffenkleinle, line 42: Could we add the colors to the constants and use hex codes?

native/src/components/base/IconButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  style?: StyleProp<ViewStyle>
+}
+
+const DEFAULT_BUTTON_SIZE = 40
+
+const IconButton = ({
+  accessibilityLabel,
+  icon,
+  onPress,
+  size = DEFAULT_BUTTON_SIZE,
```
steffenkleinle, line 30: Not sure if we should use a prop for this as width/height/border-radius can easily be styled using styled components. My personal opinion is to not use a prop for this.

native/src/components/MapView.tsx
LeandraH: Should be fine, we don't move the icon at the moment either.

native/src/components/SettingItem.tsx
LeandraH: Hmmm, these are pretty random colors I picked to roughly match the previous ones. The design system doesn't have feedback colors yet, I'd rather wait for those.

steffenkleinle: For me actually an even bigger reason to add it to the constants. Then we can easily change it later once we have the correct colors and don't just forget it.

native/src/components/ThemedSearchBar.tsx
```diff
@@ -1,50 +1,62 @@
...
+const StyledIcon = styled(Icon)`
+  color: ${props => props.theme.colors.textSecondaryColor};
+  height: 20px;
+`
+
+const StyledInput = styled(TextInput)`
+  padding: 4px;
+  flex: 1;
+  color: ${props => props.theme.colors.textColor};
+  font-size: 18;
```
steffenkleinle, line 31: I am getting a warning here:
> Expected style "fontSize: 18" to contain units

native/src/components/MapView.tsx
steffenkleinle: Actually react-native automatically moves this to the other side for RTL layouts so this is fine.

steffenkleinle (approved):
Tested on android

native/src/components/SettingItem.tsx
LeandraH: I think I'll leave it because they can easily found by search for red or green in the code, and I assume that will happen then.

native/src/components/MapView.tsx
LeandraH: Huh. Cool

native/src/components/ThemedSearchBar.tsx
LeandraH: Me too, but when I add units, the size is completely ignored, so I decided to ignore the warning 🤷

steffenkleinle: For me this has the same result, so I would strongly prefer this
```suggestion
  font-size: 18px;
```

LeandraH: It did for me now too. No idea what I did wrong previously. Grrrr.

Issue #2430, opened 2023-09-06
Settings icon in header too close to the edge


LeandraH:
### Describe the Bug

The settings icon is too close to the edge of the screen now, it seems to have lost some padding.

### Steps to Reproduce

1. Open the app in iOS or Android
2. Look at the header

### Expected Behavior

I'd like the icon to not look squashed against the edge.

### Actual Behavior

It looks a little hard to actually tap on the icon because it is so far right.

### Additional Information

<img width="211" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/18513943/6272ce42-a97c-4fe5-bfa8-de6addb393ee">

anxbt:
### **Cant run it locally** 
 tried the following commands 
-installed `yarn`
-ran `yarn start
yarn android`

i even made changes in the  **package.json** file

`"scripts": {
    "start": "react-native start",
    "android": "react-native run-android",
    "lint": "yarn run eslint",
    // ... other scripts
  },`

SteffiStuffel:
@anxbt Have a look at our readmes: 
[General Readme](https://github.com/digitalfabrik/integreat-app)
[Setup for native](https://github.com/digitalfabrik/integreat-app/blob/main/native/README.md)

You first need to `yarn install` and then change no folder `native` and run `yarn start` and `yarn android`. You need to have set up an Android Emulator before.

anxbt:
I know this issue is closed could u help me to run this Integrat app for Android locally. I have a Android emulator installed already as I   work with React native.i ran "yarn" then "yarn start" then "yarn Android". I even copied the whole original package.json file .

@lunars97 could u please help me.

lunars97:
@anxbt Hi, sure. In order to run android emulator, you need to change to native directory and there you need to run `yarn start` and `npx react-native run-android`

PR #2437, opened 2023-09-11
2430: Add space between settings icon and the edge of the screen

lunars97:
### Short description

The settings icon is too close to the edge of the screen on the header now, it seems to have lost some padding due to refactoring probably.

[https://github.com/digitalfabrik/integreat-app/issues/2430](Here) is the link to the issue

### Resolved issues

[https://github.com/digitalfabrik/integreat-app/issues/2430](Issue)

### Additional info

image

---


native/src/components/CustomHeaderButtons.tsx
```diff
@@ -47,7 +47,8 @@ const CustomHeaderButtons = (props: {
       <OverflowMenu
         onPress={onOverflowMenuPress(cancelLabel)}
         accessibilityLabel={t('moreOptions')}
-        OverflowIcon={<Icon Icon={MenuIcon} />}>
+        OverflowIcon={<Icon Icon={MenuIcon} />}
+        style={{ marginRight: 7 }}>
```
LeandraH, line 51: We try to avoid inline styling whenever possible and instead use [styled components](https://styled-components.com/docs/basics#getting-started)

LeandraH (commented):
Very nice, thank you! But try to use a styled component here.

A different thing, I think you copied the contents from the issue that you were working on into this PR. That's not necessary, you can just fill in the template that a PR is opened with, including that it automatically links the issue if you fill it in there. The idea here is that the PR has more of the technical implementation and the issue is the larger level overview.

native/src/components/CustomHeaderButtons.tsx
lunars97: @LeandraH Thank you for explaining the pr thing :) As I understood, I do not need to fill in the pr description if there is already the similar issue, I can only link this pr to that issue. Am I right?

SteffiStuffel (changes requested):
Thanks for this PR. For all Integreat Issues we always need to keep in mind, that every design should also work for RTL-languages (e.g. like Arabic). I attached a screenshot, the margin next to the search-icon is missing, please also add this margin. For testing you need to change the device language of your emulator to Arabic.

native/src/components/CustomHeaderButtons.tsx
LeandraH: Hmm, no, we should fill in the template, it's mostly a helper for the people reviewing to make that as easy and simple as possible, plus if we need to find out in the future why we did something, that can help too. Maybe take a look at other people's PRs to see what I mean.

I guess if your PR is part of a bigger set of PRs, there are some cases where you don't need to fill in much of the description, but even then I would recommend explaining a bit of what is happening in the PR.

In this specific case, there's obviously not much to explain so you could just say that you added padding to the icon which seems to have gotten lost somewhere in all the icon refactoring of the last few weeks.

SteffiStuffel (commented):
Thanks for the changes, works as expected.

For better code, that is more easy adjustable, it would be better to add a margin just at one place. I think it would be better to add a container (a styled view with a margin) around the header HeaderButtons in `CustomHeaderButtons.tsx` where the margin is set.

native/src/components/CustomHeaderButtons.tsx
```diff
@@ -14,8 +14,8 @@ import styled from 'styled-components/native'
 import { LanguageIcon, MenuIcon, SearchIcon } from '../assets'
 import Icon from './base/Icon'
 
-const StyledOverflowMenu = styled(OverflowMenu)`
-  margin-right: 10px;
+const StyledHeaderContainer = styled.View`
+  margin: 0 10px 0 10px;
```
steffenkleinle, line 18: Could be simplified to 
```suggestion
  margin: 0 10px;
```

steffenkleinle (approved):
Tested on android, well done :+1:

SteffiStuffel (approved):
Nice work. :rocket:

Issue #2394, opened 2023-08-28
Header title for deep linking


LeandraH:
**Is your feature request related to a problem? Please describe.**
The header in the app usually shows the name of the previous route. If the app is closed and opened via a deep link, there is no previous route, it instead shows the current route (but without translation?). It should correctly show the name of the route you are sent to if you tap the button to go back.

**Describe the solution you'd like**
The title in the header should always show the name of the route that you would return to if you tap on the back button next to it.

**Describe alternatives you've considered**
Can't think of any.

**Additional context**
I opened https://integreat.app/testumgebung/de/locations and got the attached screen. Note that it says "categories" in the header. When I tap on back, I am returned to Testumgebung, so the header title should be "Testumgebung" here.
![IMG_EF6664344B16-1](https://github.com/digitalfabrik/integreat-app/assets/18513943/51088063-517e-491b-b40c-493f4640c1a4)

PR #2441, opened 2023-09-13
2394: Fix Header title for deep linking

SteffiStuffel:
### Short description
removed `t(previousRoute.name)` as a fallback, as we do not have translations for route names, so I think this does never make sense. 
Other option would be to set the back navigation to the city name, what would make sense for opening all deep links, as their back navigation is defined as the home screen of the city, but we would have to load data to convert the citycode to the appropriate city-display name. I think this quite of an overhead, and just having the back icon is fine.

### Resolved issues

Fixes: #2394 

---


LeandraH (approved):
Seems like a good enough fix to the problem to me, and not worth the hassle for a pretty rare problem (I assume)

steffenkleinle (changes requested):
Actually, this breaks quite a few cases:
* news > select one news item (header should be "News" but is empty)
* events > select one events item (header should be "Events" but is empty)
* settings > licenses (header should be "Settings" but is empty)

I agree that the previous way of handling it is maybe not the most obvious/explicit, but this is not an improvement. Perhaps we have to explicitly set the header title in every route.

SteffiStuffel:
@steffenkleinle thanks for the input, i fixed it. Also replaced the empty string for categories with the "local information" translation, but can change back to empty if this is preferred.

Issue #2443, opened 2023-09-14
Hide i18next logs in tests


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a developer executing tests it is annoying that the test logs are cluttered with i18n debug logs.

**Describe the solution you'd like**
i18n debug logs should not be shown in the test logs.

**Describe alternatives you've considered**
Keep it the way it is.

PR #2444, opened 2023-09-14
2443: Hide i18next debug logs for test

steffenkleinle:
### Short description

Hide i18next debug logs for test.

### Proposed changes

Hide i18next debug logs for test by overwriting console.log before tests and restoring it afterwards.

### Side effects

None.

### Resolved issues

Fixes: #2443

---


LeandraH (approved):
Nice! Still getting a debug statement from `sentry.ts` but I don't think that's in the scope of this ticket, plus it's single lines as opposed to translation objects...

web/src/components/__tests__/I18nProvider.spec.tsx
```diff
@@ -9,9 +9,22 @@ import I18nProvider from '../I18nProvider'
 
 jest.mock('../../utils/BrowserLanguageDetector')
 jest.mock('translations/src/loadTranslations')
+jest.mock('')
```
f1sh1918, line 12: why we need this?

steffenkleinle: we don't, forgot to remove

f1sh1918 (approved):
looks fine :)

Issue #2386, opened 2023-08-24
Make the e2e tests less flaky


LeandraH:
**Is your feature request related to a problem? Please describe.**
The e2e tests are quite flaky which reduces our confidence in the pipeline.

**Describe the solution you'd like**
We should check which tests exactly keep failing, and what we can do to improve it. Possibilities off the top of my head:
- More retries
- Longer timeout
- Rewrite the test to look for something else

**Describe alternatives you've considered**
We could leave everything like it is because it is okay, just annoying. Or we could remove the failing tests, which I think we really really shouldn't.

sarahsporck:
And of course also update the e2e dependencies.

LeandraH:
Additional info: 9 out of the last 9 failed e2e tests were because of `change language should open language selector: Error: expect(received).toBeTruthy() Received: false`.

PR #2446, opened 2023-09-15
2386: Make the E2E tests less flaky

LeandraH:
### Short description

We've been having issues with the E2E tests failing quite a bit, and working when restarting them. I checked the last 11 times that the tests failed, and it was each time because of the test checking for the overlay to change the language to appear.

### Proposed changes

- I added one additional waiting command to that test, to wait specifically for one of the language links. In 200 tests on my machine, the test hasn't failed since.

### Side effects

- If we take out the English language from the E2E-Testumgebung, the test will fail again, and then always. But I don't see why we would do that.
-

### Resolved issues

Fixes: #2386 

---


e2e-tests/web/test/specs/changeLanguage.e2e.ts
```diff
@@ -15,6 +15,7 @@ describe('change language', () => {
     await languageIcon.click()
 
     const languageSelector = await $("*[data-testid='headerActionItemDropDown']")
+    await $('=English')
```
sarahsporck, line 18: Can you explain why this makes the tests less flaky? :)

LeandraH: My guess is that the `headerActionItemDropDown` (i.e. the box that contains all the languages that you can choose) sometimes is rendered but in a state where it's still not visible, so the test for the visibility would fail. But when awaiting one of its child elements to show up, that would mean that it has to be visible, so the test passes.

SteffiStuffel:
> * I added one additional waiting command to that test, to wait specifically for one of the language links. In 200 tests on my machine, the test hasn't failed since.
> 

Did the tests failed on your local machine before?

LeandraH:
Yes, I don't have the exact numbers anymore but I put in a 50x loop for that specific test and it would fail between 30 and 40 times out of those. Only in Chrome though, not in Firefox or Safari.

SteffiStuffel (approved):

sarahsporck (approved):

Issue #2304, opened 2023-08-14
Write info about skipping a job to channels


SteffiStuffel:
Lately we forgot to unset the "SKIP_JOB_PROMOTE_INTEGREAT_IOS variable and Integreat was not updated for a few months. To prevent this in the future and to be more transparent, I suggest we write a message to ~releases or ~tech-react (to project channels may be to much effort for the benefit), so we (and everyone else) gets informed that a promotion or whatever is skiped.

steffenkleinle:
In my opinion this should go to ~integreat-app-notifications for integreat and perhaps a new channel ~lunes-app-notifications for lunes

PR #2449, opened 2023-09-18
2304: Nofity skip deliver and promote jobs

SteffiStuffel:
### Short description

Added notification to skip_job command

### Proposed changes

- moved notify script to tools
- changed order of skip_job and checkout

### Side effects

- hopefull none
-

### Resolved issues

Fixes: #2304 

---


.circleci/src/commands/skip_job.yml
```diff
@@ -6,15 +6,17 @@ parameters:
...
       command: |
         function skip_job {
           build_config_name=$1
           variable_name=SKIP_JOB_${CIRCLE_JOB%-*} # Remove everything after "-". e.g. "check-1" -> check
           if [[ -n ${!variable_name} && ($build_config_name =~ ${!variable_name} || "all" =~  ${!variable_name}) ]]; then
-            echo "This job is skipped because the variable $variable_name is set"
+            message="Job is skipped because the variable $variable_name is set"
+            echo $message
             circleci step halt
+            source tools/notify-mattermost main integreat-app-notifications "$message"
```
steffenkleinle, line 19: Why are we using the tool here directly instead of the `notify` command?

tools/notify-mattermost
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
steffenkleinle, line 1: Did you change anything here except extracting this?

.circleci/src/commands/skip_job.yml
SteffiStuffel: Because we cannot call the notify command conditionally depending on a environment variable. CircleCI does not support calling commands conditionally on environment variables, as they are evaluated on run time, whereas the conditional-statements are evaulated on compile time.

tools/notify-mattermost
SteffiStuffel: adding the option to pass a message to the command, but beside of this it is the same.

steffenkleinle (approved):
LGTM

tools/notify-mattermost
```diff
@@ -0,0 +1,41 @@
...
+    if [ "$STATUS" -ne "200" ]; then
+        echo "Notification not sent due to an error. Status: $STATUS. Please check the webhook URL"
+        exit 1
+    fi
+
+    echo "Notification sent!"
+    exit 0
+  }
+
+Send_notification $1 $2 "$3"
```
LeandraH, line 41: Why does the third one get quotation marks?

LeandraH (commented):
I don't get it, why are we moving the `skip_job` to after the `checkout`? Doesn't that mean unnecessary checking out of the repo?

tools/notify-mattermost
SteffiStuffel: Because we are passing a string with spaces and without the quotes only the first word would be passed

SteffiStuffel:
> I don't get it, why are we moving the `skip_job` to after the `checkout`? Doesn't that mean unnecessary checking out of the repo?

Yes that is correct. We need to change the order, because the nofity-script it not be available, without checking out before. As we only do skipping jobs rarely and checking out does not take very long this is fine.

LeandraH:
Aaaah, cool, thanks!

LeandraH (approved):

Issue #2392, opened 2023-08-28
Adjust release workflow to be able to pass a branch


SteffiStuffel:
As a developer I want to deploy a hotfix easily.
We can use the release tag of the version that needs the hotfix.
Create a hotfix branch on this tag and implement the fix.

## Acceptance Criteria:

- soften the rule that the main branch will be triggered for delivery workflows
- add the ability to pass a branch to the delivery worksflows (in this case the hotfix branch)
- if no branch is provided the main branch should be used as default

PR #2452, opened 2023-09-19
2392: Remove main protection rule for delivery

SteffiStuffel:
### Short description

Removed main protection rule for delivery

### Proposed changes

- a warning is sent to mattermost, if we trigger delivery from feature branch as next delivery will fail if feature branch is not merged before
- script for triggering the pipeline is adjusted so branch can be passed. If no branch is passed, main is used.

### Resolved issues

Fixes: #2392 

---

### DIFF
--- a/tools/trigger-pipeline.ts
+++ b/tools/trigger-pipeline.ts
@@ -20,15 +20,17 @@ program
   .command('trigger <workflow-type>')
   .description(`trigger a workflow in the ci on the main branch`)
   .requiredOption('--api-token <api-token>', 'circleci api token')
-  .action(async (workflowType: string, options: { apiToken: string }) => {
-    const { apiToken } = options
+  .option('--branch <branch>', 'the branch the workflow will be triggered on', MAIN_BRANCH)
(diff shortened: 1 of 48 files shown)


tools/trigger-pipeline.ts
```diff
@@ -20,15 +20,17 @@ program
   .command('trigger <workflow-type>')
   .description(`trigger a workflow in the ci on the main branch`)
   .requiredOption('--api-token <api-token>', 'circleci api token')
-  .action(async (workflowType: string, options: { apiToken: string }) => {
-    const { apiToken } = options
+  .option('--branch <branch>', 'the branch the workflow will be triggered on', MAIN_BRANCH)
```
f1sh1918, line 23: what does this last parameter `MAIN_BRANCH` mean in this case?

f1sh1918 (approved):
Looks fine to me

tools/trigger-pipeline.ts
steffenkleinle: I think that is just the default if nothing is passed, see https://github.com/tj/commander.js#default-option-value

steffenkleinle (approved):

Issue #2381, opened 2023-08-22
Tabbing opens sidebar


steffenkleinle:
### Describe the Bug

Tabbing through the app opens the sidebar and breaks the layout/makes the app unusable.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/testumgebung/de-si
2. Press tab (keyboard key) until the sidebar opens (19 times)
3. See that the sidebar opens
4. The sidebar doesn't close on click on the button

### Expected Behavior

The sidebar should only open if clicked on the kebab menu, not on tabbing there.

### Actual Behavior

The sidebar opens on tab pressed.

### Additional Information

Check that the sidebar behaves a11y wise correct otherwise.

</details>

PR #2454, opened 2023-09-19
2381: Don't open the kebab menu when tabbing through the app

LeandraH:
### Short description

When tabbing all the way through the app, at the very end it would open the kebab menu because it was already there just visibly hidden and the tab somehow seems to trigger an `onClick` there. This also tells the HTML that the menu is hidden when the user can't see it, and now doesn't open just by tabbing through anymore.

### Resolved issues

Fixes: #2381 

---


steffenkleinle (approved):
Tested on firefox

PR #2455, opened 2023-09-19
2271: Incorrect map padding

f1sh1918:
### Short description

The map padding was not correct in all cases. The user location icon was not centered for all snap points of the bottom action sheet

### Proposed changes

- removed followUserLocation from MapLibre since its blocks the `setCamera` and this issue wasn't resolved for a while
- setUserLocation manually
- removed lots of code due to deactivate `followUserLocation` for setCamera is not needed anymore

### Side effects

- there is one issue with the default settings that i couldn't fix. I don't know why the camera is moving to the default setting when i add there a padding. Maybe someone could check. We could also remove this padding again.

### Resolved issues

Fixes: #2271 

### Testing
- Click on the userLocation icon with and without open BottomActionSheet


release-notes/unreleased/2271-incorrect-map-padding.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2271
+show_in_stores: false
```
steffenkleinle, line 2: German translation not needed/never used if this is set to false.

native/src/routes/Pois.tsx
```diff
@@ -249,12 +250,14 @@ const Pois = ({ pois: allPois, language, cityModel, route, navigation }: PoisPro
...
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        bottomSheetHeight={getBottomSheetSnapPoints(deviceHeight)[sheetSnapPointIndex]!}
         locationPermissionGranted={!!coordinates}
         onRequestLocationPermission={requestAndDetermineLocation}
         iconPosition={
-          sheetSnapPointIndex < BOTTOM_SHEET_SNAP_POINTS.length - 1
+          sheetSnapPointIndex < getBottomSheetSnapPoints(deviceHeight).length - 1
             ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              BOTTOM_SHEET_SNAP_POINTS[sheetSnapPointIndex]!
+              getBottomSheetSnapPoints(deviceHeight)[sheetSnapPointIndex]!
```
steffenkleinle, line 260: Please extract this to a separate const such that disabling eslint is only needed once.
Optional: snapPoints are also calculated 4 times atm and could be extracted.

native/src/components/MapView.tsx
```diff
@@ -146,28 +150,30 @@ const MapView = ({
     setSheetSnapPointIndex(1)
   }
 
+  const deactivateFollowUserLocation = () => {
+    if (followUserLocation) {
+      setFollowUserLocation(false)
+    }
+  }
```
steffenkleinle, line 157: Why not just, setting it to false if it already is false does not have an effect.
```suggestion
  const deactivateFollowUserLocation = () => setFollowUserLocation(false)
```

native/src/components/MapView.tsx
```diff
@@ -146,28 +150,30 @@ const MapView = ({
...
         styleJSON={mapConfig.styleJSON}
         zoomEnabled
         onPress={onPress}
         ref={mapRef}
+        onStartShouldSetResponder={() => true}
+        onResponderMove={deactivateFollowUserLocation}
         attributionEnabled={false}
         logoEnabled={false}>
-        <MapLibreGL.UserLocation visible={locationPermissionGranted} />
+        <MapLibreGL.UserLocation visible={locationPermissionGranted} onUpdate={location => setUserLocation(location)} />
```
steffenkleinle, line 170:
```suggestion
        <MapLibreGL.UserLocation visible={locationPermissionGranted} onUpdate={setUserLocation} />
```

native/src/components/MapView.tsx
```diff
@@ -97,41 +98,44 @@ const MapView = ({
...
     await onRequestLocationPermission()
-    setFollowUserLocation(true)
-  }, [onRequestLocationPermission, setFollowUserLocation])
-
-  const onUserTrackingModeChange = (
-    event: MapLibreGLEvent<'usertrackingmodechange', { followUserLocation: boolean }>,
-  ) => {
-    if (!event.nativeEvent.payload.followUserLocation) {
-      setFollowUserLocation(event.nativeEvent.payload.followUserLocation)
+    if (userLocation?.coords && cameraRef.current) {
```
steffenkleinle, line 121:
```suggestion
    if (userLocation?.coords) {
```

native/src/components/MapView.tsx
```diff
@@ -63,6 +62,7 @@ type MapViewProps = {
   setSheetSnapPointIndex: (index: number) => void
   followUserLocation: boolean
   setFollowUserLocation: (value: boolean) => void
```
steffenkleinle, line 64: Do we actually still need those props? I think we only still use it to decide which location icon to display. If that is the case, we could also compare the map coordinates with the user location coordinates and decide based on that which icon to display. At least we should be able to just move it to MapView and avoid handling it in Pois.

f1sh1918: Not sure if tracking and comparing the map/camera coordinates is more performant and easier to handle @steffen.kleinle

native/src/components/MapView.tsx
```diff
@@ -100,33 +98,34 @@ const MapView = ({
     zoomLevel: coordinates ? normalDetailZoom : defaultViewportConfig.zoom,
     centerCoordinate: coordinates,
     bounds: coordinates ? undefined : bounds,
+    padding: { paddingBottom: bottomSheetHeight },
```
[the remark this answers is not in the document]
LeandraH: So do we need them? It works for me without padding and animationDuration: 0

LeandraH (approved):
Nice! Tested on simulated iOS

native/src/components/MapView.tsx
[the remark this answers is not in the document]
f1sh1918: yes but without padding its not centered when the bottom sheet is open @LeandraH

PR #2459, opened 2023-09-21
2111: Add tests for Breadcrumbs.tsx

LeandraH:
### Short description

This adds some tests for the breadcrumbs.

### Side effects

- I also added a possible title to our Icon for better accessibility.

### Resolved issues

Fixes: #2111 

---


steffenkleinle (approved):
Good idea with the accessibility title!
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title Maybe we should use it in a few other places as well (or the aria-labelledby attribute)
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

LeandraH:
Since we're trying to keep our PRs small, I opened a new ticket instead of adding them here 😁 
#2470

ztefanie (approved):
Thanks for this PR :tada:

Issue #2445, opened 2023-09-14
Translate app constants


LeandraH:
**Is your feature request related to a problem? Please describe.**
Since the last time we translated our app constants some time has passed and a lot of new constants were added. We should therefore translate the constants again.

**Describe the solution you'd like**
The constants should be exported, converted to odt and sent to @osmers. When we get them back translated, import them again.

**Describe alternatives you've considered**
N/A.

**Additional context**
N/A.

### Describe the Bug

When I open the feedback modal in languages other than German, most of the text is in German only

### Steps to Reproduce

1. Go to 'https://integreat.app/lkansbach/en'
2. Click on 'Feedback'
3. See that the feedback modal is largely in German

### Expected Behavior

It should be all in the selected language.

### Actual Behavior

It is mostly in German.

### Additional Information

<img width="415" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/18513943/d2412b42-e46f-439f-99e6-08e2c28e01be">

steffenkleinle:
Thats not a bug, we just need to do the translations again. Would probably still make sense to do this :)

PR #2472, opened 2023-09-23
1925: Make the comment field of the search feedback optional

LeandraH:
### Short description

This PR makes the comment field in the search feedback optional and instead adds another text input where the user can enter what they were looking for (if they want). That text input is pre-filled with the search query. If the user changes the text in that second input and sends, the original search query is also sent along.

### Side effects

--

### Resolved issues

Fixes: #1925 

---


ztefanie (commented):
1925-ar

Colon is on the wrong side in web.

Shouldn't the text field of the not found word, be disabled? If deleting the text in this textfiled and trying to send the feedback it only says: "An error occurred while sending the feedback." HTTP Error is 400.

I think there should be a little more padding between the different parts of the feedback. in the design the whole content is much wider on web desktop.

LeandraH:
> Colon is on the wrong side in web.

That seems to be a browser thing with RTL and a punctuation mark at the end, it's moved to the last spot. You can also see it on the next German line where period is at a completely wrong position. Should be fixed by adding the translations.

LeandraH:
> Shouldn't the text field of the not found word, be disabled? If deleting the text in this textfiled and trying to send the feedback it only says: "An error occurred while sending the feedback." HTTP Error is 400.

I don't think it should be disabled because someone might not complete typing the word they were looking for in the search field but then be inclined to tell us what they were actually looking for if asked. But they also don't have to, to keep feedback as easy as possible for the user.

In other news, TIL: `'' ?? query` becomes `''`. Wieder was gelernt.

steffenkleinle:
> > Colon is on the wrong side in web.
> 
> That seems to be a browser thing with RTL and a punctuation mark at the end, it's moved to the last spot. You can also see it on the next German line where period is at a completely wrong position. Should be fixed by adding the translations.

Not sure why that exactly happens, but should be fixed with #2445, not a problem introduced here.

ztefanie:
> > Shouldn't the text field of the not found word, be disabled? If deleting the text in this textfiled and trying to send the feedback it only says: "An error occurred while sending the feedback." HTTP Error is 400.
> 
> I don't think it should be disabled because someone might not complete typing the word they were looking for in the search field but then be inclined to tell us what they were actually looking for if asked. But they also don't have to, to keep feedback as easy as possible for the user.
> 
> In other news, TIL: `'' ?? query` becomes `''`. Wieder was gelernt.

Sending an empty text in this textfield needs to be handled. Feel free to think of another solution as the one I suggested.

LeandraH:
> > > Shouldn't the text field of the not found word, be disabled? If deleting the text in this textfiled and trying to send the feedback it only says: "An error occurred while sending the feedback." HTTP Error is 400.
> > 
> > 
> > I don't think it should be disabled because someone might not complete typing the word they were looking for in the search field but then be inclined to tell us what they were actually looking for if asked. But they also don't have to, to keep feedback as easy as possible for the user.
> > In other news, TIL: `'' ?? query` becomes `''`. Wieder was gelernt.
> 
> Sending an empty text in this textfield needs to be handled. Feel free to think of another solution as the one I suggested.

It is handled, if the user removes the text in that text field, it sends the query instead. I set `query: searchTerm?.trim().length === 0 ? query : searchTerm` in getFeedbackData.

web/src/components/Feedback.tsx
```diff
@@ -141,10 +156,12 @@ const Feedback = (props: FeedbackProps): ReactElement => {
         <ErrorSendingStatus role='alert'>{t('failedSendingFeedback')}</ErrorSendingStatus>
       )}
 
-      <NoteContainer showContainer={sendFeedbackDisabled}>
-        <Icon src={NoteIcon} />
-        <NoteText>{t('note')}</NoteText>
-      </NoteContainer>
+      {!isSearchFeedback && (
```
f1sh1918, line 159: Instead of hiding this note you can just adjust the message for `isSearchFeedback` that shows up when the form field for missing search term is empty (deleted by user)

"Bitte geben sie den Suchbegriff ein für den sie kein Ergebnis finden konnten"

f1sh1918 (changes requested):
I found some issues with empty "Suchbegriff" field. You have to prevent that the user can send a feedback request even this field has become empty. Additionally to disabled the send button for this case you should also show a proper note to the user

LeandraH:
I find it to be much lower-threshold to not make the additional field an obligatory one and instead just send the query that the user had originally put in if they remove the one from the additional field for some reason. Which is also why e.g. it's not fine to pass in searchTerm but has the fallback of the query.

LeandraH:
I seem to not be making it clear what I mean because I've written this several times by now, so I'll try to give some examples of the flow that I've written.

1. The user is looking for information on Einbürgerung. They start typing and as they've typed in "Einbürg", the search shows him that there are no results. Ideally, the user would then fill in an additional "erung" into the field labelled searchTerm, so that it reads "Einbürgerung". When the user hits send, the query sent to the server is "Einbürgerung".
2. The user is looking for information on Einbürgerung. They start typing and when they've entered the whole word "Einbürgerung", they notice that there are no search results. They don't edit the field with the searchTerm. They hit send on the feedback, and the query that is sent to the server is "Einbürgerung".
3. This is the case that I think we are talking about, unless I am misunderstanding something: The user is looking for information on Einbürgerung. They start typing and when they've entered "Einbürg", they notice that there are no results. There is a second text edit field that now has the word "Einbürg" in it which they find confusing, so they delete the word there. If they now hit send, the query that is sent to the server is "Einbürg".

Does that make it clearer what I mean? Or did I misunderstand your comments?

f1sh1918:
I even think it should be visible what will be send to the backend. You could use the note field to show what query will be send if searchTerm is empty

f1sh1918:
I even think it should be visible what will be send to the backend. You could use the note field to show what query will be send if searchTerm is empty

LeandraH:
Good idea, thanks! I'll also add some tests for documentation :)

ztefanie:
> > > > Shouldn't the text field of the not found word, be disabled? If deleting the text in this textfiled and trying to send the feedback it only says: "An error occurred while sending the feedback." HTTP Error is 400.
> > > 
> > > 
> > > I don't think it should be disabled because someone might not complete typing the word they were looking for in the search field but then be inclined to tell us what they were actually looking for if asked. But they also don't have to, to keep feedback as easy as possible for the user.
> > > In other news, TIL: `'' ?? query` becomes `''`. Wieder was gelernt.
> > 
> > 
> > Sending an empty text in this textfield needs to be handled. Feel free to think of another solution as the one I suggested.
> 
> It is handled, if the user removes the text in that text field, it sends the query instead. I set `query: searchTerm?.trim().length === 0 ? query : searchTerm` in getFeedbackData.

Oke I see. On my previous test i got the 400 error, not sure why. But it is working now as expected.

ztefanie (approved):

native/src/components/FeedbackContainer.tsx
```diff
@@ -99,7 +111,7 @@ const FeedbackContainer = (props: FeedbackContainerProps): ReactElement => {
       city: cityCode,
       language,
       comment: commentWithMail,
-      query,
+      query: searchTerm?.trim().length === 0 ? query : searchTerm,
```
steffenkleinle, line 114: Actually I think it might make sense to just send both (similar to what we do with the contact address:
```suggestion
      query: query === searchTerm ? query : `${searchTerm} (actual query: ${query})`,
```

That would give us/the cities even more information to work with and probably also an insight if the current UI is misleading/confusing.

web/src/components/FeedbackContainer.tsx
```diff
@@ -88,7 +102,7 @@ export const FeedbackContainer = (props: FeedbackContainerProps): ReactElement =
       city: cityCode,
       language,
       comment: commentWithMail,
-      query,
+      query: searchTerm?.trim().length === 0 ? query : searchTerm,
```
steffenkleinle, line 105: Same as for native

web/src/components/Feedback.tsx
```diff
@@ -10,7 +10,7 @@ import Icon from './base/Icon'
 import TextButton from './base/TextButton'
 import ToggleButton from './base/ToggleButton'
 
-export const Container = styled.div`
+export const Container = styled.div<{ widerDisplay?: boolean }>`
```
steffenkleinle, line 13: Perhaps:
```suggestion
export const Container = styled.div<{ fullWidth?: boolean }>`
```

steffenkleinle (commented):
One thing I noticed: On native the feedback quickly pops up before the list is rendered, would you mind fixing this here or should I create an issue for it?

Personal opinion and could and probably even should be done in a separate issue: IMO Feedback and FeedbackContainer should be merged (both native and web) because we are just passing a lot of props between the two components without any real separation of concerns from a logical point of view. Common logic between web and native could be moved to the api client/endpoint. What do you think?

Tested on android and firefox.

web/src/components/__tests__/FeedbackContainer.spec.tsx
```diff
@@ -148,10 +148,6 @@ describe('FeedbackContainer', () => {
     const { getByRole, findByText, queryByRole } = renderWithTheme(
       <FeedbackContainer {...buildDefaultProps(CATEGORIES_ROUTE, true)} />,
     )
-    const buttonRating = getByRole('button', {
-      name: 'feedback:useful',
-    })
-    fireEvent.click(buttonRating)
```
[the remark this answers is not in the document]
LeandraH: But this isn't normal feedback, this is search feedback. I'll update the tests to make that clearer.

native/src/components/FeedbackContainer.tsx
LeandraH: That's a great idea!

native/src/components/Feedback.tsx
```diff
@@ -121,7 +121,7 @@ const Feedback = ({
               <HeadlineContainer>
                 <Description>{t('searchTermDescription')}</Description>
               </HeadlineContainer>
-              <Input value={searchTerm} onChangeText={setSearchTerm} />
+              <Input value={searchTerm} onChangeText={setSearchTerm} testID='searchTerm' />
```
steffenkleinle, line 124:
```suggestion
              <Input value={searchTerm} onChangeText={setSearchTerm} />
```

native/src/components/__tests__/FeedbackContainer.spec.tsx
```diff
@@ -110,4 +116,65 @@ describe('FeedbackContainer', () => {
...
+        <FeedbackContainer
+          routeType={SEARCH_ROUTE}
+          isSearchFeedback
+          language={language}
+          cityCode={city}
+          query={query}
+        />
+      </NavigationContainer>,
+    )
+    const input = getByTestId('searchTerm')
```
steffenkleinle, line 163: `getByTestId` should only be used if not possible otherwise. In this example, `getByDisplayText` can be used instead:
```suggestion
    const input = getByDisplayValue(query)
```
See https://testing-library.com/docs/queries/bydisplayvalue.
If we would use a placeholder text, https://testing-library.com/docs/queries/byplaceholdertext could be used as well.

steffenkleinle (approved):

native/src/components/__tests__/FeedbackContainer.spec.tsx
LeandraH: Thanks, didn't know about `getByDisplayValue`, I couldn't get it to work with `getByText` :)

Issue #2321, opened 2023-08-15
Add minimum space between navigation tiles and toolbar


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
If there are not that many languages available in a city, there is no space between the toolbar and the navigation tiles which doesn't look that nice imo.

**Describe the solution you'd like**
There should be a minimum space between the navigation tiles and the toolbar if the toolbar is not pushed down because of the language selector.

**Describe alternatives you've considered**
Keep it as it is.

**Additional context**
![image](https://github.com/digitalfabrik/integreat-app/assets/29095487/aac0752d-62e2-4d37-bb9b-e7ca7e4a4063)

PR #2476, opened 2023-09-25
2321: Added minimum space between navigation tiles and toolbar

relentless95:
### Short description
there is almost no space between navigation tiles and toolbar when there are only a few languages available for a city. 

### Proposed changes
Created the variable additionalWidth (35px) in dimensions.tsx  to be used inside the styled.aside element found in Layout.tsx to increase the distance between the navigation tiles and toolbar.

### Side effects

no side effects related to the changes. 
However, i noticed a test failed when running yarn test. this is related to language selection. See image below: 
image

### Resolved issues

Fixes: #2321


.history/web/src/components/Layout_20230925102115.tsx
LeandraH, line 1: I'm not sure how all of these `.history` changes appeared here but please revert them.

web/src/components/Layout.tsx
```diff
@@ -94,7 +94,7 @@ const Main = styled.main<{ fullWidth: boolean }>`
 `
 
 const Aside = styled.aside<{ languageSelectorHeight: number }>`
-  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge}px;
+  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge + dimensions.additionalWidth}px;
```
LeandraH, line 97: Nicely done, working around the magic-number limitation! I'm not sure the constant should be called `additionalWidth` though since it isn't additional space in the horizontal but in the vertical direction. How about `additionalSpacing` instead?

LeandraH (commented):
Very cool, thanks for your PR! But when I refresh a page (e.g. http://localhost:9000/lkansbach/de), the Toolbar is positioned right under the header and then jumps down after a short bit. I think it would be nicer to have it there immediately.

ztefanie (changes requested):
You messed up with git somehow, there are 10.000+ commits and a lot of changes in this PR that are not from you. 

I suggest you:
1) checkout main
2) checkout a new branch 
3) Cherry Pick the commits you have done: https://git-scm.com/docs/git-cherry-pick
4) Push
5) Open a new PR
6) Close the old PR
7) Delete the old branch

relentless95:
I had an extension which tracked the changes in each file when i saved them. This created a lot of files without me knowing. So, i reverted those changes but didn't know it would cause all these issues. I will close this pr

steffenkleinle:
@relentless95 can we delete this branch? :)

relentless95:
yes

PR #2484, opened 2023-09-27
2026: Update to yarn berry

sarahsporck:
### Short description

In lunes we already use yarn berry for building our packages, as it is supposedly much faster. 
For long I'd wanted to implement yarn berry into the integreat-app project, but last time I've reached this point we encountered an oom in the CI, when installing the dependencies, therefore, we aborted this mission. To avoid the same issue, I tried to get to the root of the problem, and it seems linking and building the native project alone already takes around 1.5GB of memory. 
Also, as switching to pnp is, for the native project only, not possible, I removed native from our yarn workspaces and used the old node_modules linker only for this project. 

### Proposed changes

- TODO
-

### Side effects

- native project is not part of the yarn workspaces anymore

### Resolved issues

Fixes: #2026 

---


sarahsporck:
I'll do the actuall pr separately, to have a cleaner history.

PR #2488, opened 2023-10-02
Remove outdated comment

LeandraH:
### Short description

Just removing a comment that we don't need anymore (the color is correct already)


steffenkleinle (approved):
Ups :D

PR #2489, opened 2023-10-02
2321: Added minimum space between navigation tiles and toolbar

relentless95:
### Short description

there is almost no space between navigation tiles and toolbar when there are only a few languages available for a city.

### Proposed changes

Created the variable additionalWidth (35px) in dimensions.tsx to be used inside the styled.aside element found in Layout.tsx to increase the distance between the navigation tiles and toolbar.

### Side effects

None

### Resolved issues

Fixes: #2321


LeandraH (commented):
Woops, that looks like a merge conflict that was committed. Do you know how to fix that?

relentless95:
i think so, but i am not sure where the merge conflict is. can you point it out?

LeandraH:
It's a conflict between the two versions of `top`. You can tell by the very typical structure that git adds to merge conflicts, which is the <<<<< Branch1 piece before the first version, then the ====== to separate the two versions and then the >>>>>> Brach2 piece. You can remove all of that and just leave your version of `top`.

web/src/components/Layout.tsx
```diff
@@ -94,13 +94,13 @@ const Main = styled.main<{ fullWidth: boolean }>`
...
 
 const Aside = styled.aside<{ languageSelectorHeight: number }>`
-  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge}px;
+  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge + dimensions.additionalSpacing}px;
   margin-top: ${props => props.languageSelectorHeight - dimensions.navigationMenuHeight}px;
   display: inline-block;
   position: sticky;
   width: ${dimensions.toolbarWidth}px;
   vertical-align: top;
-  transition: top 0.2s ease-in-out;
```
LeandraH, line 103: With a 0s transition, you can completely remove the transition :)

LeandraH (commented):
I think at a screen width of e.g. 1000px, the toolbar is too low

image

web/src/components/Layout.tsx
steffenkleinle: Which we shouldn't do, please readd. Except you have a reason for changing this?

relentless95: i was requested to remove the transition feature because when the page loads, the transition feature moves the toolbox down after some time.  This transition was unnecessary so i just set the time to 0 instead of removing it completely.

LeandraH: That was me, I think it feels weird for the toolbar to float down after everything is done. Do you not?

steffenkleinle: TBH I don't mind either, especially since I am not even sure when the transition happens/how it looks like, for some reason I can't reproduce it :D.
Then lets remove the transition completely then though.

web/src/constants/dimensions.ts
```diff
@@ -30,6 +31,7 @@ const dimensions: DimensionsType = {
   footerHeight: 50,
   navigationMenuHeight: 90,
   poiDesktopPanelWidth: 332,
+  additionalSpacing: 35,
```
steffenkleinle, line 34: In my opinion we should not put this in this dimensions.ts and instead just use it directly as a magic number. Reasons:
* It is only used in one place and probably really specific to this one use case
* The naming is really generic and might mean anything and probably leads to more confusion than it helps

Personally I would also prefer some power of two number like `32px` but thats just a personal opinion.

steffenkleinle (commented):
Congrats on your first PR here :tada: Thanks a lot!

LeandraH (approved):
Looks great, thank you!

web/src/components/Layout.tsx
```diff
@@ -93,14 +95,13 @@ const Main = styled.main<{ fullWidth: boolean }>`
...
-  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge}px;
-  margin-top: ${props => props.languageSelectorHeight - dimensions.navigationMenuHeight}px;
+const Aside = styled.aside`
+  top: ${dimensions.headerHeightLarge + topOffset}px;
   display: inline-block;
   position: sticky;
   width: ${dimensions.toolbarWidth}px;
   vertical-align: top;
-  transition: top 0.2s ease-in-out;
+
```
steffenkleinle, line 104:
```suggestion
```

web/src/components/Layout.tsx
```diff
@@ -93,14 +95,13 @@ const Main = styled.main<{ fullWidth: boolean }>`
   }
 `
 
-const Aside = styled.aside<{ languageSelectorHeight: number }>`
-  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge}px;
-  margin-top: ${props => props.languageSelectorHeight - dimensions.navigationMenuHeight}px;
+const Aside = styled.aside`
+  top: ${dimensions.headerHeightLarge + topOffset}px;
```
steffenkleinle, line 99: Why did you choose to remove/change those instead of just adding an offset on top of the previous set `top`/`margin-top`? This way, on some widths it breaks, see the following screenshot where the toolbar pokes through the language selector.
Imo this should be reverted and just an additional offset added to the top.


Can be reproduced here with a width e.g. 782px:
http://localhost:9000/testumgebung/de/events

steffenkleinle (changes requested):

web/src/components/Layout.tsx
LeandraH: I don't think the toolbar needs to be fully visible when the language selectors are open, just like the top of the rest of the page isn't visible then. Though it should also not poke through, of course.

steffenkleinle: We tried to make it not poke through before a few times and we did not manage because of various z-indices set across the app. So the workaround is to just always make it fully visible (below the language selector). However, if you got an idea on how to fix it that would be amazing of course! But otherwise, this should be fixed by adding enough top margin/top offset.

LeandraH: Aaaaaah, I see, thank you for the explanation :) Looks like the `aside` having a new stacking context from being sticky is the culprit. I'll see if I find a solution for that, or do you have one, @relentless95 ?

relentless95: I agree also agree with you both. I just thought that it didn't matter whether the toolbar is visible when trying to access the  different languages because the goal is to access change the language. But i can just revert it, just let me know your final decision.

LeandraH: It can be fixed by adding a `z-index: 0` to the `Body` (i.e. the parent of the `Aside`) in the `Layout.tsx`. That's because each stacking context is treated like one single element by the parent container when it comes to text flow. So we need to change the stacking order in that parent container (in this case the `Body`) to affect the `z-index` of the stacking context.

LeandraH: Ah, I see, that puts the feedback module behind the header because it's a child of the aside. Too bad. I think fixing that is outside the scope of this ticket, so don't worry about it.
Which means that if we don't want the toolbar poking through, we need to put it lower than the language selectors. Unless you see another solution to not having both of them overlap there?

steffenkleinle: @relentless95 are you still working on this? At least in my opinion most of the changes should be reverted and only this line should be changed in the following way:
``` ts
  top: ${props => props.languageSelectorHeight + dimensions.headerHeightLarge + additionalToolbarTopSpacing}px;
```
where `additonalToolbarTopSpacing` is something like `8`, `16` or `32` px or another order of 2. I think I like `32` but thats your choice :)

steffenkleinle: Ah, didn't know that :see_no_evil: Lets see, I think we can get this ready :)

steffenkleinle:
@relentless95 can you run prettier once? Seems like some formatting is still of in Layout.tsx.

PR #2493, opened 2023-10-02
2101: Back button from POI goes back to previous route

LeandraH:
### Short description

When clicking on a link to a location (e.g. in Testumgebung (DE) > Test von Links zu Orten), the back button used to not lead to the page you came from but to the maps overview. With this PR, the back button leads to the page you just came from.

### Proposed changes

- I added a little check if the slug is defined when the PoisContainer is rendered. If we start on the normal map page, then there is no slug yet but if we start directly inside a POI, there is a slug at render time. In that case, the back button leads to the previous route as opposed to the maps route.

### Side effects

-
-

### Resolved issues

Fixes: #2101 

---


native/src/routes/Pois.tsx
```diff
@@ -135,7 +136,7 @@ const Pois = ({ pois: allPois, language, cityModel, route, navigation }: PoisPro
       deselectAll()
     }
   }
-  useOnBackNavigation(slug || multipoi ? deselectFeature : undefined)
+  useOnBackNavigation((slug || multipoi) && enteredViaMaps ? deselectFeature : undefined)
```
f1sh1918, line 139: this check looks weird. first you check if the slug or multipoi is there and then you check if `slug === undefined` (enteredViaMaps)

native/src/routes/PoisContainer.tsx
```diff
@@ -21,6 +21,8 @@ const PoisContainer = ({ navigation, route }: PoisContainerProps): ReactElement
   const { slug, multipoi } = route.params
   const { cityCode, languageCode } = useCityAppContext()
 
+  const [enteredViaMaps] = useState(slug === undefined)
```
f1sh1918, line 24: I think you don't have to pass a state here?
You have the `slug` also in `Pois`

f1sh1918 (commented):
@LeandraH can you please provide a short explanation how to test, f.e. provide a link on some page to a poi or sth.

f1sh1918 (changes requested):
Hm i would suggestion the follow change. Since now the user will be backwarded even he/she selects and deselects any poi

If you come via link and directly click back-> back button goes back to the page you came. Then there should also be a proper headline of the PageTitle
If you come via link and then deselect and select another POI the back button should deselect the poi

sarahsporck (commented):
I don't like how the state is not preserved. E.g I deep link to a location and deselect and select the place again. Then the state is lost. And you go back to the dashboard when going back.

Not sure what's the best way to handle this. My first idea would be using nested routers probably...

LeandraH:
I spent some time trying to get a nested router to work but found some problems with that:
- If we just put in a router to control the entire page, then going back works beautifully without needing to hack anything into the navigation. But the map unfortunately reloads with every routing and we don't have the nice flyTo effect that we currently have.
- My next idea was to put just the BottomActionsSheet into a new stack but then the navigation in the header doesn't have access to the inner stack anymore, so we'd be back to hacking into that to fix it. At which point I think we may as well leave it the way it is.
So I implemented Andy's suggestions but not Sarah's.

f1sh1918 (commented):
Hm i tried to recreate a deep-link on a site in the testumgebung but it opens the web app. How did you create a deep link on a site that opens maps in the app @LeandraH ?
Or is this a new bug or sth?

Currently i can not test the issue

ztefanie (commented):
I works as expected, but the label of the back button should also be adjusted.

When clicking on a link to a POI, the POI opens correctly, the back button in the header goes back to the previous page (the page with the link), but the headers says "Map", which is not correct.

ztefanie:
> Hm i tried to recreate a deep-link on a site in the testumgebung but it opens the web app. How did you create a deep link on a site that opens maps in the app @LeandraH ? Or is this a new bug or sth?
> 
> Currently i can not test the issue

The link that was present in the testumgebung was to webnext and webnext-links are not handled by the native apps. Other mistake could be to have a different city opened as the deep link opens, this will also open the in app browser.
If both does not help, check your app settings, that the integreat.app links are handled by the integreat app, as this can be disabled by the user.

f1sh1918 (approved):

native/src/components/Header.tsx
```diff
@@ -210,8 +210,12 @@ const Header = ({
 
     const previousParams = previousRoute.params
 
-    // Poi details are not opened in a new route
-    if (route.name === POIS_ROUTE) {
+    if (
+      // Poi details are not opened in a new route
+      route.name === POIS_ROUTE &&
+      // unless coming from a deep link, and then the previous route has no slug
+      !!(previousParams as { slug?: string }).slug
```
steffenkleinle, line 217: Hmm I am not entirely sure if we can make that assumption. A poi detail can also be linked to from a category or event and then the previous route has a slug :/

native/src/routes/Pois.tsx
```diff
@@ -74,6 +74,7 @@ const Pois = ({ pois: allPois, language, cityModel, route, navigation }: PoisPro
   const { slug, multipoi } = route.params
   const [sheetSnapPointIndex, setSheetSnapPointIndex] = useState<number>(1)
   const [listScrollPosition, setListScrollPosition] = useState<number>(0)
+  const [deselectOnBackNavigation, setDeselectOnBackNavigation] = useState(slug === undefined)
```
steffenkleinle, line 77: I think this should be
```suggestion
  const [deselectOnBackNavigation, setDeselectOnBackNavigation] = useState(slug === undefined && multipoi === undefined)
```

native/src/components/Header.tsx
LeandraH: Oh no, you're right! Back to the drawing board, I guess 😭

LeandraH: Oh. Right, I overreacted. The slug trick does still work inside the Pois, just not in the header. Whew

ztefanie: You can just create two boolean variables with meaningful names to get rid of these comments.

ztefanie (approved):

PR #2494, opened 2023-10-03
2031: Combine recurring events web

steffenkleinle:
### Short description

This PR combines recurring events such that they only show up once in the list.

### Proposed changes

- Show an icon in the event list whether they are today, recurring or today and recurring
- Always show the next three dates (if available) and give the option to show the next ten dates by using a collapsible

### Side effects

None.

### Resolved issues

Fixes web part of: #2031.

---


web/src/components/DatesPageDetail.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const hasMoreDates = date.hasMoreRecurrencesThan(MAX_DATE_RECURRENCES)
+  const { t } = useTranslation('events')
+
+  if (dates.length === 1) {
+    return <PageDetail identifier={t('date_one')} information={nextDate} />
+  }
+
+  const Title = <Identifier>{t(hasMoreDates ? 'nextDate_other' : 'date_other')}: </Identifier>
+  const Dates = dates.map(it => <div key={it}>{it}</div>)
+  const StickyDates = <>{Dates.slice(0, MAX_DATE_RECURRENCES_COLLAPSED)}</>
```
LeandraH, line 31: I don't think `StickyDates` is the right name since they move around when I scroll and thus aren't sticky. How about `FirstDates` or `AlwaysShownDates`?

api-client/src/models/DateModel.ts
```diff
@@ -23,8 +40,48 @@ class DateModel {
...
+    )
+  }
+
+  recurrences(count: number): DateModel[] {
+    if (!this.recurrenceRule) {
+      return [this]
+    }
+
+    const now = DateTime.now()
+    const maxDate = now.plus({ years: 5 }).toJSDate()
```
f1sh1918, line 62: maybe you could also use a const for years

LeandraH (approved):
Looks great! Tested in Firefox

web/src/components/ListItem.tsx
```diff
@@ -38,19 +35,30 @@ const FullWidthLink = styled(StyledCleanLink)`
   flex: 1;
 `
 
+const TitleRow = styled.div`
+  display: flex;
+  flex-direction: row;
```
f1sh1918, line 40: row is the default direction. Shouldn't be needed to set

f1sh1918 (commented):
Nice implementation!

I found two issues
1. The start and end times start to change at some point

image

2. I think we should ensure that for mobile the dates fit into one line (maybe we could use short month names for mobile)

LeandraH:
Oh, one more thing, the old recurring links don't work anymore, e.g. http://localhost:9000/testumgebung/de/events/testkonzert$2023-10-09. It's annoying but I think we should keep supporting them for a while longer

steffenkleinle:
> I found two issues
> 
>     1. The start and end times start to change at some point

See #2496, this issue is not a new one. Just now its more visible.

>     2. I think we should ensure that for mobile the dates fit into one line (maybe we could use short month names for mobile)

Good catch, implemented it that way for the EventListItem already but forgot to do so here. Will change :)

f1sh1918:
> > I found two issues
> > ```
> > 1. The start and end times start to change at some point
> > ```
> 
> See #2496, this issue is not a new one. Just now its more visible.
> 
> > ```
> > 2. I think we should ensure that for mobile the dates fit into one line (maybe we could use short month names for mobile)
> > ```
> 
> Good catch, implemented it that way for the EventListItem already but forgot to do so here. Will change :)

But if its more visible should we merge this before #2496 @steffenkleinle ?
Looks very confusing for the user

steffenkleinle:
> > > I found two issues
> > > 1. The start and end times start to change at some point
> > 
> > See #2496, this issue is not a new one. Just now its more visible.
>
> But if its more visible should we merge this before #2496 @steffenkleinle ? Looks very confusing for the user

Yes, added it as blocking as well. Will have a look.

steffenkleinle:
> Oh, one more thing, the old recurring links don't work anymore, e.g. http://localhost:9000/testumgebung/de/events/testkonzert$2023-10-09. It's annoying but I think we should keep supporting them for a while longer

Done.

ztefanie (approved):
Tested on Firefox. Looks good :+1:

LeandraH (approved):
Thanks for fixing that, tested in Firefox. I'd also prefer a little additional padding before the address here, sorry to only notice that now!

image

Issue #2428, opened 2023-09-05
Remove dexing for android builds


sarahsporck:
**Is your feature request related to a problem? Please describe.**
As described in the firebase docs we enabled dexing for android builds. However since Android 21 this is done automatically by the android gradle plugin, so we just add another overhead here. 

**Describe the solution you'd like**
Remove 'androidx.multidex:multidex' from build.gradle and all other ocurrences of `multidex`

**Describe alternatives you've considered**
N/A

**Additional context**
https://developer.android.com/build/multidex

PR #2503, opened 2023-10-05
2428: Remove manual multidexing, Jetifier & deprecated package attribute

sarahsporck:
### Short description

This PR removes some unused configurations from our gradle setup:
- Multidexing is enabled and run by [default since sdk 21](https://developer.android.com/build/multidex#mdex-gradle)
- [jetifier is deprecated](https://github.com/mikehardy/jetifier) & [~it is automatically run by the react-native cli~, so no need to invoke it manually and keep the dependency](https://github.com/react-native-community/cli/pull/1662)... I still don't think that there is a place for jetifier anymore
- [since gradle 7.3 the package name should not be set in the AndroidManifest anymore
](https://developer.android.com/guide/topics/manifest/manifest-element)
### Proposed changes

- remove jetifier
- remove multidexing
- remove package name

### Resolved issues

Fixes: #2428 

---


steffenkleinle (approved):

LeandraH (approved):
I also see `android.enableJetifier=true` in the `gradle.properties`, does that one also have to go?

sarahsporck:
> I also see `android.enableJetifier=true` in the `gradle.properties`, does that one also have to go?

Sadly this cannot go yet, as some of our dependencies still use android.support libraries, and this migrates them to androidx.

Issue #2496, opened 2023-10-03
Wrong time shown for events close to daylight savings time


steffenkleinle:
### Describe the Bug

If an event is in a different timezone (e.g. in daylight savings time), the displayed time is wrong/confusing.

#### Before #2031 is done

1. Go to https://webnext.integreat.app/muenchen/de/events/lern-und-hausaufgabenhilfe-6$2023-10-31
2. See that the time is 13:00 to 16:00
3. Go to https://webnext.integreat.app/muenchen/de/events/lern-und-hausaufgabenhilfe-6$2023-10-24
4. See that the time is 14:00 to 16:00

#### After #2031 is done

1. Go to https://webnext.integreat.app/muenchen/de/events/lern-und-hausaufgabenhilfe-6
2. Expand dates
3. See that up til 24.10 the times are 14:00 til 16:00
4. Starting with the 31.10 the times are 13:00 to 15:00

### Expected Behavior

The time should always be shown in the timezone of the event.

### Actual Behavior

The time is shown in the timezone of the user.

### Additional Information

Closed issues relating to this: #1054 and #1126.
Also relates to #2031 which combines recurring events.

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #2506, opened 2023-10-06
2496: Dates of recurring events wrong if changing to daylight saving time

steffenkleinle:
### Short description

Our rrule doesn't support working with timezones, therefore the time was wrong if the start dates was in DST and the some recurrences not or the other way round. This PR also adds the offset such that times are displayed correctly.

### Proposed changes

Keep the offset of the original start date and add the offset difference to all dates returned by rrule.

### Side effects

None.

### Resolved issues

Fixes: #2496.

---

### DIFF
--- a/api-client/src/models/DateModel.ts
+++ b/api-client/src/models/DateModel.ts
@@ -8,6 +8,7 @@ class DateModel {
   _endDate: DateTime
   _allDay: boolean
   _recurrenceRule: RRuleType | null
+  _offset: number
 
   constructor({
     startDate,
@@ -20,14 +21,15 @@ class DateModel {
     allDay: boolean
     recurrenceRule: RRuleType | null
   }) {
-    const firstRecurrence = recurrenceRule?.after(DateTime.now().toJSDate())
-    const duration = endDate.diff(startDate)
+    this._recurrenceRule = recurrenceRule
+    this._offset = startDate.offset
     this._allDay = allDay
+    const duration = endDate.diff(startDate)
     // If there is a recurrence rule, the start and end dates are not updated in the CMS and are therefore most of the time outdated
     // Therefore calculate the (next) correct start and end date based on the recurrence rule if available
-    this._startDate = firstRecurrence ? DateTime.fromJSDate(firstRecurrence) : startDate
-    this._endDate = firstRecurrence ? DateTime.fromJSDate(firstRecurrence).plus(duration) : endDate
-    this._recurrenceRule = recurrenceRule
+    const first = recurrenceRule?.after(DateTime.now().toJSDate())
+    this._startDate = first ? this.rruleToDateTime(first) : startDate
+    this._endDate = this._startDate.plus(duration)
   }
 
   get startDate(): DateTime {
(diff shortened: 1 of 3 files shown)


steffenkleinle:
Please test this thoroughly, I am not really sure if this is a good/correct solution! Timezones mess with my head... :exploding_head:

f1sh1918 (approved):
looks fine :)

release-notes/unreleased/2496-wrong-dst.yml
```diff
@@ -0,0 +1,8 @@
+issue_key: 2438
+show_in_stores: true
+platforms:
+  - android
+  - ios
+  - web
+en: Fixed wrong dates caused by daylight savings time.
+de: Falsch angezeigte Daten durch Zeitverschiebung behoben.
```
ztefanie, line 8: This translation is nor really correct. "daylaight saving time" is "Sommerzeit". "time shift" would be "Zeitverschiebung".

api-client/src/models/DateModel.ts
```diff
@@ -99,6 +101,14 @@ class DateModel {
     return `${localizedStartDate} - ${localizedEndDate.toFormat(format)}`
   }
 
+  private rruleToDateTime(date: Date): DateTime {
+    const dateTime = DateTime.fromJSDate(date)
+    // rrule is not correctly handling timezones and always returning local time in the shape of UTC
+    // Therefore we have to manually apply timezone offset, e.g. for daylight savings time
+    // https://github.com/jkbrzt/rrule#important-use-utc-dates
```
ztefanie, line 108: This link suggest to handle this with luxon, like this:
```
rule.all().map(date =>
DateTime.fromJSDate(date)
  .toUTC()
  .setZone('local', { keepLocalTime: true })
  .toJSDate()
)
```

Why have you decided to handle it like you did?

LeandraH (commented):
[This](https://github.com/jkbrzt/rrule#timezone-support) sounds like we can add in the `tzid` parameter to get a timezone from Rrule, probably in the constructor. Or have you tried that already?

steffenkleinle:
> [This](https://github.com/jkbrzt/rrule#timezone-support) sounds like we can add in the `tzid` parameter to get a timezone from Rrule, probably in the constructor. Or have you tried that already?

I am pretty sure that this is sadly also not what we want. Its just a way to tell rrule which timezone the given date of the rule is in since you should always pass UTC dates.
From the linked [RFC](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.19):
> This parameter MUST be specified [...] when the value is  neither a UTC or a "floating" time.

In our case, however, the rules we get from the CMS are using UTC already. The problem is not the input of the rrule library but instead the output which uses the current timezones offset which breaks logic if some recurrence is affected by DST.

release-notes/unreleased/2496-wrong-dst.yml
steffenkleinle: Will use `Zeitumstellung`.

ztefanie (approved):

PR #2511, opened 2023-10-09
2461: Add husky

ztefanie:
### Short description

Added husky and fixed pre commit script

### Proposed changes

I adjusted the script as it has some errors, please check if the script behaves correct for you:
- If changes in the circleci config are made, the script should check for installation of circleci and fail if it is not installed. If it is installed it should generate and check the config
- prettier should be run for the changed files.
- if no files are changed prettier should not be run

### Resolved issues

Fixes: #2461 

---


ztefanie:
> Nice! I managed to get something that prettier can fix fixed by committing it instead :)
> 
> We also talk about installing the pre-commit hook in our Readme, could you update that?

There is no need to "install" the pre-commit hook manually. It is installed in the `prepare` script, which is automatically executed after `yarn install`.

package.json
```diff
@@ -21,6 +21,7 @@
     ]
   },
   "scripts": {
+    "prepare": "husky install",
```
sarahsporck, line 24: Sidenote: For this `yarn install` has to be run in the root directory.

sarahsporck (approved):
Thank you very much for this PR 💯 !

LeandraH:
> > Nice! I managed to get something that prettier can fix fixed by committing it instead :)
> > We also talk about installing the pre-commit hook in our Readme, could you update that?
> 
> There is no need to "install" the pre-commit hook manually. It is installed in the `prepare` script, which is automatically executed after `yarn install`.

Exactly, so we should not tell people to install the pre-commit in our Readme :) I'll open a ticket for it.

Issue #2501, opened 2023-10-05
Improve eslint config


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Eslint is our tool to check correct formating and ensure code consistency. But our current eslint configuration is awfully slow. While working on #2484 I realized there are a few fixes that could improve the overall linting performance.

**Describe the solution you'd like**
- enable `reportUnusedDisableDirectives: true` and get rid of all unnecessary eslint inline configs
- remove as many globs as possible from the eslint file. Especially the ignore patterns take around 30ms per file to evaluate (run `lint --debug` to see this). I think it is much quicker to lint files such as the `stylelint.config.js` as to ignore those. 
- check out https://typescript-eslint.io/linting/troubleshooting/performance-troubleshooting to see what rules are already checked by typescript and disable those and enable their typescript counterparts
- enable caching for running eslint locally (this speeds eslint up by 💯%) and remove eslint changed (as caching does basically the same) 

**Describe alternatives you've considered**
N/A

PR #2515, opened 2023-10-10
2501: Improve eslint performance

sarahsporck:
### Short description

In #2484 I found out that there are a few things that could speed up linting locally. This is implemented in this PR.

### Proposed changes

- [Use .eslintcache](https://eslint.org/docs/latest/use/command-line-interface#caching) (not in ci though ;>)
- Remove and report unused eslint directives
- removed all unneccessary globs in `eslinrc.js`

### Side effects

- It could have happened that I accidentally removed or added a glob too much. would be nice if you could pay attention to that during review
- technically [the camelcase rule is now disabled for object properties](https://eslint.org/docs/latest/rules/camelcase#properties-always). But I don't think that enabling makes much sense, as there are a lot of situations where we actually want the rule to be turned off

### Resolved issues

Fixes: #2501 

---


web/tools/webpack.config.ts
```diff
@@ -124,14 +123,10 @@ const createConfig = (
   // If version_name is not supplied read it from version file
   const versionName = passedVersionName || readVersionName()
   const shortCommitSha = passedCommitSha?.substring(0, SHORT_COMMIT_SHA_LENGTH) || 'Commit SHA unknown'
-  // eslint-disable-next-line no-console
```
steffenkleinle, line 127: Why are a lot of these not necessary anymore?

sarahsporck: tools has different linting rules. See eslint overrides: 
```ts
{
     files: ['**/tools/**', 'translations/**', 'e2e-tests/**'],
      rules: {
        'no-console': 'off',
        'import/no-extraneous-dependencies': 'off',
     }
}
```

steffenkleinle: Ahh nvm, stupid question. Sorry.

steffenkleinle (approved):

api-client/src/models/PoiModel.ts
```diff
@@ -118,7 +118,6 @@ class PoiModel extends ExtendedPageModel {
     return (
       this._category ??
       new PoiCategoryModel({
-        // eslint-disable-next-line no-magic-numbers
```
sarahsporck, line 121: I'm actually a little bit perplex why no-magic-numbers does not trigger here o.o

steffenkleinle: Maybe `id` is some kind of ignore pattern now?

steffenkleinle: https://eslint.org/docs/latest/rules/no-magic-numbers#detectobjects

That looks like it could be the reason? Its not active as default and therefore doesn't check objects if I understand correctly. Not sure why it triggered an error before then.

sarahsporck: Ahh. [This](https://eslint.org/docs/latest/rules/no-magic-numbers#detectobjects) needs to be set to true.

sarahsporck: I guess the option was added at some point and we didn't notice. I'll enable it again. 👍🏼

eslintconifg.json
```diff
@@ -0,0 +1,3269 @@
+yarn run v1.22.19
+$ /Users/digitalfabrik/projects/integreat-app/node_modules/.bin/eslint --print-config .eslintrc.js
+{
```
ztefanie, line 3: This should not be commited, should it?

ztefanie (approved):
It makes me unhappy to see all these `es-lint disable no-magic-numbers` comments. If we want this rule on error level, we should fix the errors by adding named constants. If we do not want to enforce no-magic-numbers, we should not enable the rule. Why do you think this is the best solution?

ztefanie (changes requested):
I think this is not working. If I add a linting error to e.g. `api-client/src/models/DateModel.ts` I get no error for running `yarn lint:ci` nor for `yarn lint` in root directory. Only if running `yarn lint` in subfolder `api-client` linting is failing.

sarahsporck:
> I think this is not working. If I add a linting error to e.g. `api-client/src/models/DateModel.ts` I get no error for running `yarn lint:ci` nor for `yarn lint` in root directory. Only if running `yarn lint` in subfolder `api-client` linting is failing. Edit: So our CI will not catch this linting errors.

It seems that with specifying `!(tools/**)` I matched all directories, BUT the tools directory which turned off two linting rules. dDid you by chance try adding a `console.log`? Then this should be fixed. I also have had removed the `.` in the command, which lead to a situation where lint didn't select any file, which could also have been the issue. Would you mind testing this again?

ztefanie:
Yes comments work as expected now, but I think it would be nicer if the command outputs the files that have linting errors instead of returning nothing. Writing to reports does not really make sense, as if you do not save it or am I missing something?

sarahsporck:
> Yes comments work as expected now, but I think it would be nicer if the command outputs the files that have linting errors instead of returning nothing. Writing to reports does not really make sense, as if you do not save it or am I missing something?

You are completely right. I changed it so we actually print out lint results. Also the junit report is stored as test results in the ci now. Further I realized, that for quite some time linting didn't check any files in the CI. As we did not pass a glob to the eslint command. o.o

eslintconifg.json
ztefanie: Why do we need to commit this file? I thought it is just a translation from our `.eslintrc.js` to a `.json` file. And everything in this file is just duplicated. Is it used somewhere? Wo we really need both config files?

ztefanie:
> It makes me unhappy to see all these `es-lint disable no-magic-numbers` comments. If we want this rule on error level, we should fix the errors by adding named constants. If we do not want to enforce no-magic-numbers, we should not enable the rule. Why do you think this is the best solution?

Have you seen this comment? What are your thoughts on this?

eslintconifg.json
sarahsporck: Nope. This was added accidentally when I checked what rules excactly are emitted. Removed it again. Thank you for noting

sarahsporck:
Oh no! I missed that. Hm. I personally think it is good to keep the rule. I at least remember times where I was tempted to use "magic numbers" but decided against it because of this linting rule. I'm fine with both, honestly. Therefore, I'll decide for adding constants.

steffenkleinle:
I would also like to keep the rule! We can also do a separate (good first issue) ticket to introduce new constants and remove the magic numbers and disable linting commands.

LeandraH:
I am also in favor of keeping the rule. It's a good reminder to not use magic numbers, and can easily be disabled if we really need one.

sarahsporck:
> I would also like to keep the rule! We can also do a separate (good first issue) ticket to introduce new constants and remove the magic numbers and disable linting commands.

I do think however that the rule does actually make a lot sense for objects, where each property already has a name. Are you fine with disabling the object rule again?

ztefanie (approved):
Created the ticket for no-magic-numbers constant. 

Thanks for this pr :+1:

Issue #2473, opened 2023-09-25
Remove the Wohnraumbörse


LeandraH:
**Is your feature request related to a problem? Please describe.**
The last municipality just cancelled the Wohnbörse.

**Describe the solution you'd like**
The code for the Wohnbörse can be deleted. Should we need it in the future, we can revert the PR. So if there is important information necessary to do that, we should also document that in the PR or this issue.

**Describe alternatives you've considered**
Leave it in but it's extra code to maintain and ship.

PR #2518, opened 2023-10-11
2473: Remove shelter offer

steffenkleinle:
### Short description

Remove the shelter offer which is not used by any city anymore.

### Proposed changes

Remove the route, all components, endpoints, types, translations and constants of the shelter offer.

### Side effects

The wohnraumbörse offer does not work anymore. If we want to revert that, we can just revert the commit/PR.

### Resolved issues

Fixes: #2473.

---


f1sh1918 (approved):
Looks nice.
I think this translation could also be deleted?

`"wohnraum": "Wohnraum"`

PR #2519, opened 2023-10-11
1317: Make local news offline available

steffenkleinle:
### Short description

Make local news offline available.

### Proposed changes

- Add methods to store, load or check the availability of local news to the `DatabaseConnector` and the `DefaultDataContainer`
- Add loading of local news to `useLoadCityContent`
- Use city content data instead of `useLocalNews`
- Delete now unused `useLocalNews`

### Side effects

None.

### Resolved issues

Fixes: #1317.

---


f1sh1918 (approved):
Tested on android emulators works nice :)

ztefanie (commented):
Tested on Android, works as expected.
Code looks good, but I think for news it makes sense to update more often then once a day. Maybe adding a refresh on focus the news tab makes sense, when internet is availabe.

native/src/utils/DatabaseConnector.ts
```diff
@@ -510,6 +517,34 @@ class DatabaseConnector {
     return this.readFile(path, mapPoisJson)
   }
 
+  async storeLocalNews(localNews: LocalNewsModel[], context: DatabaseContext): Promise<void> {
+    const jsonModels = localNews.map(
+      (it: LocalNewsModel): ContentLocalNewsJsonType => ({
```
LeandraH, line 522: You don't need to set the type of the `localNews` again, it's already set in the argument typing two lines earlier.

LeandraH (approved):
Nice, tested in emulated iOS. I seem to remember 24h being the requested time but I agree with Steffi that it feels a bit long for news. I like the idea of the refresh when the News tab is opened with an internet connection.

steffenkleinle:
> Tested on Android, works as expected. Code looks good, but I think for news it makes sense to update more often then once a day. Maybe adding a refresh on focus the news tab makes sense, when internet is availabe.

@ztefanie good idea! Added refreshing :)

ztefanie (approved):

PR #2521, opened 2023-10-12
1233: Implement lists as actual list

lunars97:
### Short description

Implement lists as actual list for accessibility according to **Prüfbericht
zur Barrierefreiheit**

### Proposed changes

- Searchresults
- “Integreat deiner Region vor!”

### Resolved issues

Fixes: #2167 

---


steffenkleinle (changes requested):
This issue is for both web and native and I think there is quite a few places where we don't use lists. Probably we shouldn't use lists in all these cases, but just examples (all found by searching for .map() functions in the code):
* native: Categories.jsx -> Footer
* native: CalendarChoiceModal
* native: CategoryListItem -> rendering of sub category list items
* native: Selector
* native: OpeningHours
* native: PoiFilters
* web: CitySelector
* web: OpeningHours
* web: PoiFilters
* web: SearchPage
* web: Selector

So probably we should split up the issue and do some more research in which cases it makes sense and in which cases it does not make sense.

native/src/routes/CityNotCooperating.tsx
```diff
@@ -53,30 +57,51 @@ const StepExplanation = styled.Text`
...
+  explanation: string
+}
+
 const CityNotCooperating = (): ReactElement | null => {
   const { t } = useTranslation('cityNotCooperating')
   const [isCopied, setIsCopied] = useState<boolean>(false)
   const template = buildConfig().featureFlags.cityNotCooperatingTemplate
   const CityNotCooperatingIcon = buildConfigAssets().CityNotCooperatingIcon
 
+  const steps: Step[] = [
```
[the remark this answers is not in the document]
LeandraH: I'd also think that it's a bit much but it is one of the things that has been explicitly criticized in the report from Pfennigparade.

native/src/components/SearchListItem.tsx
```diff
@@ -100,6 +102,8 @@ const SearchListItem = ({
       </DirectionContainer>
     </FlexStyledLink>
   )
+
+  return <List items={category} renderItem={({ item: category }) => category} Header={SearchList} />
```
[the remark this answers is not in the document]
LeandraH: The search results have been named as an example of where we didn't implement a list properly, so there are still things we need to do here. Might be worth looking at it with a screenreader.

native/src/routes/CityNotCooperating.tsx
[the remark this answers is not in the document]
steffenkleinle: This exact example was criticized? I think we should discuss this, we shouldn't just do everything if we don't think it makes sense.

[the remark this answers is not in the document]
lunars97: @LeandraH @steffenkleinle  should I leave it for now as it is?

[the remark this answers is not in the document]
steffenkleinle: Yes, lets wait for more opinions here. Maybe we should discuss this in a call. Sorry for that!

LeandraH:
Let's have a chat about this in Monday's Weekly?

native/src/routes/CityNotCooperating.tsx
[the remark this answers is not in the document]
ztefanie: Yes this exact example was mentioned. I am fine with keeping it as if is or with changing it to an actual list.

native/src/components/SearchListItem.tsx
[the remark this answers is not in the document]
lunars97: I am currently inserting FlatList to native: **Categories.jsx -> Footer.** Should I open for it another issue or can I push changes here? I am a bit confused right now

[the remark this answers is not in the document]
lunars97: @LeandraH SearchResults are working good with a screen reader. Should I leave it then?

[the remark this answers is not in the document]
LeandraH: Wait, so you couldn't reproduce the issues named in the report? In https://nextcloud.tuerantuer.org/index.php/apps/files/?dir=/2023-10%20Aktion%20Mensch_Barrierefreiheit%20f%C3%BCr%20alle_bis%20zu%20350k&openfile=4955578 under "Listenartige Elemente sind fehlerhaft implementiert"?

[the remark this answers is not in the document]
LeandraH: Why don't we keep this PR for native changes and open another issue for checking over the web stuff?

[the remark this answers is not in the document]
lunars97: @LeandraH It says that we need to include announcement to search results.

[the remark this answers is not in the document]
LeandraH: Yes, announcements of how long the list is, which list entry we are focusing on, and that we've reached the end of the list. Did I understand correctly that you tested these with a screen reader on the search results and found all of them?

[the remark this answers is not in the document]
lunars97: @LeandraH no, there is no announcement of list length. I have been looking for a solution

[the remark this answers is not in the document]
lunars97: Can I for now implement all cases you mentioned above @steffenkleinle here in this PR?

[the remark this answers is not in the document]
lunars97: @LeandraH seems like it should announce automatically. But so far I was not able to reproduce it with emulators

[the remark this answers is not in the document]
LeandraH: That's too bad! I don't think it makes sense to implement the other cases until we know how to do that properly, though. Do you have any other ideas how to get the announcements in there?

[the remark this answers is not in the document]
LeandraH: Thank you :)

native/src/routes/SearchModal.tsx
```diff
@@ -102,6 +102,7 @@ const SearchModal = ({
       contentWithoutHtml={item.contentWithoutHtml}
       language={languageCode}
       query={query}
+      accessibilityHint={t(`It navigates to the description of the item`)}
```
ztefanie, line 105: This should be moved to our translations.json file, so the users get this output in their language.
Also i think "description" is not really the correct term, maybe "page" is better.
Better place to put this, is in the `SearchListItem.tsx` as all searchListItems have the same hint, no need to pass it around with props.

native/src/routes/CityNotCooperating.tsx
```diff
@@ -3,19 +3,22 @@ import React, { ReactElement, useState } from 'react'
 import { useTranslation } from 'react-i18next'
```
ztefanie, line 3: Please revert this file.

web/src/routes/CityNotCooperatingPage.tsx
```diff
@@ -13,7 +13,7 @@ const Container = styled.div`
   flex-direction: column;
```
ztefanie, line 13: Please revert this file.

native/src/components/base/TextButton.tsx
```diff
@@ -25,12 +25,31 @@ type TextButtonProps = {
   text: string
   onPress: () => Promise<void> | void
   disabled?: boolean
+  accessible?: boolean
+  focusable?: boolean
+  accessibilityRole?: AccessibilityRole | undefined
```
ztefanie, line 30: No need to have the `undefined` here, as this prop is optional, because of the question mark, the `undefined` type is added automatically.

Also there is no need to have the `accessibilityRole` as a prop here, as we are in a Button component, the accessibilityRole should always be "button" here. So just add below `accessibilityRole='button'` and remove the option to pass an other prop.

native/src/components/base/TextButton.tsx
```diff
@@ -25,12 +25,31 @@ type TextButtonProps = {
   text: string
   onPress: () => Promise<void> | void
   disabled?: boolean
+  accessible?: boolean
+  focusable?: boolean
```
ztefanie, line 29: Please remove the `accessible` and `focusable` props from this component, as you do not use them and they are set by default for every `Pressable`, see the React Native docs: https://reactnative.dev/docs/accessibility#accessible

native/src/components/__tests__/SearchListItem.spec.tsx
```diff
@@ -14,6 +14,7 @@ jest.mock('react-native-webview', () => ({
 describe('SearchListItem', () => {
```
ztefanie, line 14: This file can be reverted, when you move the accessibilityHint inside the SearchListItem.tsx. See my comment in `SearchModal.tsx`

native/src/components/SearchListItem.tsx
```diff
@@ -46,6 +46,7 @@ type SearchListItemProps = {
   contentWithoutHtml: string
```
ztefanie, line 46: This file can be reverted, see my comment on the file `SearchModal.tsx`

native/src/components/List.tsx
```diff
@@ -11,10 +11,12 @@ const NoItemsMessage = styled.Text`
 
 type ListProps<T> = {
   items: Array<T>
-  noItemsMessage: ReactElement | string
+  noItemsMessage?: ReactElement | string
```
ztefanie, line 14: Why have you changed this to be optional?

native/src/components/List.tsx
```diff
@@ -11,10 +11,12 @@ const NoItemsMessage = styled.Text`
...
   renderItem: (props: { item: T; index: number }) => ReactElement
   Header?: ReactElement
   Footer?: ReactElement
+  scrollEnabled?: boolean
+  accessibilityRole?: AccessibilityRole | undefined
```
ztefanie, line 19: No need to pass the `accessibilityRole` here as a prop. This component has a FlatList inside, which should always have the `accessibiltyRole="grid"`, see [RN docs](https://reactnative.dev/docs/accessibility#accessibilityrole). This will also add the "the in/out of grid announcements to the android GridView.", e.g. "four elements in the list"

native/src/components/List.tsx
```diff
@@ -45,6 +49,8 @@ const List = <T,>({
       paddingHorizontal: 10,
     }}
     onEndReachedThreshold={1}
+    scrollEnabled={scrollEnabled}
+    accessibilityRole={accessibilityRole}
```
ztefanie, line 53: Just set it to the value `"grid"` here.

native/src/components/LayoutedScrollView.tsx
```diff
@@ -7,18 +7,20 @@ import Layout from './Layout'
 type LayoutedScrollViewProps = {
   children?: React.ReactNode
   refreshControl?: React.ReactElement<RefreshControlProps>
+  nestedScrollEnabled?: boolean
```
ztefanie, line 10: Why is this change needed?

native/src/components/CategoryListItem.tsx
```diff
@@ -82,15 +83,19 @@ const CategoryListItem = ({
...
-    {subCategories.map(subCategory => (
-      <SubCategoryListItem
-        key={subCategory.path}
-        subCategory={subCategory}
-        resourceCache={resourceCache[subCategory.path] ?? {}}
-        onItemPress={onItemPress}
-        language={language}
-      />
-    ))}
+    <List
```
ztefanie, line 86: Maybe you can rename this file and component to `CategoryList.tsx`, as this component does not render one listItem, as the name says, but the whole list. This was the case before your changes, but maybe you can nevertheless adjust it.

ztefanie (changes requested):
As Steffen suggested, it is a little over-engineered to introduce a list for only the two items int the "CityNotCooperation"-View, so please revert these two files. Sorry about that, it was originally mentioned in the ticket, but sometimes we realize the best solution, not until we review.

native/src/components/LayoutedScrollView.tsx
lunars97: @ztefanie while there was a virtual list error saying that `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality`

native/src/components/List.tsx
lunars97: because there is always a ts error saying that this prop is missing or should I include it everywhere?

native/src/routes/SearchModal.tsx
lunars97: @ztefanie if I put it inside `SearchListItem` it causes error

native/src/components/__tests__/SearchListItem.spec.tsx
lunars97: @ztefanie I was unable to insert it into SearchListItem. It caused an undefined error

native/src/routes/SearchModal.tsx
ztefanie: Not sure which error you get, but here is the solution i suggest: https://github.com/digitalfabrik/integreat-app/pull/2583/files

lunars97: @ztefanie I also did the same thing, let me try it again :)

ztefanie (approved):
Looks good :tada: Nice work. 

Tested on Android, works fine.

If you fix something we mentioned in a comment, you can click "resolve conversation", so it is easy to see, which conversations are open and need e.g. an answer and which one are "done".

Also please squash the commits into one commit.

It is really nice to see, how you never give up and always make your PRs nice and functioning code. Well done :clap:

native/src/components/LayoutedScrollView.tsx
```diff
@@ -15,6 +15,7 @@ const LayoutedScrollView = (props: LayoutedScrollViewProps): ReactElement => {
     <Layout>
       <ScrollView
         keyboardShouldPersistTaps='always'
+        nestedScrollEnabled
```
steffenkleinle, line 18: why do we need that?

native/src/routes/SearchModal.tsx
```diff
@@ -113,6 +114,7 @@ const SearchModal = ({
         <List
           items={searchResults}
           renderItem={renderItem}
+          accessibilityLabel={`${searchCategories(categories, query).length} ${t('searchResults')}`}
```
steffenkleinle, line 117: This should go to translations.json with proper plural handling and using a placeholder for the number.

web/src/routes/CityNotCooperatingPage.tsx
```diff
@@ -119,5 +119,4 @@ const CityNotCooperatingPage = ({ languageCode }: CityNotCooperatingPageProps):
     </Layout>
   )
 }
-
```
steffenkleinle, line 122: Please readd

native/src/components/SearchListItem.tsx
```diff
@@ -57,6 +59,7 @@ const SearchListItem = ({
   contentWithoutHtml,
   onItemPress,
   query,
+  t,
```
steffenkleinle, line 62: Instead of passing this you should better use the useTranslation hook.

translations/translations.json
```diff
@@ -8401,6 +8401,8 @@
     "de": {
       "nothingFound": "Es wurden leider keine passenden Ergebnisse gefunden.",
       "searchPlaceholder": "Suche",
+      "itemHint": "Es navigiert zur Beschreibung der Seite",
```
steffenkleinle, line 8404: Optional: Not sure about this but I think something like that would be better:
```suggestion
      "itemHint": "Suchergebnis öffnen",
```

native/src/components/CategoryListItem.tsx
steffenkleinle: @ztefanie I think this is not correct. This renders one list item which, in turn, includes a list of sub category items. Therefore the name should stay the way it is.

native/src/components/LayoutedScrollView.tsx
lunars97: @steffenkleinle yep, this exact error

steffenkleinle: But do you have a resource that says that this `nestedScrollEnabled` fixes the error? Or did you just try things until you found a solution?

lunars97: @steffenkleinle many developers mentioned this solution when I was searching for an answer why

lunars97: [Here is the link](https://github.com/facebook/react-native/issues/31697)

native/src/components/Categories.tsx
```diff
@@ -81,7 +81,7 @@ const Categories = ({
       path={category.path}
       AfterContent={category.organization && <OrganizationContentInfo organization={category.organization} />}
       Footer={children.map(it => (
```
steffenkleinle, line 83: So in theory we have a list of lists here, there is multiple categories that are mapped to `SubCategoryList`s. In theory the childrens should also be rendered in a list. Can also be done in a separate issue though.

native/src/components/SubCategoryList.tsx
```diff
@@ -52,35 +52,35 @@ export const CategoryThumbnail = styled(SimpleImage)<{ language: string }>`
...
   subCategories,
   resourceCache,
   onItemPress,
-}: CategoryListItemProps): ReactElement => (
+}: SubCategoryListProps): ReactElement => (
   <>
     <FlexStyledLink onPress={() => onItemPress({ path: category.path })}>
       <DirectionContainer language={language}>
-        <CategoryEntryContainer>
+        <SubCategoryEntryContainer>
```
steffenkleinle, line 73: Sorry but these renamings are wrong :see_no_evil: The sub categories are only rendered below (lines 86 until 98). If I think more about it, the name of this should actually stay `CategoryListItem`. The component renders a category list item. The fact that a part of that is that it renders sub categories is just an implementation detail and not really relevant to the usage... Sorry for the confusion, my mistake!

translations/translations.json
```diff
@@ -8412,8 +8412,9 @@
     "de": {
       "nothingFound": "Es wurden leider keine passenden Ergebnisse gefunden.",
       "searchPlaceholder": "Suche",
-      "itemHint": "Es navigiert zur Beschreibung der Seite",
-      "searchResults": "Suchergebnisse",
+      "itemHint": "Suchergebnis öffnen",
+      "searchResult": "{{count}} Suchergebnis",
+      "searchResult_other": "{{count}} Suchergebnisse",
```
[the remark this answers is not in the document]
steffenkleinle: I think if we want to support that we should add all the suffixes in all the german and englisch translations like this:
``` json
  "searchResultsCount_zero": "0 Suchergebnisse",
  "searchResultsCount_one": "1 Suchergebnis",
  "searchResultsCount_two": "2 Suchergebnisse",
  "searchResultsCount_few": "{{count}} Suchergebnisse",
  "searchResultsCount_many": "{{count}} Suchergebnisse",
  "searchResultsCount_other": "{{count}} Suchergebnisse"
```
And for most languages they will all be the same, just for arabic and a few others translators than have to change some.

native/src/components/Categories.tsx
steffenkleinle: If you don't want to do it in this PR we can just open a new issue for this :)

lunars97: @steffenkleinle I implemented it :)

steffenkleinle (approved):
Tested on android, well done!

PR #2526, opened 2023-10-17
2026: Better tsconfig breakdown

sarahsporck:
### Short description

One idea of yarn > 2 includes not hoisting dependencies. This means that e.g. jest/typescript/eslint is not hoisted to child dependencies. However we also don't want (and most of the time don't need) to install these dependencies in every workspace. Instead yarn berry allows to use global scripts, which can be accessible from anywhere. To ease the workflow with those global scripts it often helps alinging the configs to their framework specific monorepo setup. 
For typescript [this is described in the docs here](https://www.typescriptlang.org/docs/handbook/project-references.html#an-example-project).

### Proposed changes

- Each tsconfig.json specifies exactly what files it includes
- Every reference (in a tsconfig file) will only be built if it has not yet been built before
  - This speeds up `tsc --build` command visibly

### Side effects

- I had trouble with eslint running into oom in #2484. This is a known issue and can be fixed by removing `enforceConsistentCasingInFilenames` from the tsconfig and adding `--max-old-space-size=1535` to the lint:ci command. If I see the builds failing for this PR I'll add this change.

### Resolved issues

Relates: #2026 

---


native/tsconfig.json
```diff
@@ -20,7 +18,19 @@
...
+    "jest.setup.ts",
+    "index.ts",
+    "jest.config.ts",
+    "babel.config.js",
+    "metro.config.ci.js",
+    "metro.config.js",
+    ".yarn/plugins/root-command.js",
+    "react-native.config.js",
+    "run.js",
+    "src/assets/licenses.json"
```
steffenkleinle, line 32: Why do we need to explicitly include that? First, it is json and second it should be included with `src/**/*` already, shouldn't it?

tools/tsconfig.json
```diff
@@ -8,8 +8,10 @@
     "noUncheckedIndexedAccess": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "outDir": "dist"
   },
+  "include": ["*"],
```
steffenkleinle, line 14: Why can we just do that here but not in e.g. native?

sarahsporck: in tools there is only a flat list of files, e.g. `next-version.ts`, `manage-metadata.ts`. In native we have a directory and additionally some other files, therefore the glob becomes more complicated. But in short it should be prefered to use more specific globs, as (especially in native) a lot of irrelevant nested directories would be scanned for ts/tsx/js/.. files. This is definitely discouraged.

steffenkleinle (approved):
Tested, nice!

LeandraH (approved):
Very cool, and I can finally run `yarn ts:check` in the main directory 🥳

PR #2527, opened 2023-10-17
2026: Use cosmiconfigs

sarahsporck:
### Short description

With yarn `pnp` there are no node_modules anymore. Instead the dependencies are stored as zip's in the project directly. Check out the changed files in #2484 . To use those dependencies in the config files, we can use cosmiconfig files (javascript configuration files).

### Proposed changes

- convert affected config files to cosmiconfig format

### Resolved issues

Relates: #2026 

---


ztefanie:
Is this really ready for review? Seems to depend on #2484 which is still draft? Why is the base main?
Sorry I am a bit confused here :-/

steffenkleinle (approved):

steffenkleinle:
> Is this really ready for review? Seems to depend on #2484 which is still draft? Why is the base main? Sorry I am a bit confused here :-/

I am guessing changing to these cosmiconfigs works with our current setup as well but with yarn v3/pnp it is necessary?

sarahsporck:
exactly, but I'm fine with postponing/restructuring and doing it later.

PR #2528, opened 2023-10-17
2026: Use global jest configuration

sarahsporck:
### Short description

This one has a similar reasoning as #2526 . We can basically use jest projects, to combine global jest configs, run all tests together and simplify ci

### Proposed changes

- use jest projects
- combine global jest configs (such as coverage)
- run all tests together in ci

### Resolved issues

Relates: #2026 

---


ztefanie (approved):
Thanks for this PR, i like it.

As on `main` files not being covered by any test like e.g. `native/src/components/base/ChipButton.tsx` are not in the report. Is there a way to have them in the report with a coverage of 0%?

LeandraH (approved):

sarahsporck:
I added `collectCoverageFrom` glob pattern that now specifies that all `src` folders need to be covered. Nice find!

Issue #2541, opened 2023-11-06
Can't add an event to my calender


Docmale:
### Describe the Bug

Can't add an event to my calender.
Tried on android 11 (tablet), 9 (handy) and 13 (handy).
Not jet tested on ios.

On Win11 the export iCal gives:

BEGIN:VCALENDAR
VERSION:2.0
PRODID:Integreat
BEGIN:VEVENT
DTSTAMP:2023110611:24112400
UID:3e40a8d9-c5d7-52f9-a01e-6341376bff22
SUMMARY:Concert
DTSTART;TZID=[object Object]:2023113020:00200000
DTEND;TZID=[object Object]:2023113022:30223000
DESCRIPTION:single Konzert\n\nhttps://beta.integreat.app/testumgebung-reg/en/events/concert
LOCATION:Elbphilharmonie, Platz der Deutschen Einheit 4, 20457 Hamburg
END:VEVENT
END:VCALENDAR

Importing this to my calendar it says OK but no event appears (Am I doing something wrong?)

### Steps to Reproduce

1. Go to 'Testumgebung Regression'
2. Click on 'Events' than 'Testkonzert'
3.  Click on 'Zum Kalender hinzufügen'
4. Click your calendar
5. See error 'Ein Fehler ist aufgetreten'

Docmale:
Now tried on ios: export works fine event is correcly imported

PR #2542, opened 2023-11-06
2541: Fix event export

LeandraH:
### Short description

This fixes the event export and adds monitoring for it.

### Proposed changes

- For the web export, the format didn't escape the letter 'T' in the middle of the DateTime, so we got an additional localized 24-hour time in there instead of a T, e.g. getting `2023110613:17131700` instead of `20231106T131700`.
- For the Android export, we were also using a wrong format. There was an additional problem here: for allDay events, the start and end have to be at midnight. The dates that we get from the API are at midnight for the start and at 23:59:00 for the end. I added a minute and set the times to be at midnight.
- Another Android issue: in Android 13 and 14, an exported calendar event is added to the Google Calendar and then added to the calendar app at the next sync. So, usually within a minute but not within a second. This lead to an error when trying to open the event from the link in the Snackbar so I deleted that link. Not linking to the event seems to also be the default Android behavior.
- One more Android issue: Android split the permissions for the calendar into a read and a write one.
- And one iOS issue: Apple changed its calendar permissions for iOS 17 which is why I switched to the latest rn-permissions to request those. rn-calendar-events is not updated anymore.

### Side effects

- Updated rn-permissions to a beta version.
-

### Resolved issues

Fixes: #2541 

---

### DIFF
--- a/native/src/components/ExportEventButton.tsx
+++ b/native/src/components/ExportEventButton.tsx
@@ -8,6 +8,7 @@ import styled from 'styled-components/native'
 import { EventModel } from 'api-client'
 
 import useSnackbar from '../hooks/useSnackbar'
+import { reportError } from '../utils/sentry'
 import CalendarChoice from './CalendarChoiceModal'
 import TextButton from './base/TextButton'
 
@@ -56,8 +57,9 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
       } else {
         showSnackbar({ text: 'generalError' })
       }
-    } catch {
+    } catch (e) {
       showSnackbar({ text: 'generalError' })
+      reportError(e)
     }
   }
 
@@ -86,8 +88,9 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
     } else if (editableCalendars.length === 1) {
       try {
         await exportEventToCalendar(editableCalendars[0]?.id)
-      } catch {
+      } catch (e) {
         showSnackbar({ text: 'generalError' })
+        reportError(e)
       }
     } else {
       setCalendars(editableCalendars)
@@ -99,8 +102,9 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
     setShowCalendarChoiceModal(false)
     try {
       await exportEventToCalendar(id)
-    } catch {
+    } catch (e) {
       showSnackbar({ text: 'generalError' })
+      reportError(e)
     }
   }
 
--- a/native/src/components/ExportEventButton.tsx
+++ b/native/src/components/ExportEventButton.tsx
@@ -1,8 +1,8 @@
-import { DateTime } from 'luxon'
 import React, { ReactElement, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Linking, Platform } from 'react-native'
 import RNCalendarEvents, { Calendar } from 'react-native-calendar-events'
+import { check, checkMultiple, PERMISSIONS, request, requestMultiple } from 'react-native-permissions'
 import styled from 'styled-components/native'
 
 import { EventModel } from 'api-client'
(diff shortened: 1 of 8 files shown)


f1sh1918 (changes requested):
Only tested on android so far. Couldn't request permissions on Android 12. 
Added some comments/proposals to get it work.

native/src/components/ExportEventButton.tsx
```diff
@@ -27,45 +28,51 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
...
-    if (permission === 'denied' || permission === 'restricted') {
+    const iosPermission = PERMISSIONS.IOS.CALENDARS
+    const androidPermissions = [PERMISSIONS.ANDROID.READ_CALENDAR, PERMISSIONS.ANDROID.WRITE_CALENDAR]
+    const checkedPermission =
+      Platform.OS === 'ios' ? await check(iosPermission) : await checkMultiple(androidPermissions)
+    const requestPermission = () =>
+      Platform.OS === 'ios' ? request(iosPermission) : requestMultiple(androidPermissions)
+
+    const permission = checkedPermission === 'unavailable' ? await requestPermission() : checkedPermission
+    if (permission === 'denied' || permission === 'limited') {
```
f1sh1918, line 75: this check won't work because on android permission in an object.
You have to access it.
https://github.com/zoontek/react-native-permissions#checkmultiple
 You should also consider the "blocked" variant on android if you explicitly deny it. In this case you could just

native/src/components/ExportEventButton.tsx
```diff
@@ -27,45 +28,51 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
...
-      checkedPermission === 'undetermined' ? await RNCalendarEvents.requestPermissions() : checkedPermission
-    if (permission === 'denied' || permission === 'restricted') {
+    const iosPermission = PERMISSIONS.IOS.CALENDARS
+    const androidPermissions = [PERMISSIONS.ANDROID.READ_CALENDAR, PERMISSIONS.ANDROID.WRITE_CALENDAR]
+    const checkedPermission =
+      Platform.OS === 'ios' ? await check(iosPermission) : await checkMultiple(androidPermissions)
+    const requestPermission = () =>
+      Platform.OS === 'ios' ? request(iosPermission) : requestMultiple(androidPermissions)
+
+    const permission = checkedPermission === 'unavailable' ? await requestPermission() : checkedPermission
```
f1sh1918, line 74:
```suggestion
    const permission = await requestPermission()
```
This check handles the library itself afaik. (Only tested on android)

native/src/components/ExportEventButton.tsx
```diff
@@ -79,7 +86,12 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
     if (editableCalendars.length === 0) {
       showSnackbar({ text: 'noCalendarFound' })
     } else if (editableCalendars.length === 1) {
```
f1sh1918, line 88:
```suggestion
    } else if (editableCalendars[0]) {
```

native/src/components/ExportEventButton.tsx
```diff
@@ -79,7 +86,12 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
...
-      exportEventToCalendar(editableCalendars[0]?.id)
+      try {
+        await exportEventToCalendar(editableCalendars[0]?.id)
```
f1sh1918, line 90:
```suggestion
        await exportEventToCalendar(editableCalendars[0].id)
```

native/src/components/ExportEventButton.tsx
LeandraH: Arg, good catch, thanks!

native/src/components/ExportEventButton.tsx
LeandraH: I don't understand why that is necessary, can you explain it to me?

native/src/components/ExportEventButton.tsx
LeandraH: You're right, very cool!

native/src/components/ExportEventButton.tsx
f1sh1918: because its cleaner to check if there is this index `0` instead doing and optional chaining afterwards

native/src/components/ExportEventButton.tsx
```diff
@@ -27,45 +28,57 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
...
+      | Record<'android.permission.READ_CALENDAR' | 'android.permission.WRITE_CALENDAR', PermissionStatus>,
+    state: PermissionStatus,
+  ): boolean => {
+    if (typeof permissions === 'string') {
+      return permissions === state
+    }
+    return (
+      permissions['android.permission.READ_CALENDAR'] === state &&
+      permissions['android.permission.WRITE_CALENDAR'] === state
+    )
```
f1sh1918, line 74:
```suggestion
    return Object.values(permissions).every(permission => permission === state)
```

native/src/components/ExportEventButton.tsx
steffenkleinle: Learned something new here, first time I see that. Quite annoying though that both checks are needed? Or would only the check `0 in editableCalendars` be enough as well?

native/src/components/ExportEventButton.tsx
```diff
@@ -27,45 +28,57 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
...
+
   const checkCalendarsAndExportEvent = async (): Promise<void> => {
-    const checkedPermission = await RNCalendarEvents.checkPermissions()
-    const permission =
-      checkedPermission === 'undetermined' ? await RNCalendarEvents.requestPermissions() : checkedPermission
-    if (permission === 'denied' || permission === 'restricted') {
+    const iosPermission = PERMISSIONS.IOS.CALENDARS
+    const androidPermissions = [PERMISSIONS.ANDROID.READ_CALENDAR, PERMISSIONS.ANDROID.WRITE_CALENDAR]
+    const permission = Platform.OS === 'ios' ? await request(iosPermission) : await requestMultiple(androidPermissions)
+    if (checkPermissionState(permission, 'limited') || checkPermissionState(permission, 'blocked')) {
```
steffenkleinle, line 78: Optional: If you use an array for iOS as well you can simplify it and the delete the function above:
```suggestion
    const iosPermission = [PERMISSIONS.IOS.CALENDARS]
    const androidPermissions = [PERMISSIONS.ANDROID.READ_CALENDAR, PERMISSIONS.ANDROID.WRITE_CALENDAR]
    const permissions = await requestMultiple(Platform.OS === 'ios' ? iosPermission : androidPermissions)
    const permissionsDenied = Object.values(permissions).some(permission =>
      ['limited', 'blocked'].includes(permission),
    )
       
    if (permissionsDenied) {
```

native/src/components/ExportEventButton.tsx
LeandraH: Yup, I didn't know that either. I don't think only checking `0 in editableCalendars` is enough either because if there is more than one editable calendar, it would also have a 0 :(

steffenkleinle (approved):
Tested on Android, nice find!

native/src/components/ExportEventButton.tsx
LeandraH: Great idea, thanks!

Issue #2477, opened 2023-09-26
Wrong app name in Feedback


Docmale:
### Describe the Bug

Wrong app name on page Feedback "... da es uns hilft **Integreat** zu verbessern." in Malte app and Hallo Aschaffenburg app 

### Expected Behavior

The text should sound somewhat like "... da es uns hilft Malte zu verbessern." "... da es uns hilft Hallo Aschaffenburg zu verbessern."

sarahsporck:
I overlooked that this is a `good first issue` ^^. Maybe sth for @relentless95?

relentless95:
yes, please give me some more.

sarahsporck:
Then feel free to work on this. If you have any questions, send me a message 👍🏼

PR #2559, opened 2023-11-10
2477 wrong app name in feedback

relentless95:
### Short description

Wrong app name on page Feedback "...da es uns hilft Integreat zu verbessern" in Malte app and Hallo Aschaffenburg app

The App name changes dynamically in  the Feedback page. 
Depending on the app it reads "...da es uns hilft Integreat zu verbessern"  or "...da es uns hilft malte zu verbessern" .
Added a placeholder in translations.json  and created a constant which is used in Feedback.tsx to change the app name dynamically.

### Side effects
None

#2477 

### Resolved issues

Fixes: #

---

### DIFF
--- a/native/src/components/Feedback.tsx
+++ b/native/src/components/Feedback.tsx
@@ -1,3 +1,4 @@
+import name from 'build-config-name'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
 import { ActivityIndicator, ScrollView, Text, TextInput } from 'react-native'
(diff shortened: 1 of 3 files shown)


LeandraH (commented):
Very cool, thanks for the PR! However, I don't like the idea of making (or renaming) a new constant in the build config just for a single use case. I think you can use it directly in the code without that step.
It's also using the wrong name, the name 'malte' (lowercase) is one that we might use in the code but not for communication with users. For that, we would use 'Malte' (uppercase) which is the `appName`.

steffenkleinle (approved):

LeandraH (approved):
Nice, thank you!

ztefanie (changes requested):
This will break the web app, as you need to pass the {{appName}} there also, where the commentDescription translation is used.

relentless95:
> This will break the web app, as you need to pass the {{appName}} there also, where the commentDescription translation is used.

I tried doing the same thing in the web app and It seems to be working. I will push and wait for the reviews

ztefanie (approved):
Thank you, looks good :+1:

LeandraH:
Good catch, and thanks for fixing it :)

PR #2560, opened 2023-11-13
Update translations

ztefanie:
The app constants have been translated. Unfortunately our portugues was all spanisch, as the translators noticed.Therefore we asked the translators, who did the translations before to fix it. But this leads to a pt.ods which was out of sync with our current app constants, so I needed to adjust he format to the current format of the .ods document. I tried to be extra carefull and hope I did not mess up anything. So currently all translations are up to date, but portugues which will be finished by the new translators soon, now portugues is incomplete and at the state of the last translation we did.


LeandraH (approved):

steffenkleinle (approved):
Checked a few translations randomly, seems to be correct! Thanks :)

Issue #2504, opened 2023-10-06
Clicking on a cluster on the map should zoom in on it


LeandraH:
**Is your feature request related to a problem? Please describe.**
When I'm on the map and click on a cluster, I expect to zoom in on that. As an example, if I go on http://localhost:9000/testumgebung/de/locations/ in a laptop browser, I see a cluster of three locations in Berlin. When I click on that "3", nothing happens. It should instead center that spot and zoom in. Same goes for native.

We currently have that behavior when the user double clicks on a cluster but other apps usually zoom in from a single click / tap, if it is on a cluster.

**Describe the solution you'd like**
I want the map to center on the cluster and zoom in (at least a little, doesn't have to be enough to open up the cluster).

**Describe alternatives you've considered**
We could leave it like it is but it's not what other maps do, so users probably expect a zoom.

nikhilgarg26:
hey @LeandraH can i work on this issue, could please assign this to me ?

LeandraH:
That's awesome, thank you!

LeandraH:
Hey @nikhilgarg26, how's it going with this? Do you have any questions that we can maybe help you with?

LeandraH:
Hey @nikhilgarg26, are you still working on this? If not, I'd remove you as the assignee so that someone else can have a go but thanks for having a go at it!

PR #2563, opened 2023-11-13
2504: Zoom cluster click

f1sh1918:
### Short description

As a user i want to zoom in if i click on a map cluster.

### Proposed changes

- add function that zooms on click of a cluster

### Resolved issues

Fixes: #2504

### Note ###
Just realized that we have different animations on web & native. On native it may make sense to decrease the animation duration

Tested on ios, chrome & firefox

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -1,5 +1,5 @@
 import MapLibreGL, { CameraSettings } from '@maplibre/maplibre-react-native'
-import type { BBox, Feature } from 'geojson'
+import type { BBox, Feature, Geometry, GeoJsonProperties } from 'geojson'
 import { Position } from 'geojson'
 import React, { ReactElement, useCallback, useLayoutEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
@@ -14,6 +14,8 @@ import {
   MapFeature,
   MapFeatureCollection,
   animationDuration,
+  clusterLayerId,
+  featureLayerId,
 } from 'api-client'
 
 import { LocationFixedIcon, LocationNotFixedIcon, LocationOffIcon } from '../assets'
@@ -64,8 +66,6 @@ type MapViewProps = {
   Overlay?: ReactElement
 }
 
-const featureLayerId = 'point'
-
 // Has to be set even if we use map libre
 MapLibreGL.setAccessToken(null)
 const MapView = ({
@@ -100,11 +100,11 @@ const MapView = ({
     bounds: coordinates ? undefined : bounds,
   }
 
-  const moveTo = useCallback((location: Position, bottomSheetHeight: number) => {
+  const moveTo = useCallback((location: Position, bottomSheetHeight: number, zoomLevel = normalDetailZoom) => {
     if (cameraRef.current) {
       cameraRef.current.setCamera({
         centerCoordinate: location,
-        zoomLevel: normalDetailZoom,
+        zoomLevel,
         animationDuration,
         padding: { paddingBottom: bottomSheetHeight },
       })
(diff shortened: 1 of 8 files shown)


native/src/components/MapView.tsx
```diff
@@ -130,6 +130,20 @@ const MapView = ({
...
+      return
+    }
+    const clusterCollection = await mapRef.current.queryRenderedFeaturesAtPoint(
+      [pressedLocation.properties.screenPointX, pressedLocation.properties.screenPointY],
+      undefined,
+      [clusterLayerId],
+    )
+    if (clusterCollection && 0 in clusterCollection.features) {
+      const feature = clusterCollection.features[0] as MapFeature
+      moveTo(feature.geometry.coordinates, bottomSheetHeight, (await mapRef.current.getZoom()) + 2)
```
steffenkleinle, line 144: Not sure if I understand this, we are always just zooming in `2` steps more than the current zoom level? We don't have a specific zoom level we always want as default for clicking on a cluster?

Also, why do we need to query features just to zoom in to the pressed location we already moved to? And why are we calling moveTo again instead of just setting the zoom level (https://github.com/maplibre/maplibre-react-native/blob/main/docs/Camera.md#zoomtozoomlevel-animationduration)?

native/src/components/MapView.tsx
```diff
@@ -144,6 +158,8 @@ const MapView = ({
     const feature = featureCollection?.features.find((it): it is MapFeature => it.geometry.type === 'Point')
     selectFeature(feature ?? null)
     setSheetSnapPointIndex(1)
+
+    zoomOnClusterPress(pressedLocation)
```
steffenkleinle, line 162: You could just pass the first element from `featureCollection` (and even `await mapRef.current.getZoom`) to avoid the checks and the duplicate logic in your new `zoomOnClusterPress` method. Also, isn't this also called for pressing on points, not only clusters?

web/src/components/MapView.tsx
```diff
@@ -110,6 +110,21 @@ const MapView = forwardRef(
...
 
+    const zoomOnClusterPress = useCallback(
+      (event: MapLayerMouseEvent) => {
+        if (mapRef) {
+          const clusterFeatures = mapRef.queryRenderedFeatures(event.point)
+          if (0 in clusterFeatures && clusterFeatures[0].properties.cluster) {
+            mapRef.flyTo({
+              center: event.lngLat,
+              zoom: mapRef.getZoom() + 2,
+            })
```
steffenkleinle, line 121: Same question as above, why don't we just use `setZoom`? https://docs.mapbox.com/mapbox-gl-js/api/map/#map#setzoom

native/src/components/MapView.tsx
f1sh1918: well because you don't want to specify a zoom level but just zoom a little bit closer on click, since you start on a very low zoom level its a very ugly user experience to zoom directly to detail click. And you want to center the point you are clicking thats why i use `flyTo`

web/src/components/MapView.tsx
f1sh1918: answered above

native/src/components/MapView.tsx
f1sh1918: no because you want to center on the cluster and not on the feature in the feature collection thats why i'm using a different layerId (`clusterLayerId`) therefore you get different items
you can just log the `featureCollection` and the `clusterCollection`

```js
Feature undefined
 LOG  ClusterFeature {"geometry": {"coordinates": [10.859384536743164, 48.51899218172588], "type": "Point"}, "id": 138, "properties": {"cluster": true, "cluster_id": 138, "point_count": 2, "point_count_abbreviated": "2"}, "type": "Feature"}

```

LeandraH (approved):
Very cool, thank you! Tested on Firefox and emulated iOS

native/src/components/MapView.tsx
steffenkleinle: Oh sorry, I confused multipoints with clusters... thanks for the explanation :)

native/src/components/MapView.tsx
steffenkleinle: Thanks, missed that.

web/src/components/MapView.tsx
```diff
@@ -110,6 +110,21 @@ const MapView = forwardRef(
...
     }, [])
 
+    const zoomOnClusterPress = useCallback(
+      (event: MapLayerMouseEvent) => {
+        if (mapRef) {
+          const clusterFeatures = mapRef.queryRenderedFeatures(event.point)
+          if (0 in clusterFeatures && clusterFeatures[0].properties.cluster) {
+            mapRef.flyTo({
+              center: event.lngLat,
+              zoom: mapRef.getZoom() + 2,
```
steffenkleinle, line 120: Optional: Should we perhaps extract the `2` to a common constant in the api-client with a speaking name?

steffenkleinle (commented):
Nice work, totally confused clusters with multipois :eyes:. One thing: If the clusters are clickable, I think the mouse cursor should also indicate that on web. Atm it doesn't change like for pois or multipois.

steffenkleinle (approved):
Tested on android and web.

steffenkleinle (approved):

Issue #2547, opened 2023-11-08
Can not send feedback after opening the keyboard 


clarabracklo:
I am using the native app on an iPhone (15.6.1). When opening the feedback funktion at the top right and start typing feedback I can not close the keyboard anymore and therefore can not send any feedback. 

~~I also received the feedback from Munich that no feedback has been send since changing the functionality. Might be connected could also be because the functionality is not visible on every page footer anymore (harder to find).~~ Handled in #2548

ztefanie:
@LeandraH have you tested on Android or why have you added the android label?

LeandraH:
Nope, sorry, must have mistyped or something

f1sh1918:
@clarabracklo 
you should be able to close the keyboard by clicking on the screen above the keyboard?
Since this is a Textarea and not a Textfield you are able to create multiple lines by clicking "Enter" Button.
You can either use the "Enter" Button for new line or leaving the TextField
I think the main issue is not closing the keyboard but to find the functionality in the overflow menu

steffenkleinle:
> @clarabracklo you should be able to close the keyboard by clicking on the screen above the keyboard? Since this is a Textarea and not a Textfield you are able to create multiple lines by clicking "Enter" Button. You can either use the "Enter" Button for new line or leaving the TextField

 Not easily being able to close the feedback was also already reported in the A11y report, so perhaps we should reassign `enter` to close the feedback? Does being able to add linebreaks to the feedback comment add any benefits?

>I think the main issue is not closing the keyboard but to find the functionality in the overflow menu

I think these are just two different issues and therefore should have two different tickets.

f1sh1918:
yes you are right i also thought about that. Gonna test how the `TextArea` behaves if you don't break by enter

f1sh1918:
okay i think as Steffen described the best option is to switch the "Enter" Button by "Done" button and close the keyboard. The user can not break lines manually but it will happen automatically.
i will also add a keyboard aware scroll view so the text remains visible during typing

clarabracklo:
Thank you very much! This feedback has come up during our community workshops and if there are any changes implemented I think it would be great to communicate that in our facebook group so that our users feel heard . Do you think you could possibly ping Leonie when you are done @f1sh1918 ?

PR #2564, opened 2023-11-14
2547: Send feedback open keyboard

f1sh1918:
### Short description

Currently the user is not able to close the keyboard by virtual keyboard because the "Enter" button makes new lines. So the user may be confused how to close the keyboard.

### Proposed changes

- replace enter button by "done" button
- close keyboard on done instead of new line
- add keyboard aware view to be able to read the written text when keyboard is open

### Resolved issues

Fixes: #2547 

---

### DIFF
--- a/native/src/components/Feedback.tsx
+++ b/native/src/components/Feedback.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
 import { ActivityIndicator, ScrollView, Text, TextInput } from 'react-native'
+import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
 import { useTheme } from 'styled-components'
 import styled from 'styled-components/native'
 
@@ -116,7 +117,7 @@ const Feedback = ({
   const renderBox = (): React.ReactNode => {
     if (['idle', 'failed'].includes(sendingStatus)) {
       return (
-        <>
+        <KeyboardAwareScrollView>
           {isSearchFeedback ? (
             <>
               <HeadlineContainer>
@@ -133,6 +134,7 @@ const Feedback = ({
               <FeedbackButtons isPositiveFeedback={isPositiveFeedback} setIsPositiveFeedback={setIsPositiveFeedback} />
             </>
           )}
+
           <HeadlineContainer>
             <Description>{t('commentHeadline')}</Description>
             <Text>({t('optionalInfo')})</Text>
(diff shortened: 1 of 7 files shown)


LeandraH (approved):
Very cool, thanks! I really like the input scrolling into focus when the keyboard is opened. Tested on real iOS

native/src/components/Feedback.tsx
```diff
@@ -133,26 +134,39 @@ const Feedback = ({
...
           <DescriptionContainer>
             <Text>{t('commentDescription', { appName: buildConfig().appName })}</Text>
           </DescriptionContainer>
-          <CommentInput onChangeText={onCommentChanged} value={comment} multiline numberOfLines={3} />
+          <CommentInput
+            onChangeText={onCommentChanged}
+            value={comment}
+            multiline
+            numberOfLines={3}
+            blurOnSubmit
```
LeandraH, line 150: Why did you decide to only blur this one on submit?

ztefanie (approved):
Looks good :+1: 

Tested on Android, works as expected

Two small thoughts:
- react-native-keyboard-aware-scroll-view was not updated for the last two years
- you can move the mocking of the keyboardAwareView to the .mocks folder, so it only needs to be implemented once.

steffenkleinle (approved):

f1sh1918:
> Looks good :+1: 
> 
> Tested on Android, works as expected
> 
> Two small thoughts:
> - react-native-keyboard-aware-scroll-view was not updated for the last two years
> - you can move the mocking of the keyboardAwareView to the .mocks folder, so it only needs to be implemented once.

I didn't find any alternative for this library. Also it's not a crucial feature, so I think it's okay.
Gonna move it to mocks, thx for the hint

native/src/components/Feedback.tsx
f1sh1918: because i want to leave the textfield? what would you suggest?

LeandraH: I'm fine with that, I just figure that it should then blur the other inputs as well.

f1sh1918: Basically because its default true for single line inputs

_If true, the text field will blur when submitted. The default value is true for single-line fields and false for multiline fields. Note that for multiline fields, setting blurOnSubmit to true means that pressing return will blur the field and trigger the onSubmitEditing event instead of inserting a newline into the field._

steffenkleinle: Why is blurring the text input connected to leaving the input field?

f1sh1918: because you loose the focus of the field

steffenkleinle: Thanks for explaining, makes sense!

Issue #2524, opened 2023-10-16
Merge Feedback and FeedbackContainer


LeandraH:
**Is your feature request related to a problem? Please describe.**
@steffenkleinle wrote in https://github.com/digitalfabrik/integreat-app/pull/2472: 

"IMO Feedback and FeedbackContainer should be merged (both native and web) because we are just passing a lot of props between the two components without any real separation of concerns from a logical point of view. Common logic between web and native could be moved to the api client/endpoint. What do you think?"

I agree :)

### Test information

Check if sending feedback in different variations and form different routes still works on both native and web.

ztefanie:
I disagree. 
There is a separation of concerns.
Feedback is for rendering and all the visual stuff (all the styled components live there)
and FeedbackContainer is for the logical stuff.
In the tests it becomes really visisble, FeedbackContainer tests are `function xy should have been called...` and so on, whereas in Feedback all the tests are just `button should be disabled`, `correct text should be displayed`.

I think this two files show really good, how it is possible to split up components and how to realize single responsibly principle in frontend components easily. If we change something, we should more often split up components than merge these two into one.

steffenkleinle:
> I disagree. There is a separation of concerns. Feedback is for rendering and all the visual stuff (all the styled components live there) and FeedbackContainer is for the logical stuff. In the tests it becomes really visisble, FeedbackContainer tests are `function xy should have been called...` and so on, whereas in Feedback all the tests are just `button should be disabled`, `correct text should be displayed`.
> 
> I think this two files show really good, how it is possible to split up components and how to realize single responsibly principle in frontend components easily. If we change something, we should more often split up components than merge these two into one.

I agree that just merging the two files is not a good idea because it becomes to big. However, I don't think the current way of splitting is a good one. First of all, things that are duplicated in native and web should be moved to the api-client which already reduces the size of the FeedbackContainer quite a bit. Second, at least in my opinion, UI should go normally in smaller modular components rather than in one big `Feedback` file which has all styled components. Instead, There should be separate smaller components like `Note`, `FeedbackButtons`, `TextInput` and so on. My main problem at the moment is that ALL the UI lives in ONE component (Feedback) and all the props are just passed down do that one component. And if we have those smaller visual components, we can a) reuse them in other places and b) just use them directly in `FeedbackContainer` because there is not that much need for additional styled components anymore.

ztefanie:
[...]

PR #2566, opened 2023-11-16
2524: Refactor feedback native

steffenkleinle:
### Short description

Refactor feedback on native by moving shared logic to api-client and extracting things to components.

### Proposed changes

- Extract shared feedback logic with web to api-client
- Extract code to new separate components (InputSection, Note, FeedbackButtons)
- Adjust styling

### Side effects

Feedback on web is broken, will be done in a separate PR ().

### Resolved issues

Partly fixes: #2524.

---


f1sh1918 (commented):
Tested in ios.
Even if i select a feedback button the note does not disappear.
Not sure if the design changed but the comment field was 3 lines default. Now on ios its only a single line field. Since the textfield lines increase if you type more i'm find with that behavior if ui/ux is fine

native/src/components/InputSection.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <TitleContainer>
+        <Title>{title}</Title>
+        {showOptional && <Text>({t('optional')})</Text>}
+      </TitleContainer>
+      {description ? <Text>{description}</Text> : null}
+      <Input
+        onChangeText={onChange}
+        value={value}
+        multiline={multiline}
+        numberOfLines={multiline ? MULTI_LINE_NUMBER : undefined}
```
f1sh1918, line 71: This works only for android
https://reactnative.dev/docs/textinput#numberoflines-android

You have to set a height to the container as before like `<height in px> * numberOflines`

steffenkleinle: Good to know, thanks!

native/src/components/Note.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { ReactElement } from 'react'
+import styled from 'styled-components/native'
+
+import { NoteIcon } from '../assets'
+import Icon from './base/Icon'
+
+const NoteBox = styled.View<{ visible: boolean }>`
+  background-color: ${props => props.theme.colors.themeColor};
```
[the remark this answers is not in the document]
steffenkleinle: I will create a separate issue since I don't know which color to use and this was introduced before: #2577.

f1sh1918 (approved):
Looks good on ios now

native/src/components/Feedback.tsx
```diff
@@ -109,87 +53,56 @@ const Feedback = ({
...
-            <>
-              <Caption title={t('headline')} />
-              <DescriptionContainer>
-                <Text>{t('description')}</Text>
-              </DescriptionContainer>
-              <FeedbackButtons isPositiveFeedback={isPositiveFeedback} setIsPositiveFeedback={setIsPositiveFeedback} />
-            </>
-          )}
+  if (sendingStatus === 'sending') {
+    return <ActivityIndicator size='large' color='#0000ff' />
```
sarahsporck, line 62: Why did you choose this color? Why not use the Loading component?

native/src/components/Note.tsx
```diff
@@ -0,0 +1,39 @@
...
+  margin-right: 12px;
+`
+
+type NoteProps = {
+  text: string
+  visible: boolean
+}
+
+const Note = ({ text, visible }: NoteProps): ReactElement => (
+  <NoteBox visible={visible} aria-hidden={!visible}>
```
sarahsporck, line 33: I'd prefer it if the button moves up when the note disappears. I also checked back with @hauf-toni regarding this. Why did you prefer hiding it?

api-client/src/endpoints/createFeedbackEndpoint.ts
```diff
@@ -19,56 +35,84 @@ export enum FeedbackType {
...
 
-      if (params.comment !== null) {
-        formData.append('comment', params.comment)
+      if (isPositiveRating !== null) {
+        formData.append('rating', isPositiveRating ? POSITIVE_RATING : NEGATIVE_RATING)
       }
 
-      if (params.query !== undefined) {
-        formData.append('query', params.query)
+      const queryWithSearchTerm = query === searchTerm ? query : `${searchTerm} (actual query: ${query})`
```
sarahsporck, line 103: Don't like how this looks in the backend. 
Maybe just the query in brackets without the "actual query"? Because the CMS is actually in german.

native/src/components/Note.tsx
steffenkleinle: That was the way it was implemented bevor. It was at least partly on my initiative though since I don't like jumping layouts. If I want to click the button and then it moves up that is annoying to me. However, if UI/UX says thats how it should be done we can easily do that.

native/src/components/Feedback.tsx
steffenkleinle: This is the previous implementation I just didn't change.

api-client/src/endpoints/createFeedbackEndpoint.ts
steffenkleinle: This is also previous implementation 🙈

steffenkleinle:
@sarahsporck do you want me to implement the changes you would like in this PR? I think the PRs are already big enough and in general I thought we wanted to try to not mix big refactorings and behavior changes. Therefore I think I would prefer if you could open a new ticket for these things.

sarahsporck:
Jep, I'd also prefer a soearate pr :)

steffenkleinle:
> Jep, I'd also prefer a soearate pr :)

Created https://github.com/digitalfabrik/integreat-app/issues/2597.

sarahsporck (approved):

PR #2567, opened 2023-11-16
2524: Refactor feedback web

steffenkleinle:
### Short description

Refactor feedback on web by reusing shared logic from api-client and extracting things to components.

I chose not to merge Feedback and FeedbackContainer as both are still ~ 100 lines.

### Proposed changes

- Extract shared feedback logic with web to api-client
- Extract code to new separate components (InputSection, Note, FeedbackButtons)
- Adjust styling
- Align code with native

### Side effects

None.

### Resolved issues

Fixes: #2524.

---

### DIFF
--- a/web/src/components/InputSection.tsx
+++ b/web/src/components/InputSection.tsx
@@ -1,0 +1,73 @@
+import React, { HTMLInputTypeAttribute, ReactElement } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+
+import TextInput from './TextInput'
+
+const MULTI_LINE_NUMBER = 7
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+`
+
+const TitleContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`
+
+export const Title = styled.label`
+  font-weight: bold;
+`
+
+const StyledTextArea = styled.textarea`
+  resize: none;
+  min-height: 60px;
+`
+
+type InputSectionProps = {
+  id: string
+  title: string
+  description?: string
+  value: string
+  onChange: (input: string) => void
+  type?: HTMLInputTypeAttribute
+  multiline?: boolean
+  showOptional?: boolean
+}
+
+const InputSection = ({
+  id,
+  title,
+  description,
+  value,
+  onChange,
+  type,
+  multiline = false,
+  showOptional = false,
+}: InputSectionProps): ReactElement => {
+  const { t } = useTranslation('common')
+  return (
+    <Container>
+      <TitleContainer>
+        <Title htmlFor={id}>{title}</Title>
+        {showOptional && <div>({t('optional')})</div>}
+      </TitleContainer>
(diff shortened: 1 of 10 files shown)


f1sh1918 (approved):
Looks fine to me. Some small suggestions

web/src/components/Note.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactElement } from 'react'
+import styled from 'styled-components'
+
+import { NoteIcon } from '../assets'
+import Icon from './base/Icon'
+
+const NoteContainer = styled.div<{ visible: boolean }>`
+  display: flex;
+  background-color: ${props => props.theme.colors.themeColor};
```
f1sh1918, line 9: please use a separate `warningColor` instead of theme

web/src/components/InputSection.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { HTMLInputTypeAttribute, ReactElement } from 'react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+
+import TextInput from './TextInput'
+
+const MULTI_LINE_NUMBER = 7
```
f1sh1918, line 7: i would use an optional parameter with a default value instead of setting it here fixed to 7 for every multiline field

web/src/components/Note.tsx
steffenkleinle: Created #2577.

sarahsporck (approved):
Nice! I like the separation of the components :)
When sending search feedback there is still the message that no results are found. Which should be removed after sending feedback (at least in my opinion). But feel free to merge without changing this. :)

steffenkleinle:
> Nice! I like the separation of the components :) When sending search feedback there is still the message that no results are found. Which should be removed after sending feedback (at least in my opinion). But feel free to merge without changing this. :)

I would create a new (good first issue) for this :) Do you think we should hide the whole failure (smiley + message) after a feedback has been sent?

web/src/components/Note.tsx
```diff
@@ -0,0 +1,32 @@
...
+  font-size: ${props => props.theme.fonts.decorativeFontSizeSmall};
+`
+
+type NoteProps = {
+  text: string
+  visible: boolean
+}
+
+const Note = ({ text, visible }: NoteProps): ReactElement => (
+  <NoteContainer visible={visible}>
```
sarahsporck, line 26: Same question as in native: I'd prefer it if the button moves up when the note disappears. I also checked back with @hauf-toni regarding this. Why did you prefer hiding it?

sarahsporck (changes requested):

web/src/components/Note.tsx
steffenkleinle: That was the way it was implemented bevor. It was at least partly on my initiative though since I don't like jumping layouts. If I want to click the button and then it moves up that is annoying to me. However, if UI/UX says thats how it should be done we can easily do that.

steffenkleinle: Created https://github.com/digitalfabrik/integreat-app/issues/2597.

sarahsporck (approved):

PR #2570, opened 2023-11-20
Add portugues translations

ztefanie:
Added translations for portugues.

Fixes: https://github.com/digitalfabrik/integreat-app/issues/2445


steffenkleinle (approved):
Tested randomly, looks good

ztefanie:
> Looks good to me. Can i somehow enable portugues in the cms testumgebung to test to ui? couldn't find an option

This just adds the translations, we need to enable it, by adding it to `config.ts`

ztefanie:
> Looks good to me. Can i somehow enable portugues in the cms testumgebung to test to ui? couldn't find an option

as it pt is already enabled, you should only need to change your system language to portugues

Issue #2575, opened 2023-11-20
Error Open Feedback in BottomActionSheet


f1sh1918:
### Describe the Bug

When the user opens the feedback modal inside the BottomActionSheet it results in a FocusTrap Error.

### Steps to Reproduce

1. [Go to '...'](http://localhost:9000/testumgebung/de/locations)
2. Switch to mobile viewport
3. Open Feedback Modal
### Expected Behavior

The Feedback Modal will open

### Actual Behavior

Crashes with focus trap error

<img width="376" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/37902063/416c2d80-6e5c-4475-8e21-4ccca1c3d02b">

PR #2576, opened 2023-11-20
2575: Focus trap bottom action sheet

f1sh1918:
### Short description

Focus trap lib since `10.x` can not detect nodes from a portal (bottom action sheet) and will throw an error (opening feedback modal)

### Proposed changes

- added a legacy displayCheck for finding the node in the portal

### Resolved issues

Fixes: #2575 

---

### DIFF
--- a/web/src/components/Modal.tsx
+++ b/web/src/components/Modal.tsx
@@ -62,7 +62,7 @@ const Modal = ({ title, closeModal, children, direction, wrapInPortal = false }:
 
     return () => layoutElement?.setAttribute('aria-hidden', 'false')
   }, [])
-  // display check option is needed for portals - https://github.com/focus-trap/tabbable/blob/master/CHANGELOG.md#532
+  // display check option is needed for portals - https://github.com/focus-trap/tabbable/blob/master/CHANGELOG.md#600
   const Modal = (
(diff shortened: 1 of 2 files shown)


web/src/components/Modal.tsx
```diff
@@ -62,9 +62,9 @@ const Modal = ({ title, closeModal, children, direction, wrapInPortal = false }:
 
     return () => layoutElement?.setAttribute('aria-hidden', 'false')
   }, [])
-
+  // display check option is needed for portals - https://github.com/focus-trap/tabbable/blob/master/CHANGELOG.md#532
```
steffenkleinle, line 65: I think you linked the wrong release, shouldnt it be https://github.com/focus-trap/tabbable/blob/master/CHANGELOG.md#600?

It says
> restore old (incorrect) behavior

Should we investigate a better fix?

steffenkleinle (approved):
Tested on firefox

web/src/components/Modal.tsx
f1sh1918: Gonna adjust the link. Yes I can create an New issue for that. I think we should use this fix as a workaround

ztefanie (approved):
Looks good.

Issue #2574, opened 2023-11-20
Create Button base component


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user I want consistently looking UI elements. As a developer, I don't want to change attributes in every usage of a button.

**Describe the solution you'd like**
Introduce a new base component `Button` that is used instead of all `<button>` tags or `role='button'` attributes. Also our other base components `TextButton`, `ToggleButton` and `ChipButton` should use that base component.
Use the `<button>` tag in favor of `role='button'` attribute (see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role).

Also check that there are no more appearances of deprecated `onKeyPressed` in our code.

**Describe alternatives you've considered**
None.

**Additional context**
None.

Issue #2572, opened 2023-11-20
Tabbing through poi details opens and closes the collapsibles


steffenkleinle:
### Describe the Bug

If I am tabbing through the poi details, the collapsibles collapse and open seemingly randomly.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/testumgebung/de/locations/ballonfabrik
2. Tab until reaching "Kontaktdetails"
3. See collapsing
4. Tab some more forward/backwards to see more collapsing/opening

### Expected Behavior

The collapsibles should not open/close on tab. Instead, they should only open/close if another key is pressed.
Also, the whole button/collapsible header should have an accessibility label, not just the icon.

### Actual Behavior

The collapsibles open/close on tab.

### Additional Information

None.

steffenkleinle:
Already fixed in #2574.

PR #2578, opened 2023-11-21
2574: Introduce base button component

steffenkleinle:
### Short description

Introduce a base button component and replace all usages of `<button>` and `role='button'`.

### Proposed changes

- Introduce a new `Button` component which overrides default styling
- Replace occurrences of  `<button>` and `role='button'` with new component
- Add aria-label in a few places

### Side effects

None.

### Resolved issues

Fixes: #2574.

---


f1sh1918 (approved):
One small thing to adjust. Rest looks fine to me. Nice refactoring

web/src/components/base/Button.tsx
```diff
@@ -0,0 +1,34 @@
...
+  cursor: ${props => (props.disabled ? 'default' : 'pointer')};
+  pointer-events: ${props => (props.disabled ? 'none' : 'default')};
+  padding: 0;
+  border: none;
+  text-align: start;
+`
+
+type ButtonProps = {
+  onClick: () => void
+  children?: ReactNode
```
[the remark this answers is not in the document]
steffenkleinle: Okay to me, the only case where we dont pass children is in the Modal for the Overlay. Of course we could pass a empty div or fragment. What do you think?

translations/translations.json
```diff
@@ -6667,6 +6667,8 @@
       "detailsInformation": "Informationen",
       "detailsNextPoi": "nächster",
       "detailsPreviousPoi": "vorheriger",
+      "previousPoi": "Vorheriger Ort",
```
ztefanie, line 6670: No need to add these new translations you can just use ${`detailsPreviousPoi` `poi`}.

web/src/components/Collapsible.tsx
```diff
@@ -53,10 +53,9 @@ const Collapsible = ({
   return (
     <Container className={className}>
       <CollapsibleHeader
-        role='button'
+        ariaLabel={t(collapsed ? 'showMore' : 'showLess')}
```
ztefanie, line 56: Better use the aria-expanded, which is explicitly for collapiables: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

web/src/components/Collapsible.tsx
```diff
@@ -53,10 +53,9 @@ const Collapsible = ({
...
         onClick={() => setCollapsed(!collapsed)}
-        tabIndex={0}
-        onKeyUp={() => setCollapsed(!collapsed)}>
```
ztefanie, line 59: Why have you decided to disable the key handling?

web/src/components/KebabMenu.tsx
```diff
@@ -95,9 +90,9 @@ const KebabMenu = ({ items, direction, show, setShow, Footer }: KebabMenuProps):
 
   return (
     <ToggleContainer>
-      <ToggleButton onClick={onClick} aria-label={t('sideBarOpenAriaLabel')}>
+      <Button onClick={onClick} ariaLabel={t('sideBarOpenAriaLabel')}>
```
ztefanie, line 93: Would also make sense to use aria-expanded here:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

web/src/components/base/Button.tsx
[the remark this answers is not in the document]
ztefanie: I prefer making it child prop mandatory.

web/src/components/base/Button.tsx
```diff
@@ -0,0 +1,34 @@
...
+  tabIndex?: number
+  className?: string
+}
+
+const Button = ({ onClick, children, ariaLabel, tabIndex, className, disabled = false }: ButtonProps): ReactElement => (
+  <StyledButton
+    type='button'
+    onClick={onClick}
+    disabled={disabled}
+    aria-label={ariaLabel}
```
ztefanie, line 27: Please also add aria role button:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

ztefanie (commented):
Tested on chrome, looks good.

translations/translations.json
steffenkleinle: Not the case for every language, in some languages this (i.e. the order) could differ.

web/src/components/Collapsible.tsx
steffenkleinle: The key handling is still there, just now with the default button behavior that only specific keys (e.g. enter) trigger the onClick. This also fixes e.g. this bug: #2572.

web/src/components/base/Button.tsx
steffenkleinle: Why? This doesn't make sense for a `<button>` tag.
> The button role is for clickable elements that trigger a response when activated by the user. Adding role="button" tells the screen reader the element is a button, but provides no button functionality. Use [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) or [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) with type="button" instead.

web/src/components/Collapsible.tsx
steffenkleinle: Good suggestion. However, I think the `ariaLabel` should still not be removed. In the example both is used together. Will add the `aria-expanded` though.

web/src/components/KebabMenu.tsx
steffenkleinle: Same as above.

web/src/components/Collapsible.tsx
ztefanie: I agree

ztefanie (approved):
Not tested again

Issue #2505, opened 2023-10-06
Filter buttons too close to each other in iPhone 13 Mini


LeandraH:
### Describe the Bug

When I open the locations filter on my iPhone 13 Mini, the tiles are touching. They shouldn't be.

### Steps to Reproduce

1. Open the native app
2. Go to Testumgebung
3. Go to locations
4. Click on filters
5. See the tiles too close 

### Expected Behavior

There should be a reasonable gap between the tiles (maybe 8px?)

### Actual Behavior

There is no gap between them.

### Additional Information

Screenshot:
![Image](https://github.com/digitalfabrik/integreat-app/assets/18513943/a15051f3-c3e4-4ba8-b493-e216b65fd317)

PR #2582, opened 2023-11-21
2505: Filter button too close

f1sh1918:
### Short description

On small device the filter buttons are too close together.

### Proposed changes

- remove padding of view, since there is alright sufficient padding

### Resolved issues

Fixes: #2505

Sorry didn't want to grab this easy issue. I expected it to be more complicated like only rendering less issue per row on small devices by creating a grid or sth. But seems like there was just a unneeded padding

### DIFF
--- a/release-notes/unreleased/2505-filter-button-too-close.yml
+++ b/release-notes/unreleased/2505-filter-button-too-close.yml
@@ -1,0 +1,7 @@
+issue_key: 2505
+show_in_stores: false
+platforms:
+  - ios
+  - android
+en: Remove unneeded padding for filter buttons
+de: Unnötiger Abstand von Poi-Filter-Buttons entfernt
--- a/native/src/components/PoiFiltersModal.tsx
+++ b/native/src/components/PoiFiltersModal.tsx
@@ -17,7 +17,6 @@ const Container = styled.View`
   flex: 1;
   flex-direction: column;
   align-items: flex-start;
-  padding: 0 16px;
 `
 
 const SubTitle = styled(Text)`


sarahsporck (commented):
I don't really like the look when there are only a few filter categories. Maybe we can change this from space-between to `left`/content-dir-dependent?
Bildschirmfoto 2023-11-22 um 10 30 03

f1sh1918:
@sarahsporck  and @steffenkleinle 
what do you think about centering and make a gap.
If we do `flex-start` we have the issue that maybe there will space right depending on device width

image

release-notes/unreleased/2505-filter-button-too-close.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2505
+show_in_stores: false
+platforms:
+  - ios
+  - android
+en: Remove unneeded padding for filter buttons
+de: Unnötiger Abstand von Poi-Filter-Buttons entfernt
```
steffenkleinle, line 7: I think this PR is more about adding padding between the filter buttons instead of removing it, isn't it?
```suggestion
en: Add padding between filter buttons.
```

steffenkleinle (approved):

steffenkleinle (changes requested):
I think the gap between multiple tile rows is now too big (margin-bottom of 24px and gap of 16px)

:

f1sh1918:
> I think the gap between multiple tile rows is now too big (margin-bottom of 24px and gap of 16px):
> 
> 

This the same as we have in Web @steffenkleinle

f1sh1918:
> I think the gap between multiple tile rows is now too big (margin-bottom of 24px and gap of 16px):
> 
> 

Hm we have the same spacings on web and it depends on the device width if we have two or three items a row.
So what do you suggest?
We can define how many items a row and make tiles flexible or keep it as it is

release-notes/unreleased/2505-filter-button-too-close.yml
f1sh1918: Not really. I removed the padding around to be able to have a gap between

steffenkleinle:
> > I think the gap between multiple tile rows is now too big (margin-bottom of 24px and gap of 16px):
> > 
> 
> Hm we have the same spacings on web and it depends on the device width if we have two or three items a row. So what do you suggest? @steffenkleinle We can define how many items a row and make tiles flexible or keep it as it is. I mean I can reduce the gap but I just followed design and it will even come to a limit with small devices

I don't mean the horizontal but the vertical gap. We have a gap of 40px there which I would personally say is too much. If the design has a vertical gap of 40px fine for me as well.

release-notes/unreleased/2505-filter-button-too-close.yml
steffenkleinle: Like fine for me, I just meant that the issue was specifically about the buttons not having a gap between and therefore being right next to each other which was the problem for the users imo. If there is now more or less space around is not the relevant part, is it?

f1sh1918:
> > > I think the gap between multiple tile rows is now too big (margin-bottom of 24px and gap of 16px)
> 
> > Hm we have the same spacings on web and it depends on the device width if we have two or three items a row. So what do you suggest? @steffenkleinle We can define how many items a row and make tiles flexible or keep it as it is. I mean I can reduce the gap but I just followed design and it will even come to a limit with small devices
> 
> I don't mean the horizontal but the vertical gap. We have a gap of 40px (24px margin-bottom + 16px gap)/there which I would personally say is too much. If the design has a vertical gap of 40px fine for me as well. The amount of items in a row and the horizontal gap is totally fine imo. Basically I think I would either remove the vertical gap or margin-bottom.

okay now i got it. thx for the explanation. I updated the margins.

release-notes/unreleased/2505-filter-button-too-close.yml
ztefanie: Margin vs padding :D

ztefanie (approved):
Looks good on android emulator.

steffenkleinle (approved):

PR #2586, opened 2023-11-22
2359: Show not found error for pois

steffenkleinle:
### Short description

Show a not found error if n poi can be found for the given slug.

### Proposed changes

- Fix the order to actually show an error if no poi is found (native)
- Introduce a new `PoiSharedChildren` component for reusability between mobile and desktop (web)
- Handle not found error in PoiSharedChildren (web)
- Show back button and hide `In deiner Nähe` for not found error (web)

### Side effects

None.

### Resolved issues

Fixes: #2359.

---

### DIFF
--- a/web/src/components/PoiSharedChildren.tsx
+++ b/web/src/components/PoiSharedChildren.tsx
@@ -1,0 +1,59 @@
+import React, { ReactElement } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { GeoJsonPoi, LocationType, PoiModel, sortMapFeatures } from 'api-client'
+import { UiDirectionType } from 'translations'
+
+import Failure from './Failure'
+import List from './List'
+import PoiDetails from './PoiDetails'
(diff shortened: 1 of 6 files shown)


web/src/components/PoiSharedChildren.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { ReactElement } from 'react'
```
steffenkleinle, line 1: Not really happy with the naming `PoiSharedChildren`, other suggestions?

LeandraH (approved):
It's a little odd to have two buttons that lead back to the overview in desktop but I think it's okay. 

image

web/src/components/PoiSharedChildren.tsx
sarahsporck: Not really 🤔

sarahsporck (approved):
Tested on ios & android simulator, as well as firefox.

PR #2588, opened 2023-11-22
2026: Migrate to yarn 4

sarahsporck:
2026: update yarn

### Short description

As already discussed multiple times I'd like to update/migrate to the current yarn version. This pr introduces yarn /w node_modules.
If ci failes due to oom I'll have to consider splitting the repo installs. Let's see. This will then be in a separate pr.

### Proposed changes

- update to yarn4
- use global scripts
    - tests are run with `yarn :test` in the corresponding folder
    - lint with `yarn :lint`
    - ...
- update styled-components due to typescript issues

### Resolved issues

Fixes: #2026

---


.circleci/src/jobs/promote_ios.yml
```diff
@@ -13,7 +13,6 @@ steps:
   - checkout
   - skip_job:
       build_config_name: << parameters.build_config_name >>
-  - restore_yarn_tools_cache
   - restore_yarn_cache # Needed for babel-node run in read_build_config.rb. May be refactored and optimized later.
```
steffenkleinle, line 16:
```suggestion
  - restore_yarn_cache
```

api-client/src/testing/mockUseLoadFromEndpoint.ts
```diff
@@ -1,6 +1,6 @@
-import { mocked } from 'jest-mock'
+import { useLoadFromEndpoint } from '..'
 
-import { useLoadFromEndpoint } from 'api-client'
+import { mocked } from 'jest-mock'
```
steffenkleinle, line 3: Shouldn't this be ordered the other way round?

native/package.json
```diff
@@ -88,8 +84,8 @@
     "react-native-webview": "^13.4.0",
     "react-navigation-header-buttons": "^11.0.0",
     "rrule": "2.7.2",
-    "styled-components": "^5.3.11",
-    "translations": "0.0.1",
+    "styled-components": "^6.1.1",
```
steffenkleinle, line 87: Why did you upgrade styled-components?

package.json
```diff
@@ -9,41 +9,27 @@
...
-    "lint:ci": "yarn node --max-old-space-size=1535 $(yarn bin eslint) --format junit -o reports/lint/junit-lint.xml .",
-    "test": "jest",
-    "test:changed": "jest -o",
-    "test:ci": "jest --coverage --ci",
+    "postinstall": "husky install",
+    ":lint": "cd $INIT_CWD && eslint . --cache --cache-location ../.eslintcache .",
+    "lint:ci": "cd $PROJECT_CWD && eslint --format junit -o reports/lint/junit-lint.xml",
+    ":test": "cd $INIT_CWD && yarn run -T jest",
+    ":test:changed": "cd $INIT_CWD && yarn run -T jest -o",
+    ":test:ci": "cd $PROJECT_CWD && yarn jest --coverage --ci",
```
steffenkleinle, line 22: Why are we sometimes using `INIT_CWD` and sometimes `PROJECT_CWD`?

web/package.json
```diff
@@ -19,22 +19,16 @@
     "build:integreat-test-cms": "yarn run webpack-ts --env config_name=integreat-test-cms",
     "build:obdach": "yarn run webpack-ts --env config_name=obdach",
     "check:build": "yarn es-check es2020 ./dist/**/*.js --verbose",
-    "test": "jest",
-    "lint": " yarn stylelint && yarn eslint --cache --cache-location ../.eslintcache .",
-    "stylelint": "stylelint --config stylelint.config.js \"src/**/*.{ts,tsx,css}\"",
```
steffenkleinle, line 24: So running `lint` in the web dir/workspace will not automatically exeute `stylelint` anymore? Did you adjust the CI?

web/src/assets/licenses.json
```diff
@@ -0,0 +1,363 @@
+{
```
steffenkleinle, line 1: I think this should be deleted. The licenses are already found in `web/assets/licenses.json` instead of `web/src/assets/licenses.json`

web/src/components/CityEntry.tsx
```diff
@@ -31,50 +30,47 @@ type CityEntryProps = {
...
-              {index !== aliases.length - 1 && <span>, </span>}
-            </span>
-          ))}
-          {aliases.length > MAX_NUMBER_OF_ALIASES && <span> ... </span>}
-        </div>
-      </CityListItem>
-    )
-  }
+  const normalizedFilter = normalizeString(filterText)
+  const aliases = getMatchedAliases(city, normalizedFilter)
```
steffenkleinle, line 45: Optional: Probably it would be better to use a `useMemo` here instead of a `useCallback` and then call that: 
``` ts
  const aliases = useMemo((): string[] => {
    const normalizedFilter = normalizeString(filterText)
    if (city.aliases && normalizedFilter.length >= 1) {
      return Object.keys(city.aliases).filter(alias => normalizeString(alias).includes(normalizedFilter))
    }
    return []
  }, [city, filterText])

```

steffenkleinle (changes requested):
Tested on firefox and android, looks really nice already! Thanks :yellow_heart:

steffenkleinle:
One more thing: The integreat icon in the header is now super small.
Screenshot_20231124-132345_IntegreatTestCms

.circleci/src/commands/restore_yarn_cache.yml
```diff
@@ -3,16 +3,13 @@ steps:
...
       name: Restore Yarn Package Cache
       keys:
-        - 8-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}-{{ checksum "native/package.json" }}-{{ checksum "web/package.json" }}
-        - 8-yarn-{{ arch }}-
+        - 9-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}-{{ checksum "native/package.json" }}-{{ checksum "web/package.json" }}
+        - 9-yarn-{{ arch }}-
   - run:
       name: Yarn
-      command: yarn --frozen-lockfile --cache-folder ~/.cache/yarn
+      command: yarn --immutable
```
ztefanie, line 10: Have you checked if the caching is working as expected?

ztefanie (commented):
I think you need to adjust the new commands, like `:test:ci` instead of `test:ci`

sarahsporck:
Sadly this seems to be out of scope right now. I don't think I'll be able to finish this. The main problem is still [the oom in the CI](https://github.com/yarnpkg/berry/issues/3996), which can be worked around, but that only feels like a workaround. 
It also irks me that yarn pnp will never have [proper react native support](https://github.com/react-native-community/cli/issues/27) as it seems. 
A feasable alternative I've already checked out would be [pnpm](https://pnpm.io/), which should pretty much work out of the box with most of our tools. You can checkout the benchmarks here: https://pnpm.io/benchmarks . I'll take a look at the effort of moving to another package manager :) and we can discuss this in our next weekly.

PR #2589, opened 2023-11-22
258: Embed videos

f1sh1918:
### Short description

With this feature media via iframe like vimeo videos (and youtube if source is enabled) can be embedded.
I decided to split up the prs

### Proposed changes

- [ ]  implement native with opt-in and source permissions in settings #2602
- [ ] implement web with opt-in and source permissions with cookies and manage it on separate route (external source) this page should be linked on data privacy page

### Side effects

- hopefully none, please test properly all links and features of the webview we are using

### Resolved issues

Fixes: #1213

### DIFF
--- a/build-configs/BuildConfigType.ts
+++ b/build-configs/BuildConfigType.ts
@@ -51,6 +51,7 @@ export type CommonBuildConfigType = {
   hostName: string
   // Hostnames from which resources are automatically downloaded for offline usage.
   allowedHostNames: Array<string>
+  allowedIframeSources: string[]
   // Regex defining which urls to intercept as they are internal ones.
   internalLinksHijackPattern: string
   featureFlags: FeatureFlagsType
--- a/build-configs/BuildConfigType.ts
+++ b/build-configs/BuildConfigType.ts
@@ -51,7 +51,7 @@ export type CommonBuildConfigType = {
   hostName: string
   // Hostnames from which resources are automatically downloaded for offline usage.
   allowedHostNames: Array<string>
-  allowedIframeSources: string[]
+  whiteListedIframeSources: string[]
   // Regex defining which urls to intercept as they are internal ones.
   internalLinksHijackPattern: string
   featureFlags: FeatureFlagsType
--- a/native/src/components/RemoteContent.tsx
+++ b/native/src/components/RemoteContent.tsx
@@ -79,7 +79,6 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
-      // load allowed iframe sources on request
       if (buildConfig().allowedIframeSources.some(source => event.url.indexOf(source) > 0)) {
         return true
       }
--- a/native/src/components/RemoteContent.tsx
+++ b/native/src/components/RemoteContent.tsx
@@ -80,7 +80,7 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
       // load allowed iframe sources on request
-      if (buildConfig().allowedIframeSources.some(el => event.url.indexOf(el) > 0)) {
+      if (buildConfig().allowedIframeSources.some(source => event.url.indexOf(source) > 0)) {
         return true
       }
       if (event.url === new URL(resourceCacheUrl).href) {
--- a/native/src/components/RemoteContent.tsx
+++ b/native/src/components/RemoteContent.tsx
@@ -6,6 +6,7 @@ import { useTheme } from 'styled-components/native'
 
 import { ErrorCode } from 'api-client'
 
+import buildConfig from '../constants/buildConfig'
 import { userAgent } from '../constants/endpoint'
 import { HEIGHT_MESSAGE_TYPE, WARNING_MESSAGE_TYPE } from '../constants/webview'
 import renderHtml from '../utils/renderHtml'
@@ -76,13 +77,19 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
+      // load allowed iframe sources on request
+      if (buildConfig().allowedIframeSources.some(el => event.url.indexOf(el) > 0)) {
+        return true
+      }
       if (event.url === new URL(resourceCacheUrl).href) {
         // Needed on iOS for the initial load
         return true
       }
       // If it takes too long returning false the webview loads the pressed url anyway on android.
       // Therefore only set it to state and execute onLinkPress in useEffect.
-      setPressedUrl(event.url)
+      if (event.navigationType === 'click') {
+        setPressedUrl(event.url)
+      }
       return false
     },
     [resourceCacheUrl],
@@ -106,6 +113,7 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
       dataDetectorTypes='none'
       userAgent={userAgent}
       domStorageEnabled={false}
+      allowsFullscreenVideo
       showsVerticalScrollIndicator={false}
       showsHorizontalScrollIndicator={false}
       scrollEnabled={false} // To disable scrolling in iOS
--- a/native/src/utils/renderHtml.ts
+++ b/native/src/utils/renderHtml.ts
@@ -94,7 +94,7 @@ const renderJS = (cacheDictionary: ParsedCacheDictionaryType, allowedIframeSourc
     const iframes = document.querySelectorAll('iframe')
     const allowedIframeSources = ${JSON.stringify(allowedIframeSources)}
 
-    iframes.forEach((el) => {
+    iframes.forEach((iframe) => {
       if (allowedIframeSources.some(url => el.src.indexOf(url) > 0)) {
         const url = new URL(el.src)
         // Add do not track parameter
--- a/native/src/utils/renderHtml.ts
+++ b/native/src/utils/renderHtml.ts
@@ -5,94 +5,111 @@ import { ERROR_MESSAGE_TYPE, getFontFaceSource, HEIGHT_MESSAGE_TYPE, WARNING_MES
 
 // language=JavaScript
 const renderJS = (cacheDictionary: Record<string, string>) => `
-  function reportError (message, type) {
-    if (!window.ReactNativeWebView) {
-      return window.setTimeout(function() { reportError(message, type) }, 100)
-    }
-
-    window.ReactNativeWebView.postMessage(JSON.stringify({ type, message: message }))
-  }
-
-  // Catching occurring errors
-  (function() {
-    window.onerror = function(msg, url, lineNo, columnNo, error) {
-      // from https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
-      var string = msg.toLowerCase()
-      var substring = "script error"
-      if (string.indexOf(substring) > -1) {
-        reportError('Script Error: See Browser Console for Detail: ' + msg + JSON.stringify(error),
-          '${ERROR_MESSAGE_TYPE}')
-      } else {
-        var message = [
-          'Message: ' + msg,
-          'URL: ' + url,
-          'Line: ' + lineNo,
-          'Column: ' + columnNo,
-          'Error object: ' + JSON.stringify(error)
-        ].join(' - ')
-        reportError(message, '${ERROR_MESSAGE_TYPE}')
-      }
-      return false
-    };
-  })();
-
-  (function() {
-    var hrefs = document.querySelectorAll('[href]')
-    var srcs = document.querySelectorAll('[src]')
-    var cacheDictionary = ${JSON.stringify(cacheDictionary)}
-
-    console.debug('Resources to inject:')
-    console.debug(cacheDictionary)
-
-    for (var i = 0; i < hrefs.length; i++) {
-      var item = hrefs[i]
-      try {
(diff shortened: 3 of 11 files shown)


native/src/utils/renderHtml.ts
```diff
@@ -5,237 +5,259 @@ import { ERROR_MESSAGE_TYPE, getFontFaceSource, HEIGHT_MESSAGE_TYPE, WARNING_MES
...
+        container.setAttribute('style', '');
+      }
 
+      window.addEventListener('load', adjustHeight);
+      window.addEventListener('resize', adjustHeight);
+      var details = document.querySelectorAll("details")
+      details.forEach(detail => detail.addEventListener("toggle", adjustHeight))
+    })();
+
+    // iframe handling - remove not allowed iframe src from DOM and add tracking parameter
```
f1sh1918, line 97: just added this. rest was not correct formatted

native/src/components/RemoteContent.tsx
```diff
@@ -76,13 +77,19 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
+      // load allowed iframe sources on request
+      if (buildConfig().allowedIframeSources.some(el => event.url.indexOf(el) > 0)) {
```
steffenkleinle, line 81:
```suggestion
      if (buildConfig().allowedIframeSources.some(source => event.url.indexOf(source) > 0)) {
```

native/src/components/RemoteContent.tsx
```diff
@@ -76,13 +77,19 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
+      // load allowed iframe sources on request
```
steffenkleinle, line 80: Optional: I think the comment is not really needed as the variable names are pretty imo.
```suggestion
```

native/src/utils/renderHtml.ts
```diff
@@ -5,237 +5,259 @@ import { ERROR_MESSAGE_TYPE, getFontFaceSource, HEIGHT_MESSAGE_TYPE, WARNING_MES
 
 // language=JavaScript
 const renderJS = (cacheDictionary: Record<string, string>) => `
-  function reportError (message, type) {
-    if (!window.ReactNativeWebView) {
-      return window.setTimeout(function() { reportError(message, type) }, 100)
-    }
+    function reportError (message, type) {
```
steffenkleinle, line 8: Please revert the formatting, now there are 4 spaces instead of our default of 2. Imo the previous was correct. I think this new indentation is just coming from the fact that you added brackets around the string/didn't use direct return.

release-notes/unreleased/258-embed-video.yml
```diff
@@ -0,0 +1,8 @@
+issue_key: 258
+show_in_stores: true
+platforms:
+  - ios
+  - android
+  - web
+en: Embedding of vimeo videos is now supported
+de: Einbinden von Vimeo videos wird jetzt unterstützt
```
steffenkleinle, line 8:
```suggestion
en: Embedding of videos is now supported
de: Einbinden von Videos wird jetzt unterstützt
```

web/src/components/RemoteContent.tsx
```diff
@@ -14,7 +14,6 @@ const SandBox = styled.div<{ centered: boolean; smallText: boolean }>`
 
   ${props => (props.centered ? 'text-align: center;' : '')}
   ${props => (props.centered ? 'list-style-position: inside;' : '')}
-
```
steffenkleinle, line 17: Please readd

web/src/components/RemoteContent.tsx
```diff
@@ -145,10 +149,25 @@ const RemoteContent = ({
         node.addEventListener('click', handleClick)
       }
     })
+
+    // iframe handling - remove not allowed iframe src from DOM and add tracking parameter
+    const iframes = sandBoxRef.current.getElementsByTagName('iframe')
+    Array.from(iframes).forEach((node: HTMLIFrameElement) => {
+      if (buildConfig().allowedIframeSources.some(el => node.src.indexOf(el) > 0) && sandBoxRef.current) {
```
steffenkleinle, line 156: You already using `sandBoxRef.current` above without a non-null check, so it should be fine here as well.
```suggestion
      if (buildConfig().allowedIframeSources.some(source => node.src.indexOf(source) > 0)) {
```

web/src/components/RemoteContent.tsx
```diff
@@ -145,10 +149,25 @@ const RemoteContent = ({
...
+        var url = new URL(node.src)
+        url.searchParams.append('dnt', '1')
```
steffenkleinle, line 158: Is this a convention that the do not track parameter is called `dnt` or is this just the case for vimeo? Because otherwise we need to keep that in mind for other video platforms.

web/src/components/RemoteContent.tsx
```diff
@@ -145,10 +149,25 @@ const RemoteContent = ({
         node.addEventListener('click', handleClick)
       }
     })
+
+    // iframe handling - remove not allowed iframe src from DOM and add tracking parameter
+    const iframes = sandBoxRef.current.getElementsByTagName('iframe')
+    Array.from(iframes).forEach((node: HTMLIFrameElement) => {
+      if (buildConfig().allowedIframeSources.some(el => node.src.indexOf(el) > 0) && sandBoxRef.current) {
+        var url = new URL(node.src)
```
steffenkleinle, line 157:
```suggestion
        const url = new URL(node.src)
```

web/src/components/RemoteContent.tsx
```diff
@@ -145,10 +149,25 @@ const RemoteContent = ({
         node.addEventListener('click', handleClick)
       }
     })
+
+    // iframe handling - remove not allowed iframe src from DOM and add tracking parameter
```
steffenkleinle, line 153: Optional:
```suggestion
    // Remove disallowed iframes from DOM and add do not track parameter
```

web/src/components/RemoteContent.tsx
```diff
@@ -145,10 +149,25 @@ const RemoteContent = ({
...
+    const iframes = sandBoxRef.current.getElementsByTagName('iframe')
+    Array.from(iframes).forEach((node: HTMLIFrameElement) => {
+      if (buildConfig().allowedIframeSources.some(el => node.src.indexOf(el) > 0) && sandBoxRef.current) {
+        var url = new URL(node.src)
+        url.searchParams.append('dnt', '1')
+        node.setAttribute('src', url.href)
+      } else if (sandBoxRef.current) {
```
steffenkleinle, line 160:
```suggestion
      } else {
```

native/src/utils/renderHtml.ts
```diff
@@ -5,237 +5,259 @@ import { ERROR_MESSAGE_TYPE, getFontFaceSource, HEIGHT_MESSAGE_TYPE, WARNING_MES
...
+        if (allowedIframeSources.some(url => el.src.indexOf(url) > 0)) {
+          var url = new URL(el.src)
+          url.searchParams.append('dnt', '1')
+          el.setAttribute('src', url.href)
+      } else {
+          el.parentNode.removeChild(el)
+        }
+      })
+    })();
+  `
```
steffenkleinle, line 112: Question (because I am actually not sure whether it would be easily possible or even if it would be possible if it would make sense): Could we directly strip the `<iframe>` tags from the html in the api-client already?

Then we would only need to do this in one place (but probably for all endpoints with html). So probably it would still not make sense.

steffenkleinle (changes requested):
Tested on android, works fine. Good job!

I can't open any links anymore though (neither deeplinks nor internal ones), without the check it still works
 ``` ts
if (event.navigationType === 'click') {
```

Also, we should take care of what @svenseeberg said [here](https://github.com/digitalfabrik/integreat-app/issues/1213#issuecomment-1822845297):
> Just a note: we need an active opt in of users before content from Vimeo or other services can be loaded.

**Comment by f1sh1918** [comment]
> Tested on android, works fine. Good job!
> 
> I can't open any links anymore though (neither deeplinks nor internal ones), without the check it still works
> 
> ```ts
> if (event.navigationType === 'click') {
> ```
> 
> Also, we should take care of what @svenseeberg said [here](https://github.com/digitalfabrik/integreat-app/issues/1213#issuecomment-1822845297):
> 
> > Just a note: we need an active opt in of users before content from Vimeo or other services can be loaded.

Alright gonna check that. The problem here is that iframe src will also be handled which shouldn't since they are no click event and will be opened in external browser which shouldn't happen.

File: native/src/utils/renderHtml.ts (hunk 4 above)
Comment (f1sh1918) [reply native/src/utils/renderHtml.ts:10]: @steffenkleinle can you specify which string you mean? :)

File: web/src/components/RemoteContent.tsx (hunk 11 above)
Comment (f1sh1918) [reply web/src/components/RemoteContent.tsx:160]: had to declare a const for the `sandBofRef.current` to avoid this check

File: native/src/utils/renderHtml.ts (hunk 12 above)
Comment (f1sh1918) [reply native/src/utils/renderHtml.ts]: I mean we could add this `do not track` in the `api-client` but the rest of the `iframe` handling is different on `native` and `web` so even we need a separate handling.
Additionally i think we not just want to strip this but show a proper message that it was blocked to show the maintainer of the page that he/she is not allowed to use this source instead of just stripping it out.

File: web/src/components/RemoteContent.tsx (hunk 8 above)
Comment (f1sh1918) [reply web/src/components/RemoteContent.tsx:158]: you are right this only works for vimeo, gonna add a check

File: native/src/utils/renderHtml.ts [hunk 13]
```diff
@@ -14,16 +16,15 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
   }
 
   // Catching occurring errors
-  (function() {
+  (function  catchErrors() {
```
LeandraH, line 19:
```suggestion
  (function catchErrors() {
```

LeandraH: I think we can remove the comment above now

LeandraH (approved):
Looks great, thanks! Tested in Firefox and Safari

build-configs/BuildConfigType.ts
```diff
@@ -51,7 +51,7 @@ export type CommonBuildConfigType = {
   hostName: string
   // Hostnames from which resources are automatically downloaded for offline usage.
   allowedHostNames: Array<string>
-  allowedIframeSources: string[]
+  whiteListedIframeSources: string[]
```
steffenkleinle, line 54: I think we should avoid this term if possible bc of discrimination, why did you change it from `allowed`?

native/src/utils/renderHtml.ts
```diff
@@ -81,22 +79,41 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
-    var details = document.querySelectorAll("details")
+    const details = document.querySelectorAll("details")
     details.forEach(detail => detail.addEventListener("toggle", adjustHeight))
   })();
+
+  (function handleIframes() {
+    const iframes = document.querySelectorAll('iframe')
+    const allowedIframeSources = ${JSON.stringify(allowedIframeSources)}
+
+    iframes.forEach((el) => {
```
steffenkleinle, line 97: Please use speaking names:
```suggestion
    iframes.forEach((iframe) => {
```

web/src/utils/doNotTrack.ts
```diff
@@ -0,0 +1,8 @@
+// Vimeo supports do not track
+export const addDoNotTrackParameter = (iframe: HTMLIFrameElement): void => {
+  if (iframe.src.indexOf('vimeo') > -1) {
```
steffenkleinle, line 3:
```suggestion
  if (iframe.src.includes('vimeo')) {
```

web/src/utils/doNotTrack.ts
```diff
@@ -0,0 +1,8 @@
+// Vimeo supports do not track
```
steffenkleinle, line 1: I think this is self explanatory, isn't it? :see_no_evil:

steffenkleinle: Does it make sense to move this to the api-client as this is probably the same logic for native?

web/src/components/RemoteContent.tsx
```diff
@@ -139,16 +145,30 @@ const RemoteContent = ({
...
     Array.from(collection).forEach(node => {
       if (HIJACK.test(node.href)) {
         node.addEventListener('click', handleClick)
       }
     })
+
+    // Remove disallowed iframes from DOM and add do not track parameter for vimeo
+    const iframes = currentSandBoxRef.getElementsByTagName('iframe')
+    Array.from(iframes).forEach((iframe: HTMLIFrameElement) => {
+      if (buildConfig().whiteListedIframeSources.some(source => iframe.src.indexOf(source) > 0)) {
```
steffenkleinle, line 159:
```suggestion
      if (buildConfig().whiteListedIframeSources.some(iframe.src.includes)) {
```

steffenkleinle (commented):
There is a few changes for native here, will you wait for native to merge this? Should we review those changes here? I think it would be better to actually split them.

web/src/utils/doNotTrack.ts
f1sh1918: In general this makes sense, but the problem is we can not import function in the string template. Or lets say import functions and pass them parameters. I tried this several time with the `t()` to pass sources to the translations

f1sh1918:
> There is a few changes for native here, will you wait for native to merge this? Should we review those changes here? I think it would be better to actually split them.

Yes your are right sorry for the confusion. I'm gonna create two separate branches and pr and integrate your suggestions

Issue #2593, opened 2023-11-24
Fix external link styling


steffenkleinle:
### Describe the Bug

If external links don't fit in one line, the external link icon is shown smaller, at a weird place and repeated.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/minden-luebbecke/es/trabajo-y-profesión/encontrar-trabajo/acceso-al-mercado-laboral
2. Scroll down to the linked pdf Zugang_SGB_II_und_Arbeitsmarkt.pdf
3. See external link icon (screenshot attached)

### Expected Behavior

The external link icon looks the same as everywhere else.

### Actual Behavior

The external link icon is shown smaller, at a weird place and repeated.

### Additional Information

![image](https://github.com/digitalfabrik/integreat-app/assets/29095487/7fdfacb5-9514-4f2c-850d-b683fd2c525c)

I found the bug while checking #2487, actually at least partially related to it.

Issue #2487, opened 2023-09-30
Fix mobile usability


steffenkleinle:
### Describe the Bug

Google search console notifies us about two urls that are not usable on mobile: https://search.google.com/search-console/mobile-usability?resource_id=https%3A%2F%2Fintegreat.app%2F
Found issues are:
* [Text too small to read](https://search.google.com/search-console/mobile-usability/drilldown?resource_id=https%3A%2F%2Fintegreat.app%2F&item_key=CAw)
* [Clickable elements too close together](https://search.google.com/search-console/mobile-usability/drilldown?resource_id=https%3A%2F%2Fintegreat.app%2F&item_key=CAs)
* [Content wider than screen](https://search.google.com/search-console/mobile-usability/drilldown?resource_id=https%3A%2F%2Fintegreat.app%2F&item_key=CAk)

All three issues affect the following three websites:
* https://integreat.app/lkstarnberg/de/deutsche-sprache/deutschkurse-im-landkreis/deutschkurs-typen-im-landkreis
* https://integreat.app/minden-luebbecke/es/trabajo-y-profesi%C3%B3n/encontrar-trabajo/acceso-al-mercado-laboral
* https://integreat.app/minden-luebbecke/uk/%D0%BD%D0%B0%D0%B2%D1%87%D0%B0%D0%BD%D0%BD%D1%8F-%D1%82%D0%B0-%D0%BF%D1%80%D0%BE%D1%84%D0%B5%D1%81%D1%96%D0%B9%D0%BD%D0%B0-%D0%BF%D1%96%D0%B4%D0%B3%D0%BE%D1%82%D0%BE%D0%B2%D0%BA%D0%B0/%D0%B2%D0%B8%D0%B7%D0%BD%D0%B0%D0%BD%D0%BD%D1%8F-%D1%96%D0%BD%D0%BE%D0%B7%D0%B5%D0%BC%D0%BD%D0%B8%D1%85-%D0%BE%D1%81%D0%B2%D1%96%D1%82%D0%BD%D1%96%D1%85-%D0%BA%D0%B2%D0%B0%D0%BB%D1%96%D1%84%D1%96%D0%BA%D0%B0%D1%86%D1%96%D0%B9

At least content wider than screen was easily reproducable for the first page with a small viewport.

### Additional Information

These are probably content issues. Investigate and tell the service team what should be changed about the pages (or fix if it is an app problem).

If problems are caused by content, contact service team to fix it.

[...]

PR #2594, opened 2023-11-24
2593: Fix external link styling

steffenkleinle:
### Short description

Remove inline-flex to fix external link styling especially for multiline links.

### Proposed changes

- Remove inline-flex on external links (caused content being wider than screen and the issue described in #2593).
- Change margin-left to be direction independent

### Side effects

None.

### Resolved issues

Fixes: #2593 (and partly #2487).

---


LeandraH (approved):
Nice! Tested in Firefox and in emulated iOS.

I think the icon could be marginally higher in a web
image

and marginally lower in native
image

steffenkleinle:
> I think the icon could be marginally higher in a web

Also using firefox I think the bottom lines of the icon and the link underline are same height (even though it might seem differently because the lines are thicker.

image

> and marginally lower in native

Agreed, wasn't really able to center it though (margin top doesn't worK) and `vertical-align: middle` is not really in the middle... Used `vertical-align: middle;` anyway now though. If there are other ideas how to actually center it let me know :pray: 
Screenshot_20231127-140306_IntegreatTestCms

ztefanie (approved):
Nice :+1: 

Please keep in mind, we have currently a big need for good first issues. :D

Issue #2580, opened 2023-11-21
Search Icon not centered


f1sh1918:
### Steps to Reproduce

1. [Go to '...'](https://webnext.integreat.app/landing/de)
2. Check the magnifier icon next to the textfield

### Expected Behavior

Icon should be centered with the text

### Actual Behavior

the text icon is not centered

### Additional Information

<img width="630" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/37902063/90d8276c-340a-45b1-98f6-c89e703711f1">

PR #2599, opened 2023-11-28
2580: Center search icon

Njivaniaina2407:
Short description
The Search Icon is not centered with the input "Suche"

Proposed changes
Center and align the Search Icon with the input, set it a little lower

Side effects
Resolved issues
The Search Icon is aligned with the Input "Suche"

Fixes: https://github.com/digitalfabrik/integreat-app/issues/2580

---

### DIFF
--- a/web/src/components/SearchInput.tsx
+++ b/web/src/components/SearchInput.tsx
@@ -41,6 +41,8 @@ const Wrapper = styled.div`
 const StyledIcon = styled(Icon)`
   width: 20px;
   height: 20px;
+  position: relative;
+  top: 4px;
 `
 
 type SearchInputProps = {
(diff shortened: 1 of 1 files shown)


LeandraH (commented):
That solves the issue here very nicely, thank you! But I think we can make the code more future-proof by setting the parent display to flex and aligning its items with the center. That would also keep us safe in case we e.g. change the icon sizing at some point.

f1sh1918:
> That solves the issue here very nicely, thank you! But I think we can make the code more future-proof by setting the parent display to flex and aligning its items with the center. That would also keep us safe in case we e.g. change the icon sizing at some point.

sorry i probably should have mentioned it again in the ticket here. This task was about learning sth. about flexbox :)

f1sh1918 (commented):
Good Work :) Added a comment about `inline-flex`

LeandraH (approved):
Nicely done!

f1sh1918 (approved):

Issue #2603, opened 2023-12-04
Screen is missing background opacity when feedback open


sarahsporck:
### Describe the Bug

When opening the feedback modal in web the background color is missing.

### Steps to Reproduce

1. Use firefox, doesn't happen on Chrome
2. Go to [testumgebung](https://webnext.integreat.app/testumgebung/de)
3. Click on 'Feedback'
4. See missing background

### Expected Behavior

Tha background color is present, as it was in the past

### Actual Behavior

There is no background

### Additional Information

Recorded in firefox:
<img width="999" alt="Bildschirmfoto 2023-12-04 um 10 16 09" src="https://github.com/digitalfabrik/integreat-app/assets/28013368/6c7bf9c4-8abd-43f5-b80c-25777fadc056">

steffenkleinle:
Introduced in #2578.

PR #2604, opened 2023-12-04
2603: Screen is missing background opacity when feedback open

Njivaniaina2407:
### Short description

When opening the feedback modal in web the background color is missing.
The background color does not appear on Firefox but does on Chrome 

### Proposed changes

A 100% width and height to be added to the object that seems to have disappeared

### Resolved issues

Fixes: #2603

---


LeandraH (approved):
Great find!

PR #2605, opened 2023-12-04
1213: Embed videos native

f1sh1918:
### Short description

I decided to split up this issue to web and native, since it was more effort than expected.
In this pr i implemented an `opt-in` section in settings to manage external sources for iframes. Since we may have more sources in future, the sources are easily extendible via `buildConfig`.

Important: This issue is about embedding external video sources via iframe (vimeo, youtube, etc), not making html videos working (external sources blocked by csp and video upload restricted by cms)

### Proposed changes

- if no permissions to external sources are set (initial state) show a opt-in with a checkbox to allow this external source
- all permission to external sources can be managed in a separate settings section to withdraw or allow a permission again

### Side effects
- hopefully none. please proper test internal/external and deep links on ios and android

### Testing
- go to http://localhost:9000/testumgebung/de/video-test-verschiedene-sources
- to get the initial unset permission state again, just copy this into `RemoteContent`
`appSettings.setExternalSourcePermissions([]).catch(reportError)`
- you can just add several more source in `build-config` f.e.
`allowedIframeSources: ['vimeo.com', 'youtube.com', 'tagesschau.de'],`

Note: For web i will create a separate issue using cookies

fixes: #1213


native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,144 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+    function handleAllowedIframeSources(iframe, allowedExternalSourcePermissions, iframeSource, iframeContainer) {
+      if (externalSourcePermissions.some(source => source.type === iframeSource && source.allowed)) {
+        // Add do not track parameter (only working for vimeo)
+        if (iframeSource === 'Vimeo') {
+          const url = new URL(iframe.src)
+          url.searchParams.append('dnt', '1')
+          iframe.setAttribute('src', url.href)
+        }
+        const message = ${JSON.stringify(
+          t('remoteContent:knownResourceContentMessageOne'),
```
f1sh1918, line 172: i couldn't find a proper way to pass variables (except of defined strings) to a passed function inside this `JSON.stringify`
If someone finds a solution that would simplify a lot.

native/src/components/RemoteContent.tsx
```diff
@@ -57,29 +72,67 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
...
+        reportError(error)
+        setError(error.message)
+      }
+    },
+    [externalSourcePermissions, navigateTo],
+  )
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
+      if (buildConfig().allowedIframeSources.some(source => event.url.indexOf(source) > 0)) {
```
sarahsporck, line 125:
```suggestion
      if (buildConfig().whitelistedIframeSources.some(source => event.url.indexOf(source) > 0)) {
```

native/src/components/RemoteContent.tsx
```diff
@@ -99,13 +152,22 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
     <WebView
       source={{
         baseUrl: resourceCacheUrl,
-        html: renderHtml(content, cacheDictionary, theme, language),
+        html: renderHtml(
+          content,
+          cacheDictionary,
+          buildConfig().allowedIframeSources,
```
sarahsporck, line 158:
```suggestion
          buildConfig().whitelistedIframeSources,
```

native/src/components/Consent.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { updateSourcePermissions } from '../utils/helpers'
+import { reportError } from '../utils/sentry'
+import Caption from './Caption'
+import ConsentSection from './ConsentSection'
+import Layout from './Layout'
+
+export const ItemSeparator = styled.View`
+  background-color: ${props => props.theme.colors.textDecorationColor};
+  height: ${StyleSheet.hairlineWidth}px;
+`
```
steffenkleinle, line 20: Optional: We already have some Divider/Separator components in use (e.g. HorizontalLine, Divider in two files). Maybe we should merge them at some point as well.

native/src/components/Consent.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+  const getInitialValue = (src: string): boolean =>
+    externalSourcePermissions.find(source => source.type === src)?.allowed ?? false
+
+  return (
+    <Layout>
+      <Caption title={t('headline')} />
+      <ItemSeparator />
+      <View>
+        {capitalizeFirstLetter(buildConfig().allowedIframeSources).map(src => (
```
steffenkleinle, line 45: We should probably use a list for this

native/src/components/Consent.tsx
```diff
@@ -0,0 +1,59 @@
...
+    const updatedSources = updateSourcePermissions(externalSourcePermissions, { type, allowed })
+    setExternalSourcePermissions(updatedSources)
+    appSettings.setExternalSourcePermissions(updatedSources).catch(reportError)
+  }
+  const getInitialValue = (src: string): boolean =>
+    externalSourcePermissions.find(source => source.type === src)?.allowed ?? false
+
+  return (
+    <Layout>
+      <Caption title={t('headline')} />
```
steffenkleinle, line 42: Optional: We probably have to discuss this but in my opinion we should use speaking names for translations as well that tell me what they are instead of where they are used. Reading this I have no idea what kind of text is shown and I think its just an implementation detail.

sarahsporck (commented):
Couldn't get it to work on ios and android simulator. Do you need a real device? When opening the video test page I get -> `[Error: "Message: ReferenceError: Can't find variable: el - URL: http://localhost:8080/ - Line: 236 - Column: 46 - Error object: {}"]` which seems to be a cms issue due to the localhost:8080. Am I missing sth?

native/src/components/ConsentSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+            true: theme.colors.themeColor,
+            false: theme.colors.textSecondaryColor,
+          }}
+          accessibilityRole='switch'
+          value={allow}
+          onValueChange={val => {
+            setAllow(val)
+            onPress(title, val)
+          }}
+        />
```
steffenkleinle, line 55: There is already a `SettingsSwitch` component that has exactly the same props/styling. Maybe you could reuse that? If you want, you can already move it to the `base` directory.

native/src/components/ConsentSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+  initialSwitchValue: boolean
+  onPress: (type: string, value: boolean) => void
+}
+
+const ConsentSection = ({ title, description, initialSwitchValue, onPress }: ConsentSectionProps): ReactElement => {
+  const [allow, setAllow] = useState<boolean>(initialSwitchValue)
+
+  useEffect(() => {
+    setAllow(initialSwitchValue)
+  }, [initialSwitchValue])
```
steffenkleinle, line 33: Do we actually need this? Or does this mean that this state is persisted twice (once here and once in the parent component/settings)? If that is the case, there should be no state in this component and the value should just be passed down (renamed to `allowed` or similar).

translations/translations.json
```diff
@@ -8633,7 +8659,9 @@
       "trackingLeaveDescription": "Du hast das Tracking nicht erlaubt. Das Tracking hilft uns, die {{appName}}-App zu verbessern und bessere Informationen für Einwanderer bereitzustellen. Bist du sicher, dass du den Bildschirm verlassen möchten, ohne das Tracking zu erlauben?",
       "enabled": "aktiviert",
       "disabled": "deaktiviert",
-      "decline": "Ablehnen"
+      "decline": "Ablehnen",
+      "externalResourcesTitle": "Externe Quellen",
+      "externalResourcesDescription": "Legen sie fest welche externen Quellen zugelassen werden"
```
steffenkleinle, line 8664:
```suggestion
      "externalResourcesDescription": "Legen sie fest, welche externen Quellen zugelassen werden."
```

translations/translations.json
```diff
@@ -8772,7 +8800,9 @@
       "trackingLeaveDescription": "You have disabled tracking. Tracking helps us to improve the {{appName}} app and provide better information for immigrants. Are you sure you want to leave the screen without enabling tracking?",
       "enabled": "activated",
       "disabled": "deactivated",
-      "decline": "Refuse"
+      "decline": "Refuse",
+      "externalResourcesTitle": "External Sources",
+      "externalResourcesDescription": "Specify which external sources are permitted"
```
steffenkleinle, line 8805:
```suggestion
      "externalResourcesDescription": "Specify which external sources are permitted."
```

translations/translations.json
```diff
@@ -8408,6 +8408,32 @@
       "showOwnLocation": "显示自己的位置"
     }
   },
+  "remoteContent": {
+    "de": {
+      "knownResourceOptIn": "Ich bin damit einverstanden, dass mir Inhalte dieser externen Quelle angezeigt werden: ",
+      "knownResourceContentMessageOne": "Dieses Element enthält Daten von ",
```
steffenkleinle, line 8414: Since the word order might be different in other languages, the variable should be included in the translation. Same below. If `knownResourceContentMessageTwo` is used directly after, they could be merged.

native/src/utils/AppSettings.ts
```diff
@@ -16,6 +19,7 @@ export type SettingsType = {
   jpalTrackingEnabled: boolean | null
   jpalTrackingCode: string | null
   jpalSignals: Array<SignalType>
+  externalSourcePermissions: ExternalSourcePermission[]
```
steffenkleinle, line 22: Maybe a key/value map would be easier to manage? No need to check for duplicates and so on.

native/src/components/RemoteContent.tsx
```diff
@@ -57,29 +72,67 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
...
-      reportError(error)
-      setError(error.message)
+  // Webview content does not reload on opt-in click
+  useEffect(() => {
+    if (reloadSettings) {
+      appSettings.loadExternalSourcePermissions().then(setExternalSourcePermissions).catch(reportError)
+      setReloadSettings(false)
     }
-  }, [])
+  }, [reloadSettings])
```
steffenkleinle, line 87: Why did you choose to use a `useEffect` for this instead of just calling a function directly?

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+
+    function showBlockMessageWithSettings(text, iframeContainer) {
+      showMessage(text, iframeContainer)
+      showSettingsButton(iframeContainer)
+    }
+
+    function handleAllowedIframeSources(iframe, allowedExternalSourcePermissions, iframeSource, iframeContainer) {
+      if (externalSourcePermissions.some(source => source.type === iframeSource && source.allowed)) {
+        // Add do not track parameter (only working for vimeo)
+        if (iframeSource === 'Vimeo') {
```
steffenkleinle, line 165: Shouldn't this be `vimeo`?

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
-    window.addEventListener('resize', adjustHeight);
-    var details = document.querySelectorAll("details")
-    details.forEach(detail => detail.addEventListener("toggle", adjustHeight))
+    window.addEventListener('load', adjustHeight)
+    window.addEventListener('resize', adjustHeight)
+    const details = document.querySelectorAll('details')
+    details.forEach(detail => detail.addEventListener('toggle', adjustHeight))
   })();
+
+  (function handleIframes() {
```
steffenkleinle, line 108: I haven't checked for alternatives yet but this feels super hard to read/review and looks pretty error prone. Maybe we should discuss this a little more. Or are we sure that this is the best option?

native/src/components/RemoteContent.tsx
steffenkleinle: Why? I think we should not use the terms `whitelisted` or `blacklisted` anymore because of discriminating language.

translations/translations.json
LeandraH: I think we should stick to the informal you that we're using in the other translations.

native/src/components/RemoteContent.tsx
sarahsporck: Jap. You're right. Just couldn't test this without changing 🤷🏼‍♀️

native/src/utils/renderHtml.ts
f1sh1918: dunno what other possibilities we have inside this template literals

native/src/utils/AppSettings.ts
f1sh1918: you can not use Maps in this template literal because `Map` is es6 @steffenkleinle

f1sh1918:
> Couldn't get it to work on ios and android simulator. Do you need a real device? When opening the video test page I get -> `[Error: "Message: ReferenceError: Can't find variable: el - URL: http://localhost:8080/ - Line: 236 - Column: 46 - Error object: {}"]` which seems to be a cms issue due to the localhost:8080. Am I missing sth?

not sure if you are on the correct branch @sarahsporck i can't find any `el` in the code

native/src/components/ConsentSection.tsx
f1sh1918: ok i found the issue. i didn't check for externalSource to be not empty

f1sh1918: ok i had to add a check if externalPermissions where loaded before rendering the ConsentSections

native/src/components/Consent.tsx
f1sh1918: you mean the react native component or the data type?

native/src/components/Consent.tsx
f1sh1918: so `title` would be appropriate?

native/src/utils/renderHtml.ts
f1sh1918: it was capitalized before but i adjusted the check and added a `toLowerCase()`

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+      iframeContainer.appendChild(label)
+      iframeContainer.appendChild(checkbox)
+    }
+
+    function showBlockMessageWithSettings(text, iframeContainer) {
+      showMessage(text, iframeContainer)
+      showSettingsButton(iframeContainer)
+    }
+
+    function handleAllowedIframeSources(iframe, allowedExternalSourcePermissions, iframeSource, iframeContainer) {
```
steffenkleinle, line 162: Maybe a better name is actually `supportedIframeSources` instead of allowed since they are only allowed from our perspective and not from the users perspective and this might make the logic more clear. At least I was confused about the two different types of "allowed". If you agree, please rename the params and build config properties as well.

native/src/utils/renderHtml.ts
steffenkleinle: I think its just not possible sadly since the functions are evaluated before the html is executed.

One idea though: I think we should write the messages in the way that they are universal and not iframe source specific, e.g.
```
vimeo.com
External content from this source was blocked. You can allow this source in the settings.
```
Then it should a) be easier to read/implement and b) more error proof in other translations (bc of things like word order etc.).

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+          const url = new URL(iframe.src)
+          url.searchParams.append('dnt', '1')
+          iframe.setAttribute('src', url.href)
+        }
+        const message = ${JSON.stringify(
+          t('remoteContent:knownResourceContentMessageOne'),
+        )} + iframeSource + '. ' + ${JSON.stringify(t('remoteContent:knownResourceContentMessageTwo'))}
+          showMessageWithSettings(message, iframeContainer)
+      } else if (allowedExternalSourcePermissions.some(source => source.type === iframeSource && !source.allowed)) {
+          const translation = ${JSON.stringify(t('remoteContent:knownResourceBlocked'))}
```
steffenkleinle, line 175: All these `JSON.stringify`s can be avoided by using `'` quotes:
```suggestion
          const translation = '$(t('remoteContent:knownResourceBlocked')}'
```

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+    const iframes = document.querySelectorAll('iframe')
+    const allowedIframeSources = ${JSON.stringify(allowedIframeSources)}
+    const externalSourcePermissions = ${JSON.stringify(externalSourcePermissions)}
+
+    iframes.forEach((iframe) => {
+      const iframeContainer = document.createElement('div')
+      iframeContainer.classList.add('iframe-container')
+      iframe.parentNode.appendChild(iframeContainer)
+      const allowedIframeSource = allowedIframeSources.find(src => iframe.src.includes(src))
+      if (allowedIframeSource) {
```
steffenkleinle, line 198:
```suggestion
      const isSupported = allowedIframeSources.find(iframe.src.includes)
      if (allowedIframeSource) {
```

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+      iframeContainer.appendChild(checkbox)
+    }
+
+    function showBlockMessageWithSettings(text, iframeContainer) {
+      showMessage(text, iframeContainer)
+      showSettingsButton(iframeContainer)
+    }
+
+    function handleAllowedIframeSources(iframe, allowedExternalSourcePermissions, iframeSource, iframeContainer) {
+      if (allowedExternalSourcePermissions.some(source => source.type === iframeSource && source.allowed)) {
```
steffenkleinle, line 163:
```suggestion
      const supportedExternalSource = allowedExternalSourcePermissions.find(source => source.type === iframeSource)
      if (supportedExternalSource?.allowed) {
```

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+        if (iframeSource.toLowerCase() === 'vimeo') {
+          const url = new URL(iframe.src)
+          url.searchParams.append('dnt', '1')
+          iframe.setAttribute('src', url.href)
+        }
+        const message = ${JSON.stringify(
+          t('remoteContent:knownResourceContentMessageOne'),
+        )} + iframeSource + '. ' + ${JSON.stringify(t('remoteContent:knownResourceContentMessageTwo'))}
+          showMessageWithSettings(message, iframeContainer)
+      } else if (allowedExternalSourcePermissions.some(source => source.type === iframeSource && !source.allowed)) {
```
steffenkleinle, line 174:
```suggestion
      } else if (supportedExternalSource?.allowed === false)) {
```

native/src/utils/renderHtml.ts
```diff
@@ -71,34 +84,143 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+          const translation = ${JSON.stringify(t('remoteContent:knownResourceBlocked'))}
+          const message = iframeSource + ' ' + translation
+          showBlockMessageWithSettings(message, iframeContainer)
+          iframe.remove()
+      }
+      // No permissions set for listed sources
```
steffenkleinle, line 180: Should be clear enough by the code, e.g. by changing the order and using the suggestion above
```suggestion
```
E.g.
``` ts
      const supportedExternalSourcePermission = allowedExternalSourcePermissions.find(source => source.type === iframeSource)
      if (!supportedExternalSourcePermission) {
        // show opt in
      } else if (supportedExternalSourcePermission.allowed) {
        // show iframe
      } else {
        // show blocked message
      }
```

native/src/components/Consent.tsx
steffenkleinle: I mean our own `List` component, not sure what you mean with `data type` in this case.

translations/translations.json
steffenkleinle: Yeah missed that, sorry.

native/src/utils/renderHtml.ts
f1sh1918: alright i also changed them in build config

native/src/utils/renderHtml.ts
f1sh1918: outdated since switched to record

native/src/components/ConsentSection.tsx
```diff
@@ -22,12 +22,12 @@ const Description = styled(Text)`
...
   description: string
-  initialSwitchValue: boolean
+  allowed: boolean
   onPress: (type: string, value: boolean) => void
 }
 
-const ConsentSection = ({ title, description, initialSwitchValue, onPress }: ConsentSectionProps): ReactElement => {
-  const [allow, setAllow] = useState<boolean>(initialSwitchValue)
+const ConsentSection = ({ title, description, allowed, onPress }: ConsentSectionProps): ReactElement => {
+  const [allow, setAllow] = useState<boolean>(allowed)
```
steffenkleinle, line 30: This can be removed :see_no_evil: Otherwise we have duplicated state
```suggestion
```

The `onPress` method updates the state in the parent component already.

native/src/components/RemoteContent.tsx
```diff
@@ -113,7 +113,7 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
 
   const onShouldStartLoadWithRequest = useCallback(
     (event: WebViewNavigation): boolean => {
-      if (buildConfig().allowedIframeSources.some(source => event.url.indexOf(source) > 0)) {
+      if (buildConfig().supportedIframeSources.some(source => event.url.indexOf(source) > 0)) {
```
steffenkleinle, line 116:
```suggestion
      if (buildConfig().supportedIframeSources.some(source => event.url.includes(source))) {
```
or even
```suggestion
      if (buildConfig().supportedIframeSources.some(event.url.includes)) {
```

translations/translations.json
```diff
@@ -8411,17 +8411,13 @@
   "remoteContent": {
     "de": {
       "knownResourceOptIn": "Ich bin damit einverstanden, dass mir Inhalte dieser externen Quelle angezeigt werden: ",
-      "knownResourceContentMessageOne": "Dieses Element enthält Daten von ",
-      "knownResourceContentMessageTwo": "Du kannst die Einbettung solcher Inhalte hier blockieren: ",
-      "knownResourceBlocked": "wurde geblockt. Unter Einstellungen können sie Inhalte dieser Quelle zulassen.",
-      "unknownResourceBlocked": "Ein Inhalt von einer externen Quelle wurde geblockt:"
+      "knownResourceContentMessage": "Dieses Element enthält Daten von dieser externen Quelle. Du kannst die Einbettung solcher Inhalte in den Einstellungen blockieren",
+      "knownResourceBlocked": "Externer Inhalt von dieser Quelle wurde geblockt. Unter Einstellungen kannst du diese Quelle zulassen."
```
steffenkleinle, line 8415:
```suggestion
      "knownResourceBlocked": "Externer Inhalt von dieser Quelle wurde blockiert. In den Einstellungen kannst du diese Quelle zulassen."
```

translations/translations.json
```diff
@@ -8411,17 +8411,13 @@
...
     },
     "en": {
       "knownResourceOptIn": "I agree that content from this external source may be displayed to me: ",
-      "knownResourceContentMessageOne": "This element contains data from ",
-      "knownResourceContentMessageTwo": "You can block the embedding of such content here: ",
-      "knownResourceBlocked": "has been blocked. You can allow content from this source in settings.",
-      "unknownResourceBlocked": "Content from an external source has been blocked:"
+      "knownResourceContentMessage": "This element contains data from that external source. You can block the embedding of such content in settings.",
```
steffenkleinle, line 8419:
```suggestion
      "knownResourceContentMessage": "This element contains data from this external source. You can block the embedding of such content in the settings.",
```

translations/translations.json
```diff
@@ -8411,17 +8411,13 @@
...
+      "knownResourceBlocked": "External content from this source has been blocked. You can allow this source in settings."
```
steffenkleinle, line 8420:
```suggestion
      "knownResourceBlocked": "External content from this source has been blocked. You can allow this source in the settings."
```

native/src/utils/renderHtml.ts
```diff
@@ -33,72 +52,174 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+
+    function showMessageWithSettings(text, iframeContainer, iframeSource) {
+      showMessage(text, iframeContainer, 'iframe-info-text', iframeSource)
+      showSettingsButton(iframeContainer)
+    }
+
+    function showOptIn(text, iframeContainer, source) {
+      function onClickHandler() {
+        window.ReactNativeWebView.postMessage(
+          JSON.stringify({ type: '${IFRAME_MESSAGE_TYPE}', allowedSource: { type: source, allowed: true } }),
```
steffenkleinle, line 151: Optional: Since `allowed` is only ever set to `true`, we could change this to:
```suggestion
          JSON.stringify({ type: '${ALLOW_EXTERNAL_SOURCE_MESSAGE_TYPE}', source }),
```

steffenkleinle (approved):
Tested on android, amazing job!

native/src/components/ConsentSection.tsx
f1sh1918: true realized that in the web implementation

steffenkleinle:
@f1sh1918 created a proposal PR to fix the duplicated state: #2614

f1sh1918:
> @f1sh1918 created a proposal PR to fix the duplicated state: #2614

Thx everything is merged and should be fine now.

native/src/utils/renderHtml.ts
```diff
@@ -33,72 +54,191 @@ const renderJS = (cacheDictionary: Record<string, string>) => `
...
+
+    const iframes = document.querySelectorAll('iframe')
+    const supportedIframeSources = ${JSON.stringify(supportedIframeSources)}
+
+    iframes.forEach(iframe => {
+      const supportedIframeSource = supportedIframeSources.find(src => iframe.src.includes(src))
+      if (supportedIframeSource) {
+        handleSupportedIframeSources(iframe, supportedIframeSource)
+      } else {
+        iframe.remove()
```
ztefanie, line 223: I think it would be better to show a short message, about whats going on. "A video was implemented here, but cannot be displayed due to content policies" or whatever.

ztefanie (commented):
Tested on real Android device, works as expected.

native/src/utils/renderHtml.ts
f1sh1918: well i've implemented it like this before, but as discussed before with @steffenkleinle we decided not to make it more complicated than needed since the authors should be informed about which content can be embedded and which not. So this case shouldn't happen

ztefanie (approved):

Issue #2316, opened 2023-08-14
Zammad contact form


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Malte needs a new offer which allows embedding/attaching a contact form into a page.
The collected information shoul look sth. like this:
```
Zimmernummer:

Name:

[x] Ich möchte digital kontaktiert werden:
---- 8< -----
(Eins von beiden Pflicht, wenn Häkchen gesetzt. Ansonsten verschwindet die Auswahl.)

Telefonnummer:

E-Mail:
---- >8 -----

Ich möchte von einer [männlichen/weiblichen/egal] Person kontaktiert werden.

Optional Textfeld:
```

**Describe the solution you'd like**
A contact form is rendered (as an offer) for Malteser.

- [x] create an offer "hilfebutton" similar to sprungbrett
- [ ] implement the design for the "hilfebutton" form
- [x] ~~as the backend ticket is not yet merged & released hard code the form in the content and comment it out before merging~~

**Describe alternatives you've considered**
- HTML forms in content

**Additional context**
- #2315 
- https://nextcloud.tuerantuer.org/index.php/f/4782047
- https://nextcloud.tuerantuer.org/index.php/f/3906203

sarahsporck:
@hauf-toni We can schedule an appointment, to discuss what's necessary for this.

LeandraH:
Here are some user stories from Malteser for this: https://nextcloud.tuerantuer.org/index.php/apps/files/?dir=/Projects/Malte/50_Notizen%20und%20Themensammlungen%20f%C3%BCr%20Calls&fileid=522263

hauf-toni:
Design-proposal can be found [📍 here ](https://www.figma.com/file/EiRpMhSqscuaChGdhNVZQ4/MALTE-App?type=design&node-id=3%3A22&mode=design&t=BFPUggPWcwUwFBC4-1)

LeandraH:
Looks great!

sarahsporck:
@osmers it would be nice if you could also look over the design proposal whether everything is present :)

osmers:
@sarahsporck I'll have a look tmr :)

osmers:
@sarahsporck I made some comments within the design - but overall I think it looks really good.

Issue #2497, opened 2023-10-04
Better Contrast for Malte Buttons


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Malte buttons do have this incredibly bright-red-danger background. While the text is displayed in some grayish color. This is not only badly accessible, but also hard to read in general.

**Describe the solution you'd like**
Calculate font color from background color, when using colored backgrounds.

**Describe alternatives you've considered**
Use static font colors in build configs for colored backgrounds.

**Additional context**
<img width="300" alt="Bildschirmfoto 2023-10-04 um 09 41 27" src="https://github.com/digitalfabrik/integreat-app/assets/28013368/e3a3e4bf-8d80-4ad1-9e6a-b35432a4bded"><img width="331" alt="Bildschirmfoto 2023-10-04 um 09 44 12" src="https://github.com/digitalfabrik/integreat-app/assets/28013368/996cee3e-4b43-4a34-a1ab-94b5be351117">

hauf-toni:
design/color-proposal can be found [📍 here](https://www.figma.com/file/EiRpMhSqscuaChGdhNVZQ4/MALTE-App?type=design&node-id=54%3A114&mode=design&t=DvW55uNHC9uYQzIX-1) @sarahsporck

hauf-toni:
@sarahsporck I've added a screen design for the feedback view (default + seleocted). Let me know if you have any questions 📭

sarahsporck:
exactly. red is malte and green is aschaffenburg. Both suffer from bad contrast. I also mentioned this in the figma design.

f1sh1918:
> Is this one ready or still in review now @f1sh1918? :) It's a little bit unclear where we ended off. 

I think we have to check the Design but atm I have to time for that.

steffenkleinle:
@hauf-toni is this actually `In Review`? Is this waiting for #2414?

hauf-toni:
yes, it is not waiting for #2414 . please leave some feedback or alternatively implement these changes =)

steffenkleinle:
> yes, it is not waiting for #2414 . please leave some feedback or alternatively implement these changes =)

Ok, thank you. I think this would need some changes on our side and we should try to make this uniform for all build configs (i.e. also for integreat, aschaffenburg and obdach) in a way that the logic is the same (but the specific colors obviously differ). Otherwise, we'd need a lot of logic dependent on the currently used build config.

[...]

Issue #2467, opened 2023-09-23
Delete offers tab


sarahsporck:
## New description

**Is your feature request related to a problem? Please describe.**
We want to remove the offers tab and move all of them to embedded offers.

**Describe the solution you'd like:**
Remove all offer specific routes, pages, helpers, ... that are not needed for embedded offers.
Make sure embedded offers continue to work.
Remove screenshots, intro slide, ...

**Testing information**
Make sure embedded offers continue to work.

## Old description
As a first step:
- Offers should be moved to local information. That means:  If there is no "Arbeit" 1st level category all the current offers (as they are all work related) should be childs of that 1st level "Arbeit category)
- If there already is an "Arbeit" 1st level category, a 2nd level category "Offers" should be added as a child which has all the current offers as child.
- In the offers tab, there should be a hint that offers where moved including an internal link to the 1st or 2nd level category where the offers are now shown.

Acceptance criteria:
- Offline usage is handled on native (either offers are saved offline as well or at least a good error is displayed)
- Hint shown in the offers tab
- Offers as child of 1st or 2nd level category
- Delete offers screenshot from store metadata

Things to be discussed/unclear:
- If page based offers are nearly ready, does it really make sense to do this first and than change it a short time later again?
- How can we now in the offers if we have to link to a 1st or 2nd level category?
- How do we know in other languages than German whether there is a 1st level category "Arbeit"? Do we check depending on the title or the path or something else?
- How to handle offline usage?
- Can we find the offers category from the search?
- How does feedback work?

In the future we will remove the offers navigation item and change to page based offers (#2313).

steffenkleinle:
I think this is blocked by #2313 right?

sarahsporck:
Hmm. Don't really think so. We discussed this at the conference. And our first decision was to just move all the offers into one Root Category, I think. @ztefanie do you remember?

ztefanie:
Yes I also remember this was discussed and also @dkehne agreed on this?

steffenkleinle:
I adjusted the description with things that are unclear here/not that easy imo. I think this is not worth it if we are switching to page based offers with support from the cms and a new design anyway...

ztefanie:
Please also delete the offers-tab screenshots for the store presence (google and apple)

PR #2609, opened 2023-12-06
2316: Zammad form web

sarahsporck:
### Short description

Implement the web zammad contact form. 
Regarding a few things I'm still awaiting feedback from toni, that's why this is still a draft.

### Proposed changes

- add submit function to api client
- add embedded_offer to categories
- adjust inputsection component for material ui
- add radio button base components
- implement zammad form for web

Test here: 
- http://localhost:9000/augsburg/de/hilfebutton-test
- http://localhost:9000/augsburg/de/offers/help

### Resolved issues

Fixes: #2316 

---


translations/translations.json
```diff
@@ -9604,5 +9604,26 @@
     "zh-CN": {
       "noOffersAvailable": "目前没有提供项目"
     }
+  },
+  "zammad": {
```
steffenkleinle, line 9608: I think this should be renamed, using `zammad` is just an implementation detail and not really speaking if you don't know what it is.
If we choose to stick to malte specific naming it should probably be 
```suggestion
  "malteHelpForm": {
```

translations/translations.json
```diff
@@ -9604,5 +9604,26 @@
...
+    "de": {
+      "supportNote": "Wir möchten, dass Sie sich sicher und unterstützt fühlen, wenn Sie Probleme oder Fragen haben. Durch das folgende Kontaktformular haben Sie die Möglichkeit, mit Ansprechpartner:innen in Kontakt zu treten. Sie können die Art Ihres Anliegens angeben und aus verschiedenen Kontaktoptionen wählen.",
+      "securityNote": "Ihre Sicherheit hat für uns oberste Priorität. Ihre Daten werden mit Vertraulichkeit behandelt, damit Ihre Privatsphäre jederzeit gewahrt bleibt.",
+      "name": "Name, Vorname",
+      "roomNumber": "Zimmernummer",
+      "howToBeContacted": "So möchte ich Kontaktiert werden:",
```
steffenkleinle, line 9614:
```suggestion
      "howToBeContacted": "So möchte ich kontaktiert werden:",
```

translations/translations.json
```diff
@@ -9604,5 +9604,26 @@
...
+      "eMail": "E-Mail",
+      "telephone": "Telefon",
+      "telephoneNumber": "Tel.-Nr.",
+      "inPerson": "in Person",
+      "contactPerson": "Kontaktperson:",
+      "contactPersonAnyGender": "Das Geschlecht des Ansprechpartners/der Ansprechpartnerin spielt keine Rolle",
+      "contactPersonGenderFemale": "Ich möchte von einer weiblichen Person kontaktiert werden",
+      "contactPersonGenderMale": "Ich möchte von einer männlichen Person kontaktiert werden",
+      "causeOfContact": "Kurze Beschreibung des Anliegens:",
```
steffenkleinle, line 9623:
```suggestion
      "contactReason": "Kurze Beschreibung des Anliegens:",
```

web/src/components/base/Button.tsx
```diff
@@ -11,21 +11,30 @@ const StyledButton = styled.button<{ disabled: boolean }>`
 `
 
 type ButtonProps = {
-  onClick: () => void
+  onClick?: () => void
```
steffenkleinle, line 14: AFAICS this is never undefined
```suggestion
  onClick: () => void
```

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -0,0 +1,57 @@
...
+  languageCode: string
+  baseUrl: string | (() => Promise<string>)
+}
+
+type SubmitHelpFormReturn = Promise<{
+  helpButtonOffer: OfferModel
+  submit: () => Promise<void>
+}>
+
+const submitHelpForm = async ({ cityCode, languageCode, baseUrl }: SubmitHelpFormParams): SubmitHelpFormReturn => {
```
steffenkleinle, line 17: Why did you choose to custom implement this instead of using the endpoint classes (like in createFeedbackEndpoint)?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -0,0 +1,57 @@
...
+  const apiUrl = typeof baseUrl === 'string' ? baseUrl : await baseUrl()
+  const helpButtonOfferAlias = MALTE_HELP_FORM_OFFER_ROUTE
+  const offersPayload = await createOffersEndpoint(apiUrl).request({
+    city: cityCode,
+    language: languageCode,
+  })
```
steffenkleinle, line 23: Why do we need this here?

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+    0 2px 4px rgb(0 0 0 / 15%);
+
+  :hover {
+    background-color: ${props => props.theme.colors.textSecondaryColor}D9;
+  }
+
+  :active {
+    box-shadow: none;
+  }
+`
```
steffenkleinle, line 48: Why did you use custom styling here instead of directly applying it int the `TextButton` component and using that?

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+const Note = styled.div`
+  display: flex;
+  padding-bottom: 10px;
+`
+
+const StyledIcon = styled(Icon)`
+  width: 25px;
+  height: 25px;
+  padding-right: 20px;
```
steffenkleinle, line 19: This breaks on RTL layouts

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+  }
+`
+
+const Form = styled.form`
+  display: flex;
+  flex-direction: column;
+`
+
+type ContactChannel = 'email' | 'telephone' | 'person'
+type ContactType = 'male' | 'female' | 'any'
```
steffenkleinle, line 56:
```suggestion
type ContactGender = 'male' | 'female' | 'any'
```

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+  border-radius: 100px;
+  width: fit-content;
+  align-self: center;
+  padding: 15px;
+  box-shadow:
+    0 2px 4px rgb(0 0 0 / 15%),
+    0 2px 4px rgb(0 0 0 / 15%);
+
+  :hover {
+    background-color: ${props => props.theme.colors.textSecondaryColor}D9;
```
steffenkleinle, line 42: Hmm perhaps we should move that to a constant/use opacity instead?

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  return (
+    <>
+      <Note>
+        <StyledIcon src={SupportIcon} />
+        {t('zammad:supportNote')}
+      </Note>
+      <Note>
+        <StyledIcon src={SecurityIcon} />
+        {t('zammad:securityNote')}
```
steffenkleinle, line 98:
```suggestion
        {t('supportNote')}
      </Note>
      <Note>
        <StyledIcon src={SecurityIcon} />
        {t('securityNote')}
```

web/src/components/base/InputSection.tsx
```diff
@@ -22,47 +20,143 @@ export const Title = styled.label`
...
+  | {
+      title: string
+      compact?: false
+      subtitle?: string
+      description?: string
+    }
+  | { title: string; compact: true }
+) &
+  ({ optional: true; showOptional?: boolean } | { optional?: false; showOptional?: false }) &
+  Omit<InputProps, 'required'>
```
steffenkleinle, line 86: Why is this complicated type needed? What cases does this prevent?
For example I don't get that title needs to be of type string in both cases and compact can either be true or false or undefined?
Also, why do we need optional, showOptional and required?

translations/translations.json
```diff
@@ -9604,5 +9604,26 @@
...
+    "de": {
+      "supportNote": "Wir möchten, dass Sie sich sicher und unterstützt fühlen, wenn Sie Probleme oder Fragen haben. Durch das folgende Kontaktformular haben Sie die Möglichkeit, mit Ansprechpartner:innen in Kontakt zu treten. Sie können die Art Ihres Anliegens angeben und aus verschiedenen Kontaktoptionen wählen.",
+      "securityNote": "Ihre Sicherheit hat für uns oberste Priorität. Ihre Daten werden mit Vertraulichkeit behandelt, damit Ihre Privatsphäre jederzeit gewahrt bleibt.",
+      "name": "Name, Vorname",
+      "roomNumber": "Zimmernummer",
+      "howToBeContacted": "So möchte ich Kontaktiert werden:",
+      "eMail": "E-Mail",
+      "telephone": "Telefon",
+      "telephoneNumber": "Tel.-Nr.",
+      "inPerson": "in Person",
```
steffenkleinle, line 9618: Perhaps
```suggestion
      "personally": "Persönlich"
```

api-client/src/routes/index.ts
```diff
@@ -26,6 +26,9 @@ export const OFFERS_ROUTE: OffersRouteType = 'offers'
 export type SprungbrettOfferRouteType = 'sprungbrett'
 export const SPRUNGBRETT_OFFER_ROUTE: SprungbrettOfferRouteType = 'sprungbrett'
 
+export type MalteHelpFormOfferRouteType = 'hilfebutton'
+export const MALTE_HELP_FORM_OFFER_ROUTE: MalteHelpFormOfferRouteType = 'hilfebutton'
```
steffenkleinle, line 30: Shouldn't we use an english term here?

web/src/components/base/InputSection.tsx
```diff
@@ -22,47 +20,143 @@ export const Title = styled.label`
...
+        border-color: ${props => props.theme.colors.invalidInput};
+      }
+    `}
+`
+
+export const CompactTitle = styled.label`
+  position: relative;
+  padding: 0 4px;
+  top: 0.9rem;
+  left: 10px;
```
steffenkleinle, line 45: This breaks on RTL layouts

web/src/components/base/Button.tsx
sarahsporck: Jep. I changed that at some poin. But now it can be reverted, I think :)

api-client/src/endpoints/submitHelpForm.ts
sarahsporck: I tried following the pattern of the sprungbrett offers endpoint. I'll try finishing the page based offer pr today, so that I can correctly implement that. Using the offers endpoint should also be perfectly okay, you're right.

web/src/components/MalteHelpForm.tsx
sarahsporck: This is because the form already implements #2497, which is not yet merged yet. See design in figma -> https://www.figma.com/file/EiRpMhSqscuaChGdhNVZQ4/MALTE-App?type=design&node-id=3-22&mode=design&t=CoIwpX4BiwJXzTlC-0
However if you really dislike using a coustom styling I can also add a comment in #2497

web/src/components/base/InputSection.tsx
sarahsporck: There are cases where we apparently do not want to show that a text field is optional, even though it is optional. See [this discussion](https://www.figma.com/file/EiRpMhSqscuaChGdhNVZQ4?node-id=34:340&mode=design#593523521) in figma.

sarahsporck: This type encompasses the following options:
- compact format with only one title (which is also the label)
- non-compact format with title (which is the label) and an optional description and an optional subtitle

And the optional options:
- show that an input field is optional, when it is optional
- do not show that an input is optional, when it is optional
- a text field is required, therefore not optional, show Optional should be undefined or false

api-client/src/routes/index.ts
sarahsporck: I don't really have an opinion on that. Do you think just `/help` works better?

web/src/components/MalteHelpForm.tsx
```diff
@@ -15,10 +19,10 @@ const Note = styled.div`
   padding-bottom: 10px;
 `
 
-const StyledIcon = styled(Icon)`
+const StyledIcon = styled(Icon)<{ direction: 'ltr' | 'rtl' }>`
   width: 25px;
   height: 25px;
-  padding-right: 20px;
+  ${props => (props.direction === 'ltr' ? 'padding-right: 20px' : 'padding-left: 20px')};
```
steffenkleinle, line 25: Could we instead use a gap on the parent component to avoid these direction dependent styles?

web/src/components/base/InputSection.tsx
```diff
@@ -20,143 +20,24 @@ export const Title = styled.label`
...
-      required={required}
-      value={value}
-      maxLength={maxLength}
-      submitted={submitted}
-    />
-  )
+  title: string
+  description?: string
+  showOptional?: boolean
+  children: React.ReactComponentElement<typeof InputComponent>
```
steffenkleinle, line 28: Why do you pass the `Input` instead of rendering it directly here?

web/src/components/MalteHelpForm.tsx
```diff
@@ -209,17 +227,18 @@ const MalteHelpForm = ({ submit, dashboardRoute }: MalteHelpFormProps): ReactEle
...
-          value={comment}
-          onChange={setComment}
-          optional
-          maxLength={200}
-          submitted={submitted}
-        />
+        <InputSection id='comment' title={t('contactReason')}>
+          <Input
+            id='comment'
+            hint={t('maxCharacters', { numberOfCharacters: 200 })}
```
steffenkleinle, line 233: I think moving this to a shared constant in the api-client would be nice. Perhaps just in the same file as the `submit` happens.

translations/translations.json
```diff
@@ -9605,22 +9605,22 @@
...
       "telephone": "Telefon",
       "telephoneNumber": "Tel.-Nr.",
-      "inPerson": "in Person",
+      "personally": "in Person",
       "contactPerson": "Kontaktperson:",
       "contactPersonAnyGender": "Das Geschlecht des Ansprechpartners/der Ansprechpartnerin spielt keine Rolle",
       "contactPersonGenderFemale": "Ich möchte von einer weiblichen Person kontaktiert werden",
       "contactPersonGenderMale": "Ich möchte von einer männlichen Person kontaktiert werden",
-      "causeOfContact": "Kurze Beschreibung des Anliegens:",
+      "contactReason": "Kurze Beschreibung des Anliegens:",
```
steffenkleinle, line 9623: I think this could be shortened/simplified a little since the max chars text already tells it can't be a long text, e.g. just
```suggestion
      "contactReason": "Anliegen:",
```

api-client/src/routes/index.ts
steffenkleinle: I'd say so, yes. Depends a little if this is like an official term used in communication (like e.g. sprungbrett) or just an internal name.

web/src/components/MalteHelpForm.tsx
steffenkleinle: I think we should either change the styling for all buttons or for none. Otherwise we have different styles in different places of which at least one is not reusable/other people won't know about :see_no_evil:

web/src/CityContentSwitcher.tsx
```diff
@@ -49,6 +51,7 @@ export type CityRouteProps = {
   pathname: string
   cityCode: string
   languageCode: string
+  embedded?: boolean
```
steffenkleinle, line 54:
```suggestion
```

web/src/components/CityContentLayout.tsx
```diff
@@ -35,8 +36,13 @@ const CityContentLayout = (props: CityContentLayoutProps): ReactElement => {
     disableScrollingSafari = false,
     showFooter = true,
     city,
+    embedded,
   } = props
 
+  if (embedded) {
+    return <>{children}</>
+  }
+
```
steffenkleinle, line 45: What is this used for?

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  if (sendingStatus === 'successful') {
+    return (
+      <Container>
+        <div>{t('submitSuccessful')}</div>
+        <Link to={dashboardRoute}>{t('error:goTo.categories')}</Link>
+      </Container>
+    )
+  }
+
+  return (
```
steffenkleinle, line 132: This is quite big, maybe we can extract some code to some reusable components here?

translations/translations.json
```diff
@@ -9605,22 +9605,22 @@
...
       "maxCharacters": "maximal {{numberOfCharacters}} Zeichen",
       "responseDisclaimer": "Es kann einige Werktage dauern, bis Sie eine Rückmeldung erhalten. Wir bemühen uns, ihr Anliegen schnellstmöglich zu bearbeiten.",
```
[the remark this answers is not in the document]
sarahsporck: Probably "Sie", as the other texts already use formal language here.

web/src/components/CityContentLayout.tsx
sarahsporck: I've thought about moving this logic to the CitzContent Layout, to remove duplicate code. But as the offers tab will be removed in the future, all offers would also be "embedded", therefore I reverted this change again.

web/src/components/MalteHelpForm.tsx
sarahsporck: Not sure what you mean...

web/src/components/base/InputSection.tsx
sarahsporck: To move the input logic to the input component instead of passing all Input props through.

web/src/components/MalteHelpForm.tsx
sarahsporck: I think it's alright. As this file only contains the form. ^^ What would you split/remove?

assets/icons/security.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="24" viewBox="0 0 20 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
steffenkleinle, line 1: Can you adjust the two icons such that they are stylable via our icon component please? i.e. setting `fill="currentColor"`? Same for `support.svg`. Nearly all of our other icons have a size of 24x24 px.

native/src/routes/Offers.tsx
```diff
@@ -13,19 +13,21 @@ type OffersProps = {
...
-    if (offer.alias === SPRUNGBRETT_OFFER_ROUTE) {
+    if (offer.alias in [SPRUNGBRETT_OFFER_ROUTE, MALTE_HELP_FORM_OFFER_ROUTE]) {
       path = offer.alias
     }
     return new TileModel({
       title: t(offer.title),
       path,
       thumbnail: offer.thumbnail,
-      isExternalUrl: offer.alias !== SPRUNGBRETT_OFFER_ROUTE,
+      isExternalUrl: !internalOffers.includes(offer.alias),
```
steffenkleinle, line 30: Optional: Either use `.includes` or `offer.alias in` for both `isExternalUrl` and `path` (which could be inlined). Like this its a little confusing and I thought its doing two different things.

translations/translations.json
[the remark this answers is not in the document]
steffenkleinle: Hmm, all of our constants are `du` at the moment except the new ones for this feature. I asked in ~malteser for this, lets see what the consensus is there.

web/src/components/MalteHelpForm.tsx
steffenkleinle: You can add `gap: 20px;` on the parent component (in this case `Note`) and remove this line with the same effect and without the need for a condition/pass the direction as a prop.

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+import TextButton from './base/TextButton'
+
+const Note = styled.div`
+  display: flex;
+  padding-bottom: 10px;
+`
+
+const StyledIcon = styled(Icon)<{ direction: 'ltr' | 'rtl' }>`
+  width: 24px;
+  height: 24px;
```
steffenkleinle, line 24: This is set as a default already and can be removed.
```suggestion
```

web/src/components/MalteHelpForm.tsx
steffenkleinle: I guess most of it is fine/doesn't make sense to extract. However, I think you could create a `RadioGroup` component similar to what I did in native. That way, it would also be reusable for future use.

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+    !name.length ||
+    (!email.length && contactChannel === 'email') ||
+    (!telephone.length && contactChannel === 'telephone')
+  const dashboardRoute = cityContentPath({ languageCode, cityCode })
+
+  const submitHandler = useCallback(
+    (event: SyntheticEvent<HTMLFormElement>) => {
+      const form = event.currentTarget
+      if (!form.checkValidity()) {
+        // event.preventDefault()
```
steffenkleinle, line 83:
```suggestion
```

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+        invalidInput?.scrollIntoView({ behavior: 'smooth' })
+        setSendingStatus('idle')
+      } else {
+        setSendingStatus('sending')
+        const request = async () => {
+          await submitHelpForm({ cityCode, languageCode, helpButtonOffer })
+          setSendingStatus('successful')
+        }
+
+        request().catch(err => {
```
steffenkleinle, line 95: Please avoid mixing async/await and then/catch

web/src/routes/MalteHelpFormOfferPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+            })
+          }
+        />
+      </Wrapper>
+    )
+  }
+
+  return (
+    <Wrapper>
+      {!embedded && <Helmet pageTitle={pageTitle} languageChangePaths={languageChangePaths} cityModel={city} />}
```
steffenkleinle, line 104: Do we already have an issue to remove the offers page/tab?

web/src/components/MalteHelpForm.tsx
sarahsporck: Where is it mixed? 😅 I used the same logic as in the feedback container.

web/src/routes/MalteHelpFormOfferPage.tsx
sarahsporck: Jep. Maybe the description could be extended, though #2467.

web/src/components/base/RadioElement.tsx
```diff
@@ -0,0 +1,80 @@
...
+  groupId,
+  id,
+  label,
+  checked,
+  children,
+  onChange,
+}: Option<T>): ReactElement => (
+  <RadioElementContainer>
+    <RadioLabel htmlFor={label}>
+      <Radio
```
ztefanie, line 66: Radio should have role="radio"

web/src/components/MalteHelpForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+        <Input
+          id='roomNumber'
+          hint={`${t('roomNumber')} (${t('common:optional')})`}
+          hintIsLabel
+          value={roomNumber}
+          direction={config.getScriptDirection(languageCode)}
+          onChange={setRoomNumber}
+          submitted={submitted}
+        />
+        <RadioGroup caption={t('howToBeContacted')}>
```
ztefanie, line 144: RadioGroup should have `role="radiogroup"`
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role

ztefanie (commented):
How can this be tested?

translations/translations.json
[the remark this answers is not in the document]
sarahsporck: Consensus seems to be `du`

web/src/components/base/RadioGroup.tsx
```diff
@@ -0,0 +1,104 @@
...
+  align-self: center;
+  accent-color: ${props => props.theme.colors.textSecondaryColor};
+  ${props =>
+    props.direction === 'ltr'
+      ? css`
+          margin-right: 20px;
+        `
+      : css`
+          margin-left: 20px;
+        `};
```
steffenkleinle, line 33: Could we also use `gap: 20px` on the parent (RadioLabel)?

web/src/components/base/RadioElement.tsx
sarahsporck: For an input type="radio" you should not need the [aria role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role).
```
The first rule of ARIA is if a native HTML element or attribute has the semantics and behavior you require, use it instead of re-purposing an element and adding ARIA. Instead use the native [HTML <input type="radio">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) (with an associated [<label>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label), which natively provides all the functionality required
```

sarahsporck: Also https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role . `input type="radio"` implicitly har role "radio"

translations/translations.json
```diff
@@ -9630,5 +9630,29 @@
     "zh-CN": {
       "noOffersAvailable": "目前没有提供项目"
     }
+  },
+  "malteHelpForm": {
+    "de": {
+      "supportNote": "Wir möchten, dass Sie sich sicher und unterstützt fühlen, wenn Sie Probleme oder Fragen haben. Durch das folgende Kontaktformular haben Sie die Möglichkeit, mit Ansprechpartner:innen in Kontakt zu treten. Sie können die Art Ihres Anliegens angeben und aus verschiedenen Kontaktoptionen wählen.",
```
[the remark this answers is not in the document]
sarahsporck: Who should I ask for this? Or what to do to HIX optimize a text?

ztefanie (approved):
Looks good on Chrome :+1:

translations/translations.json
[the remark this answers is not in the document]
ztefanie: I think Laura is the only expert on this in our organization. Maybe there are online tools for that?

steffenkleinle (approved):
Not tested again, well done :)

Issue #2379, opened 2023-08-22
POI filter: show number of found POIs


Docmale:
**Is your feature request related to a problem? Please describe.**
As a user of a POI filter, I want to see if a filter has results or not, so I can see if the tap on "Orte anzeigen" is worth it or useless.

**Describe the solution you'd like**
The number of matching pois should be displayed in the `Orte anzeigen` button:
![image](https://github.com/digitalfabrik/integreat-app/assets/29095487/e8848259-5273-4751-b119-377a8f339ab7)

If no pois are matching, the button should be displayed and the message changed:
![image](https://github.com/digitalfabrik/integreat-app/assets/29095487/17ff1e73-326c-40ff-8919-812e9cc7b931)

See the design [here ](https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?type=design&node-id=1731%3A54&mode=design&t=5c6Wzwv2sdjqI6GX-1).

**Describe alternatives you've considered**
None.

**Additional context**
None.

f1sh1918:
Please provide a design for the filterbar @hauf-toni

hauf-toni:
📍 design proposal can be found [here ](https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?type=design&node-id=1731%3A54&mode=design&t=5c6Wzwv2sdjqI6GX-1)

LeandraH:
Looks good to me

steffenkleinle:
LGTM as well, thanks! Does anybody know if we have a ticket for a design somewhere for the category search (i.e. showing the number of search results in the following screen)?
<img src="https://github.com/digitalfabrik/integreat-app/assets/29095487/e47b6942-4ee0-4d57-9d97-fd66cbcf4ce3" width="200px" />

PR #2611, opened 2023-12-08
2379: Poi filter shows number of found pois

relentless95:
### Short description

The number of matching pois should be displayed in the ```Orte anzeigen``` button. If no pois are matching, the button should be deactivated and the message changed.
### Proposed changes

In the web app, i displayed the number of matching pois in the ```Orte anzeigen``` button and if there are no pois matching  the message is changed. 
However, i encountered an issue when trying the same thing in the native app.  
I could only display the number of matching pois and not the message displayed because it didn't pass the test. 
I also tweaked the test a little. 

### Side effects

None 
### Resolved issues

Fixes: #2379

---


native/src/components/PoiFiltersModal.tsx
```diff
@@ -78,6 +78,7 @@ type PoiFiltersModalProps = {
   setSelectedPoiCategory: (poiCategory: PoiCategoryModel | null) => void
   currentlyOpenFilter: boolean
   setCurrentlyOpenFilter: (currentlyOpen: boolean) => void
+  poisLength: number
```
steffenkleinle, line 81: Optional: I think something like `poisCount` or `numberOfPois` but be a little more clear

native/src/components/PoiFiltersModal.tsx
```diff
@@ -125,7 +127,7 @@ const PoiFiltersModal = ({
           </TileRow>
         </Section>
         <Section>
-          <StyledTextButton onPress={closeModal} text={t('showPois')} />
+          <StyledTextButton onPress={closeModal} text={`${poisLength} ${t('showPois')}`} />
```
steffenkleinle, line 130: You should adjust our translations accordingly instead of using a template literal, i.e. adjusting the `showPois` translation such that you can pass a count and depending on that get the right singular/plural text. The reason for this is mainly that e.g. the word order might differ in other languages. As an example you can have a look at the docs: https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals

native/src/components/__tests__/PoiFiltersModal.spec.tsx
```diff
@@ -14,16 +14,20 @@ describe('PoiFiltersModal', () => {
   beforeEach(jest.clearAllMocks)
 
   const pois = new PoiModelBuilder(2).build()
+  // const poisLength = pois.length
+
```
steffenkleinle, line 18:
```suggestion
```

web/src/components/PoiFilters.tsx
```diff
@@ -58,6 +58,7 @@ const TileRow = styled.div<{ itemCount: number }>`
 const StyledButton = styled(TextButton)`
   width: 100%;
   margin: 0;
+  /* margin: 0; background-color: ${props => props.theme.colors.textDisabledColor}; */
```
steffenkleinle, line 61:
```suggestion
```

web/src/components/PoiFilters.tsx
```diff
@@ -121,7 +124,11 @@ const PoiFilters = ({
             ))}
           </TileRow>
         </Section>
-        <StyledButton onClick={closeModal} text={t('showPois')} />
+        {poisLength !== 0 ? (
+          <StyledButton onClick={closeModal} text={`${poisLength} ${t('showPois')} `} />
+        ) : (
+          <StyledButton onClick={closeModal} text={t('noPois')} disabled />
+        )}
```
steffenkleinle, line 131: Same as for native, pls use plurals of i18next for this. Then you can just do
```suggestion
        <StyledButton onClick={closeModal} text={t('showPois', { count: poisLength })} disabled={poisLength === 0} />
```

native/src/components/PoiFiltersModal.tsx
steffenkleinle: Also, I think we should disable the button the same way as on web if there are no pois matching the filters.

steffenkleinle (commented):
Looking good already! All the tests are passing for me on native, what exactly is the problem?

relentless95:
Although the changes seem good now, it still doesn't pass the test when running on circleci. As mentioned before last week i modified my version of the test so that it works.

relentless95:
> Although the changes seem good now, it still doesn't pass the test when running on circleci. As mentioned before last week i modified my version of the test so that it works.

I also have no idea why the commit workflow check failed.

steffenkleinle (changes requested):
Thanks for applying some of the changes. Whats still missing in my opinion is correctly handling singular and plural. At the moment the message for exactly one poi is wrong since it will still use the plural (i.e. `1 Show locations`). It should use singular though. To see how to do that, please have another look at the documentation I linked:
https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals

 Also, the translations are not really correct in some languages other than German, e.g. for English the correct translation would be:
* `Show 1 location`
* `Show x locations`

Therefore, you should delete all translations where you don't know if it is correct or not.

relentless95:
> Thanks for applying some of the changes. Whats still missing in my opinion is correctly handling singular and plural. At the moment the message for exactly one poi is wrong since it will still use the plural (i.e. `1 Show locations`). It should use singular though. To see how to do that, please have another look at the documentation I linked: https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals
> 
> Also, the translations are not really correct in some languages other than German, e.g. for English the correct translation would be:
> 
> * `No places`/`No places found`
> * `Show 1 place`
> * `Show x places`
> 
> Therefore, you should delete all translations where you don't know if it is correct or not.

Thank you for pointing that out again. Also, do you have any idea why the check in circleci fails. is It related to this same issue with plurals?

steffenkleinle:
> Thank you for pointing that out again. Also, do you have any idea why the check in circleci fails. is It related to this same issue with plurals?

They fail because you changed the used translation to `noPois` but still search for `0 showPois`.

relentless95:
> > Thanks for applying some of the changes. Whats still missing in my opinion is correctly handling singular and plural. At the moment the message for exactly one poi is wrong since it will still use the plural (i.e. `1 Show locations`). It should use singular though. To see how to do that, please have another look at the documentation I linked: https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals
> > Also, the translations are not really correct in some languages other than German, e.g. for English the correct translation would be:
> > 
> > * `No places`/`No places found`
> > * `Show 1 place`
> > * `Show x places`
> > 
> > Therefore, you should delete all translations where you don't know if it is correct or not.
> 
> Thank you for pointing that out again. Also, do you have any idea why the check in circleci fails. is It related to this same issue with plurals?

thank you

translations/translations.json
```diff
@@ -6718,6 +6718,7 @@
       "onlyCurrentlyOpen": "nur aktuell geöffnete Orte anzeigen",
       "poiCategories": "Kategorien",
       "showPois": "{{numberOfPois}} Orte anzeigen",
+      "showPois_one": "{{numberOfPois}} Ort anzeigen",
```
steffenkleinle, line 6721: You should use something like this:
```suggestion
      "showPois_zero": "0 Orte gefunden",
      "showPois_one": "1 Ort anzeigen",
      "showPois_two": "{{count}} Orte anzeigen",
      "showPois_few": "{{count}} Orte anzeigen",
      "showPois_many": "{{count}} Orte anzeigen",
      "showPois_other": "{{count}} Orte anzeigen",
```

Then you can just use
``` ts
t('showPois', { count: poisCount })
```
Handling for zero pois is already handled then so you can furthermore simplify the rendering probably :)

translations/translations.json
```diff
@@ -6768,7 +6769,7 @@
       "adjustFilters": "አካባቢዎችን ያጥሩ",
       "onlyCurrentlyOpen": "በአሁኑ ሰአት ክፍት የሆኑ ቦታዎችን ብቻ ያሳዩ",
       "poiCategories": "ምድቦች",
-      "showPois": "{{numberOfPois}} ቦታዎችን ያሳዩ",
+      "showPois": " ቦታዎችን ያሳዩ",
```
steffenkleinle, line 6772: Please delete all of those translations if you don't know whether they are correct or not. 
```suggestion
```
Exceptions are probably only german and english and perhaps if you can speak other languages.

translations/translations.json
relentless95: Alright, I’ll use that instead.

translations/translations.json
relentless95: I did’t add the field showPois to each language as it was already there. I just added a placeholder before every instance of showPois. I deleted those changes in this commit.

steffenkleinle: Yes, but you change the german version of it which is the base translation. Therefore all translations in other languages are not correct anymore. If they are not deleted, they won't be fixed.

relentless95: okay, i will just revert the translation file to its version before i made any change to it.

relentless95: I will do a commit and if it's not what you meant, feel free to modify it.\n I thought the data in translations.json and deleting some fields would break the app. Guess that's not the case.

steffenkleinle: It's not, if some fields are not there, it just falls back to german.

native/src/components/PoiFiltersModal.tsx
```diff
@@ -125,7 +127,11 @@ const PoiFiltersModal = ({
           </TileRow>
         </Section>
         <Section>
-          <StyledTextButton onPress={closeModal} text={t('showPois')} />
+          {poisCount !== 0 ? (
+            <StyledTextButton onPress={closeModal} text={t('showPois', { count: poisCount })} />
+          ) : (
+            <StyledTextButton onPress={closeModal} text={t('noPois')} disabled />
+          )}
```
steffenkleinle, line 134: The translation `showPois` you created now also handle the case of `poisCount === 0`, so no need to handle this separately here. Could you try to simplify this? Same for web :)

translations/translations.json
```diff
@@ -7035,6 +7034,7 @@
       "map": "Map",
       "showOnMap": "Show on map",
       "noPois": "No places found.",
+      "noPois_one": "No place found.",
```
steffenkleinle, line 7037:
```suggestion
```

translations/translations.json
```diff
@@ -7076,7 +7076,8 @@
       "adjustFilters": "Filter locations",
       "onlyCurrentlyOpen": "only show currently open locations",
       "poiCategories": "Categories",
-      "showPois": "Show locations",
+      "showPois": "Show {{count}} places",
+      "showPois_one" : "Show {{count}} place",
```
steffenkleinle, line 7080: Could you adjust those similar to the german translations, i.e. adding `_zero`, `_one`, `_two`, `_few`, `_many` and `_other` translations.

relentless95: Alright, i'm on it.

native/src/components/PoiFiltersModal.tsx
relentless95: alright

translations/translations.json
relentless95: I am reading about test with typescript to understand exactly why the check in circleci fails.

steffenkleinle: We can also do a call some time to help you if you want. At the moment it fails because of prettier (which formats our code, e.g. new lines, indentations, spaces...) in the translations.json file. Perhaps there is a trailing comma or similar.

relentless95: A call would be nice.

native/src/components/PoiFiltersModal.tsx
```diff
@@ -127,11 +127,7 @@ const PoiFiltersModal = ({
           </TileRow>
         </Section>
         <Section>
-          {poisCount !== 0 ? (
-            <StyledTextButton onPress={closeModal} text={t('showPois', { count: poisCount })} />
-          ) : (
-            <StyledTextButton onPress={closeModal} text={t('noPois')} disabled />
-          )}
+          <StyledTextButton onPress={closeModal} text={t('showPois', { count: poisCount })} />
```
steffenkleinle, line 130: Nice, now we are nearly there. The only thing missing is passing the `disabled` prop and deciding correctly when it should be disabled :) Great work so far :tada:

translations/translations.json
steffenkleinle: Just remove this and from my point of view you are good to go! Great work :)

steffenkleinle (approved):
Tested on android and web (firefox).
On web, it seems like the number is somehow not aligned vertically the same way as the rest of the text:
image
No idea why and how though and whether that is intended or not by the browsers.

Nice work!

steffenkleinle:
Please squash your commits when merging :)

relentless95:
> Please squash your commits when merging :)

Yes, I will do that when it's approved.

sarahsporck (approved):
Tested in Firefox and Android emulator. Works as expected. Nicely done!

PR #2612, opened 2023-12-08
1213: Embed video web

f1sh1918:
### Short description

As a user i want to embed video from supported sources
As i user i want to explicitly allow these source before getting displayed.
As a user i want to be able withdraw my decision about external sources

### Proposed changes

- hide iframe initially and change source to avoid any tracking (seems to be the only option to do by source)
- preserve iframe sources from content
- show opt-in checkbox
- show iframe with hint to consent page if opt-in accepted
- add consent page to manage external sources
- add cookies class and hook to manage the technical cookies

### Side effects
- hopefully none

### Testing ###
http://localhost:9000/testumgebung/de/video-test-verschiedene-sources

### Resolved issues

Fixes: #1213


web/src/components/RemoteContent.tsx
```diff
@@ -107,6 +146,12 @@ type RemoteContentProps = {
 }
 
 const HIJACK = new RegExp(buildConfig().internalLinksHijackPattern)
+const DOMPURIFY_TAG_IFRAME = 'iframe'
+const DOMPURIFY_ATTRIBUTE_FULLSCREEN = 'allowfullscreen'
+
+export type ExternalSourcePermission = Record<string, boolean>
```
f1sh1918, line 152: gonna move this to `api-client` also for native

web/src/routes/ConsentPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+type ConsentPageProps = { languageCode: string }
+const ConsentPage = ({ languageCode }: ConsentPageProps): ReactElement => {
+  const { t } = useTranslation('consent')
+  const { value: sources, updateCookie } = useCookie(EXTERNAL_SOURCES_COOKIE_NAME)
+  const externalSourcePermissions: ExternalSourcePermissions = useMemo(
+    () => (sources ? JSON.parse(sources) : {}),
+    [sources],
+  )
+
+  const onPress = (type: string, allowed: boolean) => {
```
LeandraH, line 30: Maybe we can give this a more telling name like `setNewCookies`, and `permissionGiven` instead of `allowed`. I got confused about why we're setting all cookies to allowed in the next line 😅

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,189 @@
...
+  contentIframeSources: IframeSources,
+): void => {
+  const updatedContentSources: IframeSources = contentIframeSources
+  updatedContentSources[index] = source
+  setExistingIframes(updatedContentSources)
+}
+
+export const hideIframe = (iframe: HTMLIFrameElement): void => {
+  iframe.setAttribute('src', IFRAME_BLANK_SOURCE)
+  iframe.setAttribute('style', 'display:none')
```
LeandraH, line 30: I don't love that we're only hiding the iframe instead of removing it.

LeandraH (approved):
Very cool, thank you! Tested in Firefox

web/src/utils/iframes.ts
f1sh1918: yes but you can't remove them, because you don't know where they were in the DOM.Thats why i replace the src and hide them.
Even by replacing the source, the iframe can't track anything or deliver data

LeandraH: Huh, we can't get it via the `parentNode` and `removeChild`?

web/src/components/RemoteContent.tsx
```diff
@@ -129,20 +202,69 @@ const RemoteContent = ({
...
       if (HIJACK.test(node.href)) {
         node.addEventListener('click', handleClick)
       }
     })
-  }, [html, handleClick, sandBoxRef])
+
+    const iframes = currentSandBoxRef.getElementsByTagName('iframe')
+    Array.from(iframes).forEach((iframe: HTMLIFrameElement, index: number) => {
+      if (iframe.src !== IFRAME_BLANK_SOURCE) {
+        preserveIFrameSourcesFromContent(index, iframe.src, setContentIframeSources, contentIframeSources)
```
steffenkleinle, line 228: Why do we need to save the sources separately except from the cookie? This looks like duplicated state. Or is there any reason why we need to map sources to the iframe index?

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,189 @@
...
+  checkbox.name = 'opt-in-checkbox'
+  checkbox.id = id
+  checkbox.onclick = onClickHandler
+  const label = document.createElement('label')
+  showSource(label, source)
+  label.htmlFor = id
+  label.appendChild(document.createTextNode(text))
+  container.appendChild(label)
+  container.appendChild(checkbox)
+}
```
steffenkleinle, line 129:
```suggestion
}

```

web/src/utils/iframes.ts
f1sh1918: no @LeandraH you remove the node with the iframe and its gone. We don't recreate the old DOM every time as we do in native. So we can not relocate where it was before

web/src/hooks/useCookie.ts
```diff
@@ -0,0 +1,33 @@
...
+import { useState, useCallback } from 'react'
+
+import Cookie from '../utils/Cookie'
+
+type UseCookieReturnType = {
+  value: string | null
+  updateCookie: (newValue: string, path: string, domain: string, expires?: DateTime) => void
+  deleteCookie: (path: string, domain: string) => void
+}
+const useCookie = (name: string): UseCookieReturnType => {
```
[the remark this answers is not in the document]
f1sh1918: i think you are right. we can set a default

web/src/utils/Cookie.ts
```diff
@@ -0,0 +1,30 @@
...
+  static set(
+    name: string,
+    value: string,
+    path: string,
+    domain: string,
+    expires = DateTime.now().plus({ year: 1 }),
+  ): void {
+    document.cookie = `${encodeURIComponent(name)}=${encodeURIComponent(
+      value,
+    )}; expires=${expires.toJSDate()}; path=${path}; domain=${domain}; secure;`
```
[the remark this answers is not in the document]
f1sh1918: not sure what you expect here. This is how you update a cookie.

web/src/utils/Cookie.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  static remove(name: string, path: string, domain: string): void {
+    Cookie.set(name, '', path, domain, DateTime.fromMillis(0))
```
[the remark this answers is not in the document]
f1sh1918: yes but thats how it works. There is no other way to remove a cookie. Its just about setting a expiration date in the past. You can check cookie javascript documentation

web/src/components/RemoteContent.tsx
```diff
@@ -115,6 +173,14 @@ const RemoteContent = ({
   smallText = false,
 }: RemoteContentProps): ReactElement => {
   const sandBoxRef = React.createRef<HTMLDivElement>()
+  const { value: sources, updateCookie } = useCookie('externalSources')
+  const externalSourcePermissions: ExternalSourcePermission = useMemo(
+    () => (sources ? JSON.parse(sources) : {}),
+    [sources],
+  )
```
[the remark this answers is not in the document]
f1sh1918: yes i basically return now an empty object from generic type to avoid this check

translations/translations.json
```diff
@@ -8410,24 +8410,26 @@
...
     },
     "en": {
-      "knownResourceOptIn": "I agree that content from this external source may be displayed to me: ",
-      "knownResourceContentMessage": "This element contains data from that external source. You can block the embedding of such content in settings.",
+      "knownResourceOptIn": "I agree that content from this external source may be displayed to me.",
+      "knownResourceContentMessage": "This element contains data from an external source. You can block the embedding of such content in settings.",
       "knownResourceBlocked": "External content from this source has been blocked. You can allow this source in settings."
     }
   },
   "consent": {
```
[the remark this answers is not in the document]
f1sh1918: yes you are right. External source are a part of consent. There may be added other consent related issue in the future. So i moved everything into `consent` namespace

web/src/routes/ConsentPage.tsx
f1sh1918: i renamed to `updateSourcePermission`and `permissionGiven`

web/src/components/ConsentSection.tsx
```diff
@@ -0,0 +1,40 @@
...
+  height: 1px;
+  border: none;
+  margin: 0;
+`
+
+type ConsentSectionProps = {
+  title: string
+  description: string
+  allowed: boolean
+  onPress: (type: string, value: boolean) => void
```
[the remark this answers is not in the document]
f1sh1918: adjusted implementation

web/src/hooks/useCookie.ts
```diff
@@ -0,0 +1,33 @@
+import { DateTime } from 'luxon'
+import { useState, useCallback } from 'react'
+
+import Cookie from '../utils/Cookie'
+
+type UseCookieReturnType = {
+  value: string | null
+  updateCookie: (newValue: string, path: string, domain: string, expires?: DateTime) => void
+  deleteCookie: (path: string, domain: string) => void
+}
```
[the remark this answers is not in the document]
f1sh1918: i converted the hook into a generic one but kept the state as it is just changed return types and avoided null return

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export const hideIframe = (iframe: HTMLIFrameElement): void => {
+  iframe.setAttribute('src', IFRAME_BLANK_SOURCE)
+  iframe.setAttribute('style', 'display:none')
+}
+
+export const restoreIframe = (iframe: HTMLIFrameElement, source: string): void => {
+  iframe.setAttribute('src', source)
+  iframe.setAttribute('style', 'display:block')
+}
```
steffenkleinle, line 36: Optional and maybe not a good practice :smile::
Could we perhaps "save" the src in some of the iframes attributes, e.g. the id to avoid having to save it together with some index and update logic in some unrelated react state. At least I found the code very confusing/hard to read. I tried the following and it would work afaics:
```suggestion
export const hideIframe = (iframe: HTMLIFrameElement): void => {
  const source = iframe.getAttribute('src')
  if (source !== null && source !== IFRAME_BLANK_SOURCE) {
    iframe.setAttribute('id', source)
    iframe.setAttribute('src', IFRAME_BLANK_SOURCE)
    iframe.setAttribute('style', 'display:none')
  }
}

export const restoreIframe = (iframe: HTMLIFrameElement): void => {
  const source = iframe.getAttribute('id')
  if (source) {
    iframe.setAttribute('src', source)
    iframe.setAttribute('style', 'display:block')
  }
}
```
Just an idea though, so feel free to just ignore :)

Edit: Actually, it seems like you are already using `iframe-container${supportedSource}${iframeIndex}` as id for the container, so maybe that can easily be combined?

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+  checkbox.onclick = onClickHandler
+  const label = document.createElement('label')
+  showSource(label, source)
+  label.htmlFor = id
+  label.appendChild(document.createTextNode(text))
+  container.appendChild(label)
+  container.appendChild(checkbox)
+}
+
+const showSettingsButton = (element: HTMLDivElement, t: TFunction): void => {
```
steffenkleinle, line 131:
```suggestion
const showSettingsLink = (element: HTMLDivElement, t: TFunction): void => {
```

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+  supportedSource: string,
+  viewportSmall: boolean,
+  deviceWidth: number,
+): void => {
+  const permission = supportedSource ? externalSourcePermissions[supportedSource] : undefined
+  const iframeId = `iframe-container${supportedSource}${iframeIndex}`
+  const iframeContainer = getIframeContainer(iframeId, viewportSmall, iframe, deviceWidth)
+
+  if (permission === undefined) {
+    const message = `${t('consent:knownResourceOptIn')}`
```
steffenkleinle, line 176:
```suggestion
    const message = t('consent:knownResourceOptIn')
```

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+    showOptIn(message, iframeContainer, supportedSource, onUpdateCookie, iframeIndex)
+  } else if (permission && storedIframeSource) {
+    restoreIframe(iframe, storedIframeSource)
+    // Add do not track parameter (only working for vimeo)
+    if (supportedSource === 'vimeo.com') {
+      addDoNotTrackParameter(iframe)
+    }
+    const message = `${t('consent:knownResourceContentMessage')}`
```
steffenkleinle, line 184:
```suggestion
    const message = t('consent:knownResourceContentMessage')
```

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+  storedIframeSource: string,
+  t: TFunction,
+  onUpdateCookie: (source: string) => void,
+  iframeIndex: number,
+  supportedSource: string,
+  viewportSmall: boolean,
+  deviceWidth: number,
+): void => {
+  const permission = supportedSource ? externalSourcePermissions[supportedSource] : undefined
+  const iframeId = `iframe-container${supportedSource}${iframeIndex}`
```
steffenkleinle, line 172:
```suggestion
  const iframeContainerId = `iframe-container${supportedSource}${iframeIndex}`
```

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+
+const showSource = (element: HTMLElement, source: string): void => {
+  const span = document.createElement('span')
+  span.classList.add('iframe-source')
+  element.appendChild(span)
+  span.appendChild(document.createTextNode(source))
+  element.appendChild(document.createElement('br'))
+}
+
+export const showMessage = (text: string, element: HTMLDivElement, iframeSource?: string): void => {
```
steffenkleinle, line 91: `iframeSource` is never undefined, so you can just make it required and remove the check below.

web/src/utils/iframes.ts
```diff
@@ -0,0 +1,190 @@
...
+  deviceWidth: number,
+): void => {
+  const permission = supportedSource ? externalSourcePermissions[supportedSource] : undefined
+  const iframeId = `iframe-container${supportedSource}${iframeIndex}`
+  const iframeContainer = getIframeContainer(iframeId, viewportSmall, iframe, deviceWidth)
+
+  if (permission === undefined) {
+    const message = `${t('consent:knownResourceOptIn')}`
+    showOptIn(message, iframeContainer, supportedSource, onUpdateCookie, iframeIndex)
+  } else if (permission && storedIframeSource) {
```
steffenkleinle, line 178: `storedIframeSource` is never undefined/null
```suggestion
  } else if (permission) {
```

web/src/hooks/useCookie.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import Cookie from '../utils/Cookie'
+
+type UseCookieReturnType<T> = {
+  value: T
+  updateCookie: (newValue: T, expires?: DateTime, domain?: string, path?: string) => void
+  deleteCookie: (path?: string, domain?: string) => void
+}
+const useCookie = <T>(name: string): UseCookieReturnType<T> => {
+  const [value, setValue] = useState<string | null>(Cookie.get(name))
```
steffenkleinle, line 12: Optional: You could also save the value directly to avoid a lot of JSON.parse/JSON.stringify:
```suggestion
  const [value, setValue] = useState<T | null>(JSON.parse(Cookie.get(name) ?? 'null'))
```
Warning: See some other comments, this would break rerendering if you are continuing mutation instead of spread syntax :)

steffenkleinle (approved):
Really nice work!
For me the code was a little hard to understand in a few places, thats why I tried thinking of a few possible improvements. However, none of those are required, everything is working just fine :+1: So if you feel like the current is good enough and you don't have the patience to check my comments, just merge it :rocket:

web/src/utils/iframes.ts
f1sh1918: i don't feel like its getting cleaner with this solution tbh

web/src/utils/Cookie.ts
```diff
@@ -0,0 +1,30 @@
+import { DateTime } from 'luxon'
+
+class Cookie {
```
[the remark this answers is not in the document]
f1sh1918: you are absolutely right. I just checked the implementation on `tagesschau.de` which uses cookies and didn't think about it. But i don't see any reason in this case why not to use `localStorage`
Maybe @svenseeberg ?

steffenkleinle (approved):
Tested on firefox.

Issue #2607, opened 2023-12-06
Missing purpose strings in Info.plist


f1sh1918:
### Describe the Bug

We receive mails from apple that there are missing purpose string

### Steps to Reproduce

1. Make a release (don't do manually :) )
2. Check mail with issues (attached screenshot)

### Expected Behavior

The purpose strings can be found

### Actual Behavior

apple can't find some purpose string

### Additional Information
Please add all needed purpose strings also in the `Info.plist` even they are localized.

<img width="807" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/37902063/0ee8fd36-a3f2-449b-89eb-9f62f426303d">

PR #2616, opened 2023-12-12
2607: Missing purpose strings 

f1sh1918:
### Short description
These purpose strings are just fallbacks for avoiding messages from apple of missing purpose strings. They shouldn't be used in the app but the localized purpose strings
Please switch your ios device language to something thats not english and test if the correct localized strings will still be used

### Proposed changes

- add fallbacks to the info.plist in english

### Resolved issues

Fixes: #2607 

---


LeandraH (approved):
Looks good, can you fix the indentations though?

f1sh1918:
unfortunately not because i don't have them in the IDE.

steffenkleinle (approved):

PR #2620, opened 2023-12-13
2316: Malte help form native

steffenkleinle:
### Short description

Add malte help form for native.

Testing:
- malte help form: https://webnext.integreat.app/augsburg/de/hilfebutton-test (opening the malte help form via the offers tab is NOT supported as it will soon be removed)
- sprungbrett: https://webnext.integreat.app/testumgebung/de/arbeit/ausbildung (you have to replace `sprungbrettOffer.path` with https://web.integreat-app.de/proxy/sprungbrett/app-search-internships?location=86150 in SprungbrettOffer since for the testumgebung plz no jobs are found) 

### Proposed changes

- Use the old input design instead of the new proposed material design
- Add react-hook-forms for form handling
- Add new route malte contact form which is only activated if the offer exists
- Add new components FormInput and FormRadioButtons

### Side effects

- Submitting the form is not yet implemented

### Resolved issues

Fixes: #2316.

---


sarahsporck (commented):
Could you extend the useEmbeddedOffers hook, as well as the EmbeddedOffer component to display the help form on category pages?

native/src/routes/MalteHelpFormOffer.tsx
```diff
@@ -0,0 +1,145 @@
...
+            { key: 'any', label: t('any') },
+            { key: 'female', label: t('female') },
+            { key: 'male', label: t('male') },
+          ]}
+        />
+      </View>
+
+      <StyledHorizontalLine />
+
+      {/* TODO Extract to shared constant */}
```
sarahsporck, line 129: Should be possible now 👍🏼

steffenkleinle:
> Could you extend the useEmbeddedOffers hook, as well as the EmbeddedOffer component to display the help form on category pages?

@sarahsporck actually not exactly sure how to do that as we are not loading any data for the malte help button. How did you imagine this?

steffenkleinle:
> > Could you extend the useEmbeddedOffers hook, as well as the EmbeddedOffer component to display the help form on category pages?
> 
> @sarahsporck actually not exactly sure how to do that as we are not loading any data for the malte help button. How did you imagine this? Should I just return a mock value or `true` or something for that offer?

Actually, atm I don't get why we need the hook at all and don't just do the loading in the separate offers which are rendered conditionally in `EmbeddedOffers` :see_no_evil:

LeandraH (commented):
When writing a comment longer than 200 characters on iOS, the submit button is simply disabled but without information about why. I think there should be a comment about that. Maybe we just make the problematic field red?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,6 +1,6 @@
 import OfferModel from '../models/OfferModel'
 
-export const MAX_COMMENT_LENGTH = 200
+export const MALTE_HELP_FORM_MAX_COMMENT_LENGTH = 200
```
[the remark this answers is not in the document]
sarahsporck: This number was implicated by the design @hauf-toni made.
But I think I also remember that the comment field was initially not requested by malteser. We added it to help adding some context to the help request. This could be the reason why the length is so short. Maybe we can increase it to 400 characters?

web/src/routes/OffersPage.tsx
```diff
@@ -39,15 +38,6 @@ const OffersPage = ({ city, cityCode, languageCode }: CityRouteProps): ReactElem
...
         }
-        if (offer.alias === MALTE_HELP_FORM_OFFER_ROUTE) {
-          path = pathnameFromRouteInformation({ route: MALTE_HELP_FORM_OFFER_ROUTE, cityCode, languageCode })
-          // ignore zammad url (postData) for the tile model
-          return new TileModel({
-            title: t(offer.title),
-            path,
-            thumbnail: offer.thumbnail,
-          })
-        }
```
sarahsporck, line 50: I think this should be removed when we actually remove the offers tab. Otherwise the helpform appears in the offers tab but cannot be clicked on.

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -0,0 +1,141 @@
...
+      onSubmit()
+      showSnackbar({ text: t('submitSuccessful') })
+    } catch (e) {
+      showSnackbar({ text: t('error:unknownError') })
+    }
+  })
+
+  return (
+    <Container>
+      <Caption title={t('title')} />
```
sarahsporck, line 86: In web I don't show the title for page based offers, as it breaks the "flow". Do you prefer to keep it?

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -0,0 +1,141 @@
...
+`
+
+const InformationRow = styled.View`
+  flex-direction: row;
+  gap: 8px;
+`
+
+const InformationText = styled.Text`
+  flex: 1;
+`
```
sarahsporck, line 31: I think this should have around the same size as the content font size.

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <InformationRow>
+        <Icon Icon={SecurityIcon} />
+        <InformationText>{t('securityNote')}</InformationText>
+      </InformationRow>
+
+      <FormInput name='name' title={t('name')} control={control} rules={{ required: true }} />
+      <FormInput name='roomNumber' title={t('roomNumber')} control={control} showOptional />
+
+      <View>
+        <InputTitle>{t('howToBeContacted')}</InputTitle>
```
sarahsporck, line 101: These Subtitles could also be larger and could profit from a little bit more padding.

native/src/components/base/InputSection.tsx
```diff
@@ -30,50 +31,65 @@ const Title = styled(ThemedText)`
   text-align: left;
 `
 
-const Input = styled.TextInput<{ numberOfLines: number }>`
+const Input = styled.TextInput<{ numberOfLines: number; invalid: boolean }>`
   border-width: 1px;
-  border-color: ${props => props.theme.colors.themeColor};
-  text-align-vertical: top;
+  border-color: ${props => (props.invalid ? props.theme.colors.invalidInput : props.theme.colors.themeColor)};
```
sarahsporck, line 36: So in malte the default color is dead read? Yikes! But nothing to do now I guess...

native/src/components/SprungbrettOffer.tsx
```diff
@@ -0,0 +1,40 @@
...
+import openExternalUrl from '../utils/openExternalUrl'
+import List from './List'
+import SprungbrettListItem from './SprungbrettListItem'
+
+type SprungbrettOfferProps = {
+  sprungbrettOffer: OfferModel
+  languageCode: string
+}
+
+const SprungbrettOffer = ({ sprungbrettOffer, languageCode }: SprungbrettOfferProps): ReactElement => {
```
sarahsporck, line 21: so this code is identical to `routes/SprungbrettOffer.tsx` besides the useLoadFromEndpoint and the error handler? To me it seems this should be the route and the other one the component. 
But also probably never mind as we'll remove one of those files in the next few weeks anyway...

web/src/routes/OffersPage.tsx
steffenkleinle: I will just filter it out, good point :+1:

native/src/components/MalteHelpFormOffer.tsx
steffenkleinle: No, I will remove it :+1:

native/src/components/base/InputSection.tsx
steffenkleinle: Good point, I guess I will just make the borders grey (text decoration color).

native/src/components/SprungbrettOffer.tsx
steffenkleinle: Yes, routes/SprungbrettOffer.tsx will be removed in #2637. I disagree, since there will not be a standalone offers/sprungbrett offer route anymore, offers will only ever be shown embedded in a category. Therefore, it should be a component and not a route/container. This also makes sense since all meta tags/language change paths and so on should be set by the category page already.

native/src/components/base/InputSection.tsx
steffenkleinle: Will also affect integreat incl. the feedback component then though.

api-client/src/endpoints/submitHelpForm.ts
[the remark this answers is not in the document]
steffenkleinle: Increased to 400

steffenkleinle:
> When writing a comment longer than 200 characters on iOS, the submit button is simply disabled but without information about why. I think there should be a comment about that. Maybe we just make the problematic field red?

It doesn't show the (max. 400 chars) hint?
Anyway, I fixed validation of the fields (the boxes are now highlighted in red) as well as setting maxLength on the comment field such that it is not possible to enter more than 400 characters in the first place.

LeandraH (approved):
Very cool!

native/src/components/base/InputSection.tsx
sarahsporck: In web it is already black. So I don't think this should be an issue ;)

sarahsporck (approved):
Tested on Android Emulator again. Looks very nice!

Issue #2313, opened 2023-08-14
Page based offer


sarahsporck:
**Is your feature request related to a problem? Please describe.**
In the future we want to move offers into our page based content to avoid parallel structures and get rid of the "Offers" menu in the app.

**Describe the solution you'd like**
When fetching a page, there is an additional field `offer` which contains on single item of the offers endpoint. The offer should probably be positioned below the actual page endpoint.

**Describe alternatives you've considered**
N/A

**Additional context**
- https://github.com/digitalfabrik/integreat-cms/issues/2360

steffenkleinle:
@sarahsporck do you think we still need a design for this?

sarahsporck:
Hmm. Probably yes. The question is where and how the offer is displayed. Does it directly flow into the content? If it is below the content, as described in the text, do we just pin it below? Do we still use the tile here? What about Offers like sprungbrett? To me these are mostly design/usability questions. But we could also spend some time and discuss this in the team.

nikolahoff:
[Here](https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?type=design&node-id=1936%3A81&mode=design&t=PlawIftw1YKKESb0-1) ist the link to the figma file.

PR #2622, opened 2023-12-14
2313: Page Based Offer

sarahsporck:
### Short description

This PR implements page based offers. The design can be found here -> https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?type=design&node-id=2007-4825&mode=design&t=JpeoYWfqTRHYDOKB-0

In the future it should be possible to render jobs of different job providers in one list e.g. IHK & Sprungbrett. For that multiple offers are then returned from the backend, which is the reason `embedded_offers` is an array. However currently we only support rendering one embedded offer in the frontend and ignore the remainders.

### Proposed changes

- Web: introduce var embedded. Only render city content layout when embedded  === false
- Native: introduce a hook which loads the embedded offer if a category has an offer.
    - @steffenkleinle it was not possible to use `useLoadExtraCityContent` with the cityContentData as param for the load function, as this resulted in a rerender cycle. This could be because `data` is not always "reset" on every render and therefore it is a different object. Not sure though. Feel free to investigate, if you don't like the solution.
- Also: small adjustment to list view design

### Resolved issues

Fixes: #2313 

### How to test?

In loadSprungbrettJobs change `cityCode` to e.g. "augsburg".
Then visit http://localhost:9000/testumgebung/de/arbeit/sprungbrett or http://localhost:9000/testumgebung/de/arbeit/ausbildung

---


native/src/components/ListItem.tsx
```diff
@@ -55,7 +57,7 @@ type ListItemProps = {
 const ListItem = ({ language, title, thumbnail, children, Icon, navigateTo }: ListItemProps): ReactElement => (
   <StyledPressable onPress={navigateTo}>
     <ListItemView language={language}>
-      {!!thumbnail && <Thumbnail source={thumbnail} />}
+      {!!thumbnail && (React.isValidElement<SvgProps>(thumbnail) ? thumbnail : <Thumbnail source={thumbnail} />)}
```
ztefanie, line 60: Should this check really be here? Can't this be moved to `SimpleImage` component?

ztefanie (commented):
Code looks nice :+1: 

I could not test this. When changing "cityCode" to "Augsburg", i get: "Diese Seite wurde leider nicht gefunden."

LeandraH:
> I could not test this. When changing "cityCode" to "Augsburg", i get: "Diese Seite wurde leider nicht gefunden."

Ran into this as well, it needs to be "augsburg".

LeandraH (commented):
Works nicely in Firefox but I can't load it in emulated Android

[Uploading Screen_recording_20240108_110038.webm…]()

api-client/src/types.ts
```diff
@@ -167,3 +168,5 @@ export type TimeSlot = {
   end: string
   start: string
 }
+
+export type EmbeddedOffer = 'sprungbrett' | 'help'
```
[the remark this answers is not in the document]
sarahsporck: ahh. this should be added in the zammad contact form pr. will remove it here :)

native/src/components/ListItem.tsx
sarahsporck: It can. But I don't think that's quite right. The SimpleImage component also manages styling, resizing, ... If we pass a react element this cannot be influenced anymore, as the thumbnail is already rendered. I'd prefer keeping it as it is. But I'll move the check to a utility punction to make this clearer. You're alright with this?

native/src/hooks/useEmbeddedOffer.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type EmbeddedOfferParams = {
+  cityContentResponse: Omit<CityContentReturn, 'data'>
+  category?: CategoryModel | null
+  cityCode: string
+  languageCode: string
+}
+
+type EmbeddedOfferReturn = {
+  extra: Awaited<LoadSprungbrettJobReturn> | null
```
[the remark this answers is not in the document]
sarahsporck: Extra refers to the extra/additional content returned from the city. It is already used here
https://github.com/digitalfabrik/integreat-app/blob/dd1de8bc4252e9898fa058b76d0fa4f0932bbad9/native/src/hooks/useLoadExtraCityContent.ts#L25-L28

sarahsporck:
> Works nicely in Firefox but I can't load it in emulated Android

hmm. for me it works on the emulator. did you delete and rebuild the app? :)

native/src/hooks/useEmbeddedOffer.ts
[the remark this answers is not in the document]
LeandraH: Hmmm, so we use it there to give us an opportunity to add anything we want, am I understanding that correctly? But that's not what we have here, right?

LeandraH:
> hmm. for me it works on the emulator. did you delete and rebuild the app? :)

You're right, works now. I've never had to delete the app before, in what cases do we need to do that?

api-client/src/endpoints/createOffersEndpoint.ts
```diff
@@ -5,11 +5,7 @@ import { JsonOfferPostType, JsonOfferType } from '../types'
 
 export const OFFERS_ENDPOINT_NAME = 'offers'
 
-const createPostMap = (jsonPost: JsonOfferPostType): Map<string, string> => {
-  const map = new Map()
-  Object.keys(jsonPost).forEach(key => map.set(key, jsonPost[key]))
-  return map
-}
+export const createPostMap = (jsonPost: JsonOfferPostType): Map<string, string> => new Map(Object.entries(jsonPost))
```
sarahsporck, line 8: This should basically do the same.

native/src/routes/CategoriesContainer.tsx
```diff
@@ -56,11 +57,22 @@ const CategoriesContainer = ({ navigation, route }: CategoriesContainerProps): R
...
     data?.categories && !category && previousLanguageCode === languageCode ? ErrorCode.PageNotFound : response.error
 
+  const refresh = useCallback(() => {
+    response.refresh()
+    embeddedOffers.refresh()
+  }, [response, embeddedOffers])
+
   return (
-    <LoadingErrorHandler {...response} error={error} scrollView>
+    <LoadingErrorHandler refresh={refresh} loading={response.loading} error={error} scrollView>
```
sarahsporck, line 75: I changed this part again, so that the page content is already displayed, while the offer might still be loading. In my opinion this provides a much better user experience.

web/src/routes/SprungbrettOfferPage.tsx
```diff
@@ -64,19 +73,27 @@ const SprungbrettOfferPage = ({ city, cityCode, languageCode }: CityRouteProps):
     ),
   }
 
+  const Wrapper = !embedded
```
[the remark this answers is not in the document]
sarahsporck: nope. this can then be removed. Therefore, we don't have to bear with this ugly condition for long :)

native/src/components/EmbeddedOffer.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const EmbeddedOffer = ({ extra, languageCode }: EmbeddedOfferProps): ReactElement | null => {
+  if (extra && extra.sprungbrettJobs) {
+    return <SprungbrettOffer jobs={extra.sprungbrettJobs} language={languageCode} />
+  }
+  return null
+}
+
+export default EmbeddedOffer
```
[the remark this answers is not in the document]
sarahsporck: I changed the types so that there is a property for each embedded offer. Is that better?

native/src/hooks/useEmbeddedOffer.ts
```diff
@@ -0,0 +1,65 @@
...
+  cityContentResponse,
+}: EmbeddedOfferParams): EmbeddedOfferReturn => {
+  const load = useCallback(async () => {
+    if (category) {
+      const embeddedOfferReturn = await Promise.all(
+        category.embeddedOffers.flatMap(offer => {
+          if (offer.alias === SPRUNGBRETT_OFFER) {
+            return loadSprungbrettJobs({ cityCode, languageCode, baseUrl: determineApiUrl })
+          }
+          return []
```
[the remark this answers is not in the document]
sarahsporck: For now we will look at the embeddedOffers in a fixed order. First 'sprungbrett' then 'hilfebutton', etc. ...
At some point however we want to support multiple offers, e.g. For the 'jobbörsen' to serve a combined list of jobs from multiple provideres. I think for now it is fine that we only look at some embedded offers.

native/src/routes/CategoriesContainer.tsx
```diff
@@ -56,11 +57,20 @@ const CategoriesContainer = ({ navigation, route }: CategoriesContainerProps): R
...
+  const { extra, ...combinedResponse } = useEmbeddedOffer({
+    category,
+    cityCode,
+    languageCode,
+    cityContentResponse: response,
+  })
+
   const error =
-    data?.categories && !category && previousLanguageCode === languageCode ? ErrorCode.PageNotFound : response.error
+    data?.categories && !category && previousLanguageCode === languageCode
```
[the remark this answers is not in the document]
sarahsporck: I refactored this, so that the content can be loaded, even though the embedded offer does not load. Seems to be better ux in my opinion.

native/src/components/SprungbrettListItem.tsx
```diff
@@ -3,6 +3,7 @@ import { Text } from 'react-native'
 
 import { SprungbrettJobModel } from 'api-client'
 
+import { SprunbrettIcon } from '../assets'
```
LeandraH, line 6:
```suggestion
import { SprungbrettIcon } from '../assets'
```

LeandraH (approved):

f1sh1918:
Code looks good and its working fine. Tested on ios and chrome. Nicely done!

I just think there are some styling issues within the `ListItem`, which are optional since it was before but i mention it here because the design is not 100% fitting.

**ListItem**
`height:100%` which should be removed
In the design i couldn't measure the exact padding but it should be at least `12px` and set on the outer container instead the `Description`. (f.e. event list item look weird because there is different padding for the thumbnail)
image

f1sh1918 (approved):

native/src/hooks/useEmbeddedOffer.ts
```diff
@@ -0,0 +1,65 @@
...
+        }),
+      )
+      return Object.assign({}, ...embeddedOfferReturn)
```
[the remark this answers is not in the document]
steffenkleinle: Ah, thanks for the explanation, missed that!

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+} from 'api-client'
+
+import { determineApiUrl } from '../utils/helpers'
+
+type EmbeddedOffers = {
+  sprungbrett?: Awaited<LoadSprungbrettJobReturn>
+}
+
+type EmbeddedOffersParams = {
+  category?: CategoryModel | null
```
steffenkleinle, line 18: Imo there is no reason to connect this to a category, in theory also some other models could have embedded offers. Additionally, you could just pass an empty array as default then to avoid some of the checks below.
```suggestion
  embeddedOffers: OfferModel[]
```

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type EmbeddedOffersReturn = {
+  embeddedOffers: EmbeddedOffers | null
+  error: Error | null
+  loading: boolean
+  refresh: () => void
+}
+
+const useEmbeddedOffers = ({ category, cityCode, languageCode }: EmbeddedOffersParams): EmbeddedOffersReturn => {
```
steffenkleinle, line 30: Optional: So far our naming that fetches stuff from the cms is including `load`.
```suggestion
const useLoadEmbeddedOffers = ({ category, cityCode, languageCode }: EmbeddedOffersParams): EmbeddedOffersReturn => {
```

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+  cityCode: string
+  languageCode: string
+}
+
+export type EmbeddedOffersReturn = {
+  embeddedOffers: EmbeddedOffers | null
+  error: Error | null
+  loading: boolean
+  refresh: () => void
+}
```
steffenkleinle, line 28: Optional: Is there a reason to use custom return value with `embeddedOffers` instead of `data` property? I think this would simplify things.
```suggestion
export type EmbeddedOffersReturn = ReturnType<EmbeddedOffers>
```

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+            return {
+              sprungbrett: await loadSprungbrettJobs({ cityCode, languageCode, baseUrl: determineApiUrl }),
+            }
+          }
+          return {}
+        }),
+      )
+      return Object.assign({}, ...embeddedOfferReturn) as EmbeddedOffers
+    }
+    return {} as EmbeddedOffers
```
steffenkleinle, line 45: Typecasts are not necessary for me:
```suggestion
      return Object.assign({}, ...embeddedOfferReturn)
    }
    return {}
```

web/src/components/CategoriesContent.tsx
```diff
@@ -49,7 +57,17 @@ const CategoriesContent = ({ categories, categoryModel, t, ...props }: Categorie
...
+        Footer={
+          categoryModel.embeddedOffers[0] && (
+            <EmbeddedOffer
+              offer={categoryModel.embeddedOffers[0]}
+              city={city}
+              pathname={pathname}
+              cityCode={cityCode}
+              languageCode={languageCode}
+            />
+          )
```
steffenkleinle, line 69: This is how it is supposed to be in the end, right?
```suggestion
          categoryModel.embeddedOffers.map(embeddedOffer => (
            <EmbeddedOffer
              offer={embeddedOffer}
              city={city}
              pathname={pathname}
              cityCode={cityCode}
              languageCode={languageCode}
            />
          ))
```

native/src/components/SprungbrettListItem.tsx
```diff
@@ -18,7 +18,7 @@ class SprungbrettListItem extends PureComponent<SprungbrettListItemProps> {
     const { language, job, openJobInBrowser } = this.props
     return (
       <ListItem
-        thumbnail={<SprunbrettIcon width={25} height={25} />}
+        thumbnail={<SprungbrettIcon width={24} height={24} />}
```
steffenkleinle, line 21: Sorry, might have been not that clear here:
Our custom icon component should be used which automatically sets the icon size to the default of 24px.
```suggestion
        thumbnail={<Icon Icon={SprungbrettIcon} />}
```

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+  LoadSprungbrettJobReturn,
+  SPRUNGBRETT_OFFER,
+  loadSprungbrettJobs,
+  useLoadAsync,
+} from 'api-client'
+
+import { determineApiUrl } from '../utils/helpers'
+
+type EmbeddedOffers = {
+  sprungbrett?: Awaited<LoadSprungbrettJobReturn>
```
steffenkleinle, line 14: Optional: I think this type would be easier to read if you would change `LoadSprungbrettJobReturn` to just the object without the promise and do this here:
```suggestion
  sprungbrett?: LoadSprungbrettJobReturn
```

native/src/components/Categories.tsx
```diff
@@ -82,21 +85,23 @@ const Categories = ({
...
-              resourceCache={resourceCache}
-              language={language}
-              onItemPress={navigateToCategory}
-            />
-          )}
-          // Fixes VirtualizedLists nesting error
-          scrollEnabled={false}
-        />
+        <>
+          <EmbeddedOffer embeddedOffers={embeddedOffers} languageCode={language} />
```
steffenkleinle, line 89: I think the offers should be rendered below the list. Otherwise the categories are only after an possibly very long list.

native/src/hooks/useEmbeddedOffers.ts
```diff
@@ -0,0 +1,56 @@
...
+        category.embeddedOffers.map(async offer => {
+          if (offer.alias === SPRUNGBRETT_OFFER) {
+            return {
+              sprungbrett: await loadSprungbrettJobs({ cityCode, languageCode, baseUrl: determineApiUrl }),
+            }
+          }
+          return {}
+        }),
+      )
+      return Object.assign({}, ...embeddedOfferReturn) as EmbeddedOffers
```
steffenkleinle, line 43: Using a reduce instead of Object.assign would preserve the typechecking (Object.assign always returns any):
```suggestion
      return embeddedOfferReturn.reduce((it, acc) => ({ ...acc, ...it }), {})
```

web/src/components/CategoriesContent.tsx
sarahsporck: Probably not. It's not quite clear how we handle multiple embeddedOffers yet. The idea would be to have on component which is responsible for multiple offers then. There could be multiple ways to implement this. E.g. The cms could handle requests to external (job offer) providers and provides an endpoint, which we call. Or we implement this in the frontend. Multiple embedded offers are not yet a thing, I fear.

steffenkleinle: Okay, works for now. Thanks for explaining :)

native/src/components/Categories.tsx
sarahsporck: Yes, that's true. But I think design wise the offer belongs to the content above. Maybe we should disallow embedded offers in second level pages...

native/src/hooks/useEmbeddedOffers.ts
sarahsporck: This does not work with the dependencies. This rerenders this hook over and over again, as embeddedOffers is not memoized, but considered a different value than category. Hard to explain, but this just doesn't work...

PR #2623, opened 2023-12-22
Upgrade Headroom to v2

michael-markl:
### Short description

Upgrade ReactStickyHeadroom dependency to v2

### Side effects

ReactStickyHeadroom powers the header of the web app. No changes should be visible.


steffenkleinle (approved):
Tested on firefox, seems to work as expected. Thanks!

ztefanie (approved):
Thanks for updating :+1: 
Tested on chrome, works as expected.

Issue #2569, opened 2023-11-20
Test `CategoriesToolbar.tsx`


sarahsporck:
**Is your feature request related to a problem? Please describe.**
We currently don't have any good starter tickets for web left. Therefore, to allow newcomers to get to know the project and get familiar with react, we can use tests for that. The CategoriesToolbar is not covered in our code base entirely. 

**Describe the solution you'd like**
- add a test file `CategoriesToolbar.spec.tsx` in the `web/src/components/__tests__/` directory
- write a test that checks whether the correct pdf url is used in the case
    - a) the category is a root category (it is visible in the dashbord page)
    - b) the category is not a root category
- consider the remaining code as a black box which you don't know about

Use the other tests as helper to reuse existing patterns and get to know [@testing-library/react](https://testing-library.com/docs/react-testing-library/api). Also checkout [jest](https://jestjs.io/), to get to know the test framework.

Njivaniaina2407:
@sarahsporck  I just commited `CategoriesToolbar.spec.tsx` file. Is it necessary to make a pull request for this?

sarahsporck:
Jep. We always make pull requests for all changes :)

PR #2625, opened 2024-01-09
2569: Test CategoriesToolbar.tsx

Njivaniaina2407:
### Short description

Create a CategoriesToolbar.spec.tsx file and   write a test that checks whether the correct PDF URL is used in case:
- Category is a root category
- Category is not a root category

### Proposed changes

Creating the CategoriesToolbar.spec.tsx file and writing the test 

### Resolved issues

Fixes: #2569

---


ztefanie (changes requested):

steffenkleinle (approved):
Looking good! :tada:

ztefanie (approved):
Well done :muscle: :rocket:

web/src/components/__tests__/CategoriesToolbar.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import { CategoryModel } from 'api-client'
```
steffenkleinle, line 1: We renamed/moved the whole api-client, so you have to change this to
```suggestion
import { CategoryModel } from 'shared/api'
```

Issue #2573, opened 2023-11-20
Poi list bottom action sheet title not scrolling


steffenkleinle:
### Describe the Bug

The bottom action sheet title does not scroll with the rest of the bottom action sheet content. This looks weird and blocks rare space.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/testumgebung/de/locations on a mobile viewport (<750px)
2. Scroll in the bottom action sheet
3. See the title (`In deiner Nähe`) not scrolling

### Expected Behavior

The title (`In deiner Nähe`) scrolls the same as the poi list items. Preferably the styling also looks like the title of a poi detail.  

### Actual Behavior

The title is fixed and does not scroll with a different styling than the title of a poi detail.

### Additional Information

None.

PR #2629, opened 2024-01-15
2573: Poi list bottom action sheet title not scrolling

Njivaniaina2407:
### Short description

The bottom action sheet title does not scroll with the rest of the bottom action sheet content. This looks weird and blocks rare space. 
The title (In deiner Nähe) should scroll the same as the poi list items. Preferably the styling also looks like the title of a poi detail.

### Proposed changes

Put the title in the scrollable field

### Resolved issues

Fixes: #2573

---


web/src/components/PoisMobile.tsx
```diff
@@ -168,7 +173,7 @@ const PoisMobile = ({
         }
       />
       <BottomActionSheet
-        title={!canGoBack ? t('listTitle') : undefined}
+        title={!canGoBack ? '' : undefined}
```
LeandraH, line 176: This now chooses between two ways that don't show a title, so we can simply remove the entire title.

web/src/components/PoisMobile.tsx
```diff
@@ -180,6 +185,7 @@ const PoisMobile = ({
           </GoBackContainer>
         )}
         <ListContainer>
+          {!canGoBack ? <ListTitle>{t('listTitle')}</ListTitle> : undefined}
```
LeandraH, line 188: You can use the logical AND here as a shorthand like this: `{!canGoBack && <ListTitle>{t('listTitle')}</ListTitle>}`. But it's also fine like this, your call.

LeandraH (commented):
Very nice, thank you! Just a little extra clean-up :)
Tested in Firefox

steffenkleinle (approved):
Tested on firefox, well done! :tada:

LeandraH (approved):
Awesome, thank you!

PR #2630, opened 2024-01-15
1127: Add events and POIs to search results

LeandraH:
### Short description

The search results now also include events and ~offers~ POIs.

### Proposed changes

- Added [Minisearch](https://github.com/lucaong/minisearch) to do the searching (and sorting!) for us. It's a pretty fast, small library that searches on the frontend.
- Combined categories, events, and POIs into a big array which is passed to Minisearch for indexing
- Stuck with very simple design for the offers and the POIs

### Side effects

- Edited getCachedThumbnail to take in a path instead of a category
-

### Resolved issues

Fixes: #1127 

---


native/src/routes/SearchModal.tsx
```diff
@@ -37,51 +42,91 @@ const Wrapper = styled.View`
...
+      })
+      openExternalUrl(link, showSnackbar)
+    } else {
+      const isEvent = link.split('/')[3] === EVENTS_ROUTE
+      const routeInformation: CategoriesRouteInformationType | EventsRouteInformationType = {
+        route: isEvent ? EVENTS_ROUTE : CATEGORIES_ROUTE,
+        cityContentPath: link,
+        cityCode,
+        languageCode,
+      }
```
sarahsporck, line 106: check out https://github.com/digitalfabrik/integreat-app/blob/5dca2f60c5aa23d011145506a451f379ffcc26d5/native/src/hooks/useNavigateToDeepLink.ts#L63-L64 which does exactly what you are doing here

api-client/src/endpoints/hooks/useAllPossibleSearchResults.ts
```diff
@@ -0,0 +1,98 @@
...
+import useLoadFromEndpoint from 'api-client/src/endpoints/hooks/useLoadFromEndpoint'
+import loadSprungbrettJobs from 'api-client/src/endpoints/loadSprungbrettJobs'
+
+type useSearchParams = {
+  city: string
+  language: string
+  cmsApiBaseUrl: string
+}
+
+export type SearchResult = {
```
sarahsporck, line 15: I'm wondering if it makes sense to add the component (that will be rendered) to the search result. This could slightly speed up searching, as we are rendering all items initially anyways.

web/src/routes/SearchPage.tsx
```diff
@@ -99,20 +118,25 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
-          <SearchListItem
-            key={category.path}
-            query={query}
-            category={category}
-            contentWithoutHtml={contentWithoutHtml}
-          />
-        ))}
+        {(query.length !== 0 ? results : allPossibleResults).map(
+          ({ title, id, content, location, path, url, thumbnail }) => (
+            <SearchListItem
```
sarahsporck, line 123: The sames goes for this one. If we add the component to the minisearch search result this should be slightly faster.

sarahsporck (changes requested):
Very weird. There are different fonts used in the search result title in native. I think the highlighted font should at least be bold, so that it is obviously a highlight. Could fix this :)
Bildschirmfoto 2024-01-23 um 11 33 02

Besides that I'm very impressed by how smooth searching feels now. Definitely a good change.

sarahsporck:
Not sure whether I like searching in Offers though. Feels a little bit out of context. Other opinions?

native/src/routes/SearchModal.tsx
LeandraH: Amazing, thanks for the hint!

api-client/src/endpoints/hooks/useAllPossibleSearchResults.ts
LeandraH: It might also delay the start of the search because it would take longer to load everything in, and that is currently by far the bottleneck :(

sarahsporck: But everything is loaded anyway, when opening the search, right?

LeandraH:
> Not sure whether I like searching in Offers though. Feels a little bit out of context. Other opinions?

I see your point but I don't think people usually mind search results being too generous. But I can take it back out.

sarahsporck:
> > Not sure whether I like searching in Offers though. Feels a little bit out of context. Other opinions?
> 
> I see your point but I don't think people usually mind search results being too generous. But I can take it back out.

Maybe let's wait for another opinion. But just having "Fleischer/in xyz" as a search result doesn't say much. I think we should rather include local news or POIs instead (which can also be done in a separate pr)

api-client/src/endpoints/hooks/useAllPossibleSearchResults.ts
LeandraH: Yes, everything loads then, but an array of components is a lot bigger than an array of tiny objects, so it would take longer to add it to the Minisearch. And since that is already the moment when the most waiting happens (because we're waiting for the categories etc to load), I'd rather not make it even longer, even though it would marginally speed up the display once the search is actually underway. Which feels pretty fast to me already

native/src/routes/SearchModal.tsx
```diff
@@ -37,51 +39,76 @@ const Wrapper = styled.View`
...
-      cityCode,
-      languageCode,
-      cityContentPath: category.path,
-    })
+    search.addAll(allPossibleResults)
+    return search
+  }, [allPossibleResults])
+
+  // Minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
+  const searchResults = minisearch.search(query) as unknown as SearchResult[]
```
steffenkleinle, line 82: Couldn't we move this to the api-client?

native/src/routes/SearchModal.tsx
```diff
@@ -37,51 +39,76 @@ const Wrapper = styled.View`
...
+      sendTrackingSignal({
+        signal: {
+          name: SEARCH_FINISHED_SIGNAL_NAME,
+          query,
+          url: urlFromRouteInformation(routeInformation),
+        },
+      })
+      navigateTo(routeInformation)
+    }
   }
```
steffenkleinle, line 112: You can just use `useNavigateToLink` instead which should do exactly what you need

native/src/routes/SearchModal.tsx
```diff
@@ -112,7 +151,7 @@ const SearchModal = ({
       <SearchHeader theme={theme} query={query} closeSearchBar={onClose} onSearchChanged={setQuery} t={t} />
       <KeyboardAvoidingView behavior={Platform.OS === 'ios' ? 'padding' : undefined} style={{ flex: 1 }}>
         <List
-          items={searchResults}
+          items={query.length > 0 ? searchResults : allPossibleResults}
```
steffenkleinle, line 154: Is this necessary? Doesn't minisearch return all items if the query is empty?

native/src/components/SearchListItem.tsx
```diff
@@ -40,29 +40,37 @@ const HighlighterCategoryTitle = styled(Highlighter)<{ language: string }>`
...
 `
 
 type SearchListItemProps = {
-  category: CategoryModel
-  contentWithoutHtml: string
-  resourceCache: PageResourceCacheStateType
-  onItemPress: (category: CategoryModel) => void
+  title: string
+  contentWithoutHtml?: string
+  resourceCache?: PageResourceCacheStateType
```
steffenkleinle, line 49: You pass a default value for both so it should never be undefined

web/src/components/SearchListItem.tsx
```diff
@@ -56,12 +56,30 @@ const StyledLink = styled(Link)`
...
+      path: string
+      thumbnail?: string
+      url?: never
+    }
+  | {
+      url: string
+      path?: never
+      thumbnail?: never
+    }
+)
```
steffenkleinle, line 73: This is duplicated, can we use the exported type from `useAllPossibleSearchResults` instead?

native/src/routes/SearchModal.tsx
steffenkleinle, line 82:
```suggestion
  // Minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
I think the typing is correct, for me the return type is already `SearchResult[]` (and you are also using it without typecast in web).
```suggestion
  const searchResults = minisearch.search(query)
```
```

File: api-client/src/endpoints/hooks/useAllPossibleSearchResults.ts [hunk 9]
```diff
@@ -0,0 +1,98 @@
...
+      path: string
+      location?: string
+      url?: string
+    }
+  | {
+      location: string
+      url: string
+      content?: string
+      path?: string
+    }
```
steffenkleinle, line 31: I think we could just simplify this since we always use `path ?? url` and `location ?? content`, so we could also just do and store location in content and url in path.
```suggestion
      content: string
      path: string
    }
```

native/src/routes/SearchModal.tsx
```diff
@@ -37,51 +39,76 @@ const Wrapper = styled.View`
...
-  categories,
+  allPossibleResults,
   navigateTo,
   theme,
   languageCode,
   cityCode,
   closeModal,
   t,
   initialSearchText = '',
+  loading,
```
steffenkleinle, line 62: `loading` should be handled in the container

native/src/routes/SearchModalContainer.tsx
```diff
@@ -23,6 +25,39 @@ const SearchModalContainer = ({ navigation, route }: SearchModalContainerProps):
...
+        id: event.path,
+      })) ?? []),
+      ...(offerData?.extra.sprungbrettJobs.map(offer => ({
+        title: offer.title,
+        location: offer.location,
+        url: offer.url,
+        id: offer.url,
+      })) ?? []),
+    ],
+    [data?.categories, data?.events, offerData?.extra.sprungbrettJobs],
```
steffenkleinle, line 59: Can we move this to a common helper function in the api-client?

web/src/routes/SearchPage.tsx
```diff
@@ -64,31 +76,38 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+        />
         <LoadingSpinner />
       </CityContentLayout>
     )
   }
 
-  if (!searchResults) {
-    const error = categoriesError || new Error('Categories should not be null!')
+  if (error) {
+    reportError(error)
```
steffenkleinle, line 103: Errors should not be reported here, otherwise we report an error on every render. In general we should have this already covered.
```suggestion
```

steffenkleinle (changes requested):

native/src/routes/SearchModal.tsx
steffenkleinle: Rather use `useNavigateToLink` which does not include additional deeplinking logic and should be sufficient.

api-client/src/endpoints/hooks/useAllPossibleSearchResults.ts
sarahsporck: Fine. Then, let's not do this. :)

LeandraH:
It turns out, I misread the original issue, it requested adding locations to the search, not offers. Fixed that, will now work on the other review points.

native/src/routes/SearchModal.tsx
LeandraH: No, it unfortunately returns an empty array then.

web/src/components/SearchListItem.tsx
LeandraH: It was only this complicated because of the offers, removing the offers fixed this.

native/src/routes/SearchModal.tsx
LeandraH: Ah, yes, it returns the type `SearchResult` provided by `Minisearch` which is different from the one I defined (it doesn't have `title`, `content`, or `path`). When I remove the typecasting, the `List` yells at me because the function `renderItem` takes in our internal `SearchResults` as arguments, with title etc.

I am honestly somewhat confused about not having to typecast it in web but am happy to not need to?

steffenkleinle (changes requested):
Nearly there, well done! :tada:

native/src/routes/SearchModal.tsx
steffenkleinle: Weird stuff, seems like some weird naming problem... no idea about it tbh...

web/src/routes/SearchPage.tsx
```diff
@@ -33,14 +33,18 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+  } = useAllPossibleSearchResults({
     city: cityCode,
     language: languageCode,
+    cmsApiBaseUrl,
   })
-  const searchResults = useMemo(() => (categories ? searchCategories(categories, query) : null), [categories, query])
+
+  const minisearch = useMiniSearch(allPossibleResults)
+
+  const results = minisearch.search(query)
```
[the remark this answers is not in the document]
sarahsporck: See [here](https://github.com/lucaong/minisearch/blob/72cdfbbe866de860e53de6b37db4534ac8bd1e52/src/MiniSearch.ts#L1237). The conditional does make sense.

e2e-tests/web/test/specs/navigateToSearchResults.e2e.ts
```diff
@@ -24,7 +24,9 @@ describe('navigateToSearchResult', () => {
 
     await expect(results[0]).toExist()
     await expect(results[1]).toExist()
-    expect(await results[1]!.getAttribute('href')).toBe(`/${Routes.dashboard}/language/language-courses`)
+    expect(await results[1]!.getAttribute('href')).toBe(
+      `/${Routes.dashboard}/language/language-courses/online-language-courses`,
+    )
```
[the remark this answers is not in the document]
sarahsporck: I think because minisearch has a [different ordering](https://github.com/lucaong/minisearch/blob/72cdfbbe866de860e53de6b37db4534ac8bd1e52/src/MiniSearch.ts#L1270) for search results. It sorts the result by "score", which will change the order of the search result in some cases.

sarahsporck (approved):
Tested in native and web. Works like charm <3

e2e-tests/web/test/specs/navigateToSearchResults.e2e.ts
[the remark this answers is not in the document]
LeandraH: Exactly :)

shared/hooks/useMiniSearch.ts
```diff
@@ -0,0 +1,30 @@
...
+  thumbnail?: string
+  content: string
+  path: string
+}
+
+const useMiniSearch = (allPossibleResults: SearchResult[]): MiniSearch<SearchResult> => {
+  const minisearch = useMemo(() => {
+    const search = new MiniSearch({
+      fields: ['title', 'content'],
+      storeFields: ['title', 'content', 'path', 'location', 'url', 'thumbnail'],
```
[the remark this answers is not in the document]
LeandraH: `id` is always returned but it's a good point, it's a little confusing to not have it. I'll add it.

shared/hooks/useMiniSearch.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type SearchResult = {
+  title: string
+  id: string | number
+  thumbnail?: string
+  content: string
+  path: string
+}
+
+const useMiniSearch = (allPossibleResults: SearchResult[]): MiniSearch<SearchResult> => {
```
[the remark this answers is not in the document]
LeandraH: Hah, true, good catch, thanks!

shared/utils/formatSearchResults.ts
```diff
@@ -0,0 +1,35 @@
+import { CategoriesMapModel, EventModel, PoiModel } from '../api'
+import { SearchResult } from '../hooks/useMiniSearch'
+
+const formatSearchResults = (
```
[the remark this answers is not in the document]
LeandraH: That's a really good point, thanks!

shared/hooks/useMiniSearch.ts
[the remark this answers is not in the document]
LeandraH: Actually, removed the field `id` because that was only added in the formatting and we can use any field as `id` in Minisearch.

web/src/routes/SearchPage.tsx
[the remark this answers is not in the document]
LeandraH: Yup, conditional is important but I did move it out of the return statement :)

web/src/hooks/useAllPossibleSearchResults.ts
```diff
@@ -0,0 +1,53 @@
...
+import { SearchResult, formatSearchResults } from 'shared'
+import { createCategoriesEndpoint, createEventsEndpoint, createPOIsEndpoint, useLoadFromEndpoint } from 'shared/api'
+
+type useSearchParams = {
+  city: string
+  language: string
+  cmsApiBaseUrl: string
+}
+
+const useAllPossibleSearchResults = ({
```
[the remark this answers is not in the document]
LeandraH: https://github.com/digitalfabrik/integreat-app/issues/2691

native/src/routes/SearchModal.tsx
```diff
@@ -59,29 +47,10 @@ const SearchModal = ({
...
-    })
-    navigateTo({
-      route: CATEGORIES_ROUTE,
-      cityCode,
-      languageCode,
-      cityContentPath: category.path,
-    })
-  }
+  // Minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
+  const searchResults = minisearch.search(query) as unknown as SearchResult[]
```
steffenkleinle, line 53: Okay, I investigated this a little more. On web this just does not throw an error because we are not requiring the type to be (our) SearchResult. Instead, we are just destructuring it straight away and the return value of `.search()` includes `    [key: string]: any;`, therefore all of our fields are just of type `any`.

Since that (as well as just typecasting) is a little unsafe as we won't be notified when this is fixed or we change used fields) I would suggest the following: Include the actual `search` process in the hook and use as follows:
``` ts
const useSearch = (allPossibleResults: SearchResult[], query: string): SearchResult[] => {
  const minisearch: MiniSearch<SearchResult> = useMemo(() => {
    ...
  }, [allPossibleResults])

  // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
  return query.length === 0 ? allPossibleResults : minisearch.search(query)
}
```

The cool thing about that is that how we are doing the search is just an implementation detail, we only have to handle it once and `ts-expect-error` notifies us if this is fixed by the library.

Optional: Rename the hook to `useSearch` as using minisearch is just an implementation detail.

Then you could do this here and remove the conditional further below:
```suggestion
  const searchResults = useSearch(allPossibleResults, query)
```

steffenkleinle (commented):
I really really like the performance on web! Its sooo.... responsive xD
On native I am a little struggling though, I can't scroll and e.g. for munich it takes quite a while til the search opens... Can somebody else reproduce?

native/src/routes/SearchModal.tsx
LeandraH: Very cool, thanks for looking further into it, I like it!

LeandraH:
> On native I am a little struggling though, I can't scroll and e.g. for munich it takes quite a while til the search opens... Can somebody else reproduce?

Yes, managed to reproduce and fix it 🥳

Issue #2615, opened 2023-12-12
Set content direction (`ltr`|`rtl`) in theme


sarahsporck:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated, when I have to pass properties which are not important for the functionality to a component. This includes unnecessary uiDirection prop and the language prop.

**Describe the solution you'd like**
Currently we pass the direction to styled-component props via react properties. However, this could clearly be simplified by setting the content direction on the theme context instead. Then we could always access the current content direction ('ltr' or 'rtl') in the styled components.

I think it also makes sense to enable [this stylelint rule](https://github.com/csstools/stylelint-use-logical) to avoid forgetting about rtl & ltr.

**Describe alternatives you've considered**
Not changing a running system, I guess.

PR #2631, opened 2024-01-16
2615: Set content direction in theme

LeandraH:
### Short description

This sets the content direction in the theme so we don't need to pass it around in props anymore.

I'll also create a second ticket for the stylelint rule Sarah suggested.

### Resolved issues

Fixes: #2615 

---


sarahsporck (approved):
Wow! This PR feel nice to review. Nicely done. Tested in firefox & safari.

steffenkleinle (commented):
Nice, love it :heart_eyes: Can we do that for native as well? Or is that not possible for some reason?

LeandraH:
> Nice, love it 😍 Can we do that for native as well? Or is that not possible for some reason?

We could but we don't pass around the direction in native via props since we already have the helper functions in `contentDirection.ts`. So I think it's not really necessary.

steffenkleinle:
> > Nice, love it 😍 Can we do that for native as well? Or is that not possible for some reason?
> 
> We could but we don't pass around the direction in native via props since we already have the helper functions in `contentDirection.ts`. So I think it's not really necessary.

Hmm, while its true that it is easier to use than on web I think its still not as easy as having a prop in the theme since you still have to pass the language and use the helper function every time you want to access the content direction. The helper is still used 20+ times.

So I personally think it would be nicer to have that theme prop but you are completely right, it was much more important on web.

steffenkleinle (approved):

LeandraH:
I think I'll leave it. It's already quite inconsistent between web and native because we don't just use the direction in native, we use all kinds of helper functions based on the direction. So even if we put all those helpers into the theme, it would be an inconsistent experience between native and web. I'd say we see if it becomes a problem in the future.

PR #2632, opened 2024-01-16
2615: Linting for better rtl/ltr compatibility

LeandraH:
### Short description

I added [this stylelint rule](https://github.com/csstools/stylelint-use-logical) to remind us in the future to not use properties like margin-left because that might not have the required effect with a rtl language. And then I replaced a loooooot of properties.

### Side effects

- I updated the `yarn lint` command in the base folder to also run `stylelint` in web which it probably should have been doing already.

### Resolved issues

Fixes: #2615 

---


sarahsporck:
Maybe you could change the base branch to #2631 to simplify reviewing.

sarahsporck (commented):
So you mostly changed 
- `right` -> `inset-inline-end`
- `top` -> `inset-block-start`
- `bottom` -> `block-end`
- `left` -> `inline-start`
- `width` -> `inline-size`
- `height` -> `block-size`

Does the linting rule check these?

LeandraH:
> Maybe you could change the base branch to #2631 to simplify reviewing.

Arg, yes, sorry, I had meant to do that 😅

steffenkleinle (commented):
Nice find with the rule. I am not 100% sure if we really want to also replace all usages of top/bottom/width/height. Personally I find those easier to read/faster to write than their logical counter parts (might be just because I am used to it though). Furthermore, since we don't have this rule for native, in one place we would probably still use top/bottom/width/height/... while we won't on web which is a little confusing imo.

Other opinions? I am fine with the current solution as well.

steffenkleinle:
> > Nice find with the rule. I am not 100% sure if we really want to also replace all usages of top/bottom/width/height. Personally I find those easier to read/faster to write than their logical counter parts (might be just because I am used to it though). Furthermore, since we don't have this rule for native, in one place we would probably still use top/bottom/width/height/... while we won't on web which is a little confusing imo. Or am I missing something and there are actual reasons why we need to change the vertical properties as well?
> > Other opinions? I am fine with the current solution as well.
> 
> I think as long as stylelint tells me to change the property I'm perfectly fine with this. Also it seems this is the best way to handle multiple text directions 🤷🏼‍♀️ . Also, I'm unsure how to disable the rule for only some properties.

Disabling for some properties would be easy: https://github.com/csstools/stylelint-use-logical#except
So I think its just our preference what we want :)

LeandraH:
That's a good point, it's really particularly for ltr / rtl that we want this, right? I'll take a look

LeandraH:
Alright, now I mainly replaced:

- left: inset-inline-start
- right: inset-inline-end
- padding/margin/etc-left: -inline-start
- padding/margin/etc-right: -inline-end

web/src/components/PoisDesktop.tsx
```diff
@@ -26,7 +26,6 @@ const PanelContainer = styled.article`
   display: flex;
   flex-direction: column;
   width: ${dimensions.poiDesktopPanelWidth}px;
-  min-width: ${dimensions.poiDesktopPanelWidth}px;
```
steffenkleinle, line 29: Was this removed on purpose?

steffenkleinle (approved):
I like it, thanks!

web/src/components/PoisDesktop.tsx
LeandraH: Yes, we already have the width for exactly the size right above it so it's unnecessary.

steffenkleinle: Sometimes I still had problems with this as a `width` alone was not enough and the width was less than what was set. I think this was one of the cases.

steffenkleinle (changes requested):
Please readd min-width

steffenkleinle (approved):
Not tested again

f1sh1918:
@LeandraH just one question for the understanding. How does `padding-top` and `padding-bottom` behave different on rtl/ltr. For the left and right paddings its obvious.

sarahsporck (approved):
Tested in firefox. Works fine :)

LeandraH:
> @LeandraH just one question for the understanding. How does `padding-top` and `padding-bottom` behave different on rtl/ltr. For the left and right paddings its obvious.

It shouldn't, that's why we decided to remove it again. I did miss it in one file and removed that as well, there shouldn't be any top and bottom replacements anymore

Issue #2318, opened 2023-08-14
Contact form is submitted to Zammad


sarahsporck:
**Is your feature request related to a problem? Please describe.**
The malteser want an ability to interact with their users in form of contact forms. There needs to be a way to submit the contents of a form to Zammad

**Describe the solution you'd like**
The app has to request a token from Zammad via the [form config](https://helpdesk.integreat-app.de/api/v1/form_config) endpoint (e.g. [example](https://community.zammad.org/t/submit-form-from-custom-ui/7497)). The form has to be submitted as multipart/form-data (or maybe even as JSON, we need to test this) which includes the token to [form_submit](https://helpdesk.integreat-app.de/api/v1/form_submit)

Before submitting, the app needs to render all fields into a single formatted string. The current language should also be included into the string.

**Describe alternatives you've considered**
N/A

**Additional context**
- https://github.com/digitalfabrik/integreat-cms/issues/2361
- #2316

PR #2633, opened 2024-01-17
2318: Contact form is submitted to zammad

sarahsporck:
### Short description

This PR addresses submitting the malte form to Zammad. It can be tested here http://localhost:9000/testumgebung/de/offers/help
And an example ticket can be found here -> https://zammad-malte.tuerantuer.org/#ticket/zoom/15.
You should be able to create a customer account if you set your email in the contact form (check your mailbox). 

### Proposed changes

- create the body which is submitted to zammad.
- submit the form to zammad

### Side effects

> [!Note]
> I struggled a bit with correctly submitting the form and redirecting afterwards. As the submit function is async but a form submit runs synchronously. This was fixed by adding `event.preventDefault()` on the submit. Not sure if this has any implications... Feel free to enlighten me.

### Resolved issues

Fixes: #2318 

---


api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
...
+        title: `Hilfebutton - ${cityCode}`,
+        fingerprint,
+        token: config.token,
+      }),
+    })
+    return response.json()
+  }
   // eslint-disable-next-line no-console
-  Promise.resolve(console.log('submit'))
+  console.log(await submitToZammad())
```
LeandraH, line 139: I assume that log can go?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
...
+
+Beschreibung des Anliegens:
+${comment}
+`
+}
+
+const generateEmail = () => {
+  const RANDOM_NUMBER_LENGTH = 8
+  const array = new Uint8Array(RANDOM_NUMBER_LENGTH)
+  return `support+zammad-ticket-${crypto.getRandomValues(array).join('')}@integreat-app.de`
```
LeandraH, line 49: Random question, what's the story with the email address? Is that just for filtering?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
 import OfferModel from '../models/OfferModel'
```
steffenkleinle, line 1: I would stick to one naming if possible, so renaming this to `submitMalteHelpForm`

LeandraH (approved):
Looks good, tested in Firefox. I assume there's a second ticket still coming for native because I can't open the route there 😁 

Edit: just realised that there is a whole second PR for that, sorry!

api-client/src/endpoints/submitHelpForm.ts
steffenkleinle, line 139: Should this console.log stay?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        fingerprint,
+      }),
+    })
+    return response.json()
+  }
```
steffenkleinle, line 103: Can you extract this to a separate function outside of `submitHelpForm`?

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
...
+
+  const submitToZammad = async () => {
```
steffenkleinle, line 105: Is the wrapping in a function necessary?

api-client/src/endpoints/submitHelpForm.ts
sarahsporck: for the `form_submit` endpoint an email adress is required. In case the user doesn't provide an email address we generate a random one. It is not possible to pass anything as email address, as well. 
See here:
https://github.com/zammad/zammad/blob/de0574f09d7f6da7685abf53f6de3ca81c7a3a62/app/controllers/form_controller.rb#L49-L64

api-client/src/endpoints/submitHelpForm.ts
```diff
@@ -1,15 +1,142 @@
...
+  roomNumber,
+  contactChannel,
+  additionalContactInformation,
+  contactGender,
+  comment,
+  translate,
+}: BuildHelpFormBodyParams): string => {
+  const anyGenderText = contactGender === 'any' ? translate('contactPersonAnyGender') : undefined
+  const maleGenderText = contactGender === 'male' ? translate('contactPersonGenderMale') : undefined
+  const femaleGenderText = contactGender === 'female' ? translate('contactPersonGenderFemale') : undefined
```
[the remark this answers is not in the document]
sarahsporck: Hardcoding sounds like the way to go actually. I didn't want to pass even more arguments to this function.

api-client/src/endpoints/submitHelpForm.ts
LeandraH: Aaaah, cool, thanks for the explanation :)

sarahsporck:
I realized that we need to show a proper error message for invalid emails, as this is don by zammad. I don't see that we implement this check in the frontend, as they also check whether the mail actually exists, so let's keep the validation for zammad.

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -78,11 +85,25 @@ const MalteHelpFormOffer = ({
 
   const submit = handleSubmit(async _data => {
```
steffenkleinle, line 86: Please remove the underscore, was just used to avoid linting complaining about an unused var.

steffenkleinle (approved):
Tested on firefox and android :tada: Well done!

Issue #2635, opened 2024-01-17
Use correct thumbnails for pois


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
We have a todo in our code:
>   // TODO IGAPP-920: this has to be removed when we get proper images from CMS

folowed by the line
>  const thumbnail = poiFeature.thumbnail?.replace('-150x150', '') ?? PoiThumbnailPlaceholderLarge

See here: https://github.com/digitalfabrik/integreat-app/blob/04f954ce2e4669700df43690caf6fe43c5cd1b47/native/src/components/PoiDetails.tsx#L53
Have a look at that and what the right solution would be. The correct thumbnail/image sizes from the cms should be used.

**Describe the solution you'd like**

- [x] Remove the TODO and fix the usage if there is something to fix.
- [x] Check both web and native if everything is working as expected.
- [x] Furthermore, thumbnails are NOT required for pois in the cms, so it may be `null`. Currently, our types for this are wrong. This might lead to bugs. Therefore adjust the types and fix the usages. ~~Also talk to the cms team whether this is the case for categories/events as well. https://digitalfabrik.github.io/integreat-cms/api-docs.html#locations~~. Edit: According to the docs this is optional for categories, events and pois.

**Describe alternatives you've considered**
Just remove the TODO.

**Additional context**
The linked issue (#1881) was closed without additional information.

**Testing information**
Test all image related things in both web and native. This includes (among others):
- Navigation items (e.g. Local information, events, ...)
- Category tile icons
- Category, event and poi list icons
- Poi detail icon
- Event detail icon
- Images in the content

sarahsporck:
I think this is basically now waiting for this: https://github.com/digitalfabrik/integreat-app/issues/2277. Besides that @f1sh1918 should be able to give more context to this.

f1sh1918:
I think there should be enough context in the linked issues. There should be an Image resizer implemented and serve an array or object of image sources

steffenkleinle:
> I think there should be enough context in the linked issues. There should be an Image resizer implemented and serve an array or object of image sources

Will move this to #2277. This is an issue for at least categories and pois, for event it seems like it is already implemented.

PR #2636, opened 2024-01-17
Remove todos

steffenkleinle:
### Short description

Remove and fix TODOs in code.

### Proposed changes

- Fix the TODOs that are fixable
- Remove the TODOs that are linked in an issue
- Created #2635

### Side effects
None.

### Resolved issues
None.


native/src/utils/PushNotificationsManager.ts
```diff
@@ -108,7 +108,6 @@ export const useForegroundPushNotificationListener = ({
         if (mounted) {
           // The CMS needs some time until the push notification is available in the API response
           setTimeout(() => {
-            // TODO #2140: Uncomment and improve snackbar
```
LeandraH, line 111: Uh, the comment's still there though?

steffenkleinle: Yes, but it should be done in #2140 where this is described/mentioned. So I would just leave the comment.

ztefanie (approved):
Thanks for this cleanup :+1: 
Tested on Android, Pois work as expected, everything seems to be fine.

Issue #2598, opened 2023-11-27
User location doesn't work when a POI is selected


LeandraH:
### Describe the Bug

When a POI is selected and I tap to center on my location, the map moves in my direction a little but then centers back on the selected POI. This is both in iOS and Android

### Steps to Reproduce

1. Go to maps
2. Tap on a POI
3. Tap on the my-location icon
4. See the center of the map moving and then centering back on the selected POI.

### Expected Behavior

I would expect the map to always center on my location when I tap the button for that.

### Actual Behavior

The map only centers on my location if no POI is selected.

### Additional Information

https://github.com/digitalfabrik/integreat-app/assets/18513943/421e9bcd-e81e-4d9d-b88a-68ad0649b287

PR #2640, opened 2024-01-22
2598: Fix follow user location and decouple distance from map features (native)

steffenkleinle:
### Short description

This fixes the bug that the map always centers back on a selected map feature instead of keeping the manual set viewport or staying on the users location.
This was caused by continuous rerenders triggered by very frequent updates of the user location which made the whole Pois route rerender (and therefore trigger the useEffect responsible for centering on a newly selected map feature). To really fix this, sadly some refactoring was necessary which I split in web and native.

### Proposed changes

- Decouple pois and map features (and get rid of usages of GeoJsonPois)
- Remove distance from map features (as well as other properties not needed for displaying map features) and calculate when needed
- Only update distance when diff greater than threshold (0.1km)
- Extract and reuse filtering/multipoi/slug logic
- Some smaller refactoring to separate concerns a little better

### Side effects

Web will only work once the corresponding PR is merged (#2641).

### Testing

Please test this thoroughly. Especially please test selecting/deselecting multipois (and what is displayed in the bottom action sheet). Also check whether scroll positions are kept.

### Resolved issues

Fixes: #2598 (native).

---


native/src/components/PoiDetails.tsx
```diff
@@ -44,14 +43,14 @@ const StyledCategory = styled.Text`
...
+  distance: number | null
 }
 
-const PoiDetails = ({ poi, poiFeature, language }: PoiDetailsProps): ReactElement => {
+const PoiDetails = ({ poi, language, distance }: PoiDetailsProps): ReactElement => {
   const { t } = useTranslation('pois')
-  const thumbnail = poiFeature.thumbnail?.replace('-150x150', '') ?? PoiThumbnailPlaceholderLarge
-  const distance = poiFeature.distance
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const thumbnail = poi.thumbnail?.replace('-150x150', '') ?? PoiThumbnailPlaceholderLarge
```
steffenkleinle, line 53: Sadly our types here are wrong, `thumbnail` may very well be null (see https://digitalfabrik.github.io/integreat-cms/api-docs.html#locations).

LeandraH: Would it make sense to open a ticket to fix them?

steffenkleinle: Ahh sorry, included it in this issue: #2635.

api-client/src/utils/geoJson.ts
```diff
@@ -1,7 +1,9 @@
 import distance from '@turf/distance'
 
-import { featureLayerId, LocationType, PoiModel } from '../index'
-import { GeoJsonPoi, MapFeature, MapFeatureCollection } from '../maps'
+import { featureLayerId, LocationType, MapFeature, MapFeatureCollection } from '../maps'
+import PoiModel from '../models/PoiModel'
+
+export const MIN_DISTANCE_THRESHOLD = 0.1
```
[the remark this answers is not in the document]
steffenkleinle: Thats the precision that is atm displayed to the user. We could of course change this if we decide it should be more precise.

f1sh1918 (commented):
i'm facing a super strange issue on ios when filtering pois, just this particular category.
Can anyone test on ios if this is just a simulator issue
On android everything works fine

https://github.com/digitalfabrik/integreat-app/assets/37902063/d9af2e6e-0fa2-466d-b0c9-eccd2f995498

f1sh1918 (changes requested):
I'm facing a super strange rendering issue of `PoiListItem`
1. When filtering pois, just this particular category. Only happens if you didn't filter or select anything before.
2. Selecting a multipoi results in the same issue

Happens on ios (simulators and real device)
On android everything works fine

I played around with the style changes but couldn't find the issue so far.
On main everything is fine, so it is definitly related to this pr

https://github.com/digitalfabrik/integreat-app/assets/37902063/d9af2e6e-0fa2-466d-b0c9-eccd2f995498

https://github.com/digitalfabrik/integreat-app/assets/37902063/f40fae34-9bbe-4e5f-8f18-7271dcafc75c

native/src/components/PoiListItem.tsx
```diff
@@ -28,15 +28,14 @@ const Thumbnail = styled(SimpleImage)`
 
 const StyledPressable = styled(Pressable)<{ language: string }>`
   flex: 1;
+  height: 100%;
```
f1sh1918, line 31: this causes rendering issues on ios. please remove `flex: 1; height: 100%`

native/src/components/PoiListItem.tsx
```diff
@@ -28,15 +28,14 @@ const Thumbnail = styled(SimpleImage)`
...
   border-bottom-width: 1px;
   border-bottom-color: ${props => props.theme.colors.textDisabledColor};
   flex-direction: ${props => contentDirection(props.language)};
   padding: 24px 0;
 `
 
 const Description = styled.View`
-  flex: 1;
```
f1sh1918, line 38: readd `flex:1`

f1sh1918:
@steffenkleinle please adjust the styles for `PoiListItem` as following to solve the ios rendering issues

```js
const StyledPressable = styled(Pressable)<{ language: string }>`
  border-bottom-width: 1px;
  border-bottom-color: ${props => props.theme.colors.textDisabledColor};
  padding: 24px 0;
`

const Description = styled.View`
  flex: 1;
  flex-direction: column;
  padding: 0 32px;
  justify-content: center;
`
```
Can not find out why you added a height and so one. We just need a flex on Description

f1sh1918 (approved):

native/src/components/List.tsx
```diff
@@ -44,11 +44,7 @@ const List = <T,>({
     }
     onEndReached={onEndReached}
     showsVerticalScrollIndicator={false}
-    contentContainerStyle={{
-      flexGrow: 1,
-
-      paddingHorizontal: 10,
-    }}
+    contentContainerStyle={{ flex: 1 }}
```
ztefanie, line 47: @steffenkleinle Why have you changed this? This causes this bug: https://github.com/digitalfabrik/integreat-app/issues/2660

I reverted here: Feel free to comment if this breaks something you fixed.

steffenkleinle: Hmm, I think I had some layout problems, but @f1sh1918 suggestions already fixed them. So yes, feel free to revert.

Issue #2647, opened 2024-01-25
Wrong first recurrence for recurring event if close to date


steffenkleinle:
### Describe the Bug

If the next recurrence of an event is near, it is sometimes not displayed anymore.

### Steps to Reproduce

1. Create an recurring event that is happening in less than an hour (or mock the passed recurrence rule)
2. See the next recurrence not being displayed

### Expected Behavior

The next recurrence should be displayed until it ended.

### Actual Behavior

The next recurrence is not dispalyed anymore even before it ends.

### Additional Information

None.

PR #2648, opened 2024-01-25
2647: Fix next recurrence of events

steffenkleinle:
### Short description

Fix the displaying of the next recurrence of events if its near.

### Proposed changes

- Substract timezone offset difference from dates passed to rrule to account for usage of "local time" 
- Substract duration from dates passed to rrule to include currently happening events

### Side effects

None.

### Resolved issues

Fixes: #2647.

---

### DIFF
--- a/release-notes/unreleased/2647-next-recurrence.yml
+++ b/release-notes/unreleased/2647-next-recurrence.yml
@@ -1,0 +1,8 @@
+issue_key: 2467
+show_in_stores: true
+platforms:
+  - web
+  - ios
+  - android
+en: Fixed displaying of next date of an event if its near.
+de: Anzeige von baldigen Terminen behoben.
--- a/release-notes/unreleased/2647-next-recurrence.yml
+++ b/release-notes/unreleased/2647-next-recurrence.yml
@@ -4,5 +4,5 @@ platforms:
   - web
   - ios
   - android
-en: Fixed displaying of next date of an event if its near.
+en: Fixed displaying of next date of an event if it's soon.
 de: Anzeige von baldigen Terminen behoben.
--- a/shared/api/models/DateModel.ts
+++ b/shared/api/models/DateModel.ts
@@ -1,4 +1,4 @@
-import { DateTime } from 'luxon'
+import { DateTime, Duration } from 'luxon'
 import { RRule as RRuleType } from 'rrule'
 
 const MAX_RECURRENCE_YEARS = 5
@@ -11,6 +11,7 @@ class DateModel {
   _allDay: boolean
   _recurrenceRule: RRuleType | null
   _offset: number
+  _duration: Duration
 
   constructor({
     startDate,
@@ -26,12 +27,12 @@ class DateModel {
     this._recurrenceRule = recurrenceRule
     this._offset = startDate.offset
     this._allDay = allDay
-    const duration = endDate.diff(startDate)
+    this._duration = endDate.diff(startDate)
     // If there is a recurrence rule, the start and end dates are not updated in the CMS and are therefore most of the time outdated
     // Therefore calculate the (next) correct start and end date based on the recurrence rule if available
-    const first = recurrenceRule?.after(DateTime.now().toJSDate())
+    const first = recurrenceRule?.after(this.currentDateToRrule())
     this._startDate = first ? this.rruleToDateTime(first) : startDate
-    this._endDate = this._startDate.plus(duration)
+    this._endDate = this._startDate.plus(this._duration)
   }
 
   get startDate(): DateTime {
@@ -69,7 +70,7 @@ class DateModel {
     const duration = this._endDate.diff(this._startDate)
 
     return this.recurrenceRule
-      .between(now.toJSDate(), maxDate, true, (_, index) => index < count)
+      .between(this.currentDateToRrule(), maxDate, true, (_, index) => index < count)
       .map(it => this.rruleToDateTime(it))
       .map(
         it =>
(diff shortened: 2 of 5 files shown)


release-notes/unreleased/2647-next-recurrence.yml
```diff
@@ -0,0 +1,8 @@
+issue_key: 2467
+show_in_stores: true
+platforms:
+  - web
+  - ios
+  - android
+en: Fixed displaying of next date of an event if its near.
```
LeandraH, line 7:
```suggestion
en: Fixed displaying of next date of an event if it's soon.
```

shared/api/models/__tests__/DateModel.spec.ts
```diff
@@ -141,8 +141,8 @@ describe('DateModel', () => {
         recurrenceRule: rrulestr('DTSTART:20230414T050000\nRRULE:FREQ=WEEKLY;BYDAY=MO'),
       })
 
-      expect(date.startDate).toEqual(DateTime.fromISO('2023-10-09T07:00:00.000+02:00'))
-      expect(date.endDate).toEqual(DateTime.fromISO('2023-10-10T09:00:00.000+02:00'))
+      expect(date.startDate).toEqual(DateTime.fromISO('2023-10-02T07:00:00.000+02:00'))
```
f1sh1918, line 144: why did you change this?

shared/api/models/DateModel.ts
```diff
@@ -127,6 +128,17 @@ class DateModel {
...
+  private currentDateToRrule(): Date {
+    // rrule is not correctly handling timezones and is expecting dates in "local time", i.e. in the timezone of dstart
+    // Therefore we have to manually apply timezone offset, e.g. for daylight savings time
+    // https://github.com/jkbrzt/rrule#important-use-utc-dates
+    // Also include dates happening at the moment by subtracting the duration
+    return DateTime.now()
+      .minus(this._duration)
+      .minus({ minutes: this._offset - DateTime.now().offset })
+      .toJSDate()
+  }
```
LeandraH, line 140: I am getting a little confused by the adding and subtracting of different offsets here and in the previous method. Maybe we can instead use `Luxon` methods like `setZone`?

shared/api/models/DateModel.ts
```diff
@@ -127,6 +128,17 @@ class DateModel {
     return dateTime.plus({ minutes: this._offset - dateTime.offset })
   }
 
+  private currentDateToRrule(): Date {
+    // rrule is not correctly handling timezones and is expecting dates in "local time", i.e. in the timezone of dstart
+    // Therefore we have to manually apply timezone offset, e.g. for daylight savings time
+    // https://github.com/jkbrzt/rrule#important-use-utc-dates
+    // Also include dates happening at the moment by subtracting the duration
+    return DateTime.now()
+      .minus(this._duration)
```
LeandraH, line 137: Why are we subtracting the duration here?

f1sh1918:
@steffenkleinle can you describe how to reproduce this issue.
Just created a new event as described in issue description and can't reproduce it
https://webnext.integreat.app/testumgebung/de/events/test-recurrence

LeandraH (commented):
I wasn't able to reproduce the original issue without mocking (did manage to lock myself out of a few accounts by changing time zones on my laptop though, yay 😅 ) but I follow your reasoning. But I would prefer methods with clearer names to not turn the `rruleToDateTime` and `currentDateToRrule` methods into ones that I for one would be scared to touch in the future.

f1sh1918 (approved):
looks fine to me

shared/api/models/DateModel.ts
f1sh1918: to show events that already started but didn't end yet

shared/api/models/DateModel.ts
steffenkleinle: I didn't figure out how to make that work, since (afaik) it only changes the timezone, the actual time (relative to UTC stays the same). I.e. it changes from 00:00 UTC to 01:00 UTC +1 which is exactly the same point in time, just in a different timezone.

steffenkleinle:
> @steffenkleinle can you describe how to reproduce this issue. Just created a new event as described in issue description and can't reproduce it https://webnext.integreat.app/testumgebung/de/events/test-recurrence

Its only possible to reproduce by creating the event during summer time (or probably easier, mocking the rrule).

steffenkleinle:
> I wasn't able to reproduce the original issue without mocking (did manage to lock myself out of a few accounts by changing time zones on my laptop though, yay 😅 ) but I follow your reasoning. But I would prefer methods with clearer names to not turn the `rruleToDateTime` and `currentDateToRrule` methods into ones that I for one would be scared to touch in the future.

@LeandraH what names do you propose? I am not sure if I get your point :see_no_evil:

shared/api/models/__tests__/DateModel.spec.ts
steffenkleinle: Now recurring events that are currently happening are also shown. Therefore the first recurrence is the 02.10.2023, not just the 09.10.2023 (as the mocked date is the 02.10.2023). However, I just changed the mocked date now.

shared/api/models/DateModel.ts
LeandraH: Ah, nice, thanks!

LeandraH:
> > I wasn't able to reproduce the original issue without mocking (did manage to lock myself out of a few accounts by changing time zones on my laptop though, yay 😅 ) but I follow your reasoning. But I would prefer methods with clearer names to not turn the `rruleToDateTime` and `currentDateToRrule` methods into ones that I for one would be scared to touch in the future.
> 
> @LeandraH what names do you propose? I am not sure if I get your point 🙈

Sorry, poor phrasing, I didn't mean the names `rruleToDateTime` and `currentDateToRrule`, I meant the adding and subtracting in there, like I also wrote on those lines, and you already answered. I think the names `rruleToDateTime` and `currentDateToRrule` are pretty clear 👍

LeandraH (approved):

Issue #2610, opened 2023-12-06
Enable export of recurring event in web


LeandraH:
**Is your feature request related to a problem? Please describe.**
The user can currently export single events on the web. For a recurring event, they should also be able to export the event series.

**Additional context**
Design is here: https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?type=design&node-id=1731-44&mode=design&t=CrzJcaUD0nhgdXpm-0

PR #2651, opened 2024-01-30
2610: Export recurring events in web

LeandraH:
### Short description

Sorry, I messed up the rebasing for #2645, new branch here. Copied from the other one:

Users can so far choose to export a recurring event as a series or as a single event in native, here they can also do that in web. It looks a little different than in the [design](https://www.figma.com/file/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=2043%3A147&mode=dev) because we can't choose the calendar here and so I decided to just add this little extra radio group at the bottom instead of replacing the entire content of the event.

### Proposed changes

- Added a boolean to EventModel.toICal in order to either export a single event or the whole series
- Added two radio buttons for choosing between exporting a single event and a series of them to the EventsPage

### Side effects

- Removed the direction prop from the RadioGroup and instead used inset-inline-start there

### Resolved issues

Fixes: #2610 

---


web/src/routes/EventsPage.tsx
```diff
@@ -126,10 +146,43 @@ const EventsPage = ({ city, pathname, languageCode, cityCode }: CityRouteProps):
   if (event) {
     const { featuredImage, thumbnail, lastUpdate, content, title, location, date } = event
     const defaultThumbnail = featuredImage ? featuredImage.medium.url : thumbnail
-
-    const PageFooter = (
-      <StyledButton onClick={() => downloadEventAsIcsFile(event)} text={t('exportAsICal')} fullWidth={viewportSmall} />
-    )
+    const isRecurring = event.date.recurrenceRule && event.date.recurrenceRule.count() > 1
```
steffenkleinle, line 149: Optional: Could we move the whole event export logic in this file to a separate component, e.g. EventExport or similar? I think that way we'd have less logic in this component, clearer separation of concern, possibility of reusing the event export and easier testing.

shared/api/models/EventModel.ts
```diff
@@ -51,11 +52,11 @@ class EventModel extends ExtendedPageModel {
     return this._featuredImage
   }
 
-  toICal(baseUrl: string, appName: string): string {
+  toICal(baseUrl: string, appName: string, recurring: boolean): string {
     const { title, location, path, date, excerpt, lastUpdate } = this
     const url = `${baseUrl}${path}`
     const uid = v5(`${url}/${formatDateICal(lastUpdate)}`, v5.URL)
-    const timezone = date.startDate.zone
+    const timezone = date.startDate.zone.name
```
steffenkleinle, line 59: Was the previous implementation wrong here?

steffenkleinle (changes requested):
On mobile devices clicking on export, scrolling down and then clicking cancel or exporting leads to the footer having too much space:
image

Otherwise it works as expected.

shared/api/models/EventModel.ts
LeandraH: Yup, it lead to the `TZID=[object Object]` that Steffi found

ztefanie (approved):
Not reviewed again. Tested and output is valid.

web/src/components/ExportEventButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+  event: EventModel
+}
+
+const ExportEventButton = ({ event }: ExportEventButtonProps): ReactElement => {
+  const [isExporting, setIsExporting] = useState<boolean>(false)
+  const [exportRecurring, setExportRecurring] = useState<boolean>(false)
+  const { t } = useTranslation('events')
+  const { viewportSmall } = useWindowDimensions()
+
+  const isRecurring = event.date.recurrenceRule && event.date.recurrenceRule.count() > 1
```
steffenkleinle, line 43: I think it would perhaps be better to check using our own method. Here we might have edge cases with different timezones, durations and so on again.
```suggestion
  const isRecurring = date.hasMoreRecurrencesThan(1)
```

steffenkleinle (approved):
Not tested again, well done :rocket:

Issue #2650, opened 2024-01-30
Set page title as ticket title for zammad


sarahsporck:
**Is your feature request related to a problem? Please describe.**
As the PBO can be applied on any page it makes sense to add the name of the page to the ticket. Setting it as the ticket title makes the most sense, I think. 

**Describe the solution you'd like**
Add the title of the page to the zammad ticket. 
Set the region in the body instead.

**Describe alternatives you've considered**
- not adding it
- adding it to the body instead

**Additional context**

#2315

PR #2661, opened 2024-02-05
2650: Add page url to zammad ticket & fix submitting for native

sarahsporck:
### Short description

Also send url of page to zammad, so that it is possible to see from where the message came from.
Add crypto polyfill for native, so that we can use `crypto.getRandomValues` for email generation.

### Proposed changes

- add pagetitle & url in web & native
- polyfill crypto.getRandomValues

### Side effects

Maybe check out the event export in native, as the polyfill is apparently "required" for the usage of `uuid` (which is used in the `EventModel.toICal` method.

### Resolved issues

Fixes: #2650 

We should do a manual release after this is merged, so that this feature can be tested properly.

---


shared/api/endpoints/submitMalteHelpForm.ts
```diff
@@ -37,15 +37,15 @@ const contactChannelText = (contactChannel: ContactChannel): string => {
...
+  url,
   name,
   roomNumber,
   contactChannel,
   additionalContactInformation,
   contactGender,
   comment,
 }: BuildHelpFormBodyParams): string => `
-Sprache: ${languageCode}
+Gesendet von ${url}
```
steffenkleinle, line 48: Why no more language? Because its part of the url?

steffenkleinle (approved):
The footer is too big on web, but probably already fixed by merging in main since @LeandraH fixed that in #2651:


Another small possible improvement (which we could also create a separate issue for): Use the keyboard types email and phone numbers for the input fields.

Otherwise working fine, tested on firefox and android.

shared/api/endpoints/submitMalteHelpForm.ts
sarahsporck: Jep exactly. Do you think I should readd it? I think just the BCP tag feels a little bit weird here. The url provides more context.

steffenkleinle: No, I guess its fine. But probably more of a question for the maltesers/@osmers.

osmers: Hm, I think that might be fine - but for non-technical users maybe a bit more difficult to read? Maybe we can even use the language name instead of the code?

Having the URL definitely helps though bcs we also need to be able to send out emails only to certain groups based on the region the form was submitted in.

f1sh1918 (approved):
Tested on chrome and ios. Works fine
Kleiner typo hier noch. Vll kannst du den noch fixen: "erhältst"

image

shared/api/endpoints/submitMalteHelpForm.ts
sarahsporck: Jep but we only have the "native" language names in the app, right? Not sure if they could understand what فارسی means 🤔

Issue #2660, opened 2024-02-05
News not scrollable


Docmale:
### Describe the Bug

Neither Local nor tünewes are scrollable

### Steps to Reproduce

1. Go to Testumgebung
2. Click on Nerws
3. Scroll down (swipe up) LOCAL or tünews
4. No scrolling

### Additional Information
Web is ok
since version 24.02.0; previous version is ok

PR #2662, opened 2024-02-05
2660: Fix news not scrollable

ztefanie:
### Short description

Removed the content flex of the List.

Maybe we should trigger a beta today directly with this fix.

### Side effects

- Maybe something with the pois does not work as expected, compare to https://github.com/digitalfabrik/integreat-app/pull/2640

### Resolved issues

Fixes: #2660

---

### DIFF
--- a/native/src/components/List.tsx
+++ b/native/src/components/List.tsx
@@ -44,7 +44,6 @@ const List = <T,>({
     }
     onEndReached={onEndReached}
     showsVerticalScrollIndicator={false}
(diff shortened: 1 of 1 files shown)


LeandraH (approved):
Nice, tested on emulated iOS. Thanks for the quick fix!

f1sh1918 (approved):
Pois work fine. Just tested the issue why we changed the styles there which maybe caused these problems with the list item

Tested on ios simulator and android emulator

Issue #2659, opened 2024-02-05
Integreat cities accessible in halloaschaffenburg


steffenkleinle:
### Describe the Bug

At the moment it is possible to open all integreat cities in halloaschaffenburg. This should not be possible.

### Steps to Reproduce

1. Go to https://halloaschaffenburg.de/kreis-bergstrasse/de
2. See the integreat content of Kreis Bergstrasse

### Expected Behavior

An error should be displayed that this page does not exist.

### Actual Behavior

The integreat content is displayed.

### Additional Information

Also check the native app. Accessing any content other than the `hallo` instance should not be possible.

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #2663, opened 2024-02-05
2659: Don't load other cities when fixed city is set

sarahsporck:
### Short description

Currently it is possible to open IG cities in Aschaffenburg, when directly navigating to the url 
-> https://halloaschaffenburg.de/kreis-bergstrasse/de
This should be avoided, by not loading any other city than the fixed city.

### Proposed changes

- Add a fixedCityContentSwitcher, which checks whether the params `cityCode` is the `fixedCity`

### Resolved issues

Fixes: #2659

### DIFF
--- a/native/src/components/PoiDetails.tsx
+++ b/native/src/components/PoiDetails.tsx
@@ -49,7 +49,6 @@ type PoiDetailsProps = {
 
 const PoiDetails = ({ poi, language, distance }: PoiDetailsProps): ReactElement => {
   const { t } = useTranslation('pois')
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
   const thumbnail = poi.thumbnail?.replace('-150x150', '') ?? PoiThumbnailPlaceholderLarge
   const { title, content, email, website, phoneNumber, openingHours, temporarilyClosed, isCurrentlyOpen, category } =
     poi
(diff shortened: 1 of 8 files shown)


LeandraH (approved):
Very nice, thanks! Tested in Firefox. Maybe also add a tiny test?

ztefanie (approved):

steffenkleinle (approved):

steffenkleinle (commented):
Why did you choose to add another component instead of just adding anther condition to the if-clause in https://github.com/digitalfabrik/integreat-app/blob/ccd4c73d658f5ace835bfeca0ec92da028698825/web/src/CityContentSwitcher.tsx#L60?
``` ts
  const { fixedCity } = buildConfig().featureFlags
  if ((!city && !loading) || (fixedCity && cityCode !== fixedCity)) {
```

sarahsporck:
> Why did you choose to add another component instead of just adding anther condition to the if-clause [...]

This way we don't even have to make the request to the endpoint. Therefore, it is faster and uses less traffic.

native/src/components/PoiDetails.tsx
```diff
@@ -49,7 +49,6 @@ type PoiDetailsProps = {
 
 const PoiDetails = ({ poi, language, distance }: PoiDetailsProps): ReactElement => {
   const { t } = useTranslation('pois')
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
sarahsporck, line 52: appears to be unused and threw a linting warning

Issue #2634, opened 2024-01-17
Don't show last update for empty category with embedded offers


steffenkleinle:
### Describe the Bug

Showing a last update for an empty category with only embedded offers is pretty confusing, because the list of the offers was not updated last time on that dates.

Furthermore, the embedded offers should have more padding after the timestamp (see https://chat.tuerantuer.org/digitalfabrik/pl/7czoxkzkebfpxmt83qjfioa4cr).

### Steps to Reproduce

1. Go to https://webnext.integreat.app/testumgebung/de/arbeit/sprungbrett
2. See last update

### Expected Behavior

- [ ] No last update is shown if the category is empty.
- [ ] If there is content, make it obvious that the "last update" only refers to the content, e.g. by adding some padding.

### Actual Behavior

A last update is shown if the category is empty.

### Additional Information

![image](https://github.com/digitalfabrik/integreat-app/assets/29095487/3e33449f-88e1-450a-8604-76a7e7671664)

ztefanie:
Tested on Chrome works nice :tada:

steffenkleinle:
Native is still missing

PR #2664, opened 2024-02-06
2634: Don't show last update for empty category with embedded offers

Njivaniaina2407:
### Short description

Showing a last update for an empty category with only embedded offers is pretty confusing, because the list of the offers was not updated last time on that dates.

### Proposed changes

Set a condition before displaying the last update

### Resolved issues

Fixes: #2634 (web only)

---


web/src/components/Page.tsx
```diff
@@ -36,7 +36,7 @@ const Page = ({
   thumbnailSrcSet,
   content,
   lastUpdate,
-  showLastUpdateText = true,
+  showLastUpdateText = false,
```
steffenkleinle, line 39: I think you shouldn't just change this. This changes behavior in a lot of other use cases. Please revert this and instead pass `false` as a prop in the specific case you don't want to show it.

web/src/components/LastUpdateInfo.tsx
```diff
@@ -26,7 +26,7 @@ export const LastUpdateInfo = ({
   const { i18n, t } = useTranslation('common')
   return (
     <TimeStamp className={className}>
-      {withText && t('lastUpdate')} {lastUpdate.setLocale(i18n.language).toFormat(format)}
+      {withText && t('lastUpdate')} {withText && lastUpdate.setLocale(i18n.language).toFormat(format)}
```
steffenkleinle, line 29: This changes the behavior of the prop. The prop is just there to decide whether to just display a date or a date and the text `last update:`.  A better option would be to just not render the whole component if you don't want it to show at all.

steffenkleinle (changes requested):

steffenkleinle (changes requested):
Looking good.

Native is missing though, I can take over :)

steffenkleinle (approved):
Looks good! Lets do native in a separate PR :)

ztefanie (approved):
Tested on Chrome works nice 🎉

PR #2670, opened 2024-02-12
2634: Add space between embedded offers and last update (native)

steffenkleinle:
### Short description

Add space between embedded offers and last update (native).

### Proposed changes

Add more space between last update and embedded offers if there is content and an embedded offer.

### Side effects

None.

### Resolved issues

Fixes: #2634.

---

### DIFF
--- a/native/src/components/EmbeddedOffers.tsx
+++ b/native/src/components/EmbeddedOffers.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react'
+import styled from 'styled-components/native'
 
 import { CATEGORIES_ROUTE, MALTE_HELP_FORM_OFFER_ROUTE, SPRUNGBRETT_OFFER_ALIAS } from 'shared'
 import { CategoryModel } from 'shared/api'
@@ -6,6 +7,10 @@ import { CategoryModel } from 'shared/api'
 import urlFromRouteInformation from '../navigation/url'
 import MalteHelpFormOffer from './MalteHelpFormOffer'
 import SprungbrettOffer from './SprungbrettOffer'
+
+const Container = styled.View<{ withMargin: boolean }>`
+  ${props => props.withMargin && 'margin-top: 32px;'}
+`
 
 type EmbeddedOfferProps = {
   category: CategoryModel
@@ -18,7 +23,11 @@ const EmbeddedOffers = ({ category, cityCode, languageCode, goBack }: EmbeddedOf
   const offer = category.embeddedOffers[0]
   switch (offer?.alias) {
     case SPRUNGBRETT_OFFER_ALIAS:
-      return <SprungbrettOffer sprungbrettOffer={offer} languageCode={languageCode} />
+      return (
+        <Container withMargin={!!category.content}>
+          <SprungbrettOffer sprungbrettOffer={offer} languageCode={languageCode} />
+        </Container>
+      )
     case MALTE_HELP_FORM_OFFER_ROUTE: {
       const url = urlFromRouteInformation({
         route: CATEGORIES_ROUTE,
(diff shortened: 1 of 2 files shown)


native/src/components/EmbeddedOffers.tsx
```diff
@@ -27,13 +36,15 @@ const EmbeddedOffers = ({ category, cityCode, languageCode, goBack }: EmbeddedOf
...
       })
       return (
-        <MalteHelpFormOffer
-          categoryPageTitle={category.title}
-          url={url}
-          malteHelpFormOffer={offer}
-          cityCode={cityCode}
-          onSubmit={goBack}
-        />
+        <Container withMargin={!!category.content}>
```
ztefanie, line 39: I think we should not make this necessary on every switch-case element, but move it to a higher level. Especially when adding more cases for other offers this is a lot of duplication and makes the code harder to read and maintain.

Suggestion:
```

const EmbeddedOffer = (embeddedOfferProps: EmbeddedOfferProps): ReactElement | null => {
  const { category, cityCode, languageCode, goBack } = embeddedOfferProps
  const offer = category.embeddedOffers[0]
  switch (offer?.alias) {
    case SPRUNGBRETT_OFFER_ALIAS:
      return (
        <Container withMargin={!!category.content}>
          <SprungbrettOffer sprungbrettOffer={offer} languageCode={languageCode} />
        </Container>
      )
    case MALTE_HELP_FORM_OFFER_ROUTE: {
      const url = urlFromRouteInformation({
        route: CATEGORIES_ROUTE,
        languageCode,
        cityCode,
        cityContentPath: category.path,
      })
      return (
        <Container withMargin={!!category.content}>
          <MalteHelpFormOffer
            categoryPageTitle={category.title}
            url={url}
            malteHelpFormOffer={offer}
            cityCode={cityCode}
            onSubmit={goBack}
          />
        </Container>
      )
    }
    default:
      return null
  }
}

const EmbeddedOffers = (embeddedOfferProps: EmbeddedOfferProps): ReactElement | null => {
  const { category } = embeddedOfferProps
  return (
    <Container withMargin={!!category.content}>
      <EmbeddedOffer {...embeddedOfferProps} />
    </Container>
  )
}
```

ztefanie (changes requested):
Tested on android, works as expected.

native/src/components/EmbeddedOffers.tsx
steffenkleinle: Good point, thanks!

ztefanie (approved):
Thanks for the adjustment. :fire: 
Not tested again.

web/src/components/EmbeddedOffers.tsx
```diff
@@ -12,29 +12,29 @@ const Container = styled.div<{ withMargin: boolean }>`
   ${props => props.withMargin && 'margin-top: 48px;'}
 `
 
-type EmbeddedOfferProps = {
+type EmbeddedOffersProps = {
   category: CategoryModel
 } & CityRouteProps
 
-const EmbeddedOffers = ({ category, ...props }: EmbeddedOfferProps): ReactElement | null => {
+const EmbeddedOffer = ({ category, ...props }: EmbeddedOffersProps): ReactElement | null => {
```
LeandraH, line 19: Wait, one became singular, one plural, I am confused

LeandraH (approved):
Nice, thanks! Tested in emulated iOS

web/src/components/EmbeddedOffers.tsx
steffenkleinle: That was the suggestion from @ztefanie and I think its okay. If you have some better naming idea let me know :)

Issue #2571, opened 2023-11-20
Feature Request: Add parameters for filter selection when calling the POI map


nassabay:
**Is your feature request related to a problem? Please describe.**
It would be good to be able to include a selection via filtering when calling the map, so when the user follows this link to the map, the map already shows the selected filter.

**Describe the solution you'd like**
Add the possibility to include the filter parameters when calling the POI map. These filter parameters should be selected already and define the view.

**Describe alternatives you've considered**
None.

**Additional context**

steffenkleinle:
We'd need additional query parameters:
- Selected poi category
- Only show currently open (optional)
- Filter selection currently opened (optional)

We should probably also think about error handling (e.g. if poi specified by slug is not part of the selected filter...).

PR #2671, opened 2024-02-12
2571: Add map query params

steffenkleinle:
### Short description

Add query params for zoom and poi category filter.

Don't worry, most of the changes are tests :bug: 

### Proposed changes

- Add query param for currently selected poi category (included in share url/pathname)
- Add query param for initial zoom level (not included in share url/pathname)
- Adjust behavior as following:
  - Empty map press: Single and multi poi are deselected and filter modal is closed, filters are retained
  - Selecting map feature (single or multi poi): Previous single or multi pois are deselected, filters are retained
  - Selecting single poi in list: Multi poi and filters are retained
  - Selecting a poi category filter: Single poi and multi pois are deselected
  - Deselecting a poi category filter: Single poi and multi pois are retained
  - Selecting currently open filter: Single poi and multi pois are deselected
  - Deselecting currently open filter: Single poi and multi pois are retained
  - Back to overview button: First single poi is deselected, then multi poi, filters are retained
  - Zoom is never retained (i.e. just initially set via url)
- Extract Pois and PoiFiltersOverlayButtons components (web)
- Fix page title to account for single pois (web)

Does this make sense? Please let me know if you don't understand something or think its not intuitive/good UI/UX.

### Rejected changes

I decided against implementing query params for both currently open filter and filter selection currently shown. Reasoning:
- Currently open filter: The result of this filter is not always the same as it depends on the weekday/time. Therefore, including this in the url could lead to confusing results (i.e. "but it worked on my phone")
- Filter selection shown: I did not see a use case for this and since it would just increase url length and complexity I opted against.

Let me know if you think differently.

### Side effects

None (I think).

**Please test this very thoroughly! I am not sure about quite some behavior here!** :warning:

### Resolved issues

Fixes: #2571.

---


web/src/components/Pois.tsx
```diff
@@ -0,0 +1,165 @@
+import React, { ReactElement, useState } from 'react'
```
steffenkleinle, line 1: Mostly just extracted from `PoisPage`

LeandraH (commented):
Hmm, when I run `npx uri-scheme open integreat://integreat-app/muenchen/en/locations\?category\=4 --ios`, it opens to the map but no filters are enabled. It's currently not opening at all in Android so can't test there, sorry, I'll try again next week.
Short testing in Firefox worked.

steffenkleinle:
> Hmm, when I run `npx uri-scheme open integreat://integreat-app/muenchen/en/locations\?category\=4 --ios`, it opens to the map but no filters are enabled. It's currently not opening at all in Android so can't test there, sorry, I'll try again next week. Short testing in Firefox worked.

Good catch! Should be fixed :)

LeandraH (approved):
Very cool! Tested in Firefox and real iOS (still no dice on Android, sorry, someone else will have to test that)

steffenkleinle:
> Very cool! Tested in Firefox and real iOS (still no dice on Android, sorry, someone else will have to test that)

I think you perhaps have to set the app to support the deep link manually (at least for the test cms apk). You can do that in the settings for the app: Open by default > Add link > integreat.app. After that I was able to open https integreat.app links

LeandraH:
> > Very cool! Tested in Firefox and real iOS (still no dice on Android, sorry, someone else will have to test that)
> 
> I think you perhaps have to set the app to support the deep link manually (at least for the test cms apk). You can do that in the settings for the app: Open by default > Add link > integreat.app. After that I was able to open https integreat.app links

You're right, I did, thanks! I even vaguely remember disabling the deep linking for something a while ago 😅  Anyway, worked with that! Very nice!

f1sh1918 (approved):
Looks fine to me tested on android and chrome.
Not sure why we need the zoom as a query param @steffenkleinle ?
Since it won't be added to the url no one will share it?

steffenkleinle:
> Looks fine to me tested on android and chrome. Not sure why we need the zoom as a query param @steffenkleinle ? Since it won't be added to the url no one will share it?

It was a request by the maltesers to be able to share a map with specific zoom, so like this you can manually add a zoom query param that is then initially set.
See https://chat.tuerantuer.org/digitalfabrik/pl/hsuepxecj38b9mmcpfy4h3us9o

Issue #2677, opened 2024-02-19
Share link is not clickable


f1sh1918:
### Describe the Bug

When a user shares a link via the toolbar the link is not separated from the title and not recognized as a clickable, because there is no space or new line between page title and lonk

### Steps to Reproduce

1. https://integreat.app/lkaugsburg/de/locations/kloster-obersch%C3%B6nefeld
2. Click on 'teilen'
3. Select mail f.e.

### Expected Behavior

The link should be clickable and separated from the title 

### Actual Behavior

Link is in one line and appears directly after the page title

### Additional Information

<img width="741" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/37902063/9ad4dbb2-62d3-4441-a4e6-75e7faf06273">

PR #2681, opened 2024-02-20
2677: Share link not clickable

f1sh1918:
### Short description

When a user wants to share a page f.e. via mail the link is not clickable due to missing whitespace

### Proposed changes

- add a whitespace

### Resolved issues

Fixes: #2677

---


LeandraH (approved):
Thanks

Issue #2682, opened 2024-02-20
Adjust share message


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently the share messages differs from web to native. For native we only use `Gerade gefunden: <link>`
For web we use: `Gerade gefunden: <Page title> - <Location> <link>`
**Describe the solution you'd like**
We should use the same structure for both platform
- add location to share message if exists
- add page title (check exceptions that)

**Additional context**
Check the exception as we also have for the `Header title` for Pois, etc

PR #2684, opened 2024-02-20
2682: Add proper share message

f1sh1918:
### Short description

The share message should include a proper title and location to describe what the user is currently sharing

### Proposed changes

- add location to share message
- add page title to share message
- added unit tests

### Resolved issues

Fixes: #2682

---


native/src/components/Header.tsx
```diff
@@ -82,14 +86,45 @@ const Header = ({
...
+        return t('locations')
+      }
+    }
+
+    if (route.name === NEWS_ROUTE) {
+      const newsType = (route.params as { newsType: TuNewsType | LocalNewsType }).newsType
+      if (newsType === 'tu-news') {
+        return 'Tü News'
+      }
+      return t('localInformation')
```
steffenkleinle, line 108: I think this should be local news

steffenkleinle (changes requested):
:tada: 
Could you perhaps add some tests for this checking the different cases?

native/src/hooks/useHeader.tsx
```diff
@@ -29,10 +31,11 @@ const useHeader = <T extends RoutesType>({
           showItems={!!data}
           availableLanguages={availableLanguages}
           shareUrl={shareUrl}
+          cityName={cityName}
```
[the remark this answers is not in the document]
f1sh1918: used first one

native/src/components/Header.tsx
```diff
@@ -82,14 +86,45 @@ const Header = ({
...
+    const pageTitle = (route.params as { title?: string } | undefined)?.title
+    if (!previousRoute) {
+      // Home/Dashboard: Show city name
+      return cityName ?? ''
+    }
+
+    if (route.name === POIS_ROUTE && previousRoute.name === POIS_ROUTE) {
+      const poisRouteParams = route.params as RoutesParamsType[PoisRouteType]
+      if (poisRouteParams.slug || poisRouteParams.multipoi) {
+        return t('locations')
```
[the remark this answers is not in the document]
f1sh1918: you are right. this check is only needed for multipois to use sth like 'Gefunden: Mehrere Orte - <Location> <Link>'
The normal title is already set in the first line of the fkt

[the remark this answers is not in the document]
f1sh1918: yes we should update this also on web but thats couple of additional work which i would do in a separate task, since its not that easy to get the poi title. You can parse the url but thats not a nice solution

f1sh1918:
> 🎉 Could you perhaps add some tests for this checking the different cases?

i added tests for all special cases and one for general case (categories route)
Let me know if we need more @steffenkleinle

native/src/components/Header.tsx
```diff
@@ -82,14 +88,33 @@ const Header = ({
...
+    if (route.name === NEWS_ROUTE) {
+      const newsType = (route.params as { newsType: TuNewsType | LocalNewsType }).newsType
+      if (newsType === TU_NEWS_TYPE) {
+        return tunewsLabel
+      }
+      return t('localNews')
+    }
+
+    return pageTitle ?? t(route.name)
+  }
```
steffenkleinle, line 107: Sorry I did not realize that earlier but the share title should always just be the pageTitle/fallback to `route.name` controlled by the separate routes instead deciding here in the header what would be a good share title. That way:
- page title and share title are consistent
- it is controlled by the routes and therefore easier extendable
- No weird typecasts necessary

To fix the few occurrences (AFAICS only local news list, tünews list and tünews details are wrong), you can use `useSetRouteTitle` and set the correct route titles that way.

native/src/components/Header.tsx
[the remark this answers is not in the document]
steffenkleinle: Actually I think I already (accidentally) fixed that in #2671 by fixing the page title for pois.

native/src/components/Header.tsx
f1sh1918: hm but this `useSetRouteTitle` is not used on everyRoute f.e. Disclaimer etc. so i have to add it there too, i guess

steffenkleinle: exactly, route title with a fallback to `t(route.name)`

f1sh1918: i think the main issue in general is with the News overview pages that if we make a `GoBack` we don't get the correct route title.
Shouldn't it just be possible to use the 
`useSetRouteTitle({ navigation, title: <title> })` in News Overview and Details pages and pass there the correct title?

steffenkleinle: I think its overwritten by the usage of  `useSetRouteTitle` in `News` component

native/src/components/Header.tsx
```diff
@@ -82,14 +88,33 @@ const Header = ({
...
+    if (route.name === POIS_ROUTE && poisRouteParams.multipoi) {
+      return t('pois:multiPois')
+    }
+
+    if (route.name === NEWS_ROUTE) {
+      const newsType = (route.params as { newsType: TuNewsType | LocalNewsType }).newsType
+      if (newsType === TU_NEWS_TYPE) {
+        return tunewsLabel
+      }
+      return t('localNews')
```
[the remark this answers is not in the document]
f1sh1918: moved this implementation to `News` and fixed it

native/src/components/__tests__/Header.spec.tsx
```diff
@@ -158,11 +179,65 @@ describe('Header', () => {
...
+
+    expect(share).toHaveBeenCalledWith({
+      message: `${t('shareMessage')}: ${t('disclaimer')} - ${cityDisplayName(cityModel)} ${defaultShareUrl}`,
+      title: 'Integreat',
+    })
+    expect(sendTrackingSignal).toHaveBeenCalledWith({
+      signal: { name: SHARE_SIGNAL_NAME, url: 'https://example.com/share' },
+    })
+  })
+  it('should remove the page title in the share message if it equals the city name', () => {
```
LeandraH, line 226:
```suggestion

  it('should remove the page title in the share message if it equals the city name', () => {
```

LeandraH (approved):
😍  Very cool, tested in emulated iOS

native/src/utils/cityDisplayName.ts
```diff
@@ -4,10 +4,13 @@ import dimensions from '../constants/dimensions'
 
 export const forceNewlineAfterChar = (text: string, char: string): string => text.replace(char, `${char}\n`)
 
-const cityDisplayName = (city: CityModel | undefined, deviceWidth: number): string => {
+const cityDisplayName = (city: CityModel | undefined, deviceWidth?: number): string => {
   if (!city) {
     return ''
   }
+  if (!deviceWidth) {
```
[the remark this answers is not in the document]
f1sh1918: i adjusted the function slightly and added a `cityNameFitsSingleLine` boolean

Issue #2626, opened 2024-01-10
Link not clickable in push notification


osmers:
### Describe the Bug

A link is not always fully linked within the push notification 

### Steps to Reproduce

1. Go to https://integreat.app/castrop-rauxel/ru/news/local/5917
2. Click on the link
3. See link not opening/error

### Expected Behavior

The link should the considering the complete URL and link it entirely 

### Actual Behavior

Only part of the URL is highlighted and linked. Link does not work. 

### Additional Information

Page to edit the push notification: https://admin.integreat-app.de/castrop-rauxel/push-notifications/de/979/edit/ 

More info from Sven: https://chat.tuerantuer.org/digitalfabrik/pl/fffs5x8uxpb67gnbzc74jza6cc

LeandraH:
The link in question is https://integreat.app/castrop-rauxel/ru/информация-помощь-украине/разрешение-на-временное-го-года, we should check on the encoding here.

PR #2696, opened 2024-02-23
2626: Match non-ascii characters for links

steffenkleinle:
### Short description

Match non-ascii characters while detecting links.

### Proposed changes

- Match non-ascii characters using regex `[^\u0020-\u007F]`

### Side effects

:shrug: there might be some, any regex experts here that know? :smile: 

### Testing

Send a push notification with a link similar to this (or just replace it in the code):
> TEst odnskfjadslkfjsjf https://integreat.app/castrop-rauxel/ru/Düsseldorf/инфор.мация-помощь-украине/разрешение-на-временное-го-года?asdf sfasklfdskf.sadjfkdjaslf@test.com jfdsafsjkdfls

### Resolved issues

Fixes: #2626.

---


shared/utils/replaceLinks.ts
```diff
@@ -1,9 +1,11 @@
...
+// [^\u0020-\u007F] matches non-ASCII characters (e.g. umlaute, cyrillic, arabic)
+// https://stackoverflow.com/a/150078
 const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/
 const hostname = /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
-const path = /(?:\/(?:[+~%/.\w-]*[+~%/\w-]+)?)?/
-const query = /(?:\?([-+=&;%@.\w]*[-+=&;%@\w]+)?)?/
-const hash = /(?:#([.!/\\\w]*[!/\\\w]+)?)?/
+const path = /(?:\/(?:([+~%/.\w-]|[^\u0020-\u007F])*([+~%/\w-]|[^\u0020-\u007F]+))?)?/
+const query = /(?:\?(([-+=&;%@.\w]|[^\u0020-\u007F])*([-+=&;%@\w]|[^\u0020-\u007F])+)?)?/
+const hash = /(?:#(([.!/\\\w]|[^\u0020-\u007F])*([!/\\\w]|[^\u0020-\u007F])+)?)?/
```
sarahsporck, line 8: Something is not right here. When using this test string:
```
https://stackoverflow.com/a/150078
https://stackoverflow.com/a/150078
https://stackoverflow.com/a/150078
```
There should be three matches with the same url. But instead it now matches across the line break. 
Compare: 
old regex -> https://regex101.com/r/bjF2h4/1
new regex -> https://regex101.com/r/DuReg4/1

sarahsporck (changes requested):

shared/utils/replaceLinks.ts
sarahsporck: Maybe we should use sth like `a-zA-Z0-9\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF` instead of `\w` as suggested in the comment of [this stackoverflow post](https://stackoverflow.com/a/6041965)?

sarahsporck: Or check out xregexp as it also supports unicode symbols natively :) https://stackoverflow.com/questions/26133593/using-regex-to-match-international-unicode-alphanumeric-characters-in-javascript

shared/utils/replaceLinks.ts
```diff
@@ -1,9 +1,11 @@
 // https://stackoverflow.com/a/3809435
+// [^\u0020-\u007F] matches non-ASCII characters (e.g. umlaute, cyrillic, arabic)
+// https://stackoverflow.com/a/150078
 const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/
 const hostname = /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
-const path = /(?:\/(?:[+~%/.\w-]*[+~%/\w-]+)?)?/
-const query = /(?:\?([-+=&;%@.\w]*[-+=&;%@\w]+)?)?/
-const hash = /(?:#([.!/\\\w]*[!/\\\w]+)?)?/
+const path = /(?:\/(?:([+~%/.\w-]|[^\u0020-\u007F])*([+~%/\w-]|[^\u0020-\u007F]+))?)?/
```
ztefanie, line 6: This does not really make sense. You say it should be a word with latin leters and +~... or it should match anything. The second part is always true, when the first part is true, so the first part can be deleted.

Also i do not get why this is repeated once with the * and once with the +. Coudn't the * part be deleted as it is just duplication with no changes?

steffenkleinle: @ztefanie it does make sense. E.g. for path segemnts, dots (`.`) are allowed in the middle of a segment but not at the end, thats why there is **nearly** identical  parts in the regexes.

But I adjusted the unicode matching part according to @sarahsporck's suggestion :)

sarahsporck (approved):
Nice! Looks good :>

shared/utils/replaceLinks.ts
ztefanie: Sorry about not being clear, with first part, i meant `[+~%/.\w-]` (the first part before the or) and by second part i meant `[^\u0020-\u007F]`. 

> You say it should be a word with latin leters and +~... or it should match anything. The second part is always true, when the first part is true, so the first part can be deleted.

steffenkleinle: Ahhh okay. Thats not the case anymore though, is it?

ztefanie (approved):
:+1: Tested on android emulator with Cyrillic, Arabic and html url encoding. Works as expected. :rocket: 

:clown_face: Just one thing, when i clicked on the PN, i first saw the "this site does not exist" screen for 1 second, then i was redirected to the correct screen, but I think this is not releated to this issue.

shared/utils/replaceLinks.ts
```diff
@@ -0,0 +1,23 @@
+// Match \w and ' characters
+// https://stackoverflow.com/a/6041965
+// language=RegExp
```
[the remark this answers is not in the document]
steffenkleinle: The comments are necessary for syntax highlighting in the IntelliJ. Will keep it as it is.

steffenkleinle:
> 🤡 Just one thing, when i clicked on the PN, i first saw the "this site does not exist" screen for 1 second, then i was redirected to the correct screen, but I think this is not releated to this issue.

Sadly I can't reproduce that. Please open an issue for this with some more information on how to reproduce.

Issue #2703, opened 2024-02-28
Landing Page exists for `halloaschaffenburg`


sarahsporck:
### Describe the Bug

The landing page can be accessed for `halloaschaffenburg.de`

### Steps to Reproduce

1. Go to 'halloaschaffenburg.de/landing/de'
2. See landing page

### Expected Behavior

Either a NotFoundError or a redirect to halloaschaffenburg.de

### Actual Behavior

The landing page is visible

PR #2704, opened 2024-02-28
2703: Fix landing page visible for halloaschaffenburg

sarahsporck:
### Short description

Do not render landing route for halloaschaffenburg.

### Proposed changes

- Render landing route conditionally
- Remove unecessary redirect

### Resolved issues

Fixes: #2703 

---


web/src/RootSwitcher.tsx
```diff
@@ -84,11 +84,8 @@ const RootSwitcher = ({ setContentLanguage }: RootSwitcherProps): ReactElement =
 
         {/* Redirects */}
         <Route path='/' element={<Navigate to={fixedCityPath ?? landingPath} replace />} />
-        <Route path={LANDING_ROUTE} element={<Navigate to={fixedCityPath ?? landingPath} replace />} />
+        <Route path={RoutePatterns[LANDING_ROUTE]} element={<Navigate to={fixedCityPath ?? landingPath} replace />} />
```
steffenkleinle, line 87: Hmmm... While this works and there is no endless loop, we are basically redirecting from `/landing/de` to `/landing/de` if there is no fixed city. I'd prefer if we could avoid that by only having that redirect if there is a fixed city.
```suggestion
        {!!fixedCityPath && (
          <Route path={RoutePatterns[LANDING_ROUTE]} element={<Navigate to={fixedCityPath} replace />} />
        )}
```

web/src/RootSwitcher.tsx
```diff
@@ -84,11 +84,8 @@ const RootSwitcher = ({ setContentLanguage }: RootSwitcherProps): ReactElement =
...
         <Route path='/:cityCode' element={<Navigate to={fixedCityPath ?? language} replace />} />
```
steffenkleinle, line 88: This now also handles redirecting `/landing`. Perhaps we should make that clear using a comment :see_no_evil:

steffenkleinle (approved):
Tested on firefox :wrench:

LeandraH (approved):
😍  Nice! Tested in Firefox

PR #2714, opened 2024-03-14
2171: Rename 'Hier klicken' Button

LeandraH:
### Short description

This PR renames the button at the bottom of the screen where you choose your region. Just saying "click here" is not very understandable for people who use e.g. a screenreader.

### Resolved issues

Fixes: #2171 

---


native/src/components/CityNotCooperatingFooter.tsx
```diff
@@ -50,7 +50,7 @@ const CityNotCooperatingFooter = ({
     <FooterContainer>
       <StyledIcon Icon={CityNotCooperatingIcon} />
       <Question>{t('cityNotFound')}</Question>
-      <StyledButton text={t('clickHere')} onPress={navigateToCityNotCooperating} />
+      <StyledButton text={t('clickHere', { appName: buildConfig().appName })} onPress={navigateToCityNotCooperating} />
```
steffenkleinle, line 53: :wrench: It would be cool if you could change the translation key as well, e.g. to `suggestToRegion` or similar. Same for other occurrences.

translations/translations.json
```diff
@@ -4794,8 +4794,8 @@
       "refresh": "Liste der nahegelegenen Regionen aktualisieren",
       "pageTitle": "Lokale Informationen für Dich",
       "metaDescription": "{{appName}} ist Dein digitaler Guide für Deutschland. Finde lokale Informationen, Veranstaltungen und Beratung. Immer aktuell und in Deiner Sprache.",
-      "cityNotFound": "Deine Region nicht gefunden?",
-      "clickHere": "Hier klicken"
+      "cityNotFound": "Deine Region nicht gefunden? Werde jetzt aktiv!",
```
steffenkleinle, line 4797: :wrench: I think we should be consistent with the exclamation mark between the different languages. Either we should use it everywhere or nowhere.

steffenkleinle (approved):
:+1: Tested on firefox and android.

translations/translations.json
LeandraH: I did that on purpose because in English, exclamation marks after an imperative are really only for a direct order that someone is yelling. But in German, it's common to have an exclamation mark after an imperative even when it's just a request. So how about I add a period to the English one?

steffenkleinle: Sounds good

translations/translations.json
```diff
@@ -4812,7 +4812,7 @@
       "pageTitle": "ለእርስዎ አካባቢያዊ መረጃ",
       "metaDescription": "{{appName}} በጀርመን ላለው ህይወት የእርስዎ ዲጂታል መመሪያ ነው። አካባቢያዊ መረጃን፣ ክንዉኖችን እና ምክርን ያግኙ። በእርስዎ ቋንቋ ሁልጊዜ የቅርብ ጊዜ የሆነ እና የሚገኝ ነው።",
       "cityNotFound": "ያንተ ኣከባቢ ኣልተገኘምን?",
-      "clickHere": "እዚህ ተጫን"
+      "suggestToRegion": "እዚህ ተጫን"
```
steffenkleinle, line 4815: @osmers will the translators retranslate this next time and add the context if we keep it like this?
We changed the german translation from `Hier klicken` to `{{appName}} in Deiner Region vorschlagen`? Or should we better delete the whole translations?

osmers: As far as I know we only tell them to translate the empty lines, so if we changed something and keep the old translation, then it won't be updated. Having them check all translations costs extra generally. So I would say delete the translation.

LeandraH: Cool, I'll do that, thank you :)

translations/translations.json
```diff
@@ -4811,8 +4811,8 @@
       "refresh": "በቅርብ የሚገኙት ኣከባቢዎች መዝገብ ማሻሻል",
       "pageTitle": "ለእርስዎ አካባቢያዊ መረጃ",
       "metaDescription": "{{appName}} በጀርመን ላለው ህይወት የእርስዎ ዲጂታል መመሪያ ነው። አካባቢያዊ መረጃን፣ ክንዉኖችን እና ምክርን ያግኙ። በእርስዎ ቋንቋ ሁልጊዜ የቅርብ ጊዜ የሆነ እና የሚገኝ ነው።",
-      "cityNotFound": "ያንተ ኣከባቢ ኣልተገኘምን?",
-      "suggestToRegion": "እዚህ ተጫን"
+      "cityNotFound": "",
```
steffenkleinle, line 4814: While the result for translator is probably the same, I think it would be better to just delete the whole properties:
```suggestion
```

steffenkleinle (approved):

ztefanie (approved):
Looks good. :+1:

Issue #2646, opened 2024-01-25
Under iOS the on-screen keyboard slides over the form


Docmale:
### Describe the Bug

Under iOS the on-screen keyboard slides over the Malteser help form so that I can't see what I'm typing. On Android the form moves up as expected

### Steps to Reproduce

1. Go to Malte - Testumgebung - Deutsch
2. Click on [Krieg in der Ukraine] - [Hilfe für Ukrainer und Ukrainerinnen]
3. Scroll down to Help form
4. try to write into fields

### Expected Behavior

The form should move up as in Android

### Actual Behavior

on-screen keyboard slides over the Malteser help form

f1sh1918:
Still does not work for all input fields @LeandraH 
with an additional `<KeyboardAwareScrollView extraScrollHeight={120}>` it works

https://github.com/digitalfabrik/integreat-app/assets/37902063/99ac995e-aafb-4ddd-b159-b7ddcb0c66a1

PR #2715, opened 2024-03-14
2646: Fix keyboard hiding input in help form

LeandraH:
### Short description

I added a `KeyboardAwareScrollView` to the help form so that the keyboard doesn't slide up over the input field anymore.

### Resolved issues

Fixes: #2646 

---


release-notes/unreleased/2646-dont-hide-content-behind-keyboard.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2646
+show_in_stores: true
```
steffenkleinle, line 2: :upside_down_face: since this is only a malte feature and only active in a  few regions, I'd set this to false
```suggestion
show_in_stores: false
```

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -114,56 +114,59 @@ const MalteHelpFormOffer = ({
...
-
-      <FormInput name='name' title={t('name')} control={control} rules={{ required: true }} />
-      <FormInput name='roomNumber' title={t('roomNumber')} control={control} showOptional />
-
-      <View>
-        <InputTitle>{t('howToBeContacted')}</InputTitle>
-        <FormRadioButtons
-          name='contactChannel'
+      {/* @ts-expect-error Android avoids the keyboard automatically */}
+      <KeyboardAvoidingView behavior={Platform.OS === 'ios' ? 'padding' : ''}>
```
steffenkleinle, line 118: :x: Please set it to `undefined` instead of `''` for android to avoid the error.
```suggestion
      <KeyboardAvoidingView behavior={Platform.OS === 'ios' ? 'padding' : undefined}>
```

steffenkleinle (commented):
:+1: Tested on android with `undefined` instead of `''`, works as expected

steffenkleinle (approved):

f1sh1918:
@steffenkleinle this is an ios bug :D
Dunno if i do sth wrong on ios but for me that doesn't fix the issue

https://github.com/digitalfabrik/integreat-app/assets/37902063/d96291d3-feaa-435a-8634-603cbef8494f

steffenkleinle:
> @steffenkleinle this is an ios bug :D 

I know, but I think its still important that it still works on android? :smile:

f1sh1918 (changes requested):
For its not working and i think `position` is the better way to go.
Only tested this on ios not android. And probably  we need an extended padding for the scroll view to be able to scroll more if you focus on last Textarea
```js
<KeyboardAvoidingView behavior={Platform.OS === 'ios' ? 'position' : undefined} keyboardVerticalOffset={-80}>
```
Instead of `-80` the header height should be used

https://github.com/digitalfabrik/integreat-app/assets/37902063/b820fee8-4fad-440c-a401-29cad04beadc

LeandraH:
Arg, that one doesn't work on my iPhone because its screen is too short. I'll try some more.

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -2,6 +2,7 @@ import React, { ReactElement } from 'react'
 import { useForm } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
 import { View } from 'react-native'
+import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
```
steffenkleinle, line 5: :thought_balloon: Last update of the library was 3 years ago, perhaps we should also switch libraries at some point.

LeandraH: Good idea, I opened #2743 for it :)

native/src/components/MalteHelpFormOffer.tsx
```diff
@@ -113,58 +114,60 @@ const MalteHelpFormOffer = ({
...
-      </InformationRow>
-
-      <FormInput name='name' title={t('name')} control={control} rules={{ required: true }} />
-      <FormInput name='roomNumber' title={t('roomNumber')} control={control} showOptional />
-
-      <View>
-        <InputTitle>{t('howToBeContacted')}</InputTitle>
-        <FormRadioButtons
-          name='contactChannel'
+    <KeyboardAwareScrollView extraHeight={200}>
```
f1sh1918, line 117:
```suggestion
    <KeyboardAwareScrollView extraScrollHeight={dimensions.headerHeight + useSafeAreaInsets().top}>
```
Please avoid using magic numbers. To calculate the proper scroll height we should use the header height and the height of the safe area inset (notches etc) of the particular device.
Tested on iPhone15 and iPhone SE3

f1sh1918 (commented):
Please use proper scroll height calculation

f1sh1918 (approved):

Issue #2585, opened 2023-11-22
Hide no search result message after sending feedback


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user I don't want to see the message `Es wurden leider keine passenden Ergebnisse gefunden.` and the sad smiley after I send a feedback what is missing.

**Describe the solution you'd like**
The failure (message + sad smiley) should be hidden after a feedback is sent.

**Describe alternatives you've considered**
Keep showing it.

**Additional context**
Probably you need to pass a callback to `FeedbackContainer` that is called after submitting was successful. A new state variable has to be introduced in the search to manage showing the failure.

Related to #2531.

PR #2717, opened 2024-03-15
2585: Stop showing not found error after feedback is sent

steffenkleinle:
### Short description

Stop showing not found error and sad smiley after feedback is sent.

### Proposed changes

- Move displaying of nothing found to Feedback
- Only display while no feedback sent

### Side effects

None.

### Resolved issues

Fixes: #2585.

---


web/src/components/Feedback.tsx
```diff
@@ -48,9 +49,10 @@ type FeedbackProps = {
   onFeedbackChanged: (isPositiveFeedback: boolean | null) => void
   onSubmit: () => void
   sendingStatus: SendingStatusType
-  searchTerm?: string
+  noResults: boolean | undefined
+  searchTerm: string | undefined
```
LeandraH, line 53: ❓ Why did you switch from `searchTerm?: string` to `searchTerm: string | undefined`?

LeandraH (approved):
👍  Nice! Tested on Firefox and real iOS

web/src/components/Feedback.tsx
steffenkleinle: I think using optional types are sometimes risking not passing anything even though something should be passed. Forcing devs to explicitly pass `undefined` or `null` might prevent bugs. Since `Feedback` is and will only be used once (by `FeedbackContainer`) and `searchTerm` is always passed, I figured this lowers (minimally) the risk of missing this in a future refactoring.

So basically I think the following makes sense as a rule of thumb:
- If the component is used in multiple places and most use cases have the same parameter, an optional property (perhaps with a default) make sense (e.g. an optional property `disabled` that is `false` for most usages)
- If the component is used in only one or two places and it is not clear what the default should be, its better to do it explicitly

LeandraH: Cool, thanks for the explanation :)

ztefanie (approved):
Tested on Android, works as expected :+1: 

Please use `git mv` if you rename files. Makes it easier to review and does not break the history of the file.

steffenkleinle:
> Please use `git mv` if you rename files. Makes it easier to review and does not break the history of the file.

@ztefanie good point, thanks. It doesn't make a difference for git history tracking no matter whether you are using the IDE, `mv` or `git mv` though. The only thing that matters there is that moving/renaming the file is committed separately without any changes done to the content. I normally try to do that, but forgot this time. Will adjust and force push :)

PR #2720, opened 2024-03-18
2635: Use correct thumbnails

steffenkleinle:
### Short description

Use correct thumbnails for pois and fix thumbnail type.

### Proposed changes

- Remove unnecessary `.replace('150x150', '')` for poi thumbnails
- Make thumbnail type optional for categories, events and pois to match api
- Refactor `SimpleImage` to support local svg icons
- Refactor `NavigationTile`

### Side effects

Hopefully none.

### Resolved issues

Fixes: #2635.

---

### DIFF
--- a/native/src/components/NavigationTile.tsx
+++ b/native/src/components/NavigationTile.tsx
@@ -1,16 +1,13 @@
-import * as React from 'react'
-import { ReactNode } from 'react'
+import React, { ReactElement } from 'react'
 import { View } from 'react-native'
 import styled from 'styled-components/native'
 
-import { ThemeType } from 'build-configs'
-
 import TileModel from '../models/TileModel'
-import Icon from './base/Icon'
+import SimpleImage from './SimpleImage'
 import Pressable from './base/Pressable'
 
-const NEWS_DOT_RADIUS = 20
 const ICON_SIZE = 50
+
 const Circle = styled(View)`
   margin-top: 9px;
   margin-bottom: 5px;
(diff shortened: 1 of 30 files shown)


native/src/components/NavigationTile.tsx
```diff
@@ -33,74 +30,30 @@ const TileTitle = styled.Text`
...
 `
 
 type NavigationTileProps = {
   tile: TileModel
-  theme: ThemeType
   width: number
 }
 
-class NavigationTile extends React.Component<NavigationTileProps> {
-  getNewsDot(): ReactNode {
```
ztefanie, line 71: Why delete all this NewsDot logic? I think it is a nice feature, maybe it would make more sense to fix it, than to completely delete it.

native/src/components/NavigationTiles.tsx
```diff
@@ -34,7 +33,7 @@ const StyledIcon = styled(Icon)<{ disabled: boolean }>`
 `
 
 type NavigationTilesProps = {
-  tiles: Array<TileModel>
+  tiles: TileModel[]
```
ztefanie, line 36: :thought_balloon: Maybe we can add a linting rule or something, so we stay consistent with either of them.

native/src/components/SimpleImage.tsx
```diff
@@ -1,46 +1,59 @@
...
 type SimpleImageProps = {
   source: ImageSourceType
   style?: StyleProp<ImageStyle>
   resizeMode?: ImageResizeMode
   // In order to be able to align an image, its width or aspect ratio has to be set
   specifyAspectRatio?: boolean
-}
-
-// For ios you should not use the absolute path, since it can change with a future build version, therefore we use home directory
-// https://github.com/facebook/react-native/commit/23909cd6f62056de0cd0f7c06e3997dd967c139a
```
ztefanie, line 20: Have you tested with ios, so this still works?

native/src/utils/getCachedThumbnail.ts
```diff
@@ -0,0 +1,22 @@
+import { Platform } from 'react-native'
+
+import { URL_PREFIX } from '../constants/webview'
+import { PageResourceCacheStateType } from './DataContainer'
+
+const getCachedThumbnail = (thumbnail: string, resourceCache: PageResourceCacheStateType | undefined): string => {
+  const resource = resourceCache?.[thumbnail]
+
+  if (resource) {
```
ztefanie, line 9: :wrench: You can changes this to 
```
if(!resource){
return thumbnail}
}

    if (Platform.OS === 'ios') {
      return `~${cachedThumbnail.substring(cachedThumbnail.indexOf('/Documents'))}`
    }
    return `${URL_PREFIX}${cachedThumbnail}`
```

Makes it easier to understand, as the nested conditional is avoided.

LeandraH (approved):
👍  Nice! Tested in Firefox and real iOS

native/src/components/SimpleImage.tsx
steffenkleinle: I can't don't have an ios device. @LeandraH could you please test this specifically again, i.e. showing images while offline on ios?

native/src/components/NavigationTile.tsx
steffenkleinle: The feature/implementation is from 2019 and was never used so far while we still had to maintain it. Unused code should be removed, not kept for a theoretical future use. If we actually want to have the feature, we can easily restore it from revision. Therefore I changed the commit history and made a specific commit to remove the news dot logic. Sorry for the force push!

native/src/components/SimpleImage.tsx
LeandraH: Checked, it still works. Except for the back icon in the header but that also disappears on main. I'll check that in the production app and open a ticket if necessary

PR #2721, opened 2024-03-18
2379 poi filter show number of found pois

relentless95:
### Proposed changes

-
-

### Side effects

-
-

### Resolved issues

Fixes: #

---


steffenkleinle:
@relentless95 this is already done in #2611, right? Can we delete the branch?

relentless95:
yes it is.
it was a mistake

Issue #2723, opened 2024-03-18
Endless cycle opening internal links in the InAppBrowser


steffenkleinle:
### Describe the Bug

Opening internal links in the InAppBrowser (happens if the language or city is different than the currently selected) at the moment lead to an endless cycle of the app opening the InAppBrowser which in turn opens the app again.

### Steps to Reproduce

1. If you are using the IntegreatTestCms app on Android > 12, you first have to add `https://integreat.app` to allowed urls in settings
2. Go to https://webnext.integreat.app/testumgebung/de/test-von-links-zu-orten
3. Select the link not matching your current language
4. See cycle

### Expected Behavior

The InAppBrowser is opened and the content is shown. The app is not opened again.

### Actual Behavior

There is an endless cycle of the app opening the InAppBrowser and vice versa. 

### Additional Information

Similar to #2440.

### Testing Information

For testing on android, please make sure to allow `open supported links`, see screenshot.
Please test the behavior on iOS as well.
<img src="https://github.com/digitalfabrik/integreat-app/assets/29095487/f145e2f0-da5e-414a-81b7-46a036af7f30" width="200px" />

PR #2724, opened 2024-03-18
2723: Fix endless loop opening internal links

steffenkleinle:
### Short description

Fix endless loop opening internal links with different city or language

### Proposed changes

Open internal links with `http://` protocol to avoid the InAppBrowser opening the link in the app again. Idea taken from here: https://github.com/proyecto26/react-native-inappbrowser/issues/107#issuecomment-541260993

### Side effects

None. The InAppBrowser still redirects to `https://`.

### Resolved issues

Fixes: #2723.

---


LeandraH (approved):
👍 Looks good. I couldn't reproduce the issue in real iOS but this didn't break it there either.

f1sh1918 (approved):
👍  nice :)
Works fine for android and ios

Issue #2687, opened 2024-02-20
Support language-independent urls for events, news and pois


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to link to specific integreat features without having to specify a language. While this already works for the dashboard/city page (e.g. https://integreat.app/testumgebung), it does not yet work for:
- events (e.g. https://integreat.app/testumgebung/events)
- news (e.g. https://integreat.app/testumgebung/news)
- pois (e.g. https://integreat.app/testumgebung/locations)

**Describe the solution you'd like**
The above urls should be valid integreat urls that are redirected to the users selected language, e.g.
- `https://integreat.app/testumgebung/events` -> `https://integreat.app/testumgebung/<language>/events`
- `https://integreat.app/testumgebung/news` -> `https://integreat.app/testumgebung/<language>/news`
- `https://integreat.app/testumgebung/locations` -> `https://integreat.app/testumgebung/<language>/locations`

On web simple `<Redirect />`s should do the trick. For native, we have to adjust the internal pathname parser to use the fallback language if the second path segment is either `events`, `news` or `locations`.

Share urls don't have to be adjusted, they should continue to be language dependent.

**Describe alternatives you've considered**
Don't implement this.

**Additional context**
https://chat.tuerantuer.org/digitalfabrik/pl/bsttq6b3hfb35jg1g47hnsaqrr

Issue #2683, opened 2024-02-20
Add documentation about reserved slugs


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Reserved slugs should be tracked in the backend, so that e.g. it is impossible to create a region with a reserved slug. So when adding new paths to the router we also need to adjust the backend reserved slugs:
https://github.com/digitalfabrik/integreat-cms/blob/2a123c039bfa699b8be68ae748e23431ddcd2deb/integreat_cms/core/settings.py#L69-L87
To increase awareness for that this should be documented.

**Describe the solution you'd like**
Add documentation, that our routes need to be tracked in the backend. And that new routes require a change in the backend.

**Additional context**
See initial discussion here https://github.com/digitalfabrik/integreat-cms/pull/2391#discussion_r1473281500

PR #2725, opened 2024-03-18
2687: Support language independent urls

steffenkleinle:
### Short description

Support language independent urls for reserved routes such as events, news, pois, offers, search and disclaimer.

### Proposed changes

- Add redirects from language-independent urls (web)
- Adjust InternalPathnameParser to support language-independent urls
- Add documentation about reserved slugs

### Side effects

Hopefully none.

### Resolved issues

Fixes: #2687, #2683.

---

### DIFF
--- a/shared/routes/index.ts
+++ b/shared/routes/index.ts
@@ -81,3 +81,24 @@ export const MAIN_DISCLAIMER_ROUTE: MainDisclaimerRouteType = 'main-disclaimer'
 
 export type NotFoundRouteType = 'not-found'
 export const NOT_FOUND_ROUTE: NotFoundRouteType = 'not-found'
+
+// Adding new routes should be mentioned to the CMS team if the route has to be reserved
+export const RESERVED_TOP_LEVEL_SLUGS = [
+  LANDING_ROUTE,
+  MAIN_DISCLAIMER_ROUTE,
+  NOT_FOUND_ROUTE,
+  CONSENT_ROUTE,
+  LICENSES_ROUTE,
+  CITY_NOT_COOPERATING_ROUTE,
+  JPAL_TRACKING_ROUTE,
+]
+
+// Adding new routes should be mentioned to the CMS team if the route has to be reserved
+export const RESERVED_CITY_CONTENT_SLUGS = [
+  SEARCH_ROUTE,
+  DISCLAIMER_ROUTE,
+  EVENTS_ROUTE,
+  NEWS_ROUTE,
+  POIS_ROUTE,
+  OFFERS_ROUTE,
+]
--- a/web/src/CityContentSwitcher.tsx
+++ b/web/src/CityContentSwitcher.tsx
@@ -1,5 +1,5 @@
 import React, { FunctionComponent, ReactElement, Suspense } from 'react'
-import { Route, Routes, useLocation, useParams } from 'react-router-dom'
+import { Navigate, Route, Routes, useLocation, useParams } from 'react-router-dom'
 
 import {
   CATEGORIES_ROUTE,
@@ -7,6 +7,7 @@ import {
   DISCLAIMER_ROUTE,
   EVENTS_ROUTE,
   MALTE_HELP_FORM_OFFER_ROUTE,
+  NEWS_ROUTE,
   normalizePath,
   OFFERS_ROUTE,
   POIS_ROUTE,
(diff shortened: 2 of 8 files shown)


shared/routes/index.ts
```diff
@@ -81,3 +81,26 @@ export const MAIN_DISCLAIMER_ROUTE: MainDisclaimerRouteType = 'main-disclaimer'
...
+// Changes done to the reserved routes have to be done in the CMS as well:
+// https://github.com/digitalfabrik/integreat-cms/blob/main/integreat_cms/core/settings.py#L69-L87
+export const RESERVED_CITY_CONTENT_SLUGS = [
+  SEARCH_ROUTE,
+  DISCLAIMER_ROUTE,
+  EVENTS_ROUTE,
+  NEWS_ROUTE,
+  POIS_ROUTE,
+  OFFERS_ROUTE,
+]
```
steffenkleinle, line 106: I felt like a comment here would make more sense than a short section in the README everybody will just forget. If you disagree let me know.

web/src/__tests__/CityContentSwitcher.spec.tsx
```diff
@@ -28,43 +29,58 @@ describe('CityContentSwitcher', () => {
...
+
   it.each([{ routeName: SEARCH_ROUTE }, { routeName: DISCLAIMER_ROUTE }, { routeName: POIS_ROUTE }])(
     'should navigate to $routeName route',
     async ({ routeName }) => {
       mockUseLoadFromEndpointWithData(city)
-      const { findByText } = renderRoute(<CityContentSwitcher languageCode={languageCode} />, {
-        routePattern: cityContentPattern,
-        pathname: path(city, routeName),
-      })
+      const { findByText } = renderCityContentSwitcher(routeName)
```
LeandraH, line 52: Any chance of using a `findByRole` here and `getByRole` in the later tests?

web/src/RootSwitcher.tsx
```diff
@@ -10,8 +10,10 @@ import {
   LANDING_ROUTE,
   LICENSES_ROUTE,
   MAIN_DISCLAIMER_ROUTE,
+  NEWS_ROUTE,
```
LeandraH, line 13:
```suggestion
```
❌

web/src/RootSwitcher.tsx
```diff
@@ -37,11 +39,11 @@ const LicensesPage = lazyWithRetry(() => import('./routes/LicensesPage'))
 const RootSwitcher = ({ setContentLanguage }: RootSwitcherProps): ReactElement => {
   const { i18n } = useTranslation()
   const { fixedCity, cityNotCooperating, jpalTracking } = buildConfig().featureFlags
-  const languageCode = useMatch('/:slug/:languageCode/*')?.params.languageCode
+  const { routeParam0, routeParam1, '*': splat } = useMatch('/:routeParam0/:routeParam1/*')?.params ?? {}
   useScrollToTop()
 
   const detectedLanguageCode = i18n.language
-  const language = languageCode ?? detectedLanguageCode
+  const language = routeParam1 ?? detectedLanguageCode
```
LeandraH, line 46: ❓ Doesn't this PR make sure that `routeParam1` exactly isn't necessarily the language?

LeandraH (approved):
👍 Very nice! Tested in Firefox and emulated iOS

web/src/CityContentSwitcher.tsx
```diff
@@ -144,6 +145,13 @@ const CityContentSwitcher = ({ languageCode }: CityContentSwitcherProps): ReactE
 
       {tuNewsEnabled && render(TU_NEWS_ROUTE, TuNewsPage)}
       {tuNewsEnabled && render(TU_NEWS_DETAIL_ROUTE, TuNewsDetailPage)}
+
+      {(localNewsEnabled || tuNewsEnabled) && (
+        <Route
+          path={NEWS_ROUTE}
+          element={<Navigate to={localNewsEnabled ? LOCAL_NEWS_ROUTE : TU_NEWS_ROUTE} replace />}
+        />
+      )}
```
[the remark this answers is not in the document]
steffenkleinle: Too align behavior between native and web. And because I initially thought that we will only support features, not specific slugs (i.e. `news` instead of `news/local`). Since it doesn't hurt and allows for a little shorter urls I would keep it. Do you disagree?

web/src/RootSwitcher.tsx
steffenkleinle: Yes, but the case that it is a feature slug (e.g. `news`) is handled by the redirects in which case this will not make a difference. In every other case it is either
* a correct language, then everything will continue to work as expected
* not a correct language, than the `LanguageNotAvailablePage` will be displayed from which the user can choose the preferred language

web/src/CityContentSwitcher.tsx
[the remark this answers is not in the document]
LeandraH: Nah, sounds good, I was just wondering.

f1sh1918 (approved):
👍 nice implementation
Tested on ios and chrome

web/src/__tests__/CityContentSwitcher.spec.tsx
steffenkleinle: Since the way roles are handled will change in the react-native upgrade I'd rather not :)

PR #2728, opened 2024-03-19
2711-2: Fixes for adding apks

ztefanie:
### Short description
Added some fixes for the release asset upload

### Proposed changes
- added deliverino context and yarn_tool_cache for promote jobs 
- fixed passing pre-release parameter
- fixed setting IOS_RELEASE_ID correctly
- skip e2e asset upload


steffenkleinle (approved):

steffenkleinle:
:question: One question: During promotion the `prerelease` status is only removed from one release, right? Shouldn't we remove them from all older `prereleases`? E.g. we have a weekly beta delivery, we trigger a manual release and therefore have to beta/prereleases, only one loses its prerelease status during promotion.

steffenkleinle:
> > ❓ One question: During promotion the `prerelease` status is only removed from one release, right? Shouldn't we remove them from all older `prereleases`? E.g. we have a weekly beta delivery, we trigger a manual release and therefore have to beta/prereleases, only one loses its prerelease status during promotion.
> 
> I think it is more important to handle the opposite case correctly. We deliver a beta, which has a bug, we do not promote this beta, but release an other beta where the bug is fixed. I think the first beta should definitely stay in "pre-release".

Makes sense, I agree after your explanation. Thanks!
Just to double check: This works correctly? I.e. it removes the prerelease tag of the latest release, not the one of the oldest release?

steffenkleinle (approved):

LeandraH (approved):
👍  Very cool, thanks!

ztefanie:
> > > ❓ One question: During promotion the `prerelease` status is only removed from one release, right? Shouldn't we remove them from all older `prereleases`? E.g. we have a weekly beta delivery, we trigger a manual release and therefore have to beta/prereleases, only one loses its prerelease status during promotion.
> > 
> > 
> > I think it is more important to handle the opposite case correctly. We deliver a beta, which has a bug, we do not promote this beta, but release an other beta where the bug is fixed. I think the first beta should definitely stay in "pre-release".
> 
> Makes sense, I agree after your explanation. Thanks! Just to double check: This works correctly? I.e. it removes the prerelease tag of the latest release, not the one of the oldest release?

This is the code
```
const result = releases.data.find(release => release.tag_name.includes(platform))
  if (result && result.prerelease) {
    console.log('Unset prerelease tag of ', result.tag_name)
    return result.id
  }
```

As I understand the github api, we get all the releases in anti-cronological-order, so yes the first match is the newest release. (It is not the latest release, as a pre-release can never be the latest release).

steffenkleinle:
> > > > ❓ One question: During promotion the `prerelease` status is only removed from one release, right? Shouldn't we remove them from all older `prereleases`? E.g. we have a weekly beta delivery, we trigger a manual release and therefore have to beta/prereleases, only one loses its prerelease status during promotion.
> > > 
> > > 
> > > I think it is more important to handle the opposite case correctly. We deliver a beta, which has a bug, we do not promote this beta, but release an other beta where the bug is fixed. I think the first beta should definitely stay in "pre-release".
> > 
> > 
> > Makes sense, I agree after your explanation. Thanks! Just to double check: This works correctly? I.e. it removes the prerelease tag of the latest release, not the one of the oldest release?
> 
> This is the code
> 
> ```
> const result = releases.data.find(release => release.tag_name.includes(platform))
>   if (result && result.prerelease) {
>     console.log('Unset prerelease tag of ', result.tag_name)
>     return result.id
>   }
> ```
> 
> As I understand (and tested) the github api, we get all the releases in anti-cronological-order, so yes the first match is the newest release. (It is not the latest release, as a pre-release can never be the latest release).

Sounds great! Thanks for explaining :)

Issue #2727, opened 2024-03-19
Search results missing for events and pois


f1sh1918:
### Describe the Bug

In some cases the search does not show all events and pois.

### Steps to Reproduce

1. Go to Kaiserslautern
2. Open search bar
3. Type 'Krab'
4. Check results
5. Delete search text and type again 'Krab' -> results are fine

### Expected Behavior

The correct search results should be shown even for first search after opening search bar

### Actual Behavior

On the first search events and pois are missing

### Additional Infos ###
I could only reproduce it for android but test it also on ios

PR #2729, opened 2024-03-19
2727: Search result pois events missing

f1sh1918:
### Short description

Currently for async search not all search results were found with `minisearch`. The issue is that the indexing was not ready when the query was typed and missing results were not reloaded.

### Proposed changes

- reload search results when indexing has finished

### Side effects

- during typing search query the search results change but if we want it async/non blocking, i think we can't find a better solution

### Resolved issues

Fixes: #2727

### Testing
- Go to a location
- open the search and type in immediately your search query. All results should be shown

### DIFF
--- a/native/src/routes/SearchModal.tsx
+++ b/native/src/routes/SearchModal.tsx
@@ -44,7 +44,7 @@ const SearchModal = ({
   const theme = useContext(ThemeContext)
   const { t } = useTranslation('search')
 
-  const searchResults = useSearch(allPossibleResults, query, 'async')
+  const searchResults = useSearch(allPossibleResults, query, true)
 
   const onClose = (): void => {
     sendTrackingSignal({
--- a/native/src/routes/SearchModal.tsx
+++ b/native/src/routes/SearchModal.tsx
@@ -44,7 +44,7 @@ const SearchModal = ({
   const theme = useContext(ThemeContext)
   const { t } = useTranslation('search')
 
-  const searchResults = useSearch(allPossibleResults, query, true)
+  const searchResults = useSearch(allPossibleResults, query)
 
   const onClose = (): void => {
     sendTrackingSignal({
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -1,5 +1,5 @@
 import MiniSearch from 'minisearch'
-import { useMemo } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 
 export type SearchResult = {
   title: string
@@ -8,8 +8,10 @@ export type SearchResult = {
   path: string
 }
 
-const useSearch = (allPossibleResults: SearchResult[], query: string, mode?: 'async'): SearchResult[] => {
-  const minisearch = useMemo(() => {
+const useSearch = (allPossibleResults: SearchResult[], query: string, asyncMode = false): SearchResult[] => {
+  const [isIndexing, setIsIndexing] = useState<boolean>(asyncMode)
+  const [searchResult, setSearchResult] = useState<SearchResult[]>([])
+  const miniSearch = useMemo(() => {
     const search = new MiniSearch({
       idField: 'path',
       fields: ['title', 'content'],
@@ -20,16 +22,20 @@ const useSearch = (allPossibleResults: SearchResult[], query: string, mode?: 'as
         prefix: true,
       },
     })
-    if (mode === 'async') {
-      search.addAllAsync(allPossibleResults)
+    if (asyncMode) {
+      search.addAllAsync(allPossibleResults).then(() => setIsIndexing(false))
     } else {
       search.addAll(allPossibleResults)
     }
     return search
-  }, [allPossibleResults, mode])
+  }, [allPossibleResults, asyncMode])
 
-  // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
-  return query.length === 0 ? allPossibleResults : minisearch.search(query)
+  useEffect(() => {
+    // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
+    setSearchResult(query.length === 0 ? allPossibleResults : miniSearch.search(query))
+  }, [allPossibleResults, isIndexing, miniSearch, query])
+
+  return searchResult
 }
 
 export default useSearch
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -1,14 +1,9 @@
 import MiniSearch from 'minisearch'
 import { useCallback } from 'react'
 
-import { useLoadAsync } from '../api'
+import { CategoryModel, EventModel, PoiModel, useLoadAsync } from '../api'
 
-export type SearchResult = {
-  title: string
-  thumbnail?: string
-  content: string
-  path: string
-}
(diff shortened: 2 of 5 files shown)


shared/hooks/useSearch.ts
```diff
@@ -1,15 +1,10 @@
...
-  title: string
-  thumbnail?: string
-  content: string
-  path: string
-}
-
-const useSearch = (allPossibleResults: SearchResult[], query: string, mode?: 'async'): SearchResult[] => {
-  const minisearch = useMemo(() => {
+const useSearch = (allPossibleResults: SearchResult[], query: string, asyncMode = false): SearchResult[] => {
+  const [isIndexing, setIsIndexing] = useState<boolean>(asyncMode)
```
f1sh1918, line 5: maybe someone knows a better naming for that

native/src/routes/SearchModal.tsx
```diff
@@ -44,7 +44,7 @@ const SearchModal = ({
   const theme = useContext(ThemeContext)
   const { t } = useTranslation('search')
 
-  const searchResults = useSearch(allPossibleResults, query, 'async')
+  const searchResults = useSearch(allPossibleResults, query, true)
```
LeandraH, line 47: 🔧  I specifically decided against using a random boolean here because of Steffi's clean code talk :)

LeandraH (approved):
😍 Thanks for the quick fix! I couldn't reproduce the original issue, unfortunately, but it looks right. Would be good for someone with an Android device to check as well, though. Checked on emulated Android and real iOS

shared/hooks/useSearch.ts
```diff
@@ -20,16 +22,20 @@ const useSearch = (allPossibleResults: SearchResult[], query: string, mode?: 'as
...
     return search
-  }, [allPossibleResults, mode])
+  }, [allPossibleResults, asyncMode])
 
-  // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
-  return query.length === 0 ? allPossibleResults : minisearch.search(query)
+  useEffect(() => {
+    // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
+    setSearchResult(query.length === 0 ? allPossibleResults : miniSearch.search(query))
+  }, [allPossibleResults, isIndexing, miniSearch, query])
```
steffenkleinle, line 36: This looks a little bit complicated to me. Can't we just use `useLoadAsync` instead of `useMemo`, `useEffect` and two `useState`s?
```ts
const useSearch = (allPossibleResults: SearchResult[], query: string): SearchResult[] => {
  const initializeMiniSearch = useCallback(async () => {
    const search = new MiniSearch({
      idField: 'path',
      fields: ['title', 'content'],
      storeFields: ['title', 'content', 'path', 'thumbnail'],
      searchOptions: {
        boost: { title: 2 },
        fuzzy: true,
        prefix: true,
      },
    })
    await search.addAllAsync(allPossibleResults)
    return search
  }, [allPossibleResults])

  const { data: minisearch } = useLoadAsync(initializeMiniSearch)

  return query.length === 0 || !minisearch ? allPossibleResults : minisearch.search(query)
}
```

native/src/routes/SearchModal.tsx
f1sh1918: yes okay but then i have to check here again a string and set a boolean

shared/hooks/useSearch.ts
```diff
@@ -8,8 +8,10 @@ export type SearchResult = {
   path: string
 }
 
-const useSearch = (allPossibleResults: SearchResult[], query: string, mode?: 'async'): SearchResult[] => {
-  const minisearch = useMemo(() => {
+const useSearch = (allPossibleResults: SearchResult[], query: string, asyncMode = false): SearchResult[] => {
```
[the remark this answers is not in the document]
f1sh1918: good question. I think we should decide for async according to the amount of possible results of a location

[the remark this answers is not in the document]
LeandraH: It doesn't load in web with async

native/src/routes/SearchModal.tsx
steffenkleinle: Then I think something like this would be better:
``` ts
const useSearch = (allPossibleResults: SearchResult[], query: string, mode: 'async' | 'concurrent'): SearchResult[] => {
```
or even
``` ts
const useSearch = (allPossibleResults: SearchResult[], query: string, options: { async: boolean }): SearchResult[] => {
```

f1sh1918: refactored function

shared/hooks/useSearch.ts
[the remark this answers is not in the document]
f1sh1918: you mean a page reload?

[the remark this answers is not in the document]
f1sh1918: couldn't find any issues

f1sh1918:
> 😍 Thanks for the quick fix! I couldn't reproduce the original issue, unfortunately, but it looks right. Would be good for someone with an Android device to check as well, though. Checked on emulated Android and real iOS

checked on android.

shared/hooks/useSearch.ts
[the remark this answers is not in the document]
LeandraH: Yes, I just tested as well, you fixed that :)

steffenkleinle (approved):
On web it flickers if you have a preset query:
http://localhost:9000/kaiserslautern/de/search/?query=krab
I guess this is because we are not waiting for every request to finish in `useAllSearchResults`. Might be done in a separate issue though.

Didn't experience any other issues though. Tested on firefox and android.
However, I was also not able to reproduce the original issue.

f1sh1918:
> On web it flickers if you have a preset query:
> http://localhost:9000/kaiserslautern/de/search/?query=krab
> I guess this is because we are not waiting for every request to finish in `useAllSearchResults` as e.g. in this case for me the pois are ready first, minisearch is searching, doesn't find anything, events are added and then there are results. Might be acceptable though (or be done in a spearate issue).
> 
> Didn't experience any other issues though. Tested on firefox and android.
> However, I was also not able to reproduce the original issue.

Probably that flickering causes the errors on e2e Web Tests :/

steffenkleinle:
> > On web it flickers if you have a preset query:
> > http://localhost:9000/kaiserslautern/de/search/?query=krab
> > I guess this is because we are not waiting for every request to finish in `useAllSearchResults` as e.g. in this case for me the pois are ready first, minisearch is searching, doesn't find anything, events are added and then there are results. Might be acceptable though (or be done in a spearate issue).
> > Didn't experience any other issues though. Tested on firefox and android.
> > However, I was also not able to reproduce the original issue.
> 
> Probably that flickering causes the errors on e2e Web Tests :/
> 
> Update: There are also some type errors in `useAllPossibleSearchResults` that should be refactored
> 
> I don't have time today for this

Looking into it.

steffenkleinle:
@f1sh1918 opened a PR to fix the flickering. Basically the problem occurs because minisearch is now initialized asynchronously. So the search results are available first, but minisearch has to be initialized after and that takes at least one render. During that time, minisearch is initialized with only some of the possible search results (e.g. just pois). Therefore we have to check if `minisearch.documentCount === allSearchResults.length`. See #2730.

Issue #2675, opened 2024-02-15
Disallow nullable numbers in boolean checks


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
I was running into problems while using nullable numbers as conditions:
``` ts
  const x: number | undefined = 0
  if (!x) {
    // This is not true as 0 also evaluates to false
    console.log('x is undefined')
  }
```

**Describe the solution you'd like**
Add the eslint rule strict-boolean-expressions to avoid these checks. This might lead to a lot of errors.

**Describe alternatives you've considered**
Do not add the rule.

**Additional context**
Documentation: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md

PR #2735, opened 2024-03-22
2675: Disallow unsafe number checks

steffenkleinle:
### Short description

Add linting rule strict-boolean-expression to disallow unsafe number checks.

### Proposed changes

- Add linting rule strict-boolean-expression to disallow unsafe number checks.
- Fix errors
- Keep allowing unsafe boolean and string boolean checks

### Side effects

None.

### Resolved issues

Fixes: #2675.

---


.eslintrc.js
```diff
@@ -105,6 +105,13 @@ module.exports = {
     ],
     '@typescript-eslint/no-use-before-define': 'error',
     '@typescript-eslint/prefer-ts-expect-error': 'error',
+    '@typescript-eslint/strict-boolean-expressions': [
+      'error',
+      {
+        allowNullableBoolean: true,
+        allowNullableString: true,
```
steffenkleinle, line 112: This can be discussed. I figured to keep allowing those checks for now as quite many changes would be necessary.
``allowNullableBoolean` allows us to do the following:
``` ts
const myBoolean: boolean | null = null
if (myBoolean) {
  // do stuff
}
```
while ``allowNullableString` allows us to do the following:
``` ts
const myString: string | null = null
if (myString) {
  // do stuff
}
```
Both are potentially unsafe since it treats `null` and `false`/`null` and empty strings `''` the same way. I don't really see a problem for `allowNullableBoolean` but `allowNullableString` might lead to a bug at some point. I still think the risk is acceptable. 

Disallowing `allowNullableBoolean` leads to 89 errors while `allowNullableString` leads to 247 errors.

native/src/components/ExportEventButton.tsx
```diff
@@ -127,7 +127,7 @@ const ExportEventButton = ({ event }: ExportEventButtonType): ReactElement => {
 
   return (
     <>
-      {!!calendars?.length && (
+      {calendars && calendars.length > 1 && (
```
steffenkleinle, line 130: @LeandraH is this correct?

web/src/components/DropDownContainer.tsx
```diff
@@ -0,0 +1,37 @@
+import styled from 'styled-components'
+
+import dimensions from '../constants/dimensions'
+
+const DropDownContainer = styled.div<{ active: boolean; height?: number }>`
```
steffenkleinle, line 5: Just extracted and changed conditions.

native/src/components/MapView.tsx
```diff
@@ -101,8 +101,9 @@ const MapView = ({
   }
 
   const coordinates = selectedFeature?.geometry.coordinates
+  const defaultZoom = coordinates ? normalDetailZoom : defaultViewportConfig.zoom
   const defaultSettings: CameraSettings = {
-    zoomLevel: zoom ?? coordinates ? normalDetailZoom : defaultViewportConfig.zoom,
+    zoomLevel: zoom ?? defaultZoom,
```
LeandraH, line 106: 😍  I really like that you also took the time to refactor things just to make them easier to understand!

native/src/components/RemoteContent.tsx
```diff
@@ -99,7 +99,7 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
         log(message.message, 'warning')
       } else {
         const error = new Error(
-          message.message ? JSON.stringify(message.message) : 'Unknown message received from webview',
+          message.message !== undefined ? JSON.stringify(message.message) : 'Unknown message received from webview',
```
LeandraH, line 102: 💭  Hmmm, if the error message is an empty string, I think I'd find the `Unknown message received from webview` error clearer than the empty string. Though I'm not sure how that would ever happen either.

native/src/components/SimpleImage.tsx
```diff
@@ -8,7 +8,7 @@ import getCachedThumbnail from '../utils/getCachedThumbnail'
 import Icon from './base/Icon'
 
 const StyledImage = styled.Image<{ aspectRatio?: number }>`
-  ${props => props.aspectRatio && `aspect-ratio: ${props.aspectRatio};`}
+  ${props => props.aspectRatio !== undefined && `aspect-ratio: ${props.aspectRatio};`}
```
LeandraH, line 11: In case anybody else wonders what happens with an aspect ratio of 0: The property is ignored.

native/src/routes/Pois.tsx
```diff
@@ -90,7 +90,7 @@ const Pois = ({ pois: allPois, language, cityModel, route, navigation }: PoisPro
   }
 
   const deselect = () => {
-    if (multipoi && slug) {
+    if (multipoi !== undefined && slug) {
```
LeandraH, line 93: 💭 I understand currently why we're treating `multipoi` and `slug` differently here but I'm not sure I will understand that in the future. But I guess the linter will yell at me if I use the check incorrectly.

native/src/utils/DatabaseConnector.ts
```diff
@@ -691,18 +691,19 @@ class DatabaseConnector {
...
+                  id: jsonObject.location.id,
+                  name: jsonObject.location.name,
+                  country: jsonObject.location.country,
+                  address: jsonObject.location.address,
+                  latitude: jsonObject.location.latitude,
+                  longitude: jsonObject.location.longitude,
+                  postcode: jsonObject.location.postcode,
+                  town: jsonObject.location.town,
+                })
+              : null,
```
LeandraH, line 706: 🔧 I'd prefer turning this around.

shared/api/endpoints/createFeedbackEndpoint.ts
```diff
@@ -85,8 +85,8 @@ const getFeedbackType = (routeType: FeedbackRouteType, slug?: string): FeedbackT
   }
 }
 
-export default (baseUrl: string): Endpoint<ParamsType, void> =>
-  new EndpointBuilder<ParamsType, void>(FEEDBACK_ENDPOINT_NAME)
+export default (baseUrl: string): Endpoint<ParamsType, Record<string, never>> =>
+  new EndpointBuilder<ParamsType, Record<string, never>>(FEEDBACK_ENDPOINT_NAME)
```
LeandraH, line 89: ❓ I don't quite understand the never part here. Is that supposed to be overwritten when calling the EndpointBuilder?

LeandraH (approved):
Very nice! Tested in Firefox and emulated iOS

native/src/components/SimpleImage.tsx
steffenkleinle: Hmm I didn't test this now but when I think to recall that when implementing the aspect ratio in the first place setting `aspectRatio` somehow broke stuff. But I might be wrong...

shared/api/endpoints/createFeedbackEndpoint.ts
LeandraH: Ooooh, cool, thanks for the explanation!

f1sh1918 (approved):
Looks fine.
Tested on chrome & android

native/src/utils/DatabaseConnector.ts
steffenkleinle: Changed condition to just `jsonObject.location` instead of `jsonObject.location?.id !== undefined` so it should be better now :)

web/src/components/PoisMobile.tsx
```diff
@@ -93,7 +93,7 @@ const PoisMobile = ({
   const [mapViewRef, setMapViewRef] = useState<MapViewRef | null>(null)
   const { pois, poi, mapFeatures, mapFeature } = data
   const { height } = useWindowDimensions()
-  const canDeselect = mapFeature || !!slug
+  const canDeselect = !!mapFeature || !!slug
```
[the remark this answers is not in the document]
steffenkleinle: This check is enough, changed the one in `PoisDesktop` to match this.

Issue #2699, opened 2024-02-26
Support embedded offer for appointment booking


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user I want to be able to use the Terminbuchungstool (appointment booking tool) as simple as possible and directly from the app.

**Describe the solution you'd like**
We need to support an embedded offer for the appointment booking tool.

~~:warning: Currently waiting for cms as this embedded offer is not existing yet.~~

Acceptance criteria:
- The category with the embedded offer is displayed normally (in this case probably as 1st level tile)
- Clicking on the category with the embedded offer opens the URL in the InAppBrowser (native) or navigates there (web)

**Describe alternatives you've considered**
None.

**Additional context**
This is a feature request by the maltesers for the malte app. It should work in integreat as well though.
Mattermost thread:
https://chat.tuerantuer.org/digitalfabrik/pl/zcrkyb141ff9mnuw4ojc11ba6e

**Testing information**
This can be tested using **malte** at http://localhost:9000/testumgebung/de.
Clicking on the tile `Terminbuchung` should open https://testumgebung.termine.malteapp.de. The website doesnt work though since no instance for `testumgebung` is set up. To see the real tool checkout https://soest.termine.malteapp.de
(please dont enter any data as it is probably production).

**Related Issues**
#2712, #2702

steffenkleinle:
@dkehne is it sufficient to support the appointment booking offer on the first level (tiles)? Or do we need support in the category lists as well?

dkehne:
Tough Question. We justed discussed tiles so if it can be triggered by clicking a first level icon it's sufficient.

But I cannot promise that the requirements can change over time...

steffenkleinle:
> Tough Question. We justed discussed tiles so if it can be triggered by clicking a first level icon it's sufficient.
> 
> But I cannot promise that the requirements can change over time...

That is okay. We can add support for 2nd and 3rd level later if the requirement changes. Thanks :)

PR #2738, opened 2024-03-25
2699: Support appointment booking offer

steffenkleinle:
### Short description

Add support for appointment booking and other external embedded offers on the first level.

### Proposed changes

- Add and use helper to map categories to `TileModel`s which handles external embedded offers 
- Merge and move TileModel to shared
- Add support for external urls in the `Tile` component (native)

### Side effects

- Non-internal *embedded* offers are now all treated as external offers
- Content/children of categories with embedded external offers are not displayed anymore

### Resolved issues

Fixes: #2699.

---


LeandraH (approved):
Very cool! Tested in Firefox and emulated iOS

ztefanie (changes requested):
Tested on Chrome, going to the offers tab (http://localhost:9000/testumgebung/de/offers) throws an error now.

steffenkleinle:
> Tested on Chrome, going to the offers tab (http://localhost:9000/testumgebung/de/offers) throws an error now.

Good catch. Filtering out the offer now, should be fixed.

ztefanie (approved):
hmh now there for malte the offers tab is shown but empty, but I think we preivously had this problem, too, so I will approve.

steffenkleinle:
> hmh now there for malte the offers tab is shown but empty, but I think we preivously had this problem, too, so I will approve.

Should be removed soon anyway so I think this is okay for now.

Issue #2597, opened 2023-11-27
Adjust feedback


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
There are a few things we should/want to change in the feedback.

**Describe the solution you'd like**
Things to change:
- [x] Remove the note instead of just hiding it, such that the submit button moves up (web and native)

- [x] ~~Change the way the query/search term is sent to the cms (proposal by @sarahsporck, to be discussed with the service team) (web and native): Maybe replace 'actual query' with 'ursprüngliche Suche'?~~ The actual query should be removed completely. Instead, only the adjusted search term should be delivered.

- [x] Use LoadingSpinner instead of activity indicator with custom color (native)

- [x] ~~On smaller displays the "Abschicken" button is not visible because of the gap between email-field and button, just remove the gap (native)~~ Already fixed by the first task.

**Describe alternatives you've considered**
Keep things as they are.

**Additional context**
All mentioned in these two refactoring PRs: https://github.com/digitalfabrik/integreat-app/pull/2566 and https://github.com/digitalfabrik/integreat-app/pull/2567

Loosely related to #2585.

PR #2742, opened 2024-03-27
2597: Adjust feedback

steffenkleinle:
### Short description

Several small adjustments to the feedback.

### Proposed changes

- Remove the note instead of just hiding it
- Stop sending the actual query to the cms for search feedback
- Use our own loading spinner

### Side effects

Only the adjusted search term is sent to the CMS, the actual query is not sent anymore. This might lead to confusing feedback for the CMS users.

### Resolved issues

Fixes: #2597.

---


LeandraH (approved):
Nice! Tested in Firefox and emulated iOS

f1sh1918 (changes requested):
When i go to search and type something and no result was found, the feedback form appears, but when i then delete the query which is mandatory. I can still can send the feedback which shouldn't be the case i think?

https://github.com/digitalfabrik/integreat-app/assets/37902063/efa8f77c-f43a-4376-921a-4f02f27e3df8

steffenkleinle:
> When i go to search and type something and no result was found, the feedback form appears, but when i then delete the query which is mandatory. I can still can send the feedback which shouldn't be the case i think?

If the query is empty the search text is sent which is the same behavior as it is atm. If you think this should be changed please open a new issue. Personally I think that is okay.

f1sh1918 (approved):
Alright gonna open an issue. I think we should avoid the possibility to send empty feedback!

Issue #2747, opened 2024-04-01
Event times are wrong after language change


steffenkleinle:
### Describe the Bug

The starting times of events are sometimes wrong after a language change.

### Steps to Reproduce

1. Go to Olpe > de > Veranstaltungen
2. See correct starting times
3. Change language to Ukrainian
4. See wrong starting times

See screen recording attached below. Web seems to work fine.

### Expected Behavior

The starting times should stay correct and not change after language change.

### Actual Behavior

The starting times are displayed incorrectly.

### Additional Information

This might be related to the switch from daylight savings time to summer time in the morning of 31.03.2024 in Germany.

Might be related to reloading from database connector.

https://github.com/digitalfabrik/integreat-app/assets/29095487/ad430744-306f-4c18-8553-250b74b37c17

PR #2748, opened 2024-04-02
2747: Fix wrong event times

steffenkleinle:
### Short description

Fix wrong event times by calculating offset difference from current date.

### Proposed changes

- Always store the start date from the CMS
- Only access start and end date of recurrences
- Invalidate offline content

### Explanation

The current problem was that we set the first recurrence as start date with an adjusted offset calculated from the start date. This adjusted start date is saved to the offline content. Next time we load a start date with already adjusted offset and adjust the offset again leading to wrong times.

### Side effects

The offline content will be deleted.

### Resolved issues

Fixes: #2747.

---

### DIFF
--- a/native/src/utils/DatabaseConnector.ts
+++ b/native/src/utils/DatabaseConnector.ts
@@ -32,7 +32,7 @@ import {
 import { deleteIfExists } from './helpers'
 import { log, reportError } from './sentry'
 
-export const CONTENT_VERSION = 'v5'
+export const CONTENT_VERSION = 'v6'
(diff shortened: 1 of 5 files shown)


steffenkleinle:
This is just a quick and little dirty fix. We'll have to clean this up some time later. I will take care of creating an issue.

f1sh1918 (approved):
Thx for the quick fix.
Works fine 👍 
Tested on ios

ztefanie (approved):
Looks good. Tested on Android emulator, works as expected, could not reproduce the bug

Issue #2658, opened 2024-02-05
Update dependencies


sarahsporck:
**Is your feature request related to a problem? Please describe.**
It should be time to update dependencies again.
Especially updating react-native seems to be very rewarding, as the newest version separates the logic into very consumable parts :)

**Describe the solution you'd like**
Update dependencies. Preferably do the react-native update separately: #2745.
Also take a look at our security vulnerabilities and if we can get rid of some of them.

**Testing information**
Please test everything (the whole app) very thoroughly!

**Additional context**
https://react-native-community.github.io/upgrade-helper/?from=0.71.12&to=0.73.3

PR #2750, opened 2024-04-03
2658: Upgrade dependencies

steffenkleinle:
### Short description

Upgrade dependencies

### Proposed changes

Upgrade most dependencies that are not blocked by a react-native upgrade and fix their usages.

### Blocked

- @dr.pogodin/react-native-static-server (starting [v0.8](https://github.com/birdofpreyru/react-native-static-server/releases/tag/v0.8.0)): RN 0.72 and minimum SDK 28 
- react-native-firebase (starting [v19](https://github.com/invertase/react-native-firebase/blob/main/packages/app/CHANGELOG.md#1900-2024-02-26)): CocoaPods 1.12+, done in #2757.
- @math.gl/web-mercator: Dependency not found?
- ~~react-native-webview: TS is broken starting v13.8.3: https://github.com/react-native-webview/react-native-webview/issues/3374~~: Fixed in v13.8.6
- ~~react-navigation-header-buttons: node >= v20 required starting v11.2.0~~
- decamelize, flat: Only ESM modules available
- eslint: Lots of breaking changes in v9.0.0. Might be done in a separate issue once bugs are fixed.

### Unmaintained

- react-native-system-setting (last update 2020)
- react-native-highlight-words (last update 2018)

### Side effects

This PR is not meant as a standalone PR. Therefore you DON'T have to test this, code review is enough. This should be merged together with #2757.

### Resolved issues

Fixes: #2658.

---


web/src/components/CityEntry.tsx
```diff
@@ -31,50 +31,44 @@ type CityEntryProps = {
...
 }
 
-class CityEntry extends React.PureComponent<CityEntryProps> {
-  getMatchedAliases = (city: CityModel, normalizedFilter: string): Array<string> => {
-    if (city.aliases && normalizedFilter.length >= 1) {
-      return Object.keys(city.aliases).filter(alias => normalizeString(alias).includes(normalizedFilter))
-    }
-    return []
-  }
+const CityEntry = ({ filterText, city, language }: CityEntryProps): ReactElement => {
```
steffenkleinle, line 36: Only converted from class to function component and use `useTheme` hook.

web/src/components/CleanLink.tsx
```diff
@@ -22,12 +22,13 @@ const CleanLink = ({ to, children, ariaLabel, className, newTab }: CleanLinkProp
   const newTabProps = newTab && { target: '_blank', rel: 'noopener noreferrer' }
   if (isExternalUrl(to)) {
     return (
+      // @ts-expect-error wrong types from polymorphic 'as', see https://github.com/styled-components/styled-components/issues/4112
```
steffenkleinle, line 25: Let me know if I should rather downgrade styled-components again.

native/src/components/SearchListItem.tsx
```diff
@@ -45,7 +45,6 @@ const HighlighterCategoryTitle = styled(Highlighter)<{ language: string }>`
   flex-direction: ${props => contentDirection(props.language)};
   font-family: ${props => props.theme.fonts.native.decorativeFontRegular};
   color: ${props => props.theme.colors.textColor};
-  font-weight: bold;
```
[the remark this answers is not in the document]
LeandraH: I believe Sarah requested that the title be bold as well. And I think it does look better that way, to make it clearer what the title of the page is that you're seeing a result from.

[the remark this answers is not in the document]
steffenkleinle: Okay, I'll readd it. The test were broken then though :man_shrugging:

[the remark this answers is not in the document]
LeandraH: Huh, weird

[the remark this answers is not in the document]
steffenkleinle: Also I think this behavior is not matched in web, is it?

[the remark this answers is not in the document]
LeandraH: True, the title in web isn't bold, it's larger to make the distinction.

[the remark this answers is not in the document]
LeandraH: I'm also okay with removing the bold titles again, though I'm not sure this is the right ticket for it :)

.circleci/src/jobs/check.yml
```diff
@@ -17,7 +17,7 @@ steps:
       command: yarn prettier:check
   - run:
       name: TS check
-      command: yarn ts:check
+      command: node --max-old-space-size=1536 ./node_modules/.bin/tsc --build
```
steffenkleinle, line 20: I ran into memory problems with `ts:check` so I had to increase the node js heap size.

native/src/components/SearchListItem.tsx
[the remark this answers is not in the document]
steffenkleinle: Okay, reverted and skipping the tests now. Created https://github.com/digitalfabrik/integreat-app/issues/2764.

steffenkleinle:
Could someone please upgrade the iOS dependencies? Thanks :pray:

.circleci/src/jobs/check.yml
ztefanie: So can you please create a command in package.json for this, similar to "lint:ci" we chould introduce a "ts:check:ci"

.husky/pre-commit
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
```
ztefanie, line 3: Why have you removed this? Or what was this for in the first place?

e2e-tests/native/wdio-browserstack.conf.ts
```diff
@@ -7,7 +7,7 @@ const getCapability = () => {
   }
   const capability = capabilities[capabilityName]
 
-  if (!capability) {
+  if (capability === undefined) {
```
ztefanie, line 10: :upside_down_face: Wouldn't it (although it should never be null) more robust if we stay with checking like before?

native/src/components/__tests__/SearchListItem.spec.tsx
```diff
@@ -69,7 +69,7 @@ describe('SearchListItem', () => {
     assertHighlighting(getByText(contentWithoutHtml), false)
   })
 
-  it('should show excerpt around query if only match in content', () => {
+  it.skip('should show excerpt around query if only match in content', () => {
```
ztefanie, line 72: Why do you skip them? Have you created an issue where we reactivate them?

web/src/components/CityEntry.tsx
ztefanie: We have the `<ThemeProvider>` and the `useTheme` hook. As our WHoe app is wrapped in the <ThemeProvider>, why do we still need/use the hook?

web/src/testing/render.tsx
```diff
@@ -17,14 +17,14 @@ type RenderRouteOptions = {
...
-const AllTheProviders = ({ children, options }: { children: ReactElement; options?: { pathname: string } }) => (
+const AllTheProviders = ({ children, options }: { children: ReactNode; options?: { pathname: string } }) => (
   <MemoryRouter initialEntries={options ? [options.pathname] : ['/']}>
     <ThemeProvider theme={theme}>{children}</ThemeProvider>
   </MemoryRouter>
 )
 
 export const renderWithRouterAndTheme = (ui: ReactElement, options?: { pathname: string }): RenderResult =>
-  render(ui, { wrapper: (props: { children: ReactElement }) => <AllTheProviders {...props} options={options} /> })
+  render(ui, { wrapper: (props: { children: ReactNode }) => <AllTheProviders {...props} options={options} /> })
```
ztefanie, line 27: Why would we allow ReactNodes here? "ReactElements" make more sense, don't they?

ztefanie (approved):
Thanks for this PR :rocket: 
Looks good!

.husky/pre-commit
steffenkleinle: See https://github.com/typicode/husky/releases/tag/v9.0.1 > How to migrate

web/src/components/CityEntry.tsx
steffenkleinle: Sorry but I don't understand the question :see_no_evil: Can you try again?

If you are asking why we need to use the hook/need the theme in this case, see line 63. We have to pass the theme color there.

f1sh1918 (commented):
I had to remove and recreate `podfile.lock` because of hermes and sentry issues during `pod install`. Not sure if that may lead to issues in the pipeline and cache version has to be increased

Facing this flipper issue and can not build on ios
https://stackoverflow.com/questions/63376471/react-native-build-error-on-ios-typedef-redefinition-with-different-types-ui
Flipper may be removed but couldn't get this work locally

I don't want to block this issue during my vacation thats why i don't use changes requested for now @steffenkleinle

f1sh1918 (approved):
Tested on web. works fine

web/src/testing/render.tsx
steffenkleinle: This is the type required for @testing-library and I could not get it to pass type checking otherwise. But feel free to come up with another solution. I think this is very minor priority since its only tests anyway...

Issue #2719, opened 2024-03-18
Fix missing API Declaration


ztefanie:
**Is your feature request related to a problem? Please describe.**
We need to fix Missing API declaration for apple uploads. Must be merged till **01.05.2024** or we will not be able to publish Integerat on Apple.

**Describe the solution you'd like**
Include a NSPrivacyAccessedAPITypes array in your app’s privacy 
manifest to provide approved reasons for these APIs used by your app’s code

**Additional context**

> *ITMS-91053: Missing API declaration* - Your app’s code in the “Integreat” file 
references one or more APIs that require reasons, including the following API 
categories: NSPrivacyAccessedAPICategoryDiskSpace. While no action is required 
at this time, starting May 1, 2024, when you upload a new app or app update, you 
must include a NSPrivacyAccessedAPITypes array in your app’s privacy manifest to 
provide approved reasons for these APIs used by your app’s code. For more 
details about this policy, including a list of required reason APIs and approved 
reasons for usage, visit: 
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.

> *ITMS-91053: Missing API declaration* - Your app’s code in the “Integreat” file 
references one or more APIs that require reasons, including the following API 
categories: NSPrivacyAccessedAPICategoryFileTimestamp. While no action is 
required at this time, starting May 1, 2024, when you upload a new app or app 
update, you must include a NSPrivacyAccessedAPITypes array in your app’s privacy 
manifest to provide approved reasons for these APIs used by your app’s code. For 
more details about this policy, including a list of required reason APIs and 
approved reasons for usage, visit: 
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.

> *ITMS-91053: Missing API declaration* - Your app’s code in the “Integreat” file 
references one or more APIs that require reasons, including the following API 
categories: NSPrivacyAccessedAPICategorySystemBootTime. While no action is 
required at this time, starting May 1, 2024, when you upload a new app or app 
update, you must include a NSPrivacyAccessedAPITypes array in your app’s privacy 
manifest to provide approved reasons for these APIs used by your app’s code. For 
more details about this policy, including a list of required reason APIs and 
approved reasons for usage, visit: 
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.

[...]

PR #2760, opened 2024-04-15
2719: Fix missing api declarations

f1sh1918:
### Short description

Apple wants all  accessed apis to be declared

### Proposed changes

- add an array of values for the different api types according to docs. See additional info

### Resolved issues

Fixes: #2719

### Testing
Please properly check if the reasons are fine. I'm not really sure about them.
Find information here:
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

### DIFF
--- a/native/ios/Integreat/Info.plist
+++ b/native/ios/Integreat/Info.plist
@@ -81,7 +81,7 @@
 				<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
 				<key>NSPrivacyAccessedAPITypeReasons</key>
 				<array>
-					<string>3B52.1</string>
+					<string>C617.1</string>
 				</array>
 			</dict>
 			<dict>


native/ios/Integreat/Info.plist
```diff
@@ -66,6 +66,41 @@
...
+                <string>E174.1</string>
+                </array>
+            </dict>
+            <dict>
+                <key>NSPrivacyAccessedAPIType</key>
+                <string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+                <key>NSPrivacyAccessedAPITypeReasons</key>
+                <array>
+                <string>3B52.1</string>
+                </array>
```
LeandraH, line 85: ❓ Why do we actually access timestamps? Is that for the cache?

LeandraH (approved):
Nice, thank you! I'll fix the formatting in a sec

native/ios/Integreat/Info.plist
f1sh1918: no idea as i mentioned. i guess its for offline data

steffenkleinle: I think this should be `C617.1`, i,e
> Declare this reason to access the timestamps, size, or other metadata of files inside the app container, app group container, or the app’s CloudKit container.

If it is about the offline data (which I strongly assume as well) it refers to files inside the app container and NOT files the user granted access to (which we don't have at all in Integreat).

native/ios/Integreat/Info.plist
```diff
@@ -66,6 +66,41 @@
...
 	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
 	<string>Integreat wants to use your location to find nearby cities.</string>
+	<key>NSPrivacyAccessedAPITypes</key>
+		<array>
+			<dict>
+				<key>NSPrivacyAccessedAPIType</key>
+				<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+				<key>NSPrivacyAccessedAPITypeReasons</key>
+				<array>
+					<string>E174.1</string>
```
steffenkleinle, line 76: Okay all of these are really hard...
I think this might be a problem:
> The app must behave differently based on disk space in a way that is observable to users.

I think this is not the case for us (except perhaps the app crashing or not working if there is no space anymore). However, I still agree that this is probably the right category.

native/ios/Integreat/Info.plist
```diff
@@ -66,6 +66,41 @@
...
+				<array>
+					<string>3B52.1</string>
+				</array>
+			</dict>
+			<dict>
+				<key>NSPrivacyAccessedAPIType</key>
+				<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+				<key>NSPrivacyAccessedAPITypeReasons</key>
+				<array>
+					<string>35F9.1</string>
```
steffenkleinle, line 92: Not sure if this is correct. I don't see a single case where we access the boot time... Perhaps even for bug reports?

native/ios/Integreat/Info.plist
```diff
@@ -66,6 +66,41 @@
...
+				</array>
+			</dict>
+			<dict>
+				<key>NSPrivacyAccessedAPIType</key>
+				<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+				<key>NSPrivacyAccessedAPITypeReasons</key>
+				<array>
+					<string>CA92.1</string>
```
steffenkleinle, line 100: :+1:

steffenkleinle (commented):
Hmmm... I am not sure if we are using any of the APIs directly or whether that is all just 3rd party libraries/react-native/hermes directly.

According to https://github.com/react-native-community/discussions-and-proposals/discussions/776 react-native will automatically create a privacy manifest during `pod install` in upcoming releases. So we might actually just need to upgrade to a later react-native version once that is added.

For now we might use something like https://github.com/Wooder/ios_17_required_reason_api_scanner/tree/main to check whether we are directly using such an API.

native/ios/Integreat/Info.plist
f1sh1918: i mean its requested. probably due to writing files the libraries somehow use this before writing on local device?!

f1sh1918:
> https://github.com/Wooder/ios_17_required_reason_api_scanner/tree/main

not sure which folder i should scan @steffenkleinle ?

LeandraH:
Since we need to get this out before May, it might be a good idea to proceed with the API declarations as they are done here and remove them in a future ticket in which we upgrade to the relevant React Native version. What do you think?

f1sh1918:
Can I close this PR @steffenkleinle ?

steffenkleinle:
I think so, yes. At least afaics the last release was promoted successfully. So seems like it was fixed with #2772.

Issue #2465, opened 2023-09-23
Add e2e test for push notifications


ztefanie:
**Is your feature request related to a problem? Please describe.**
Push Notifications are having quite a lot of problems recently, we want to make them more reliable.

**Describe the solution you'd like**
Add an e2e-test (for android and ios) that should check if
- the correct permissions are requested
- the correct city and language is subscribed (also after changing a city or language)

**Describe alternatives you've considered**
A more complex, but also more flaky way would be to really send a push notification and check if it is received. But we decided this would be not a really good solution.

**Additional context**
Relates: #2438 #2451 #2464

PR #2761, opened 2024-04-17
2465: Push Notifications in E2e Tests

LeandraH:
### Short description

The plan for this PR was to check that the app requests permission to send notifications, and to check if the app subscribes to the correct location and language. Unfortunately, I didn’t manage to access the console logs in the e2e environment so this PR only enables notifications in e2e and accepts the request. That makes it a bit smaller than I was hoping to do but it means that the tests will fail if the permission to send notifications is not requested. That should catch some of the notification problems.

### Proposed changes

- Enable push notifications in the e2e environment
- Allow push notifications during the build
- Completely wipe iOS devices before each test because the permissions in iOS aren't reset with a new installation (they are reset in Android when you re-install an app)

### Side effects

- The iOS tests are even slower because the device needs to be wiped between test runs

### Resolved issues

Fixes: #2465 

---


e2e-tests/native/wdio-ios.conf.ts
```diff
@@ -10,6 +10,7 @@ const iosCapabilities: Capabilities.DesiredCapabilities = {
   'appium:orientation': 'PORTRAIT',
   'appium:automationName': 'XCUITest',
   'appium:language': 'EN',
+  'appium:fullReset': true,
```
LeandraH, line 13: A second installation of an app in iOS doesn't reset the permissions; without this full reset, only one in the four tests would request permission to send notifications.

steffenkleinle: :wrench: Please add a comment in the code to explain this :)

steffenkleinle:
> The plan for this PR was to check that the app requests permission to send notifications, and to check if the app subscribes to the correct location and language. Unfortunately, I didn’t manage to access the console logs in the e2e environment so this PR only enables notifications in e2e and accepts the request. That makes it a bit smaller than I was hoping to do but it means that the tests will fail if the permission to send notifications is not requested. That should catch some of the notification problems.

Did you try to actually test/assert on push notifications? It seems like this should be possible according to some blogposts (e.g. this one for android). Or was it just too much effort?

LeandraH:
> > The plan for this PR was to check that the app requests permission to send notifications, and to check if the app subscribes to the correct location and language. Unfortunately, I didn’t manage to access the console logs in the e2e environment so this PR only enables notifications in e2e and accepts the request. That makes it a bit smaller than I was hoping to do but it means that the tests will fail if the permission to send notifications is not requested. That should catch some of the notification problems.
> 
> Did you try to actually test/assert on push notifications? It seems like this should be possible according to some blogposts (e.g. this one for android). Or was it just too much effort?

I spent an hour or two researching it but didn't actually try. I also came to the conclusion that it is possible to test notifications directly but that it would be fragile because of secrets changing and the CMS resetting, so I decided to not add those. That is also what Steffi wrote in the issue description ("A more complex, but also more flaky way would be to really send a push notification and check if it is received. But we decided this would be not a really good solution.") but I have to confess that I don't remember that discussion.

steffenkleinle:
> > > The plan for this PR was to check that the app requests permission to send notifications, and to check if the app subscribes to the correct location and language. Unfortunately, I didn’t manage to access the console logs in the e2e environment so this PR only enables notifications in e2e and accepts the request. That makes it a bit smaller than I was hoping to do but it means that the tests will fail if the permission to send notifications is not requested. That should catch some of the notification problems.
> > 
> > 
> > Did you try to actually test/assert on push notifications? It seems like this should be possible according to some blogposts (e.g. this one for android). Or was it just too much effort?
> 
> I spent an hour or two researching it but didn't actually try. I also came to the conclusion that it is possible to test notifications directly but that it would be fragile because of secrets changing and the CMS resetting, so I decided to not add those. That is also what Steffi wrote in the issue description ("A more complex, but also more flaky way would be to really send a push notification and check if it is received. But we decided this would be not a really good solution.") but I have to confess that I don't remember that discussion.

Thanks for explaining, sounds good to me.

steffenkleinle (approved):

f1sh1918 (approved):
finally we have e2e tests for pn. Nicely done :)

Issue #2674, opened 2024-02-15
Embed declaration on accessibility (DOA) in integreat (app and web)


dkehne:
**Is your feature request related to a problem? Please describe.**
Municipalities are forced from 2026 to embed a DOA into their web-offers (Erkärung der Barrierefreiheit EZB). 

**Describe the solution you'd like**
Embed the DOA the same way as the Data Privacy Declaration.

This means:
Add a new link to the footer linking to https://integreat-app.de/barrierefreiheit/. The text should read `Barrierefreiheit` and be added to translations.json.

sarahsporck:
This should only be done for integreat?

dkehne:
Yes for Integreat only at the moment. Aschaffenburg needs possibly another declaration. Malte is not forced to have it.

steffenkleinle:
> Yes for Integreat only at the moment. Aschaffenburg needs possibly another declaration. Malte is not forced to have it.

@dkehne would it hurt just having it for malte and aschaffenburg as well?

dkehne:
No, would not hurt at all.

PR #2762, opened 2024-04-17
2674: Add accessibility declaration to the footer

LeandraH:
### Short description

Adding the accessibility declaration to the footer for Integreat.

### Proposed changes

- Adding the accessibility declaration to the build config for Integreat. We probably need a separate one for Aschaffenburg at some point, and Malte doesn't need one.

### Resolved issues

Fixes: #2674 

---

### DIFF
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -38,7 +38,7 @@ const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFoote
   const { t } = useTranslation(['layout', 'settings'])
   const aboutUrl = aboutUrls[language] || aboutUrls.default
   const privacyUrl = privacyUrls[language] || privacyUrls.default
-  const accessibilityUrl: string | undefined = accessibilityUrls?.[language] ?? accessibilityUrls?.default
+  const accessibilityUrl = accessibilityUrls?.[language] ?? accessibilityUrls?.default
   const disclaimerPath = pathnameFromRouteInformation({
     route: DISCLAIMER_ROUTE,
     cityCode: city,
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -34,10 +34,11 @@ type CityContentFooterProps = {
 }
 
 const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFooterProps): ReactElement => {
-  const { aboutUrls, privacyUrls } = buildConfig()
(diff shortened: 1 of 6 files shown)


ztefanie (approved):
Looks good :rocket: 

Tested on Integreat, with and without adding an englisch version, also tested malte. All works as expected :+1:

web/src/components/CityContentFooter.tsx
```diff
@@ -34,10 +34,11 @@ type CityContentFooterProps = {
 }
 
 const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFooterProps): ReactElement => {
-  const { aboutUrls, privacyUrls } = buildConfig()
+  const { aboutUrls, privacyUrls, accessibilityUrls } = buildConfig()
   const { t } = useTranslation(['layout', 'settings'])
   const aboutUrl = aboutUrls[language] || aboutUrls.default
   const privacyUrl = privacyUrls[language] || privacyUrls.default
+  const accessibilityUrl: string | undefined = accessibilityUrls?.[language] ?? accessibilityUrls?.default
```
steffenkleinle, line 41: :wrench: Is the type actually necessary?
```suggestion
  const accessibilityUrl = accessibilityUrls?.[language] ?? accessibilityUrls?.default
```

web/src/components/GeneralFooter.tsx
```diff
@@ -13,18 +13,20 @@ type GeneralFooterProps = {
 }
 
 const GeneralFooter = ({ language }: GeneralFooterProps): ReactElement => {
-  const { aboutUrls, privacyUrls } = buildConfig()
+  const { aboutUrls, privacyUrls, accessibilityUrls } = buildConfig()
   const { t } = useTranslation('layout')
 
   const aboutUrl = aboutUrls[language] || aboutUrls.default
   const privacyUrl = privacyUrls[language] || privacyUrls.default
+  const accessibilityUrl: string | undefined = accessibilityUrls?.[language] ?? accessibilityUrls?.default
```
steffenkleinle, line 21:
```suggestion
  const accessibilityUrl = accessibilityUrls?.[language] ?? accessibilityUrls?.default
```

steffenkleinle (approved):
Not tested, but looks good :+1:

web/src/components/CityContentFooter.tsx
LeandraH: You're right, it's not necessary. The linter was complaining about the missing types but that must have been at a different version, it's not complaining now 🤷

Issue #2535, opened 2023-10-25
Check alt tags for images


LeandraH:
**Is your feature request related to a problem? Please describe.**
We should have an alt tag for all images that we show, otherwise a screenreader might e.g. read out the server address that the image is saved at. If the image doesn't contain useful information, we should have an empty string as alt tag.

**Describe the solution you'd like**
- If an image in the remote content doesn't have an alt tag, we should give it an empty string.
- For our icons on buttons, e.g. the search one, we should have an empty alt tag and handle the description of the action in the button around it, so a screenreader should read "search" for the button and not "magnifying glass" for the icon.
- If we have other icons that are not buttons (I can't think of any), we should think about whether they have a function that we want a screenreader to say. Maybe testing with a screenreader would be a good idea for that.

svenseeberg:
https://www.w3.org/WAI/tutorials/images/decorative/ provides more context.

svenseeberg:
> * Maybe testing with a screenreader would be a good idea for that.

I just tested Pericles and did not have any problems in the web app.

PR #2765, opened 2024-04-22
2535: Alt tags for icons

LeandraH:
### Short description

The main task in this issue was to check if we apply alt tags to images that come from the CMS without a manually supplied alt tag. Those already get one applied in the backend. However, the default behaviour of using an image in the CMS is that the image is added as a link to the image alone; if there is no alt text provided, then the screenreader will read out only the link address which is not great UX but there’s not much we can do about that.

The other task was to check on the icons we use in the frontend. I only found that the back-icon (in the header) isn't described well in iOS, and the buttons that handle the left- / right-scrolling in the navigation didn't have a label when disabled.

### Proposed changes

- Added an accessbilityLabel to the back button in the header
- Added an accessibilityLabel to the buttons for scrolling left and right in the nav bar

### Side effects

- In the translations file, I removed the field back from the events folder. I couldn't find it being used anywhere but I might have missed something.

### Resolved issues

Fixes: #2535 
Fixes: #2174 

---


ztefanie (approved):
Looks good. :yellow_heart:

translations/translations.json
```diff
@@ -490,7 +490,10 @@
       "close": "Schließen",
       "showMore": "Mehr anzeigen",
       "showLess": "Weniger anzeigen",
-      "optional": "optional"
+      "optional": "optional",
+      "back": "Zurück",
+      "scrollLeft": "Nach links",
+      "scrollRight": "Nach rechts"
```
steffenkleinle, line 496: :wrench: I think this is wrong on RTL layouts. Better would be something like `previous` and `next` I guess.

steffenkleinle:
> The main task in this issue was to check if we apply alt tags to images that come from the CMS without a manually supplied alt tag. Those already get one applied in the backend. However, the default behaviour of using an image in the CMS is that the image is added as a link to the image alone; if there is no alt text provided, then the screenreader will read out only the link address which is not great UX but there’s not much we can do about that.

Can't this behavior be changed in the CMS? Shouldn't we open an issue there then? Also we could set an empty alt tag is none is set to prevent that, right?

LeandraH:
> > The main task in this issue was to check if we apply alt tags to images that come from the CMS without a manually supplied alt tag. Those already get one applied in the backend. However, the default behaviour of using an image in the CMS is that the image is added as a link to the image alone; if there is no alt text provided, then the screenreader will read out only the link address which is not great UX but there’s not much we can do about that.
> 
> Can't this behavior be changed in the CMS? Shouldn't we open an issue there then? Also we could set an empty alt tag is none is set to prevent that, right?

The empty alt tag is set in the CMS already, I don't think we need to implement that a second time.

LeandraH:
Arg, mis-clicked, sorry

LeandraH:
> > The main task in this issue was to check if we apply alt tags to images that come from the CMS without a manually supplied alt tag. Those already get one applied in the backend. However, the default behaviour of using an image in the CMS is that the image is added as a link to the image alone; if there is no alt text provided, then the screenreader will read out only the link address which is not great UX but there’s not much we can do about that.
> 
> Can't this behavior be changed in the CMS? Shouldn't we open an issue there then?

Hmm, I don't really have an idea of what to change it to. I guess we could not turn the images into links but I assume that was a requested feature.

steffenkleinle:
> > > The main task in this issue was to check if we apply alt tags to images that come from the CMS without a manually supplied alt tag. Those already get one applied in the backend. However, the default behaviour of using an image in the CMS is that the image is added as a link to the image alone; if there is no alt text provided, then the screenreader will read out only the link address which is not great UX but there’s not much we can do about that.
> > 
> > 
> > Can't this behavior be changed in the CMS? Shouldn't we open an issue there then?
> 
> Hmm, I don't really have an idea of what to change it to. I guess we could not turn the images into links but I assume that was a requested feature.

Sorry but I am really confused now. Is there now an issue with missing/wrong alt tags or not? What is the current situation/behavior in both the apps and the cms?

LeandraH:
Sorry, when using an image from the Medienbibliothek in the CMS, the default behaviour is that the image is added as a link in the HTML that we get, so something like `<a href="https://integreat-test.tuerantuer.org/media/global/2023/01/3102_4803_1_g.jpg"></a>`. This example is read by the screenreader as "dreihundertzehnmillionenzweihundertachtundvierzigtausendeinunddreißig Gramm JPG Link". So it reads the last part of the link url, which is the name of the file that is uploaded. That's not the greatest but I don't have a better idea. It's also only a problem if the file doesn't have a clear name. Maybe we could discuss that at the conference, but I'm not sure there is an easy solution to this.

steffenkleinle:
> Sorry, when using an image from the Medienbibliothek in the CMS, the default behaviour is that the image is added as a link in the HTML that we get, so something like `<a href="https://integreat-test.tuerantuer.org/media/global/2023/01/3102_4803_1_g.jpg"></a>`. This example is read by the screenreader as "dreihundertzehnmillionenzweihundertachtundvierzigtausendeinunddreißig Gramm JPG Link". So it reads the last part of the link url, which is the name of the file that is uploaded. That's not the greatest but I don't have a better idea. It's also only a problem if the file doesn't have a clear name. Maybe we could discuss that at the conference, but I'm not sure there is an easy solution to this.

I think we should set an `aria-label` on the link in this case (preferably in the CMS) to something like `Open image` or similar. This should definitely be better than reading random links/link segments. That should fix the problem, right?

steffenkleinle:
> > > Sorry, when using an image from the Medienbibliothek in the CMS, the default behaviour is that the image is added as a link in the HTML that we get, so something like `<a href="https://integreat-test.tuerantuer.org/media/global/2023/01/3102_4803_1_g.jpg"></a>`. This example is read by the screenreader as "dreihundertzehnmillionenzweihundertachtundvierzigtausendeinunddreißig Gramm JPG Link". So it reads the last part of the link url, which is the name of the file that is uploaded. That's not the greatest but I don't have a better idea. It's also only a problem if the file doesn't have a clear name. Maybe we could discuss that at the conference, but I'm not sure there is an easy solution to this.
> > 
> > 
> > I think we should set an `aria-label` on the link in this case (preferably in the CMS) to something like `Open image` or similar. This should definitely be better than reading random links/link segments. That should fix the problem, right?
> 
> The aria-label would then indeed only read out that label but I'm not sure it's that easy. Do users want to know that they can open a link there to an image that doesn't even have an alt-text? I'll ask tomorrow in the Tech Sync if anybody has any ideas for how to solve this.

Hmmm, probably we'd need an accessibility expert on what would be the right thing to do. In my opinion a descriptive `Open image` I can just ignore is better than reading a random url/file name I have no idea what that means (and takes potentially way longer). There might be better options though.

LeandraH:
@steffenkleinle We discussed this in the weekly tech meeting and agreed that those links should be skipped by a screen reader. The idea behind the links to the images is that you can enlarge a picture more easily that way, and this is unlikely to be a concern for someone using a screen reader, especially since only purely decorative images should have no alt-tag. I opened the issue up there in the CMS repo.

steffenkleinle:
> @steffenkleinle We discussed this in the weekly tech meeting and agreed that those links should be skipped by a screen reader. The idea behind the links to the images is that you can enlarge a picture more easily that way, and this is unlikely to be a concern for someone using a screen reader, especially since only purely decorative images should have no alt-tag. I opened the issue up there in the CMS repo.

Alright, thanks for explaining. The CMS issue: https://github.com/digitalfabrik/integreat-cms/issues/2801

native/src/components/NavigationTiles.tsx
```diff
@@ -64,10 +65,12 @@ const NavigationTiles = ({ tiles }: NavigationTilesProps): ReactElement => {
   const scrollToStart = () => scrollViewRef.current?.scrollTo({ x: 0, y: 0, animated: true })
   const scrollToEnd = () => scrollViewRef.current?.scrollToEnd({ animated: true })
 
+  const { t } = useTranslation('common')
+
   return (
     <TilesRow>
       {isScrollable && (
-        <StyledPressable onPress={scrollToStart}>
+        <StyledPressable onPress={scrollToStart} accessibilityLabel={t('back')}>
```
steffenkleinle, line 73: Sorry for the nitpicking here, but is back/forward really the right term here? I think previous/next would be more fitting.

LeandraH: I think "forward" and "back" is better in this case because I would expect "previous" and "next" to be part of a pagination which we don't have here. But I'm not precious about it either 🤷

steffenkleinle: Hmm. One more thought: Should these button even be visible to screenreaders? Aren't all items read to the user anyway? In that case it would just be confusing to announce the arrows at all. E.g. for Süddeutsche Zeitung and airbnb similar kind of arrows were just aria hidden.

native/src/components/HeaderBox.tsx
```diff
@@ -35,10 +36,16 @@ type HeaderBoxProps = {
 const HeaderBox = ({ goBack, canGoBack = true, text }: HeaderBoxProps): ReactElement => {
   const deviceWidth = useWindowDimensions().width
   const theme = useTheme()
+  const { t } = useTranslation('common')
```
steffenkleinle, line 39: :x: I think there is no key `back` in `common` atm since you reverted those changes, right?

LeandraH: You're right. Grrrr. That was so annoying to move `back` into `common`. I guess I'll be doing that again.

steffenkleinle:
For future reference:
Short hack to move keys between namespaces (as from `events.back` to `common.back` here), you can use the export/import functionality:
1. Export the constants
``` bash
yarn manage convert translations.json translations-csv csv
```
2. Search and replace all usages of `namespace_old.key` with `namespace_new.key`in all exported csvs (e.g. `events.back` with `common.back`)
3. Import the constants again
``` bash
yarn manage convert translations-csv translations.json json
```
4. Don't forget to change the usages in native and web

I feel like this is less error prone and faster than moving keys for all languages manually one by one (and even more so if you have to change multiple keys)

steffenkleinle (approved):
Nice, thank you! Sorry for that many comments about this issue....

I didn't find any usages of `events.back` so should be fine.

PR #2771, opened 2024-04-30
2767: Fix text cut of proposal

f1sh1918:
### Short description
Removed the adjusted styles and added `flex-shrink:1` to the title container, because there is a `flex-shrink:0` on the category icon container. Same issue for the search item.
You can not tell one item in a row not to shrink and the other to increase. Thats the issue i think
Tested on android and ios

image

image


steffenkleinle (changes requested):
:x: For me this decreases the sub category text size on android :/


Before:

steffenkleinle:
Already done in #2774.

Issue #2498, opened 2023-10-04
Minimize main thread work


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
Google still thinks our mobile performance is too slow, i.e. we have a LCP longer than 2.5 on most sites on mobile. Therefore we should try to speed up our web app more.

**Describe the solution you'd like**
Investigate whether we can minimize main thread work: https://developer.chrome.com/docs/lighthouse/performance/mainthread-work-breakdown/?utm_source=lighthouse&utm_medium=lr#how-to-minimize-main-thread-work

**Describe alternatives you've considered**
None.

**Additional context**
LCP:
https://search.google.com/search-console/core-web-vitals/drilldown?resource_id=https%3A%2F%2Fintegreat.app%2F&item_key=CAMQAhgC

Part of #2371 
relates to: #2499

ztefanie:
"When I look at the performance tab I can see that the main bundle is taking around 5s to be fully loaded. As we are using HTTP/2 in production, splitting the bundle into smaller chunks can result in better performance. Check out the AggressiveSplittingPlugin. There are several guides on how to implement this: e.g. splitting react and react-dom into separate chunks or just by setting the chunk max size. I think this block article is a good starting point what can be achived here -> https://medium.com/@addyosmani/progressive-web-apps-with-react-js-part-2-page-load-performance-33b932d97cf2"

PR #2705, opened 2024-02-28
2499: Proposals to improve js execution time

sarahsporck:
### Short description

Proposal to reduce js execution time and minimize main thread work. Feel free to use this pr as inspiration/starting point for #2499 and #2498. Also feel free to close this PR if you don't see the changes working

### Proposed changes

- set `sideEffects: false` in `package.json` to allow tree shaking our packages
- use splitChunks plugin to split the code into react, dompurify (as it is quite big and cannot be tree-shaken), node_modules & our code. This can probably be optimized further
- get rid of lodash in production code (check out the bundle analyzer)

### Related issues

Relates: #2498 #2499 

---

Issue #2499, opened 2023-10-04
Reduce javascript execution time


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
Google still thinks our mobile performance is too slow, i.e. we have a LCP longer than 2.5 on most sites on mobile. Therefore we should try to speed up our web app more.

**Describe the solution you'd like**
Investigate whether we can reduce javascript execution time: https://developer.chrome.com/docs/lighthouse/performance/bootup-time/?utm_source=lighthouse&utm_medium=lr#how-to-speed-up-javascript-execution

**Describe alternatives you've considered**
None.

**Additional context**
LCP:
https://search.google.com/search-console/core-web-vitals/drilldown?resource_id=https%3A%2F%2Fintegreat.app%2F&item_key=CAMQAhgC

Part of #2371 , relates to #2498.

sarahsporck:
Maybe check out https://web.dev/learn/performance

sarahsporck:
This adresses bundle size and code compression:

It seams tree shaking does not work for lodash. The bundle still contains unused components.
When switching to function imports this is fixed e.g. from `import { isEqual } from 'lodash'` to `import isEqual from 'lodash'`. Compare with the bundle analyzer. 
Sadly, i18next and luxon don't support tree-shaking.

We could also add `"sideEffects": false` to the `translation`, `shared` and `build-configs` project. This allows webpack to remove unused code from thes projects. https://webpack.js.org/configuration/optimization/#optimizationsideeffects

All over these changes could reduce the bundle size from around 791kb gzipped to 777kb gzipped.

steffenkleinle:
Hmm, doesn't sound like a lot (< 2%) but I guess it still makes sense to do what we can. No other optimization potential here?

sarahsporck:
Steps to get the same report:
1. Download Chrome (to have access to lighthouse) 
2. Open the Chromedevtools on Integreat
3. Run the dev server in production mode: `yarn run webpack-dev-server-ts --env config_name=integreat-test-cms`
4. Run lighthouse to checkout the results

> [!Warning] 
> Page Speed Insights might use slightly different throttling, resulting in slightly different results

To reproduce the same results:
1.. Create a new hardwareprofile in the network tap using [these settings](https://github.com/GoogleChrome/lighthouse/blob/main/docs/throttling.md#the-mobile-network-throttling-preset): 
 - Download: 1638kbit/s
 - Upload: 750kbit/s
[...]

PR #2776, opened 2024-05-02
2498 and 2499: Reduce execution time and reduce bundle size

steffenkleinle:
### Short description

A few optimizations to reduce execution time and reduce bundle size, mostly taken from #2705.

### Proposed changes

- Remove lodash in web, shared and translations
- Set `sideEffects: false` (see https://webpack.js.org/guides/tree-shaking/)
- Remove splash screen

### Side effects

There is no more splash screen on web (only visible with really slow internet).

### Resolved issues

Partly fixes: #2498, #2499.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,5 +80,8 @@ or use the Windows Subsystem for Linux following the information [here](./docs/w
 
 ## Further Reading
 
-More information on both the webapp and the native app can be found in the [web documentation](web/docs) and the [native documentation](native/docs).
-Documentation on [CI/CD](docs/cicd.md), [contributing](docs/contributing.md) and our [conventions](docs/conventions.md) can be found [here](docs).
+More information on both the webapp and the native app can be found in the [web documentation](web/docs)
+and the [native documentation](native/docs). Documentation on [CI/CD](docs/cicd.md),
+[contributing](docs/contributing.md) and our [conventions](docs/conventions.md) can be found [here](docs).
+And we have a file with some small [tricks](docs/development-tips.md) that at least some developers have found
+useful in working on this repo.
--- a/docs/development-tips.md
+++ b/docs/development-tips.md
@@ -1,0 +1,33 @@
+# Development Tips and Tricks
+
+This is a file in which we can collect some tricks that we might otherwise just put in private note files.
+We have a file to share these in instead here.
+
+### Changing the API in the web app
+
+In the browser console, run `window.localStorage.setItem('api-url', 'https://cms-test.integreat-app.de')`
+or whichever API (e.g. ''https://cms.integreat-app.de) you want to call. You can also do it manually in
+the local storage of the browser.
+
+See also [here](../web/README.md#cms).
+
+### Triggering a deployment
+
+In `/tools`, use the following command: `yarn trigger-pipeline trigger <workflow from trigger-pipeline.ts>—api-token <CircleCI token>`.
+
+See also [here](./cicd.md#triggering-a-delivery).
+
+### Opening a URL in the native app
+
+For Android: `npx uri-scheme open https://integreat.app/etc/pp/ --android`
+
+For iOS: `npx uri-scheme open integreat://integreat-app/etc/pp/ --ios`
+
+### Moving keys between namespaces in translations.json
+
+If you e.g. want to move the key "back" from the namespace "events" to "common":
+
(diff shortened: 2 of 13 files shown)


LeandraH (commented):
❓ Does /tools have side effects?

❓ This looks great but when I check the stat sizes in the bundle analyzer, I get a smaller size for main.blabla.js (from 2.86 MB down to 2.28 MB!) but the overall size stays exactly at 5.58 MB. Any ideas?

Otherwise it looks great, and I think it's definitely a good idea to especially remove Lodash but I'm a little confused about the sizing.

f1sh1918 (approved):
I already told sarah that i don't think we should put more effort into this, so not sure what was your motivation here. Did you find any improvements according execution time and bundle size @steffenkleinle ?
I checked the bundle sizes and the execution time as i also did in sarahs pr and i didn't find any significant improvements tbh.

If we really want a significant performance boost we might have to switch to NextJS.
All we do here is just "A drop in the ocean" in my opinion

I'm gonna approve this pr because removing lodash is a good choice and everything works fine but i think we should put effort into other topics or really switch to NextJS.

steffenkleinle:
> ❓ Does /tools have side effects?

It should not have. Why?

> ❓ This looks great but when I check the stat sizes in the bundle analyzer, I get a smaller size for main.blabla.js (from 2.86 MB down to 2.28 MB!) but the overall size stays exactly at 5.58 MB. Any ideas?

I'll check.

steffenkleinle:
> I already told sarah that i don't think we should put more effort into this, so not sure what was your motivation here. Did you find any improvements according execution time and bundle size @steffenkleinle ? I checked the bundle sizes and the execution time as i also did in sarahs pr and i didn't find any significant improvements tbh.
> 
> If we really want a significant performance boost we might have to switch to NextJS. All we do here is just "A drop in the ocean" in my opinion
> 
> I'm gonna approve this pr because removing lodash is a good choice and everything works fine but i think we should put effort into other topics or really switch to NextJS.

Agreed that this is only a drop in the ocean and for real performance improvements we have to look at SSR (for example with NextJS). Remix (https://remix.run/) might also be an option as in the next release it will be merged with react-router which we are already using (https://remix.run/blog/merging-remix-and-react-router), so migrating could be a lot less effort.

However, I think that removing lodash still makes sense. Since all the work (for web) was basically already done, I figured it makes sense to extract these changes and merge them nevertheless (even though the performance benefits are small or even non-existent). If you don't think we should merge this, please say.

LeandraH:
> > ❓ Does /tools have side effects?
> 
> It should not have. Why?

I think you could add the `sideEffects: false` to that `package.json` too, then

steffenkleinle:
> > > ❓ Does /tools have side effects?
> > 
> > 
> > It should not have. Why?
> 
> I think you could add the `sideEffects: false` to that `package.json` too, then

Probably won't make any difference as the tools package is not part of the monorepo setup. Won't hurt either though.

ztefanie (approved):
Not tested, code looks good. I agree with all of you, not a big win here, but also a good step to remove lodash.

steffenkleinle:
Okay, I figured out why lodash is still included in the bundle: `loadTranslations` used `merge`. Therefore I replaced it with `deepmerge-ts` as well. I used https://statoscope.tech for deeper insights in the bundle size.
You can create a `stats.json` file using `yarn build bundle_analyzer` (put in `web/reports/bundle/stats.json`) which can then be uploaded there.

This PR reduces the size of the main chunk/entry point from 1.37 mb to 1.10 mb (raw from 2.9 mb to 1.8 mb), so hopefully we'll see some (little) improvement here...

steffenkleinle:
@f1sh1918 what did you use to check for improvements?

LeandraH:
Looks like quite a few methods from lodash are also used in `translations/tools/manage.ts`, not sure how those are running at the moment 😅

steffenkleinle:
> Looks like quite a few methods from lodash are also used in `translations/tools/manage.ts`, not sure how those are running at the moment 😅

Those are not run at all in web/native, its only utility methods used to manage the translations. If you have a look at the bundle, lodash should not be included. If you think we should ditch altogether from our whole codebase, please open an issue.

LeandraH (approved):
Very nice, I now see a static size of 5.06 MB with main...js being 1.76 MB. Very cool, thank you!

PR #2772, opened 2024-05-01
2739 and 2588: Update circleci android images and xcode version

steffenkleinle:
### Short description

Update circleci android images and xcode version (to upgrade to node v20, the minimal node version is v20 since #2750).

### Proposed changes

- Upgrade android images to v2024.04.01
- Upgrade xcode to v15.3.0

### Side effects

None.

### Resolved issues

Fixes: #2739, #2558.

---

PR #2783, opened 2024-05-08
2739: Fix e2e tests

steffenkleinle:
### Short description

Fix execution of e2e tests in CI by using image including ruby and ruby bundler.

### Proposed changes

In #2772 I upgraded/changed the used images in the e2e-android and e2e-ios jobs to just plain node ones (before it was cimg/android). Since neither are the right choice (android including unnecessary stuff and being confusing, node not including ruby and ruby bundler), I changed the used images to cimg/ruby. 

### Side effects

None.

### Resolved issues

Fixes: None.

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,14 +146,16 @@ commands:
         steps:
             - restore_cache:
                 keys:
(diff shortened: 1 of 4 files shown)


LeandraH (approved):
Looks reasonable, thanks!

Issue #2767, opened 2024-04-25
Texts are cut off at the end


Docmale:
### Describe the Bug
In some cases, the text is cut off at the end instead of wrapping. This issue was first described in https://chat.tuerantuer.org/digitalfabrik/pl/dmurxw7c9jykmeunzinqsts3dy
It seems to me that this effect is due to the width of the viewport as I can also reproduce this effect on my devices, although less pronounced. In the emulator (Pixel_3a_API_34) the effect is so small that I didn't recognize it at first.
(Web is ok)

### Steps to Reproduce

1. Go to region 'Landeshauptstadt München'
2. for ios language German search for 'Konver' and find 'Konversationskurse - Deutsch sprechen üben'
3. for android language Romanian search for 'privitoare' and find 'Informatii privitoare la sedere'
4. see some lines are more or less cut off

### Additional Information

Android:
![PHOTO-2024-04-24-16-28-26](https://github.com/digitalfabrik/integreat-app/assets/21236324/af2303ae-b5b5-46d4-aa45-70c013229342)
IOS:
![PHOTO-2024-04-24-16-26-11](https://github.com/digitalfabrik/integreat-app/assets/21236324/c51257cb-71ba-464c-8d80-f226ac64ea44)

PR #2774, opened 2024-05-02
2767: Fix cut off texts in category list

steffenkleinle:
### Short description

Fix cut off texts in category list. Both #2769 and #2771 did not work for me/broke some other styling. This PR should resolve all those problems with minimal changes.

### Proposed changes

- Remove unused thumbnail for SubCategoryListItem
- Set `flex-shrink: 1` on `CategoryListTitle` to fix wrapping

### Side effects

None.

### Resolved issues

Fixes: #2767.

---

PR #2784, opened 2024-05-08
2767: Fix cut off search results

LeandraH:
### Short description

I wanted to join the club and also open a PR for this.

No, seriously, none of the other fixes worked for me and I solved it very differently so I decided to go with a new branch.

### Proposed changes

- Only block space for the thumbnail in the search results if there is one
- Set the container to wrap around

### Side effects

- Should be none

### Resolved issues

Fixes: #2767 

---


LeandraH:
Closed in favour of #2774

Issue #2666, opened 2024-02-12
Fix promote ios error


ztefanie:
The pipeline is failing for the ios promotion jobs in ~50% of the runs, with this error 
`Cannot update languages - could not find an editable version for 'IOS'`

See [here](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/10653/workflows/4b0f915e-1b9c-440b-9c68-eee34a0178e8/jobs/58796)

Expected behaviour: The pipeline should finish without errors.

**Additional context**
- Maybe this is helpful: https://github.com/fastlane/fastlane/issues/16991#issuecomment-692188666
- Keep in mind that this error only occurs on the first run, if you retry it succeeds.

Should be fixed by upgrading fastlane to [v2.222.0](https://github.com/fastlane/fastlane/releases/tag/2.220.0).

Related: #2698. The CI also fails there.

steffenkleinle:
According to https://github.com/fastlane/fastlane/pull/21742, this is caused by degraded apple servers. The linked PR tries to mitigate the issue, wasn't released yet. Therefore I'd say that we have to wait until a new fastlane release is done.

Issue #2698, opened 2024-02-26
iOS promotion failing because of timeout


steffenkleinle:
### Describe the Bug

The iOS promotion job is failing frequently for integreat because of a timeout after uploading the screenshots.

### Steps to Reproduce

Examples:
- https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/10753/workflows/17251e5f-91b3-4219-aa21-242036331de0/jobs/59202

### Expected Behavior

The promotion doesn't fail.

### Actual Behavior

The promotion fails.

### Additional Information

Should be fixed by upgrading fastlane to [v2.222.0](https://github.com/fastlane/fastlane/releases/tag/2.220.0).

Related: #2666. The CI also fails there.

steffenkleinle:
Fixed by https://github.com/fastlane/fastlane/pull/21693. Therefore we have to wait until the next fastlane release. Alternatively we could use the master branch directly by doing the following:
```
gem 'fastlane', github: 'fastlane/fastlane', branch: 'master'
```

However, I think we can just restart the CI for now if it fails as this is not that time critical.

steffenkleinle:
Done in #2785.

PR #2785, opened 2024-05-08
2666 and 2698: Fix ios promotion by upgrading fastlane (and other gems)

steffenkleinle:
### Short description

Fix ios promotion by upgrading fastlane (and other gems)

### Proposed changes

Update all gems in native, android and ios.
Updating fastlane should fix both #2666 and #2698.

### Side effects

None.

### Resolved issues

Fixes: #2666, #2698.

---


native/ios/Gemfile.lock
```diff
@@ -272,13 +273,14 @@ GEM
       xcpretty (~> 0.2, >= 0.0.7)
 
 PLATFORMS
-  ruby
```
steffenkleinle, line 275: Any idea why this only updates for ios but not the android or native? Same for `BUNDLED WITH` below...

LeandraH:
I think the Fastfiles still need to be updated.

native/ios/Gemfile.lock
f1sh1918: The platform should be readded!!
As i found out the bundler check the version of the current `Gemfile.lock` and uses this. When updating rubygems you have to remove the `Gemfile.lock` and rereun `bundle install` to get correct version and platforms.
Please do that for `native` & `android` & `ios`

f1sh1918 (approved):
looks good 👍

PR #2787, opened 2024-05-08
2244: Use new cms endpoints

steffenkleinle:
### Short description

Use new cms endpoints that are now versioned.

### Proposed changes

Use the new cms endpoints according to https://digitalfabrik.github.io/integreat-cms/api-docs.html#locations in favor of the deprecated ones.

### Side effects

None.

### Resolved issues

Fixes: #2244.

---


LeandraH (approved):
Very nice! Tested in Firefox and emulated iOS

ztefanie (approved):
Not tested, but can we move this `v3` string to a separate constant, so when we change it in the future, we only need to change it once?

PR #2790, opened 2024-05-10
2588: Fix failing iOS delivery

steffenkleinle:
### Short description

Fix failing iOS delivery.

### Proposed changes

Add gem `rb-readline` to fix error during delivery on macos image.

### Side effects

None.

### Resolved issues

Fixes: None.

---


LeandraH (commented):
Thanks for the fast fix!

PR #2795, opened 2024-05-15
Add native module for reading high contrast

ztefanie:
### Short description
Introduced a native module to use the Android function to read if high contrast is set.

### Proposed changes
- Added the Native module
- Use [this](https://android.googlesource.com/platform/frameworks/base/+/a4725ef/core/java/android/view/accessibility/AccessibilityManager.java#303) Android function
- React native rejected the issue to add this function to the native modules [here](https://github.com/facebook/react-native/issues/30869)

This is just one solution, we can discuss if this PR is even what we want. Not tested for ios / No handling for ios yet. No tests or cleanup. Will do if we decide, we want to do it this way.

### DIFF
--- a/native/.idea/misc.xml
+++ b/native/.idea/misc.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_20" project-jdk-name="11 (2)" project-jdk-type="JavaSDK" />
+</project>
--- a/native/android/app/src/main/java/tuerantuer/app/integreat/IntegreatPackage.java
+++ b/native/android/app/src/main/java/tuerantuer/app/integreat/IntegreatPackage.java
@@ -6,6 +6,7 @@ import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.uimanager.ViewManager;
 import tuerantuer.app.integreat.constants.NativeConstantsModule;
 import tuerantuer.app.integreat.fetcher.FetcherModule;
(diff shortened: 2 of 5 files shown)


steffenkleinle:
@ztefanie thank you for the proposal. Seems to work as expected. I think we should discuss this in a dedicated meeting (perhaps at the next conference).

Not tested yet. But if it works and there is an iOS/web equivalent I think we could do something like
- the system setting decides whether we automatically enable the high contrast mode in the app
- there is a toggle in the setting to override that (should be easy to implement and we would not need a design/change the UI anywhere)

On web there seems to be a CSS media query: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast

Notes:
- The android setting is only for "high contrast text", not for a high contrast mode in general. So while this is an indication the user would like to use high contrast mode in general, this is not necessarily the case.
- Do we now since what Android version this setting exists? ~I did not find any concrete information here, but first mentions on google seem to be from two years ago which would be Android 12 probably. So this only works for a subgroup of users.~ Seems like this exists in an experimental version for a while since Android 5.
- It seems like there is an iOS equivalent, right?

LeandraH:
What's the status on this? Does anybody remember?

steffenkleinle:
Probably we can close this for now as we still don't have a finished design system. Also I think we thought of this (or high contrast/themes in general) as a topic for @lunars97 final topic, right @ztefanie ?

ztefanie:
yes, but it is not decided yet, i also thought about this, but i think if we close it, we may do the work from the start. 

ok i will close this and link it in the issue.

Issue #2755, opened 2024-04-08
Do not show search results, when there is no search term yet


ztefanie:
**Is your feature request related to a problem? Please describe.**
From our target group research we got the feedback, that people get confused, when they open the search, by the "results".

**Describe the solution you'd like**
Do not show results, if the search term has zero length

**Describe alternatives you've considered**
No alternatives yet discussed. If you have another idea, it needs to be discussed with the target group team and with the ui team.

**Additional context**
https://webnext.integreat.app/augsburg/de/search

[See the issue in open project](https://openproject.tuerantuer.org/projects/community-feedback/work_packages/3234/activity)

LeandraH:
Yes, that had already been changed at some point, I opened a bug ticket to hide the results in web again #2896

steffenkleinle:
It is fixed and just wasn't promoted to production yet. See webnext/beta.

PR #2796, opened 2024-05-16
2755: Do not show search results without search term

marko-ignjatovic:
### Short description

If the length of the search term is equal to 0, no results are displayed

### Proposed changes

- Add conditional for checking the length of the query prop

### Resolved issues

Fixes: #2755 

---


LeandraH (commented):
Amazing, thank you for your contribution! This looks great already, just two small things:
1. I think you have different white space rules in your IDE than our default, we use a tab that is two spaces wide. Running `yarn lint --fix` might already fix that.
2. In the browser, you now see the button for sending feedback if you didn't find what you were looking for. I think we should also hide that if there is no search query.
What do you think?

ztefanie (commented):
Thanks for your contribution :rocket: 

please add the conditional rendering also for the feedback part below the search (as @LeandraH mentioned)

Maybe you also want to add a small test for this?

marko-ignjatovic:
@LeandraH I hope it's alright now. I am not sure what to do about `vscode/launch.json` file tho. That config doesn't work for me, so I had to change. Any idea?

LeandraH:
Yup, looks great, thank you! But I have a failing test now, `SearchPage › should display results`, does that run for you? (I usually run a `yarn lint`, `yarn test`, `yarn ts:check` and `yarn prettier:write` because those are the steps that the CI will run and yell at me if I missed something)

About the `vscode/launch.json`, I'm not sure, I don't use it. But since these environment files tend to cause problems between different setups, I would say update the file so that it works for you (as you did already), and then add it to your personal gitignore file.

marko-ignjatovic:
You are right @LeandraH , I am also failing `SearchPage.spec.tsx` test. I have a issue with `yarn lint` tho, when I run it I get about 200 errors for `@typescript-eslint/strict-boolean-expressions`. What should we do about that one?

LeandraH:
Oooof, that sounds like the linter might be using the wrong `.eslintrc` file. The one in this repo does have the `@typescript-eslint/strict-boolean-expressions` setting enabled, but with exceptions for nullable booleans and nullable strings. Those exceptions are probably the ones that the linter is complaining about for you. The IDE might have a different setup for the linter?

marko-ignjatovic:
image
image


I am using `.eslintrc.js` from the root. I tried to change `error` to `warn` for `strict-boolean-expression` and I could see the change. That's how I checked which file is used

LeandraH:
That is weird. Especially since the variables that lint is complaining about are booleans. @steffenkleinle did you experience this when you implemented this? Or any ideas from @ztefanie or @f1sh1918 ?

LeandraH (approved):
Awesome, test fixed! Last thing to do is removing the `.vscode/launch.json` file (it might need a `.vscode/` in the `.gitignore`).

I'm already approving the PR now though because I'll be on vacation until 17 June and don't want to block things :)

LeandraH:
Did you manage to figure out the linter problem or did you work around it?

marko-ignjatovic:
> Did you manage to figure out the linter problem or did you work around it?

@LeandraH I haven't solved it. Just ignored

web/src/routes/SearchPage.tsx
```diff
@@ -100,16 +100,17 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
-          <SearchListItem
-            title={title}
-            contentWithoutHtml={parseHTML(content)}
-            key={path}
-            query={query}
-            path={path}
-            thumbnail={thumbnail}
-          />
-        ))}
+        {query.length > 0 &&
```
steffenkleinle, line 103: :wrench: Could we move this up and only render the `List` conditionally as well? Or does that lead to problems?

web/src/routes/SearchPage.tsx
```diff
@@ -100,16 +100,17 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+              title={title}
+              contentWithoutHtml={parseHTML(content)}
+              key={path}
+              query={query}
+              path={path}
+              thumbnail={thumbnail}
+            />
+          ))}
       </List>
       <SearchFeedback
```
steffenkleinle, line 115: :wrench: I think it would be better to also conditionally render the whole `SearchFeedback` instead of returning an empty `div` if the query is empty.

.gitignore
```diff
@@ -127,4 +127,8 @@ runConfigurations.xml
 # Mac OS X
 .DS_Store
 
-reports/
\ No newline at end of file
+reports/
+
+
+#VSCode
+.vscode/launch.json
```
steffenkleinle, line 134: Did you add custom run configs or why did you have changed `launch.json` files? If those launch configs make sense I think we could just include them in the repo/commit those changes.

steffenkleinle (commented):
:fire: :rocket: :fire: 
Thank you for another contribution! Amazing :green_heart:

steffenkleinle:
>> I am using `.eslintrc.js` from the root. I tried to change `error` to `warn` for `strict-boolean-expression` and I could see the change. That's how I checked which file is used

> That is weird. Especially since the variables that lint is complaining about are booleans. @steffenkleinle did you experience this when you implemented this? Or any ideas from @ztefanie or @f1sh1918 ?

Hmm no, not really. One idea: Did you run `yarn ts:check` before that? If typescript didnt compile the project yet, eslint sometimes fails as well (since the necessary types are not available). Make sure to run it for the whole project in the root to also include types from the build-config and shared folders.

steffenkleinle:
Hey @marko-ignjatovic,
will you still be able to finish your PRs or should we take over?
Thank you!

marko-ignjatovic:
Feel free to take it over as I am currently busy

steffenkleinle:
Closing in favor of #2859.

Issue #2756, opened 2024-04-08
Fix sending empty search feedback


f1sh1918:
### Describe the Bug

A user should not be able to send an empty feedback

### Steps to Reproduce

1. [Go to '...'](https://webnext.integreat.app/augsburg/de/search/?query=dasdadsa)
2. Clear input field
3. Send feedback

### Expected Behavior

When the input field gets cleared, the submit button will be disabled and a note should be shown. Similar to the normal feedback if no reaction was chosen

### Actual Behavior

An empty feedback will send

PR #2802, opened 2024-05-23
2756: Disable sending empty search feedback

LeandraH:
### Short description

When someone wants to give feedback on a search result but then deletes the query, they can't send the feedback anymore with this PR.

### Resolved issues

Fixes: #2756 

---


steffenkleinle (changes requested):
:x: This does only disable the search feedback if no comment is set. If I delete the search term but set a comment I can still send feedback.

:wrench: It would also be nice if you could add some tests for this.

LeandraH:
> ❌ This does only disable the search feedback if no comment is set. If I delete the search term but set a comment I can still send feedback.

But do we really want to disable that? If someone removes the query but then puts something in the comment field, that seems to me like they have thought that through, and probably couldn't figure out how to send the feedback from somewhere else.

steffenkleinle:
> > ❌ This does only disable the search feedback if no comment is set. If I delete the search term but set a comment I can still send feedback.
> 
> But do we really want to disable that? If someone removes the query but then puts something in the comment field, that seems to me like they have thought that through, and probably couldn't figure out how to send the feedback from somewhere else.

Not entirely sure. But I think we should keep it consistent and personally I don't see a case why it would make sense to send an empty query. Do you?

LeandraH:
> > > ❌ This does only disable the search feedback if no comment is set. If I delete the search term but set a comment I can still send feedback.
> > 
> > 
> > But do we really want to disable that? If someone removes the query but then puts something in the comment field, that seems to me like they have thought that through, and probably couldn't figure out how to send the feedback from somewhere else.
> 
> Not entirely sure. But I think we should keep it consistent and personally I don't see a case why it would make sense to send an empty query. Do you?



> > > ❌ This does only disable the search feedback if no comment is set. If I delete the search term but set a comment I can still send feedback.
> > 
> > 
> > But do we really want to disable that? If someone removes the query but then puts something in the comment field, that seems to me like they have thought that through, and probably couldn't figure out how to send the feedback from somewhere else.
> 
> Not entirely sure. But I think we should keep it consistent and personally I don't see a case why it would make sense to send an empty query. Do you?

Yes, I think if someone can't find the feedback button when trying to give feedback on something else. It's obviously not the best place to do it but it's better than there not being any feedback at all.

web/src/components/__tests__/SearchFeedback.spec.tsx
```diff
@@ -50,4 +51,19 @@ describe('SearchFeedback', () => {
...
+
+  it('should not allow sending search feedback if query term is removed', async () => {
+    const { getByText, rerender } = renderWithTheme(
+      <SearchFeedback cityCode={cityCode} languageCode={languageCode} query='ab' noResults />,
+    )
+    expect(getByText('feedback:send')).toBeTruthy()
+
+    // the query is controlled in the parent of SearchFeedback, so we need to update the props
+    rerender(
+      <ThemeProvider theme={theme}>
```
steffenkleinle, line 63: Can't we use `renderWithTheme` here as well?

web/src/components/__tests__/SearchFeedback.spec.tsx
```diff
@@ -50,4 +51,19 @@ describe('SearchFeedback', () => {
     )
     expect(getByText('feedback:send')).toBeTruthy()
   })
+
+  it('should not allow sending search feedback if query term is removed', async () => {
+    const { getByText, rerender } = renderWithTheme(
+      <SearchFeedback cityCode={cityCode} languageCode={languageCode} query='ab' noResults />,
+    )
+    expect(getByText('feedback:send')).toBeTruthy()
```
steffenkleinle, line 59: Probably we should add something like
```suggestion
    expect(getByText('feedback:send')).toBeTruthy()
    expect(getByText('feedback:send')).toBeEnabled()
```

steffenkleinle (approved):
We send the actual query if the user deletes the query. So the feedback is not what the user would expect if the query is deleted and feedback is still sent.
I don't really mind though.

Since this does not do what the issue said to do, maybe @f1sh1918 has another opinion on this.

ztefanie (approved):
I think this is a good solution. If one of the fields contains data it is valuable feedback, so i think it should be able to be send

> Not entirely sure. But I think we should keep it consistent and personally I don't see a case why it would make sense to send an empty query. Do you?

For example if I search for "kita" and get no results and then i search for "hausaufgabenberetung" and also get no results, I maybe delete the term above and write in the comments that i seached for two terms and had no success with neither of them.

web/src/components/__tests__/SearchFeedback.spec.tsx
LeandraH: Unfortunately not, we need to use the `rerender` that the earlier `renderWithTheme` returns. Otherwise the `SearchFeedback` gets rendered a second time with the new props, as opposed to rerendering the first one with new props.

steffenkleinle: Ahh sorry I missed that. Nvm :)

web/src/components/__tests__/SearchFeedback.spec.tsx
LeandraH: Ooooops, you're absolutely right, thanks!

PR #2804, opened 2024-05-23
Update iOS lock files

LeandraH:
### Short description

Just updating the lock files in the iOS folder


steffenkleinle (approved):
Thanks, looks good.

f1sh1918 (approved):

Issue #2801, opened 2024-05-22
Header has shadows behind icons (mobile view only)


nikolahoff:
### Describe the Bug

See the image attached
The error only appears on mobile view
I tested different browsers (firefox, chrome) and the result is the 

![Image](https://github.com/digitalfabrik/integreat-app/assets/97434837/243483c6-5695-411b-a608-30e6d705c7fe)

PR #2814, opened 2024-06-03
2801: Remove shadow in header on mobile

steffenkleinle:
### Short description

Remove the shadow in the header on mobile (web).

### Proposed changes

Remove shadow and padding for header and first children on mobile web.

### Side effects

Less padding between the icons (but imo not needed).

### Resolved issues

Fixes: #2801.

---

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -57,12 +57,6 @@ const Row = styled.div`
     min-height: ${dimensions.headerHeightSmall}px;
     overflow-x: auto;
     padding: 8px 0;
-    box-shadow: 0 2px 5px -3px rgb(0 0 0 / 20%);
-
-    :first-child {
-      box-shadow: 0 2px 5px -3px rgb(0 0 0 / 12%);
-      padding: 0 4px;
-    }
   }
 `
 


web/src/components/Header.tsx
```diff
@@ -57,12 +57,6 @@ const Row = styled.div`
     min-height: ${dimensions.headerHeightSmall}px;
     overflow-x: auto;
     padding: 8px 0;
-    box-shadow: 0 2px 5px -3px rgb(0 0 0 / 20%);
-
-    :first-child {
-      box-shadow: 0 2px 5px -3px rgb(0 0 0 / 12%);
-      padding: 0 4px;
-    }
```
steffenkleinle, line 65: Does anybody have an idea why this was needed? And why this was not a problem before and only started leading to this weird shadow now?

f1sh1918: the answer to that question can be found here:
https://github.com/digitalfabrik/integreat-app/blob/821b2b0fb666f6f90a602fc869c659f8466377f5/web/package.json#L58 @steffenkleinle 
But no idea why sarah added this shadows tbh
At least update of styled components library made this style working

f1sh1918 (approved):
Tested on chrome & firefox.
Thx for the quick fix

LeandraH (approved):
Very nice, thanks! Tested in Firefox and Safari (OS)

Issue #2697, opened 2024-02-26
Gradle build daemon disappeared unexpectedly


steffenkleinle:
### Describe the Bug

We frequently experience failing android build jobs in the CI with the following error:
> Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

### Steps to Reproduce

Examples:
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/10762/workflows/24ca6c54-7cd7-4bf4-a56c-228c8fcb8baf/jobs/59259
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11011/workflows/3f838890-980e-4218-9ecd-19f24baa07ba/jobs/60263

### Expected Behavior

Build doesn't fail.

### Actual Behavior

Builds fail.

### Additional Information

None.

PR #2815, opened 2024-06-04
2697: Optimize gradle builds

ztefanie:
### Short description
Improvments for gradle builds

### Proposed changes
- Enable parallel build
- Disable hermes undefined warnings

See scan before here: https://scans.gradle.com/s/iquknvkkj7wjg
See scan after here: https://scans.gradle.com/s/jjay2v7tbxmvw

### Side effects

- gradle builds in the CI 

### Resolved issues

Fixes: #2697 

---


steffenkleinle (approved):
LGTM

native/android/app/build.gradle
```diff
@@ -55,6 +55,7 @@ react {
     //
     //   The list of flags to pass to the Hermes compiler. By default is "-O", "-output-source-map"
     // hermesFlags = ["-O", "-output-source-map"]
+    hermesFlags =  ["-O", "-output-source-map", "-Wundefined-variable"] // this will disable all undefined warnings (they are created due to hermes ahead of time compilation)
```
f1sh1918, line 58: you may remove the comment above

native/android/gradle.properties
```diff
@@ -37,4 +37,6 @@ newArchEnabled=false
 
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
-hermesEnabled=true
\ No newline at end of file
+hermesEnabled=true
+
+org.gradle.parallel=false
```
f1sh1918, line 42: should this really be false?
Due to your description parallel builds are enabled with this pr.
Then its not very intuitive to enable sth by setting it false. This just confuses me

f1sh1918 (commented):
Build time was really good. Nice work! 👍 
I'm just a little bit confused about the `parallel` flag. Check comment :)

native/android/gradle.properties
ztefanie: You are right, I messed up, will set it to true.

f1sh1918 (approved):
looks good

Issue #2579, opened 2023-11-21
Create link base component


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user, I want to have similarly looking UI elements including links. As a developer I don't want to have to apply the same styling every time I use a link and don't want to be confused by multiple components/styles.

**Describe the solution you'd like**
At the moment we use different Link components and styles: CleanLink, CleanAnchor, helpers.removeLinkHighlighting or just use `<a>` or `Link` directly. This should be unified in one base component that is used everywhere. This base component should go to the `base` directory.

Make sure a11y best practices are followed.

**Describe alternatives you've considered**
None.

**Testing information**
All links in the app should look and behave the same way as before.

**Additional context**
This was done for buttons in #2578.

PR #2817, opened 2024-06-05
2579: Implement link component

marko-ignjatovic:
### Short description

Added new `Link` component which represents merge of `AnchorLink` and `CleanLink` 

### Proposed changes

- Created `Link` component
- Removed `AnchorLink.tsx`
- Removed `CleanLink.tsx`

### Side effects

- All link components inside app

### Resolved issues

Fixes: #2579

---


LeandraH (commented):
That looks like a great start, thank you! But there are some more places where we could replace the current links, it seems to be places where we use HTML's anchor tags like in `SharingPopup` or `react-router-dom`s `Link` like in `HeaderActionItemLink`. I found these by searching for `href`, that might help in finding more.

web/src/components/base/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const getUrl = () => {
+    if (to) {
+      return to
+    }
+    if (href) {
+      return href
+    }
+    return '/'
+  }
```
steffenkleinle, line 55: :x: This allows for wrong usage of the component since neither `to` nor `href` are required. Personally I'd favor to unify the behavior and always just have either `to` or ` href` and remove the other prop completely.

Another option to make this typesafe (also less unified/more confusing) you could do something like:
``` ts
type ToProps = {
  href: string
  to: undefined
} | {
  href: undefined
  to: string
}
```

However, I think only having either `to` or `href`` would be the better choice since
- less logic needed
- streamlined usage
- no confusion that different properties have the same result

Personally I don't mind whether we keep `to` or `href`.

web/src/components/base/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+      // @ts-expect-error wrong types from polymorphic 'as', see https://github.com/styled-components/styled-components/issues/4112
+      <StyledLink
+        as='a'
+        href={url}
+        aria-label={ariaLabel}
+        className={className}
+        data-testid='externalLink'
+        $removeHighlight={removeHighlight}
+        $adaptiveFontSize={adaptiveFontSize}
+        {...newTabProps}>
```
steffenkleinle, line 78: :x: I think this changes behavior, right? Atm not all external links are opened in a new tabs. Maybe we should investigate whether that makes sense but this PR should not change behavior without discussion about it.

web/src/components/base/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+      </StyledAnchor>
+    )
+  }
+
+  if (isExternalUrl(url)) {
+    return (
+      // @ts-expect-error wrong types from polymorphic 'as', see https://github.com/styled-components/styled-components/issues/4112
+      <StyledLink
+        as='a'
+        href={url}
```
steffenkleinle, line 72: :question: why do we need this separate check/typecast? Isn't this nearly exactly the same as `StyledAnchor` (except the flex and helpers which we could also introduce here).

web/src/components/StyledToolbarItem.tsx
```diff
@@ -1,9 +1,11 @@
 import styled from 'styled-components'
 
 import dimensions from '../constants/dimensions'
-import CleanAnchor from './CleanAnchor'
+import Link from './base/Link'
 
-const StyledToolbarItem = styled(CleanAnchor)`
+const StyledToolbarItem = styled(Link).attrs(() => ({
+  isCleanAnchor: true,
```
steffenkleinle, line 7: :wrench: this prop can be removed, see explanation below.

web/src/components/base/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+type LinkProps = {
+  to?: string
+  children: ReactNode
+  ariaLabel?: string
+  className?: string
+  newTab?: boolean
+  removeHighlight?: boolean
+  adaptiveFontSize?: boolean
+  href?: string
+  isCleanAnchor?: boolean
```
steffenkleinle, line 31: :wrench: AFAICS this is only used once with an external url and therefore this should not be necessary as it is already handled with the check for `isExternalUrl`.

web/src/components/ListItem.tsx
```diff
@@ -32,7 +32,9 @@ const Title = styled.div`
   font-weight: 700;
 `
 
-const FullWidthLink = styled(StyledCleanLink)`
+const FullWidthLink = styled(Link).attrs(() => ({
+  stylized: true,
```
steffenkleinle, line 36: :question: Why is this necessary?

web/src/components/ContactItem.tsx
```diff
@@ -11,12 +10,13 @@ const Marker = styled.img`
...
 
-const Link = styled(CleanLink)`
+const StyledLink = styled(Link).attrs(() => ({
+  adaptiveFontSize: true,
+}))`
   align-items: center;
   padding-top: 4px;
   gap: 8px;
   overflow-wrap: anywhere;
-  ${helpers.adaptiveFontSize};
```
steffenkleinle, line 19: :question: why did you change this?

web/src/components/base/Link.tsx
```diff
@@ -0,0 +1,98 @@
...
+const Link = ({
+  to,
+  href,
+  children,
+  ariaLabel,
+  isCleanAnchor,
+  className,
+  newTab,
+  removeHighlight,
+  adaptiveFontSize,
```
steffenkleinle, line 43: :question: why are these questions necessary/why can't we use regular styling for this?

steffenkleinle (changes requested):
Thanks for another contribution @marko-ignjatovic :tada: It is very much appreciated :)
I left some reviews with minor changes in mind, let me know if you need anything.

steffenkleinle:
Closing in favor of #2860.

Issue #2811, opened 2024-06-03
Fix screenshots for apple


ztefanie:
**Is your feature request related to a problem? Please describe.**
Newest integreat release was rejected due to wrong informations on the screenshots. See [here](https://appstoreconnect.apple.com/apps/1072353915/distribution/reviewsubmissions/details/4169418b-c2a5-418a-bb7a-331985c079a2)

> Review Environment
Submission ID: 4169418b-c2a5-418a-bb7a-331985c079a2
Review date: May 27, 2024
Version reviewed: 2024.5.1
Guideline 2.3.7 - Performance - Accurate Metadata
The app screenshots include references to the price of the app or the service it provides, which is not considered a part of these metadata items.
Next Steps
To resolve this issue, please remove any references to pricing from the app's screenshots, including any references to the app or service being free or discounted. If you would like to advertise changes to the app's price, consider including this information in the app description.
Resources
For information on how to revise the app name, please review [Renaming a Project or App](http://help.apple.com/xcode/mac/8.2/#/dev3db3afe4f).
Support
Reply to this message in your preferred language if you need assistance. If you need additional support, use the [Contact Us module](https://developer.apple.com/contact/topic/#!/topic/select).
Consult with fellow developers and Apple engineers on the [Apple Developer Forums](https://developer.apple.com/forums/).
Help improve the review process or identify a need for clarity in our policies by [suggesting guideline changes](https://developer.apple.com/contact/app-store/?topic=guideline).

**Describe the solution you'd like**
Fix the issue, by adding adjusted screenshots.

**Describe alternatives you've considered**
Contact apple support and convince them to accept the screenshots.

**Additional context**
Maybe we need help from uiux to create new screenshots

hauf-toni:
New Screens can be found here: [Nextcloud](https://nextcloud.tuerantuer.org/index.php/apps/files/files/6464428?dir=/Projects/Integreat/I%205.%20Design%20and%20Graphics/I%205.3.%20Digital/I%205.3.10%20Store%20Content/Playstore_Screenshots%20Ordner/Austauschscreen%20mit%20%C3%9Cbersetzungen)

hauf-toni:
[...]

PR #2828, opened 2024-06-11
2811: Readd screenshots

steffenkleinle:
### Short description

Readd the screenshots we removed in #2819 without the problematic wording.

### Resolved issues

Fixes: #2811.

---


ztefanie (commented):
Hmh, it was like this before, but I do not like that the screenshots are always in englisch and only the heading is different. One of the key features of the app ist, that it supports different languages, so I would highly perfer if the app constants and content is in the language the screenshot is for, instead of english.

steffenkleinle:
> Hmh, it was like this before, but I do not like that the screenshots are always in englisch and only the heading is different. One of the key features of the app ist, that it supports different languages, so I would highly perfer if the app constants and content is in the language the screenshot is for, instead of english.

I agree. Could you open a separate issue for that though please? I also already commented something that the screenshots are quite outdated/not showing all our features in mattermost.

ztefanie (approved):

ztefanie:
> > Hmh, it was like this before, but I do not like that the screenshots are always in englisch and only the heading is different. One of the key features of the app ist, that it supports different languages, so I would highly perfer if the app constants and content is in the language the screenshot is for, instead of english.
> 
> I agree. Could you open a separate issue for that though please? I also already commented something that the screenshots are quite outdated/not showing all our features in mattermost.

Created this one: https://github.com/digitalfabrik/integreat-app/issues/2830
Feed free to add your thoughts there.

LeandraH (approved):

Issue #2837, opened 2024-06-19
Mobile layout Broken


f1sh1918:
### Steps to Reproduce

1. Go to local news page on mobile mode
2. Scroll down and see the issue. Layout is broken

### Expected Behavior

scrolling should stop after toolbar

### Actual Behavior

You can scroll way further than bottom toolbar

### Additional Information

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #2838, opened 2024-06-19
2837: Mobile layout broken

f1sh1918:
### Short description

The mobile layout is broken on many pages

### Proposed changes

- remove `sideEffects:false`

### Side effects

- hopefully none

### Resolved issues

Fixes: #2837

---

### Note
I removed all sideEffects flag because i dunno what else issue they make cause

### DIFF
--- a/build-configs/package.json
+++ b/build-configs/package.json
@@ -2,7 +2,6 @@
   "name": "build-configs",
   "version": "0.0.1",
   "main": "index.ts",
-  "sideEffects": false,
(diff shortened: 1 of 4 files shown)


steffenkleinle (approved):
Tested on firefox, thanks for the fix and good catch!
That was a pretty big side effect.... :astonished:

LeandraH (approved):
Ouch, thanks for the quick fix! Tested in Firefox

Issue #2665, opened 2024-02-07
Error when deep linking to another city on ios


sarahsporck:
### Describe the Bug

On ios when I deep link to another city an error is sent to sentry. I think we try to open the in-app browser twice.

### Steps to Reproduce

1. Go to 'Augsburg Stadt/Willkommen'
2. Deep link with `npx uri-scheme open integreat://integreat.app/testumgebung/de/willkommen/bei-uns/stadtplan --ios ` 
3. See the following
    1. A web view is opened
    2. The app navigates to the Dashboard page (as the deep link leads to an error)

### Expected Behavior

The web view is opened and the app does **not** navigate to the Dashboard Page. Also, no error is reported.

### Actual Behavior

The app navigates to the Dashboard page and an error is reported.

### Additional Information

See corresponding [sentry issue](https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/3807/events/5f20f0078dcf4641a016aacfca68269f/).

steffenkleinle:
Possibly related: https://github.com/digitalfabrik/integreat-app/issues/1922
We attempted to fix the same issue there.

PR #2839, opened 2024-06-21
2665: Fix error in opening deep links in iOS

LeandraH:
### Short description

When trying to open a deep link in a different city, the inAppBrowser is opened twice in iOS, and the user ends up being sent to the dashboard page with an error in the Snackbar and in Sentry.

### Proposed changes

When redirecting to a deep link, we give it a little timeout first to allow the inAppBrowser to start. To allow for older devices, we also start an interval to keep trying to open the inAppBrowser. We don't check if the first method with the timeout worked before attempting a second time. I now added a little check to see if the inAppBrowser is already open before trying to open it again.

### Side effects

Hopefully nothing.

### Resolved issues

Fixes: #2665


f1sh1918 (approved):
Tested on ios.
Works nice 👍

native/src/components/RedirectContainer.tsx
```diff
@@ -30,8 +31,11 @@ const RedirectContainer = ({ route, navigation }: RedirectContainerProps): React
 
   useEffect(() => {
     // To support potentially older devices taking longer we setup a separate interval to retry the navigation
-    const interval = setInterval(() => {
-      navigateToDeepLink(url)
+    const interval = setInterval(async () => {
+      const noInAppBrowserOpenYet = !(await InAppBrowser.isAvailable())
+      if (noInAppBrowserOpenYet) {
+        navigateToDeepLink(url)
+      }
```
steffenkleinle, line 38: :wrench: I find the current naming hard to understand, perhaps the following (a positive wording) would be better?
```suggestion
      const urlRedirected = await InAppBrowser.isAvailable()
      if (!urlRedirected) {
        navigateToDeepLink(url)
      }
```

steffenkleinle: :question: why does `InAppBrowser.isAvailable()` indicate whether it is already open? According to the docs:
> Detect if the device supports this plugin.

AFAICS, it only tells whether we are able to open a link in general. What am I missing?

steffenkleinle (approved):

native/src/components/RedirectContainer.tsx
LeandraH: You're right, I misunderstood what it did! The code that I wrote is completely useless, it just doesn't run the interval on devices that can open the `inAppBrowser`, as in it never does. Good catch, thanks!

PR #2840, opened 2024-06-21
2176: Add a11y label to pagination dots

LeandraH:
### Short description

Adding a label to the pagination dots in the slide carousel.

### Proposed changes

- Add a label to the pagination dots.
- Add a translation for `goTo.pageNumber` in the error name space. It's not great but that's where all the other `goTo` keys were, so I guess it's a decent spot.

### Resolved issues

Fixes: #2176


native/src/components/Pagination.tsx
```diff
@@ -29,11 +32,20 @@ type PaginationProps = {
 
 const Pagination = ({ slideCount, currentSlide, goToSlide }: PaginationProps): ReactElement => {
   const goToSlideIndex = (index: number) => () => goToSlide(index)
+  const { t } = useTranslation()
```
f1sh1918, line 35: since we are using not different namespaces in this component i would prefer using here directly and remove the `ns: 'error'`
```suggestion
  const { t } = useTranslation('error')
```

steffenkleinle (approved):
While I am okay with the current solution, I am not sure if this accessibility label really helps with understanding what is happening (are intro slides called "page/seite"?). Perhaps it would be better to just hide the dots for screen readers as all options are available anyway (next/previous/skip).

LeandraH:
> While I am okay with the current solution, I am not sure if this accessibility label really helps with understanding what is happening (are intro slides called "page/seite"?). Perhaps it would be better to just hide the dots for screen readers as all options are available anyway (next/previous/skip).

We use it as pagination here, so sending the user to a page seems to be the expected behavior. We could also hide it but since this was something that the report from Pfennigparade specifically complained about, I'm in favor of adding the labels instead.

Issue #2763, opened 2024-04-19
iOS VoiceOver doesn't read settings toggles as toggles


LeandraH:
### Describe the Bug

When using iOS VoiceOver to navigate the settings in the app, the system reads the text for the toggle buttons ("Push notifications Receive push notifications for the latest local news") without telling me that there is a toggle button, what state the button is in, and how to interact with it.

Android TalkBack does do exactly that.

![Image](https://github.com/digitalfabrik/integreat-app/assets/18513943/c185a61a-da0b-4aa2-98e1-42399758b47f)

### Steps to Reproduce

1. Use iOS VoiceOver
2. Go to settings
3. Listen to description of the toggle buttons

### Expected Behavior

I would like VoiceOver to tell me the text of the button, that it is a toggle, what state the toggle is in,<s> and how to change that state (VoiceOver usually says to double tap to interact or something along those lines)</s> _Edit: that's only TalkBack, VoiceOver only tells you that it's a button_

### Actual Behavior

VoiceOver tells me the text of the button.

### Additional Information

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #2841, opened 2024-06-21
2763: Read out setting toggles in iOS

LeandraH:
### Short description

The iOS screen reader, VoiceOver, used to read out the words on the page of the settings without telling me that there are toggle buttons to push. Now it does say that.

### Proposed changes

- The `react-native` component `Pressable` has the prop `accessible` by default set to `true`. That means that all the children of the `Pressable` are grouped into one. That is usually great but meant in this case that the toggle isn't read out as such. So I set the `Pressable` to have `accessible: false`. More about the prop: https://reactnative.dev/docs/accessibility#accessible

### Side effects

- When swiping through the page in Android, the button is also read out in several swipes.

### To test

I have not been able to get VoiceOver running in Xcode, I think it needs an actual iPhone to test this.

### Resolved issues

Fixes: #2763


f1sh1918 (approved):
Tested on ios real device settings toggle was read out 👍

steffenkleinle (approved):
LGTM

Issue #2789, opened 2024-05-08
Console cluttered by unknown properties caused by styled-component props


steffenkleinle:
### Describe the Bug

Our web console is cluttered with a lot of errors and warnings (webpack dev mode only):

> styled-components: it looks like an unknown prop "hasPadding" is being sent through to the DOM, which will likely trigger a React console error. If you would like automatic filtering of unknown props, you can opt-into that behavior via `<StyleSheetManager shouldForwardProp={...}>` (connect an API like `@emotion/is-prop-valid`) or consider using transient props (`$` prefix for automatic filtering.)

> Warning: React does not recognize the `hasPadding` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `haspadding` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

### Steps to Reproduce

1. Run `yarn start`
2. Open `localhost:9000`
3. Open console
4. See errors

### Expected Behavior

No errors/warnings.

### Additional Information

This was introduced with the styled-components upgrade to v6 done in #2750: https://styled-components.com/docs/faqs#shouldforwardprop-is-no-longer-provided-by-default

It can be fixed using transient props (`$` prefix). See https://styled-components.com/docs/faqs#why-am-i-getting-html-attribute-warnings.

Alternatively, the previous styled-components v5 behavior can be restored. See https://styled-components.com/docs/faqs#shouldforwardprop-is-no-longer-provided-by-default.

It should be discussed what makes more sense for us/what are the implications of just using the behavior of v5.

PR #2843, opened 2024-06-25
2789: Fix the console warnings caused by the styled-components update

LeandraH:
### Short description

After updating `styled-components` to version 6, we started getting a lot of warnings in the browser console about passing in the wrong props to DOM elements (e.g. "Received `false` for a non-boolean attribute `submitted`.") That's pretty annoying when trying to debug using the browser console.

The problem is that `styled-components` now passes unknown props on to the HTML element it creates. I think that makes a lot of sense, in case there is a prop that just needs to be passed straight on that `styled-components` isn't aware of. So in our example, the HTML button received a `submitted="false"` as one prop and doesn't know what to do with it.

There are two ways of resolving this issue, one is to add a workaround in `styled-components`' `StyleSheetManager` and then props are not by default passed on to the HTML element as previously. Or the other one is to explicitly tell `styled-components` that a prop shouldn't be passed on by adding a `$` before the variable name. I think that is the cleaner way, so I implemented that here.

Unfortunately, there so far is no rule for `eslint` or `stylelint` that I could find to check for this behavior, so we will need to keep a bit of an eye on it. Luckily, the console will yell at us if we pass in a prop incorrectly.

### Proposed changes

- Added `$` to parameters that are passed to a `styled-components` component
- Removed the unsupported property `-webkit-print-color-adjust:exact`
- Changed the prop `ariaLabel` to `label` because `styled-components` is unhappy with `ariaLabel` and `react` is unhappy with `aria-label`.

### Side effects

- I did not find any but please particularly test implementations of `Button`

### Resolved issues

Fixes: #2789


ztefanie (approved):
Looks good. Thanks for fixing :rocket:

Issue #2845, opened 2024-06-30
No tooltip when hovering over calendar icon in event list


LeandraH:
### Describe the Bug

When I go to the list of events, I have calendar icons on the right of each event. When hovering over that, I used to get a tooltip telling me what the icon means. That is currently not happening.

### Steps to Reproduce

1. Go to 'https://integreat.app/augsburg/de/events'
2. Hover over the first calendar icon <img width="28" alt="image" src="https://github.com/digitalfabrik/integreat-app/assets/18513943/4e20642f-2836-4840-a6ab-e7b68d05025f">
3. See no tooltip

### Expected Behavior

A tooltip pops up when I hover over the icon.

### Actual Behavior

No tooltip pops up.

steffenkleinle:
@LeandraH is this fixed in #2847 as a side-effect of replacing our tooltip component with react-tooltip?

LeandraH:
> @LeandraH is this fixed in #2847 as a side-effect of replacing our tooltip component with react-tooltip?

Yes, exactly. Sorry, hadn't linked it yet

PR #2846, opened 2024-06-30
2845: Fix tooltip

LeandraH:
### Short description

The tooltip hover wasn't working anymore in the events list and the language switcher. The hover needs to have a parent element that is being hovered over to then show the newly visible tooltip. I added one into the `Tooltip` so that we don't need to necessarily add one around the implementations.

### Resolved issues

Fixes: #2845

### DIFF
--- a/web/src/components/Tooltip.tsx
+++ b/web/src/components/Tooltip.tsx
@@ -290,7 +290,7 @@ export default ({
       trigger={trigger}
       mediumViewportFlow={fixedMediumFlow}
       smallViewportFlow={fixedSmallFlow}>
-      {children}
+      <div>{children}</div>
     </TooltipContainer>
   )
 }


f1sh1918 (commented):
I'm not really sure if this is a good idea. we duplicate now div containers for all other (working) tooltips like social media sharing.
Instead we could just wrap the `DateIcon` with a div to get it work.
Not sure what is the better way to go

LeandraH:
That's true but I think having the extra div inside the `Tooltip` means that future uses of it are easier. I may have spent several hours trying to figure out why a `Portal` was breaking the `Tooltip` when really, I just needed an extra `div` somewhere in there 😅 

I guess I could put the `div` in the events and the language switcher, and add a note to the `Tooltip`? But I also don't think that one extra `div` is that bad.

LeandraH:
I'll fix this in #2847

PR #2854, opened 2024-07-16
Fix annoying iOS 'progress' listener warning

LeandraH:
### Short description

In the iOS emulator, particularly when in Testumgebung, I keep getting lots of the console warning "Sending 'progress' with no listeners registered". This is pretty annoying when trying to use the console for debugging.

According to https://reactnative.dev/docs/native-modules-ios#sending-events-to-javascript, we shouldn't be emitting events at all if no listeners for those events are registered, so I followed that advice.

image


steffenkleinle (approved):
LGTM, not tested

f1sh1918 (approved):
tested on ios, works fine 👍 
Thx for that fix :)

PR #2859, opened 2024-07-24
2755: Hide search results if no query is entered

steffenkleinle:
### Short description

Hide search results if no query is entered to avoid confusion.

### Proposed changes

- Hide search results and nothing found feedback if no query is entered to avoid confusion.
- Add launch.json vscode config files to .gitignore

### Side effects

None.

### Resolved issues

Fixes: #2755 

---


f1sh1918 (approved):
Tested on android, ios and chrome

LeandraH (approved):
Looks good, not tested

Issue #2865, opened 2024-08-01
Not found error flickering when clicking on push notification


steffenkleinle:
### Describe the Bug

A not found error is quickly shown when clicking on push notification before the actual news is loaded.

### Steps to Reproduce

1. Create a new news in the CMS and send a push notification
2. Keep the app in the foreground and directly click on the PN
3. See the error before the news actually loading

### Expected Behavior

No not found error should be displayed if the news is available.

### Actual Behavior

A not found error is displayed.

### Additional Information

This is due to a race condition of the initial news load without force refresh and the refresh triggered afterwards.
Therefore, we set loading to false before the refresh is done.

PR #2866, opened 2024-08-01
2865: Fix not found error flickering for clicking on push notification

steffenkleinle:
### Short description

Fix not found error flickering for clicking on push notification.

### Proposed changes

- Force refresh for initial local news load in NewsContainer
- Remove effect to refresh which caused a race condition

### Side effects

None.

### Resolved issues

Fixes: #2865.

---


native/src/hooks/useLoadCityContent.ts
```diff
@@ -27,6 +27,7 @@ import useSnackbar from './useSnackbar'
 type Params = {
   cityCode: string
   languageCode: string
+  refreshLocalNews?: boolean
```
LeandraH, line 30: 🔧  Can we call this something like `refreshForLocalNews`? `refreshLocalNews` sounds to me like it should be a function.

LeandraH (commented):
Unfortunately, I get the flickering on this branch on my iPhone :(

steffenkleinle:
> Unfortunately, I get the flickering on this branch on my iPhone :(

Are you sure you are on the right branch? Can you record this? This sadly does not give me any information on where to start. :see_no_evil:

LeandraH:
Just rechecked, definitely on the branch "2865-not-found-flickering". I do seem to remember this only being a problem with the test CMS though, and that the production CMS is fast enough?

https://github.com/user-attachments/assets/6e59e11b-1493-42c2-a11e-bca191a9406f

steffenkleinle:
> Just rechecked, definitely on the branch "2865-not-found-flickering". I do seem to remember this only being a problem with the test CMS though, and that the production CMS is fast enough?

@LeandraH thanks for the video, but it seems like you are testing different stuff than what the issue is about? The issue is about clicking a PN and then shortly seeing a not found error.
In your video, you are not doing that and only a `no languages available` error is displayed :see_no_evil: Am I missing something?

LeandraH:
Whooops, indeed, wrong video, sorry!


https://github.com/user-attachments/assets/59ec717e-bce8-4500-a4bb-7e37f3e68c3b

LeandraH (approved):
Sorry, this one also works once the hot-reloading works 🤦‍♀️  Tested on real iOS.

f1sh1918 (approved):
Tested on android. works fine

PR #2867, opened 2024-08-02
2732: Add available language to the test

LeandraH:
@steffenkleinle Something along these lines is what I was thinking. Just to have a small example of what the API response looks like and what we do with it.


steffenkleinle:
@LeandraH thank you for this, I really did not get what you were referring to but this makes a lot of sense, thanks! :heart:

Issue #2856, opened 2024-07-23
POI filter is too long for display and no scrolling available


dkehne:
### Describe the Bug
Opening the POI filter on 5,5" phone leads two categories to be cut (see screenshot) and it's not possible to scroll.

### Expected Behavior
Scrolling Option

### Actual Behavior

![signal-2024-07-23-105049](https://github.com/user-attachments/assets/e940f3d0-ab78-4362-a1f6-23ea74d6eb76)

f1sh1918:
@dkehne can you provide some information about the location you use here. Would be easier to reproduce

dkehne:
@f1sh1918 Munich City

steffenkleinle:
Done in #2869.

PR #2869, opened 2024-08-06
2856: poi filter is too long for display and no scrolling available

bahaaTuffaha:
### Short description

The POI filter on 5,5" phone leads two categories to be cut and it's not possible to scroll.

### Proposed changes

~- Used FlatList instead of .map() for build-in scrolling.~
~- Defined numbers of numColumns 3 to make it consistent on all devices~
~- Computed the flatListHeight dynamically from poiCategories length divided by numOfColumns to get number of rows multiplied by toggleButtonHeight.~
~- Used a ScrollView that wraps around opening hours section and categories section.~

- Fixed an issue with the scrollView at the modal where I replaced `{flex:1}` with `{flexGrow: 1}`

Tested on virtual devices with the following screen: 5.0, 5.5, 6.7 inches.

### Side effects

None.
~Note: I disabled scrollView from Modal (scrollView={false}) because VirtualizedLists doesn't like to be nested in ScrollViews.~

### Resolved issues

Fixes: [#2856](https://github.com/digitalfabrik/integreat-app/issues/2856).

---

### DIFF
--- a/native/src/components/PoiFiltersModal.tsx
+++ b/native/src/components/PoiFiltersModal.tsx
@@ -93,9 +93,9 @@ const PoiFiltersModal = ({
 }: PoiFiltersModalProps): ReactElement => {
   const { t } = useTranslation('pois')
   const { height } = Dimensions.get('window')
-  const toggleButtonHeight = 70
-  const flatListHeight = height - (toggleButtonHeight * poiCategories.length) / 2
-
+  const toggleButtonHeight = 100
+  const numOfColumns = 3
+  const flatListHeight = Math.ceil(height - (poiCategories.length / numOfColumns) * toggleButtonHeight)
   return (
     <Modal
       modalVisible={modalVisible}
@@ -121,7 +121,7 @@ const PoiFiltersModal = ({
           </Row>
           <FlatList
             data={poiCategories}
-            numColumns={3}
+            numColumns={numOfColumns}
             style={{ marginVertical: 10, height: flatListHeight }}
             contentContainerStyle={{ gap: 16, alignItems: 'center' }}
             columnWrapperStyle={{ gap: 16 }}
--- a/native/src/components/PoiFiltersModal.tsx
+++ b/native/src/components/PoiFiltersModal.tsx
@@ -68,9 +68,7 @@ const StyledToggleButton = styled(ToggleButton)`
 const StyledTextButton = styled(TextButton)`
   margin-bottom: 16px;
 `
-const StyledScrollView = styled.ScrollView`
-  height: 100%;
-`
+
 type PoiFiltersModalProps = {
   modalVisible: boolean
   closeModal: () => void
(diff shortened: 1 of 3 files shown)


native/src/components/PoiFiltersModal.tsx
```diff
@@ -55,11 +56,11 @@ const FlexEnd = styled.View`
...
-const TileRow = styled(Row)`
-  place-content: space-between center;
-  flex-wrap: wrap;
-  gap: 16px;
-`
+// const TileRow = styled(Row)`
+//   place-content: space-between center;
+//   flex-wrap: wrap;
+//   gap: 16px;
+// `
```
LeandraH, line 63: You can remove the commented out code.

native/src/components/PoiFiltersModal.tsx
```diff
@@ -110,17 +119,22 @@ const PoiFiltersModal = ({
...
           </Row>
-          <TileRow>
-            {poiCategories.map(it => (
+          <FlatList
+            data={poiCategories}
+            numColumns={numOfColumns}
+            style={{ marginVertical: 10, height: flatListHeight }}
+            contentContainerStyle={{ gap: 16, alignItems: 'center' }}
+            columnWrapperStyle={{ gap: 16 }}
+            renderItem={({ item }) => (
```
LeandraH, line 128: Very nice, thank you for making the code more humanly readable by changing `it` to `item`!

LeandraH (commented):
Super cool, thank you very much! The changes look good already but when I open it in an emulated Nexus 6, the tops of the buttons are cut off, so you might want to add some padding there. Another possibility, which I think I would prefer, is to make the entire page scrollable and not just the category icons. What do you think?

image

native/src/components/PoiFiltersModal.tsx
bahaaTuffaha: Does Nexus 6 emulation let you scroll?
hmm.. so do you prefer to see the whole buttons on small devices and make the scrolling on the entire page?

LeandraH (approved):
Nicely done, thank you!

LeandraH:
Oh, sorry, completely forgot: this should probably get a release note. You can find a template in the folder `release-notes`

steffenkleinle (commented):
Hey, thanks for your first PR in this project! :tada: Our `<Modal>` component should already implement scrolling behavior (including the title, which should also be scrollable in my opinion). For some reason it is not working though, could you perhaps investigate that?

bahaaTuffaha:
> Hey, thanks for your first PR in this project! 🎉 Our `<Modal>` component should already implement scrolling behavior (including the title, which should also be scrollable in my opinion). For some reason it is not working though, could you perhaps investigate that? I think that would be a little better solution compared to adding another scroll view.

Sure I will take a look at the modal...

bahaaTuffaha:
> Oh, sorry, completely forgot: this should probably get a release note. You can find a template in the folder `release-notes`
Sure after I investigate about the modal first..
I should add one at folder 2024.8.0 right?

steffenkleinle:
> > Oh, sorry, completely forgot: this should probably get a release note. You can find a template in the folder `release-notes`
> 
> Sure after I investigate about the modal first.. I should add one at folder 2024.8.0 right?

Absolutely no worries. No, it always goes in the `unrealeased` dir and is automatically moved to the correct folder on release.

native/src/components/PoiFiltersModal.tsx
```diff
@@ -111,13 +111,13 @@ const PoiFiltersModal = ({
...
-                key={it.id}
-                text={it.name}
-                active={it.id === selectedPoiCategory?.id}
-                onPress={() => setSelectedPoiCategory(it.id === selectedPoiCategory?.id ? null : it)}
-                Icon={<SvgUri uri={it.icon} />}
+                key={item.id}
+                text={item.name}
+                active={item.id === selectedPoiCategory?.id}
+                onPress={() => setSelectedPoiCategory(item.id === selectedPoiCategory?.id ? null : item)}
+                Icon={<SvgUri uri={item.icon} />}
```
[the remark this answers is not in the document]
bahaaTuffaha: @LeandraH told me its better for readability 🙂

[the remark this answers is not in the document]
steffenkleinle: Ohh, sorry about that then! We might have to discuss this again how to name vars in these cases :D The previous naming is probably from me based on kotlin naming conventions because I like the lean and short syntax :)

steffenkleinle (approved):
Good work!

Issue #2871, opened 2024-08-07
Header on licenses and disclaimer page are off


LeandraH:
### Describe the Bug

The header of the [licenses page](https://integreat.app/licenses) and the [disclaimer page](https://integreat.app/main-disclaimer) are off, the button to choose a region should be on the right hand side, like on our other web pages.
<img width="1147" alt="image" src="https://github.com/user-attachments/assets/4663b63c-1720-4446-8514-4d460d7a5b61">

PR #2872, opened 2024-08-07
2871: Fix header

LeandraH:
### Short description

Header before:
image

Header after:
image

### Side effects

- None

### Resolved issues

Fixes: #2871


web/src/components/Header.tsx
```diff
@@ -80,10 +80,10 @@ const ActionBar = styled.nav`
   justify-content: flex-end;
   gap: 12px;
   padding: 0 16px;
+  flex: 1 1 0;
```
steffenkleinle, line 83: :question: Any reason why we use this instead of just `flex: 1;`?

steffenkleinle (approved):
Tested on firefox

web/src/components/Header.tsx
LeandraH: I believe the flex-shrink is necessary to not get things like this: 
image

steffenkleinle: Hmmm. I can't reproduce that on my device, I get the same result with both `flex: 1;` and `flex: 1 1 0;` and both look very different than this :see_no_evil: 
Also, according to MDN, the last property defines `flex-basis`.

Also, the current behavior still seems to be not optimal: Even though there is more than enough space, the title already wraps since the nav item takes a lot of space:
image

Could you perhaps investigate that?

LeandraH: Ah, sorry, I thought you were asking about the `flex-shrink`. Though now that I think about it, I think that is `1` by default anyway... I'll take a look at what's going on with the smaller screen.

steffenkleinle (approved):
Looks good, thanks!

f1sh1918 (commented):
i think there is a more robust solution to fix this.
Even we don't use `rtl` here it would break if we do (just set flex-direction to `row-reverse`.

So there is a more robust solution i think

Remove the `margin-inline-start` and add `justify-content: space-between;` to the `Row` container.

Maybe i'm missing sth but this shouldn't create any issues afaics and works language independent

f1sh1918 (approved):

PR #2876, opened 2024-08-11
2130: Download only new resources

LeandraH:
### Short description

So far, we've been downloading assets (i.e. images and PDFs, really) once a day, even if they haven't changed. That was because files in the CMS kept their names even if they were edited; but that is not the case anymore, files that are changed in the CMS now get a new name each time. That means we don't need to download everything new every day, we can instead only download assets that aren't already cached because those are assets whose names are not in the cache yet.

On a side note, I decided to keep our current once-a-day-in-the-background check in `useLoadCityContent` the same because I think it's still not necessary to check for new files more than once a day. Any other opinions on that?

### Proposed changes

- When the `ResourceURLFinder` builds the map of new resources to fetch, it now filters out resources that are already in the cache.

### Side effects

- I don't see any.

### Resolved issues

Fixes: #2130 

---


native/src/utils/loadResourceCache.ts
```diff
@@ -51,8 +51,18 @@ const loadResourceCache = async ({
...
 
-  const fetchMap = resourceURLFinder.buildFetchMap(input, (url, urlHash) =>
-    buildResourceFilePath(url, cityCode, urlHash),
+  const currentResourceCache = await dataContainer.getResourceCache(cityCode, languageCode)
+
+  const currentlyCachedFiles: string[] = []
+  Object.values(currentResourceCache).forEach(filesObject => {
+    const fileUrls = Object.keys(filesObject)
+    fileUrls.forEach(url => currentlyCachedFiles.push(url))
+  })
```
steffenkleinle, line 60: :wrench: 
```suggestion
  const currentlyCachedFiles = Object.values(currentResourceCache).map(Object.keys).flat()
```

steffenkleinle (approved):
Tested on android, works great.

native/src/utils/loadResourceCache.ts
LeandraH: Woooooow, that is some serious JS magic! Thanks!

f1sh1918 (approved):
Tested on ios. works fine 👍

Issue #2875, opened 2024-08-09
Localization Issues at poi filter


bahaaTuffaha:
### Describe the Bug

When I switched the system language to Arabic language I noticed that the working hours at the poi filter is on the left also the education button in the categories it's not in Arabic.

### Steps to Reproduce

1. Change the device language to Arabic.
2. Go to Map.
3. Click on Filter location

### Additional Information

![Screenshot_1723208616](https://github.com/user-attachments/assets/3aac5305-1e76-4b1e-81c2-cb7bf846e524)

<details>
  <summary>Traceback</summary>

```

```

</details>

steffenkleinle:
Nice find! The missing translation for `Bildung`/`Education` is a CMS issue. Perhaps you could open an issue here: https://github.com/digitalfabrik/integreat-cms/issues

PR #2878, opened 2024-08-12
2875: localization issues at poi filter when choosing (münchen - Landeshauptstadt)

bahaaTuffaha:
### Short description

Opening hours Title's Alignment not on the right when arabic is chosen as the device language.

### Proposed changes

- flex-direction is needed when changing the text alignment so I added wrapped the SubTitle with Row component that has a flex-direction already.

Note: the category about education is missing the Arabic translation.

### Side effects

None

### Resolved issues

Fixes: #2875 

---


release-notes/unreleased/2875-Localization-Issues-at-poi-filter.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2875
+show_in_stores: false
+platforms: # relevant platforms, possible values: web, android and ios
+  - ios
+  - android
+en: Fixed issue with TextAlignment for RTL languages at the opening hours SubTitle
```
LeandraH, line 6: 🙃 Can you rephrase that for our users instead of our developers?
But also, I think this should be shown in the stores because it is again something users can notice.

LeandraH (commented):
Nicely done! Just the release note needs a fix and then we're good to go :)

release-notes/unreleased/2875-Localization-Issues-at-poi-filter.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2875
+show_in_stores: true
+platforms: # relevant platforms, possible values: web, android and ios
+  - ios
+  - android
+en: Fixed an issue at the map filter where opening hours not aligned correctly
```
LeandraH, line 6:
```suggestion
en: Fixed an issue at the map filter where opening hours were not aligned correctly
```

LeandraH (approved):
Awesome, thank you! Tested in real iOS

steffenkleinle (approved):
Tested on Android :tada:

ztefanie:
No need to adjust anything, but please never use non standard chars in branch names, could lead to nasty errors:
for ü just use "ue" (for ä use ae, for ö use oe)

Issue #2870, opened 2024-08-06
POI filters: Button to only show currently open locations can be cut off


LeandraH:
### Describe the Bug

When opening the overlay for filtering POIs on my iPhone 13 Mini with the phone language being set to French, the toggle button to only show currently open locations is cut off.

### Steps to Reproduce

1. Be on a small device (in this example, an iPhone 13 Mini)
2. Set the device language to French
3. Go to Maps
4. Click on the button to open the filter overlay ('Filter les lieux')
5. See error

### Expected Behavior

The description of the button should have a line break if the button otherwise doesn't fit onto the page.

### Actual Behavior

The description is all on one line, so if the language is one that takes up more space, the toggle gets pushed off the page.

### Additional Information
![Capture d’écran   2024-08-06 à 12 09 48](https://github.com/user-attachments/assets/70b733a6-4dab-454d-8ab2-03066279ab0d)

bahaaTuffaha:
This issue has been fixed at #2856 Tested on virtual device Android 4.65' screen

bahaaTuffaha:
Mistaken the button cut off with the toggle button I will reopen this .. my bad😅

PR #2879, opened 2024-08-12
2870: POI filters: Button to only show currently open locations can be cut off

bahaaTuffaha:
### Short description

The toggle button is cut off due to long sentences from languages like French.

### Proposed changes

- Added a max-width to StyledText to leave some space to the toggle button.
- Added `alignItems: 'center'` to Row style just to make the text,icon and toggle aliened vertically on the same level. 

### Side effects

None

### Resolved issues

Fixes: #2870 

---


release-notes/unreleased/2870-POI-filters-fix-toggle-button.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2870
+show_in_stores: false
+platforms: # relevant platforms, possible values: web, android and ios
+  - android
+  - ios
+en: A Fix for the toggle button at the poi filter.
```
LeandraH, line 6:
```suggestion
en: A fix for the toggle button in the map filter
de: Der Toggle-Button im Kartenfilter wurde verbessert
```
We call them pois internally but most users probably don't know that that means point of interest.

LeandraH (commented):
Thanks for the PR! It looks pretty good already but I think the button should ideally be in the same line as the icons under it. Maybe you can do something with flex positioning instead of a hard-coded width?

IMG_255E967AC9BF-1

LeandraH (approved):
Nicely done, thank you! Tested in real iOS

steffenkleinle (approved):
Tested on android, nice solution! :)

LeandraH:
By the way, you can enable auto-merge so that you don't have to come back to the PR to merge it when the pipeline is finished :)

Issue #2880, opened 2024-08-13
Weekly and monthly recurring events are sometimes off by one day


david-venhoff:
### Describe the Bug

See https://github.com/digitalfabrik/integreat-cms/issues/2988

### Steps to Reproduce

1. Go to https://admin.integreat-app.de/testumgebung/events/de/6252/edit/
2. See that the date is set to thursday Aug 15th
3. Go to https://integreat.app/testumgebung/de/events/jeden-donnerstag
4. See dates set to be 16th and then every week

### Expected Behavior

Date in the app should be date set in the cms

### Actual Behavior

Date in the app is +1

### Additional Information

https://github.com/digitalfabrik/integreat-cms/issues/2988#issuecomment-2285938397

LeandraH:
Looks so far like this happens when the start day is during summer time, and the start time is before 2 am, including all-day events. The offset during summer time in the German time zone is 2 hours, we're deducting and re-adding those 2 hours because our rrule package doesn't support time zones. I assume the deducting of 2 hours puts us on the previous day and for reasons that I haven't figured out yet, adding the 2 hours back in doesn't put us back on the correct day.

LeandraH:
The problem is that deducting the 2 hours puts us on the previous day, then rrule tells us which day is the next one based on that time, but then we add the 2 hours back in and that can shift us to the next day.
Example: We have an event with starting day 14.8.2024 at 0:00, in UTC that is 13.8.2024 at 22:00. We now want to have the next Friday from rrule, that is 16.8.2024 at 22:00. In Berlin time, that is 17.8.2024 at 0:00, i.e. the Saturday instead of the Friday.

PR #2882, opened 2024-08-14
2880: Fix repeating dates. Again.

LeandraH:
### Short description

The repeating dates were off again. Reminder, the underlying problem is that the package `rrule` handles time zones by ignoring them and pretending that everything is in UTC. This time, we had the problem that e.g. all-day events arrive with a starting time of 0:00+2 which translates to 22:00 on the previous day in UTC. If we then go and get for example the next Saturday from that, we get Saturday at 22:00 in UTC, which translates to midnight on Sunday in local time.

Links for testing: http://localhost:9000/testumgebung/de/events/jeden-donnerstag and https://cms-test.integreat-app.de/testumgebung/events/de/4596/edit/

Note: I haven't found a well-supported alternative package, but man, would I be happy to switch.

### Proposed changes

- Move the start date in the recurrence rule to local time, then get the next recurrences, then set those back to UTC time
- Added a couple of more tests for different situations I could think of

### Side effects

- It should be none. I decided against keeping the offset on the recurrences to keep them in the existing (and frankly, correct) date format.

### Resolved issues

Fixes: #2880

### DIFF
--- a/shared/api/models/DateModel.ts
+++ b/shared/api/models/DateModel.ts
@@ -83,9 +83,6 @@ class DateModel {
     // bigger than the distance from midnight (e.g. 1 am with a 2h offset during CET summer time)
     // https://github.com/jkbrzt/rrule#important-use-utc-dates
     const localRecurrenceRule = this.getRecurrenceRuleInLocalTime(this.recurrenceRule)
-    if (!localRecurrenceRule) {
-      return [this]
-    }
 
     return localRecurrenceRule
       .between(minDate, maxDate, true, (_, index) => index < count)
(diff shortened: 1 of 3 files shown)


shared/api/models/DateModel.ts
```diff
@@ -128,23 +138,16 @@ class DateModel {
...
-  private currentDateToRrule(): Date {
-    // rrule is not correctly handling timezones and is expecting dates in "local time", i.e. in the timezone of dstart
-    // Therefore we have to manually apply timezone offset, e.g. for daylight savings time
-    // https://github.com/jkbrzt/rrule#important-use-utc-dates
-    // Also include dates happening at the moment by subtracting the duration
-    return DateTime.now()
-      .minus(this._duration)
-      .minus({ minutes: this._offset - DateTime.now().offset })
-      .toJSDate()
+  private getRecurrenceRuleInLocalTime(recurrenceRule: RRuleType): RRuleType | undefined {
```
steffenkleinle, line 141: :wrench: No need for `undefined`. Please also remove check above.
```suggestion
  private getRecurrenceRuleInLocalTime(recurrenceRule: RRuleType): RRuleType {
```

steffenkleinle (approved):
Looks a lot simpler than before, great job! I don't understand everything happening and why it works, but looks good and seems to work! Great job :tada:

shared/api/models/DateModel.ts
LeandraH: Good catch, thanks! Too much refactoring 🤪

f1sh1918 (approved):
nicely done 👍

Issue #2850, opened 2024-07-03
POI filter button shines through language selection overlay


LeandraH:
### Describe the Bug

When I open the language selection overlay
- in a region that has enough languages that it's two rows of languages to select
- and from the POIs route,
the button to filter POIs with is still visible through the language overlay.

### Steps to Reproduce

1. Go to 'https://integreat.app/lkaugsburg/de/locations'
2. Click on the icon to change the language
3. See error

### Expected Behavior

The POI filter button is hidden under the language selection overlay.

### Actual Behavior

The POI filter button is visible through the language selection overlay.

### Additional Information

![Image](https://github.com/digitalfabrik/integreat-app/assets/18513943/d9557e8e-e98f-4b77-9c28-ba14031e77ed)

PR #2883, opened 2024-08-14
2850: POI filter button shines through language selection overlay

bahaaTuffaha:
### Short description

POI filter button above the  language selection overlay.

### Proposed changes

- Changed the z-index for OverlayContainer from 1 to 0 Located at MapView component.

### Side effects

None

### Resolved issues

Fixes: #2850 

---


release-notes/unreleased/2850-POI-filter-button-shines-through.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2850
+show_in_stores: true
```
LeandraH, line 2: Since this is only a web issue, it doesn't need to be shown in the stores :)

LeandraH (approved):
Very nice, thanks! Tested in Firefox

release-notes/unreleased/2850-POI-filter-button-shines-through.yml
steffenkleinle: Accordingly, no German translation is necessary :) We only need German translations for ios/android issues that are shown in the stores.

web/src/components/MapView.tsx
```diff
@@ -43,7 +43,7 @@ const OverlayContainer = styled.div`
   display: flex;
   padding: 12px 8px;
   flex: 1;
-  z-index: 1;
+  z-index: 0;
```
steffenkleinle, line 46: As the default is zero, this could be omitted, correct?
```suggestion
```

steffenkleinle (approved):
Tested on firefox

web/src/components/MapView.tsx
LeandraH: Good point!

LeandraH:
@bahaaTuffaha Though in this case, please still remove the `z-index` and test it once more before merging 🙂

Issue #2831, opened 2024-06-14
Improve language selection header item (desktop)


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user I want to be able to easily change the language of the app. Therefore I want it to be intuitive where to do that.

**Describe the solution you'd like**
We want to highlight the language selection on web (desktop) by showing the full language name. On mobile everything should stay as it is for now.
See the design [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3033-1392&t=QrdQnY7YaUcvl3jl-1).

**Describe alternatives you've considered**
Keep it as it is.

**Additional context**
Issue for native/mobile with additional discussion (still TBD): #2549.

PR #2884, opened 2024-08-15
2831: Improve language selection header item (desktop)

bahaaTuffaha:
### Short description

We want to highlight the language selection on web (desktop) by showing the full language name. On mobile everything should stay as it is for now.
See the design [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3033-1392&t=QrdQnY7YaUcvl3jl-1).

### Proposed changes

- Modified the `HeaderActionItemDropDown` component to accept a props called innerText.
- If `innerText` has a value, the component will display the `ActionBox` otherwise, it will remain unchanged.
- Used config.hasRTLScript to change the placement of the text and icon in the box depending on language direction.

**Note** : There is some weird duplication of the Tooltip when hovering above the button.

### Side effects

None

### Resolved issues

Fixes: #2831 

---


web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -44,7 +44,10 @@ const HeaderLanguageSelectorItem = ({
     }
 
     return (
-      <HeaderActionItemDropDown iconSrc={LanguageIcon} text={t('changeLanguage')}>
+      <HeaderActionItemDropDown
+        iconSrc={LanguageIcon}
+        text={t('changeLanguage')}
+        innerText={selectorItems.find(item => item._code === activeItemCode)?._name}>
```
steffenkleinle, line 50:
```suggestion
        innerText={selectorItems.find(item => item.code === activeItemCode)?.name}>
```

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -11,19 +13,44 @@ const StyledIcon = styled(Icon)`
   width: 28px;
   height: 28px;
 `
+const ActionBox = styled(Button)<{ $isRTL?: boolean }>`
+  height: 36px;
+  min-width: 122px;
+  display: flex;
+  border-color: black;
```
steffenkleinle, line 20:
```suggestion
  border-color: ${props => props.theme.colors.textColor};
```

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -11,19 +13,44 @@ const StyledIcon = styled(Icon)`
...
+  border-width: 1px;
+  border-style: solid;
+  border-radius: 4px;
+  justify-content: center;
+  align-items: center;
+  flex-direction: ${props => (props.$isRTL ? 'row-reverse' : 'row')};
```
steffenkleinle, line 26: You can just use the following, no need to pass this value:
```suggestion
  flex-direction: ${props => (props.theme.contentDirection === 'rtl' ? 'row-reverse' : 'row')};
```

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -11,19 +13,44 @@ const StyledIcon = styled(Icon)`
...
+`
+
+const StyledText = styled.span`
+  font-weight: bold;
+  font-size: 14px;
+  margin: 0 5px;
+  letter-spacing: 1.25px;
+`
```
steffenkleinle, line 34: I could reproduce the issue with two tooltips... This is because this style is also applied to the tooltip text :see_no_evil: If you remove `font-weight` and `letter-spacing`, the issue is gone. Obviously not a good solution though. This might be fixed with the new tooltip introduced in #2847 by @LeandraH. So perhaps it makes sense to wait for that :see_no_evil:

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -11,19 +13,44 @@ const StyledIcon = styled(Icon)`
...
+  border-radius: 4px;
+  justify-content: center;
+  align-items: center;
+  flex-direction: ${props => (props.$isRTL ? 'row-reverse' : 'row')};
+`
+
+const StyledText = styled.span`
+  font-weight: bold;
+  font-size: 14px;
+  margin: 0 5px;
```
steffenkleinle, line 32: I'd prefer a `gap` on the parent component, what do you think?

steffenkleinle (commented):
Tested on firefox. I couldn't find any issues apart from the duplicated tooltip.

ztefanie (commented):
Tested looks good, beside the double tooltip problem @steffenkleinle mentioned

web/src/components/HeaderActionItemDropDown.tsx
bahaaTuffaha: So I will put it on hold for now..

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -11,19 +13,44 @@ const StyledIcon = styled(Icon)`
   width: 28px;
   height: 28px;
 `
+const ActionBox = styled(Button)<{ $isRTL?: boolean }>`
+  height: 36px;
+  min-width: 122px;
```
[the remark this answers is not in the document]
bahaaTuffaha: Some languages takes more space and the design in Figma says 122 width 😅
Do I increase it?

web/src/components/HeaderActionItemDropDown.tsx
steffenkleinle: i think that PR should be merged soonish, hopefully today :) Will let you know

LeandraH: It's merged 🥳

LeandraH: If you pull from `main`, you should get the new tooltip which lets you use components or text :)

web/src/components/HeaderActionItemDropDown.tsx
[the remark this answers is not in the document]
steffenkleinle: One minor thing: In general we tend to use values which are a power of two if applicable, e.g., 2, 4, 8, 16,.... If those are not sufficient, values like 12 and 24 might be used as well. This allows us to have similar paddings, sizes, ... At some point we plan to introduce a design system with fixed values for these things, which is easier if we follow the convention now already.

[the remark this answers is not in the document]
bahaaTuffaha: Ok then the padding will be 4px 6px instead 3px 5px ...

web/src/components/HeaderActionItemDropDown.tsx
bahaaTuffaha: Ok thanks..

ztefanie (approved):
Thanks for making these adjustments so fast. Looks fine, good work. :rocket: 

Just one small hint: Please do not use special characters (like brackets) in branch names. No need to adjust it, just for the future

web/src/components/HeaderActionItemDropDown.tsx
```diff
@@ -12,19 +12,41 @@ const StyledIcon = styled(Icon)`
...
 `
+const ActionBox = styled(Button)`
+  padding: 4px 6px;
+  display: flex;
+  border: 1px solid ${props => props.theme.colors.textColor};
+  border-radius: 4px;
+  justify-content: center;
+  align-items: center;
+  flex-direction: ${props => (props.theme.contentDirection === 'rtl' ? 'row-reverse' : 'row')};
+  gap: 0 5px;
```
steffenkleinle, line 23:
```suggestion
  gap: 4px;
```

steffenkleinle (approved):
Well done! Tested on firefox :)

Issue #2833, opened 2024-06-17
Improve useLocalStorage hook


f1sh1918:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2887, opened 2024-08-21
2833: Refactor useLocalStorage and ChatController

steffenkleinle:
### Short description

Adjusted and refactored the `useLocalStorage` hook and `ChatController` component.

I couldn't reproduce the issue that the chat gets deleted after some time (apparently reported by someone from the city of munich). Perhaps this PR prevents this.

### Proposed changes

- Refactor `useLocalStorage` for type safety and providing initial values
- Merge `ChatInitializedView` into `ChatController` for better handling of sending/receiving messages
- Prevent unnecessary message refreshes by directly parsing and adding the sent message to the messages

### Side effects

Hopefully none.

### Resolved issues

Fixes: #2833.

### Testing

Specifically test the chat function, both with old and new chats (e.g. in private windows). Check loading spinners and otherwise unexpected behavior (available in testumgebung).

Quickly check consent for playing external content.

---


web/src/components/ChatController.tsx
```diff
@@ -14,48 +19,56 @@ type ChatControllerProps = {
 }
 
 const LOCAL_STORAGE_ITEM_CHAT_MESSAGES = 'Chat-Device-Id'
+const POLLING_INTERVAL = 16000
+
 const ChatController = ({ city, language }: ChatControllerProps): ReactElement => {
```
steffenkleinle, line 24: I merged ChatInitializedView into this component for easier handling of device id, sending of messages and updating the messages.

shared/api/endpoints/createSendChatMessageEndpoint.ts
```diff
@@ -23,5 +24,12 @@ export default (baseUrl: string): Endpoint<ParamsType, Array<ChatMessageModel>>
       formData.append('message', message)
       return formData
     })
-    .withMapper(() => [])
```
LeandraH, line 26: ❓ I don't understand why the mapper here used to be empty?

LeandraH (approved):
Very nice! Tested on Firefox

shared/api/endpoints/createSendChatMessageEndpoint.ts
steffenkleinle: I guess because the response was not used before. There is one endpoint to send a message (this one) and another one to fetch all messages. Instead of using the return of this message, so far only the other endpoint was refetched to get the new message as well.

LeandraH: Ah, I see, thanks :)

f1sh1918 (approved):
nice refactor works fine.

Tested videos and chat

Issue #2888, opened 2024-08-21
Warnings cluttering tests about invalid CSS selectors (react-tooltip)


steffenkleinle:
### Describe the Bug

We get a lot of warnings in our tests at the moment, e.g.:

> [react-tooltip] "#layout:search" is not a valid CSS selector
>  [react-tooltip] "#layout:changeLanguage" is not a valid CSS selector

Investigate these and fix the warnings.
Probably introduced in #2847.

### Steps to Reproduce

1. Run our tests in `web`
2. See log

### Expected Behavior

No warnings are shown.

### Actual Behavior

A lot of warnings are shown.

### Additional Information

None.

PR #2889, opened 2024-08-21
2888: Fix warnings from react-tooltip

LeandraH:
### Short description

Sorry about that, I didn't notice all the warnings cluttering up the console from react-tooltips when running tests. Some more mocking of the tooltip fixes it.

### Resolved issues

Fixes: #2888 

---


steffenkleinle (approved):
:sponge: :rocket:

ztefanie (approved):
Thanks for fixing. Maybe we should think about enabling the automock feature in the future. I think we will sure forget about mocking this if in future tests if there is only a warning.

PR #2892, opened 2024-08-21
2891: Override meta description

f1sh1918:
### Short description

Override meta description for search engine purposes

### Proposed changes

- override meta description
- import overrides in buildConfig

### Side effects

- none

### Resolved issues

Fixes: #2891

---

### DIFF
--- a/translations/override-translations/obdach.json
+++ b/translations/override-translations/obdach.json
@@ -1,0 +1,7 @@
+{
+  "categories": {
+    "de": {
+      "metaDescription": "{{appName}} - die mehrsprachige Plattform, die wohnungslose und von Wohnungslosigkeit bedrohte Menschen mit wichtigen Anlaufstellen wie Beratungen und Notunterkünften verbindet."
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


translations/override-translations/obdach.json
```diff
@@ -0,0 +1,7 @@
+{
+  "categories": {
+    "de": {
+      "metaDescription": "{{appName}} - die mehrsprachige Plattform, die wohnungslose und von Wohnungslosigkeit bedrohte Menschen mit wichtigen Anlaufstellen wie Beratungen und Notunterkünften verbindet."
+    }
+  }
+}
```
LeandraH, line 7: Man, that description sounds like a lawyer wrote it but I guess that's not up to us to decide...

LeandraH (approved):

translations/override-translations/obdach.json
f1sh1918: ugh, don't mess with @FritjofK :D

LeandraH: Good point, I don't really have a better suggestion either :D

steffenkleinle (approved):
Changed the metadescription for landing too.
@f1sh1918 @FritjofK keep in mind that this is probably already handled or overwritten by the CMS.

Issue #2868, opened 2024-08-05
Android e2e tests failing


steffenkleinle:
### Describe the Bug

Recently, our android e2e tests are frequently failing and are blocking this weeks release. Investigate and fix this.

### Steps to Reproduce

Examples:
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11676/workflows/1250bcf0-ec31-4755-a519-fec60b923aed/jobs/63639
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11676/workflows/da426656-3b16-4bb8-b7d9-da3b7c9747e7/jobs/63631
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11769/workflows/ab66e223-f7ae-4239-9e42-7b4024011574/jobs/64036
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11774/workflows/2d91c832-0b2f-4da5-9b61-861d5535653f/jobs/64083
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/11778/workflows/785d40c9-3f7e-4c7f-b8e1-6b5c3fa60efb/jobs/64118

### Additional Information

None.

PR #2893, opened 2024-08-22
2868: Fix e2e android

ztefanie:
### Short description
Fix e2e android tests

### Proposed changes
- Fixed local tests by adding a description of a workaround on how to get appium work locally. As our project in not working with npm, but appium tries to, we need to configure appium android driver globally. Fore more context see how [How Appium handels extenstions](https://appium.io/docs/en/2.0/guides/managing-exts/)
- Fixed race condition of permission dialog and landing page
- Small adjustments
- Runned this 5x times and all tests succeeded each time, see [here](https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app?branch=2868-fix-e2e-tests)

### Idea
Maybe it would make sense to rename the "Testumgebung E2e", to "A-Testumgebung E2E" or "E2E - Testumgebung", as the scrolling takes quite long and we could speed the E2E tests up by changing this.

### Side effects

None.

### Testing:
not testable

### Resolved issues

Fixes: #2868 

---


e2e-tests/native/test/pageobjects/page.ts
```diff
@@ -6,6 +6,6 @@ export class Page {
   }
 
   public async get(): Promise<true | void> {
-    return $(`~${this.pageId}`).waitForExist()
+    return $(`~${this.pageId}`).waitForExist({ timeout: 10000 })
```
ztefanie, line 9: I can also undo this, it was causing the errors in approximately less then 10%, so we can also undo this, but as the tests take very long I think waiting one time 5s more, but prevent a retry is worth it, but I am open for discussions on this.

steffenkleinle:
> ### Idea
> 
> Maybe it would make sense to rename the "Testumgebung E2e", to "A-Testumgebung E2E" or "E2E - Testumgebung", as the scrolling takes quite long and we could speed the E2E tests up by changing this.
> ### Side effects

Great idea, do you want to do that?

e2e-tests/native/test/pageobjects/page.ts
steffenkleinle: Sounds good to me :+1:

steffenkleinle (approved):
LGTM

ztefanie:
> Maybe it would make sense to rename the "Testumgebung E2e", to "A-Testumgebung E2E" or "E2E - Testumgebung", as the scrolling takes quite long and we could speed the E2E tests up by changing this.

Created a issue for this here: https://github.com/digitalfabrik/integreat-app/issues/2915

e2e-tests/native/test/utils/deviceUtils.ts
```diff
@@ -1,5 +1,11 @@
...
 
 export const allowNotifications = async (): Promise<void> => {
-  await $(new Selector().ByText('Allow').build()).click()
+  try {
+    const allow = await $(new Selector().ByText('Allow').build())
+    await allow.waitForDisplayed()
+    allow.click()
+  } catch (error) {
+    console.warn('Permission request was not displayed within the timeout:', error)
+  }
```
LeandraH, line 10: 💭  This looks to me like it would mean that the tests would pass even if the request for permissions never show up. If that is the case, and if that is what we want to do, we can just revert #2761

ztefanie: Yes you are right. I think if we want to check if the permission dialog is displayed we should just add a own test for this, but we do E2E tests only for really a few very basic use cases, so I think it is not needed.

Also reverting the previous PR does not really make sense, as we want to click the "apply" button and proceed.

LeandraH: I'm okay with adding a separate test for whether the permission is requested but I'm not comfortable with leaving the test like this because it doesn't actually fail if the permission isn't requested. Instead, we have a test that looks like it tests for the permissions but doesn't, leading to a false sense of security, and a completely useless test.

ztefanie: I am confused now. We have four tests "changeLanguage", "navigateToLeafCategory", "navigateToOtherLocation" and "navigateToSearchResults". Neither of them does have a test for requesting permissions, it is only one step in the "should open a city on location selection" test. So I would not expect this test to check for correct permission request handling.

LeandraH: You're right, we don't actually have a separate test for the notifications permission. I'll open an issue for it.

LeandraH (approved):

PR #2895, opened 2024-08-26
2223: Add link to POI from events

LeandraH:
### Short description

When creating an event in the CMS, you can choose a POI as a possible location. These now get displayed as links to the POI when you look at the event in the app too.

One weird thing: after a rebuild in the native app on my machine, the `locationPath` only shows up after a pull-to-refresh. Any ideas why? Anybody else have the same problem?

### Proposed changes

- Add the `locationPath` to the `EventModel`
- Turn the address in the event page / route into a link to the POI if we get a `locationPath`

### Side effects

- If no `locationPath` is present, nothing should change, not even with RTL but please double-check :)

### Testing

I would recommend using http://localhost:9000/testumgebung/de/events/test-veranstaltung-mit-link-zu-ort for testing.

### Resolved issues

Fixes: #2223 

---

### DIFF
--- a/native/src/components/PageDetail.tsx
+++ b/native/src/components/PageDetail.tsx
@@ -1,32 +1,66 @@
 import React, { ReactElement } from 'react'
+import { Pressable, Text } from 'react-native'
 import styled from 'styled-components/native'
 
+import { InternalPathnameParser } from 'shared'
+
+import buildConfig from '../constants/buildConfig'
 import { contentDirection } from '../constants/contentDirection'
+import useNavigate from '../hooks/useNavigate'
+
+const DetailContainer = styled.View<{ language: string }>`
+  flex-direction: ${props => contentDirection(props.language)};
+  font-family: ${props => props.theme.fonts.native.contentFontRegular};
+  color: ${props => props.theme.colors.textColor};
+`
 
 const Identifier = styled.Text`
   font-family: ${props => props.theme.fonts.native.contentFontBold};
   color: ${props => props.theme.colors.textColor};
 `
-const DetailContainer = styled.Text<{ language: string }>`
-  display: flex;
-  flex-direction: ${props => contentDirection(props.language)};
-  font-family: ${props => props.theme.fonts.native.contentFontRegular};
-  color: ${props => props.theme.colors.textColor};
+
+const StyledButton = styled.Pressable`
+  /* Offset for bold sibling + underline */
+  margin-top: 1px;
+`
+
+const ButtonText = styled.Text`
+  color: ${props => props.theme.colors.linkColor};
+  text-decoration: underline;
+  text-decoration-color: ${props => props.theme.colors.linkColor};
+`
+
+const StyledText = styled.Text`
+  /* Offset for bold sibling */
+  margin-top: 2px;
 `
 
 type PageDetailProps = {
   identifier: string
   information: string
   language: string
+  locationPath?: string | null
 }
 
-const PageDetail = (props: PageDetailProps): ReactElement => {
-  const { identifier, information, language } = props
+const PageDetail = ({ identifier, information, language, locationPath }: PageDetailProps): ReactElement => {
+  const { navigateTo } = useNavigate()
+
+  if (locationPath) {
+    const parsedRoute = new InternalPathnameParser(locationPath, language, buildConfig().featureFlags.fixedCity).route()
+    return (
+      <DetailContainer language={language}>
+        <Identifier>{identifier}: </Identifier>
+        <StyledButton onPress={() => navigateTo(parsedRoute)}>
+          <ButtonText>{information}</ButtonText>
+        </StyledButton>
+      </DetailContainer>
+    )
+  }
 
   return (
     <DetailContainer language={language}>
       <Identifier>{identifier}: </Identifier>
-      {information}
+      <StyledText>{information}</StyledText>
     </DetailContainer>
   )
 }
--- a/native/src/components/PageDetail.tsx
+++ b/native/src/components/PageDetail.tsx
@@ -1,5 +1,4 @@
 import React, { ReactElement } from 'react'
-import { Pressable, Text } from 'react-native'
 import styled from 'styled-components/native'
 
 import { InternalPathnameParser } from 'shared'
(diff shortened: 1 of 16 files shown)


native/src/components/PageDetail.tsx
```diff
@@ -1,27 +1,55 @@
...
   identifier: string
   information: string
   language: string
+  locationPath?: string | null
 }
 
-const PageDetail = (props: PageDetailProps): ReactElement => {
-  const { identifier, information, language } = props
+const PageDetail = ({ identifier, information, language, locationPath }: PageDetailProps): ReactElement => {
+  const { navigateTo } = useNavigate()
```
steffenkleinle, line 40: :upside_down_face: Personal preference:
```suggestion
  const { navigateTo } = useNavigate()
  const route = locationPath ? new InternalPathnameParser(locationPath, language, buildConfig().featureFlags.fixedCity).route() : null
```

native/src/components/PageDetail.tsx
```diff
@@ -1,27 +1,55 @@
...
+
+  if (locationPath) {
+    const parsedRoute = new InternalPathnameParser(locationPath, language, buildConfig().featureFlags.fixedCity).route()
+    return (
+      <DetailContainer language={language}>
+        <Identifier>{identifier}: </Identifier>
+        <StyledButton onPress={() => navigateTo(parsedRoute)}>
+          <ButtonText>{information}</ButtonText>
+        </StyledButton>
```
steffenkleinle, line 49: :upside_down_face: Personal preference:
```suggestion
        {route ? (
          <StyledButton onPress={() => navigateTo(parsedRoute)}>
            <ButtonText>{information}</ButtonText>
          </StyledButton>
        )  : information}
```

native/src/components/PageDetail.tsx
```diff
@@ -1,27 +1,55 @@
...
+  color: ${props => props.theme.colors.linkColor};
+  text-decoration: underline;
+  text-decoration-color: ${props => props.theme.colors.linkColor};
+`
+
 type PageDetailProps = {
   identifier: string
   information: string
   language: string
+  locationPath?: string | null
```
steffenkleinle, line 36: :wrench: Is there any reason this needs to be `location`-specific? Can't we just name it `path` or directly use `routeInformation`? If we choose to keep it location specific, the naming should be consistent and be `poiPath`.

native/src/utils/DatabaseConnector.ts
```diff
@@ -109,6 +109,7 @@ type ContentEventJsonType = {
   }
   location: LocationJsonType<number | null> | null
   featured_image: FeaturedImageJsonType | null | undefined
+  location_path: string | null
```
steffenkleinle, line 112: :wrench: The naming should be `poi_path`.

shared/api/models/EventModel.ts
```diff
@@ -14,6 +14,7 @@ class EventModel extends ExtendedPageModel {
   _location: LocationModel<number | null> | null
   _excerpt: string
   _featuredImage: FeaturedImageModel | null
+  _locationPath: string | null
```
steffenkleinle, line 17: :wrench: See above, the naming should be consistent with the rest of our codebase:
```suggestion
  _poiPath: string | null
```

web/src/components/PageDetail.tsx
```diff
@@ -1,19 +1,27 @@
...
 
+const StyledLink = styled(Link)`
+  color: ${props => props.theme.colors.linkColor};
+  text-decoration: underline;
+`
+
 type PageDetailProps = {
   identifier: string
   information: string
+  pathToInformation?: string | null
```
steffenkleinle, line 18: :upside_down_face: As the `ToInformation` does not really add any additional information, I'd prefer just `path`.
```suggestion
  path?: string | null
```

web/src/components/PageDetail.tsx
```diff
@@ -1,19 +1,27 @@
...
+import Link from './base/Link'
+
 const Identifier = styled.span`
   font-weight: 700;
 `
 
+const StyledLink = styled(Link)`
+  color: ${props => props.theme.colors.linkColor};
+  text-decoration: underline;
+`
```
steffenkleinle, line 13: :upside_down_face: I think it would be cool if we could just use this style by passing one prop to the `Link` component as this is reused in a few places and it would make adjusting and using the style easier and guarantees consistent styling. I am thinking of a boolean prop called `styled` or `highlighted` or similar which uses this style if true and no styling if false.

Since this was not introduced in this PR, you don't have to do this obviously.

steffenkleinle (changes requested):
Tested on firefox and Android. Works as expected :+1: 

On Android, the link seems a little bit off though (and I am not sure the usual link styling fits for native apps as I think it indicates an external link, but I am fine with it):

native/src/components/PageDetail.tsx
LeandraH: I just kept the name as we get it from the CMS but I can rename it :)

native/src/components/PageDetail.tsx
```diff
@@ -1,32 +1,60 @@
...
-import { contentDirection } from '../constants/contentDirection'
+import { InternalPathnameParser } from 'shared'
+
+import buildConfig from '../constants/buildConfig'
+import useNavigate from '../hooks/useNavigate'
+
+const DetailContainer = styled.View`
+  flex-direction: row;
+  color: ${props => props.theme.colors.textColor};
+  align-items: center;
```
LeandraH, line 12: The direction reversal isn't necessary anymore since it's now a View instead of a Text. So I also removed the tests for that.

LeandraH:
> (and I am not sure the usual link styling fits for native apps as I think it indicates an external link, but I am fine with it)

I decided to add the blue and the underline to make it clear that it is a link at all. We mark external links with the same style plus an icon to show that it's external. That seems like standard practice to me, to have internal and external links styled the same way, potentially with an extra icon for external ones.

f1sh1918 (approved):
Tested on chrome, android and ios. works fine for me 👍

steffenkleinle (approved):
Thanks! Tested on android and firefox, works like a charm.

Issue #2886, opened 2024-08-21
Not working Link in Aschaffenburg


Docmale:
### Describe the Bug
This is a lowest priority bug, since the effect is max hidden and self-solving by additional tapping on same link. Found it on Pixel 3-emulator (android 14) and can reproduce on Samsung A53 (14) where te effect is something different, see steps.

### Steps to Reproduce

1. Open 'Hallo Aschaffenburg' - tile 'Hallo!' (perhaps this only happens in production apps)
2. Swipe up and tap 'Aschaffenburg in Zahlen'
3. Swipe all the way up 
4. Tap at 'Broschüre zum Familienleben' -> page opens; all fine now tap at back
5. Again tap at 'Broschüre zum Familienleben' -> page jumps by about 250 pixel down; 
6a. on emulator: one of the above links is opend
6b. on device: no link opens
7. Again tap at 'Broschüre zum Familienleben' -> page opens; all fine

### Additional Information

I can reproduce this behavior as every second tap on one of the two last links on this page (Bildung, Familienleben) when page is maximum swiped up. It can't be reproduced for the first link (Aschaffenburg in Zahlen) and I havn't seen this effect at any other point.

LeandraH:
@Docmale  Is it possible that it's the same behaviour as in #2160 ?

LeandraH:
Hmm, I just tested it a little and the jumping happens for me only when I tap the first time. In fact, if I go on a new page, it jumps down when I tap somewhere the first time, and if I go back to a previous page, it jumps up on the first tap.

I attached a longer video, is that the same behaviour that you are seeing, or is this a different problem?

[Screen_recording_20240830_122122.webm](https://github.com/user-attachments/assets/e68a3344-52df-4e63-b57d-d15cd54bbe45)

Docmale:
Yes, this is the behaviour I saw. Sorry for beeing late, didn't notice your Question.

steffenkleinle:
Done in #2955.

PR #2899, opened 2024-08-28
2886: Not working Link in Aschaffenburg

bahaaTuffaha:
### Short description

When pressing an anchor tag inside the WebView , the scroll position jumps down.

### Proposed changes

- Injecting JavaScript to webView to prevent <a> tag from scrolling.
- The code should listen to presses and go through nested elements to find <a> tag and forward the href to the window. 

### Side effects

None.

### Testing

1. Open 'Hallo Aschaffenburg' - tile 'Hallo!'
2. Swipe up and tap 'Aschaffenburg in Zahlen'
3. Swipe all the way up
4. Tap at 'Broschüre zum Familienleben' -> page opens; all fine now tap at back
5. Again tap at 'Broschüre zum Familienleben' -> page jumps by about 250 pixel down;

### Resolved issues

Fixes: #2886

---

### DIFF
--- a/native/src/components/RemoteContent.tsx
+++ b/native/src/components/RemoteContent.tsx
@@ -133,9 +133,37 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
   if (error) {
     return <Failure code={ErrorCode.UnknownError} />
   }
+  const injectedJavaScript = `
+  (function() {
+    document.addEventListener('touchend', function(e) {
+      var target = e.target || e.srcElement;
+      
+      function findClosestAnchor(element) {
+        while (element) {
+          if (element.tagName.toLowerCase() === 'a') {
+            return element;
+          }
+          element = element.parentElement;
+        }
+        return null;
+      }
+      if (['a', 'li', 'span', 'strong'].includes(target.tagName.toLowerCase())) {
+        var anchor = findClosestAnchor(target);
+        if (anchor) {
+          e.preventDefault();
+          window.location.href = anchor.getAttribute('href');
+        }
+      } else {
+        e.stopPropagation();
+      }
+    }, false);
+  })();
+  true;
+`
 
   return (
     <WebView
+      injectedJavaScript={injectedJavaScript}
       source={{
         baseUrl: resourceCacheUrl,
         html: renderHtml(
--- a/release-notes/unreleased/ReleaseNoteTemplate.yml
+++ b/release-notes/unreleased/ReleaseNoteTemplate.yml
@@ -1,0 +1,6 @@
+issue_key: 2886
+show_in_stores: false
+platforms: # relevant platforms, possible values: web, android and ios
+  - android
+  - ios
+en: Improved touch response for better navigation


LeandraH (commented):
Thank you for yet another PR 🚀 I have three points about this one:

1. To be honest, I don't quite understand the code, why do you want to disable scrolling?

2. I tested it on an emulator (Pixel 7 API 34), and the problem still occurred.
[Screen_recording_20240830_122828.webm](https://github.com/user-attachments/assets/5bdc4467-19ee-4d17-a22c-fc1df96df9e3)

It seems to not be happening when you open the "Zahlen aus Aschaffenburg" link a second time, so maybe you forgot to go back to the previous screen to reopen the page in your development? That at least sounds like something I would forget...

3. As written in the issue ticket about this, I think it's a larger issue that doesn't really have anything to do with the remote content. Maybe we should discuss this on Monday at the Weekly to see if other people have some ideas about what's going on here, but it really sounds a lot like #2160 to me.

bahaaTuffaha:
> Thank you for yet another PR 🚀 I have three points about this one:
> 
> 1. To be honest, I don't quite understand the code, why do you want to disable scrolling?
> 2. I tested it on an emulator (Pixel 7 API 34), and the problem still occurred.
>    [Screen_recording_20240830_122828.webm](https://github.com/user-attachments/assets/5bdc4467-19ee-4d17-a22c-fc1df96df9e3)
> 
> It seems to not be happening when you open the "Zahlen aus Aschaffenburg" link a second time, so maybe you forgot to go back to the previous screen to reopen the page in your development? That at least sounds like something I would forget...
> 
> 3. As written in the issue ticket about this, I think it's a larger issue that doesn't really have anything to do with the remote content. Maybe we should discuss this on Monday at the Weekly to see if other people have some ideas about what's going on here, but it really sounds a lot like [IGAPP-1225: Found articles jump upwards and hide title #2160](https://github.com/digitalfabrik/integreat-app/issues/2160) to me.


https://github.com/user-attachments/assets/1bfeae0e-6f29-41a3-87fd-4712f2b15a0d

Tested on Pixel 6a (API 33)
I will test it on 34..
I'm trying to stop the scrolling behavior when clicking on an anchor 😅

LeandraH:
> Tested on Pixel 6a (API 33)
> I will test it on 34..
> I'm trying to stop the scrolling behavior when clicking on an anchor 😅

Aaaah, I see, the difference is that I'm not clicking on an anchor, I'm just clicking somewhere on the screen. So I guess it's half-way there? But also, I would recommend not working on this ticket more until we've discussed it on Monday. If this is really a bug in react-native-webview, we'll have to decide if we maybe want to patch the package instead.

bahaaTuffaha:
> > Tested on Pixel 6a (API 33)
> > I will test it on 34..
> > I'm trying to stop the scrolling behavior when clicking on an anchor 😅
> 
> Aaaah, I see, the difference is that I'm not clicking on an anchor, I'm just clicking somewhere on the screen. So I guess it's half-way there? But also, I would recommend not working on this ticket more until we've discussed it on Monday. If this is really a bug in react-native-webview, we'll have to decide if we maybe want to patch the package instead.

About patching this package someone did this already but for some reason it got closed:
https://github.com/react-native-webview/react-native-webview/pull/3201

LeandraH:
> About patching this package someone did this already but for some reason it got closed: [react-native-webview/react-native-webview#3201](https://github.com/react-native-webview/react-native-webview/pull/3201)

Nice find! Looks like that could be a good starting point; we'd have to make sure that we indeed don't have inputs or other possible elements that this bug fix might misbehave with.

steffenkleinle:
@bahaaTuffaha I think this does not fix the issue, right? Should we close the PR then?

Issue #2553, opened 2023-11-09
Add Accessibility Role to components


ztefanie:
**Is your feature request related to a problem? Please describe.**
For better accessibility the `role` should be set.

**Describe the solution you'd like**
Find all relevant ui elements, where the `role` should be set, and set it to the correct value. See https://reactnative.dev/docs/accessibility#role

Replace all usages of `accessibilityRole` with `role`.

**Describe alternatives you've considered**
Leave it as it is.

sarahsporck:
Related Issue:
- #2176

steffenkleinle:
I'd postpone this until after the react-native upgrade as it seems like there is a change from `accessibilityRole` to `role`,

PR #2900, opened 2024-08-28
2553: Add Accessibility Role to components

bahaaTuffaha:
### Short description

For better accessibility: the role should be set.

### Proposed changes

- Replaced `accessibilityRole` with `role`
- `ThemedSearchBar.tsx` has search bar with `role ="searchbox"` instead of  `role ="search"`
- Forcing role for `Pressable from './base/Pressable'`

### Side effects

None.

### Resolved issues

Fixes: #2553 

---


LeandraH:
Is this ready for review or still a draft?

bahaaTuffaha:
> Is this ready for review or still a draft?

The replacing part is done but I need to look for any other component that needs `Role` tag.
When it's ready I will click **Ready for review**

native/src/components/NavigationTile.tsx
```diff
@@ -50,7 +50,7 @@ type NavigationTileProps = {
 }
 
 const NavigationTile = ({ tile, width }: NavigationTileProps): ReactElement => (
-  <StyledPressable onPress={tile.onTilePress} width={width}>
+  <StyledPressable role='button' onPress={tile.onTilePress} width={width}>
```
LeandraH, line 53: I think this is also a link, when tapping on a NavigationTile, you are sent to the route that you just tapped on.

native/src/components/Tile.tsx
```diff
@@ -40,7 +40,7 @@ const Tile = ({ onTilePress, tile, resourceCache }: TileProps): ReactElement =>
     tile.isExternalUrl ? openExternalUrl(tile.path, showSnackbar).catch(reportError) : onTilePress(tile)
 
   return (
-    <TileContainer onPress={openTile}>
+    <TileContainer onPress={openTile} role='button'>
```
LeandraH, line 43: This also looks like a link to me

LeandraH (approved):
Nice, thank you! Not tested

steffenkleinle (commented):
Thanks for this PR!

I just found https://www.npmjs.com/package/eslint-plugin-react-native-a11y and I think we should employ it to avoid forgetting this in the future. I just installed it and it seems like the role is still missing in 5 cases (at least I am getting a linting error). Could you perhaps check out https://github.com/digitalfabrik/integreat-app/pull/2913 and merge this if you think this makes sense?

Sadly these rules seem to not work together with styled-components/custom components. Therefore it would be nice to add a required property `role` to `base/Pressable` to also fix future problems. Thank you!

bahaaTuffaha:
> Thanks for this PR!
> 
> I just found https://www.npmjs.com/package/eslint-plugin-react-native-a11y and I think we should employ it to avoid forgetting this in the future. I just installed it and it seems like the role is still missing in 5 cases (at least I am getting a linting error). Could you perhaps check out #2913 and merge this if you think this makes sense?
> 
> Sadly these rules seem to not work together with styled-components/custom components. Therefore it would be nice to add a required property `role` to `base/Pressable` to also fix future problems. Thank you!

I decided not to proceed with installing "eslint-plugin-react-native-a11y" because styled-components didn't work properly with it. also we rarely used vanilla React Native components without styled-components.

native/src/components/base/SettingsSwitch.tsx
```diff
@@ -11,14 +11,15 @@ const SettingsSwitch = ({ value, onPress }: SettingsSwitchProps): ReactElement =
   const theme = useTheme()
   return (
     <Switch
+      testID='SettingsSwitch'
```
steffenkleinle, line 14:
```suggestion
```

native/src/components/SettingItem.tsx
```diff
@@ -47,18 +47,18 @@ type SettingItemProps = {
...
 }
 
 const SettingItem = (props: SettingItemProps): ReactElement => {
-  const { title, description, onPress, value, hasBadge, hasSwitch, bigTitle, accessibilityRole } = props
+  const { title, description, onPress, value, hasBadge, hasSwitch, bigTitle, role } = props
   const { t } = useTranslation('settings')
 
   return (
-    <Pressable onPress={onPress} accessibilityRole={accessibilityRole} accessible={false}>
+    <Pressable onPress={onPress} role={role || 'none'} accessible={false}>
```
steffenkleinle, line 61:
```suggestion
    <Pressable onPress={onPress} role={role ?? 'none'} accessible={false}>
```

steffenkleinle (approved):
Nice work, thanks!

Issue #2901, opened 2024-08-28
Events recurring every last Wednesday of the month are broken


LeandraH:
### Describe the Bug

First described in https://github.com/digitalfabrik/integreat-cms/issues/3022
When setting an event recurring every last Wednesday of the month, it instead shows up as being on the same calendar day as in the first month (e.g. showing up 28.8.24 - correct - 28.9.24 - wrong - 28.10.24 - wrong)

### Steps to Reproduce

1. Go to https://cms-test.integreat-app.de/testumgebung/events/de/4601/edit/
2. See that the recurrence rule is set to be every last Wednesday of the month.
3. Go to http://localhost:9000/testumgebung/de/events/test-monthly
4. See that the dates are 28.8., 28.9., 28.10.

### Expected Behavior

The dates should be 28.8., 25.9., 30.10.

### Actual Behavior

The dates are 28.8., 28.9., 28.10.

PR #2902, opened 2024-08-28
2901: Fix recurring dates yet again

LeandraH:
### Short description

When rewriting the recurrent dates a few weeks ago, I missed a test case: repeating an event every nth week of the month.

In that case, the recurrence rule gets the parameter `BYDAY=+3TH`. When parsing that, our package `rrule` turns that not into the parameter `byweekday` but into the parameter `bynweekday` (the n stands for nth weekday). When turning the parameter `bynweekday` back into a string, we get the parameter `BYNWEEKDAY=3,+3`, which doesn't seem to be a valid parameter.

So I decided to stick to the string of the recurrence rule as the source of truth and not the recurrence rule object from our `rrule` package.

You can also see some fun workarounds in this issue: https://github.com/jkbrzt/rrule/issues/326

### Proposed changes

- Get the recurrence rule that has been offset by our timezone via the string of the rule, not the object from the package
- MORE TESTS (if you see more missing test cases, please tell me)

### Side effects

- Hopefully none

### Testing

Run the tests in your environment
Check particularly that events repeating every nth day (e.g. every 4th Friday) are showing correctly http://localhost:9000/testumgebung/de/events/test-monthly
Any other repeating event combinations that you can think of

### Resolved issues

Fixes: #2901 

---


shared/api/models/DateModel.ts
```diff
@@ -137,14 +137,12 @@ class DateModel {
...
-      interval: recurrenceRule.options.interval,
-      until: recurrenceRule.options.until,
-      dtstart: offsetStartDate,
-    })
+    const offsetStartDate = DateTime.fromJSDate(startDate)
+      .minus({ minutes: startDate.getTimezoneOffset() })
+      .toUTC()
+      .toFormat("yyyyMMdd'T'HHmmss")
+    const regexForFindingDate = /\d{8}T\d{6}/
+    return rrulestr(recurrenceRule.toString().replace(regexForFindingDate, offsetStartDate))
```
steffenkleinle, line 145: Perhaps you could explain why you are doing what you are doing here together with a link to this PR/the issue in the library.

steffenkleinle (approved):
Works like a charm, thanks for fixing all this stuff! :green_heart:

LeandraH:
@ztefanie or @f1sh1918 it would be cool if one of you could still review this PR this week :)

f1sh1918 (approved):

Issue #2786, opened 2024-05-08
Upgrade react-native to v0.74


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
A new react-native version got released. See the upgrade helper for necessary things to do:
https://react-native-community.github.io/upgrade-helper/?from=0.73.8&to=0.74.1&package=tuerantuer.app.integreat&name=Integreat

**Testing information**
Please test everything (the whole app) very thoroughly!

**Additional context**
~~Please check whether it is possible with this to upgrade to yarn v3: #2026.~~
Not possible, the high memory usage issue which would require use to upgrade our CI images still exists :/

LeandraH:
Ideally, upgrade to 0.75, or if not possible, open an issue to do it.

PR #2907, opened 2024-09-02
2786: Upgrade react native to v0.74

steffenkleinle:
### Short description

Upgrade react native to v0.74.

I upgraded dependencies in #2912, so if you review both PRs at the same time, you only need to test #2912.

Switching to yarn modern as proposed by react-native will not be done as the memory/oom issue in the CI is still open.

Upgrading to react-native v0.75 is currently blocked by react-native-static-server(https://github.com/birdofpreyru/react-native-static-server) as the version we are currently using is not compatible and new versions require a higher minSDK version.

### Proposed changes

- Upgrade react native 0.74
- Upgrade a few libraries to fix build
- Patch malibre-react-native (see #https://github.com/maplibre/maplibre-react-native/issues/336), should be fixed with the next major release

### Side effects

Hopefully none.

### Testing

Nothing. Testing should be done with #2912 as both PRs will be merged together.

### Resolved issues

Partly fixes: #2786.

---


native/.gitignore
```diff
@@ -35,7 +35,7 @@ yarn-error.log*
 .hprof
 
 # CocoaPods
-/ios/Pods/
+/**/Pods/
```
LeandraH, line 38: ❓ In case the pods are installed elsewhere?

LeandraH (approved):

ztefanie (approved):
Looks good to me, i expect these changes are all the changes from react native updater.

native/.gitignore
steffenkleinle: Hmm I guess, this change was suggested by the update helper. Definitely should not hurt though.

PR #2910, opened 2024-09-02
2173: Send language to iOS VoiceOver

LeandraH:
### Short description

As part of the test of our accessibility from Pfennigparade a while ago, we got the feedback that the app doesn't send the language chosen in the app to OS so that you might get e.g. German texts read with an English pronunciation.

Android does a good job of that natively already, and the prop [accessibilityLanguage](https://reactnative.dev/docs/accessibility#accessibilitylanguage-ios) can be used to achieve the same thing in iOS.

### Proposed changes

- Added the prop accessibilityLanguage in many places.
- Wrote a little function telling us what language the text in the header should be.

### Side effects

- I turned the Pressable in the NewsItemList into two Pressables that lead to the same place because the first half is in the language that the app is in and the second part ("Continue") is in the system language. I didn't see a visual or behavioural difference but maybe I missed something.

### Testing

1. Be on an iPhone (VoiceOver doesn't exist on emulators, yay.)
2. Turn on VoiceOver
3. Set the app language to one that is different from your phone's language
4. Walk around through the app

Alternatively, you can edit the `HeaderBox` to show the title and the language, and that should be good enough for a test.

### Resolved issues

Fixes: #2173


native/src/components/Header.tsx
```diff
@@ -218,13 +220,30 @@ const Header = ({
     }
 
     const previousRouteTitle = (previousParams as { title?: string } | undefined)?.title
-    return previousRouteTitle ?? t(previousRoute.name === 'categories' ? 'localInformation' : previousRoute.name)
+    return previousRouteTitle ?? t(previousRoute.name === CATEGORIES_ROUTE ? 'localInformation' : previousRoute.name)
+  }
+
+  const isHeaderTextInPhoneLanguage = (): boolean => {
```
steffenkleinle, line 226: Hmm... not sure if I completely understand this function. I also think its not entirely correct, e.g. if following an internal link, right?
How about: `getHeaderText()` returns both a text and a boolean indicating the language it is in? e.g. `{ text: previousRouteTitle, systemLanguage: false }` and `{ text: t('locations'), systemLanguage: true }`. Should be less complexity and less error prone without duplicated logic/if else.

native/src/components/NewsListItem.tsx
```diff
@@ -86,7 +86,7 @@ const NewsListItem = ({ index, newsItem, navigateToNews, isTunews }: NewsListIte
     <>
       <Divider firstItem={index === 0} />
       <ListItemWrapper>
-        <StyledPressable onPress={navigateToNews}>
+        <StyledPressable onPress={navigateToNews} accessibilityLanguage='de'>
```
steffenkleinle, line 89: :question: Why is this always `de`?

steffenkleinle (commented):
Thanks for the PR!

Can you explain a little bit more when we have to set the language? Whenever the language is different than the system language, right? Wouldn't that include stuff like the webview as well?

native/src/components/NewsListItem.tsx
LeandraH: Ooops, thanks, that's test code 🤦

LeandraH:
> Can you explain a little bit more when we have to set the language? Whenever the language is different than the system language, right? Wouldn't that include stuff like the webview as well? How did you check where you need to set this?

Exactly, we have to set the language for components that should be read out in the app language as opposed to the system language. I checked it manually by going through the app on my phone with VoiceOver enabled. Webview seems to be doing this correctly already.

native/src/components/Header.tsx
LeandraH: Hmm, that's an interesting idea. I also realised that I was going on the possibly incorrect idea that if the route back is the name of the region (e.g. Testumgebung), it should be in the language of the app, not the one of the system. But now I'm not so sure, it might make more sense to use the system language. What do you think?

On a side note, I checked, we don't translate region names (e.g. Munich is München in all languages)

steffenkleinle: If it is not translated anyway, I don't see a reason why I would prefer one over the other. We can also hardcode it to `German` if its the city name.

native/src/components/NewsListItem.tsx
```diff
@@ -81,12 +82,13 @@ export const ReadMore = styled(Text)<{ isTunews: boolean }>`
 const NewsListItem = ({ index, newsItem, navigateToNews, isTunews }: NewsListItemProps): ReactElement => {
   const { t, i18n } = useTranslation('news')
   const timestamp = newsItem instanceof LocalNewsModel ? newsItem.timestamp : null
+  const { languageCode } = useContext(AppContext)
```
steffenkleinle, line 85:
```suggestion
  const { languageCode } = useAppContext()
```

LeandraH: Ooooh, nice, thanks!

native/src/components/Header.tsx
```diff
@@ -213,18 +215,35 @@ const Header = ({
...
+    } else if (previousRouteTitle || previousRoute.name === CATEGORIES_ROUTE) {
+      language = languageCode
+    } else {
+      language = '' // systen language
+    }
+
+    return {
+      text: previousRouteTitle ?? t(previousRoute.name === CATEGORIES_ROUTE ? 'localInformation' : previousRoute.name),
+      language,
+    }
```
steffenkleinle, line 235: :wrench:
I think something like the following would reduce complexity:
```suggestion
    const previousRoutePath = (previousParams as { path?: string } | undefined)?.path

    if (previousRouteTitle) {
      return { text: previousRouteTitle, language: languageCode }
    }
    if (previousRoute.name === CATEGORIES_ROUTE) {
      return { text: t('localInformation'), language: previousRoutePath === undefined ? DEFAULT_LANGUAGE : languageCode }
    }
    return { text: previousRoute.name, language: undefined }
```

However, please check the if-clauses again, but basically I just took the old `text` return and compared with your if-clauses.

native/src/components/Header.tsx
```diff
@@ -213,18 +215,35 @@ const Header = ({
     if (currentRouteIsPoi && notFromDeepLink) {
       const poisRouteParams = route.params as RoutesParamsType[PoisRouteType]
       if (poisRouteParams.slug || poisRouteParams.multipoi !== undefined) {
-        return t('locations')
+        return { text: t('locations'), language: '' } // system language
```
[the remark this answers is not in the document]
LeandraH: We can use undefined but I'm not sure that makes it clear to someone reading the code that the language used will be the system language.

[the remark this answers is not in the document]
steffenkleinle: Oh right, yes, feel free to keep the comment :)

steffenkleinle (approved):
Nice job! Not tested

native/src/components/Header.tsx
LeandraH: Nice, thanks!

native/src/components/Header.tsx
```diff
@@ -213,18 +214,35 @@ const Header = ({
...
+    }
+
+    if (previousRoute.name === CATEGORIES_ROUTE) {
+      return {
+        text: t('localInformation'),
+        language: previousRouteTitle === undefined ? config.sourceLanguage : languageCode,
+      }
+    }
+
+    return { text: previousRoute.name, language: undefined } // system language
```
ztefanie, line 234: :wrench: Feel free to fix this CodeScene issue, if you are motivated.

ztefanie (approved):
Thanks for this PR, code looks good. Not tested, because i do not have a iphone.

:thought_balloon: Should handle most cases properly, but we sometimes have edge cases, that still are handled wrongly, e.g. news that are send in german for all languages or things link this.

native/src/components/Header.tsx
LeandraH: I hadn't even noticed it, thanks! Looking into it also meant finally looking into why my emulator hasn't been updating in forever so that I could install the CodeScene extension, so thank you for that too!

PR #2912, opened 2024-09-04
2786: Upgrade dependencies

steffenkleinle:
### Short description

Upgrade dependencies.

### Proposed changes

- Upgrade most dependencies
- Increase memory limit for linting in CI

### Blocked

- @react-native-community/geolocation: Somehow crashes the app if using different location providers than google, see https://github.com/Agontuk/react-native-geolocation-service/issues/433#issuecomment-2131176759
- entities: htmlparser2 expects v4.5.0 and does not work otherwise
- @dr.pogodin/react-native-static-server (starting [v0.8](https://github.com/birdofpreyru/react-native-static-server/releases/tag/v0.8.0)): RN 0.72 and minimum SDK 28 
- @math.gl/web-mercator: Dependency not found?
- react-navigation-header-buttons: Lots of changes necessary and using of unstable flags.
- webdriverio: E2E tests fail without any apparent reason.
- webpack-dev-server: `Error: Cannot set headers after they are sent to the client` stops the server in the CI
- decamelize, flat: Only ESM modules available
- eslint: Lots of breaking changes in v9.0.0. Might be done in a separate issue once bugs are fixed.
- react-native and react: Blocked by react-native-static-server.

### Side effects

Hopefully none.

### Testing

:warning: Test the whole app, both web and native!

### Resolved issues

Fixes: #2786.

---


LeandraH (commented):
The code looks fine but I'm getting an error when trying to build it for iOS (Redefinition of module 'ReactCommon'), I'll try to fix that next week :)

web/package.json
```diff
@@ -18,7 +18,7 @@
     "build:aschaffenburg": "yarn run webpack-ts --env config_name=aschaffenburg",
     "build:integreat-test-cms": "yarn run webpack-ts --env config_name=integreat-test-cms",
     "build:obdach": "yarn run webpack-ts --env config_name=obdach",
-    "check:build": "yarn es-check es2020 ./dist/**/*.js --verbose",
+    "check:build": "yarn es-check es2021 ./dist/**/*.js --verbose",
```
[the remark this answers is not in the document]
LeandraH: Sounds good to me :)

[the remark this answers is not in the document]
ztefanie: yes lets update it

e2e-tests/web/test/pageobjects/page.ts
```diff
@@ -9,7 +9,11 @@ export default class Page {
    * @param path path of the sub page (e.g. /path/to/page.html)
    */
 
-  open(path: string): Promise<string> {
-    return browser.url(`${path}`)
+  async open(path: string): Promise<string> {
```
ztefanie, line 12: :upside_down_face: nitpick: i personally dislike mixing async/await and Promises like this, but no need to change it.

ztefanie (approved):
Thanks for updating :rocket: 

quickly tested an Android and firefox, everything works as expected.

e2e-tests/web/test/pageobjects/page.ts
steffenkleinle: How would you change this? I think I don't really mix async/await and promises, `Promise` is just the correct return type of a async function.

LeandraH (approved):
Awesome, thanks for (still) slogging through that! Tested on Firefox and on real iOS, works nicely!

native/ios/Podfile.lock
```diff
@@ -34,33 +57,72 @@ PODS:
     - GCDWebServer/Core (= 3.5.4)
   - GCDWebServer/Core (3.5.4)
   - glog (0.3.5)
+  - GoogleAppMeasurement (10.24.0):
```
f1sh1918, line 60: if i run `pod install` i don't get these resolved pods here. My podfile.lock differs a lot. Even i ran `yarn install` before. But i'm using newer cocoapods version and

f1sh1918 (commented):
Unfortunately i can't build the app on ios with Xcode 15.4

Steps so far:
- `yarn install`
- `pod install` (leads to differences in podfile.lock (ios folder), check my comment
- "Clean product "in xcode
- run build

Error
image

I was also checking this issue to fix it, but the resolved `podfile.lock` is still differing from your verion and the error remains
https://github.com/reactwg/react-native-releases/issues/284
But couldn't get it work.
Even i run a build without `pod install` using your podfile.lock results in same error

LeandraH:
Urg, can you try deleting your node_modules and Pods and then reinstalling?

steffenkleinle:
One thing perhaps: After you installed the pods/did the installation for iOS @LeandraH, I reverted some upgrades and dependency resolutions. So differing pod file is probably okay, just commit them.

LeandraH:
It doesn't build for me anymore either, though that may partially be because I stupidly let my computer update Xcode this morning 🤦  I'm getting lots of "Could not build module XYZ" errors in React-RCTAppDelegate. Will try to figure that out tomorrow.

LeandraH:
I got it working again, @f1sh1918 can you test again? Probably with deleting node_modules and deintegrating Pods

LeandraH:
@f1sh1918 Can you give this another check? Just to see if it builds for you, I don't think you need to test the app itself if you don't have the time :)

PR #2913, opened 2024-09-04
2553:  [Proposal] Add linting rules for a11y in native

steffenkleinle:
### Short description

Add linting rules for a11y in native.

### Proposed changes

Add linting rules for a11y in native.

### Side effects

Failing linting.

### Resolved issues

Fixes: #

---


steffenkleinle:
Won't do as styled-components/custom names are not supported.

Issue #2897, opened 2024-08-27
Background app refresh not happening


LeandraH:
### Describe the Bug

Report from Sascha on Mattermost:

He opened the mobile app for the first time in a while. It's usually on Stadt Augsburg (and also in this case). The nav bar at the top didn't show the map option even though that had been enabled a while ago. After a pull-to-refresh, the map route showed up. On an iPhone 12 Pro Max with iOS 17.5.1, connected to the live CMS

The app should have checked if the last update was more than a day ago, seen that it was indeed more than a day ago, and then updated the data.

### Steps to Reproduce

1. Change one of the feature flags for 'Testumgebung' deciding which routes are possible
2. Wait more than a day
3. Open the native app
4. See that the routes haven't updated

Note: I haven't actually tested this. Presumably, one can change the period for checking the last update.

### Expected Behavior

The available routes in the nav bar have updated.

### Actual Behavior

The available routes in the nav bar haven't updated.

### Additional Information

Thread in Mattermost:

https://chat.tuerantuer.org/digitalfabrik/pl/tcwqhkhmjpn9djpjjbmn97kqfc

LeandraH:
First things I found: The function `useLoadCities` that checks what feature flags are enabled doesn't get called repeatedly like `useLoadCityContent` and `useLoadWithCache`. I haven't gotten around to putting it in yet; I also haven't checked yet if the other two hooks do work in the background.

PR #2921, opened 2024-09-11
2897: Refresh cities every day

steffenkleinle:
### Short description

At the moment, cities are not refreshed at all anymore if a city has been selected.
In Landing, we call the refresh manually as a workaround.
However, if a city is selected, Landing is not rendered anymore and therefore the workaround is not triggered.

This PR works around that by using `useLoadWithCache` instead of `useLoadCities` in the `useCityContent` hook.

### Proposed changes

Use `useLoadWithCache` over `useLoadCities` in the `useCityContent` hook.

### Side effects

None.

### Testing

Instead of manually changing values in the CMS and waiting for the next day for the refresh to happen, you can also do the following:
- Change any property in `mapCityJson`, e.g. the following line to
https://github.com/digitalfabrik/integreat-app/blob/4529e219cb0c8ea866dff6348e1ce8b318b22b8c/shared/api/mapping/mapCityJson.ts#L29

``` ts
eventsEnabled: true
```
- Change the refresh condition in `useLoadWithCache`, i.e. the following line to
https://github.com/digitalfabrik/integreat-app/blob/4529e219cb0c8ea866dff6348e1ce8b318b22b8c/native/src/hooks/useLoadWithCache.ts#L41

``` ts
const shouldUpdate = forceUpdate || !lastUpdate || lastUpdate < DateTime.utc().startOf('day').plus({ day: 1 })
```

### Resolved issues

Fixes: #2897.

---


LeandraH (commented):
Hmmm, I deactivated events in the CMS and set `shouldUpdate = true` in `useLoadWithCache`; I then tapped around in the app a little, in the hopes that that would trigger a refresh but only an actual pull-to-refresh did. Any ideas?


https://github.com/user-attachments/assets/1ef5a334-09c1-4ba4-9f2c-f6ef18eb8062

LeandraH (approved):
But it does work when the app is restarted (i.e. removed from the RAM and opened again), which seems to be the case that didn't work here :)

f1sh1918:
@steffenkleinle 
i'm really not sure how to test what and what is expected. I did the changes you proposed to do.
Then i cleaned app from ram and restarted. I don't have any change.
If i set `eventEnabled: false` i expected that events will disappear.
But that only worked after i switched city.

ztefanie (approved):
Tested on android, works as expected :+1:

Issue #2805, opened 2024-05-25
Banner to install app from store is not displayed when using web-app


dkehne:
**Is your feature request related to a problem? Please describe.**
We want users to install the app on their phones over using the web app.
Therefore, we want to notify the users about the app in the google play store.

**Describe the solution you'd like**
When the web app is opened on Android phones and the android app is not installed, a banner that the app is available in the google play store.
Since this is natively implemented on iOS, this is only an android issue. Reimplement the banner on iOS, see screenshot below.

This should work for for the other whitelables as well (e.g. malte, aschaffenburg).

If the banner is either closed or the link to the store is pressed, the banner should not be shown anymore until a specific amount of time has passed (to be discussed).
To this end, a value might be saved in the local storage (preferable over cookies).

**Describe alternatives you've considered**
* Do not show again after closing the banner once
* Do not show a banner

**Additional context**
<img src="https://github.com/user-attachments/assets/d425daa7-e7a1-4260-9496-ec4afb8ff54f" width="400px" />
Side note, in iOS, the banner is only shown on real devices, not in emulators.

LeandraH:
I thought I might find something in the translations about this banner but I didn't. I think a design would be a good idea.

steffenkleinle:
> I thought I might find something in the translations about this banner but I didn't. I think a design would be a good idea.

@LeandraH we never provided the banner ourselves. This is something that should be displayed automatically by browsers using corresponding meta tags in the web app, see e.g. for ios here: https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners

LeandraH:
It's still there in iOS but you have to reset some settings if you've ever dismissed the banner on your phone. 

![IMG_ADB1EF45C13C-1](https://github.com/digitalfabrik/integreat-app/assets/18513943/f3196500-7489-40a1-8c8c-c2471b1f9db8)

It looks like Android never really implemented this banner. Or does someone with an Android remember things differently?

vthejay:
Android:
With a brand new device, Samsung A23 5G: Wenn I open integreat.app on the browser, there is no banner appearing, asking me to install the app.
This happens on the landing page and also when I choose a region.
Of course this is a brand new device and the app is definitely not installed.

steffenkleinle:
I think this was never implemented for Android, at least not with the react native app.

steffenkleinle:
[...]

PR #2936, opened 2024-09-24
2805: Banner to install app from store is not displayed

bahaaTuffaha:
### Short description

When the web app is opened on Android phones a banner similar to iOS should show up to redirect user to google Play or the app if installed.
### Proposed changes

- Created a component called MobileBanner.
- It has a state `isInstalled` to change design between if app installed or not similar to ios. (for now its false be becouse we cannot determine if app installed or not before running the deep-link).
- Closing the banner should add a date in `localStorage` (plus 3 months) to show the banner again after these 3 month. 

### Side effects

This component added in the layout.

### Testing
You can test it (Emulated) in 2 ways:
- Open webapp with plug-in like User-Agent-Switcher and select android device.
or
- Run android emulator + run webapp locally in google chrome in the emulator.

### Resolved issues

Fixes: #2805 

---


assets/icons/star.svg
```diff
@@ -0,0 +1,20 @@
...
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"><defs
+     id="defs1" /><g
+     id="layer1"
+     style="display:inline"><g
+       id="g1"
+       transform="matrix(1.0402761,0,0,1.0402761,-32.20264,-61.436195)"><path
+         d="m 43.084173,78.555943 4.660243,2.451552 c 0.962251,0.507703 2.088893,-0.313088 1.902719,-1.382942 l -0.890923,-5.189728 c -0.07374,-0.426722 0.06769,-0.860685 0.377179,-1.160498 l 3.769288,-3.676144 c 0.779699,-0.757947 0.349356,-2.082861 -0.727739,-2.237593 l -5.2114,-0.757944 c -0.426721,-0.06166 -0.797832,-0.331222 -0.987628,-0.718052 L 43.645265,61.16153 c -0.482333,-0.974322 -1.872514,-0.974322 -2.354846,0 l -2.330648,4.723064 c -0.192206,0.38683 -0.559699,0.656406 -0.987628,0.718052 l -5.211399,0.757944 c -1.077095,0.158367 -1.507438,1.479617 -0.727739,2.237593 l 3.769289,3.676144 c 0.309467,0.299795 0.452099,0.736179 0.377176,1.160498 l -0.89092,5.189728 c -0.182534,1.071029 0.940498,1.890645 1.902719,1.382942 l 4.660242,-2.451552 c 0.383211,-0.200682 0.83893,-0.200682 1.222141,0 z"
+         id="path1"
+         style="fill:"currentColor";stroke-width:0.309465" /></g></g></svg>
```
steffenkleinle, line 20:
```suggestion
 /></g></g></svg>
```

steffenkleinle (changes requested):
Great solution and nice implementation! :rocket: 

Maybe this is fine, but I think we should still discuss this:
The workaround with the timeout only works if
- a) the browser does not ask for confirmation whether to allow opening of the link. For firefox mobile, this is the case, such that it opens the page in the playstore even though the app is installed. See attached video. For e.g. brave this does not seem to be the case, if this is also not the case for the default browsers and chrome probably not a problem either.
- b) only one app that supports the `integreat://` scheme is installed. This should be the case for nearly all our users, so probably fine as well.


https://github.com/user-attachments/assets/bfdcc3a3-82a2-431c-b11a-e61c3d15c017


https://github.com/user-attachments/assets/fe0d7768-1483-44b6-8b13-6b43058299be

steffenkleinle (approved):
Two more things:
- Do you think we should also hide the banner if the user clicks on `View`/`Open`? I am not sure about this.
- If it is never ever possible to check whether the app is installed before opening the link, we should perhaps also remove the logic for that. Also not sure here though.

translations/translations.json
```diff
@@ -5552,7 +5552,7 @@
       "disclaimer": "هيئة التحرير",
       "sideBarOpenAriaLabel": "فتح الشريط الجانبي",
       "sideBarCloseAriaLabel": "غلق الشريط الجانبي",
-      "openInApp": "افتح في تطبيق Integreat",
+      "openInApp": "افتح في تطبيق {{appName}}",
```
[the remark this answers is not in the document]
steffenkleinle: Looks good! And yes, `Anzeigen` also sounds better :D

LeandraH (approved):
Very nice, tested on emulated Android

translations/translations.json
```diff
@@ -5744,7 +5756,11 @@
       "shareFailDefaultMessage": "Le partage du contenu a malheureusement échoué.",
       "disclaimer": "Mentions légales",
       "sideBarOpenAriaLabel": "Ouvrir la barre latérale",
-      "sideBarCloseAriaLabel": "Fermer la barre latérale"
+      "sideBarCloseAriaLabel": "Fermer la barre latérale",
+      "openInApp": "Ouvrir dans l'app {{appName}}",
+      "open": "Ouvrir",
+      "view": "Voir",
+      "getOnPlayStore": "Obtenir - Sur le Google Play Store"
```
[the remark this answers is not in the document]
bahaaTuffaha: Yes I took it from the screenshot...
Do you want me to remove all french translations?

web/src/components/MobileBanner.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <StyledAppName>{appName}</StyledAppName>
+            {isInstalled === false && (
+              <>
+                <StyledDescription $screenSize={window.innerWidth}>Tür an Tür - Digitalfabrik gGmbH</StyledDescription>
+                <StyledStars>
+                  <StyledStarIcon src={StarIcon} />
+                  <StyledStarIcon src={StarIcon} />
+                  <StyledStarIcon src={StarIcon} />
+                  <StyledStarIcon src={StarIcon} />
+                  <StyledStarIcon src={StarIcon} />
```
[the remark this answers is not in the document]
bahaaTuffaha: I heard there is a 3rd party APIs that can scrap reviews but I just found this https://developers.google.com/android-publisher/api-ref/rest/v3/reviews/list 
But unfortunately this needs a token...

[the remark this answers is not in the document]
LeandraH: We should have that token though? Maybe @ztefanie knows more about it?

translations/translations.json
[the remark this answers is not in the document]
LeandraH: Actually, that might be best, let the professional translators handle it :)

translations/translations.json
```diff
@@ -5506,7 +5506,7 @@
       "openInApp": "In der {{appName}} App öffnen",
       "open": "Öffnen",
       "view": "Anzeigen",
-      "getOnPlayStore": "Im Google Play Store anschauen",
+      "getOnPlayStore": "Holen Sie es - Im Google Play Store anschauen",
```
LeandraH, line 5509: Oh sorry, that's a misunderstanding. The whole translation value should be "Im Google Play Store anschauen", without that "Holen Sie es". "Holen Sie es" sounds weird. Like a McDonald's ad or something.

web/src/components/MobileBanner.tsx
```diff
@@ -51,14 +50,16 @@ const StyledDivText = styled.div`
 
 const StyledAppName = styled.span`
   font-weight: bold;
-  font-size: 14px;
+  font-size: 12px;
   color: ${props => props.theme.colors.themeContrast};
 `
 
-const StyledDescription = styled.span`
+const smallScreenSize = 400
```
steffenkleinle, line 57: Please use either our hook

 ``` tsx
```
or a media query:
``` css
@media ${dimensions.smallViewport} {
...
}
```

bahaaTuffaha: image
smallViewport is 768px.. I need smaller size for iphone SE for example

bahaaTuffaha: Also the banner shows up at the smallViewport so there is no need for media query at this point.

steffenkleinle: Sorry, I  missed that. Thanks for explaining.

LeandraH:
@bahaaTuffaha This looks like it can be merged? 🥳

PR #2950, opened 2024-10-05
2916: Add tts on android

bahaaTuffaha:
### Short description

Adding text-to-speech functionality.

### Proposed changes

- `TtsPlayer.tsx` : deals with html content and it passes it to `react-native-tts` as sentences. used debounce function from lodash to handle rapid volume change. I added an expandPlayer state just to keep player expanded when changes from sentence to sentence. This component located at app level where it wraps around the NavigationContainer to be able to be floating at the bottom of the screen while scrolling.

- `useTtsPlayer.ts`: Is a hook used to pass data (content, title, visibility, volume) from across the app to `TtsPlayer.tsx` using `useContext`. 

- Categories.tsx, News.tsx, Events.tsx each one is using the `useTtsPlayer.ts` hook to pass content and title.

- Created a component `Slider.tsx` instead of adding more libraires and used the ones we have: `react-native-reanimated` & `react-native-gesture-handler`  to create slider that can be used anywhere other than the volume.

- Added ReadAloud HeaderButtonTitle to be selected from the kebab menu to show the player. This button will show up only in Integreat/IntegreatTestCms and at third level pages.

- The TtsPlayer will stop when the app is in the background also if you back out from a 3rd level page.

- Notes: 
   - Not Tested on iOS !
   - Why I passed sentences to `tts.speak()`? due to `react-native-tts` doesn't have a functional pause also passing word by word will sound more robotic.
   - Persian is not supported at `react-native-tts`.

### Side effects

- Categories.tsx, News.tsx, Events.tsx, Header.tsx

### Testing

- At integreat : Go to any 3rd level content in categories or news.
- Open the kebab menu on the right and select Read aloud.
- The player will show up press play to make it start reading content.

### Resolved issues

Fixes: #2916 

---


LeandraH (commented):
Works pretty well so far! A couple of points about the functionality:

- If I change the volume while it's playing, it starts re-reading the current fragment. If possible, I think it should keep reading at the same spot.
- In using, I thought the volume bar was for updating the speed of reading. Maybe that's just me but maybe not, we should see if other people have that problem.
- For Persian (and any other languages where it doesn't work), I think we should hide the button to read the text aloud, and close the overlay if it's open when someone switches languages.
- In emulated iOS, no sound comes out and I get this warning: `Excessive number of pending callbacks: 501. Some pending callbacks that might have leaked by never being called from native code: {"1000":{},"1001":{},"1002":{},"1003":{},"1004":{},"1005":{},"1006":{},"1007":{},"1008":{},"1009":{},"1010":{},"1011":{},"1012":{},"1013":{},"1014":{},"1015":{},"1016":{},"1017":{},"1018":{},"1019":{},"1020":{},"1021":{},"1022":{},"1023":{},"1024":{},"1025":{},"1026":{},"1027":{},"1028":{},"1029":{},"1030":{},"1031":{},"1032":{},"1033":{},"1034":{},"1035":{},"1036":{},"1037":{},"1038":{},"1039":{},"1040":{},"1041":{},"1042":{},"1043":{},"1044":{},"1045":{},"1046":{},"1047":{},"1048":{},"1049":{},"...(truncated keys)...":451}` But I guess you did say that it's not yet tested for iOS :D

As discussed, I only half-checked the PR, let me know when it's ready for review again :)

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,348 @@
...
+  const sentences: string[] | [] = useMemo(
+    () => (content ? [title, ...extractSentencesFromHtml(content)] : []),
+    [title, content],
+  )
+  const isPersian = languageCode === 'fa'
+
+  const initializeTts = useCallback((): void => {
+    Tts.getInitStatus()
+      .then(async (status: string) => {
+        // Status does not have to be 'success'
```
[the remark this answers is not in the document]
bahaaTuffaha: I took reference from lunes:
https://github.com/digitalfabrik/lunes-app/blob/41123b9ff0984f62046987c1d2fdb0ccfd71ef8d/src/components/AudioPlayer.tsx#L45

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,348 @@
...
+        if (status === 'success') {
+          // await Tts.setDefaultLanguage('de-DE')
```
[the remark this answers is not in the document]
bahaaTuffaha: I guess I need to remove it but again this is from lunes😅
https://github.com/digitalfabrik/lunes-app/blob/41123b9ff0984f62046987c1d2fdb0ccfd71ef8d/src/components/AudioPlayer.tsx#L49

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,348 @@
...
+  setVisible: React.Dispatch<React.SetStateAction<boolean>>
+  title: string
+  setTitle: React.Dispatch<React.SetStateAction<string>>
+  volume: number
+  setVolume: React.Dispatch<React.SetStateAction<number>>
+}
+export const ttsContext = createContext<ttsContextType>({
+  content: null,
+  setContent: () => {
+    // setContent
```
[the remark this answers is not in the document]
bahaaTuffaha: eslint is complaining about empty functions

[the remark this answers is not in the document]
LeandraH: I would say disable lint for empty functions before all of these and enable it again after then. It currently looks to me like the code isn't finished and something still needs to be added there.

native/src/components/TtsPlayer.tsx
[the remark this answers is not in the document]
LeandraH: @steffenkleinle ! I object to your code from two years ago 😂

But seriously, I think we should explain why it doesn't have to be a success.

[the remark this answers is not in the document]
LeandraH: Wait, in Lunes the comment explains the disabling of the linter in the next line. But you're not disabling the linter here, so you don't need it. Objection withdrawn, Steffen 😅

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+  const stopTts = async () => {
+    await Tts.stop()
+    const TTS_STOP_DELAY = 100 // delay to make sure tts is stopped
+    await new Promise(resolve => {
+      setTimeout(resolve, TTS_STOP_DELAY)
+    })
+  }
+  const startReading = () => {
+    if (!isPersian && sentences.length > 0) {
```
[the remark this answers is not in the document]
bahaaTuffaha: I Added a condition in the header to check language and hide the button from the menu

assets/icons/no-sound.svg
LeandraH, line 1: This can be simplified to

```
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
   <path
      d="m 16.358167,22.405586 c -0.263674,0 -0.51229,-0.05647 -0.745817,-0.169406 -0.233543,-0.113091 -0.478379,-0.29376 -0.734524,-0.542469 L 10.063938,17.128393 C 9.988598,17.060663 9.894424,17.026719 9.781425,17.026719 H 6.5382598 c -0.7608761,0 -1.3447246,-0.207144 -1.7515378,-0.621494 C 4.3799135,15.99089 4.1765092,15.373145 4.1765092,14.55199 v -4.203688 c 0,-0.8136106 0.2034043,-1.4275911 0.6102128,-1.8419412 C 5.1935352,8.084497 5.7773837,7.8735574 6.5382598,7.8735574 H 9.781425 c 0.112999,0 0.207175,-0.033897 0.282513,-0.1017052 l 4.813888,-4.5200968 c 0.286277,-0.2787344 0.534878,-0.4783725 0.745817,-0.5989084 0.21094,-0.1205358 0.448247,-0.1808029 0.711922,-0.1808029 0.384202,0 0.696847,0.1318357 0.937919,0.3955087 0.241071,0.2561453 0.361599,0.5687747 0.361599,0.937919 V 21.139979 c 0,0.361646 -0.120528,0.662967 -0.361599,0.904115 -0.233543,0.240995 -0.538643,0.361492 -0.915317,0.361492 z"
      fill="currentColor" />
</svg>
```

It's not necessary but it's good practice to keep SVGs as small and clean as possible.

native/src/components/Header.tsx
```diff
@@ -184,14 +191,20 @@ const Header = ({
...
   const overflowItems = showOverflowItems
     ? [
         ...(shareUrl ? [renderOverflowItem(HeaderButtonTitle.Share, onShare)] : []),
         ...(!buildConfig().featureFlags.fixedCity
           ? [renderOverflowItem(HeaderButtonTitle.Location, () => navigation.navigate(LANDING_ROUTE))]
           : []),
         renderOverflowItem(HeaderButtonTitle.Settings, () => navigation.navigate(SETTINGS_ROUTE)),
+        ...[
+          content &&
+          (buildConfig().appName === 'IntegreatTestCms' || buildConfig().appName === 'Integreat') &&
```
f1sh1918, line 203: please use a buildConfig flag (`ttsEnabled` or sth similar) for tts instead of checking the app name here

native/src/components/Header.tsx
```diff
@@ -184,14 +191,20 @@ const Header = ({
...
+          languageCode !== 'fa'
```
f1sh1918, line 204: i would prefer to use sth like a not supported languages array if the purpose here is to exclude not supported languages

native/src/components/News.tsx
```diff
@@ -63,6 +64,16 @@ const News = ({
   const selectedNewsItem = news.find(_newsItem => _newsItem.id === newsId)
   const { t } = useTranslation('news')
 
+  const { setTitle, setContent } = useTtsPlayer()
+
+  useEffect(() => {
```
f1sh1918, line 69: couldn't you just pass these values to the hook instead of using a separate useEffect here. I think the lifecycle depending things should be handled in the hook

native/src/components/RemoteContent.tsx
```diff
@@ -133,7 +135,7 @@ const RemoteContent = (props: RemoteContentProps): ReactElement | null => {
   if (error) {
     return <Failure code={ErrorCode.UnknownError} />
   }
-
+  const ttsPlayerHeight = 50
```
f1sh1918, line 138: we have a `dimensions.ts` where it should be moved

f1sh1918 (commented):
i couldn't test it since its crashing for on emulator when i click play on Android 13

native/src/components/News.tsx
bahaaTuffaha: I didn't get what exactly you want to me to do here... just pass them to the hook without useeffect?
I think in this case it will keep setting these states each time the News re-render right?

f1sh1918 (commented):
Works for me on android. good work.
Since its a draft some improvements are missing i guess

image
Shouldn't here stand the title of what will be read?
Even add shadows/elevation for the container and the play button
image
The icons should switch i guess since the direction arrows are wrong

image
I think we should create a fixed area for that player to not overlap content

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
...
+  gap: 5px;
+  width: 176px;
+`
+
+const CloseView = styled.View`
+  flex-direction: column;
+  gap: 10px;
+`
+
+export type ttsContextType = {
```
f1sh1918, line 95: should be `TtsContext`

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
...
+  title: string
+  setTitle: React.Dispatch<React.SetStateAction<string>>
+  volume: number
+  setVolume: React.Dispatch<React.SetStateAction<number>>
+}
+
+/* eslint-disable @typescript-eslint/no-empty-function */
+export const ttsContext = createContext<ttsContextType>({
+  content: null,
+  setContent: () => {},
```
f1sh1918, line 111:
```suggestion
  setContent: () => undefined,
```
for all the empty functions and remove es-lint disable

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
...
+  initialVisibility?: boolean
+}
+
+const TtsPlayer = ({ initialVisibility = false, children }: TtsPlayerProps): ReactElement => {
+  const { languageCode } = useContext(AppContext)
+  const { t } = useTranslation('layout')
+  const [isPlaying, setIsPlaying] = useState(false)
+  const [expandPlayer, setExpandPlayer] = useState(false)
+  const defaultVolume = 50
+  const [volume, setVolume] = useState(defaultVolume)
```
f1sh1918, line 134: i think you can initialize this with the value from your context, no?
`{volume} = useTtsPlayer()`

native/src/components/Header.tsx
```diff
@@ -185,13 +192,22 @@ const Header = ({
     renderItem(HeaderButtonTitle.Language, 'language', showItems, goToLanguageChange),
   ]
 
+  const unsupportedLanguagesForTts = ['fa']
+  const ttsEnabled =
+    content &&
+    (buildConfig().appName === 'IntegreatTestCms' || buildConfig().appName === 'Integreat') &&
+    !unsupportedLanguagesForTts.includes(languageCode)
+      ? renderOverflowItem(t(`${HeaderButtonTitle.ReadAloud}`), onRead)
```
f1sh1918, line 200: just use here
```js
()=>setVisible(true)
```
and remove onRead, since its not reused somewhere and unneeded code

native/src/components/Header.tsx
```diff
@@ -82,6 +84,7 @@ const Header = ({
   // Save route/canGoBack to state to prevent it from changing during navigating which would lead to flickering of the title and back button
   const [previousRoute] = useState(navigation.getState().routes[navigation.getState().routes.length - 2])
   const [canGoBack] = useState(navigation.canGoBack())
+  const { setVisible, content } = useTtsPlayer()
```
f1sh1918, line 87:
```suggestion
  const { setVisible: setTtsPlayerVisible , content } = useTtsPlayer()
```
To have a better idea whats happens if you use it below

LeandraH:
It unfortunately also doesn't play any sound on real iOS, there is just this warning that pops up:

IMG_023107052847-1

LeandraH:
> > It unfortunately also doesn't play any sound on real iOS, there is just this warning that pops up:
> 
> Maybe I should define `iosVoiceId: '',` for ios https://github.com/ak1394/react-native-tts?tab=readme-ov-file#speaking

Hmm, putting in the example `'com.apple.ttsbundle.Moira-compact'` didn't change the behavior. 

There is also a very minor bug that I found: when opening the TTS player, the line with the last update jumps down, and back up when you close it. I attached a video of the behavior.

https://github.com/user-attachments/assets/930195ce-5929-4764-a4fe-04030d74edd1

bahaaTuffaha:
> > > It unfortunately also doesn't play any sound on real iOS, there is just this warning that pops up:
> > 
> > 
> > Maybe I should define `iosVoiceId: '',` for ios https://github.com/ak1394/react-native-tts?tab=readme-ov-file#speaking
> 
> Hmm, putting in the example `'com.apple.ttsbundle.Moira-compact'` didn't change the behavior.
> 
> There is also a very minor bug that I found: when opening the TTS player, the line with the last update jumps down, and back up when you close it. I attached a video of the behavior.

This behavior is intended to leave space for the player to prevent it from covering the content.
I think iosVoiceId should be taken from `voices()`
`Tts.voices().then(voices => console.log(voices));`
The voices should be filtered by the language and passed to speak.

LeandraH:
> This behavior is intended to leave space for the player to prevent it from covering the content.

That's a good idea but leads to jumping in this case. The padding should probably be added under the last update stamp instead of before.

LeandraH:
> I think iosVoiceId should be taken from `voices()`
> `Tts.voices().then(voices => console.log(voices));`
> The voices should be filtered by the language and passed to speak.

I unfortunately still don't get any sound out. But the iOS can be fixed in a separate issue, I think you can for now work on getting the PR ready for reviewing for Android :)

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  const pauseReading = () => {
+    Tts.stop()
+    setIsPlaying(false)
+  }
+
+  const handleBackward = async () => {
+    await stopTts()
+    setSentenceIndex(prev => Math.max(0, prev - 1)) // it return the bigger number so no negative values
```
steffenkleinle, line 241: :wrench: Please don't abbreviate words like this. Also, I think `index` is more clear.
```suggestion
    setSentenceIndex(index => Math.max(0, index - 1))
```

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
...
+  const debounceDelay = 500
+
+  const debouncedVolumeChange = debounce((newVolume: number) => {
+    setVolume(newVolume)
+    if (isPlaying) {
+      Tts.stop().then(() =>
+        setTimeout(() => {
+          startReading()
+          // eslint-disable-next-line no-magic-numbers
+        }, 200),
```
steffenkleinle, line 260: :wrench: Please extract to constant instead of disabling linting.

native/src/routes/Events.tsx
```diff
@@ -41,6 +42,8 @@ export type EventsProps = {
 
 const Events = ({ cityModel, language, navigateTo, events, slug, refresh }: EventsProps): ReactElement => {
   const { t } = useTranslation('events')
+  const event = slug ? events.find(it => it.slug === slug) : null
+  useTtsPlayer(event?.content ?? '', event?.title ?? 'Events')
```
steffenkleinle, line 46: :wrench: 
```suggestion
  useTtsPlayer(event?.content, event?.title)
```

native/src/hooks/useTtsPlayer.ts
```diff
@@ -0,0 +1,32 @@
+import { useContext, useEffect } from 'react'
+
+import { ttsContext, TtsContextType } from '../components/TtsPlayer'
+
+const useTtsPlayer = (content?: string, title?: string): TtsContextType => {
```
steffenkleinle, line 5: :thinking: How about extracting the sentence creation logic to here? This way, we would not need duplicated state (title, content, sentences) and using this would be more easy.
```ts
const useTtsPlayer = (model?: PageModel | LocalNewsModel): TtsContextType => {
  const sentences = useMemo(
    () => {
      if (model) {
        const content = parseHTML(model.content) 
        return [model.title, ...segment(languageCode, content)]
      }
      return null
    },
    [model, languageCode],
  )

  useEffect(() => {
    tts.setSentences(sentences)
    return () => {
      tts.setSentences([])
    }
  }, [sentences, tts])
```

native/src/components/News.tsx
```diff
@@ -62,6 +63,7 @@ const News = ({
 }: NewsProps): ReactElement => {
   const selectedNewsItem = news.find(_newsItem => _newsItem.id === newsId)
   const { t } = useTranslation('news')
+  useTtsPlayer(selectedNewsItem?.content ?? '', selectedNewsItem?.title ?? 'News')
```
steffenkleinle, line 66:
```suggestion
  useTtsPlayer(selectedNewsItem?.content, selectedNewsItem?.title)
```

native/src/components/Categories.tsx
```diff
@@ -34,6 +35,7 @@ const Categories = ({
 }: CategoriesProps): ReactElement => {
   const children = categories.getChildren(category)
   const cityCode = cityModel.code
+  useTtsPlayer(categories.isLeaf(category) ? category.content : '', categories.isLeaf(category) ? category.title : '')
```
steffenkleinle, line 38:
```suggestion
  useTtsPlayer(categories.isLeaf(category) ? category  : undefined)
```

native/src/components/RemoteContent.tsx
```diff
@@ -181,7 +183,7 @@ const RemoteContent = ({
       // https://github.com/react-native-webview/react-native-webview/issues/1869
       setSupportMultipleWindows={false}
       style={{
-        height: webViewHeight,
+        height: visible ? webViewHeight + dimensions.ttsPlayerHeight : webViewHeight,
```
steffenkleinle, line 186: :x: I sadly think that this is not the right place for this as the `lastUpdate` timestamp is still hidden behind the player. Probably you have to move it one level up to `Page`.

steffenkleinle (changes requested):
Thanks for the PR, looks quite good already and I like the implementation. I just think that the `TtsPlayer` component should be split up to separate concerns and increase readability.

assets/icons/sound.svg
```diff
@@ -0,0 +1 @@
... 0 0 1 .281-.442c.14-.097.293-.13.459-.097a.636.636 0 0 1 .41.282c.311.407.553.896.724 1.464.177.569.266 1.162.266 1.779a5.95 5.95 0 0 1-.266 1.778 4.563 4.563 0 0 1-.724 1.464.636.636 0 0 1-.41.282.627.627 0 0 1-.459-.096zm2.994 2.1a.565.565 0 0 1-.25-.403.726.726 0 0 1 .137-.506 7.6 7.6 0 0 0 1.014-2.133c.247-.8.37-1.628.37-2.486a8.55 8.55 0 0 0-.362-2.487 7.338 7.338 0 0 0-1.022-2.132.695.695 0 0 1-.145-.5.597.597 0 0 1 .258-.41.61.61 0 0 1 .474-.104.636.636 0 0 1 .41.281c.521.698.923 1.516 1.208 2.455a9.94 9.94 0 0 1 .426 2.897c0 .992-.145 1.955-.434 2.888-.285.934-.684 1.755-1.2 2.463a.636.636 0 0 1-.41.281.658.658 0 0 1-.474-.104zm3.017 2.132a.568.568 0 0 1-.257-.418.699.699 0 0 1 .136-.5 12.144 12.144 0 0 0 1.778-4.257c.167-.798.25-1.627.25-2.485 0-.864-.083-1.693-.25-2.487-.165-.8-.402-1.553-.707-2.261-.3-.714-.657-1.379-1.07-1.996a.726.726 0 0 1-.137-.507.563.563 0 0 1 .257-.41.556.556 0 0 1 .467-.097.626.626 0 0 1 .418.298c.462.681.861 1.42 1.2 2.213.337.789.597 1.623.78 2.503.182.88.273 1.794.273 2.744 0 .95-.09 1.864-.273 2.744-.177.88-.435 1.716-.773 2.51a12.15 12.15 0 0 1-1.207 2.205.663.663 0 0 1-.418.298.556.556 0 0 1-.467-.097z" style="stroke-width:1.09867"/></svg>
```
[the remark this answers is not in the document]
bahaaTuffaha: If you noticed this icon in Figma has size 28x28 so I had to change it to 24x24 and clean it up...
some Icons like the play button is not separated (black circle, play icon) so I had to improvise 😅.
I use Inkscape and [svgomg](https://jakearchibald.github.io/svgomg/) for cleaning.

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,344 @@
+import { debounce } from 'lodash'
```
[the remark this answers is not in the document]
bahaaTuffaha: Sure I will look for a way to split it up 🙂

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,172 @@
...
+        />
+        {expandPlayer && (
+          <StyledBackForthButton accessibilityLabel='Forward Button' onPress={handleForward}>
+            <Icon Icon={PlaybackIcon} />
+            <StyledText>{t('next')}</StyledText>
+          </StyledBackForthButton>
+        )}
+      </StyledPanel>
+      {expandPlayer && (
+        <StyledPanel style={{ paddingHorizontal: 10, flexDirection: 'row' }}>
```
steffenkleinle, line 155: :wrench: Please use styled-components for this

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  return (
+    <StyledTtsPlayer $isPlaying={expandPlayer} style={elevatedButton}>
+      <StyledPanel>
+        {expandPlayer && (
+          <StyledBackForthButton accessibilityLabel='backward Button' onPress={handleBackward}>
+            <StyledText>{t('prev')}</StyledText>
+            <FlipView>
+              <Icon Icon={PlaybackIcon} />
+            </FlipView>
```
steffenkleinle, line 131: :wrench: Please use the `reverse` prop instead
```suggestion
            <Icon Icon={PlaybackIcon} reverse />
```

assets/icons/sound.svg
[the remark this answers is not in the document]
steffenkleinle: I think it would be cool to request cleaned up ones from UI/UX instead as they have some weird or most likely unnecessary syntax/clipping rules.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,219 @@
...
+  volume: 50,
+  setVolume: () => undefined,
+  sentences: [],
+  setSentences: () => undefined,
+  languageCode: '',
+})
+
+type TtsContainerProps = {
+  children: ReactElement
+  initialVisibility?: boolean
```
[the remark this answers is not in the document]
bahaaTuffaha: Yes at the test file actually: `TtsContainer.spec.tsx`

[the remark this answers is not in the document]
steffenkleinle: In general, behavior of tests should be similar and try to replicate actual usage of components. I think adding additional functionality just for tests should therefore be avoided.
How about doing the following in your test:
``` tsx
  const TestChild = () => {
    const { setVisible } = useTtsPlayer(dummyPage)
    useEffect(() => {
      setVisible(true)
    }, [setVisible])
    return null
  }
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,219 @@
...
+}
+
+const TtsContainer = ({ initialVisibility = false, children }: TtsContainerProps): ReactElement => {
+  const { languageCode } = React.useContext(AppContext)
+  const { t } = useTranslation('layout')
+  const [isPlaying, setIsPlaying] = useState(false)
+  const [expandPlayer, setExpandPlayer] = useState(false)
```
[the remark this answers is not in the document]
steffenkleinle: Hmm, shouldn't the icon then switch between paused and playing after every sentence then too as you are doing:
``` tsx
          icon={<PlayButtonIcon Icon={isPlaying ? PauseIcon : PlayIcon} />}
```

native/src/hooks/useTtsPlayer.ts
```diff
@@ -1,23 +1,35 @@
...
+      tts.setSentences(sentences)
+      tts.setTitle(model?.title ?? '')
     }
     return () => {
-      tts.setContent(null)
+      tts.setSentences([])
+      tts.setTitle('')
     }
-  }, [content, title, tts])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
bahaaTuffaha: tts is an object that changes every time so its wrong to add it as dependency.. any suggestions?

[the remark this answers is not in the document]
steffenkleinle: It shouldn't change on every render as you wrap it in a `useMemo` and all of its dependencies are from `useState`. However, the feeling that it changes too often might be because you also expose `sentenceIndex` which changes frequently. Since its not used anywhere, you can just remove that though.

assets/icons/sound.svg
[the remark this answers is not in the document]
bahaaTuffaha: sure

build-configs/common/theme/colors.ts
```diff
@@ -16,6 +16,7 @@ export type ColorsType = {
   invalidInput: string
   warningColor: string
   linkColor: string
+  themeContrast: string
   grayBackgroundColor: string
   slightlyDarkGray: string
```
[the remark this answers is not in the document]
bahaaTuffaha: Yes ... and these shades could make a difference in the look.
image

native/src/hooks/useTtsPlayer.ts
```diff
@@ -1,23 +1,35 @@
...
+    return null
+  }, [model, tts.languageCode])
 
   useEffect(() => {
-    if (content && title) {
-      tts.setContent(content)
-      tts.setTitle(title)
+    if (sentences) {
+      tts.setSentences(sentences)
+      tts.setTitle(model?.title ?? '')
```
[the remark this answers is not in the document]
bahaaTuffaha: I need it to show the title on the player...

native/src/hooks/useTtsPlayer.ts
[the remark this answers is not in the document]
bahaaTuffaha: Tried to put setSentenceIndex  in a callback function at TtsContainer but still showing:
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.

native/src/hooks/useTtsPlayer.ts
```diff
@@ -1,23 +1,35 @@
...
+import segment from 'sentencex'
 
-import { ttsContext, TtsContextType } from '../components/TtsPlayer'
+import { parseHTML } from 'shared'
+import { EventModel, LocalNewsModel, PageModel, TunewsModel } from 'shared/api'
 
-const useTtsPlayer = (content?: string, title?: string): TtsContextType => {
+import { ttsContext, TtsContextType } from '../components/TtsContainer'
+
+const useTtsPlayer = (model?: PageModel | LocalNewsModel | TunewsModel | EventModel | undefined): TtsContextType => {
```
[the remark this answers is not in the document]
bahaaTuffaha: I kept the model optional just for `useTtsPlayer` that is located at the Header to get the hook variables without passing any content.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,219 @@
...
+      subscription.remove()
+    }
+  }, [])
+
+  const stopTts = async () => {
+    await Tts.stop()
+    const TTS_STOP_DELAY = 100 // delay to make sure tts is stopped
+    await new Promise(resolve => {
+      setTimeout(resolve, TTS_STOP_DELAY)
+    })
```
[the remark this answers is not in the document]
bahaaTuffaha: Sometimes Tts.stop() wont make the tts stop so I experimented with this way and it works so I left it as it is.
I will remove the comment.

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,172 @@
...
+`
+
+const StyledPlayIcon = styled(IconButton)`
+  background-color: ${props => props.theme.colors.textColor};
+  width: 50px;
+  height: 50px;
+  border-radius: 50px;
+`
+
+const StyledBackForthButton = styled.TouchableOpacity`
```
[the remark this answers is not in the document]
bahaaTuffaha: This is native we don't have `base\Button.tsx` as we do in web.

native/src/hooks/useTtsPlayer.ts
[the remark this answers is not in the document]
steffenkleinle: Then you should be able to remove `undefined` though I think?

native/src/components/TtsPlayer.tsx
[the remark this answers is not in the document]
steffenkleinle: Sorry, I meant the `Pressable` component, my bad :)

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+        callOnEnd()
+      }
+    })
+
+    return () => {
+      Tts.removeAllListeners('tts-finish')
+      Tts.removeAllListeners('tts-progress')
+      Tts.removeAllListeners('tts-cancel')
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
steffenkleinle, line 109: :x: Please avoid disabling this linting rule. Instead, you can wrap `startReading` in a `useCallback` and add it as dependency.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const [sentenceIndex, setSentenceIndex] = useState(0)
+  const [visible, setVisible] = useState(false)
+  const [sentences, setSentences] = useState<string[]>([])
+
+  const unsupportedLanguage = unsupportedLanguagesForTts.includes(languageCode)
+  const maxTitle = 20
+  const title = sentences[0] || t('readAloud')
+  const longTitle = title.length > maxTitle ? title.substring(0, maxTitle).concat('...') : title
+
+  const callOnEnd = () => {
```
steffenkleinle, line 51: :upside_down_face: Please use consitent naming
```suggestion
  const stopPlaying = () => {
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const initializeTts = useCallback((): void => {
+    Tts.getInitStatus().catch(async error => {
+      reportError(`Tts-Error: ${error.code}`)
+      if (error.code === 'no_engine') {
+        await Tts.requestInstallEngine().catch((e: string) => reportError(`Failed to install tts engine: : ${e}`))
+      }
+    })
+  }, [])
+
+  const startReading = (index = sentenceIndex) => {
```
steffenkleinle, line 67: :wrench: Please use consistent naming and either stick to `playing` or `reading`

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+    if (!unsupportedLanguage && sentences.length > 0) {
+      // Persian not supported
```
steffenkleinle, line 69: :wrench: This should never be called in the first place if it is an unsupported language (aka tts is not enabled). Therefore this can be omitted.
```suggestion
    const currentSentence = sentences[index]
    if (currentSentence) {
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+      Tts.setDefaultLanguage(languageCode)
+      Tts.speak(String(sentences[index]), {
```
steffenkleinle, line 71: :wrench: Please avoid type casts like this, they are dangerous and can lead to unexpected behavior. Instead, properly cover edge cases (see suggestion above).
```suggestion
      Tts.speak(sentences[index], {
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+        iosVoiceId: '',
+        rate: 1,
+      })
+      if (sentences.length === 0) {
+        setSentenceIndex(0)
+        setIsExpanded(false)
+      }
+    } else {
+      setIsPlaying(false)
+    }
```
steffenkleinle, line 86: :x: I think responsibilities are currently a little unclear. This method should never have to set isPlaying to false again. Instead, it should only this method should set isPlaying to true at all (and only if there is actually something to play).
In my opinion, there should be the following methods/callback handlers:
- play
- pause
- stop (and/or close)
- playNext
- playPrevious

Each one of those should handle setting the correct state and edge cases (e.g. in the case of goForward, play should not be called if there is no next sentence).
So `play` would look something like:
``` ts
const play = (index = sentenceIndex) => {
  Tts.stop() // or custom stop() if necessary
  const sentence = sentences[index]
  if (sentence) {
    setIsPlaying(true)
    Tts.speak(...)
  }
}
```
while `playNext` would look something like this:
``` tsx
const newIndex = index + 1
if (nextIndex < sentences.length) {
  setSentenceIndex(nextIndex)
  play(nextIndex)
}
```
and similarly for the other functions.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+      if (error.code === 'no_engine') {
+        await Tts.requestInstallEngine().catch((e: string) => reportError(`Failed to install tts engine: : ${e}`))
+      }
+    })
+  }, [])
+
+  const startReading = (index = sentenceIndex) => {
+    if (!unsupportedLanguage && sentences.length > 0) {
+      // Persian not supported
+      Tts.setDefaultLanguage(languageCode)
```
steffenkleinle, line 70: :wrench: This belongs to `initializeTts`, or is it necessary to call this again every single time?

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+    Tts.addEventListener('tts-cancel', () => setIsPlaying(false))
+    Tts.addEventListener('tts-finish', () => {
+      if (sentenceIndex < sentences.length - 1) {
+        const nextIndex = sentenceIndex + 1
+        setSentenceIndex(nextIndex)
+        startReading(nextIndex)
+      } else {
+        callOnEnd()
+      }
+    })
```
steffenkleinle, line 102: Using the suggestions from above, this could be changed to simply
```suggestion
    Tts.addEventListener('tts-finish', playNext)
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+    } else {
+      setIsPlaying(false)
+    }
+  }
+
+  useEffect(() => {
+    initializeTts()
+
+    Tts.addEventListener('tts-progress', () => setIsPlaying(true))
+    Tts.addEventListener('tts-cancel', () => setIsPlaying(false))
```
steffenkleinle, line 93: :question: Why are these callbacks necessary (especially `tts-progress`)? When do they ever happen? If they are actually necessary, they should call our functions, e.g.
```suggestion
    Tts.addEventListener('tts-cancel', stop)
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+type TtsContainerProps = {
+  children: ReactElement
+}
+
+const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
+  const { languageCode } = React.useContext(AppContext)
+  const { t } = useTranslation('layout')
+  const [isPlaying, setIsPlaying] = useState(false)
+  const [isExpanded, setIsExpanded] = useState(false)
```
steffenkleinle, line 41: :x: I am still very sure that this `expanded` state is redundant if the other suggestions are applied. In my opinion, there should only be three states necessary: enabled/disabled (if disabled in the build config, no actual useState necessary), visible/hidden, playing/not playing. If it is playing, it should automatically expand (as it is now implemented). If it is paused, it should be collapsed. If the value of the useState `isPlaying` does not change anymore after every sentence/forward/backward, this property will therefore not be needed anymore.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const [sentenceIndex, setSentenceIndex] = useState(0)
+  const [visible, setVisible] = useState(false)
+  const [sentences, setSentences] = useState<string[]>([])
+
+  const unsupportedLanguage = unsupportedLanguagesForTts.includes(languageCode)
+  const maxTitle = 20
+  const title = sentences[0] || t('readAloud')
```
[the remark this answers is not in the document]
bahaaTuffaha: This only will show up if the player is showing but the content is gone if you went back to the categories list for example. what do you think @hauf-toni ?

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const longTitle = title.length > maxTitle ? title.substring(0, maxTitle).concat('...') : title
```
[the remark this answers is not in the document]
bahaaTuffaha: Nice!

native/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
hauf-toni: My recommendation for the display of the tts-Feature if there's no readable text would be to show a message like "Nothing to read" in place of the function when the button is pressed. We should also disabled the button in this state to avoid unnecessary user interaction.

Is this what you were hoping to find out?

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const callOnEnd = () => {
+    Tts.stop()
+    setIsPlaying(false)
+    setIsExpanded(false)
+    setSentenceIndex(0)
+  }
+
+  const initializeTts = useCallback((): void => {
```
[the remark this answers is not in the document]
bahaaTuffaha: If sentenceIndex or  sentences.length changes at `useEffect` this will get reinitiated again right?

native/src/components/TtsContainer.tsx
bahaaTuffaha: This needed if I changed the language while staying in the same component lifecycle I guess.

native/src/components/TtsContainer.tsx
bahaaTuffaha: Sorry for overcomplicating things 😓

native/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
bahaaTuffaha: Yes Thank you! I didn't expect a fast replay 🙂

native/src/components/TtsContainer.tsx
bahaaTuffaha: I removed them both.. you are right we are the ones who dictates when its playing not the eventListener 👍

native/src/components/Header.tsx
```diff
@@ -83,6 +85,7 @@ const Header = ({
   // Save route/canGoBack to state to prevent it from changing during navigating which would lead to flickering of the title and back button
   const [previousRoute] = useState(navigation.getState().routes[navigation.getState().routes.length - 2])
   const [canGoBack] = useState(navigation.canGoBack())
+  const { enabled: isTtsActive, setVisible: setTtsPlayerVisible } = useTtsPlayer()
```
steffenkleinle, line 88: :upside_down_face: Please keep the naming consistent
```suggestion
  const { enabled: isTtsEnabled, setVisible: setTtsPlayerVisible } = useTtsPlayer()
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+export type TtsContextType = {
+  enabled: boolean
+  setEnabled: (enable: boolean) => void
+  visible: boolean
+  setVisible: (visible: boolean) => void
+  sentences: string[] | null
+  setSentences: (sentences: string[]) => void
+}
+
+export const ttsContext = createContext<TtsContextType>({
```
steffenkleinle, line 23: :wrench: So far we are using upper case names for contexts, please change accoringly
```suggestion
export const TtsContext = createContext<TtsContextType>({
```

native/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
steffenkleinle: Fair point, you are right. Lets keep this as it is :)

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+    return () => {
+      Tts.removeAllListeners('tts-finish')
+    }
+  }, [enabled, initializeTts, playNext, sentenceIndex, sentences.length])
+
+  useEffect(() => {
+    if (sentences.length <= 1) {
+      stop()
+    }
+  }, [sentences])
```
steffenkleinle, line 129: :wrench: This `useEffect` should be removed by only passing a modified `setSentences` down that both sets the sentences and stops playing.

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+      {visible && (
+        <TtsPlayer
+          isPlaying={isPlaying}
+          setIsPlaying={setIsPlaying}
+          sentenceIndex={sentenceIndex}
+          handleBackward={playPrevious}
+          handleForward={playNext}
+          handleClose={close}
+          pauseReading={pause}
+          startPlaying={play}
```
steffenkleinle, line 173: :wrench: 
```suggestion
          playPrevious={playPrevious}
          playNext={playNext}
          close={close}
          pause={pause}
          play={play}
```

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+  )
+
+  return (
+    <ttsContext.Provider value={ttsContextValue}>
+      {children}
+      {visible && (
+        <TtsPlayer
+          isPlaying={isPlaying}
+          setIsPlaying={setIsPlaying}
+          sentenceIndex={sentenceIndex}
```
steffenkleinle, line 168: :wrench: Both `setIsPlaying` and `sentenceIndex` should not be necessary anymore, could you adjust `TtsPlayer` accordingly?

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+        play(nextIndex)
+      } else {
+        stop()
+      }
+    },
+    [play, sentenceIndex, sentences.length],
+  )
+
+  const playPrevious = (index = sentenceIndex) => {
+    const prevIndex = Math.max(0, index - 1)
```
steffenkleinle, line 105: :question: The `max` should not be needed anymore, right?
```suggestion
    const previousIndex = index - 1
```

native/src/components/TtsContainer.tsx
steffenkleinle: No worries! I think you are doing a great job :)

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,181 @@
...
+  children: ReactElement
+}
+
+const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
+  const { languageCode } = React.useContext(AppContext)
+  const { t } = useTranslation('layout')
+  const [isPlaying, setIsPlaying] = useState(false)
+  const [sentenceIndex, setSentenceIndex] = useState(0)
+  const [visible, setVisible] = useState(false)
+  const [enabled, setEnabled] = useState(false)
```
steffenkleinle, line 42: :x: There is no state needed for this as it does need to be manually set (enabled is just dependent on other states/constant values). See comments https://github.com/digitalfabrik/integreat-app/pull/2950#discussion_r1868047660 and https://github.com/digitalfabrik/integreat-app/pull/2950#discussion_r1868056687
This should be done here and only `enabled` should be passed down in the context and exposed by `useTtsPlayer`.

native/src/hooks/useTtsPlayer.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (sentences) {
+      tts.setSentences(sentences)
+    }
+    tts.setEnabled(enabled)
+    return () => {
+      tts.setSentences([])
+      tts.setEnabled(false)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [enabled, sentences, languageCode])
```
steffenkleinle, line 41: :wrench: There is no need to manually set `enabled`, it should only be defined by other values.
```suggestion
  useEffect(() => {
    setSentences(sentences)
    return () => setSentences([])
  }, [setSentences, sentences])
```

steffenkleinle (commented):
This looks really good already, thanks for taking care of all those comments! A little more to go but we are getting there :)

steffenkleinle:
One more thing: Personally I'd be in favor of always showing the header item `read aloud` if the feature is available (that is, the current language is supported and it is enabled in the build config). So in contrast to now, we would also show the header item even if there is nothing to read. The player would then just say `nothing to read`. That would keep the UI mostly consistent (except if switching to an unsupported language).

native/src/components/TtsContainer.tsx
bahaaTuffaha: Passing `sentenceIndex` is needed for `playPrevious(sentenceIndex)` and `playNext(sentenceIndex)`

steffenkleinle: There is no need to pass sentence index to these functions, they can just directly read and use it

native/src/components/TtsContainer.tsx
bahaaTuffaha: This will go minus index when you keep pressing playPrevious

steffenkleinle: Yes, but you handle that in the if in the next line, right?

native/src/components/TtsContainer.tsx
bahaaTuffaha: If I pass it directly it won't keep playing.. it will stop at an index for some reason.

native/src/components/Header.tsx
steffenkleinle: This is still valid

native/src/components/TtsContainer.tsx
steffenkleinle: @bahaaTuffaha check out my proposal here: https://github.com/digitalfabrik/integreat-app/pull/3038 For me, this works just fine without stopping at any index.

steffenkleinle:
@bahaaTuffaha please have a look at my proposal here, this fixes a few issues still open from this comment and simplifies the code in a few places. Let me know if you have any questions. Otherwise this works perfectly for me!

Changes done in the proposal:
- Avoid passing sentence indices for playNext/playPrevious
- playPrevious if already at first sentence now just starts over
- `Read aloud` is always displayed in the header overflow if enabled
- play icon is greyed out if no sentences are available

hauf-toni:
@steffenkleinle @bahaaTuffaha I would recommend using a slightly customized version of Steffen's suggestion. Users might assume that the feature doesn't work if it's missing, instead of realizing that there's no readable content on the page. On the other hand, displaying the feature all the time could lead to confusion: If a user clicks on “Read aloud” and hears nothing, she/he might feel that the feature is not working. We should instead always display the function, but show it as disabled when there is no content to read. It could also say “there is nothing to read on this page” if the user clicks on it anyway.

steffenkleinle:
> @steffenkleinle @bahaaTuffaha I would recommend using a slightly customized version of Steffen's suggestion. Users might assume that the feature doesn't work if it's missing, instead of realizing that there's no readable content on the page. On the other hand, displaying the feature all the time could lead to confusion: If a user clicks on “Read aloud” and hears nothing, she/he might feel that the feature is not working. We should instead always display the function, but show it as disabled when there is no content to read. It could also say “there is nothing to read on this page” if the user clicks on it anyway.

@hauf-toni Hmm, okay for me. What should happen then if the player is opened and a user then navigates to another site that does not have any content (e.g. just to the parent)? Do we hide the close the player then? Or just show the player as it is displayed at the moment?

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const initializeTts = useCallback((): void => {
+    Tts.getInitStatus().catch(async error => {
+      reportError(`Tts-Error: ${error.code}`)
+      if (error.code === 'no_engine') {
+        await Tts.requestInstallEngine().catch((e: string) => reportError(`Failed to install tts engine: : ${e}`))
+      }
+    })
+  }, [])
+
+  const enabled = Boolean(buildConfig().featureFlags.tts) && !unsupportedLanguagesForTts.includes(languageCode)
```
steffenkleinle, line 55: :wrench: The `Boolean` is not needed, please remove. This is due to local linting server not getting the updated build config yet. Please run `yarn ts:check` in the project root integreat-app.

steffenkleinle (approved):
Works like a charm, nice work :rocket: Good to go from my side :)

native/src/components/Header.tsx
```diff
@@ -218,7 +230,7 @@ const Header = ({
       return { text: t('locations'), language: undefined } // system language
     }
 
-    const previousRouteTitle = (previousRoute.params as { title?: string } | undefined)?.title
+    const previousRouteTitle = (previousRoute.params as { path?: string } | undefined)?.path
```
steffenkleinle, line 233: :x: Why did you change this? Please revert!

native/src/components/Header.tsx
```diff
@@ -231,7 +243,7 @@ const Header = ({
       }
     }
 
-    return { text: t(previousRoute.name), language: undefined } // system language
+    return { text: previousRoute.name, language: undefined } // system language
```
steffenkleinle, line 246: :x: Why did you change this? Please revert!

native/src/components/TtsContainer.tsx
```diff
@@ -52,14 +54,16 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
     })
   }, [])
 
-  const enabled = Boolean(buildConfig().featureFlags.tts) && !unsupportedLanguagesForTts.includes(languageCode)
+  const enabled = buildConfig().featureFlags.tts && !unsupportedLanguagesForTts.includes(languageCode)
+  const isContentAvailable = Array.isArray(sentences) && sentences.length > 0
```
steffenkleinle, line 58: :upside_down_face: Makes things clearer imo as otherwise I am wondering why it would not be an array.
```suggestion
  const isContentAvailable = sentences !== null && sentences.length > 0
```

bahaaTuffaha: if it's not enabled it will show the snackbar ...

steffenkleinle: Sorry, I don't understand, can you rephrase please?

bahaaTuffaha: I got confused by my own code 😅 NVM

native/src/components/Header.tsx
```diff
@@ -85,7 +85,11 @@ const Header = ({
   // Save route/canGoBack to state to prevent it from changing during navigating which would lead to flickering of the title and back button
   const [previousRoute] = useState(navigation.getState().routes[navigation.getState().routes.length - 2])
   const [canGoBack] = useState(navigation.canGoBack())
-  const { enabled: isTtsEnabled, setVisible: setTtsPlayerVisible } = useTtsPlayer()
+  const {
+    enabled: isTtsEnabled,
+    setVisible: setTtsPlayerVisible,
+    isContentAvailable: isTtsContentEnabled,
```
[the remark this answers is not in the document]
bahaaTuffaha: imo this looks a bit off from the rest but I will go with it..

steffenkleinle (approved):
Not tested again :) Nice job!

hauf-toni:
>@hauf-toni Hmm, okay for me. What should happen then if the player is opened and a user then navigates to another site that does not have any content (e.g. just to the parent)? Do we hide the close the player then? Or just show the player as it is displayed at the moment?

Good catch @steffenkleinle – The player should be kept open but we should also offer feedback. This solution avoids abrupt ui changes, the user retains control and there is no unexpected closing behaviour: 

➡️ The player remains open but is shown as disabled. If a user clicks anyway, it reads “There is nothing to read on this page” or similar. If the user navigates back to a page with content, the player resumes normal behavior.

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,146 @@
...
+  width: 176px;
+`
+
+const CloseView = styled.View<{ $isPlaying?: boolean }>`
+  flex-direction: column;
+  gap: 10px;
+  margin-bottom: ${props => (props.$isPlaying ? verticalMargin : 0)}px;
+`
+
+const elevatedButton = {
```
f1sh1918, line 82: is there a reason why this is set here seperately and not in the styled component?

native/src/routes/Events.tsx
```diff
@@ -57,8 +60,6 @@ const Events = ({ cityModel, language, navigateTo, events, slug, refresh }: Even
   }
 
   if (slug) {
-    const event = events.find(it => it.slug === slug)
```
f1sh1918, line 60: why this change was needed?

native/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,146 @@
...
+  width: ${props => (props.$isPlaying ? '90%' : '80%')};
+  display: flex;
+  flex-direction: ${props => (props.$isPlaying ? 'column' : 'row')};
+  justify-content: center;
+  align-items: center;
+  align-self: center;
+  position: absolute;
+  bottom: 5px;
+  min-height: 93px;
+  gap: ${props => (props.$isPlaying ? '0px;' : '10px')};
```
f1sh1918, line 23:
```suggestion
  gap: ${props => (props.$isPlaying ? '0px;' : '20px')};
```

To increase the gap between play and close button

bahaaTuffaha:
> ### Design
> Some general opinions. I would keep the size of playing and not playing container equal size. I think this jumping of the size is a bit annoying.

I followed the design from here:
https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3874-441&t=fkqv8aniSlohmWPe-0

image

f1sh1918:
> > ### Design
> > Some general opinions. I would keep the size of playing and not playing container equal size. I think this jumping of the size is a bit annoying.
> 
> I followed the design from here: https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3874-441&t=fkqv8aniSlohmWPe-0
> 
> image

yes but now there is not "Vorlesefunktion" but title which may be longer.
Sometimes you have to challenge some design if you see it in action.
I wouldn't expect a player is changing its size tbh

Unfortunately its not working for ios. Player can't be stopped.
We should either disable it for ios or not merge it yet.
I try to find a solution in the meantime

steffenkleinle:
> Unfortunately its not working for ios. Player can't be stopped. We should either disable it for ios or not merge it yet. I try to find a solution in the meantime
> 
> Could be related to this but not sure [ak1394/react-native-tts#256](https://github.com/ak1394/react-native-tts/issues/256)

There is a separate issue for iOS (#2917) and it is disabled on production (as the feature flag for everything but integreat-test-cms is set to false.

f1sh1918:
Ok i put some investigation into this.

The issue is that the action `Tts.stop()` triggers different event depending on the platform.
On Android it sends the `tts-cancel` which is fine because we don't listen to that
On Ios it sends the `tts-finish` which should be sent only on "speak finish" 
We listen to this to continue with the next sentence thats why it triggers automatically the next sentence on ios.
Since you can even on beta not stop the tts on ios i would disable it if we can not fix it.

I put also some effort in checking the native modules and thought this shouldn't be too difficult to patch but in the java and swift code the particular functions send the correct events. So i'm not pretty sure where is the issue and how to fix that

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const initializeTts = useCallback((): void => {
+    Tts.getInitStatus().catch(async error => {
+      reportError(`Tts-Error: ${error.code}`)
+      if (error.code === 'no_engine') {
+        await Tts.requestInstallEngine().catch((e: string) => reportError(`Failed to install tts engine: : ${e}`))
+      }
+    })
+  }, [])
+
+  const enabled = buildConfig().featureFlags.tts && !unsupportedLanguagesForTts.includes(languageCode)
```
steffenkleinle, line 57: Alright, lets just disable it like this:
```suggestion
  const enabled = Platform.OS === 'android' && buildConfig().featureFlags.tts && !unsupportedLanguagesForTts.includes(languageCode)
```

f1sh1918:
Yes we can do that. As far as i can see this issue was even in older version (tested several), so i'm not sure if it will be fixed or we have to think about a different implementation.

native/src/routes/Events.tsx
bahaaTuffaha: for useTtsPlayer and for less redundant code..
image

native/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,173 @@
...
+    },
+    [languageCode, sentenceIndex, sentences],
+  )
+
+  const stop = async () => {
+    setIsPlaying(false)
+    setSentenceIndex(0)
+    await Tts.stop()
+    const TTS_STOP_DELAY = 100
+    await new Promise(resolve => {
```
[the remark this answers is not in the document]
bahaaTuffaha: Screenshot from 2025-01-23 15-09-40

f1sh1918:
> > On Android it sends the `tts-cancel` which is fine because we don't listen to that
> 
> How is the behavior of tts-cancel can we use it for ios to keep the chain of speech going?

@bahaaTuffaha `tts-cancel` is never send since the `didCancelSpeechUtterance` from the `AVSpeechSynthesize` in ios is never trigged internally see the link i posted before. I added lots of comments in my pr and try to explain what happens there

One question i didn't get an answer was why the first level texts will not be read or why they are disabled for tts.

steffenkleinle:
> One question i didn't get an answer was why the first level texts will not be read or why they are disabled for tts.

That was part of the original design and discussion of the issue. Perhaps @hauf-toni can shed some more light on this. Personally I also think we should perhaps change this decision and allow to read all content (not counting children or similar). I think otherwise it is not really obvious for users how the feature works.

bahaaTuffaha:
> > > On Android it sends the `tts-cancel` which is fine because we don't listen to that
> > 
> > 
> > How is the behavior of tts-cancel can we use it for ios to keep the chain of speech going?
> 
> @bahaaTuffaha `tts-cancel` is never send since the `didCancelSpeechUtterance` from the `AVSpeechSynthesize` in ios is never trigged internally see the link i posted before. I added lots of comments in my pr and try to explain what happens there
> 
> One question i didn't get an answer was why the first level texts will not be read or why they are disabled for tts.

image

bahaaTuffaha:
> Alright, I think everything clear here for now, lets get this merged. @bahaaTuffaha can you check the failing CI?

image
 
 and what about this?

steffenkleinle:
> > Alright, I think everything clear here for now, lets get this merged. @bahaaTuffaha can you check the failing CI?
> 
> image
> 
> and what about this?

That is a separate issue/task, so we are good to merge this (if it is disabled on ios as suggested here: https://github.com/digitalfabrik/integreat-app/pull/2950#discussion_r1926637870)

hauf-toni:
> That was part of the original design and discussion of the issue. Perhaps @hauf-toni can shed some more light on this. Personally I also think we should perhaps change this decision and allow to read all content (not counting children or similar). I think otherwise it is not really obvious for users how the feature works.

Yes, I agree. This enhances predictability & clarity for users and they don't have to guess, when content is likely to be read aloud. What is the status of the design decision concerning the sizing, do you still need some input on this? @bahaaTuffaha

steffenkleinle:
> > That was part of the original design and discussion of the issue. Perhaps @hauf-toni can shed some more light on this. Personally I also think we should perhaps change this decision and allow to read all content (not counting children or similar). I think otherwise it is not really obvious for users how the feature works.
> 
> Yes, I agree. This enhances predictability & clarity for users and they don't have to guess, when content is likely to be read aloud. What is the status of the design decision concerning the sizing, do you still need some input on this? @bahaaTuffaha

@hauf-toni we have a new issue for that, see https://github.com/digitalfabrik/integreat-app/issues/3064. I think @f1sh1918 said he'll take a look at it, so perhaps you can discuss the necessary changes (in the new issue)?

Issue #2931, opened 2024-09-18
[Chat] Polling leads to scrolling to the end of the chat


steffenkleinle:
### Describe the Bug

Due to our polling, the chat scrolls to the end of all messages which can be rather annoying if you are trying to read previous or long messages.

### Steps to Reproduce

1. Go to [testumgebung](https://integreat.app/testumgebung/de)
2. Click chat button
3. Start a new conversation with lengthy messages
4. Scroll up
5. See the chat scrolling down automatically

### Expected Behavior

The expected behavior is as follows:
- No scrolling happens if no new messages are received
- The chat scrolls to the top of the first new message if there are new messages
- The text area should not loose focus

### Actual Behavior

The chat always scrolls to the end.

### Additional Information

None.

svenseeberg:
There is one other problem that is closely related: the text area looses focus when new messages are polled.

PR #2956, opened 2024-10-14
2931: Fix scrolling to end of chat if polling

steffenkleinle:
### Short description

Fix scrolling to end of chat if polling and instead only scroll to the end of the chat if there are new messages.

### Proposed changes

Fix scrolling to end of chat if polling and instead only scroll to the end of the chat if there are new messages.

### Side effects

None.

### Testing

- Test that if the chat is opened scrolling to the bottom happens
- Test that if there are new messages, scrolling to the bottom happens (wait for the next polling)
- Test that if there are no new messages, scrolling to the bottom happens (wait for the next polling)

### Resolved issues

Fixes: #2931.

---


LeandraH (approved):
Very nice, thanks! Tested in Firefox

web/src/components/ChatController.tsx
```diff
@@ -37,11 +37,11 @@ const ChatController = ({ city, language }: ChatControllerProps): ReactElement =
   const isBrowserTabActive = useIsTabActive()
 
   useEffect(() => {
-    if (!isBrowserTabActive) {
-      return undefined
+    if (isBrowserTabActive) {
```
ztefanie, line 40: I liked it the over way round more. Early exit is clearer and avoids nested if/elses if more complexity is added to this function in the future

web/src/components/ChatConversation.tsx
```diff
@@ -24,27 +24,26 @@ const ErrorSendingStatus = styled.div`
...
+    if (messagesCount < messages.length) {
+      messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' })
+      setMessagesCount(messages.length)
+    }
+  }, [messages, messagesCount])
 
   return (
     <Container className={className}>
-      {hasConversationStarted ? (
+      {messages.length > 0 ? (
```
ztefanie, line 46: (minor) I liked this named constant

ztefanie (commented):
Looks good. Just minor comments. Tested works as expected

ztefanie (approved):

web/src/components/ChatController.tsx
steffenkleinle: I get the point, but as long as there is only this little complexity, I prefer this as no negation is needed. When additional complexity is added, this can be changed again if necessary.

web/src/components/ChatConversation.tsx
steffenkleinle: The prop was redundant and a relict of old code. I therefore chose to remove it.

Issue #2948, opened 2024-10-02
[Chat] Chat ID should be unique per region


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
The chat ID should be unique per region, otherwise the chat breaks.

**Describe the solution you'd like**
The chat ID should be unique per region.

**Describe alternatives you've considered**
None.

**Additional context**
integreat-chat issue: https://github.com/digitalfabrik/integreat-chat/issues/54

steffenkleinle:
@svenseeberg it does not have to be unique per language, right?

PR #2957, opened 2024-10-14
2948: Start a new chat for each region

steffenkleinle:
### Short description

Start a new chat for each region instead of sharing the chat instance.
This would break the chat in the backend.

### Proposed changes

Store the id of the chat in the local storage related to the city.

### Side effects

:warning: Previous chats won't work anymore (unless the chat id is copied manually to the new local storage entry).

### Testing

Open the chat in testumgebung and send a message.
Then go to muenchen and open the chat and confirm that there is no chat yet.

### Resolved issues

Fixes: #2948, https://github.com/digitalfabrik/integreat-chat/issues/54.

---

### DIFF
--- a/release-notes/unreleased/2948-unique-chat-id.yml
+++ b/release-notes/unreleased/2948-unique-chat-id.yml
@@ -1,0 +1,5 @@
+issue_key: 2948
+show_in_stores: false
+platforms:
+  - web
+en: Chats are not shared across regions.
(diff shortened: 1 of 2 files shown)


steffenkleinle:
⚠️ @svenseeberg  Previous chats won't work anymore (except the chat id is copied manually to the new local storage entry).

LeandraH (approved):
Nice, tested in Firefox

ztefanie (approved):
Tested on firefox and chrome :+1:

svenseeberg:
> Previous chats won't work anymore after this PR is released to production

This is totally fine. We are currently in an alpha stage and breaking changes are totally acceptable.

PR #2960, opened 2024-10-17
2350: Show snackbar instead of opening settings

steffenkleinle:
### Short description

Explain that the settings have to be changed in order to enable push notifications if the permissions are denied.
Also fixes a bug that the push notifications are shown as enabled even if the permissions are denied. 

### Proposed changes

- Show a snackbar if trying to enable PNs but permissions are denied
- Correctly synchronize the app settings in PushNotificationsManager by using `appContext` settings

### Side effects

None.

### Testing

- Push notifications are received if enabled in the settings
- Push notifications are not received if disabled in the settings
- If permissions are denied and the user tries to enable push notifications, a snackbar is shown
- If permissions are granted during first app start, the setting is enabled
- If permissions are NOT granted during first app start, the setting is disabled
- A snackbar is shown if an error occurs during `onPress` (can be tested by just throwing an error there)

### Resolved issues

Show snackbar instead of directly opening settings, refactor settings to 

Fixes: #2350.

---


native/src/utils/PushNotificationsManager.ts
```diff
@@ -41,7 +43,7 @@ export const requestPushNotificationPermission = async (): Promise<boolean> => {
 
   if (permissionStatus !== RESULTS.GRANTED) {
     log(`Permission denied, disabling push notifications in settings.`)
-    await appSettings.setSettings({ allowPushNotifications: false })
+    updateSettings({ allowPushNotifications: false })
```
steffenkleinle, line 46: This led to unsynchronized settings.

LeandraH (approved):
Very cool, thanks! Tested in emulated Android and real iOS

native/src/routes/Settings.tsx
```diff
@@ -31,36 +29,25 @@ const SectionHeader = styled.Text`
...
-    updateSettings(newSettings)
-
-    try {
-      const successful = changeAction ? await changeAction(newSettings) : true
-
-      if (!successful) {
-        updateSettings(oldSettings)
-      }
-    } catch (e) {
+    update()?.catch(e => {
```
ztefanie, line 39: :wrench: : I am not quite sure what you want to achive, but this will only handle errors that are from async functions. If a update function is synchronous it will not be handled (like in a try/catch block). This may be no problem currently as the synchronous function do not need this error handling, bu I think it may be hard to understand / maintain. So maybe we can either also catch also errors from synchronous functions or rename this function so at least the naming makes clear this only safes async functions.

native/src/routes/Settings.tsx
```diff
@@ -31,36 +29,25 @@ const SectionHeader = styled.Text`
...
       log('Failed to persist settings.', 'error')
       reportError(e)
-      updateSettings(oldSettings)
-    }
+      appContext.updateSettings(oldSettings)
+      showSnackbar({ text: t('error:unknownError') })
```
ztefanie, line 43: :wrench: A more specific error message would be possible and nice, but would need translation of course.

native/src/utils/createSettingsSections.ts
```diff
@@ -71,34 +61,29 @@ const createSettingsSections = ({
...
-                  settings => ({
-                    allowPushNotifications: !settings.allowPushNotifications,
-                  }),
-                  async (newSettings): Promise<boolean> => {
-                    if (!cityCode) {
-                      // No city selected so nothing to do here (should not ever happen since settings are only available from city content routes)
-                      return true
-                    }
+              onPress: async () => {
+                const allowPushNotifications = !settings.allowPushNotifications
```
ztefanie, line 65: I get linting errors here, that are not on main:
ESLint: Unexpected any value in conditional. An explicit comparison or type cast is required.(@typescript-eslint/strict-boolean-expressions)

translations/translations.json
```diff
@@ -2338,6 +2338,7 @@
       "notSupportedByDevice": "Diese Funktion wird auf diesem Gerät nicht unterstützt.",
       "languageSwitchFailedTitle": "Leider ist der Sprachwechsel fehlgeschlagen.",
       "languageSwitchFailedMessage": "Die ausgewählte Sprache ist nicht offline verfügbar. Eine aktive Internetverbinding ist notwendig.",
+      "noPushNotificatidbonPermission": "Berechtigung erforderlich",
```
ztefanie, line 2341:
```suggestion
      "pushNotificationPermissionRequired": "Berechtigung erforderlich",
```

ztefanie: I would prefer to rename the key, if you disagree at least the typo needs to be fixed.

native/src/utils/createSettingsSections.ts
steffenkleinle: I cannot reproduce this and it does not happen in the CI, so I am pretty sure that it only happens on your machine. Did you try running `yarn` and `yarn ts:check` before to make sure that all types are up-to-date?

native/src/routes/Settings.tsx
steffenkleinle: I am not sure if this is desirable/helpful for users. Could you perhaps give a suggestion on what would be a better error message in this case? Something like `Failed to apply settings`/`Fehler beim Anwenden der Einstellungen`? Would this really be better/more helpful to users?

ztefanie: yes i think this is better than "unknown error", but if you disagree i am fine with leaving it as it is

native/src/utils/createSettingsSections.ts
ztefanie: 2350-linting

Did run both commands successfully. Not sure why this is only in Intellij but not in CI

ztefanie (approved):

Issue #2961, opened 2024-10-17
App crashing for news


steffenkleinle:
### Describe the Bug

The app is crashing if opening the news.

### Steps to Reproduce

0. Make sure there is at least one news item
1. Go to news (in an LTR language)
2. See app crash

### Expected Behavior

The app does not crash.

### Actual Behavior

The app does crash.

### Additional Information

This seems to be due to https://github.com/software-mansion/react-native-svg/issues/2442.

PR #2962, opened 2024-10-17
2961: Fix app crash for scaleX on svg

steffenkleinle:
### Short description

Fix app crash for scaleX on svg.

### Proposed changes

Upgrade react-native-svg.

### Side effects

None.

### Testing

See issue.

### Resolved issues

Fixes: #2961.

---

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -86,7 +86,7 @@
     "react-native-reanimated": "^3.15.1",
     "react-native-safe-area-context": "^4.10.9",
     "react-native-screens": "^3.34.0",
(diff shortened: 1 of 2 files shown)


ztefanie (approved):
Not tested

LeandraH (approved):
I couldn't reproduce the issue in emulated Android or iOS but as long as it fixed the issue for you, I'd say that's good enough.

ztefanie:
It works as expected on your branch, the app does not crash, but also for me it does not crash on main.

Issue #2966, opened 2024-10-21
Migrating yarn from classic to modern


bahaaTuffaha:
### Describe the Problem

- `yarn install` takes 98% of my memory (16 GB) for some reason.
- slow installing times (on windows)

### Describe the Preferred solution
I'm upgrading to Modern yarn as the docs recommends...

### Related Issues

Duplicates: https://github.com/digitalfabrik/integreat-app/issues/2026
Blocked by: https://github.com/yarnpkg/berry/issues/3996

### Additional Information

Last time we checked (2024-12) the blocked issue still led to problems in the CI with oom errors. For more context see https://github.com/digitalfabrik/integreat-app/pull/2968#pullrequestreview-2477815188.

PR #2967, opened 2024-10-21
2966: Yarn install takes a lot of memory while installing

bahaaTuffaha:
### Short description

`yarn install` takes 98% of my memory (16 GB) for some reason.

### Proposed changes

- Deleting yarn.lock and do yarn install again.

### Side effects

none.

Fixes: #2966 

---

### DIFF
--- a/native/src/components/__mocks__/MapView.tsx
+++ b/native/src/components/__mocks__/MapView.tsx
@@ -16,7 +16,7 @@ const MockMapView = ({
   <>
     {features.map(feature => {
       const title =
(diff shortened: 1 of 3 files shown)


LeandraH (approved):
Nice, thank you!

By the way, I would recommend making your branch names about the solution instead of the problem, you could e.g. have named this one 2966-speed-up-yarn. Also, shorter branch names are generally preferred :)

bahaaTuffaha:
I tried to reinstall now and it looks like the problem still persists 😞

Issue #2731, opened 2024-03-19
Add linting rule to use only one style of typing arrays


LeandraH:
**Is your feature request related to a problem? Please describe.**
We have places where arrays are typed as `content[]` and others where they are typed as `Array<content>`. We should decide on one and implement a linting rule to enforce that
https://typescript-eslint.io/rules/array-type/ 

**Describe the solution you'd like**

- [x] Decide on how we want to type arrays: `content[]` should be used
- [ ] Add the linting rules
- [ ] Update existing types

**Describe alternatives you've considered**
Both cases are functionally equivalent so we could also leave it.

steffenkleinle:
I guess lets vote on this one :)
- `content[]` :bread: 
- `Array<content>` :bullettrain_front: 
- I want to use both :shrug:

ztefanie:
I vote for [], as this is just the short version and [] supports readonly type modifier.

steffenkleinle:
Lets go for `content[]` as @f1sh1918 is also in favor of that. Any objections @LeandraH?

LeandraH:
No, works for me

PR #2970, opened 2024-10-24
2731: Add linting rule for arrays

simomps:
### Short description

Add a linting style for arrays

### Proposed changes

- All arrays displayed as arrayName[]
- other array style changed according to new rule

### Side effects

- None

### Testing

Check that all settings still work.

### Resolved issues

Fixes: [#2731](https://github.com/orgs/digitalfabrik/projects/2/views/3?pane=issue&itemId=56983718&issue=digitalfabrik%7Cintegreat-app%7C2731) 

---


tools/yarn.lock
```diff
@@ -136,40 +136,37 @@
     universal-github-app-jwt "^2.2.0"
     universal-user-agent "^7.0.0"
 
-"@octokit/auth-oauth-app@^7.1.0":
```
steffenkleinle, line 139: Do you know why these changes of the `yarn.lock` are present here in your PR? Usually, this file should only be modified if dependencies in a package.json are added/removed/upgraded.
Please revert this file if you did not actively or purposefully changed this file.

steffenkleinle (commented):
Congrats to your first PR, nicely done :rocket: Just one small comment regarding the `yarn.lock`, I think this is otherwise good to go!

tools/yarn.lock
simomps: Ah no. I didn't change it on purpose. I will revert the changes. 
Thanks for the notice :)

steffenkleinle (approved):
Thank you :+1:

LeandraH (commented):
Very cool, thank you very much for cleaning up  🧽 

I have found two more places where we still use the `Array` type, one is `Selector.ts` in the e2e-tests, the other one in `helpers.spec.ts` in native, might be worth a check why the linter doesn't find that. But then again, maybe not 😂

LeandraH (approved):
I'll approve it already since I'll be on vacation for a bit and don't want to block anything :)

steffenkleinle:
> Very cool, thank you very much for cleaning up 🧽
> 
> I have found two more places where we still use the `Array` type, one is `Selector.ts` in the e2e-tests, the other one in `helpers.spec.ts` in native, might be worth a check why the linter doesn't find that. But then again, maybe not 😂

Actually I am quite sure that this is something different. The two occurrences mentioned by you are both referring to
something like the following:
``` ts
const keys = new Array<string>()
```

These are initialization of a new array instance (of type string), opposed to something like the following  which actually throws an error in both cases:
``` ts
const keys: Array<string> = new Array<string>()
```

I therefore think we are actually good to go here :)

LeandraH:
Woooops, you're right, sorry!

PR #2973, opened 2024-10-25
Speed up local development android builds

steffenkleinle:
### Short description

Speed up local development android builds using `--active-arch-only` flag.
See https://reactnative.dev/docs/build-speed#build-only-one-abi-during-development-android-only.

### Testing

Android build  still works.

### Resolved issues

Fixes: None.

---

### DIFF
--- a/native/run
+++ b/native/run
@@ -21,7 +21,7 @@ program
   .option('--production', 'whether a production (release) build should be made')
   .action((buildConfigName, options) => {
     const { production } = options
-    const productionFlag = production ? '--mode=release' : ''
+    const buildFlag = production ? '--mode=release' : '--active-arch-only'
 
     const jsonBuildConfig = execSync(
       `yarn workspace --silent build-configs --silent manage to-json ${buildConfigName} android`,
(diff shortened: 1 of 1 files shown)


LeandraH (approved):
That's awesome, great find! Not tested though 😅

native/run
```diff
@@ -21,7 +21,7 @@ program
   .option('--production', 'whether a production (release) build should be made')
   .action((buildConfigName, options) => {
     const { production } = options
-    const productionFlag = production ? '--mode=release' : ''
+    const productionFlag = production ? '--mode=release' : '--active-arch-only'
```
ztefanie, line 24: This should be renamed to "buildFlag" or something similar.

ztefanie (approved):
Thanks for this improvement :rocket:

Tested on an emulated andoid (ABI x86_64). ABI was detected properly and build successfully in 3m 4s

Just one small comment.

native/run
ztefanie, line 24:
```suggestion
    const buildFlag = production ? '--mode=release' : '--active-arch-only'
```

Issue #2924, opened 2024-09-12
App crashing if local storage access is denied


steffenkleinle:
### Describe the Bug

We have sentry reports that the web app is crashing if access to the local storage is disabled:

Sentry Issue: [INTEGREAT-APP-4Y5](https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/5308/?referrer=github_integration)

### Steps to Reproduce
1. Block `integreat.app` from accessing local storage/cookies as explained in the last section [here](https://support.mozilla.org/en-US/kb/storage)
2. Go to https://integreat.app
3. See the app crash

### Expected Behavior

The app crashes.

### Actual Behavior

The app does not crash.

### Additional Information

This might be due to anti tracking restrictions. Investigate if you can reproduce this.

steffenkleinle:
I could reproduce this on firefox (desktop) by explicitly adding `integreat.app` as blocked as defined here in the last section: https://support.mozilla.org/en-US/kb/storage

steffenkleinle:
I linked the wrong issue in #2975, reopening.

PR #2974, opened 2024-10-25
2924: Catch SecurityError when accessing localStorage

steffenkleinle:
### Short description

Catch SecurityError when accessing localStorage and prevent app from crashing.

### Proposed changes

- Wrap access of localStorage in a try/catch block

### Side effects

None.

### Testing

I did not find a way to test this locally (as in: using a development server on localhost) due to me being unable to block localhost from accessing the local storage. Let me know if you can think of another way.

### Resolved issues

Fixes: #2924

---

### DIFF
--- a/web/src/hooks/useLocalStorage.ts
+++ b/web/src/hooks/useLocalStorage.ts
@@ -1,5 +1,7 @@
 import { useState, useCallback } from 'react'
 
+import { reportError } from '../utils/sentry'
+
 type UseLocalStorageProps<T> = {
   key: string
   initialValue: T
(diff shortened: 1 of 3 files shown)


ztefanie (approved):
Thanks for this PR and fixing production :rocket: :fire: 

I did test this on chrome, with adding manually an error when accessing localhost by adding this code:

```
  if (window.location.hostname === 'localhost') {
    Object.defineProperty(window, 'localStorage', {
      value: {
        getItem: () => {
          throw new Error('not allowed')
        },
        setItem: () => null,
        removeItem: () => null,
        clear: () => null,
      },
      writable: false,
    })
  }
```

App does not crash. No error is shown

:thought_balloon: At least for the chat a hint that chats cannot be save would be needed, or even the chat feature should be disabled, when it is not possible to save a chat id.

web/src/hooks/useLocalStorage.ts
```diff
@@ -12,17 +14,35 @@ type UseLocalStorageReturn<T> = {
...
 
   const updateLocalStorageItem = useCallback(
     (newValue: T) => {
-      localStorage.setItem(key, JSON.stringify(newValue))
+      try {
+        localStorage.setItem(key, JSON.stringify(newValue))
+      } catch (e) {
+        // Prevent the following error crashing the app if the browser blocks access to local storage (see #2924)
+        // SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
+        const accessDenied = e instanceof Error && e.message.includes('Access is denied for this document')
```
ztefanie, line 41: Would be nice if this code duplication can be avoided

steffenkleinle: Imo two cases of something like this in the same file is okay. If there are more, this should be extracted.

steffenkleinle:
> 💭 At least for the chat a hint that chats cannot be save would be needed, or even the chat feature should be disabled, when it is not possible to save a chat id.

I think we should not disable the feature as it is still usable as long as the tab is not closed/refreshed. I think this corresponds to opening a chat in private windows or similar. So I'd not put additional work into this right now, but if we get feeback regarding this we can still do that at a later point.

Issue #2520, opened 2023-10-11
Show number of search results


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
As a user using the search I want to know how many results are found.

**Describe the solution you'd like**
This is relevant on both web and native and probably for both the normal category search and the city search. Somewhere a number of search results should be shown. Or do you disagree and think the number of search results is not necessary?

<img src="https://github.com/digitalfabrik/integreat-app/assets/29095487/206c5384-ca62-4468-bd26-956edde6c1a0" width="200px" />
<img src="https://github.com/digitalfabrik/integreat-app/assets/29095487/4e2da6ec-686f-46b4-8be4-0a789d0438e7" width="200px" />

**Describe alternatives you've considered**
None.

**Additional context**
Relates to #2352.

nikolahoff:
@steffenkleinle can you tell me what these items are that are shown when no search term has been entered yet?
Is it possible in this case to shoe "most searched" items?
![Uploading Bildschirmfoto 2024-05-14 um 12.18.54.png…]()

nikolahoff:
<img width="1279" alt="Bildschirmfoto 2024-05-14 um 12 18 54" src="https://github.com/digitalfabrik/integreat-app/assets/97434837/ee62afff-6bc3-4ca8-85da-30b3393df150">

steffenkleinle:
> Ok, I think then we don't need to implement it. What was the reason for implementing the number of search results? @steffenkleinle

I think it was somehow related to accessibility/have a better overview. However, if you think it does not make sense from an UI/UX perspective we can also just reject this :)

nikolahoff:
@steffenkleinle [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=2944-794&t=brGhCmrpcn6PQWxj-1) is the link to the design for web and native.

steffenkleinle:
> @steffenkleinle [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=2944-794&t=brGhCmrpcn6PQWxj-1) is the link to the design for web and native.

Looks great, thanks. No comments or requests from my side :)

PR #2976, opened 2024-10-25
2520: Show number search results

simomps:
### Short description

Show the amount of search results

### Proposed changes

- Add counter in the SearchPage
- Use searchResultsCount from translation.json

### Side effects

None.

### Testing

Check css styling

### Resolved issues

Fixes: #2520 

---


web/src/routes/SearchPage.tsx
```diff
@@ -102,6 +102,9 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
       {query.length > 0 && (
         <>
           <List>
+            <p style={{ margin: '10px 5px', fontSize: 'small' }}>
```
steffenkleinle, line 105: :wrench: Generally, we use styled-components wherever possible to style our components. Please adjust your PR accordingly. For examples, just search for `= styled`, e.g. in L20 above.
You can find the docu here: https://styled-components.com/docs/basics#getting-started

web/src/routes/SearchPage.tsx
```diff
@@ -102,6 +102,9 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+              {`${t('searchResultsCount', { count: results.length })}`}
```
steffenkleinle, line 106: :upside_down_face: I think this line can be simplified a little,

steffenkleinle (commented):
Thanks for the PR, going in the right direction :tada: 
I added a few minor comments, please have a look at styled components.
Implementing this for native is still missing, let me know if you feel comfortable addressing this already, otherwise I can also take over.

One additional thing: We tend to add release notes for issues that are user facing to improve our store presence. Could you please add one for this change? Have a look here: https://github.com/digitalfabrik/integreat-app/blob/main/docs/contributing.md#release-notes

JFYI: The fact that the build is failing is not your fault, to avoid this, simply rebase your branch to main or merge main into your branch. Can also be done at a later point though.

web/src/routes/SearchPage.tsx
```diff
@@ -25,6 +25,13 @@ const List = styled.ul`
   }
 `
 
+const SearchCounter = styled.p`
+  margin: 10px 0;
+  padding-left: 5px;
```
ztefanie, line 30: As Integreat also supports languages that are right to left, like arabic, just adding padding left, will only work properly for left-to-right-languages. So you also need to add this padding right.

web/src/routes/SearchPage.tsx
```diff
@@ -25,6 +25,13 @@ const List = styled.ul`
...
+  fontsize: 'small';
```
ztefanie, line 31:
```suggestion
  font-size: 'small';
```

Also "small" is not valid value, but you can reuse the size defined in the theme
`font-size: ${props => props.theme.fonts.decorativeFontSizeSmall};`

web/src/routes/SearchPage.tsx
```diff
@@ -25,6 +25,13 @@ const List = styled.ul`
   }
 `
 
+const SearchCounter = styled.p`
+  margin: 10px 0;
```
ztefanie, line 29: I think this margin is not really needed, is it?

release-notes/unreleased/2520-Show-number-of-search-results.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2520
+show_in_stores: false
+platforms:
+  - web
+  - android and ios
+en: Show number of search results
```
ztefanie, line 6: (optional) You can also add the release note in german

release-notes/unreleased/2520-Show-number-of-search-results.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2520
+show_in_stores: false
```
ztefanie, line 2:
```suggestion
show_in_stores: true
```

Please set this to true, because this is a new feature we want this information in the stores.

native/src/routes/SearchModal.tsx
```diff
@@ -22,6 +22,12 @@ const Wrapper = styled.View`
   background-color: ${props => props.theme.colors.backgroundColor};
 `
 
+const SearchCounter = styled.Text`
+  margin: 10px 20px;
+  fontsize: 'medium';
```
ztefanie, line 27: Same as below:
1) should be "font-size" 
2) please use the value from the theme

Not quite sure, but I think if you just want to use the "normal" font size, you do not have to specify it.

ztefanie (changes requested):
Thanks for this PR. :rocket: 

Tested on Web and Android, looks already really good. :+1: 

I just added some small comments that should be easy to fix.

simomps:
Thank you for the review 🌸 

I think i did resolved all issue suggestions :))

ztefanie (approved):
Nice! Thank you for the adjustments. Looks good :rocket:

release-notes/unreleased/2520-Show-number-of-search-results.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2520
+show_in_stores: true
+platforms:
+  - web
+  - android and ios
```
steffenkleinle, line 5: :x: This won't work, this will be parsed by some of our scripts which require valid yaml files:
```suggestion
  - android
  - ios
```

For more information on yaml, have a look at https://www.yaml.info/learn/index.html or some other site :)

steffenkleinle (commented):
Nice, looks great. Tested on firefox and real android, works as expected.

You still have to run prettier
``` sh
yarn prettier:write
```

One other thing: The ticket also mentions to add this to the city search (i.e. http://localhost:9000/landing/de) on both web and native, could you still add the same functionality there? If not, thats okay as well, then we can just do a separate PR for it.

steffenkleinle (approved):
Nice work, looks great! Tested on firefox and android

ztefanie:
Please squash the commits when you are finished, buy clicking "Squash and merge" here:

native/src/components/CitySelector.tsx
```diff
@@ -26,6 +26,12 @@ const SearchBar = styled.View`
   padding: 0 10%;
 `
 
+const SearchCounter = styled.Text`
+  margin: 15px 0 10px 0;
```
steffenkleinle, line 30: We have a linting rule for shorthand properties, please use either of the following:
```suggestion
  margin: 15px 0 10px;
```
or
``` suggestion
  margin-top: 15px;
  margin-bottom: 10px;
```
Just a side note: We generally try to use sizes/paddings that are a factor of 2, i.e. 2, 4, 8, 16, 32, ... or their composites (6, 12, 24, ...) if possible

Issue #2642, opened 2024-01-22
Add tests for embedded offers/malte help offer


sarahsporck:
**Is your feature request related to a problem? Please describe.**
We are making an effort to increase our test coverage and stability, therefore we want to enforce a good practice and always write tests, when implementing a new feature ❤️ .
Due to time pressure we weren't really able to add tests for these components. Let's add them once we are finished.

**Describe the solution you'd like**
Add tests for 
- [ ] embedded offers
- [ ] malte help offer

**Describe alternatives you've considered**
Don't add tests.

**Additional context**
https://codeclimate.com/github/digitalfabrik/integreat-app/code?q=embedded
https://codeclimate.com/github/digitalfabrik/integreat-app/code?q=MalteHelp

steffenkleinle:
@LeandraH just a short heads up: A lot of offers stuff is removed (and a little refactored) in #2467. So perhaps it makes sense to wait with this for a little while still.

PR #2979, opened 2024-10-30
2642: Tests for EmbeddedOffers and MalteHelpForm

LeandraH:
### Short description

Adding some unit tests to the EmbeddedOffers and the MalteHelpForm from a while ago.

### Testing

Run the tests in your local machine, please :)

### Resolved issues

Fixes: #2642 

---


native/src/components/__tests__/EmbeddedOffers.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React from 'react'
+
+import { MALTE_HELP_FORM_OFFER_ALIAS, SPRUNGBRETT_OFFER_ALIAS } from 'shared'
+import { CategoryModel, OfferModel } from 'shared/api'
+
+import render from '../../testing/render'
+import EmbeddedOffers from '../EmbeddedOffers'
+
+jest.mock('../SprungbrettOffer.tsx', () => ({
+  __esModule: true,
```
steffenkleinle, line 11: :thinking: Is the `.tsx` and the `__esModule: true` really needed? I have never seen/needed this.

native/src/components/__tests__/MalteHelpFormOffer.spec.tsx
```diff
@@ -0,0 +1,411 @@
...
+}))
+jest.mock('react-native-safe-area-context', () => mockSafeAreaContext)
+const mockShowSnackbar = jest.fn()
+jest.mock('../../hooks/useSnackbar.ts', () => ({
+  __esModule: true,
+  default: () => mockShowSnackbar,
+}))
+
+// Typescript shenanigans to be able to use mockImplementation in the tests
+const mockedSubmitMalteHelpForm = submitMalteHelpForm as jest.MockedFunction<typeof submitMalteHelpForm>
```
steffenkleinle, line 23: :wrench: This should not be needed. Instead, you can just do `mocked(submitMalteHelForm).mockImplementation(...)`.

web/src/components/__tests__/EmbeddedOffers.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { DateTime } from 'luxon'
+import React from 'react'
+
+import { MALTE_HELP_FORM_OFFER_ALIAS, SPRUNGBRETT_OFFER_ALIAS } from 'shared'
+import { CategoryModel, OfferModel } from 'shared/api'
+
+import EmbeddedOffers from '../EmbeddedOffers'
+
+jest.mock('../SprungbrettOffer.tsx', () => ({
+  __esModule: true,
```
steffenkleinle, line 11: :wrench: Same as above.

web/src/components/__tests__/MalteHelpForm.spec.tsx
```diff
@@ -0,0 +1,362 @@
...
+import MalteHelpForm from '../MalteHelpForm'
+
+jest.mock('react-i18next')
+jest.mock('shared/api', () => ({
+  ...jest.requireActual('shared/api'),
+  submitMalteHelpForm: jest.fn(),
+}))
+
+// Typescript shenanigans to be able to use mockImplementation in the tests
+const mockedSubmitMalteHelpForm = submitMalteHelpForm as jest.MockedFunction<typeof submitMalteHelpForm>
```
steffenkleinle, line 16: :wrench: See above.

native/src/components/base/InputSection.tsx
```diff
@@ -95,6 +95,7 @@ const InputSection = ({
         returnKeyType='done'
         blurOnSubmit
         role='searchbox'
+        testID={title ?? value}
```
steffenkleinle, line 98: :upside_down_face: Testing by `testID` should be avoided if possible. Have you tried querying by role/placeholder or using `.nearest()` on the title?

steffenkleinle (approved):
All tests succeed on my machine.

native/src/components/__tests__/EmbeddedOffers.spec.tsx
LeandraH: Yes, we need it because we're mocking a default export. There's a tiny bit more information about it here: https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options

native/src/components/__tests__/MalteHelpFormOffer.spec.tsx
LeandraH: Amazing, thank you! That is so much better!

native/src/components/base/InputSection.tsx
LeandraH: I agree but I didn't manage to get it to work with any other ways I tried, including placeholder, title, or role. I don't really like going by role and then the nth of that because that makes the test pretty fragile.

native/src/components/__tests__/EmbeddedOffers.spec.tsx
steffenkleinle: Hmmm, okay. Now I am confused why we don't have to use this in a lot of other places...

steffenkleinle: Oh maybe its because you are returning an object with default instead of just returning the mock directly?

LeandraH: I'm not sure how often we mock the default export of one of our own files 🤷 Looks like we so far only use it in the `DashboardNavigationTiles`.

LeandraH: Ooooh, you're right, I found a better way, thanks!

native/src/components/__tests__/EmbeddedOffers.spec.tsx
```diff
@@ -7,21 +7,21 @@ import { CategoryModel, OfferModel } from 'shared/api'
 import render from '../../testing/render'
 import EmbeddedOffers from '../EmbeddedOffers'
 
-jest.mock('../SprungbrettOffer.tsx', () => ({
-  __esModule: true,
-  default: ({ sprungbrettOffer }: { sprungbrettOffer: OfferModel }) => {
+jest.mock('../SprungbrettOffer.tsx', () => {
```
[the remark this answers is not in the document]
LeandraH: Arg, thanks 😅

native/src/components/__tests__/EmbeddedOffers.spec.tsx
```diff
@@ -7,21 +7,21 @@ import { CategoryModel, OfferModel } from 'shared/api'
...
+  const MockedSprungbrettOffer = ({ sprungbrettOffer }: { sprungbrettOffer: OfferModel }) => {
     const { Text } = require('react-native')
     return <Text>This is the Sprungbrett offer called {sprungbrettOffer.title}</Text>
-  },
-}))
+  }
+  return MockedSprungbrettOffer
```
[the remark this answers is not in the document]
LeandraH: I honestly find it more legible than a function to a function, but I don't feel strongly about it. I can also change it if you want me to 🤷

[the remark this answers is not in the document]
steffenkleinle: I don't really have a preference on the specific code style either. I just think it should generally be uniform and consistent. I don't know though why the linter does not complain though...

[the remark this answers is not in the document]
steffenkleinle: And my IDE highlights this (which I don't like... and yes, I could just turn it off but I think generally it is useful) xD
image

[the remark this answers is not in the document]
LeandraH: I think the linter skips test files. Or at least it does in my environment 🤔  I assumed it's on purpose?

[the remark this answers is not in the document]
steffenkleinle: Some rules are definitely skipped for test files, but not this one. The no redundant local variable is just an IDE thing (see above) though. And since there is another thing than just a return in this function body, the linting rule does not complain.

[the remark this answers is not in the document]
LeandraH: Ooooh, of course. Hmm. I'm still in favour of keeping it with the extra variable just because I find functions returning functions somewhat confusing but at the same time, it's just a test file, nobody really needs to be able to understand how exactly the mocking worked here. I also definitely feel your pain of the linter marking the file as problematic, even if it's only the local one. Well, I think I just convinced myself 😂

f1sh1918 (approved):

Issue #2949, opened 2024-10-02
Granting approximate location is treated as location blocked completely


steffenkleinle:
### Describe the Bug

Since Android 12 it is possible to decide whether to grant an app precise location or just approximate location. Since we only ask for precise location (i.e. `ACCESS_FINE_LOCATION`, we get `denied` if the user only grants approximate location.

Instead of this, we should be okay with `ACCESS_COARSE_LOCATION` as well. 

### Steps to Reproduce

0. Use a device with android 12+
1. Remove location permission
2. Go the city selection
3. Click on the refresh button next to `Location access required`
4. Select `approximate`
5. See the snackbar showing that location access is blocked

### Expected Behavior

The app should take the approximate location and still work with that.

### Actual Behavior

Selecting approximate location leads to app showing the access is blocked completely.

### Additional Information

Code snippet that could be useful: https://github.com/zoontek/react-native-permissions/issues/668

PR #2980, opened 2024-10-31
2949: Allow coarse location in Android

LeandraH:
### Short description

Since Android 12, it has been possible to grant an app either the precise location or the approximate location. As default, android always asks whether to grant precise or approximate location when requesting fine location. We have been considering an accepted coarse location as a rejected location (because the request for the exact location has, indeed, been denied). This PR changes that.

### Proposed changes

- Put in a small function that, for Android, checks if we got permission for the exact location, and if not, checks for a permission for an approximate location.

### Side effects

- Shouldn't be any

### Testing

Remove location permissions, then open the app and give access to the rough position instead of the exact one, you should have the map functioning normally after. Please test this particularly on a real Android device, I only used an emulator.

### Resolved issues

Fixes: #2949 

---


release-notes/unreleased/2949-coarse-location-android.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 2949
+show_in_stores: true
+platforms:
+  - android
+en: A coarse location permission is now seen as an accepted permission
+de: Erlaubnis zur groben Ortung zählt jetzt als erlaubt
```
steffenkleinle, line 6: :wrench: Since release notes are visible to users, I think we should focus a little more on actual changes for them (and not what changed internally):
```suggestion
en: Fix location detection if only granted approximate permission.
de: Die Standortbestimmung funktioniert nun auch, wenn nur eine grobe Bestimmung erlaubt wird.
```

steffenkleinle (commented):
Tested on real android 12. Works as expected for city selection/nearby cities. Sadly, for maps, user location is still marked as blocked/not available if selecting coarse location only.

LeandraH:
Hmm, damn. The user location in the map shows up as allowed when I accept only the coarse selection, but I'm also on an emulated device. Damn. Can you check if you get the same behaviour in an emulator?

native/src/hooks/useUserLocation.ts
```diff
@@ -35,6 +32,22 @@ const getUserLocation = async (): Promise<LocationStateType> =>
     )
   })
 
+const locationPermissionIOS = PERMISSIONS.IOS.LOCATION_WHEN_IN_USE
+const fineLocationPermissionAndroid = PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION
+const coarseLocationPermissionAndroid = PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION
+
+const getLocationPermissionStatus = async (requestPermission: boolean) => {
+  const permissionFunctionToCall = requestPermission ? request : check
```
ztefanie, line 40: :upside_down_face: I think this naming is quite confusing, but might be a personal opinion
```suggestion
  const checkOrRequest = requestPermission ? request : check
```

native/src/hooks/useUserLocation.ts
```diff
@@ -35,6 +32,22 @@ const getUserLocation = async (): Promise<LocationStateType> =>
...
+  if (Platform.OS === 'ios') {
+    return permissionFunctionToCall(locationPermissionIOS)
+  }
+  const finePermission = await permissionFunctionToCall(fineLocationPermissionAndroid)
+  if (finePermission === RESULTS.GRANTED || finePermission === RESULTS.LIMITED) {
```
ztefanie, line 45: :upside_down_face: 
```suggestion
  if ([RESULTS.GRANTED, RESULTS.LIMITED].includes(finePermissionStatus)) {
  ) {
```

native/src/hooks/useUserLocation.ts
```diff
@@ -35,6 +32,22 @@ const getUserLocation = async (): Promise<LocationStateType> =>
...
+const locationPermissionIOS = PERMISSIONS.IOS.LOCATION_WHEN_IN_USE
+const fineLocationPermissionAndroid = PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION
+const coarseLocationPermissionAndroid = PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION
+
+const getLocationPermissionStatus = async (requestPermission: boolean) => {
+  const permissionFunctionToCall = requestPermission ? request : check
+  if (Platform.OS === 'ios') {
+    return permissionFunctionToCall(locationPermissionIOS)
+  }
+  const finePermission = await permissionFunctionToCall(fineLocationPermissionAndroid)
```
ztefanie, line 44: :upside_down_face: 
```suggestion
  const finePermissionStatus = await permissionFunctionToCall(fineLocationPermissionAndroid)
```

ztefanie (approved):
Thanks for this adjustment. I tested map and city-selection by location on real android. Works as expected. When granting only coarse location i see this view some times, but I expect this to be android intended behaviour:
image

Added only a few nit-pick/personal opinion code style comments.

ztefanie:
> Tested on real android 12. Works as expected for city selection/nearby cities. Sadly, for maps, user location is still marked as blocked/not available if selecting coarse location only. Am I doing something wrong?

On my android device i did not get this behaviour, but a location near my location was displayed on the map as the "standort"

steffenkleinle (approved):

steffenkleinle:
> > Tested on real android 12. Works as expected for city selection/nearby cities. Sadly, for maps, user location is still marked as blocked/not available if selecting coarse location only. Am I doing something wrong?
> 
> On my android device i did not get this behaviour, but a location near my location was displayed on the map as the "standort"

Hmm, I guess this is related to the fact that my phone is degoogled (i.e. does not have any google services that are used to determine the location installed).
I created a proposal in #2987 that mitigates that issue on my phone (by simply initializing the user location with the return of `refreshPermissionAndLocation` which we are calling anyway. Let me know what you think :)

Issue #2918, opened 2024-09-11
Add TTS on web


ztefanie:
**Is your feature request related to a problem? Please describe.**
As a analphabetic user/less fluent reader I want to have the possibility to get texts read out loud such that I can still use the app and understand the content. This feedback was given the target group contact team

**Describe the solution you'd like**
 Add an option in the UI to read the current text: See design [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3358-3&t=mEcwdGGYbx4HVbjl-1)
 This should work for categories
 It should work in all supported languages (if possible)

Leading platforms prioritize accessibility and user control over imposing character-based restrictions. I would recommend making the tts functionality available for all third level pages, providing a consistent experience for our users.

We may use these [library](https://github.com/morganney/tts-react) or search for a better option.

- We only read texts of third level categories
- target group for this feature are people with basic reading skills and/or basic language skills, not for people who cannot read at all
- should be available offline
- we do not want to read "lists with 3 elements" or stuff like this. We expect user to have basic reading skills and read the text while the tts reads it.
- we aim to support web and native, but we will start with native
- pausing should be possible
- we want to have this feature enabled by default, but design should not be annoying for people who do not want to this feature or add an easy way to turn it off / hide it.
- this feature should only be available for integreat, not for malte or aschaffenburg

**Additional context**
- See discussion this "meta/story" issue https://github.com/digitalfabrik/integreat-app/issues/2821
- See additional information on the motiviation [here](https://openproject.tuerantuer.org/projects/community-feedback/work_packages/3171/activity)

PR #2985, opened 2024-11-06
2918: Add tts on web

bahaaTuffaha:
### Short description

Adding text-to-speech functionality for web.

### Proposed changes

- `TtsPlayer.tsx` will receive content and title from `useTtsPlayer.ts` hook then it will process content as following:
     - Adding period to any `<p>` or `<li>` tag if there is non at the end.
     - Splitting the content using a library `sentencex` to support most languages.
     - Adding the title to the array.
- `TtsPlayer.tsx` using EasySpeech library for web speech api .. why I used it instead of directly using the native api read here: https://github.com/leaonline/easy-speech/blob/master/FAQ.md#why-does-this-library-exists-if-i-can-use-tts-natively-in-the-browser
- Web Speech API has some limitation on devices like Android I can't pause so I implemented a way to increment `currentSentencesIndex` at each sentence so pausing will be depend on sentence index . Another issue with Firefox about triggering onEnd function at any `.cancel()` method unlike chromium based browsers.
- This is different from tts-native: voices needs to be installed from OS so a modal will show up `TtsHelpModal` to guide users for each platform. Note: by default for windows for example you have what the keyboard layout have of languages also it may need a restart to see the newly installed language.

- Modified Modal.tsx to accept new props: icon and styling just for (backgroundColor, borderRadius)

### Side effects

- CityContentToolbar.tsx, CategoriesPage.tsx, LocalNewsPage.tsx, TuNewsDetailPage.tsx, RemoteContent.tsx, RootSwitcher.tsx

### Testing

- At integreat : Go to any 3rd level content in categories or news.
- Click on ToolbarItem on the side called readAloud
- The player will show up press play to make it start reading content.

### Observations
I tested this PR on linux and I got different result at each browser:
Chrome: works as expected + nice voices. 
Firefox: I can't pause (linux) + ~~using opensource voice library from `mbrola mb-en1` sounds robotic~~ better sounded voices can be downloaded from [pied](https://github.com/Elleo/pied?tab=readme-ov-file) (Note: you need to restart the browser after selecting a voice).
Brave: (linux) not working at all because there is no build in speech to text like google chrome does.
Windows on the other hand.. most browsers can run good as long you download the voices from the OS it self.
I didn't test it on mac OS...

### Resolved issues

Fixes: #2918 

---


ztefanie (commented):
Nice work :rocket: 

Tested this on chrome and firefox.

Firefox voice is really crappy, but chrome is really nice :rocket: Maybe we add hint, that some browsers better support this feature? Maybe @hauf-toni can share some input, how to handle this best from a UI perspective.

Found a few bug:

issue 1:
1) Open this page: http://localhost:9000/augsburg/de/willkommen/willkommen-in-augsburg-2
2) Click "read aloud"
3) Click "play"
4) Click on "Weiter"
Result: The reader starts reading out the beginning of the current sentance, when clicking "weiter" for the first time. When clicking "weiter" a second time, it correctly jumps to the next sentance
Expected result: When clicking "weiter" for the first time, it should directly jump to the next sentance.

issue 2:
When opening this page: http://localhost:9000/augsburg/de/willkommen
the "read aloud" button is displayed, but does nothing when clicked.
Expected: Should either not be displayed or does something when clicked

issue 3: 
Open this page: http://localhost:9000/augsburg/de/behoerden-beratung/behoerden/auslaenderbehoerde
Click "next" to jump to the last sentance
in this line "Weitere Informationen finden Sie unter [hier]" the reader switches to the "start"-mode, but there is still this whole contact blog that is read 


issue 4: 
When clicking pause I see the initial view of the tts-player, with the first word of the page in the heading. I think this is confusing, as clicking play does not continue reading from the beginning. I would expect that the player just changes the pause button to showing a play button.
what i get when i click pause
pause-is
what i expected
pause2

issue 5:
on firefox when the tts reaches the end of the text, the player does not reset to start as it is for other browsers.

bahaaTuffaha:
> issue 1:
> 
> 1. Open this page: http://localhost:9000/augsburg/de/willkommen/willkommen-in-augsburg-2
> 2. Click "read aloud"
> 3. Click "play"
> 4. Click on "Weiter"
>    Result: The reader starts reading out the beginning of the current sentance, when clicking "weiter" for the first time. When clicking "weiter" a second time, it correctly jumps to the next sentance
>    Expected result: When clicking "weiter" for the first time, it should directly jump to the next sentance.
> 
> issue 2: When opening this page: http://localhost:9000/augsburg/de/willkommen the "read aloud" button is displayed, but does nothing when clicked. Expected: Should either not be displayed or does something when clicked
> 
> issue 3: Open this page: http://localhost:9000/augsburg/de/behoerden-beratung/behoerden/auslaenderbehoerde Click "next" to jump to the last sentance in this line "Weitere Informationen finden Sie unter [hier]" the reader switches to the "start"-mode, but there is still this whole contact blog that is read

can you please mention which browser you used for each issue?

ztefanie:
> 
> Can you please mention which browser you used for each issue?

I mostly used Chrome, sometimes I double checked with Firefox

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,177 @@
...
+  if (isVisible) {
+    return (
+      <>
+        {showHelpModal && <TtsHelpModal closeModal={() => setShowHelpModal(false)} />}
+        <StyledTtsPlayer $isPlaying={isPlaying}>
+          <StyledPanel
+            $isPlaying={isPlaying}
+            style={{ flexDirection: theme.contentDirection === 'rtl' ? 'row-reverse' : 'row' }}>
+            {isPlaying && (
+              <StyledBackForthButton label='backward-button' onClick={playPrevious}>
```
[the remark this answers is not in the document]
bahaaTuffaha: Its used by default in the button:
image

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,165 @@
...
+        pitch: 1,
+        rate: 1,
+        volume: 0.6,
+        boundary: e => handleBoundary(e),
+        end: () => {
+          if (!isFirefox) {
+            resetOnEnd()
+          }
+        },
+      }).catch(() => null)
```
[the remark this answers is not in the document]
bahaaTuffaha: At chrome this will throw interrupted errors..

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,165 @@
...
+    } finally {
+      allowIncrement.current = false
+    }
+  }
+
+  const togglePlayPause = () => {
+    try {
+      if (isPlaying) {
+        pause()
+      } else if (currentSentencesIndex !== 0 && !isAndroid) {
```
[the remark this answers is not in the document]
bahaaTuffaha: I added a comment:
`// FirefoxAndLinux and isAndroid can't resume`

build-configs/integreat/index.ts
```diff
@@ -39,6 +39,7 @@ const commonIntegreatBuildConfig: CommonBuildConfigType = {
     fixedCity: null,
     cityNotCooperatingTemplate,
     chat: true,
+    tts: false,
```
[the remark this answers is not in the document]
bahaaTuffaha: @steffenkleinle what do you think?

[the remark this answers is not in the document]
steffenkleinle: Yes, I think so. Until iOS is ready as well I think we only want this on beta :)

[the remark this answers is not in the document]
bahaaTuffaha: Ok I will turn this to true 👍️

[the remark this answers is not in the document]
steffenkleinle: sorry, badly phrased from my side. This should stay false until ios is ready.

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,177 @@
...
+  display: flex;
+  flex-direction: ${props => (props.$isPlaying ? 'column' : 'row')};
+  justify-content: center;
+  align-items: center;
+  padding: 6px;
+  position: sticky;
+  bottom: 5px;
+  min-height: 93px;
+  gap: ${props => (props.$isPlaying ? '5px;' : '10px')};
+  border-color: transparent;
```
LeandraH, line 24: I don't think this line does anything

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,177 @@
...
+    transform 0.1s ease;
+
+  &:active {
+    box-shadow: none;
+    transform: translateY(2px);
+  }
+`
+
+const StyledPlayIcon = styled(Button)`
+  ${buttonBaseStyles}
```
LeandraH, line 55: 🔧 You could make a styled base button that you then use instead of importing the styles, so something like this:

```javascript
const BaseButton = styled(Button)`
  display: flex;
  justify-content: center;
  align-items: center;
  transition:
    box-shadow 0.2s ease,
    transform 0.1s ease;
  &:active {
    box-shadow: none;
    transform: translateY(2px);
  }
`

const StyledPlayIcon = styled(BaseButton)`
...
```

web/src/components/TtsPlayer.tsx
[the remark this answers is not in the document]
LeandraH: Nice! The labels should be translated, though :)

web/src/components/__tests__/RemoteContent.spec.tsx
```diff
@@ -1,6 +1,8 @@
 import { fireEvent } from '@testing-library/react'
 import React from 'react'
+import { MemoryRouter } from 'react-router-dom'
 
+// Import MemoryRouter if not included in renderWithTheme
```
LeandraH, line 5: Please delete the comments from the generator ;)

web/src/components/__tests__/RemoteContent.spec.tsx
```diff
@@ -10,7 +12,9 @@ describe('RemoteContent', () => {
   it('should render the html content', () => {
     const content = 'Test html'
     const { getByText } = renderWithTheme(
-      <RemoteContent html={`<div>${content}</div>`} onInternalLinkClick={() => undefined} />,
+      <MemoryRouter>
+        <RemoteContent html={`<div>${content}</div>`} onInternalLinkClick={() => undefined} />
+      </MemoryRouter>,
```
LeandraH, line 17: Maybe write a short function to not repeat this so often

web/src/components/__tests__/TtsContainer.spec.tsx
LeandraH, line 1: 💭 I don't love this test because it really tests the implementation of the component as opposed to testing roughly how a user would use the component. On the other hand, it's really hard to test third-party components, and then ones playing sounds. Not sure what to do here

web/src/contexts/TtsContextProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+import buildConfig from '../constants/buildConfig'
+
+export type TtsContextType = {
+  enabled?: boolean
+  canRead: boolean
+  visible: boolean
+  setVisible: (visible: boolean) => void
+  sentences: string[]
+  setSentences: (sentences: string[]) => void
+}
```
LeandraH, line 12:
```suggestion
}

```

web/src/contexts/TtsContextProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const TtsContext = createContext<TtsContextType>({
+  enabled: false,
+  canRead: false,
+  visible: false,
+  setVisible: () => undefined,
+  sentences: [],
+  setSentences: () => undefined,
+})
```
LeandraH, line 20:
```suggestion
})

```

web/src/contexts/TtsContextProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+type TtsContextProviderProps = {
+  children: ReactElement
+  initialVisibility?: boolean
+}
```
LeandraH, line 24:
```suggestion
}

```

web/src/hooks/useTtsPlayer.ts
```diff
@@ -0,0 +1,72 @@
...
+      const paragraphs = tempDiv.querySelectorAll('p')
+      appendPeriod(paragraphs)
+
+      const textContent = tempDiv.textContent || tempDiv.innerText
+      const content = parseHTML(textContent)
+      return [model.title.concat('.'), ...segment(languageCode, content)]
+    }
+
+    return []
+  }, [model, languageCode])
```
LeandraH, line 43: 💭 Hmmm, I don't like this hack to break up the content, and I see that you didn't do it in the native one. Maybe there is a better solution here?

LeandraH (commented):
This looks very good already, thank you! Tested in Firefox on OS X

A couple of small things:

~I would prefer that the overlay doesn't change size based on whether or not the text is being read.~ I hear that will be a separate issue :)

There were a lot of very random CSS numbers, please already apply the multiples-of-four rule of thumb that Toni mentioned when introducing the new style guide.

Also, I'm getting an unfortunate word separation in German, maybe increase the width of the side bar.
image

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const togglePlayPause = () => {
+    try {
+      const canResume = currentSentencesIndex !== 0 && !isAndroid && !isFirefoxAndLinux
+      if (isPlaying) {
+        if (isFirefoxAndLinux) {
+          stop()
+        }
+        pause()
+      } else if (canResume) {
+        // FirefoxAndLinux and isAndroid can't resume
```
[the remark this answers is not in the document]
bahaaTuffaha: I think steffi asked for these comments.. sure I will remove them 😐️

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [currentSentencesIndex, setCurrentSentenceIndex] = useState<number>(0)
+  const [showHelpModal, setShowHelpModal] = useState(false)
+  const title = sentences[0] || t('nothingToRead')
+  const longTitle = truncate(title, { maxChars: MAX_TITLE_DISPLAY_CHARS })
+  const userAgent = navigator.userAgent
+  const isAndroid = Boolean(/android/i.test(userAgent))
+  const isFirefox = userAgent.toLowerCase().includes('firefox')
+  const isLinux = userAgent.toLowerCase().includes('linux')
+  const isFirefoxAndLinux = isFirefox && isLinux
+  const enableOnEnd = useRef(true)
```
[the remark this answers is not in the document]
bahaaTuffaha: `onEnd` is function that will run at the end of a speech .. this ref is for guarding the `onEnd` from running.

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,175 @@
...
+            })
+          }
+        },
+      }).catch(() => null) // at chrome this will throw interrupted errors
+    } catch (_) {
+      setShowHelpModal(true)
+    }
+  }
+
+  const playOnFallback = (index?: number) => {
```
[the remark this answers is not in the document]
bahaaTuffaha: It will play in case of the player didn't play... simple 😎

web/src/hooks/useTtsPlayer.ts
bahaaTuffaha: hmm at native it stops normal at the end of sentences but at web I noticed lists and paragraphs without full stop will continue reading the next sentence as it one.

Other option is maybe changes from CMS ?

f1sh1918:
> This looks very good already, thank you! Tested in Firefox on OS X
> 
> A couple of small things:
> 
> ~I would prefer that the overlay doesn't change size based on whether or not the text is being read.~ I hear that will be a separate issue :)
> 
> There were a lot of very random CSS numbers, please already apply the multiples-of-four rule of thumb that Toni mentioned when introducing the new style guide.
> 
> Also, I'm getting an unfortunate word separation in German, maybe increase the width of the side bar.
> image
> 

Maybe can also adjust css styles to only break word

web/src/components/Modal.tsx
```diff
@@ -30,11 +30,20 @@ const ModalContainer = styled.div`
...
+const ModalContentContainer = styled.div<{ $style?: ModalContentStyling }>`
   position: relative;
   display: flex;
   flex-direction: column;
-  background-color: ${props => props.theme.colors.backgroundColor};
+  background-color: ${props =>
+    props.$style?.backgroundColor != null
+      ? props.theme.colors.ttsPlayerWarningBackground
+      : props.theme.colors.backgroundColor};
+  border-radius: ${props => (props.$style?.borderRadius != null ? props.$style.borderRadius : 0)}px;
```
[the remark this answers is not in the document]
bahaaTuffaha: I think I've already tried that

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const { sentences, setSentences, visible, setVisible, enabled } = useTtsPlayer()
+  const [currentSentencesIndex, setCurrentSentenceIndex] = useState<number>(0)
+  const [showHelpModal, setShowHelpModal] = useState(false)
+  const title = sentences[0] || t('nothingToRead')
+  const longTitle = truncate(title, { maxChars: MAX_TITLE_DISPLAY_CHARS })
+  const userAgent = navigator.userAgent
+  const isAndroid = Boolean(/android/i.test(userAgent))
+  const isFirefox = userAgent.toLowerCase().includes('firefox')
+  const isLinux = userAgent.toLowerCase().includes('linux')
+  const isFirefoxAndLinux = isFirefox && isLinux
```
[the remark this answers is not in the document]
bahaaTuffaha: It matters... different platform have different behaviors with tts

web/src/components/TtsPlayer.tsx
bahaaTuffaha: without it:
image

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
LeandraH: Your code doesn't use `isLinux` anywhere other than in `isFirefoxAndLinux`, and you don't use `isFirefox` anywhere other than `isFirefoxAndLinux`. So you can do `const isFirefoxAndLinux = userAgent.toLowerCase().includes('firefox') && userAgent.toLowerCase().includes('linux')` without affecting the legibility.

web/src/components/CityContentToolbar.tsx
```diff
@@ -47,19 +49,52 @@ const CityContentToolbar = (props: CityContentToolbarProps) => {
...
   return (
     <Toolbar iconDirection={iconDirection} hideDivider={hideDivider}>
       {children}
+
+      <Tooltip
+        id='tts-icon'
+        openOnClick
+        isOpen={ttsToolTip}
+        place={tooltipDirection}
+        tooltipContent={t('layout:nothingToReadFullMessage')}>
```
LeandraH, line 87: You're already in the layout context in the translation file :)

web/src/components/ModalContent.tsx
```diff
@@ -41,20 +41,33 @@ const StyledIcon = styled(Icon)`
...
+const ModalContent = ({ title, icon, style, closeModal, children, small }: ModalProps): ReactElement => {
   const { t } = useTranslation('common')
 
   return (
-    <Container>
+    <Container style={style}>
       <Header $small={small}>
-        <span>{title}</span>
+        <StyledDiv>
+          {!!icon && icon}
```
LeandraH, line 68:
```suggestion
          {icon}
```

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
LeandraH: Hmmm, I think a better name would help making it clearer why we need it :)

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
LeandraH: So, this plays in case there is an error with playing the sound but there shouldn't be, am I understanding this correctly? Then I would rename it to `fallbackPlay`, I think

LeandraH (approved):
Very cool, thank you so much! This is such a complicated feature, and you made it through that!

Tested in Firefox

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
bahaaTuffaha: This is the fallback ... sometimes the player will stop for no reason so this will check its running while `isPlaying` is true and speechSynthesis.speaking is not .. so it triggers the play again 🙂

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
LeandraH: Good thinking, though!

build-configs/common/theme/colors.ts
```diff
@@ -36,6 +38,8 @@ export const commonLightColors = {
   invalidInput: '#B3261E',
   warningColor: '#FFA726',
   linkColor: '#0b57d0',
-  grayBackgroundColor: '#dedede',
-  slightlyDarkGray: '#b9b9b9',
+  ttsPlayerWarningBackground: 'rgba(255, 253, 230, 1)',
```
f1sh1918, line 41: :wrench: is there any reason for using rgba here?
If not please convert it to hex

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,172 @@
...
+      if (selectedVoice == null || EasySpeech.status().status === 'created') {
+        setSentences([])
+        setShowHelpModal(true)
+        return
+      }
+
+      await EasySpeech.speak({
+        text: String(sentences[index]),
+        voice: selectedVoice,
+        volume: 0.6,
```
f1sh1918, line 67: we should add a rate here. For me on firefox its super fast and on chrome its okay with default rate.
It doesn't look like the default is for both browsers the same
`rate: 0.8`
worked good for me

web/src/components/TtsHelpModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+`
+
+const ModalContent = styled.div`
+  padding: 0 16px 16px 48px;
+`
+
+const StyledWarningText = styled.div`
+  font-family: ${props => props.theme.fonts.web.contentFont};
+  font-size: 12px;
+  width: 72%;
```
f1sh1918, line 26: can you explain why you use this width? Looks quite random

f1sh1918 (approved):
👍 Nice work.
Voice on macos safari is really crappy but thats not you fault

I have on styling suggestion. Can you please increase the spacing between these two button. On mobile its very close and closing and playing should have a proper distance

image

web/src/components/TtsHelpModal.tsx
bahaaTuffaha: because it's a multiple of 4 number 😅

web/src/components/TtsHelpModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <StyledText>{title}</StyledText>
+    </StyledLink>
+  </div>
+)
+
+const TtsHelpModal = ({ closeModal }: { closeModal: () => void }): ReactElement => {
+  const theme = useTheme()
+  const { t } = useTranslation('layout')
+  return (
+    <Modal
```
steffenkleinle, line 85: :x: I am not able to close the modal, neither by pressing the x button nor the overlay around the modal.

web/src/components/CityContentToolbar.tsx
```diff
@@ -47,19 +49,52 @@ const CityContentToolbar = (props: CityContentToolbarProps) => {
     }, COPY_TIMEOUT)
   }
 
+  const showTtsWarning = () => {
+    setTtsToolTip(true)
+    setTimeout(() => {
+      setTtsToolTip(false)
+    }, COPY_TIMEOUT)
```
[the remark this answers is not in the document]
steffenkleinle: :question: Why do we need a timeout to show the tooltip? I think the item should just be disabled if there is nothing to read with a normal tooltip explaining the reason. Compare to the tooltips for language change/location change for reference.

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,175 @@
...
+            closeModal={() => {
+              setShowHelpModal(false)
+              togglePlayPause()
+            }}
+          />
+        )}
+        <StyledTtsPlayer $isPlaying={isPlaying}>
+          <StyledPanel
+            $isPlaying={isPlaying}
+            style={{ flexDirection: theme.contentDirection === 'rtl' ? 'row-reverse' : 'row' }}>
```
steffenkleinle, line 143: :wrench: Please avoid inline styling, pass the language instead and apply the style in the styled component.

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  flex-direction: row;
+  gap: 4px;
+  align-items: flex-end;
+`
+
+const PlayButtonIcon = styled(Icon)`
+  color: ${props => props.theme.colors.ttsPlayerBackground};
+`
+
+const BackForthIcon = styled(Icon)<{ $flip: boolean }>`
```
steffenkleinle, line 73: :wrench: We have this functionality implemented. Please use the prop `reverse` instead.

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const theme = useTheme()
+
+  if (isVisible) {
+    return (
+      <>
+        {showHelpModal && (
+          <TtsHelpModal
+            closeModal={() => {
+              setShowHelpModal(false)
+              togglePlayPause()
```
steffenkleinle, line 136: :wrench: Is it necessary to pause here? If I understood correctly, the help modal is only shown if it is not possible to read anything anyway as no voice is installed?

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,175 @@
...
+            }}
```
steffenkleinle, line 137: :question: From a user flow I find it very confusing that this modal is only shown once I press play on the tts player. I would expect it to be directly shown once clicking on `read aloud` :thinking:

web/src/components/TtsPlayer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  playPrevious,
+  playNext,
+  close,
+  title,
+  togglePlayPause,
+}: TtsPlayerProps): ReactElement | null => {
+  const { t } = useTranslation('layout')
+  const theme = useTheme()
+
+  if (isVisible) {
```
steffenkleinle, line 129: :wrench: Please flip the condition to avoid the indentation which makes it less readable.
```suggestion
  if (!isVisible) {
    return null
  }
  
  return (
  ...
```

web/jest.setup.ts
```diff
@@ -6,6 +6,9 @@ import 'raf/polyfill'
 
 global.fetch = require('jest-fetch-mock')
 
+jest.mock('easy-speech')
+jest.mock('sentencex', () => jest.fn())
```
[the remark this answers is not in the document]
bahaaTuffaha: I remember jest was complaining so I added these 2 lines.

web/src/components/TtsContainer.tsx
```diff
@@ -0,0 +1,172 @@
+import EasySpeech from 'easy-speech'
```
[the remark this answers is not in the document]
bahaaTuffaha: This will leave everything in one place... but sure I will do it anyway 😐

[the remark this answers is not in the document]
bahaaTuffaha: I just remembered : because the useTtsPlayer uses `useLocation` hook and if I did placed <TtsContainer> in app like native it will show errors due to its location above (parent) the Router.

web/src/components/TtsHelpModal.tsx
bahaaTuffaha: what is more annoying is the font-size is smaller now because it was 14 /4 = 3.5 so I had to move it to 12.
I'm thinking of changing it to 16px..

web/src/components/CityContentToolbar.tsx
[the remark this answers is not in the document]
bahaaTuffaha: This timeout to hide the tooltip not to show it 😅

web/src/components/Modal.tsx
```diff
@@ -45,12 +49,14 @@ const ModalContentContainer = styled.div`
 
 type ModalProps = {
   title: string
+  icon?: ReactElement
+  style?: ModalContentStyling
```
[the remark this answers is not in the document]
bahaaTuffaha: Oh I did changes and forgot to remove these... I was passing value from css to styled-components or the opposite.
Nice catch!

web/src/components/StyledSmallViewTip.tsx
```diff
@@ -6,7 +6,7 @@ const StyledSmallViewTip = styled.p`
   font-weight: 400;
   margin-bottom: 0;
   margin-top: 8px;
-  white-space: normal;
+  white-space: nowrap;
```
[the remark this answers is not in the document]
bahaaTuffaha: image

web/src/components/TtsHelpModal.tsx
bahaaTuffaha: what browser did you use?

web/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
steffenkleinle: Any reason not to move it inside of the router?

web/src/components/TtsPlayer.tsx
bahaaTuffaha: Yes sometimes it says undefined

web/src/components/TtsPlayer.tsx
bahaaTuffaha: I will remove it anyway I move the visibility check at ttsContainer

web/src/components/StyledSmallViewTip.tsx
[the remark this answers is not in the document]
steffenkleinle: Thanks for explaining, looks good :+1:

web/src/components/TtsHelpModal.tsx
steffenkleinle: firefox

bahaaTuffaha: I have seen that with brave because it doesn't have voices/api so easySpeech will throw an error ... not sure if we can catch it 🫠

steffenkleinle: I don't think that is related. This is just that the modal cant be closed.

web/src/components/TtsHelpModal.tsx
LeandraH: The multiples of four is mostly for the size of elements and empty spaces between them, a font-size of 14px is totally fine.

f1sh1918:
@bahaaTuffaha i just found one issue.
1. Go to a third level page and open the  player
2. Move to a first level page
3. The Play button should be disabled if there is nothing to read. At least its the case for native implementation

image

native/src/hooks/useTtsPlayer.ts
```diff
@@ -17,8 +16,7 @@ const useTtsPlayer = (model?: PageModel | LocalNewsModel | TunewsModel | undefin
   useFocusEffect(
     useCallback(() => {
       if (model && model.content.length > 0) {
-        const content = parseHTML(model.content)
-        const sentences: string[] = segment(languageCode, content)
+        const sentences: string[] = segmentation(languageCode, addingPeriodsToDom(model.content))
```
steffenkleinle, line 19: :question: Why do we need periods here now?

native/src/components/TtsContainer.tsx
```diff
@@ -92,6 +92,10 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
       const ttsStopDelay = 100
       setTimeout(resolve, ttsStopDelay)
     })
+  }
+
+  const stopPlayer = useCallback(async () => {
+    await removeTtsListenerAndStop()
   }, [])
```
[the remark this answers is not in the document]
bahaaTuffaha: This is just for codeSence (the plugin suggested this)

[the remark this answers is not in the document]
steffenkleinle: Hmm, if CodeScene suggests this I don't think its a good suggestion :D  It only makes the code more complicated, not less. I will revert this.

web/src/hooks/useTtsPlayer.ts
```diff
@@ -0,0 +1,51 @@
...
+    // I disabled eslint due to tts changes every render
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [sentences, tts.setSentences])
+
+  useEffect(() => {
+    const synth = window.speechSynthesis
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (synth !== undefined) {
+      synth.cancel()
+    }
```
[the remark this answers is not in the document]
steffenkleinle: Okay, I will check.

native/src/hooks/useTtsPlayer.ts
bahaaTuffaha: hmm .. not actually needed for native but keeping the web and native consistent.
In general I think tts for native works differently and as I remember it didn't need a period to figure out when to stop unlike web.

native/src/components/TtsContainer.tsx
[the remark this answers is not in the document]
bahaaTuffaha: The health increased from this BTW 😂😂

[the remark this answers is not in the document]
steffenkleinle: it did get better or worse?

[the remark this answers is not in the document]
bahaaTuffaha: better from what the plugin says...

native/src/hooks/useTtsPlayer.ts
steffenkleinle: Could reproduce on firefox linux.

bahaaTuffaha: > Actually, I can't. Everything works as expected for me and the tts correctly jumps to the next sentence after each utterance/sentence, even without adding extra periods. Could you please test this and add concrete info on how to reproduce this?

Comment out the following lines at `ttsDomManipulation.ts`:
  appendPeriod(paragraphs)
  appendPeriod(listItems)

At firefox (anything doesn't have a full stop at the end will continue reading like lists): 
http://localhost:9000/muenchen/en/everyday-life/information-on-insurance-contracts-shopping-invoices-reminders-finances/bank-account

web/src/hooks/useTtsPlayer.ts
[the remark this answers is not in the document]
steffenkleinle: Couldn't find any problems with removing so removed in #3150.

steffenkleinle (approved):

f1sh1918 (changes requested):
Unfortunately safari breaks with this error if i click pause

image

f1sh1918 (commented):
@steffenkleinle i fixed some issues for safari
@LeandraH could you please test the tts player especially for safari. The voice should now sound better and start/stop should work

I tested tts on chrome, safari and firefox

f1sh1918 (approved):
Tested on chrome, firefox and safari

steffenkleinle:
> Tested on chrome, firefox and safari
> 
> Update: Okay the big issue is that not all voices work good on all browsers. For a particular language are different voices available so i can not define one standard voices which leads to the question how to solve that. We would have to check for every language all the voices for every browser

Thanks for the testing and fixing! I think we should then delegate this to the user, i.e. it is then the users responsibility to select a voice that works well in their language + browser. I don't think we can/should test all combinations of voices + languages + browsers...

steffenkleinle:
> Tested on chrome, firefox and safari
> 
> Update: Okay the big issue is that not all voices work good on all browsers. For a particular language are different voices available so i can not define one standard voices which leads to the question how to solve that. We would have to check for every language all the voices for every browser

Created https://github.com/digitalfabrik/integreat-app/issues/3174 to handle the voices issue on safari.

Issue #2996, opened 2024-11-20
Settings at native(android) is showing an error


bahaaTuffaha:
### Describe the Bug

This error look like it has to do with undefined slug when opening the settings view.

### Steps to Reproduce

1. Go to main branch (native)
2.  when I press on the settings from the menu it shows this error.

### Expected Behavior

- Showing the settings

### Additional Information

Probably introduced in #2910.

![Screenshot_1732105336](https://github.com/user-attachments/assets/39928524-5a7a-456c-96ae-4ecfb56efd69)

PR #2998, opened 2024-11-20
2996: Fix settings

LeandraH:
### Short description

I created an error via typecasting that didn't take the possibility of an undefined into account, in https://github.com/digitalfabrik/integreat-app/pull/2910 . This fixes it.

### Proposed changes

- Change typecasting to also allow undefined

### Side effects

- None

### Testing

See if you can open all pages in the app, particularly settings.

### Resolved issues

Fixes: #2996 

---

### DIFF
--- a/native/src/components/Header.tsx
+++ b/native/src/components/Header.tsx
@@ -207,8 +207,8 @@ const Header = ({
       return { text: currentTitle ?? '', language: config.sourceLanguage }
     }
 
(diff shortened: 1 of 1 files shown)


f1sh1918 (approved):
smoke tested on ios. looks good to me. thx for the quick fix

ztefanie (approved):
looks good. thanks for the quick fix.

steffenkleinle (approved):
LGTM

Issue #2745, opened 2024-03-27
Upgrade react-native


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
Its time to upgrade our dependencies again. We should update react-native as well.

This blocks
- #2553

**Testing information**
Please test everything (the whole app) very thoroughly!

**Additional context**
Other dependencies should be upgraded in #2658.

steffenkleinle:
Fixed in #2757.

steffenkleinle:
Fixed in #2757.

PR #2999, opened 2024-11-20
Update lockfiles

LeandraH:
### Short description

Just updating lockfiles

### Testing

Maybe see if a clean installation gives you the same ones?


steffenkleinle (commented):
Can you elaborate a little more on why you opened this PR and what exactly you did here? Just delete the lockfiles and add them again? Please make sure that all checks succeed and additionally run the e2e tests in the CI... when upgrading in #2745 I really struggled with that and had to revert it in the end.

LeandraH:
Closing this for now since @steffenkleinle showed me a way of not needing to delete all node_modules when switching branches and getting an error about integreat-app/shared/node_modules/@turf/distance/node_modules/tslib not existing. I just need to delete the folder shared/node_modules/@turf .

Issue #2548, opened 2023-11-08
Feedback functionality too hard to find and less used


clarabracklo:
I have received the feedback from Munich that since the new design of the feedback functionality no more feedback has been send to them whereas before they received 2 a week. This could be because the option of giving feedback is not as visible anymore without the smileys in the footer. I have also observed that users in user testings have a hard time finding the feedback functionality by themselves. 

Needs to be discussed and maybe checked with other municipalities. 

------------------------------------------------------------------------------------

Edit 23.9.24:
We will for now only change things in the web app, not the native one. There is currently one feedback button in the sidebar (footer on web mobile), we will change that to the two smileys saying useful and not useful that each open the feedback overlay with the pressed button being selected.

The design is here: https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3708-16025&node-type=frame&t=NnheR3Z60BO01kgM-0

hauf-toni:
as discussed with Martin, it might be useful to re-evaluate the visibility of the feedback function. possible solutions could be to explicitly point out the function in the intro slides and encourage users to use it. an assessment by the service team would certainly be helpful here.

clarabracklo:
I think adding this to the intro slides could be helpful however I have observed that most people just skip through the slides and even if they read them they might forget about the feedback from the time they first start the up until the time they want to give feedback?

steffenkleinle:
The easiest and most sure way to get users to send feedback would be to display a footer on every page again with feedback.

I am a little surprised anyway that its that much less feedback than before we because we only moved it to the overflow menu on native, on web the feedback button should still be shown below every page. The only change there was that we removed the two smileys and now use a comment icon. Maybe that is also not clear enough?

steffenkleinle:
The easiest and most sure way to get users to send feedback would be to display a footer on every page again with feedback.

I am a little surprised anyway that its that much less feedback than before we because we only moved it to the overflow menu on native, on web the feedback button should still be shown below every page. The only change there was that we removed the two smileys and now use a comment icon. Maybe that is also not clear enough?

dkehne:
After we have received the same feedback again at Kommunenumfrage we should discuss on the next conference. For me personally it feels like as we have hidden the feedback function on app.

clarabracklo:
Just fyi: We have also gotten the same feedback again from Kreis Viersen.

steffenkleinle:
[...]

PR #3004, opened 2024-11-23
2548: Split up web feedback at the first level into two smileys again

LeandraH:
### Short description

The municipalities have been receiving way less feedback since we made some changes there about half a year ago. In this PR, I changed the sidebar in web desktop / the footer in web mobile to show the two smileys instead of the button saying Feedback.

### Proposed changes

- Instead of showing the FeedbackIcon in the sidebar / footer, this shows the HappySmileyIcon and the SadSmileyIcon
- A click on either of those icons opens the feedback overlay with the clicked icon selected
- I added a Portal around the sidebar in web desktop to avoid the sidebar poking through when the language selection screen is open

### Side effects

- If the device is narrower than 408px, the sad smiley jumps to a new row. Do we have a minimum width that we support?

### Testing

- Check that sending feedback still works
- Check that the sidebar isn't hidden behind things it shouldn't be
- Check that the sidebar isn't hiding other things that it shouldn't be hiding

### Resolved issues

Fixes: #2548 

---

### DIFF
--- a/web/src/components/FeedbackToolbarItem.tsx
+++ b/web/src/components/FeedbackToolbarItem.tsx
@@ -13,10 +13,9 @@ import ToolbarItem from './ToolbarItem'
 type FeedbackToolbarItemProps = {
   route: RouteType
   slug?: string
-  isInBottomActionSheet: boolean
 }
 
-const FeedbackToolbarItem = ({ route, slug, isInBottomActionSheet }: FeedbackToolbarItemProps): ReactElement => {
+const FeedbackToolbarItem = ({ route, slug }: FeedbackToolbarItemProps): ReactElement => {
   const { cityCode, languageCode } = useCityContentParams()
   const [isFeedbackOpen, setIsFeedbackOpen] = useState(false)
   const [isSubmitted, setIsSubmitted] = useState(false)
@@ -27,7 +26,7 @@ const FeedbackToolbarItem = ({ route, slug, isInBottomActionSheet }: FeedbackToo
   return (
     <>
       {isFeedbackOpen && (
-        <Modal title={title} closeModal={() => setIsFeedbackOpen(false)} wrapInPortal={isInBottomActionSheet}>
+        <Modal title={title} closeModal={() => setIsFeedbackOpen(false)} wrapInPortal>
           <FeedbackContainer
             onClose={() => setIsFeedbackOpen(false)}
             onSubmit={() => setIsSubmitted(true)}
--- a/web/src/components/FeedbackToolbarItem.tsx
+++ b/web/src/components/FeedbackToolbarItem.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 
 import { FeedbackRouteType } from 'shared/api'
 
-import { FeedbackIcon } from '../assets'
+import { HappySmileyIcon, SadSmileyIcon } from '../assets'
 import useCityContentParams from '../hooks/useCityContentParams'
 import { RouteType } from '../routes'
 import FeedbackContainer from './FeedbackContainer'
@@ -20,6 +20,7 @@ const FeedbackToolbarItem = ({ route, slug, isInBottomActionSheet }: FeedbackToo
   const { cityCode, languageCode } = useCityContentParams()
   const [isFeedbackOpen, setIsFeedbackOpen] = useState(false)
   const [isSubmitted, setIsSubmitted] = useState(false)
+  const [isPositive, setIsPositive] = useState<boolean | null>(null)
   const { t } = useTranslation('feedback')
   const title = isSubmitted ? t('thanksHeadline') : t('headline')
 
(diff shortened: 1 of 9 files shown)


bahaaTuffaha:
image
here's how it looks from my device...

LeandraH:
Oooooh, good catch, thanks!

web/src/components/Layout.tsx
```diff
@@ -93,25 +94,6 @@ const Main = styled.main<{ $fullWidth: boolean }>`
   }
 `
 
-const Aside = styled.aside<{ $languageSelectorHeight: number }>`
-  top: ${props => props.$languageSelectorHeight + dimensions.headerHeightLarge + additionalToolbarTopSpacing}px;
-  margin-top: ${props => props.$languageSelectorHeight - dimensions.navigationMenuHeight}px;
```
bahaaTuffaha, line 98: ❓If it's removed I think the language selector will cover the toolbar now..
You can test it at testumgebung.

web/src/styles/Aside.css
```diff
@@ -0,0 +1,18 @@
...
+
+  @media (min-width: 1100px) {
+    left: 10%;
+  }
+
+  &:empty {
+    display: none;
+  }
+
+  &:empty + * {
```
bahaaTuffaha, line 15: ❓ hmm what this is for ? and what the adjacent component in this case?

bahaaTuffaha (changes requested):
It Looks great 👍
There is some observations .. At iPad Air screen for example: it's too close to the content:
Main:
main
This PR:

web/src/components/Layout.tsx
LeandraH: That's why I added a `Portal` :)

web/src/styles/Aside.css
LeandraH: Ooooh, good point, the adjacent component was the body, but that's not adjacent anymore!

bahaaTuffaha (approved):

web/src/components/SearchFeedback.tsx
```diff
@@ -35,6 +35,7 @@ const SearchFeedback = ({ cityCode, languageCode, query, noResults }: SearchFeed
           routeType={SEARCH_ROUTE}
           query={query}
           noResults={noResults}
+          isPositive={null}
```
ztefanie, line 38: :question: should be "false", shouldn't it?

web/src/components/FeedbackToolbarItem.tsx
```diff
@@ -13,31 +13,47 @@ import ToolbarItem from './ToolbarItem'
...
   slug?: string
-  isInBottomActionSheet: boolean
 }
 
-const FeedbackToolbarItem = ({ route, slug, isInBottomActionSheet }: FeedbackToolbarItemProps): ReactElement => {
+const FeedbackToolbarItem = ({ route, slug }: FeedbackToolbarItemProps): ReactElement => {
   const { cityCode, languageCode } = useCityContentParams()
   const [isFeedbackOpen, setIsFeedbackOpen] = useState(false)
   const [isSubmitted, setIsSubmitted] = useState(false)
+  const [isPositive, setIsPositive] = useState<boolean | null>(null)
```
ztefanie, line 22: It does not really make sense to have the same value in the parent and child as a state. I think it would make more sense to only have it in the parent and pass it (and the set function) down to the child.

ztefanie (approved):
Thanks for this PR. Tested on firefox with big and small size, works as expected :rocket:

web/src/components/SearchFeedback.tsx
LeandraH: You're right, I thought it should be null for the search but looking at the existing functionality, we send a negative feedback

LeandraH: After some research I found that the existing functionality I looked at is from a year ago and that we changed the search feedback to be neutral a while ago. I think it makes more sense for it to be negative though, so I'll open an issue for that.

web/src/components/FeedbackToolbarItem.tsx
```diff
@@ -13,31 +13,48 @@ import ToolbarItem from './ToolbarItem'
...
+        }}
+      />
+      <ToolbarItem
+        icon={SadSmileyIcon}
+        text={t('notUseful')}
+        onClick={() => {
+          setIsFeedbackOpen(true)
+          setIsPositiveRating(false)
+        }}
+      />
```
steffenkleinle, line 57: :question: Wouldn't it make more sense to just render the `FeedbackToolbarItem` twice? I think this way its pretty confusing that one feedback toolbar item renders two toolbar items. That way, there would also be no need to raise the `ispositiveRating` state to this compoennt

LeandraH: Hmm, interesting idea, I guess we could do that? Or at least rename the component. I guess I'll open another PR for that 😅

steffenkleinle: Sorry, I didn't get that the PR is already merged. If you feel motivated, otherwise this is alright as well.

LeandraH: All good, you're right, it's a good idea :)

Issue #3008, opened 2024-11-27
Don't show the loader when opening the search page


LeandraH:
**Is your feature request related to a problem? Please describe.**
When I open the [search page](https://integreat.app/testumgebung/de/search), I see a loading sign which implies to me that I can't start typing until it's done loading.

**Describe the solution you'd like**
I would like the search bar to show up without the loader under it.

**Describe alternatives you've considered**
We haven't gotten user feedback about it, so it's not exactly essential.

PR #3009, opened 2024-11-27
3008: Hide the loader when first opening the search page

LeandraH:
### Short description

When the search page opens, it shouldn't show a loader while we're loading stuff that the user can't see.

### Proposed changes

- Show only the search bar when the query is empty.

### Side effects

- Should be none

### Testing

Compare the loading of https://integreat.app/testumgebung/de/search and http://localhost:9000/testumgebung/de/search

### Resolved issues

Fixes: #3008 

---

### DIFF
--- a/web/src/routes/SearchPage.tsx
+++ b/web/src/routes/SearchPage.tsx
@@ -82,6 +82,15 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
     />
   )
 
+  if (query.length === 0) {
+    return (
+      <CityContentLayout isLoading {...locationLayoutParams}>
+        <Helmet pageTitle={pageTitle} languageChangePaths={languageChangePaths} cityModel={city} />
+        {SearchBar}
+      </CityContentLayout>
+    )
+  }
+
   if (loading || !results) {
     return (
       <CityContentLayout isLoading {...locationLayoutParams}>
--- a/web/src/routes/SearchPage.tsx
+++ b/web/src/routes/SearchPage.tsx
@@ -58,11 +58,19 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
     code,
   }))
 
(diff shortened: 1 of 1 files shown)


web/src/routes/SearchPage.tsx
```diff
@@ -82,7 +82,16 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+  if (query.length === 0) {
+    return (
+      <CityContentLayout isLoading {...locationLayoutParams}>
+        <Helmet pageTitle={pageTitle} languageChangePaths={languageChangePaths} cityModel={city} />
+        {SearchBar}
+      </CityContentLayout>
+    )
+  }
+
+  if (loading) {
```
steffenkleinle, line 94: :wrench: How about deleting both these two blocks (`if (query.length === 0)` and `if (loading)`) and just adding the suggestion below? i think the behavior should be the (nearly the) same with a lot less code. Also, it avoids not showing an error just because the query is empty.
The only difference is that results will be displayed faster even though not all endpoints have loaded yet.

web/src/routes/SearchPage.tsx
```diff
@@ -107,8 +116,8 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
       {query.length > 0 && (
         <>
```
steffenkleinle, line 117:
``` tsx
          {loading && <LoadingSpinner />}
```

web/src/routes/SearchPage.tsx
```diff
@@ -122,7 +131,7 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
           <SearchFeedback
             cityCode={cityCode}
             languageCode={languageCode}
-            noResults={results.length === 0}
+            noResults={results?.length === 0}
             query={filterText}
           />
```
steffenkleinle, line 136:
``` tsx
          {loading && <LoadingSpinner />}
```

steffenkleinle (approved):
Nice improvement! Tested on firefox

PR #3016, opened 2024-12-03
Hotfix: Fix map layout

LeandraH:
### Short description

We currently have the beautiful line `width: ${props => (props.$fullWidth ? '100%' : dimensions.maxWidth - 2 * dimensions.toolbarWidth)}px;` in our code which leads to the `width: 100%px;` when `$fullWidth` is `true`. Until merging #3004, that was not a problem (I assume it was overwritten somewhere but I honestly don't see it). After merging #3004, we have a slight layout issue here: https://webnext.integreat.app/augsburg/en/locations This PR fixes that.

### Proposed changes

- Puts the `px` where it belongs
- Updates the `max-width` to account for the `$fullWidth` prop

### Side effects

- I hope no new ones!

### Testing

Look at http://localhost:9000/testumgebung/de/locations in a desktop browser, compare it to https://webnext.integreat.app/augsburg/en/locations.

### DIFF
--- a/web/src/components/Layout.tsx
+++ b/web/src/components/Layout.tsx
@@ -74,8 +74,8 @@ const Body = styled.div<{ $fullWidth: boolean; $disableScrollingSafari: boolean 
 
 const Main = styled.main<{ $fullWidth: boolean }>`
   display: inline-block;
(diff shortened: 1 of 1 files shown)


steffenkleinle (approved):
Thanks for the quick fix! Solves the problem for me, tested on firefox.

f1sh1918 (approved):
Tested on chrome. looks good to me. thx for the quick fix

Issue #3020, opened 2024-12-04
Unexpected scrolling behavior in desktop mobile


LeandraH:
### Describe the Bug

If I scroll down on a longer page and follow a link to a shorter page, there is a lot of extra space at the bottom of the shorter page. The browser maintains the height from the previous page as a minimum.

### Steps to Reproduce

1. Restrict your browser to a smartphone width
2. Go to 'https://integreat.app/aichach-friedberg/de'
3. Scroll down to the bottom
4. Click on 'Informationen für Ehrenamtliche'
5. See that the page continues past the error

https://github.com/user-attachments/assets/9ed16336-f9f0-496d-94c3-5046f14f6987

### Expected Behavior

The page should end with the footer.

### Actual Behavior

There is a lot of "white" space below the footer.

steffenkleinle:
I remember this issue somehow, I also found this https://github.com/digitalfabrik/integreat-app/issues/2299 (which was closed as not reproducible). Can reproduce it now though.

LeandraH:
Oooooh, good memory, I'd forgotten all about that!

I have now tested this behavior in Chrome, Firefox, and Safari, it occurs in all of those.

bahaaTuffaha:
I tried to reproduce the issue on Firefox and what interesting is :
if you scrolled by swiping down .. everything is as expected but if you used the mouse to scroll it will break like this..

bahaaTuffaha:
My observation: depending on the scrolled height at the previous page it will result as a space under the html.
My failed attempt: at `web/src/index.ejs`
```
 @media (max-width: 768px) {
            html, body {
                height: 100%;
                overflow-x: hidden; 
            }
        }
```

This will cause another issue : the header won't show up until I scroll to the top.(obviously not a good solution as it could affect any scrolling behavior)

PR #3025, opened 2024-12-12
3020: Fixing the unexpected scrolling behavior at mobile

bahaaTuffaha:
### Short description

If I scroll down on a longer page and follow a link to a shorter page, there is a lot of extra space at the bottom of the shorter page. The browser maintains the height from the previous page as a minimum.

### Proposed changes

- I just added added a `height: 100% ` and `overflow-x: hidden` for mobile screen only to prevent any unexpected issues that could occur for desktop.

### Side effects

I hope none despite this applied on body/html across the app.

### Testing
You can test this also on web.
- on Chrome go to `Landkreis Aichach-Friedberg`
- Select law and finance and scroll to the end (you can notice the space is gone).

Fixes: #3020 

---

### DIFF
--- a/web/src/index.ejs
+++ b/web/src/index.ejs
@@ -110,7 +110,12 @@
                 opacity: 1;
             }
         }
-
+        @media (max-width: 768px) {
+            html, body {
+                height: 100%;
+                overflow-x: hidden; 
+            }
+        }
         #splash {
             position: fixed;
             top: 0;


bahaaTuffaha:
This solution introduced another (not sure if I want to call it a bug)... it wont show the header unless you scroll to the top.
https://github.com/user-attachments/assets/f4720f56-7b4d-46cc-9974-2f5166c8c693

steffenkleinle:
> This solution introduced another (not sure if I want to call it a bug)... it wont show the header unless you scroll to the top.
> 2024-12-12.15-56-33.mp4

Hmmm, I think this should be fixed and the header should show even if not at the top. Could you perhaps try to find another solution?

Issue #2909, opened 2024-09-02
Recommend integreat text cut off


steffenkleinle:
### Describe the Bug

The recommend integreat text is cut off on native. See screenshot below.

### Steps to Reproduce

1. Go to landing/city selection
2. Scroll to the end
3. Click on `Suggest Integreat to your region`
4. Scroll down
5. See that the end is cut off

### Expected Behavior

The text not to be cut off.

### Actual Behavior

The text is cut off.

### Additional Information

<img src="https://github.com/user-attachments/assets/67f6b48a-26e9-4bb0-bab2-efbfdef479e1" width="200px" />

PR #3033, opened 2025-01-09
2909: Fix copy text being cut off in the suggest your region page

lunars97:
### Short description

Currently the copy text is not shown fully, half of it is cut off. The text should be fully shown.

### Proposed changes

Fix the scroll view and the template text's style

### Side effects

none

### Resolved issues

Fixes: #2909 

---

### DIFF
--- a/native/src/routes/CityNotCooperating.tsx
+++ b/native/src/routes/CityNotCooperating.tsx
@@ -8,7 +8,7 @@ import TextButton from '../components/base/TextButton'
 import buildConfig, { buildConfigAssets } from '../constants/buildConfig'
 
 const Container = styled.ScrollView`
-  display: flex;
+  flex: 1;
   padding: 30px;
 `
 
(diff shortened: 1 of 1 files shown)


LeandraH (commented):
Woohoo, welcome back! 🥳 🥳 🥳

This does the job but I think there are easier ways to do it, especially since the text in the template is always German, and so the length shouldn't change too much. For example, I was able to get the entire text to show up just by increasing the `margin-bottom` in the `TemplateText`.

lunars97:
> Woohoo, welcome back! 🥳 🥳 🥳
> 
> This does the job but I think there are easier ways to do it, especially since the text in the template is always German, and so the length shouldn't change too much (though we do still have different device sizes to take into account). For example, I was able to get the entire text to show up just by increasing the `margin-bottom` in the `TemplateText`.

yes, you are totally right. I think I oversaw the changes when I was trying it :)

native/src/routes/CityNotCooperating.tsx
```diff
@@ -8,7 +8,7 @@ import TextButton from '../components/base/TextButton'
 import buildConfig, { buildConfigAssets } from '../constants/buildConfig'
 
 const Container = styled.ScrollView`
-  display: flex;
```
LeandraH, line 11: Good catch!

LeandraH (approved):
Nicely done! Tested on an emulated iPhone 16 and an emulated iPad Air 13-inch.

bahaaTuffaha (approved):
Tested (emulation) on Pixel 2: 5.0″ screen and Pixel 7 Pro: 6.71″ screen
Works great!👍️

Issue #2926, opened 2024-09-14
Search function should have german fallback


ztefanie:
**Is your feature request related to a problem? Please describe.**
As a user who's native language is arabic and who also speak german, i am frustrated when entering a german word to the search an no results are displayed

**Describe the solution you'd like**
- always download the german language for a region
- do display german results if there are no results in the language the user selected
- if in both languages there are no results show a text "no results found in German and Arabic, ~want to search in a different language "button to change langugage"~
- implement new design of giving feedback about missing information

**Describe alternatives you've considered**
We discussed only adding hint text, but decided against it, as this is a very important key feature of Integreat

**Additional context**
This feature request is a requirement from the target group, see [here](https://openproject.tuerantuer.org/projects/community-feedback/work_packages/3232) 

If needed split this issue into multiple PRs, e.g. 
- one pr for downloading the additional language: #3024 
- redesign of the search in native: #3065 
- redesigns of the search in web: #3035

hauf-toni:
design can be found [📍 here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=3462-118&t=0pYuzllOyXtdKEhN-1)

hauf-toni:
fyi: decided against including a CTA for language change, as too many direct options/ctas would overwhelm users and this function should be generally known

LeandraH:
I would say the design looks good and we can send it on to the Integreat-App board?

ztefanie:
> I would say the design looks good and we can send it on to the Integreat-App board?

done

PR #3035, opened 2025-01-09
2926: Redesign of search feedback in web

LeandraH:
### Short description

This is the implementation of the redesign of the search feedback in web.

### Proposed changes

- When there are no search results, the text makes it clearer that we also searched in the fallback language, and a suggestion to try a different language.
- When there are no search results, there is a very clear CTA to open the feedback form if the user thinks there is information missing anyway.
- The design also calls for the number of search results to disappear if there aren't any but I decided against that so that a screen reader can still yell at someone to stop typing if they already have no results (#2352)

### Side effects

- The sad smiley in the search feedback disappears (though I'm not sure that qualifies as a side effect since that's the design)
- I used German as the hard-coded fallback language in the translations. Should we ever have the app with a different fallback language by going international, we'll have to fix that. I decided to kick that problem down the road because it's not that likely a scenario, and we would probably have to check through the entire app anyway in that case.

### Testing

You can switch the app to English and search for "alleinerziehend".

### Resolved issues

Partially fixes: #2926 

---


web/src/components/SearchFeedback.tsx
```diff
@@ -13,6 +15,22 @@ const Container = styled.div`
...
+const CenteredContainer = styled.div`
+  text-align: center;
+`
+
+const SemiBoldText = styled.p`
+  font-weight: 600;
+`
+
+const MiddleText = styled.p`
+  padding-bottom: 1rem;
```
steffenkleinle, line 27: :question: Hmm, did I miss something or why did you switch to rem? Is this already part of the new design system? Otherwise I'd prefer to keep it consistent.

web/src/components/SearchFeedback.tsx
```diff
@@ -26,21 +44,35 @@ const SearchFeedback = ({ cityCode, languageCode, query, noResults }: SearchFeed
...
 
+  if (noResults) {
+    const fallbackLanguage = config.sourceLanguage
+
+    return (
+      <CenteredContainer>
+        <SemiBoldText>
+          {languageCode === fallbackLanguage ? t('noResultsInOneLanguage') : t('noResultsInTwoLanguages')}
+        </SemiBoldText>
+        <MiddleText>{t('checkQuery', { appName: buildConfig().appName })}</MiddleText>
```
steffenkleinle, line 69: :upside_down_face: I'd try not to give names based on positioning as that might change in the future and instead rather use names more related to function.
```suggestion
        <Hint>{t('checkQuery', { appName: buildConfig().appName })}</Hint>
```

translations/translations.json
```diff
@@ -4763,7 +4763,12 @@
       "positiveRating": "Diese Seite ist hilfreich",
       "searchTermDescription": "Für den folgenden Begriff konnte kein Ergebnis gefunden werden:",
       "contactMailAddress": "E-Mail für Rückfragen",
-      "note": "Bitte wähle eine Reaktion oder schreibe einen Kommentar, um das Feedback abschicken zu können."
+      "note": "Bitte wähle eine Reaktion oder schreibe einen Kommentar, um das Feedback abschicken zu können.",
+      "giveFeedback": "Feedback geben",
+      "checkQuery": "Überprüfen Sie Ihren Suchbegriff oder ändern Sie die eingestellte Sprache der {{appName}}-App.",
+      "informationMissing": "Hier fehlen Informationen?",
+      "noResultsInOneLanguage": "Es wurden leider keine passenden Ergebnisse in Deiner Sprache gefunden.",
+      "noResultsInTwoLanguages": "Es wurden leider keine passenden Ergebnisse in Deiner Sprache oder auf Deutsch gefunden."
```
steffenkleinle, line 4771: :wrench: I think `oneLanguage` and `twoLanguage` are not really telling. Instead I would just do something like
```suggestion
      "noResultsInUserLanguage": "Es wurden leider keine passenden Ergebnisse in Deiner Sprache gefunden.",
      "noResultsInUserAndSourceLanguage": "Es wurden leider keine passenden Ergebnisse in Deiner Sprache oder auf Deutsch gefunden."
```

steffenkleinle: :thinking: If we ever change the source language, for example for a new whitelabel in some other country, this kinda breaks. Since this is not planned at the moment and we have translation overrides, this is probably okay. But definitely something to keep in mind for the future. Perhaps `Originalsprache` would also work? But probably not as clear/easily understandable.

steffenkleinle (approved):
Nice work!

translations/translations.json
```diff
@@ -4947,7 +4952,12 @@
       "positiveRating": "This site is useful",
       "searchTermDescription": "No result could be found for the following term:",
       "contactMailAddress": "E-Mail for further questions",
-      "note": "Please select a reaction or write a comment in order to send feedback."
+      "note": "Please select a reaction or write a comment in order to send feedback.",
+      "giveFeedback": "Give feedback",
+      "checkQuery": "Check your search term or select a different language in the {{appName}} app",
```
steffenkleinle, line 4957:
```suggestion
      "checkQuery": "Check your search term or select a different language in the {{appName}} app.",
```

web/src/components/SearchFeedback.tsx
LeandraH: You're right, not sure why I did that 😅

steffenkleinle (approved):

bahaaTuffaha (approved):
Great work  .. tested on Firefox and Chrome

Issue #2881, opened 2024-08-13
POI filter overlay is off-center in Arabic in iOS


LeandraH:
### Describe the Bug

When opening the POI filter overlay on my iPhone 13 Mini when it's set to Arabic, the entire overlay is too far to the right and looks slightly cut off.

### Steps to Reproduce

1. Be on iOS, possibly a 13 mini
2. Set the phone language to Arabic
3. Open Integreat
4. Open the maps screen
5. Open the filter overlay
6. See that everything is too far to the right

### Expected Behavior

The contents of the overlay should be centered.

### Actual Behavior

The contents of the overlay are off to the right.

### Additional Information

![IMG_7DF659D47E80-1](https://github.com/user-attachments/assets/6719cc8c-a626-4c0e-9a09-3a92438a72e5)

PR #3037, opened 2025-01-11
2881: Fix weird modal bug

LeandraH:
### Short description

I couldn't really find anything about why but our `Modal` centers its content (correctly) on Android for rtl and ltr and on iOS for ltr. On iOS for rtl, the content is aligned to the right. This fixes it by using margin instead of padding. It makes no sense to me but I decided an hour of googling is enough for a very minor bug.

### Proposed changes

- Use margin instead of padding.

### Side effects

- None

### Testing

Test on iOS, set the device language to a rtl language (e.g. Arabic), go to the map and open the overlay with the POI filters. It should be centered now.

### Resolved issues

Fixes: #2881

---


steffenkleinle (approved):
Very weird.... tested on iOS and Android, fixes the bug as expected. Nice!

f1sh1918 (approved):
works fine. thx for the fix 👍 
Tested only on ios

PR #3038, opened 2025-01-16
2916: TTS android proposal

steffenkleinle:
### Proposed changes

-
-

### Side effects

-
-

### Resolved issues

Fixes: #

---

### DIFF
--- a/native/src/components/TtsContainer.tsx
+++ b/native/src/components/TtsContainer.tsx
@@ -52,7 +52,7 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
     })
   }, [])
 
-  const enabled = buildConfig().featureFlags.tts && !unsupportedLanguagesForTts.includes(languageCode)
+  const enabled = Boolean(buildConfig().featureFlags.tts) && !unsupportedLanguagesForTts.includes(languageCode)
 
   const play = useCallback(
     (index = sentenceIndex) => {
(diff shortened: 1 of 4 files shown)


native/src/components/TtsContainer.tsx
```diff
@@ -93,25 +89,20 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
+    if (nextIndex < sentences.length) {
+      setSentenceIndex(nextIndex)
+      play(nextIndex)
+    } else {
+      stop()
     }
+  }, [play, sentenceIndex, sentences.length])
+
+  const playPrevious = () => {
+    const previousIndex = Math.max(0, sentenceIndex - 1)
```
steffenkleinle, line 103: This aligns the behavior to that of most media players to just start the first sentence again

native/src/components/TtsContainer.tsx
```diff
@@ -93,25 +89,20 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
-    },
-    [play, sentenceIndex, sentences.length],
-  )
-
-  const playPrevious = (index = sentenceIndex) => {
-    const previousIndex = index - 1
-    if (previousIndex >= 0) {
-      setSentenceIndex(previousIndex)
-      play(previousIndex)
+  const playNext = useCallback(() => {
```
steffenkleinle, line 92: @bahaaTuffaha for me it is definitely not necessary to pass the sentece index here.

native/src/components/TtsContainer.tsx
bahaaTuffaha: I think that what it was before 🫣

PR #3046, opened 2025-01-23
2916: Add tts on ios

f1sh1918:
### Short description

Ios internally uses the wrong function for stop tts and is therefore sending the wrong Events and even twice.
I don't think we can do here anything, i already checked the native bindings if we could patch there something. 
Therefore i added some tweaks here to workaround the event sending issues we have on ios.
Thats really not a clean way so feel free to provide a better solution

### Proposed changes

- implement a suppress function for ios to skip wrongly triggered finish event
- use pause function instead of stop for ios. This enables to continue at the exact position and avoids event finish issues
- since manual play next or previous triggers immeditately  'tts-finish' that triggers playNextAuto (Event listener) , i had to adjust the particular function for ios.
- add a working speech rate for ios

### Issues
- Note at the first sentence in some case it repeats the sentence instead of moving forward. I think thats okay for some beta state.

### Side effects

- hopefully none

### Testing

1. Go to second level text and open "Vorlesefunktion" via "..." Menu. Start and stop reading. Even go next and previous. 

### Resolved issues

Fixes: #2917

---

### DIFF
--- a/native/src/components/TtsContainer.tsx
+++ b/native/src/components/TtsContainer.tsx
@@ -33,7 +33,7 @@ export const TtsContext = createContext<TtsContextType>({
 type TtsContainerProps = {
   children: ReactElement
 }
-
+// TODO Sometimes on ios when clicking playNext it reads the sentences and stop.
 const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
   const { languageCode } = React.useContext(AppContext)
   const { t } = useTranslation('layout')
@@ -44,6 +44,7 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
   const title = sentences[0] || t('nothingToRead')
   const longTitle = truncate(title, { maxChars: MAX_TITLE_DISPLAY_CHARS })
   const unsupportedLanguagesForTts = ['fa', 'ka', 'kmr']
+  const [suppressFinishEvent, setSuppressFinishEvent] = useState(false)
 
   const initializeTts = useCallback((): void => {
     Tts.getInitStatus().catch(async error => {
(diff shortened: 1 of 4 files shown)


native/src/components/TtsContainer.tsx
```diff
@@ -72,42 +92,64 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
+    } else {
+      stop().then()
     }
-  }, [play, sentenceIndex, sentences.length])
+  }, [play, sentenceIndex, sentences.length, stop])
 
   const playPrevious = () => {
-    const previousIndex = Math.max(0, sentenceIndex - 1)
+    // Since ios immediately triggers 'tts-finish' which updates the index, we have to reduce by 2.
+    const correctIndex = Platform.OS === 'ios' ? 2 : 1
```
f1sh1918, line 149: i tried here to use also the suppress function instead of reducing by two but then the previous sentence will be read and than it stops

native/src/components/TtsContainer.tsx
```diff
@@ -72,42 +92,64 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
-      stop()
+      stop().then()
+    }
+  }, [play, sentenceIndex, sentences.length, stop])
+
+  const playNext = useCallback(() => {
+    const nextIndex = sentenceIndex + 1
+    if (nextIndex < sentences.length) {
+      // do not update index  because ios immediately triggers playNextAuto due to `tts-finish` event, so it will be updated there
+      if (Platform.OS === 'ios') {
```
f1sh1918, line 136: also tried here use suppress function instead of not updating the index but it also leads to that the next sentence will be read and then it will be stopped

f1sh1918:
I really tried to create a less "hacky" solution by introducing this "suppressEvent" function but to use it in all "manual" triggers like previous, next and pause also leads to issue, so i had to use even more hacky workarounds.
Even it seem to work quite fine now for both platforms, i'm aware thats far away from a good solutions, so feel free to challenge that and propose sth better. I really spent a huge amount of hours in trying ;)

Also this ios issue that i linked is opened since ios15 (4years+) i don't expect that this will be fixed soonish tbh.

LeandraH (approved):
Wow, that looks pretty ugly indeed, thank you for explaining at each step of the way why you were doing what, and for slogging through that mess! Tested on real iOS

f1sh1918:
> Wow, that looks pretty ugly indeed, thank you for explaining at each step of the way why you were doing what, and for slogging through that mess! Tested on real iOS

When I realized what the issue is should have come the conclusión to just remove the event listener and readd. So I guess this PR is outdated and we should check Steffens. Sometimes i get lost in the woods 😁

f1sh1918:
In #3063 a cleaner solution can be found

PR #3050, opened 2025-01-26
3020: Fix unexpected scrolling behaviour in mobile desktop

lunars97:
### Short description

Currently the category pages are having additional scrollable blank height after the toolbar footer on the small screens if the side kebab-menu is activated. This might be happening because of the `visibility: hidden` attribute of the `Portal` potentially taking up unintended space due to it being a part of the DOM, therefore it could still occupy space.

### Proposed changes

- use display instead of visibility
- add relative position to the parent element (body), because of the child `Portal` element having the absolute position
- add z-index to be placed above the category elements
- refactor `window.scrollY`  to calculate and memoize the value and to make sure the side menu starts at the top of the screen

### Testing

- Open the side-menu on the small screen and check the category and its children to see if there is additional height after the toolbar footer and try to open the side-menu as well as inside category children

### Side effects

-should be none?

### Resolved issues

Fixes: #3020 

---

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -16,6 +16,8 @@ import { initSentry } from './utils/sentry'
 
 const GlobalStyle = createGlobalStyle`
   body {
+    position: relative;
+  
     /* Styling for react-tooltip: https://react-tooltip.com/docs/getting-started#styling */
     --rt-color-dark: ${props => props.theme.colors.textSecondaryColor};
     --rt-color-white: ${props => props.theme.colors.backgroundColor};
--- a/web/src/components/KebabMenu.tsx
+++ b/web/src/components/KebabMenu.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, ReactNode } from 'react'
+import React, { ReactElement, ReactNode, useLayoutEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
@@ -77,6 +77,13 @@ const StyledIcon = styled(Icon)`
 const KebabMenu = ({ items, show, setShow, Footer }: KebabMenuProps): ReactElement | null => {
   useLockedBody(show)
   const { t } = useTranslation('layout')
+  const [scrollY, setScrollY] = useState<number>(0)
+
+  useLayoutEffect(() => {
+    if (show) {
+      setScrollY(window.scrollY)
+    }
+  }, [show])
 
   const onClick = () => {
     setShow(!show)
@@ -95,8 +102,8 @@ const KebabMenu = ({ items, show, setShow, Footer }: KebabMenuProps): ReactEleme
         className='kebab-menu'
         show={show}
         style={{
-          visibility: show ? 'visible' : 'hidden',
-          top: window.scrollY > 0 ? `${window.scrollY}px` : undefined,
+          display: show ? 'block' : 'none',
+          top: scrollY > 0 ? `${scrollY}px` : undefined,
         }}>
         {/* disabled because this is an overlay for backdrop close */}
         {/* eslint-disable-next-line styled-components-a11y/no-static-element-interactions,styled-components-a11y/click-events-have-key-events */}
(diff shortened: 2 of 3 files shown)


steffenkleinle (commented):
@lunars97 are all of these changes necessary to fix the issue? Can you elaborate which change fixes which problem?

web/src/App.tsx
```diff
@@ -17,6 +17,8 @@ import { initSentry } from './utils/sentry'
 const GlobalStyle = createGlobalStyle`
   body {
     /* Styling for react-tooltip: https://react-tooltip.com/docs/getting-started#styling */
+    position: relative;
```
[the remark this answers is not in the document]
lunars97: oh yeah, you are right, it should be on the line 19 because the side menu's position is absolute and the parent element which is body should have relative position, there are cases when then side-menu breaks and appears in the middle under the category...this is what I noticed during the investigation

lunars97:
> @lunars97 are all of these changes necessary to fix the issue? Can you elaborate which change fixes which problem?

yes, these changes should fix the issue

- using display instead of visibility fixes the unexpected scroll behaviour under the footer, because `visibility:hidden` when the side-menu is hidden visually, it stays as a part of DOM which might occupy the space after using the side-menu and lead to unexpected extra space after hiding the side-menu.
- adding relative position to the parent element (body), because of the child Portal element having the absolute position which can fix the side-menu appearing in the middle of the view
- adding z-index to be placed above the category elements makes sure that it will appear above the category view not under
- by refactoring window.scrollY to calculate and memoize the value and to make sure the side menu starts at the top of the screen also helps with the problem of side-menu appearing in the middle of the screen

web/src/App.tsx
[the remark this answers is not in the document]
steffenkleinle: Thanks for explaining, noticed the same :+1:

web/src/App.tsx
```diff
@@ -16,6 +16,8 @@ import { initSentry } from './utils/sentry'
 
 const GlobalStyle = createGlobalStyle`
   body {
+  position: relative;
```
steffenkleinle, line 19: I think the indentation is off:
```suggestion
    position: relative;
```

steffenkleinle (approved):
Great fix, thank you!

web/src/components/KebabMenu.tsx
```diff
@@ -95,8 +102,8 @@ const KebabMenu = ({ items, show, setShow, Footer }: KebabMenuProps): ReactEleme
         className='kebab-menu'
         show={show}
         style={{
-          visibility: show ? 'visible' : 'hidden',
-          top: window.scrollY > 0 ? `${window.scrollY}px` : undefined,
+          display: show ? 'block' : 'none',
+          top: scrollY > 0 ? `${scrollY}px` : undefined,
```
bahaaTuffaha, line 106: :thought_balloon:  I tried to remove this top completely and it didn't affect anything at this page not sure if we can remove it...

bahaaTuffaha (approved):
Nice solution 🤩 from now on I will keep an eye on any visibility propriety that could cause issues like this ... that was unexpected.

web/src/components/KebabMenu.tsx
lunars97: fair question,I think we do not need it here completely

steffenkleinle: @f1sh1918 do you still remember why we needed this in the first place? To me it also looks like it works just fine without setting anything as `top`.

f1sh1918: this is required!
This ensures that the content of the kebab menu will be opened on top position even if you scroll down the page and then open the kebab menu

image

steffenkleinle: Oh, very true... I thought I tested that, but seems like I didn't. Well, then please keep this @lunars97 :) I think we are good to merge this as it is then!

bahaaTuffaha: 🙈

PR #3053, opened 2025-01-27
2356: Show nearby cities for web

bahaaTuffaha:
### Short description

Showing nearby cities for web.

### Proposed changes

- Created a small component just to show nearby cities similar to `groups`.
- Used the `getUserLocation()` the same way it's been used at `PoisPage.tsx`.
- Added some mocks.

### Side effects

- `CitySelector.tsx`

### Testing

- Change location or http://localhost:9000/landing/en
- Enable location and refresh the page.
- If you are not located in Germany you can override you location from F12 -> 3 dots menu -> more tools -> sensors -> change it to Berlin.

Fixes: #2356 

---

### DIFF
--- a/web/src/components/CitySelector.tsx
+++ b/web/src/components/CitySelector.tsx
@@ -1,11 +1,12 @@
-import React, { ReactElement, useState } from 'react'
+import React, { ReactElement, useCallback, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
-import { filterSortCities } from 'shared'
-import { CityModel } from 'shared/api'
+import { filterSortCities, getNearbyCities, LocationType } from 'shared'
+import { CityModel, useLoadAsync } from 'shared/api'
 
 import buildConfig from '../constants/buildConfig'
+import getUserLocation from '../utils/getUserLocation'
 import CityEntry from './CityEntry'
 import CrashTestingIcon from './CrashTestingIcon'
 import Failure from './Failure'
@@ -25,9 +26,36 @@ const CityListParent = styled.div<{ $stickyTop: number }>`
   background-color: ${props => props.theme.colors.backgroundColor};
   border-bottom: 1px solid ${props => props.theme.colors.themeColor};
 `
+
 const SearchCounter = styled.p`
   color: ${props => props.theme.colors.textSecondaryColor};
 `
+
+type NearbyCitiesProps = {
+  userLocation: LocationType | null
+  cities: CityModel[]
+  language: string
+  filterText: string
+}
+
+const NearbyCities = ({ userLocation, cities, language, filterText }: NearbyCitiesProps) => {
+  const { t } = useTranslation('landing')
+  const nearCities = userLocation
+    ? getNearbyCities(
+        userLocation,
+        cities.filter(city => city.live),
+      )
+    : []
+
+  return nearCities.length > 0 ? (
+    <div key='nearbyCities'>
+      <CityListParent $stickyTop={0}>{t('nearbyCities')}</CityListParent>
+      {nearCities.map(city => (
+        <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+      ))}
+    </div>
+  ) : null
+}
 
 type CitySelectorProps = {
   cities: CityModel[]
@@ -38,6 +66,13 @@ const CitySelector = ({ cities, language }: CitySelectorProps): ReactElement => 
   const [filterText, setFilterText] = useState<string>('')
   const [stickyTop, setStickyTop] = useState<number>(0)
   const { t } = useTranslation('landing')
+
+  const { data: userLocation } = useLoadAsync(
+    useCallback(async () => {
+      const userLocation = await getUserLocation()
+      return userLocation.status === 'ready' ? userLocation.coordinates : null
+    }, []),
+  )
 
   const resultCities = filterSortCities(cities, filterText, buildConfig().featureFlags.developerFriendly)
 
@@ -62,6 +97,7 @@ const CitySelector = ({ cities, language }: CitySelectorProps): ReactElement => 
         spaceSearch={false}
         onStickyTopChanged={setStickyTop}>
         <SearchCounter>{t('search:searchResultsCount', { count: resultCities.length })}</SearchCounter>
+        <NearbyCities userLocation={userLocation} cities={cities} language={language} filterText={filterText} />
         {resultCities.length === 0 ? <Failure errorMessage='search:nothingFound' /> : groups}
       </ScrollingSearchBox>
     </Container>
--- a/web/src/__tests__/RootSwitcher.spec.tsx
+++ b/web/src/__tests__/RootSwitcher.spec.tsx
@@ -16,6 +16,7 @@ import { renderWithRouterAndTheme } from '../testing/render'
 jest.mock('shared/api', () => ({
   ...jest.requireActual('shared/api'),
   useLoadFromEndpoint: jest.fn(),
+  useLoadAsync: jest.fn(() => ({ data: null, error: null })),
 }))
 jest.mock('../CityContentSwitcher')
 
--- a/web/src/components/__tests__/CitySelector.spec.tsx
+++ b/web/src/components/__tests__/CitySelector.spec.tsx
@@ -9,6 +9,10 @@ import CitySelector from '../CitySelector'
 
 jest.mock('react-inlinesvg')
 jest.mock('react-i18next')
+jest.mock('shared/api', () => ({
+  ...jest.requireActual('shared/api'),
+  useLoadAsync: jest.fn(() => ({ data: null, error: null })),
+}))
 
 describe('CitySelector', () => {
   const previousConfig = buildConfig()


web/src/components/CitySelector.tsx
```diff
@@ -25,10 +26,53 @@ const CityListParent = styled.div<{ $stickyTop: number }>`
...
+      {nearbyCities.map(city => (
+        <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+      ))}
+    </div>
+  ) : (
+    <NearbyMessageContainer>
+      <NearbyMessage>{userLocation ? t('noNearbyCities') : t('locationError')}</NearbyMessage>
+    </NearbyMessageContainer>
+  )
+}
```
steffenkleinle, line 74: :wrench: Please move this to an own component

f1sh1918 (approved):
Nice work! Looks good to me 👍 
Tested on chrome, safari and firefox

web/src/components/CitySelector.tsx
```diff
@@ -39,6 +42,13 @@ const CitySelector = ({ cities, language }: CitySelectorProps): ReactElement =>
   const [stickyTop, setStickyTop] = useState<number>(0)
   const { t } = useTranslation('landing')
 
+  const { data: userLocation } = useLoadAsync(
+    useCallback(async () => {
+      const userLocation = await getUserLocation()
+      return userLocation.status === 'ready' ? userLocation.coordinates : null
+    }, []),
+  )
```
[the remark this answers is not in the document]
steffenkleinle: Furthermore, would you mind creating a shared hook for this that can be reused in `getUserLocation` called `useUserLocation`? That would be cool :)

web/src/components/NearbyCities.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { t } = useTranslation('landing')
+  const nearbyCities = userLocation
+    ? getNearbyCities(
+        userLocation,
+        cities.filter(city => city.live),
+      )
+    : []
+
+  return nearbyCities.length > 0 ? (
+    <div key='nearbyCities'>
```
steffenkleinle, line 40: :upside_down_face: Why is the div/key needed here? Can't we just return a `CityListParent` straight?

web/src/components/NearbyCities.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <CityListParent $stickyTop={0}>{t('nearbyCities')}</CityListParent>
+      {nearbyCities.map(city => (
+        <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+      ))}
+    </div>
+  ) : (
+    <NearbyMessageContainer>
+      <NearbyMessage>{userLocation ? t('noNearbyCities') : t('locationError')}</NearbyMessage>
```
steffenkleinle, line 48: :wrench: I think the message should also be wrapped in the `CityListParent` such that the context is clear. Also, it looks better imo. Not sure if we should not also display more specific error messages and need a retry button like in native. It might happen that location is allowed, but not found.

web/src/components/CitySelector.tsx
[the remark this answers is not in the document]
bahaaTuffaha: Reused in getUserLocation? 
I'm already using getUserLocation is this block.. do you mean at PoisPage.tsx?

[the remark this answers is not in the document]
steffenkleinle: This code is a duplicate of the code at `PoisPage.tsx` and might be needed in other places in the future. So it would be cool to extract this code to `getUserLocation` and export it from there and use that both here and in `PoisPage.tsx` like this:
``` ts
  const { data: userLocation } = useUserLocation()
```

Probably the file `getUserLocation` should then be renamed to `useUserLocation`.

web/src/components/NearbyCities.tsx
bahaaTuffaha: There is a `nearbyCities.length > 0` check and if it's not null it means enabled.. it should show noNearbyCities.
 I will add a refresh button

steffenkleinle: Please have a look how it is implemented on native. There is two things to my comment:
- The text and separator `Nearby Cities` should still be displayed, even if we show an error message and not actual nearby cities
- The refresh button should be shown to try to load the position if it was not successful initially

web/src/components/NearbyCities.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div>
+          {nearbyCities.map(city => (
+            <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+          ))}
+        </div>
+      ) : (
+        <NearbyMessageContainer>
+          <StyledMessageWrapper>
+            <NearbyMessage>{userLocation?.coordinates ? t('noNearbyCities') : t('locationError')}</NearbyMessage>
+            <RetryButtonContainer label={t('refresh')} onClick={() => refresh()}>
```
steffenkleinle, line 69: :upside_down_face: 
```suggestion
            <RetryButtonContainer label={t('refresh')} onClick={refresh}>
```

web/src/components/NearbyCities.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <>
+      <CityListParent $stickyTop={stickyTop}>{t('nearbyCities')}</CityListParent>
+      {nearbyCities.length > 0 ? (
+        <div>
+          {nearbyCities.map(city => (
+            <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+          ))}
+        </div>
```
steffenkleinle, line 64: :wrench: The div should not be needed
```suggestion
        <>
          {nearbyCities.map(city => (
            <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
          ))}
        </>
```

web/src/hooks/__tests__/useUserLocation.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return <div>{data?.coordinates?.join(',') ?? 'null'}</div>
+}
+
+describe('useUserLocation', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should show coordinates when available', async () => {
+    ;(useUserLocation as jest.Mock).mockReturnValue({
```
steffenkleinle, line 23: :wrench: Please change this to the following, same for other occurrences:
```suggestion
    mocked(useUserLocation).mockReturnValue({
```

web/src/hooks/__tests__/useUserLocation.spec.tsx
```diff
@@ -0,0 +1,51 @@
+import { render, waitFor } from '@testing-library/react'
+import React from 'react'
+
+import useUserLocation from '../useUserLocation'
+
+jest.mock('../useUserLocation', () => ({
+  __esModule: true,
+  default: jest.fn(),
+  getUserLocation: jest.fn(),
+}))
```
steffenkleinle, line 10: :wrench: Hmm... you are basically testing that the mock returns exactly what you mock it to return. The only thing that should be mocked in this test file is `navigator.geolocation.getCurrentPosition`. Can you try to adjust your tests to actually test something? If you need help, let me know :)

web/src/hooks/useUserLocation.ts
```diff
@@ -40,4 +43,19 @@ const getUserLocation = async (): Promise<UserLocationType> =>
...
+
+  const refresh = useCallback(() => {
+    setRefreshState(prev => !prev)
+  }, [])
+
+  return {
+    data,
+    refresh,
+  }
+}
```
steffenkleinle, line 59: :wrench: This really complicates things, the additional state and everything should not be needed. Please just copy the hook as it was before in PoisPage and CitySelector and return everything and not just the data. That includes a `refresh` function already.

web/src/components/NearbyCities.tsx
bahaaTuffaha: Bad habits 😅

web/src/hooks/useUserLocation.ts
```diff
@@ -43,18 +43,21 @@ export const getUserLocation = async (): Promise<UserLocationType> =>
     )
   })
 
-const useUserLocation = (): { data: UserLocationType | null; refresh: () => void } => {
-  const [refreshState, setRefreshState] = useState(false)
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const { data } = useLoadAsync(useCallback(async () => getUserLocation(), [refreshState]))
+type useUserLocationReturn = {
```
steffenkleinle, line 46: :wrench: Should be uppercase
```suggestion
type UseUserLocationReturn = {
```

web/src/hooks/useUserLocation.ts
```diff
@@ -43,18 +43,21 @@ export const getUserLocation = async (): Promise<UserLocationType> =>
...
+  data: UserLocationType | null
+  refresh: () => void
+  error: Error | null
+  loading: boolean
+}
 
-  const refresh = useCallback(() => {
-    setRefreshState(prev => !prev)
-  }, [])
+const useUserLocation = (): useUserLocationReturn => {
```
steffenkleinle, line 53: :wrench: You can just do
```suggestion
const useUserLocation = (): Return<UserLocationType>
```

web/src/hooks/useUserLocation.ts
```diff
@@ -43,18 +43,21 @@ export const getUserLocation = async (): Promise<UserLocationType> =>
...
+  const { data, refresh, error, loading } = useLoadAsync(useCallback(async () => getUserLocation(), []))
```
steffenkleinle, line 54: :upside_down_face: Why not just
```suggestion
  return useLoadAsync(useCallback(getUserLocation, []))
```

steffenkleinle: The callback isn't even needed as `getUserLocation` is not defined inside the hook. It would only be needed if you did the following as I thought would make sense (but perhaps this is better):
``` ts
return useLoadAsync(
    useCallback(async () => {
      const userLocation = await getUserLocation()
      return userLocation.status === 'ready' ? userLocation.coordinates : null
    }, []),
  )
```

web/src/components/NearbyCities.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const nearbyCities = userLocation
+    ? getNearbyCities(
+        userLocation,
+        cities.filter(city => city.live),
+      )
+    : []
+
+  return (
+    <>
```
steffenkleinle, line 57: :wrench: I am sorry, but seems like we should readd the `div` here. Otherwise we have this weird visual effect:
```suggestion
    <div>
```
image

steffenkleinle (approved):
Nice work!

LeandraH:
@bahaaTuffaha Can this be merged?

Issue #3054, opened 2025-01-28
Display Issue in RTL Languages for Event Calendar


nassabay:
### Describe the Bug

In all languages that are read from right to left, such as Sorani, Farsi, and Arabic, there seems to be a display issue in the event calendar. For some events, the image appears on the left instead of the right (see screenshot).

You can find a few examples here: https://integreat.app/muenchen/ar/events
See attached image as well. 

![Image](https://github.com/user-attachments/assets/48df9c0c-6748-4ce4-9752-778ec7f1df1c)

PR #3055, opened 2025-01-28
3054: Fix rtl display events

f1sh1918:
### Short description

It looks like due to not needed direction tags, the rtl events are not displayed correctly,

### Proposed changes

- removed not needed direction tags for title row, since they will be inherited from `ListItemContainer`

### Side effects

- may affect other components that use ListItems (Licenses, Pois, Breadcrumbs)

### Testing

Open: http://localhost:9000/testumgebung/ar/events

### Resolved issues

Fixes: #3054

---

### DIFF
--- a/web/src/components/ListItem.tsx
+++ b/web/src/components/ListItem.tsx
@@ -58,7 +58,7 @@ const ListItem = ({ path, title, thumbnail, thumbnailSize, children, Icon }: Lis
       {!!thumbnail && <Thumbnail alt='' src={thumbnail} $thumbnailSize={thumbnailSize} />}
       <Description>
         <TitleRow>
-          <Title dir='auto'>{title}</Title>
+          <Title>{title}</Title>
           {Icon}
         </TitleRow>
         {children}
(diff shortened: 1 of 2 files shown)


web/src/components/ListItem.tsx
```diff
@@ -58,7 +58,7 @@ const ListItem = ({ path, title, thumbnail, thumbnailSize, children, Icon }: Lis
       {!!thumbnail && <Thumbnail alt='' src={thumbnail} $thumbnailSize={thumbnailSize} />}
       <Description>
         <TitleRow>
-          <Title dir='auto'>{title}</Title>
+          <Title>{title}</Title>
```
steffenkleinle, line 61: :thought_balloon: Could you explain a little more why this breaks the the layout only for non-recurring events? This seems very odd to me...

steffenkleinle (approved):
Tested on firefox, fixed as expected.

lunars97 (approved):
Thank you for your quick fix :+1: LGTM

web/src/components/ListItem.tsx
f1sh1918: I can not explain why I guess due to duplication maybe. At least setting this is not needed at this place

Issue #3029, opened 2025-01-07
Add "Typing" message for automatic replies


svenseeberg:
As long as we expect automatic replies, we should show a placeholder message with 3 dots (...). This is the default indicator that the other side is typing. This message should be shown for about 60s when the last message is a user message. After 60s the message should disappear.

This is not a perfect mechanism but should do the trick in most cases. It will provide a hint to the user that an answer can most likely be expected. We have no good indication (yet?) for the front end that the back end is still working on an answer.

In the long run we probably want to indicate that the back end is actually working on an answer. However, this is only in reach when we merge the chat back end into the Integreat CMS and start implementing websockets.

### Additional information

Tracking issue in integreat-chat: https://github.com/digitalfabrik/integreat-chat/issues/120

svenseeberg:
There is one caveat I "discovered" yesterday: the first response should be ignored as it is returned every time and is returned "quickly". This message usually just says that our bot is going to try to answer.

steffenkleinle:
> There is one caveat I "discovered" yesterday: the first response should be ignored as it is returned every time and is returned "quickly". This message usually just says that our bot is going to try to answer.

@svenseeberg can you clarify a little more why we should not display the typing indicator in that case? Our current polling interval is 16s, so the user might have to wait 16s until he gets the automatic message. I think that is not really quick.
Furthermore, I think it is even fine if the typing indicator is only displayed one second until the first message is received, isn't it?

Or do you mean the typing indicator should still be shown after the first automatic response because we are still going to receive the chatbot message?

svenseeberg:
> Or do you mean the typing indicator should still be shown after the first automatic response because we are still going to receive the chatbot message?

Yes the first message, that comes back within seconds, should be ignored. This is always the same boiler plate that is not really relevant.

> @svenseeberg can you clarify a little more why we should not display the typing indicator in that case?

The real automatic response takes about 60s. However, the user has no idea how long it will take. Indicating that a message might come back is IMHO useful to keep the user waiting for that amount of time. If no message arrives within 60 or 90s, usually the back end is broken and no automatic answer will arrive (soon).

steffenkleinle:
We discussed the topic in person and the result was the following:
[...]

PR #3056, opened 2025-01-29
3029: Add typing indicator for automatic replies in chat

lunars97:
### Short description

Typing indicator message (...) for 60 seconds should be shown after automatic answer in chat. It will be a temporary solution untill the websocket implementation

### Proposed changes

- Add typing indicator function
- Style typing indicator in a similar manner

### Side effects

- none

### Resolved issues

Fixes: #3029 

---

### DIFF
--- a/web/src/components/ChatConversation.tsx
+++ b/web/src/components/ChatConversation.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 
 import ChatMessageModel from 'shared/api/models/ChatMessageModel'
 
-import ChatMessage from './ChatMessage'
+import ChatMessage, { Message } from './ChatMessage'
 
 const Container = styled.div`
   font-size: ${props => props.theme.fonts.hintFontSize};
@@ -13,6 +13,10 @@ const Container = styled.div`
 
 const InitialMessage = styled.div`
   margin-bottom: 12px;
+`
+
+const TypingIndicatorWrapper = styled(Message)`
+  width: max-content;
 `
 
 const ErrorSendingStatus = styled.div`
@@ -29,10 +33,27 @@ type ChatConversationProps = {
   className?: string
 }
 
+type TypingIndicatorProps = {
+  isVisible: boolean
+}
+
+const TypingIndicator = ({ isVisible }: TypingIndicatorProps): ReactElement | null =>
+  isVisible ? (
+    <TypingIndicatorWrapper>
+      <strong>...</strong>
+    </TypingIndicatorWrapper>
+  ) : null
+
+const TYPING_INDICATOR_TIMEOUT = 60000
+
 const ChatConversation = ({ messages, hasError, className }: ChatConversationProps): ReactElement => {
   const { t } = useTranslation('chat')
   const [messagesCount, setMessagesCount] = useState(0)
+  const [typingIndicatorVisible, setTypingIndicatorVisible] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
+  const isLastMessageFromUser = messages[messages.length - 1]?.userIsAuthor
+  const hasOnlyReceivedInfoMessage = messages.filter(message => !message.userIsAuthor).length === 1
+  const waitingForAnswer = isLastMessageFromUser || hasOnlyReceivedInfoMessage
 
   useEffect(() => {
     if (messagesCount < messages.length) {
@@ -40,6 +61,19 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
       setMessagesCount(messages.length)
     }
   }, [messages, messagesCount])
+
+  useEffect(() => {
+    if (waitingForAnswer) {
+      setTypingIndicatorVisible(true)
+
+      const typingIndicatorTimeout = setTimeout(() => {
+        setTypingIndicatorVisible(false)
+      }, TYPING_INDICATOR_TIMEOUT)
+
+      return () => clearTimeout(typingIndicatorTimeout)
+    }
+    return () => undefined
+  }, [waitingForAnswer])
 
   return (
     <Container className={className}>
@@ -53,6 +87,7 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
               showIcon={messages[index - 1]?.userIsAuthor !== message.userIsAuthor}
             />
           ))}
+          <TypingIndicator isVisible={typingIndicatorVisible} />
           <div ref={messagesEndRef} />
         </>
       ) : (
--- a/web/src/components/__tests__/ChatConversation.spec.tsx
+++ b/web/src/components/__tests__/ChatConversation.spec.tsx
@@ -1,3 +1,4 @@
+import { act } from '@testing-library/react'
 import React from 'react'
 
 import ChatMessageModel from 'shared/api/models/ChatMessageModel'
@@ -7,6 +8,7 @@ import ChatConversation from '../ChatConversation'
 
 jest.mock('react-i18next')
 window.HTMLElement.prototype.scrollIntoView = jest.fn()
+jest.useFakeTimers()
 
 const render = (messages: ChatMessageModel[], hasError: boolean) =>
   renderWithRouterAndTheme(<ChatConversation messages={messages} hasError={hasError} />)
@@ -21,8 +23,20 @@ describe('ChatConversation', () => {
     }),
     new ChatMessageModel({
       id: 2,
+      body: 'Willkommen in der Integreat Chat Testumgebung auf Deutsch. Unser Team antwortet werktags, während unser Chatbot zusammenfassende Antworten aus verlinkten Artikeln liefert, die Sie zur Überprüfung wichtiger Informationen lesen sollten.',
+      userIsAuthor: false,
+      automaticAnswer: true,
+    }),
+    new ChatMessageModel({
+      id: 3,
       body: 'Informationen zu Ihrer Frage finden Sie auf folgenden Seiten:',
       userIsAuthor: false,
+      automaticAnswer: false,
+    }),
+    new ChatMessageModel({
+      id: 4,
+      body: 'Wie kann ich mein Deutsch verbessern?',
+      userIsAuthor: true,
       automaticAnswer: false,
     }),
   ]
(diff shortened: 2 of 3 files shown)


lunars97:
The typing indicator also appearing before the automatic reply, I have no idea how to limit this behaviour... Any ideas?

steffenkleinle (changes requested):
Looks good already, I just  think the condition when to display the typing indicator is still wrong. I only get it if the last message is not from the user.

steffenkleinle:
Also see my question here: https://github.com/digitalfabrik/integreat-app/issues/3029#issuecomment-2621290318

web/src/components/ChatConversation.tsx
```diff
@@ -41,6 +54,19 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
     }
   }, [messages, messagesCount])
 
+  // eslint-disable-next-line consistent-return
+  useEffect(() => {
+    if (!lastMessage?.userIsAuthor && lastMessage?.isAutomaticAnswer) {
```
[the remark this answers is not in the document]
lunars97: I put it there to limit somehow the automatic reply, because everytime I send something, the automatic answer comes, should I activate something in order to receive an answer from bot?

web/src/components/ChatConversation.tsx
```diff
@@ -15,6 +15,15 @@ const InitialMessage = styled.div`
...
 `
 
+const TypingIndicator = styled.div`
+  text-align: center;
+  border-radius: 5px;
+  padding: 8px;
+  width: 20px;
+  border: 1px solid ${props => props.theme.colors.textDecorationColor};
+  font-size: ${props => props.theme.fonts.contentFontSize};
+`
```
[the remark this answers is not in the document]
lunars97: good idea :)

web/src/components/ChatConversation.tsx
```diff
@@ -29,10 +33,15 @@ type ChatConversationProps = {
...
 
+const TYPING_INDICATOR_TIMEOUT = 60000
+
 const ChatConversation = ({ messages, hasError, className }: ChatConversationProps): ReactElement => {
   const { t } = useTranslation('chat')
   const [messagesCount, setMessagesCount] = useState(0)
+  const [showTypingIndicator, setShowTypingIndicator] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
+  const lastUserMessage = messages[messages.length - 1]
+  const beforelastUserMessage = messages[messages.length - 2]
```
steffenkleinle, line 44: :wrench: The naming is misleading here, those are not necessarily user messages.
```suggestion
  const lastMessage = messages[messages.length - 1]
  const secondLastMessage = messages[messages.length - 2]
```
Perhaps even better would be the following to have a speaking variable:
```suggestion
  const showTypingIndicator = messages[messages.length - 1]?.userIsAuthor || messages[messages.length - 2]?.userIsAuthor
```

steffenkleinle (approved):

web/src/components/ChatConversation.tsx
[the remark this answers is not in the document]
steffenkleinle: I misunderstood Svens message, sorry. You implemented it perfectly! See here for clarification: https://github.com/digitalfabrik/integreat-app/issues/3029#issuecomment-2626466355
Since you exactly implemented it like this, nothing to do here anymore :)

web/src/components/ChatConversation.tsx
```diff
@@ -29,10 +33,15 @@ type ChatConversationProps = {
...
 
+const TYPING_INDICATOR_TIMEOUT = 60000
+
 const ChatConversation = ({ messages, hasError, className }: ChatConversationProps): ReactElement => {
   const { t } = useTranslation('chat')
   const [messagesCount, setMessagesCount] = useState(0)
+  const [showTypingIndicator, setShowTypingIndicator] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
+  const showIndicatorOfTyping =
+    messages[messages.length - 1]?.userIsAuthor || messages[messages.length - 2]?.userIsAuthor
```
bahaaTuffaha, line 44: This means that if the message before the last one is a user it will show the 3 dot.
Shouldn't be shown just if the last message is coming from a user?

steffenkleinle: Fair point and I think you are right. Seems like I was a little tired already when I made my review, sorry :D Probably we would need a condition in the following way:
Show the indicator if either
- the last message was sent by the user OR
- there is only one message yet that was NOT sent by the user (as that is the automatic answer)

web/src/components/ChatConversation.tsx
```diff
@@ -41,7 +41,7 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
   const [showTypingIndicator, setShowTypingIndicator] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
   const showIndicatorOfTyping =
-    messages[messages.length - 1]?.userIsAuthor || messages[messages.length - 2]?.userIsAuthor
+    messages[messages.length - 1]?.userIsAuthor || !messages[messages.length - 1]?.isAutomaticAnswer
```
steffenkleinle, line 44: :x: Hmm, I think sadly this is not quite correct yet. LLM answers are also automatic answers and might be the last answer a user gets, so displaying a typing indicator would be confusing in that case.
Instead, I think you have to go through all previous message and check if there is only one automatic answer yet only display the indicator in that case.

web/src/components/ChatConversation.tsx
```diff
@@ -41,7 +41,7 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
   const [showTypingIndicator, setShowTypingIndicator] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
   const showIndicatorOfTyping =
```
steffenkleinle, line 43: :upside_down_face: 
```suggestion
  const showTypingIndicator =
```

lunars97: we have the same naming for the state above :)

steffenkleinle: Oh sorry, I missed that. I think that we should think of a more explicit naming. How about something like `shouldShowTypingIndicator` and `typingIndicatorVisible`?

web/src/components/ChatConversation.tsx
lunars97: we will have then one condition to check if there was only one automaticAnswer and if there is one then we will show the indicator, am i right?

web/src/components/ChatConversation.tsx
```diff
@@ -29,10 +33,14 @@ type ChatConversationProps = {
...
 }
 
+const TYPING_INDICATOR_TIMEOUT = 60000
+
 const ChatConversation = ({ messages, hasError, className }: ChatConversationProps): ReactElement => {
   const { t } = useTranslation('chat')
   const [messagesCount, setMessagesCount] = useState(0)
+  const [showTypingIndicator, setShowTypingIndicator] = useState(false)
   const messagesEndRef = useRef<HTMLDivElement>(null)
+  const typingIndicatorVisible = messages.filter(msg => msg.isAutomaticAnswer).length === 1
```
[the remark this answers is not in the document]
lunars97: sorry, I misunderstood you above. it is looks very nice, thank you for your clean and comprehensive solution :100:

steffenkleinle (commented):
Nice work, tested on firefox and works as expected! I think it would be cool to add some tests for this, could you try to do add a few test cases for this?

web/src/components/__tests__/ChatConversation.spec.tsx
```diff
@@ -37,7 +51,29 @@ describe('ChatConversation', () => {
...
+  it('should display typing indicator before the initial automatic answer and after for 60 seconds', () => {
+    const { getByText, queryByText, getByTestId } = render(testMessages, false)
+    expect(getByTestId(testMessages[0]!.id)).toBeTruthy()
+    expect(getByText('...')).toBeTruthy()
     expect(getByTestId(testMessages[1]!.id)).toBeTruthy()
+    expect(getByText('...')).toBeTruthy()
+
+    act(() => {
+      jest.advanceTimersByTime(TYPING_INDICATOR_TIMEOUT)
+    })
```
steffenkleinle, line 66: Tip: You can just use this to avoid exporting/importing the actual length of the timeout (same below):
```suggestion
    act(() => jest.runAllTimers())
```

web/src/components/ChatConversation.tsx
```diff
@@ -41,6 +51,28 @@ const ChatConversation = ({ messages, hasError, className }: ChatConversationPro
...
+    if (answer) {
+      return typingIndicatorTimer()
+    }
+    return () => undefined
+  }, [])
+
+  useEffect(() => {
+    const typingIndicator = handleWaitingForAnswer(waitingForAnswer)
+    return typingIndicator
+  }, [waitingForAnswer, handleWaitingForAnswer])
```
steffenkleinle, line 74: As discussed: please revert :) We'll just ignore the error for now :)

steffenkleinle (approved):
Nice work!

web/src/components/ChatConversation.tsx
lunars97: Well at least it is not 11 :smile: `ChatConversation has a cyclomatic complexity of 10, threshold = 10.`

steffenkleinle (approved):
Nice :rocket:

bahaaTuffaha (approved):
Nice work ✨
💭 I noticed that if I refreshed the page the `TypingIndicator` will show up again... we could add a flag at localStorage to store last message id just to check if we need to show the `TypingIndicator` again. Otherwise all good.
Note: (couldn't test in case of an actual replay)

lunars97:
> Nice work ✨ 💭 I noticed that if I refreshed the page the `TypingIndicator` will show up again... we could add a flag at localStorage to store last message id just to check if we need to show the `TypingIndicator` again. Otherwise all good. Note: (couldn't test in case of an actual replay)

I think this is how it should behave if the user has not received a reply yet, this way the person will know that it is being still processed?

PR #3057, opened 2025-01-29
Include ios and android for native and allow use of root project in IntelliJ

steffenkleinle:
### Short description

Include ios and android for native and allow use of root project in IntelliJ.
This allows searching these files as well.

### Proposed changes

- Remove exclusion of ios and android on native
- Remove exclusion of native and web in the root project
- Mark directories correctly

### Side effects

The docs saying "open web or native" separately is kinda outdated/not the only option anymore. However, I would first try to work with the root project only for a bit before changing that recommendation.

### Testing

Open the root or native project and play around. Let me know if you find this annoying.

### Resolved issues

Fixes: N/A

---

### DIFF
--- a/.idea/integreat-app.iml
+++ b/.idea/integreat-app.iml
@@ -5,14 +5,20 @@
     <exclude-output />
     <content url="file://$MODULE_DIR$">
       <sourceFolder url="file://$MODULE_DIR$/locales" isTestSource="false" />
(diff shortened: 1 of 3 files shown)


f1sh1918 (commented):
Works fine. I think you could even filter by filetype if it gets annoying

f1sh1918 (approved):

lunars97 (approved):
LGTM :+1:

Issue #2746, opened 2024-03-27
Remove react-test-renderer and rewrite tests


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
We still have two legacy tests using react-test-renderer in native.

**Describe the solution you'd like**
- Rewrite the two tests (`DashboardNavigationTiles.spec.tsx` and `SearchListItem.spec.tsx`) using react native testing library
- Remove `react-test-renderer` and `@types/react-test-renderer dependencies` (if possible, might be needed as peer dependency)

**Describe alternatives you've considered**
None.

**Additional context**
None.

LeandraH:
React-test-renderer might be a peer dep of react testing library, but we can still remove our own dependencies :)

PR #3062, opened 2025-01-29
2746: Remove react-test-renderer and rewrite it with react native tes…

lunars97:
…ting library

### Short description

Two legacy tests `DashboardNavigationTiles.spec.tsx` and `SearchListItem.spec.tsx` which are still using react-test-renderer should removed and be rewritten using react-native testing library. 

### Proposed changes

- Remove dependencies from package.json
- Rewrite tests

### Side effects

- I saw no peerDependencies, so should be none? :see_no_evil: 

### Testing

run test in native

### Resolved issues

Fixes: #2746

---


steffenkleinle (changes requested):
Nice, looks quite good already! I think we should avoid the mock of `NavigationTiles` if possible though, could you perhaps try that please? Thanks :)

native/src/components/__tests__/DashboardNavigationTiles.spec.tsx
```diff
@@ -1,15 +1,21 @@
...
 
 jest.mock('react-i18next')
+jest.mock('react-native-safe-area-context', () => {
+  const actual = jest.requireActual('react-native-safe-area-context')
+  return {
+    ...actual,
+    useSafeAreaInsets: () => ({ top: 0, bottom: 0, left: 0, right: 0 }),
+    SafeAreaProvider: ({ children }: { children: React.ReactNode }) => children,
+  }
+})
```
lunars97, line 18: I needed to add it because of the error `Error: No safe area value available. Make sure you are rendering <SafeAreaProvider> at the top of your app.`

steffenkleinle: Mocking `useSafeAreaInsets` should be enough as that is the only thing used by `NavigationTiles`, `SafeAreaProvider` is only used in `App.tsx`. Alternatively, you can also directly use the mock provided by `react-native-safe-area-context`, see e.g. `MalteHelpFormOffer.spec.tsx`.

steffenkleinle (approved):
Nice refactoring, thanks!

steffenkleinle:
Actually, I think we need `react-test-renderer`. The message about peer dependencies was not actually correct though. So you can just readd the dependency to the package.json and keep everything else as it is!
image

native/src/components/__tests__/DashboardNavigationTiles.spec.tsx
lunars97: @steffenkleinle thanks, good to know :)

LeandraH (approved):
Nicely done!

PR #3063, opened 2025-01-30
2917: Fix tts on ios proposal

steffenkleinle:
### Short description

This PR fixes tts on iOS and a few other bugs and problems with tts.
Furthermore, I added some tests and did some refactoring.

### Proposed changes

- Fix tts on iOS by removing the listener before stopping/pausing tts
- Allow reading content of 2nd level categories (see discussion [here](https://chat.tuerantuer.org/digitalfabrik/pl/53zutewbotnzjn7b5rkbwzpi3e))
- Fix tts sentences when navigating between pages
- Handle opening of player in `TtsContainer` instead of header for clear responsibilities
- Only initialize tts once actually needed (fixes annoying issues if no tts engine installed)
- Error handling (show users a message if initializing fails)
- Create a new useAppStateListener hook
- Improve tests for TtsContainer and useTtsPlayer
- Further improvements and refactoring

### Side effects

N/A, hopefully everything works now on both android and ios.

### Testing

Test the tts player thoroughly on both ios and android! Press all buttons multiple times (especially pause, play, previous, next) and navigate between different pages while the tts player is open.

### Resolved issues

Fixes: #2917

---


native/src/hooks/useTtsPlayer.ts
```diff
@@ -9,33 +9,22 @@ import { AppContext } from '../contexts/AppContextProvider'
...
-  const sentences = useMemo(() => {
-    if (model) {
-      const content = parseHTML(model.content)
-      return [model.title, ...segment(languageCode, content)]
-    }
-
-    return []
-  }, [model, languageCode])
+  const ttsContext = useContext(TtsContext)
+  const [previousSentences] = useState(ttsContext.sentences)
```
steffenkleinle, line 13: I did not find another way to ensure that the correct sentences are set after navigating back.

native/src/components/TtsContainer.tsx
```diff
@@ -1,31 +1,41 @@
...
 const MAX_TITLE_DISPLAY_CHARS = 20
+const TTS_UNSUPPORTED_LANGUAGES = ['fa', 'ka', 'kmr']
+const TTS_OPTIONS: Options = {
+  androidParams: {
+    KEY_PARAM_PAN: 0,
+    KEY_PARAM_VOLUME: 0.6,
+    KEY_PARAM_STREAM: 'STREAM_MUSIC',
+  },
+  iosVoiceId: '',
+  // This must not be 1 on iOS
```
steffenkleinle, line 23: @f1sh1918 do you have an issue link for this?

f1sh1918: No

native/ios/Integreat.xcodeproj/project.pbxproj
```diff
@@ -641,6 +641,14 @@
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-NativeModulesApple/React_NativeModulesApple.framework/Headers",
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-graphics/React_graphics.framework/Headers",
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-graphics/React_graphics.framework/Headers/react/renderer/graphics/platform/ios",
+					" ${PODS_CONFIGURATION_BUILD_DIR}/ReactCommon/ReactCommon.framework/Headers",
```
f1sh1918, line 644: why we need this changes?

native/src/components/AppStateListener.ts
```diff
@@ -22,7 +20,7 @@ const AppStateListener = (): ReactElement | null => {
     }
   }
 
-  useEffect(() => AppState.addEventListener('change', handleAppStateChange).remove, [])
+  useAppStateListener(handleAppStateChange)
```
f1sh1918, line 23: can you explain this change, please

f1sh1918 (approved):
Some minor comments. But looks good. 👍 
Glad that you found a better solution :)
Tested on ios real device

native/ios/Integreat.xcodeproj/project.pbxproj
steffenkleinle: They were autogenerated, I can also revert them

native/src/components/AppStateListener.ts
steffenkleinle: I created a custom hook to make it reusable/tidy up the TtsContainer. I just reused it here in this place as well.

LeandraH (approved):
Very cool! Tested on real iOS

native/ios/Integreat.xcodeproj/project.pbxproj
f1sh1918: ah okay i found them

bahaaTuffaha:
Sorry I was late to check on it out
Great work...
one thing 
You changed something at  `useTtsPlayer.ts` so when I change the language while its speaking the player will be disabled and nothing to read.

PR #3065, opened 2025-02-04
2926: Redesign search feedback native

LeandraH:
### Short description

This is the implementation of the redesign of the search feedback in native.

### Proposed changes

- When there are no search results, the text makes it clearer that we also searched in the fallback language, and a suggestion to try a different language.
- When there are no search results, there is a very clear CTA to open the feedback form if the user thinks there is information missing anyway.
- The design also calls for the number of search results to disappear if there aren't any but I decided against that so that a screen reader can still yell at someone to stop typing if they already have no results (#2352)

### Side effects

- The sad smiley in the search feedback disappears (though I'm not sure that qualifies as a side effect since that's the design)
- I used German as the hard-coded fallback language in the translations. Should we ever have the app with a different fallback language by going international, we'll have to fix that. I decided to kick that problem down the road because it's not that likely a scenario, and we would probably have to check through the entire app anyway in that case.

### Testing

You can switch the app to English and search for "alleinerziehend".

### Resolved issues

Partially fixes: #2926

### DIFF
--- a/native/src/components/FeedbackContainer.tsx
+++ b/native/src/components/FeedbackContainer.tsx
@@ -21,11 +21,11 @@ const Container = styled.View`
   gap: 8px;
 `
 
-const BoldText = styled(Text)`
+const Title = styled(Text)`
   font-weight: 600;
 `
 
-const TextBeforeButton = styled(BoldText)`
+const Hint = styled(Title)`
   margin-top: 8px;
   text-align: center;
 `
@@ -88,11 +88,9 @@ const FeedbackContainer = ({ query, language, routeType, cityCode, slug }: Feedb
     })
   }
 
-  const fallbackLanguage = config.sourceLanguage
-
-  return (
-    <Container>
-      {showFeedback ? (
+  if (showFeedback) {
+    return (
+      <Container>
         <Feedback
           comment={comment}
           contactMail={contactMail}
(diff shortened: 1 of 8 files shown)


bahaaTuffaha (approved):
LGTM.. Nice work 👌

native/src/components/FeedbackContainer.tsx
```diff
@@ -1,17 +1,33 @@
...
+  gap: 8px;
+`
+
+const BoldText = styled(Text)`
+  font-weight: 600;
+`
+
+const TextBeforeButton = styled(BoldText)`
+  margin-top: 8px;
+  text-align: center;
```
steffenkleinle, line 30: :wrench: See comments for web, I think we should avoid naming things based on the style/position they use as they can easily change. Instead, they should focus more on functionality/usage structure imo.

native/src/components/FeedbackContainer.tsx
```diff
@@ -70,20 +88,33 @@ const FeedbackContainer = ({ query, language, routeType, cityCode, slug }: Feedb
...
+          <BoldText>
+            {language === fallbackLanguage ? t('noResultsInOneLanguage') : t('noResultsInTwoLanguages')}
+          </BoldText>
+          <Text>{t('checkQuery', { appName: buildConfig().appName })}</Text>
+          <TextBeforeButton>{t('informationMissing')}</TextBeforeButton>
+          <TextButton text={t('giveFeedback')} onPress={() => setShowFeedback(true)} />
+        </>
+      )}
     </Container>
   )
```
steffenkleinle, line 119: :upside_down_face: Very personal opinion, so feel free to ignore. I think if there is only this little things to reuse, its a lot easier to read to add a separate if-clause instead of an JSX inline if:
``` tsx
if (showFeedback) {
  return <Container><Feedback.../></Container>
}
return <Container><Hint.../></Container>
```

steffenkleinle (approved):
Tested on android, looks great!

steffenkleinle (approved):

Issue #3060, opened 2025-01-29
Fix placement of kebab side menu close icon


steffenkleinle:
### Describe the Bug

As a user I'd expect the close icon to be in the same place the three dots icon of the kebab side menu is.
However, it is actually slightly off.

### Steps to Reproduce

0. Use a mobile device
1. Go to https://webnext.integreat.app/testumgebung/en
2. Click on the three dots menu in the top right
3. See that the close icon is in a different position

### Additional Information

https://github.com/user-attachments/assets/1bd12202-9991-4d46-9443-f066370c03f5

hannaseithe:
Is that only concerning the horizontal positioning or also the vertical? Because if there should also be vertical alignment we would have to change the height (70px which is the dimensions.headerHeightSmall) of the header in the KebabMenu to 86px

steffenkleinle:
> Is that only concerning the horizontal positioning or also the vertical? Because if there should also be vertical alignment we would have to change the height (70px which is the dimensions.headerHeightSmall) of the header in the KebabMenu to 86px

Good question. Personally I think that it looks better if both the header and the kebab menu have the same height, so I would say both, especially since we definitely do not have any problems with too little space there. What is your opinion on this?

hauf-toni:
Both elements should have the same width and height. The main goal is to ensure that the close icon is positioned exactly where the three-dots menu was, so users can tap the same spot on the screen without having to adjust their finger placement. Thanks for asking!

PR #3072, opened 2025-02-05
3060: KebabMenu - Aligned CloseButton based on Position of Expand Button

hannaseithe:
### Short description

I aligned the close button with the expand button and changed the height of the header, so that now the KebapMenu Header has the same height as the main header

### Proposed changes

- I have added the container ActionBar container (analoguos to the Header structure in Header.tsx) 
-  changed height to min-height,so that now the header of the KebapMenu has the same height as the main header
-  adjusted the paddings

### Side effects

- none that I am aware of

### Testing

- Is the close button correctly positioned under all circumstances?

### Resolved issues

Fixes: #3060

### DIFF
--- a/web/src/components/KebabMenu.tsx
+++ b/web/src/components/KebabMenu.tsx
@@ -61,8 +61,21 @@ const Heading = styled.div`
   justify-content: ${props => (props.theme.contentDirection === 'rtl' ? `flex-start` : `flex-end`)};
   background-color: ${props => props.theme.colors.backgroundAccentColor};
   box-shadow: -3px 3px 3px 0 rgb(0 0 0 / 13%);
-  height: ${dimensions.headerHeightSmall}px;
-  padding: 0 8px;
+  min-height: ${dimensions.headerHeightSmall}px;
+  padding: 8px 8px;
+`
+
+const ActionBar = styled.nav`
+  order: 3;
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  gap: 12px;
+  padding: 0 16px;
+
+  @media ${dimensions.smallViewport} {
+    order: 2;
+  }
 `
 
 const Content = styled.div`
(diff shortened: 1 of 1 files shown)


web/src/components/KebabMenu.tsx
```diff
@@ -61,8 +61,21 @@ const Heading = styled.div`
...
+  order: 3;
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  gap: 12px;
+  padding: 0 16px;
+
+  @media ${dimensions.smallViewport} {
+    order: 2;
+  }
```
LeandraH, line 78: 🔧 I understand that you copied the entire content from the `Header` but I think we should prune it for things we actually need right now. For example, the `ActionBar` here is the only child of its parent `Heading`, so there is no need for `order`. Maybe remove each line one at a time and see which ones you still need?

LeandraH (commented):
Amazing, thank you for your first PR with us! It looks pretty good already, except I think you positioned the closing icon on a page without a scroll bar. When I open it on a page with scrollbar, the closing icon is still a little bit off from where the kebob icon was.

https://github.com/user-attachments/assets/802cf07b-dbbc-460d-984f-827eac3064b3

web/src/components/KebabMenu.tsx
hannaseithe: Yes, you are absolutely right. That was a bit hasty of me, I will do so

hannaseithe: Regarding the scroll bar issue: I could replicate it on the edge browser only. Because elsewhere ( chrome and firefox on my windows machine and even on my android mobile) the scroll bar is an overlay. So I am not entirely sure if we should aim to fix that. Maybe more people should test it on their actual mobile devices and see if it shows up anywhere.

steffenkleinle (approved):
Congrats and thanks a lot for your first contribution :tada: :heart_eyes: Nice work, works and looks as expected! Tested on firefox :)

:thought_balloon: Unrelated to your PR and does not need to be done here: I think the hit targets are quite small, perhaps we should improve add some padding around the icons to increase the button size a bit?

steffenkleinle:
> Amazing, thank you for your first PR with us! It looks pretty good already, except I think you positioned the closing icon on a page without a scroll bar. When I open it on a page with scrollbar, the closing icon is still a little bit off from where the kebob icon was.
> Screen.Recording.2025-02-05.at.15.48.51.mov

I am also not able to reproduce this neither on firefox desktop nor mobile (or if so, only negligible).

LeandraH (approved):
Then I think we're good to go :)

LeandraH:
You can now click "Update branch" in this PR, and then enable auto-merge 🎉  Or alternatively, you can click "Update branch", wait for that to be finished, and then click to merge the PR.

PR #3077, opened 2025-02-10
3058: Fix truncated path links in push notifications

lunars97:
### Short Description
The underscore does not exist in regex for path which is leading for. ex. this path to pdf file `https://bildung-fuerth.de/fileadmin/media/pdf/Sprachwegweiser_2024.pdf ` being truncated in push notifications. 
### Proposed Changes

- add `_` to path in regex

### Side Effects

- none

### Testing

- Go to Testumgebung http://localhost:9000/testumgebung/de/news/local/11356
- Click on the link and see that the page is not found

### Resolved Issues

Fixes: #3058 

---

### DIFF
--- a/release-notes/unreleased/3058-fix-truncated-links-in-push-notifications.yml
+++ b/release-notes/unreleased/3058-fix-truncated-links-in-push-notifications.yml
@@ -1,0 +1,8 @@
+issue_key: 3058
+show_in_stores: true
+platforms:
+  - ios
+  - android
+  - web
+en: Fixed an issue with truncated links in local news
+de: Ein Problem mit abgeschnittenen Links in lokalen Nachrichten wurde behoben
--- a/shared/utils/replaceLinks.ts
+++ b/shared/utils/replaceLinks.ts
@@ -7,7 +7,7 @@ const unicode = 'a-zA-Z0-9\\u00A0-\\uD7FF\\uF900-\\uFDCF\\uFDF0-\\uFFEF'
 const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/
 const hostname = /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
 // language=RegExp
-const path = `(?:\\/(?:[+~%/.${unicode}-]*[+~%/${unicode}-]+)?)?`
+const path = `(?:\\/(?:[+~%/_.${unicode}-]*[+~%/${unicode}-]+)?)?`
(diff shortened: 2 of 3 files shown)


shared/utils/replaceLinks.ts
```diff
@@ -7,7 +7,7 @@ const unicode = 'a-zA-Z0-9\\u00A0-\\uD7FF\\uF900-\\uFDCF\\uFDF0-\\uFFEF'
 const protocol = /([A-Za-z]{3,9}:(?:\/\/)?)/
 const hostname = /(?:[-;:&=+$,\w]+@)?([A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*)/
 // language=RegExp
-const path = `(?:\\/(?:[+~%/.${unicode}-]*[+~%/${unicode}-]+)?)?`
+const path = `(?:\\/(?:[+~%/_.${unicode}-]*[+~%/_${unicode}-]+)?)?`
```
bahaaTuffaha, line 10: So the second underscore is for file name right?

bahaaTuffaha (approved):
Tested on Firefox and Chrome 👍️

shared/utils/replaceLinks.ts
steffenkleinle: I think the second part of the regex is just ensuring that the last character is not a dot (`.`) and is therefore matches the extension (`pdf`) in this case, so the second underscore should actually not be needed.
```suggestion
const path = `(?:\\/(?:[+~%/_.${unicode}-]*[+~%/{unicode}-]+)?)?`
```

steffenkleinle (commented):
Nice, tested on firefox and works like a charm! Could you perhaps add a small test case and write a release note? Thanks :)

release-notes/unreleased/3058-fix-truncated-links-in-push-notifications.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 3058
+show_in_stores: false
+platforms:
+  - ios
+  - android
+  - web
```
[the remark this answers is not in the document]
steffenkleinle: > No push notifications for web :)

Actually, this is not really about push notifications and more about local news (which exist on web). Therefore this is correct, but perhaps the note below should be adjusted to `Fixed an issue with truncated links in local news`.

@lunars97 any specific reason why we should not show the release notes in the stores? If not, could you please change the flag to true and add a german translation? Thanks :)

steffenkleinle (approved):
:tada:

PR #3081, opened 2025-02-12
2917: Fixes for tts on language change and app state change

steffenkleinle:
### Short Description

This PR fixes two issues I (and @bahaaTuffaha) discovered with the current native tts implementation:
- On language change, the wrong sentences were set to be read
- Some errors were logged that it is not possible to stop tts

### Proposed Changes

- Use useFocuseEffect instead of useLayoutEffect to only set sentences if the current page is in the foreground
- Only stop tts playing if actually plaing/enabled
- Since using a useFocusEffect requires wrapping in NavigationContainer for testing, I made this the default with opt out (sadly leads to many changes)

### Side Effects

N/A

### Testing

- Ensure no more stopping errors are logged to the console
- Ensure tts works as expected on language change:
  1. Navigate to a page with content
  2. Open the tts player and press play
  3. Change the language
  4. Press play again and ensure that the correct text (of the current page) is read
  5. Navigate back and ensure that still the correct text is played
  6. Just navigate a bit around the app and try different things

### Resolved Issues

Fixes: N/A

---


native/src/hooks/useTtsPlayer.ts
```diff
@@ -13,16 +14,18 @@ const useTtsPlayer = (model?: PageModel | LocalNewsModel | TunewsModel | undefin
...
+      if (model && model.content.length > 0) {
+        const content = parseHTML(model.content)
+        const sentences: string[] = segment(languageCode, content)
+        setSentences([model.title, ...sentences].filter(sentence => sentence.length > 0))
+      } else {
+        setSentences([])
+      }
+      return () => setSentences(previousSentences)
+    }, [previousSentences, setSentences, model, languageCode]),
+  )
```
steffenkleinle, line 28: Only wrapped this in useFocusEffect/useCallback

steffenkleinle:
Hint for reviewing: The relevant changes are in TtsContainer and useTtsPlayer

native/src/hooks/useTtsPlayer.ts
bahaaTuffaha: Nice

bahaaTuffaha (approved):
Tested.. works great 🚀 thanks.
I wonder if we gonna keep the ReadAloud accessible for the events?

steffenkleinle:
I wonder if we gonna keep the ReadAloud accessible for the events?

What do you mean by that? Can you elaborate a little please? :thinking:

bahaaTuffaha:
> I wonder if we gonna keep the ReadAloud accessible for the events?
> 
> What do you mean by that? Can you elaborate a little please? 🤔

Tts depends on the language selected but if the event contains german or fully in german it will be read with that selected language...

steffenkleinle:
> > I wonder if we gonna keep the ReadAloud accessible for the events?
> > What do you mean by that? Can you elaborate a little please? 🤔
> 
> Tts depends on the selected language but if the event contains german or fully in german it will be read with that selected language...

Hmm, you are talking about events that are in German, even though the selected language is e.g. arabic? I think we don't really have anything we can do here, since the cities can write the events in whatever way they want.

native/src/components/__tests__/SearchListItem.spec.tsx
```diff
@@ -36,17 +35,15 @@ describe('SearchListItem', () => {
 
   const renderWithNavigator = (query: string) =>
```
LeandraH, line 36: 🔧  I'd probably rename this since the navigator is gone now

LeandraH (approved):
I'm a little confused, the only functional change I see when changing the language while the text is being read is that it stops playing immediately when I open the overlay for the language change. Otherwise, both in main and in this branch, the text is being read from the beginning of the page after a language change. Is that the expected outcome?

Either way, I certainly don't see any issues with this PR.

bahaaTuffaha:
image

why useTtsPlayer test file doesn't have .spec to it?

bahaaTuffaha:
> I'm a little confused, the only functional change I see when changing the language while the text is being read is that it stops playing immediately when I open the overlay for the language change. Otherwise, both in main and in this branch, the text is being read from the beginning of the page after a language change. Is that the expected outcome?
> 
> Either way, I certainly don't see any issues with this PR.

I think when you switch language the screen will be changed so the sentences will be empty/null then the stop() will reset the setSentenceIndex to 0

image

LeandraH:
> I think when you switch language the screen will be changed so the sentences will be empty/null then the stop() will reset the setSentenceIndex to 0

Yes, that makes sense, I was just surprised because the description talked about fixing some behavior there, and I didn't notice a difference :)

steffenkleinle:
> I'm a little confused, the only functional change I see when changing the language while the text is being read is that it stops playing immediately when I open the overlay for the language change. Otherwise, both in main and in this branch, the text is being read from the beginning of the page after a language change. Is that the expected outcome?
> 
> Either way, I certainly don't see any issues with this PR.

For me, when changing the language on some nested category, the wrong sentences were often set. E.g. when going to Education > School > Primary School, opening the tts player and then changing language to German, the content of Schule (School) instead of Grundschule (Primary School) is read (just a made up example). This was due to a race condition of the use (layout) effects (which also trigger if not in the foreground). useFocusEffect only triggers on foreground which fixes the behavior as only the top route tries to set the tts sentences.

PR #3082, opened 2025-02-12
2189: Refactor and upgrade pois and bottom sheet

steffenkleinle:
### Short Description

Upgrade our native bottom sheet library and refactor its usages and pois.

### Proposed Changes

- Upgrade @gorhom/bottom-sheet
- Make the list title scroll as well
- Correctly use scroll views and flat lists
- Refactor pois

### Side Effects

N/A

### Testing

Checkout the pois and scroll through the bottom sheet. Select and deselect pois.

### Resolved Issues

Fixes: #2189

---


native/src/routes/Pois.tsx
```diff
@@ -177,11 +128,8 @@ const Pois = ({ pois: allPois, language, cityModel, route, navigation, refresh }
     </>
   )
 
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const currentBottomSheetHeight = snapPoints[sheetSnapPointIndex]!
-
   return (
-    <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
+    <View style={{ flex: 1 }}>
```
LeandraH, line 132: ❓ Does the styling still need to be inline here?

LeandraH (approved):
Nice! Tested on real iOS

native/src/routes/Pois.tsx
steffenkleinle: I don't think so, good point :+1:

lunars97 (approved):
Tested on Android Emulator. Looks good :+1:

bahaaTuffaha (approved):
It works great 🚀
- It got broken once that the yellow circle couldn't keep up with the list but I couldn't reproduce this problem

PR #3086, opened 2025-02-13
Fix iOS build scripts

steffenkleinle:
### Short Description

Fix iOS build scripts as xcode does not find homebew (and therefore yarn) in my path.
Otherwise I always get `yarn: command not found`.

### Proposed Changes

- Add homebrew to path before executing build scripts

### Side Effects

homebrew is always added to the path even if not available/necessary. Should not lead to any problems though.

### Testing

Build the app on iOS.

### Resolved Issues

Fixes: N/A

---


f1sh1918 (approved):
Build worked for me 👍

LeandraH (approved):
Me too :)

Issue #3018, opened 2024-12-04
Issue with Message Linking


nassabay:
### Describe the Bug

The Leipzig District Office informed me about the following issue:
A message was sent and successfully delivered to the phone. However, when clicking on the message, it does not lead to Messages (Nachrichten) but instead to Local Information.

### Steps to Reproduce
Tests in Testumgebung:
1. Go to Testumgebung 
2. Nachrichten
3. New "Nachricht erstellen"
4. Send Nachricht

Android: Everything works flawlessly. Svenja and Natacha have tested and confirmed this.
iPhone: There are discrepancies:
Devices and Versions:
Both devices: iPhone SE
Both devices: iOS version 18.1.1

Results:
Personal iPhone:
The message was received once. Clicking on it leads to Local Information.
Work iPhone:
The message was received twice.
- First message: Clicking on it leads to page "Gesellschaftliches Leben".
- Second message: Clicking on it leads to Local Information.

### Expected Behavior

The local news corresponding to the push notification is opened.

### Additional Information

Discussion: https://chat.tuerantuer.org/digitalfabrik/pl/i54axc4k7bnqxctkq5jxem5iko

<img width="150px" src="https://github.com/user-attachments/assets/9af761b9-0501-4a68-84c3-ea1c7d84d504" />
<img width="150px" src="https://github.com/user-attachments/assets/02cca25f-2975-4838-84a4-2d615a738fda" />
<img width="150px" src="https://github.com/user-attachments/assets/ade40703-df94-4d77-9700-6ced8bde9f93" />

steffenkleinle:
> I just tested this on my iPhone, sending notifications in Testumgebung in the test CMS and the prod CMS. In both cases, I received only a single notification, and tapping on it opened the app at the page that it had been at before (in this case, the categories home screen) and not to the news page.

Same for me. I think we should prioritize this a little :)

PR #3089, opened 2025-02-13
3018: Issue with message linking

f1sh1918:
### Short Description

It looks like notifee is blocking some background notification listener on the particular version we used.
Thats why the deeplinking was not executed

### Proposed Changes

- update notifee to latest

### Side Effects

- none

### Testing

1. Close the app
2. Send a PN via cms
3. Open the PN and you should be redirected to the correct route
4. I face the issue that if the app is open the foreground notification push redirects to 404, since the page is not yet ready. I believe this is just a testCMS issue (https://github.com/digitalfabrik/integreat-cms/issues/3378) and can be solved by increasing the TIMEOUT or using the production cms to test

Note: I will test it on productions cms when i have an account for it.

### Resolved Issues

Fixes: #3018
Tested on android and ios

---


steffenkleinle (approved):
Tested on iOS simulator and android real device. Seems to work. On android it did not open the message when the app was in a quit state (force killed). On iOS simulator I only got a notification in the foreground or when opening the app from the background. I did not get a notification in quit state (however, that is most likely due to it being a simulator and not an iOS real device).
I'll create a follow up issue for the quit state issue.
Thanks for the fix!

LeandraH (approved):
Nicely done, thank you! Tested on real iOS

Issue #3074, opened 2025-02-05
Improving responsiveness for the content search


bahaaTuffaha:
### Describe the Problem

While typing in search its laggy sometimes and it could affect the user experience.

### Describe Your Preferred Solution

Using a Debounce function to delay the query/search while typing a few milliseconds just to finish typing.
Ideally, we would do as much as possible of this functionality in /shared to share with native.

### Describe Possible Alternatives

Keeping it the way it is.

PR #3103, opened 2025-02-18
3074: Add debounce function to query search to improve responsiveness

lunars97:
### Short Description

Return the results after several milliseconds to user so that the user could have time to finish typing

### Proposed Changes

- Add debounce function to `useSearch` hook
- Add timeout for 250 Milliseconds

### Side Effects
- should be none

### Testing

Search for something, for ex. "Dolmetscher"
See that the results for the query appear after 250 Milliseconds

### Resolved Issues

Fixes: #3074 

---

### DIFF
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -7,6 +7,7 @@ import ExtendedPageModel from '../api/models/ExtendedPageModel'
 import PoiModel from '../api/models/PoiModel'
 
 export type SearchResult = ExtendedPageModel
+const DEBOUNCED_QUERY_TIMEOUT = 250
 
 export const prepareSearchDocuments = (
   categories?: CategoriesMapModel | null,
@@ -22,6 +23,18 @@ export const prepareSearchDocuments = (
 // Modifying single documents or replacing documents with a same length array will therefore NOT trigger an update
 const useSearch = (documents: SearchResult[], query: string): SearchResult[] => {
   const [indexing, setIndexing] = useState(false)
+  const [debouncedQuery, setDebouncedQuery] = useState(query)
+
+  useEffect(() => {
+    const debounceQueryTimeout = setTimeout(() => {
+      setDebouncedQuery(debouncedQuery)
+    }, DEBOUNCED_QUERY_TIMEOUT)
+
+    return () => {
+      clearTimeout(debounceQueryTimeout)
+    }
+  }, [debouncedQuery])
+
   const [search] = useState(
     new MiniSearch({
       idField: 'path',
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -25,16 +25,6 @@ const useSearch = (documents: SearchResult[], query: string): SearchResult[] => 
   const [indexing, setIndexing] = useState(false)
   const [debouncedQuery, setDebouncedQuery] = useState(query)
 
-  useEffect(() => {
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
Very cool, thank you!

shared/hooks/useSearch.ts
```diff
@@ -22,6 +23,18 @@ export const prepareSearchDocuments = (
...
+
+  useEffect(() => {
+    const debounceQueryTimeout = setTimeout(() => {
+      setDebouncedQuery(debouncedQuery)
+    }, DEBOUNCED_QUERY_TIMEOUT)
+
+    return () => {
+      clearTimeout(debounceQueryTimeout)
+    }
+  }, [debouncedQuery])
```
steffenkleinle, line 36: :upside_down_face: Could you perhaps move this below the mini search initialization/useState? Feels more like the correct order (initialization first, effects and behavior second).

shared/hooks/useSearch.ts
```diff
@@ -22,6 +23,18 @@ export const prepareSearchDocuments = (
...
+  const [debouncedQuery, setDebouncedQuery] = useState(query)
+
+  useEffect(() => {
+    const debounceQueryTimeout = setTimeout(() => {
+      setDebouncedQuery(debouncedQuery)
+    }, DEBOUNCED_QUERY_TIMEOUT)
+
+    return () => {
+      clearTimeout(debounceQueryTimeout)
+    }
```
steffenkleinle, line 35: :upside_down_face: 
```suggestion
    return () => clearTimeout(debounceQueryTimeout)
```

shared/hooks/useSearch.ts
```diff
@@ -22,6 +23,18 @@ export const prepareSearchDocuments = (
 // Modifying single documents or replacing documents with a same length array will therefore NOT trigger an update
 const useSearch = (documents: SearchResult[], query: string): SearchResult[] => {
   const [indexing, setIndexing] = useState(false)
+  const [debouncedQuery, setDebouncedQuery] = useState(query)
+
+  useEffect(() => {
+    const debounceQueryTimeout = setTimeout(() => {
+      setDebouncedQuery(debouncedQuery)
```
steffenkleinle, line 30: :question: Shouldn't this be set to `query` after the timeout?
```suggestion
      setDebouncedQuery(query)
```

shared/hooks/useSearch.ts
```diff
@@ -22,6 +23,18 @@ export const prepareSearchDocuments = (
 // Modifying single documents or replacing documents with a same length array will therefore NOT trigger an update
 const useSearch = (documents: SearchResult[], query: string): SearchResult[] => {
   const [indexing, setIndexing] = useState(false)
+  const [debouncedQuery, setDebouncedQuery] = useState(query)
```
steffenkleinle, line 26: :x: It seems to me like this is never actually used. Did you forget to use it in L61?

steffenkleinle (changes requested):
Thanks for the PR, I think this is a very nice implementation. I just think you forgot to actually make use of it :thinking: 

How about adding a release note for this (along the lines of `making search more responsive`)?

shared/hooks/useSearch.ts
lunars97: yes, you are absolutely right :speak_no_evil:

shared/hooks/useSearch.ts
lunars97: :see_no_evil: yes

PR #3113, opened 2025-02-19
3112: Prevent unnecessary tts stop

steffenkleinle:
### Short Description

Prevent unnecessary tts stop causing errors if no tts engine installed.

### Proposed Changes

- Prevent unnecessary tts stop causing errors if no tts engine installed.
- Refactor reportError

### Side Effects

- [This error](https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/1502) will not be sent to sentry anymore. We have 537k events for it, which seems more than enough. I [investigated this](#1759) in the past, but wasn't able to fix it. However, it is handled and not a problem.
- ~~I thought of preventing sending a sentry stacktrace for known issues (such as this one), but decided against as it still might be helpful to detect issues such as this one. However, we will continue to get this error for people that try to open the tts player but dont have a tts player installed.~~ Due to the unexpected behavior on language change described below, I reverted the previous changes and ignore the error instead.

### Testing

Not sure how to test this. Have a degoogled android phone I guess. Or just throw an error in `stop` and ensure that it is not called without actually opening the tts function.

### Resolved Issues

Fixes: #3112

---

### DIFF
--- a/native/src/components/TtsContainer.tsx
+++ b/native/src/components/TtsContainer.tsx
@@ -138,11 +138,9 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
   const updateSentences = useCallback(
     (newSentences: string[]) => {
       setSentences(newSentences)
-      if (isPlaying) {
-        stop()
-      }
+      stop()
     },
-    [stop, isPlaying],
+    [stop],
   )
 
   const ttsContextValue = useMemo(
--- a/native/src/components/TtsContainer.tsx
+++ b/native/src/components/TtsContainer.tsx
@@ -138,9 +138,11 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
   const updateSentences = useCallback(
     (newSentences: string[]) => {
       setSentences(newSentences)
-      stop()
+      if (isPlaying) {
+        stop()
+      }
     },
-    [stop],
+    [stop, isPlaying],
   )
 
   const ttsContextValue = useMemo(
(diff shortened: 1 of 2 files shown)


f1sh1918 (changes requested):
Looks like your changes crashed sth on ios. On main everything works fine.
Tested on ios simulator (ios 18.0)

If i click play it reads only the headline and shows "nichts zu lesen" and stops
Guess the `updateSentence` is broken somehow

https://github.com/user-attachments/assets/382751ca-f9ae-44fb-b4e8-e278d743232c

native/src/components/TtsContainer.tsx
```diff
@@ -138,9 +138,11 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
   const updateSentences = useCallback(
     (newSentences: string[]) => {
       setSentences(newSentences)
-      stop()
+      if (isPlaying) {
+        stop()
+      }
     },
-    [stop],
+    [stop, isPlaying],
```
bahaaTuffaha, line 145:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [stop],
```

steffenkleinle: :-1: Then this will always call stop or never call stop depending whether `isPlaying` is true or false the first time this is run.

bahaaTuffaha (approved):
Works as expected on Android...
I don't have a degoogled phone to test it on 😐️

Issue #3107, opened 2025-02-19
Too long loading time in search


Docmale:
### Describe the Bug

When openening the search page in region  Landeshauptstadt München and tapping the first letter, screen says 'loading' and loads for 1.5 to 2 minutes till the result is shown. This happens for every new opening of search page.
Measured with Samsung A53; on Pixel 3a Emulator I canceled search after aprox. 5 minutes.
Web is ok

### Steps to Reproduce

1. Go to 'Landeshauptstadt München'
2. Open search page
3. Enter a single letter eg. 'S'
4. Wait til results are shown
5. Clos search page and repeat from (2.)

### Expected Behavior

No or at least short waiting time only once opening search page

### Addition Information

In Augsburg Stadt it takes more than 10 seconds what is also too long

PR #3115, opened 2025-02-19
3107: Reduce search loading times

steffenkleinle:
### Short Description

Reduce search loading times by fixing minisearch initialization and rename to documents

### Proposed Changes

- Decouple initialization of minisearch and adding documents
- Avoid rerender if documents are of the same length
- Avoid returning null if not all documents indexed
- Change naming to `(search) documents`

### Side Effects

- Search results might appear after some delay

### Testing

Test search on native and web thoroughly! Test both searching in German and other languages and make sure that German search terms are also found in other languages.

### Resolved Issues

Fixes: #3107

---

### DIFF
--- a/native/src/routes/CategoriesContainer.tsx
+++ b/native/src/routes/CategoriesContainer.tsx
@@ -3,6 +3,7 @@ import { useWindowDimensions } from 'react-native'
 
 import { CATEGORIES_ROUTE, CategoriesRouteType, cityContentPath } from 'shared'
 import { ErrorCode } from 'shared/api'
+import { config } from 'translations'
 
 import Categories from '../components/Categories'
 import DashboardNavigationTiles from '../components/DashboardNavigationTiles'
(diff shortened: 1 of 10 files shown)


bahaaTuffaha (approved):
Native feels faster already Nice work 🚀
Tested on Android and for web Firefox and Brave
I think the web will be better when 3074 is done.

steffenkleinle:
> Note: The search route now starts off with a loading screen without showing the search bar. I think it's a better UX to have the search bar shown while the results load. Either way, doesn't have to be done here.

Hmm, I can't reproduce that (or I can, but that is nothing new). Also, I am very sure that if that happens now it already happened before, since I only removed loading spinners and did not add any. If you open a new city/language, this might happen since loading the city content is not yet finished. This is the same behavior as before though.

LeandraH:
> > Note: The search route now starts off with a loading screen without showing the search bar. I think it's a better UX to have the search bar shown while the results load. Either way, doesn't have to be done here.
> 
> Hmm, I can't reproduce that (or I can, but that is nothing new). Also, I am very sure that if that happens now it already happened before, since I only removed loading spinners and did not add any. If you open a new city/language, this might happen since loading the city content is not yet finished. This is the same behavior as before though.

https://github.com/digitalfabrik/integreat-app/issues/3008 Correct, it used to happen before, until I fixed it last month :D

Issue #2908, opened 2024-09-02
Clicking on search result only closes keyboard


steffenkleinle:
### Describe the Bug

Clicking on a search result only closes the keyboard. In order to actually navigate to the search result, I have to click on the search result again.

### Steps to Reproduce

~0. (Use an android device? not sure if this happens on iOS too)~
1. Go to search
2. Enter search text
3. Click on search result
4. See that only the keyboard closes and no navigation happens

### Expected Behavior

Clicking on a search result should navigate to the search result.

### Actual Behavior

Only the keyboard closes.

### Additional Information

To be discussed. Perhaps the current behavior is also best practice?

LeandraH:
Also happens on my iPhone

steffenkleinle:
@hauf-toni is the current behavior good ux or should clicking a search result directly open it instead of just closing the keyboard?

hauf-toni:
best practice is to ensure seamless navigation when clicking on a search result, without requiring an additional click to close the keyboard. When a user clicks on a search result, the keyboard should close and the user should navigate to the selected search result @steffenkleinle

PR #3122, opened 2025-02-22
2908: Immediately open link in search

LeandraH:
### Short Description

When you're in the mobile search, have typed in your query, and then tapped on the result you wanted, the keyboard would close and you would have to tap a second time on the result to actually get there. This PR fixes that, you are now sent immediately to the selected result.

### Proposed Changes

- Added [keyboardShouldPersistTaps](https://reactnative.dev/docs/scrollview#keyboardshouldpersisttaps) to the list of search items

### Side Effects

- Hopefully none, the behavior of the other lists shouldn't be affected.

### Testing

Type in a search, tap on a result, see if it opens immediately.

### Resolved Issues

Fixes: #2908 

---

### DIFF
--- a/native/src/components/List.tsx
+++ b/native/src/components/List.tsx
@@ -20,6 +20,7 @@ type ListProps<T> = {
   refresh?: () => void
   onEndReached?: () => void
   style?: ViewStyle
+  keyboardShouldPersistTaps?: 'always' | 'never' | 'handled'
 }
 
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):
Tested on android, works as expected. Should we add a release note for this?

bahaaTuffaha (approved):
Nice✨ tested on Android

Issue #3094, opened 2025-02-15
Push notifications not correctly opened in quit state


steffenkleinle:
### Describe the Bug

At the moment, push notifications are not correctly handled if the app is in a quit state on android (and most likely on iOS, too). The app only opens on the last opened page. In theory, this might also be a development issue though.

### Steps to Reproduce

1. Force kill the app
2. Send a push notification from the CMS
3. Click the push notificaiton
4. See the app open on the start screen/city dashboard.

### Expected Behavior

Clicking the push notification opens the correct local news item.

### Related Issues

Follow up for #3018 which fixed a similar issue on iOS (background app state).

### Additional Information

Apparently all messages should now be handled with notifee (see [rnfirebase docs](https://rnfirebase.io/messaging/usage#background--quit-state-messages)):
> Note: If you use @notifee/react-native, since v7.0.0, onNotificationOpenedApp and getInitialNotification will no longer trigger as notifee will handle the event.

Probably we need to implement the `setBackgroundMessageHandler` in the same/similar way to `onMessage` , see the [notifee docs](https://docs.page/invertase/notifee/react-native/integrations/fcm#subscribing-to-messages).

Please check if we can get rid of some of the current rnfirebase listeners (backgroundAppStatePushNotificationListener, quitAppStatePushNotificationListener).

LeandraH:
I just tested and could reproduce this.

PR #3125, opened 2025-02-24
3094: Push notifications quit state solution

bahaaTuffaha:
### Short Description

At the moment, push notifications are not correctly handled if the app is in a quit state on android (and most likely on iOS, too). The app only opens on the last opened page. In theory, this might also be a development issue though.

### Proposed Changes

- Used setBackgroundMessageHandler instead of  importFirebaseMessaging
- I don't think it needs unsubing like it was from my searching over internet.
- there was an error for jest about:
```
Invariant Violation: `new NativeEventEmitter()` requires a non-null argument. 
```
I solved it by `jest.mock('react-native/Libraries/EventEmitter/NativeEventEmitter')` according to [here](https://stackoverflow.com/a/70724310)

### Side Effects

- I hope none

### Resolved Issues

Fixes: #3094

---


LeandraH (commented):
Oooof, I'm always scared of touching the notifications, kudos for daring! But I have a few problems:

1. The title of the PR says that it's about a fix for when the app is in quit state, as in not in the background but actually closed. The code is a fix for the app in background state. So this doesn't actually resolve the linked ticket, or am I missing something here?

2. Can you say a bit more about why the unsubscribing wasn't necessary?

3. When testing this PR on my iPhone with the app in the background, I got no notification at all. I had been testing notifications pretty extensively on my update branch before, so it's probably not because of my device :(

Issue #3088, opened 2025-02-13
Remove prerelease flag from all beta releases when promoting


steffenkleinle:
### Describe the Problem

Currently, only the latest prerelease flag of the latest github release of the current platform is removed during a promotion. Accordingly, there are quite a few releases which still have the prerelease flag set.
However, this is wrong, as also those releases (or whatever is included in those releases) is now on production.

### Describe Your Preferred Solution

When promoting releases for a platform (android, ios, web) with github-promote-release, all releases corresponding to that platform should loose their prerelease flag.
Releases for other platforms should not be affected.

### Describe Possible Alternatives

Keep it as it is.

### Related Issues

❗ Check the status of integrating the app toolbelt, this will lead to conflicts: #2644
Related to/blocked by: #3069 

### Additional Information

N/A

PR #3079, opened 2025-02-11
3069: Add a note for promotions

bahaaTuffaha:
### Short Description

We need a way to show what is went from beta to prod.

### Proposed Changes

- Created a function called `isLatestReleasePreRelease` to check the releases and it looks for preRelease tag for each platform
- Added the line `[The most recent beta version was promoted to production]` if the last release is beta on that platform.
Note: I can go a step further and read the body of each release to get more details if needed !

### Side Effects

- `manage-metadata.ts`

### Testing

n/a

### Resolved Issues

Fixes: #3069 

---

PR #3127, opened 2025-02-25
3088: Remove all previous prereleases

bahaaTuffaha:
### Short Description

Currently, only the latest prerelease flag of the latest github release of the current platform is removed during a promotion. Accordingly, there are quite a few releases which still have the prerelease flag set.
However, this is wrong, as also those releases (or whatever is included in those releases) is now on production.

### Proposed Changes

- I filtered all Ids that are related to that platform then updateRelease for each id of these.
Note: this PR will have conflects with #3079

### Side Effects

None

### Testing

N/A

### Resolved Issues

Fixes: #3088

---

### DIFF
--- a/tools/github-promote-release.ts
+++ b/tools/github-promote-release.ts
@@ -25,12 +25,11 @@ const getAllPreReleased = async ({ deliverinoPrivateKey, owner, repo, platform }
   const releasesWithTags = releases.data.filter(
     (release: ReleaseType) => release.tag_name.includes(platform) && release.prerelease,
   )
-  const filteredIds = releasesWithTags.map((release: ReleaseType) => release.id)
   const filteredTagNames = releasesWithTags.map((release: ReleaseType) => release.tag_name)
 
-  if (filteredIds.length > 0) {
+  if (releasesWithTags.length > 0) {
     console.warn('Unset prerelease tags of ', filteredTagNames)
-    return filteredIds
+    return releasesWithTags
   }
 
   console.warn('No release found to unset the prerelease tag for. Latest release may already be non-prerelease')
--- a/tools/github-promote-release.ts
+++ b/tools/github-promote-release.ts
@@ -6,7 +6,6 @@ import authenticate from './github-authentication.js'
 
 const octokit = new Octokit()
 type Releases = GetResponseTypeFromEndpointMethod<typeof octokit.repos.listReleases>
-type ReleaseType = Releases['data'][number]
 
 type Options = {
   deliverinoPrivateKey: string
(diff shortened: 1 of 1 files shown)


steffenkleinle (changes requested):
Thanks for the PR :rocket: I think in general this does everything needed, but perhaps a little bit too much :D It should not interfere with the promotion message. It is enough to remove the prerelease flag of older releases, no need to actually return them/log them.

tools/github-promote-release.ts
```diff
@@ -14,39 +15,47 @@ type Options = {
...
+        repo,
+        release_id: preRelease.id,
+        prerelease: false,
+        make_latest: platform === 'android' ? 'true' : 'false', // We always want android to be the latest release, so a link to the latest github release will go to the apk
+      })
+      console.warn('Http response code of updating the result: ', result.status)
+    }),
+  )
+  // Returning the least release with PreRelease tag
+  return allPreReleases[0]
```
steffenkleinle, line 58: :x: Hmm... with the current implementation the information is lost whether there was actually a release promoted since there is no way anymore to tell whether the latest release was a prerelease or a production release. We should only return a prerelease here if it was actually the latest release. To work around that, the current `getAllPreReleased` should just return all releases with the filtering (and udpating) done in this function (removePreRelease, should probably be renamed to `removePreReleases`).

steffenkleinle: yes, but you filter beforehand to only include prereleases, meaning production releases are not part of the array anymore. So if the latest release was a production release, we dont actually promote anything, but there might still be prerelease somewhere.

bahaaTuffaha: We need to discuss this 🙈

tools/github-promote-release.ts
```diff
@@ -14,39 +14,42 @@ type Options = {
...
+      console.warn('Http response code of updating the result: ', result.status)
+    }),
+  )
 
-const removePreRelease = async ({ deliverinoPrivateKey, owner, repo, platform }: Options) => {
-  const release = await getReleaseIdBody({ deliverinoPrivateKey, owner, repo, platform })
-  if (!release) {
-    return null
+  const filteredTagNames = preReleases.map(release => release.tag_name)
+  console.warn('Unset prerelease tags of ', filteredTagNames)
```
steffenkleinle, line 46: :upside_down_face: Optional: How about moving this to L41 inside the `.map()`?
e.g. something like this:
``` ts
      console.warn(`Release ${release.tag_name} promoted with status:`, result.status)
```

steffenkleinle (approved):
Looks good, nice work :+1:

steffenkleinle (approved):
Nice work! Lets hope this works as expected :)

LeandraH (approved):
Looks great, and easier to understand, thank you!

Issue #3117, opened 2025-02-20
Search term half-stays on language change


LeandraH:
### Describe the Bug

When changing the language from a search, the search term stays in the search bar but disappears in the URL, and no results are shown.

### Steps to Reproduce

1. Go to 'http://localhost:9000/augsburg/de/search/?query=alleine'
2. Switch to a different language
3. See that the search term is still in the search bar but the results disappear and the query in the URL also disappears.

### Expected Behavior

I would like the search query to disappear from the search bar.

Alternatively, we could keep the query in the URL and keep showing results, but my expectation as a user would be that the query disappears.

PR #3134, opened 2025-03-04
3117: Fix search term half-stays on language change

lunars97:
### Short Description
Currently search term stays in input after language change but the url query gets cleared out. The query and search term input should be there after language change

### Proposed Changes

- change `defaultValue` to `value` in TextInput because defaultValue contains the default value, while value contains the current value after some changes have been made. 
- use UseEffect to update filterText

### Side Effects

- none

### Testing

- Go to the search page
- Search for `Augsburg`
- Change the language
- See that url query and search input text disappear

### Resolved Issues

Fixes: #3117 

---

### DIFF
--- a/web/src/routes/SearchPage.tsx
+++ b/web/src/routes/SearchPage.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 import { useLocation, useNavigate } from 'react-router-dom'
 import styled from 'styled-components'
 
-import { parseHTML, pathnameFromRouteInformation, SEARCH_ROUTE, useSearch } from 'shared'
+import { parseHTML, pathnameFromRouteInformation, SEARCH_ROUTE, useSearch, SEARCH_QUERY_KEY } from 'shared'
 import { config } from 'translations'
 
 import { CityRouteProps } from '../CityContentSwitcher'
@@ -32,7 +32,7 @@ const SearchCounter = styled.p`
 `
 
 const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps): ReactElement | null => {
-  const query = new URLSearchParams(useLocation().search).get('query') ?? ''
+  const query = new URLSearchParams(useLocation().search).get(SEARCH_QUERY_KEY) ?? ''
   const [filterText, setFilterText] = useState<string>(query)
   const { t } = useTranslation('search')
   const navigate = useNavigate()
@@ -60,7 +60,7 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
   }
 
   const languageChangePaths = city.languages.map(({ code, name }) => ({
-    path: pathnameFromRouteInformation({ route: SEARCH_ROUTE, cityCode, languageCode: code }),
+    path: `${pathnameFromRouteInformation({ route: SEARCH_ROUTE, cityCode, languageCode: code })}/?${SEARCH_QUERY_KEY}=${query}`,
     name,
     code,
   }))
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
Very nice, thank you!

f1sh1918 (approved):
Looks good 👍 
Nicely done!

steffenkleinle (changes requested):
Just requesting changes to avoid merging to discuss this. I am not 100% sure but shouldn't the solution be to change the url and keep the search term instead of clearing the search? I'd expect that the reason why users would change the language while searching something would be to get (different) results in another language instead of starting with a clean search? I am not entirely sure though, but I think I would expect that the query is the same after a language change.

LeandraH:
No, I would expect the query to disappear when changing languages, I would expect to not be using the same words in a different language, especially since German is already handled.

Playing around with changing language during a search, I have found both possibilities on other sites.

steffenkleinle:
> No, I would expect the query to disappear when changing languages, I would expect to not be using the same words in a different language, especially since German is already handled.
> 
> Playing around with changing language during a search, I have found both possibilities on other sites.

As discussed [here](https://chat.tuerantuer.org/digitalfabrik/pl/64ckgrwci7d15efq4hh79yk8ey) the query should be persistent between language changes as it seems like the more user-friendly option.

web/src/routes/SearchPage.tsx
```diff
@@ -38,6 +38,11 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
   const navigate = useNavigate()
   const fallbackLanguage = config.sourceLanguage
 
+  useEffect(() => {
+    const appendToUrl = filterText.length !== 0 ? `?query=${filterText}` : ''
+    navigate(`${pathname}/${appendToUrl}`, { replace: true })
+  }, [filterText, navigate, pathname])
+
```
steffenkleinle, line 45: :wrench: Do you think this `useEffect` is really needed? What do you think of setting the query to the `languageChangePaths` in L68?

lunars97: you are right, it makes more sense just to include it here :+1:

web/src/routes/SearchPage.tsx
```diff
@@ -60,7 +60,7 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
   }
 
   const languageChangePaths = city.languages.map(({ code, name }) => ({
-    path: pathnameFromRouteInformation({ route: SEARCH_ROUTE, cityCode, languageCode: code }),
+    path: `${pathnameFromRouteInformation({ route: SEARCH_ROUTE, cityCode, languageCode: code })}/?query=${query}`,
```
steffenkleinle, line 63: :wrench: Would be cool to use the constant for this instead of hardcoding:
``` ts
    path: `${pathnameFromRouteInformation({ route: SEARCH_ROUTE, cityCode, languageCode: code })}/?${SEARCH_QUERY_KEY}=${query}`,
```

steffenkleinle (approved):

web/src/constants/urls.ts
```diff
@@ -3,3 +3,4 @@ import buildConfig from './buildConfig'
 
 export const cmsApiBaseUrl = safeLocalStorage.getItem(API_URL_KEY) || buildConfig().cmsUrl
 export const tunewsApiBaseUrl = 'https://tunews.integreat-app.de'
+export const SEARCH_QUERY_KEY = 'query'
```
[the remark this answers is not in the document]
lunars97: ahh okay, I did not know about it...

PR #3137, opened 2025-03-04
Add reusable debounce hook to improve search query responsiveness

lunars97:
### Short Description

This is an improved version of already merged PR #3103. Thanks to @bahaaTuffaha who noticed that the responsiveness has not been that much improved and suggested to apply debounce function directly in SearchPage.tsx (web) and SearchModal.tsx (native). I also noticed that the responsiveness has become much better.

### Proposed Changes

- Removed changes made in useSearch hook in #3103
- Created a reusable hook `useDebounce` in shared
- Applied hook in SearchPage.tsx and SearchModal.tsx

### Side Effects
should be none

### Resolved Issues

Fixes: #

---


bahaaTuffaha (changes requested):
Now I can see the difference .. nice work 
I didn't test native yet...

bahaaTuffaha:
Adding memo to SearchListItem will significantly make a difference from my observation because it will skip re-rendering when props are changed. [memo](https://react.dev/reference/react/memo) 
const SearchListItem = React.memo(({ title, contentWithoutHtml, query, path, thumbnail }: SearchListItemProps): const 
```
SearchListItem = React.memo(
  ({ title, contentWithoutHtml, query, path, thumbnail }: SearchListItemProps): ReactElement => {
  ```

File: web/src/routes/SearchPage.tsx [hunk 1]
```diff
@@ -38,6 +39,13 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
   const navigate = useNavigate()
   const fallbackLanguage = config.sourceLanguage
 
+  const debouncedQuery = useDebounce(filterText, DEBOUNCED_QUERY_TIMEOUT)
+
+  useEffect(() => {
+    const appendToUrl = filterText.length !== 0 ? `?query=${filterText}` : ''
+    navigate(`${pathname}/${appendToUrl}`, { replace: true })
+  }, [debouncedQuery, navigate, location.pathname])
```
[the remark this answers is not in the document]
lunars97: we need here navigate and location dependencies

native/src/routes/SearchModal.tsx
```diff
@@ -49,8 +50,10 @@ const SearchModal = ({
   const resourceCache = useResourceCache({ cityCode, languageCode })
   const { t } = useTranslation('search')
 
-  const contentLanguageResults = useSearch(documents, query)
-  const fallbackLanguageResults = useSearch(fallbackLanguageDocuments, query)
+  const debouncedQuery = useDebounce(query, DEBOUNCED_QUERY_TIMEOUT)
```
steffenkleinle, line 53: :wrench: I think it would be nice to use `DEBOUNCED_QUERY_TIMEOUT` as default, such that there is no need to import/explicitly pass this here.
```suggestion
  const debouncedQuery = useDebounce(query)
```

shared/hooks/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
+import { useState, useEffect } from 'react'
+
+const useDebounce = <T>(value: T, delay: number) => {
```
steffenkleinle, line 3: :wrench: I think it would be nice to set a default here. As long as it is only used here, there is no need to keep it in `constants`.
```suggestion
const DEBOUNCE_TIMEOUT = 250

const useDebounce = <T>(value: T, delay = DEBOUNCE_TIMEOUT) => {
```

shared/hooks/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
...
+  const [debouncedValue, setDebouncedValue] = useState<T>(value)
+
+  useEffect(() => {
+    const debouncedValueTimeout = setTimeout(() => {
+      setDebouncedValue(value)
+    }, delay)
```
steffenkleinle, line 9: :upside_down_face: 
```suggestion
    const debouncedValueTimeout = setTimeout(() => setDebouncedValue(value), delay)
```

web/src/routes/SearchPage.tsx
```diff
@@ -38,6 +39,13 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
   const navigate = useNavigate()
   const fallbackLanguage = config.sourceLanguage
 
+  const debouncedQuery = useDebounce(filterText, DEBOUNCED_QUERY_TIMEOUT)
```
steffenkleinle, line 42:
```suggestion
  const debouncedQuery = useDebounce(filterText)
```

web/src/routes/SearchPage.tsx
```diff
@@ -38,6 +39,13 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
...
+
+  useEffect(() => {
+    const appendToUrl = debouncedQuery.length !== 0 ? `?query=${debouncedQuery}` : ''
+    navigate(`${pathname}/${appendToUrl}`, { replace: true })
+  }, [debouncedQuery, navigate, pathname])
```
steffenkleinle, line 47: :question: Is there a reason why we should not directly update the url? I think the search input and the url should always be in sync to avoid weird side effects. I think its okay that the url is updated before the search results are.

Edit: I see, problems arise due to the use of a `useState` for `filterText`. I think we should remove it and only use/update the query in the url search params directly like this by replacing L36 and L37
``` ts
  const query = new URLSearchParams(useLocation().search).get('query') ?? ''
  const [filterText, setFilterText] = useState<string>(query)
```
with
``` ts
  const [searchParams, setSearchParams] = useSearchParams()
  const query = searchParams.get('query') ?? ''
```

It is always desirable to avoid duplicated states, in this case the url search params are enough. Obviously, this was not introduced in your PR but I think it would be cool if you could refactor this here accordingly?
Let me know if you have any questions :)

web/src/routes/SearchPage.tsx
```diff
@@ -82,8 +90,6 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
 
   const handleFilterTextChanged = (filterText: string): void => {
     setFilterText(filterText)
-    const appendToUrl = filterText.length !== 0 ? `?query=${filterText}` : ''
-    navigate(`${pathname}/${appendToUrl}`, { replace: true })
```
steffenkleinle, line 86: :wrench: As mentioned above, I think we should directly update the url, so this should probably be reverted. With the change from above, you could then do the following here:
``` ts
    setSearchParams({ query })
    const appendToUrl = query.length !== 0 ? `?query=${query}` : ''
    navigate(`${pathname}/${appendToUrl}`, { replace: true })
```

steffenkleinle (changes requested):
Tested on web, nice idea @bahaaTuffaha and well done @lunars97! 

This is closely related and might conflict with #3117 if my proposals are implemented. In fact, #3117 is actually caused imo by the duplication of state in the url search params and filter text use state.

web/src/routes/SearchPage.tsx
lunars97: true. Your version with direct update seems better :+1:

web/src/routes/SearchPage.tsx
```diff
@@ -33,11 +33,12 @@ const SearchCounter = styled.p`
 `
 
 const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps): ReactElement | null => {
-  const query = new URLSearchParams(useLocation().search).get(SEARCH_QUERY_KEY) ?? ''
-  const [filterText, setFilterText] = useState<string>(query)
+  const [searchParams, setSearchParams] = useSearchParams()
+  const query = searchParams.get('query') ?? ''
```
steffenkleinle, line 37: 🔧 
```suggestion
  const query = searchParams.get(SEARCH_QUERY_KEY) ?? ''
```

web/src/routes/SearchPage.tsx
```diff
@@ -83,9 +84,9 @@ const SearchPage = ({ city, cityCode, languageCode, pathname }: CityRouteProps):
     )
   }
 
-  const handleFilterTextChanged = (filterText: string): void => {
-    setFilterText(filterText)
-    const appendToUrl = filterText.length !== 0 ? `?${SEARCH_QUERY_KEY}=${filterText}` : ''
+  const handleFilterTextChanged = (query: string): void => {
+    setSearchParams({ query })
+    const appendToUrl = query.length !== 0 ? `?query=${query}` : ''
     navigate(`${pathname}/${appendToUrl}`, { replace: true })
```
steffenkleinle, line 90: 🤔 I actually think these two lines are not necessary anymore as `setSearchParams` already updates the url, so you could just remove the whole function and use `query => setSearchParams({ query })` dierctly in L137 (or just keep the function but remove L89 and L90).

web/src/components/SearchListItem.tsx
```diff
@@ -63,24 +63,26 @@ type SearchListItemProps = {
   thumbnail: string | null
 }
 
-const SearchListItem = ({ title, contentWithoutHtml, query, path, thumbnail }: SearchListItemProps): ReactElement => {
-  const excerpt = getExcerpt(contentWithoutHtml, { query, maxChars: EXCERPT_MAX_CHARS })
+const SearchListItem = React.memo(
```
steffenkleinle, line 66: 🤔 Does it make sense to wrap this in a `React.memo()`? In what cases are these reused? The list item should change on every new char typed in the query as the highlighting also changes then.

steffenkleinle (commented):
Nice work!

web/src/components/SearchListItem.tsx
steffenkleinle: This is not correct. Using `memo` skips re-rendering when props are UNchanged, not when they are changed. From the [docs](https://react.dev/reference/react/memo#skipping-re-rendering-when-props-are-unchanged):
> React normally re-renders a component whenever its parent re-renders. With memo, you can create a component that React will not re-render when its parent re-renders so long as its new props are the same as the old props. Such a component is said to be memoized.

Since the props will basically change on every render as the query will change on nearly every render, I think this will not lead to an increased performance. Or am I missing something here?

bahaaTuffaha: I will see if I can do some sort of profiling to see with and without how it performs.

lunars97: I agree with Steffen. On every input change it will anyway change so...

steffenkleinle (approved):
Works as expected!

steffenkleinle (changes requested):
The search params are not correctly updated when the query is deleted, it stays as http://localhost:9000/muenchen/en/search?query=.

steffenkleinle:
One other thought: Perhaps it makes sense to use the debounced query for search highlight instead of the actual query to be in sync?

steffenkleinle (approved):

f1sh1918 (changes requested):
Unfortunately this break searching for location on web for me

image

f1sh1918 (approved):
Nice work! Tested on chrome and ios

bahaaTuffaha (approved):
LGTM ✅
I left the sub-branch that could minimize the lag I'm experiencing.

Issue #3135, opened 2025-03-04
Tooltip is not fully seen on language change hover


lunars97:
### Describe the Bug
Half of the tooltip text disappears under the language container when the user hovers over it in big view screens

### Steps to Reproduce

1. Go to http://localhost:9000/augsburg/de
2. Click on 'language change button'
3. Hover over the language change button
4. See the tooltip being half seen

### Expected Behavior
The tooltip should be above the language container to be fully seen on hover (or just not displayed if the language change dropdown is already open).

### Additional Information

![Image](https://github.com/user-attachments/assets/a7a532fe-ae74-4c09-85db-037dcc59c860)

PR #3140, opened 2025-03-05
3135: Fix tooltip being half hidden on language change hover

lunars97:
### Short Description

Tooltip is being half hidden when a user hovers over `Change Language` button in big width view. 

### Proposed Changes

- Set higher z-index to Tooltip

### Side Effects

- none

### Testing

Hover over `Change Language` button in web and see that the language selector container covers the half of the tooltip

### Resolved Issues

Fixes: #3135 

---


bahaaTuffaha (approved):

f1sh1918 (commented):
Looks quite good. just one hint

f1sh1918 (approved):

Issue #3084, opened 2025-02-13
iOS only: Error thrown if app is open in the background and link (e.g. from QR-Code) is tried to open.


FritjofK:
### Describe the Bug

If the app is still open in the background in another language and I then scan a QR code in another language and want to open it, the inappbrowser does not open and I get the error message. 
If I have closed the app and scan a code in another language, the inappbrowser opens. 

### Steps to Reproduce

Error was reported for **hallo aschaffenburg**

1. Open hallo aschaffenburg on iOS device  
2. Do not close the app but switch to camera and scan QR-code with different language than previously chosen in app
3. See error

### Expected Behavior

Steffen explained (https://chat.tuerantuer.org/digitalfabrik/pl/puf543enktb4jnxeyz1zi9nhsw): inappbrowser should open the requested page 

### Related Issues

None that I know of. 

### Additional Information

I (Fritjof) tested on: 
- iPhone 14
- iOS 18.1 
- hallo aschaffenburg 2025.1. 

Flyer with QR-codes from Aschaffenburg: [hallo aschaffenburg_flyer.pdf](https://github.com/user-attachments/files/18781653/hallo.aschaffenburg_flyer.pdf)

Error Screen: 
[hallo aschaffenburg_error message.pdf](https://github.com/user-attachments/files/18781721/hallo.aschaffenburg_error.message.pdf)

Screen Recording of error reproduction on iOS: 
https://github.com/user-attachments/assets/6ebfe673-e7de-49fe-b048-9485562412b8

steffenkleinle:
I can't reproduce this on a iOS simulator :/

f1sh1918:
I checked the issue on an **iPhoneSE ios 18.2.1** and tried to reproduce it with the following steps. Just to check that i didn't forget anything:

1. Open Aschaffenburg app and choose english language
2. Keep the app in background and switch to the camera app and open the qr code with the german link
3. InApp browser opens (which is the expected behavior) if a link in a different language is opened

I tried all the qr codes in the pdf on that page in different languages and couldn't reproduce the error.
If english the app opens in english.
Other languages inApp browser opens.

I would recommend to update to the latest ios version and then try to reproduce it again @FritjofK

FritjofK:
@f1sh1918 und @steffenkleinle 
I updated to 18.3.1 and still recieve the same error with the path explained above. 

I even met the error when closing the app. 
[...]

PR #3142, opened 2025-03-05
3084: InApp browser open link timeout

f1sh1918:
### Short Description

Currently on ios the InAppBrowser seem not to properly open a new link of an inAppBrowser is already running. The close function seem not to be sufficient.

### Proposed Changes

- add a timeout to wait until a new instance of the inapp browser will open

### Side Effects

- none

### Testing

- Only testable on ios real devices. Check the issue for further instructions
- I'm not 100% sure if the timeout time is sufficient so please test properly. I couldn't reproduce the error.

### Resolved Issues

Fixes: #3084

---


native/src/utils/openExternalUrl.ts
```diff
@@ -27,6 +28,10 @@ const openExternalUrl = async (rawUrl: string, showSnackbar: (snackbar: Snackbar
         },
       })
       InAppBrowser.close()
+      // On ios InAppBrowser seems to need some time to close the browser until it can be opened again properly
```
steffenkleinle, line 31: Nice, could you perhaps link the issue here? 
```suggestion
      // On ios InAppBrowser seems to need some time to close the browser until it can be opened again properly
      // https://github.com/digitalfabrik/integreat-app/issues/3084
```

steffenkleinle (approved):
Thanks for the fix! Not tested

LeandraH (approved):
I sometimes managed to reproduce the error on main and not on this branch, so yay? Tested on real iOS

Also, thank you for finding this interim solution!

PR #3145, opened 2025-03-06
Add hasProp helper to avoid ts error

steffenkleinle:
### Short Description

Add hasProp helper to avoid ts error due to typescript limitation.

### Proposed Changes

- Add `hasProp` to check for null/undefined props with ts acknowledging its result

### Side Effects

N/A

### Testing

Open the app, change a language and the city and check that everything still works.

### Resolved Issues

Fixes: N/A

---


shared/utils/index.ts
```diff
@@ -45,3 +45,8 @@ export const getCategoryTiles = ({
       isExternalUrl: externalOfferUrl !== undefined,
     })
   })
+
+export const hasProp = <P extends PropertyKey, O extends { [p in P]: unknown }>(
+  obj: O,
+  p: P,
+): obj is O & { [p in P]: NonNullable<unknown> } => obj[p] !== undefined && obj[p] !== null
```
LeandraH, line 52: 🙃  Can we call them object and prop? I don't think we really use abbreviations anywhere else, and that would make it marginally more readable, at least to me.

steffenkleinle:
> Hmm, I'm not getting an error when switching cities or languages. Where does this issue come from?

Yes, this is only about removing the typescript error which we suppressed with a `@ts-expect-error` so far. We introduced this in the entitlementcard at the same point, so I figured it makes sense to do this here. This was not a bug.

LeandraH:
> > Hmm, I'm not getting an error when switching cities or languages. Where does this issue come from?
> 
> Yes, this is only about removing the typescript error which we suppressed with a `@ts-expect-error` so far. We introduced this in the entitlementcard at the same point, so I figured it makes sense to do this here. This was not a bug, only a way to improve typesafety in our code.

Ah, nice, thanks for the explanation. So is there a `@ts-expect-error` that we can remove?

steffenkleinle:
> > > Hmm, I'm not getting an error when switching cities or languages. Where does this issue come from?
> > 
> > 
> > Yes, this is only about removing the typescript error which we suppressed with a `@ts-expect-error` so far. We introduced this in the entitlementcard at the same point, so I figured it makes sense to do this here. This was not a bug, only a way to improve typesafety in our code.
> 
> Ah, nice, thanks for the explanation. So is there a `@ts-expect-error` that we can remove?

See `AppContextProvider`, I already removed it:
image

LeandraH:
> > > > Hmm, I'm not getting an error when switching cities or languages. Where does this issue come from?
> > > 
> > > 
> > > Yes, this is only about removing the typescript error which we suppressed with a `@ts-expect-error` so far. We introduced this in the entitlementcard at the same point, so I figured it makes sense to do this here. This was not a bug, only a way to improve typesafety in our code.
> > 
> > 
> > Ah, nice, thanks for the explanation. So is there a `@ts-expect-error` that we can remove?
> 
> See `AppContextProvider`, I already removed it: image

Is that in a different PR? Because I don't see this change on this PR, and [the line is still there on main](https://github.com/digitalfabrik/integreat-app/blob/main/native/src/contexts/AppContextProvider.tsx#L103). Or is there a commit that you haven't pushed yet?

LeandraH (approved):
Sorry, my bad

bahaaTuffaha (approved):
LGTM .. Tested on Android Emulation (changed language + changed the area also changed the system's language)

Issue #3132, opened 2025-02-27
Misaligned icons/checkbox


hauf-toni:
### Describe the Bug

Some elements appear misaligned, resulting in an uneven layout.

### Steps to Reproduce

1. Go to the “Orte filtern” in the app.
2. Observe the first row.
3. Notice that the icon and checkbox are misaligned.

### Expected Behavior

Aligned, neat layout.

### Related Issues

Related to: #3131 

### Additional Information

See screenshot that shows icons that are mispositioned in the first row:

<img width="328" alt="Image" src="https://github.com/user-attachments/assets/def6cad1-48af-4bd0-81ca-15737cd77a31" />

red-apple38:
Can I take this issue?

steffenkleinle:
> Can I take this issue?

Sure, go ahead and take it :) Please make sure to have a look at our [conventions](https://github.com/digitalfabrik/integreat-app/blob/main/docs/conventions.md) before you start :) Let us know if you need anything!

red-apple38:
Hey I opened the app on the landing page you have to select a location, I didn't find the option to filter locations there (Appendix 1). When I selected a specific location, this option was also not available (Appendix 2). I did not find this option in the web version either.
Can you perhaps help me where I need to navigate to reproduce the error? 

![Image](https://github.com/user-attachments/assets/5caa0901-d32f-4b80-974d-1865c676aea6)
![Image](https://github.com/user-attachments/assets/20c89315-5a69-4bdb-be7d-4a3df039151d)

steffenkleinle:
> Hey I opened the app on the landing page you have to select a location, I didn't find the option to filter locations there (Appendix 1). When I selected a specific location, this option was also not available (Appendix 2). I did not find this option in the web version either. Can you perhaps help me where I need to navigate to reproduce the error?

I'd say you are pretty close already. The overlay can be found if you hit `Map` > `Filter locations`. However, it seems like the issue only occurs on web, so the place to reproduce this would be here: https://webnext.integreat.app/testumgebung/en/locations and then `Filter locations`. Sorry for the confusion.

Let us know if we can help you any further :)

red-apple38:
Hello :), 
[...]

PR #3147, opened 2025-03-07
3132: Fix misalignment of icons and checkboxes in Orte filtern

red-apple38:
### Short Description

This PR fixes the the appearance of icons and checkboxes in the "Orte filtern" section by adjusting the Icon size of the clock and Checkbox Size for desktop appearance.

### Proposed Changes

Modified the Row styled component in PoiFilters.tsx to use gap between the checkbox and the Icon.
Added align-items: center and a min-height to maintain consistent vertical alignment and prevent layout shifts.
Increased the Icon Size for a better appearance in desktop and web.

### Side Effects

none

### Testing

Verify that the icons and checkboxes in the "Orte filtern" section are properly aligned, starting from the left and have proper gap in desktop version.
Test the layout responsiveness by resizing the browser window to ensure the grid adjusts correctly with varying numbers of columns.

Check that the alignment remains consistent across different screen sizes (desktop, tablet, mobile).

### Resolved Issues

Fixes: #3132

---

### DIFF
--- a/web/src/components/PoiFilters.tsx
+++ b/web/src/components/PoiFilters.tsx
@@ -38,7 +38,7 @@ const Section = styled.div`
 const Row = styled.div`
   display: flex;
   gap: 8px;
-  align-items: center
+  align-items: center;
 `
 
 const SortingHint = styled.div`
--- a/web/src/components/PoiFilters.tsx
+++ b/web/src/components/PoiFilters.tsx
@@ -37,6 +37,8 @@ const Section = styled.div`
 
 const Row = styled.div`
   display: flex;
+  gap: 10px;
+  align-items: center
 `
 
 const SortingHint = styled.div`
@@ -60,8 +62,17 @@ const StyledButton = styled(TextButton)`
 `
 
 const StyledIcon = styled(Icon)`
-  width: 24px;
-  height: 24px;
+  width: 30px;
+  height: 30px;
+`
+
+const StyledCheckboxContainer = styled.div`
+  
+    @media (width >= 768px){
+        input[type="checkbox"] {
+            transform: scale(1.3); 
+        }
+    }
 `
 
 type PoiFiltersProps = {
@@ -94,12 +105,14 @@ const PoiFilters = ({
           <SubTitle>{t('openingHours')}</SubTitle>
           <Row>
             <StyledIcon src={ClockIcon} />
+            <StyledCheckboxContainer>
             <Checkbox
               id='poi-filters-currently-opened'
               checked={currentlyOpenFilter}
               setChecked={setCurrentlyOpenFilter}
               label={t('onlyCurrentlyOpen')}
             />
+            </StyledCheckboxContainer>
           </Row>
         </Section>
         <Section>
--- a/web/src/components/PoiFilters.tsx
+++ b/web/src/components/PoiFilters.tsx
@@ -99,7 +99,7 @@ const PoiFilters = ({
           <SubTitle>{t('openingHours')}</SubTitle>
           <Row>
             <StyledIcon src={ClockIcon} />
-            <Checkbox
+            <StyledCheckbox
               id='poi-filters-currently-opened'
               checked={currentlyOpenFilter}
               setChecked={setCurrentlyOpenFilter}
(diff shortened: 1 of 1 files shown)


web/src/components/PoiFilters.tsx
```diff
@@ -60,8 +62,17 @@ const StyledButton = styled(TextButton)`
 `
 
 const StyledIcon = styled(Icon)`
-  width: 24px;
-  height: 24px;
+  width: 30px;
+  height: 30px;
```
steffenkleinle, line 66: :wrench: Am I assuming correctly that you changed this to fix the issue that the width of the checkbox is smaller than actually set? I think it would be better to set `min-width` instead:
```suggestion
  min-width: 24px;
```

web/src/components/PoiFilters.tsx
```diff
@@ -60,8 +62,17 @@ const StyledButton = styled(TextButton)`
...
+`
+
+const StyledCheckboxContainer = styled.div`
+  
+    @media (width >= 768px){
+        input[type="checkbox"] {
+            transform: scale(1.3); 
+        }
+    }
 `
```
steffenkleinle, line 76: :wrench: Instead of wrapping the checkbox in an extra div, you could just pass the `className` prop to `StyledCheckbox` in [Checkbox.tsx](https://github.com/digitalfabrik/integreat-app/blob/8525cfc1526ffad4c0c381cb2ca43be6073c9437/web/src/components/base/Checkbox.tsx#L33), for more information see the [styled-components docs](https://styled-components.com/docs/basics#styling-any-component).

See an example here: https://github.com/digitalfabrik/integreat-app/blob/main/web/src/components/LastUpdateInfo.tsx

Then you can just do:
```suggestion
const StyledCheckbox = styled(Checkbox)`
  min-width: 20px;
`
```

steffenkleinle (commented):
Great work and congrats to your first contribution here :tada: Thank you :)
Tested on firefox and looks as great as expected, just some minor stylistic changes.

Another note: At the moment the checks are still failing, please make sure to run `yarn prettier:write` before commiting.

web/src/components/PoiFilters.tsx
red-apple38: The size of the checkbox was increased from 24px to 30px so that it appears larger as suggested in the issue. In the mobile view, the checkbox also looked better to me, so I just increased it. 
The problem with the checkbox could not be reproduced by you, so I think it might just be a problem on my device :)

web/src/components/PoiFilters.tsx
```diff
@@ -37,6 +37,8 @@ const Section = styled.div`
 
 const Row = styled.div`
   display: flex;
+  gap: 10px;
```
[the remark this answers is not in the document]
red-apple38: Thanks, this is good to know

web/src/components/PoiFilters.tsx
steffenkleinle: Thank you! I guess the instructions where not clear and I previously did not get the underlying issue why it looked the way it did. Thanks for your patience here :)

web/src/components/PoiFilters.tsx
```diff
@@ -60,8 +62,11 @@ const StyledButton = styled(TextButton)`
...
 
 const StyledIcon = styled(Icon)`
-  width: 24px;
-  height: 24px;
+  min-width: 24px;
+`
+
+const StyledCheckbox = styled(Checkbox)`
+  min-width: 20px;
 `
```
steffenkleinle, line 70: :wrench: Sadly this does not take effect yet because
a) No classname is passed in `Checkbox.tsx`
b) The actual checkbox (i.e. the `<input>` is wrapped in another div

Therefore, I think the best choice would be to remove this style here and add `min-width: 20px` after https://github.com/digitalfabrik/integreat-app/blob/7767605ed89e6a336c508fd956b4d751ff42518f/web/src/components/base/Checkbox.tsx#L14.
This should also fix the issue you describe in #3132.

steffenkleinle:
@red-apple38 are you still working on this or should we finish this? Thanks for your contribution!

steffenkleinle:
Closing in favor of #3193

Issue #3104, opened 2025-02-18
Make BottomSheet Slider in Pois accessible for screen readers


lunars97:
### Describe the Problem
Make the BottomSheet Slider in Pois accessible to screen readers

### Describe Your Preferred Solution
 
Add announcement to the main component when focused by adding `accessibilityHint` with sth like `Locations near you` (currently it is announcing `BottomSheet Slider`) and add announcement on how many items are in the list. Add if necessary additional enhancements to make the experience with BottomSheet Slider for screen readers better. See the related pictures below.

### Additional Information

![Image](https://github.com/user-attachments/assets/3b7b97b0-0d60-4736-a52d-3dcecc0e974a)
![Image](https://github.com/user-attachments/assets/8777bd7a-1952-4567-a3ce-bb67ec664f68)

PR #3153, opened 2025-03-14
3104: Improve bottomSheet accessibility

bahaaTuffaha:
### Short Description
Make the BottomSheet Slider in Pois accessible to screen readers

Add announcement to the main component when focused by adding accessibilityHint with sth like Locations near you (currently it is announcing BottomSheet Slider) and add announcement on how many items are in the list.

### Proposed Changes

Noticable problems:
First of all I can't change the "BottomSheet Slider" accessibility label from `PoisBottomSheet`. I had an idea of using patch-package to change this label but it could be bad idea for reusability.

The other difficulty is trying to make the bottom-sheet easier to be selected by the screen reader because it's tricky to select the bottom-sheet by tapping (the selection will go to the map behind) at android so I found out the only way is to disable the map's accessibility by using `importantForAccessibility='no'`.

- Added an accessibilityLabel to announce number of locations near the user.
- Included translations.

Note: not tested on iOS.

### Side Effects

- none

### Testing

- Enable talkback.
- Go to pois/map and check the list.
(while using talkback swipe left/right to reach the list if it's hard to tap it)

### Resolved Issues

Fixes: #3104 

---


native/src/components/PoisBottomSheet.tsx
```diff
@@ -92,6 +92,8 @@ const PoisBottomSheet = ({
       enableContentPanningGesture={enableContentPanningGesture}
       enableDynamicSizing={false}
       animateOnMount
+      accessibilityRole={'list'}
+      accessibilityLabel={t('nearYou', { count: pois.length, nearYou: t('listTitle') })}
```
steffenkleinle, line 96: :wrench: Since the bottom sheet might contain either a list or a single poi, this accessibility role and label is only correct in the former.

translations/translations.json
```diff
@@ -10158,7 +10158,13 @@
       "makeAppointment": "Termin vereinbaren",
       "appointmentNecessary": "Termin erforderlich",
       "makeAppointmentTooltipWithLink": "Vereinbare an diesem Tag einen Termin über <1>die Webseite</1> oder telefonisch.",
-      "theWebsite": "die Webseite"
+      "theWebsite": "die Webseite",
+      "nearYou_zero": "0 Standorte {{nearYou}}",
```
steffenkleinle, line 10162: :wrench: The naming here should be `Orte` instead of `Standorte`. Furthermore, using `In deiner Nähe` like this does sadly not really work as `In` is uppercase while it would need to be lowercase. Accordingly, please just duplicate the translation.

bahaaTuffaha: I'm not using it here as a text. Its only for talkback to be heard.

steffenkleinle: This might still alter how things are spoken. Furthermore, this might create completely grammatically incorrect sentences (due to word order, different translations, ...). Therefore we really don't want to combine translations in a hacky way like this.

steffenkleinle:
> The other difficulty is trying to make the bottom-sheet easier to be selected by the screen reader because it's tricky to select the bottom-sheet by tapping (the selection will go to the map behind) at android so I found out the only way is to disable the map's accessibility by using importantForAccessibility='no'.

I think this would be okay as we explicitly exclude the map from being accessible in our accessibility declaration as maps are just by design not really accessible. So feel free to implement this.

translations/translations.json
```diff
@@ -10159,12 +10159,12 @@
...
-      "nearYou_two": "{{count}} Standorte {{nearYou}}",
-      "nearYou_few": "{{count}} Standorte {{nearYou}}",
-      "nearYou_many": "{{count}} Standorte {{nearYou}}",
-      "nearYou_other": "{{count}} Standorte {{nearYou}}"
+      "nearYou_zero": "0 Orte",
+      "nearYou_one": "1 Ort",
+      "nearYou_two": "{{count}} Orte",
+      "nearYou_few": "{{count}} Orte",
+      "nearYou_many": "{{count}} Orte",
+      "nearYou_other": "{{count}} Orte"
```
steffenkleinle, line 10167: :wrench: The name does not really fit anymore:
```suggestion
      "poisCount_zero": "0 Orte",
      "poisCount_one": "1 Ort",
      "poisCount_two": "{{count}} Orte",
      "poisCount_few": "{{count}} Orte",
      "poisCount_many": "{{count}} Orte",
      "poisCount_other": "{{count}} Orte"
```

native/src/components/PoisBottomSheet.tsx
```diff
@@ -86,24 +86,26 @@ const PoisBottomSheet = ({
 
   return (
     <StyledBottomSheet
+      accessibilityLabel=''
```
steffenkleinle, line 89: :thinking: Is it necessary to set this to an empty string?

native/src/components/PoisBottomSheet.tsx
```diff
@@ -86,24 +86,26 @@ const PoisBottomSheet = ({
...
       enableDynamicSizing={false}
       animateOnMount
-      accessibilityRole={'list'}
-      accessibilityLabel={t('nearYou', { count: pois.length, nearYou: t('listTitle') })}
       handleComponent={BottomSheetHandle}
       onChange={setSnapPointIndex}>
       <BottomSheetContent>
         {slug ? (
-          <BottomSheetScrollView showsVerticalScrollIndicator={false}>{PoiDetail}</BottomSheetScrollView>
+          <BottomSheetScrollView accessibilityLabel={t('detailsInformation')} showsVerticalScrollIndicator={false}>
```
steffenkleinle, line 100: :thinking: Not sure whether it would make more sense to set this to the title of the poi instead of `Information`? What do you think?

steffenkleinle: Doesn't it make more sense to not have a label/have an empty label here then?

native/src/components/PoiDetails.tsx
```diff
@@ -55,7 +55,7 @@ const PoiDetails = ({ poi, language, distance }: PoiDetailsProps): ReactElement
     appointmentUrl ?? poi.contacts.find(contact => contact.website !== null)?.website ?? null
 
   return (
-    <PoiDetailsContainer>
+    <PoiDetailsContainer accessibilityLabel={`${title} - ${category.name}`}>
```
steffenkleinle, line 58: :thinking: Could you explain why you decided to add this as a a11y label instead of an empty string?
Isn't this just a duplicate of the two strings set below which are read in any case?

native/src/components/AddressInfo.tsx
```diff
@@ -48,7 +48,7 @@ const AddressInfo = ({ location, language }: AddressInfoProps): ReactElement =>
 
   return (
     <Container language={language}>
-      <Pressable role='button' onPress={copyLocationToClipboard}>
+      <Pressable accessibilityLabel={t('cityNotCooperating:copyText')} role='button' onPress={copyLocationToClipboard}>
```
steffenkleinle, line 51: :wrench: I think `Copy text` is not a good a11y label here, I think it should explicitly say `Copy address`. What do you think?

native/src/components/PoiDetails.tsx
bahaaTuffaha: what I found is it will skip the headers that in the Poi Details.. did you give it a test? not sure how it will work for iOS though...

native/src/components/AddressInfo.tsx
bahaaTuffaha: I think this depends on whether we want to add more translations.

steffenkleinle: Adding new translations is totally fine if it makes sense. I think in this case it is better to use a new and correct translation over a more or less incorrect (or at least imprecise) one.

native/src/components/PoiDetails.tsx
steffenkleinle: I sadly can't test this atm as my phone is google free and my emulators don't run atm :exploding_head: Sorry :/
So just to confirm: What exactly is skipped? Only the title? Or also the category? Is the rest of the content read correctly? And do you have any clue why that happens?

bahaaTuffaha: Here's an example of with this line and without it:
https://github.com/user-attachments/assets/27a1cf54-e739-43a6-a537-3d401e88ca9b

steffenkleinle: Thanks, so the change is basically to automatically read this once a poi is selected? Sounds good to me.

steffenkleinle (approved):
Not tested

native/src/components/PoiDetails.tsx
bahaaTuffaha: @LeandraH Can you test how it behaves on iOS please?

LeandraH: Sorry, it's too annoying to switch branches from the update branch, I'm not doing any code reviews at the moment. But I think @f1sh1918 also has an iPhone?

translations/translations.json
```diff
@@ -10158,7 +10159,13 @@
       "makeAppointment": "Termin vereinbaren",
       "appointmentNecessary": "Termin erforderlich",
       "makeAppointmentTooltipWithLink": "Vereinbare an diesem Tag einen Termin über <1>die Webseite</1> oder telefonisch.",
-      "theWebsite": "die Webseite"
+      "theWebsite": "die Webseite",
+      "poisCount_zero": "0 Orte",
+      "poisCount_one": "1 Ort",
+      "poisCount_two": "{{count}} Orte",
```
f1sh1918, line 10165: why do we need all the keys? Wouldn't one for singular and one for plural be enough and just pass the `count`

lunars97 (commented):
Thank you very much for this PR! Everything works nice except for one thing: See the comment with suggestion below. Plus: I think it would be much clearer for the user if we had the title of the PoisBottomSheet as `Locations near you` instead of just `Near you`. What would you say @hauf-toni ? :thinking:

lunars97 (changes requested):
See the comment above please :yellow_heart:

native/src/components/MapView.tsx
```diff
@@ -174,6 +174,8 @@ const MapView = ({
   return (
     <MapContainer>
       <StyledMap
+        importantForAccessibility='no'
```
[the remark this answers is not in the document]
bahaaTuffaha: I think its a great idea to hide them while full screen 👍️

bahaaTuffaha:
> Thank you very much for this PR! Everything works nice except for one thing: See the comment with suggestion below. Plus: I think it would be much clearer for the user if we had the title of the PoisBottomSheet as `Locations near you` instead of just `Near you`. What would you say @hauf-toni ? 🤔

I went for `"poisCount_one": "1 location"` for ex. to minimize words that needs translation.

lunars97 (approved):
LGTM :+1: Tested on Android Emulator.

Issue #3116, opened 2025-02-20
[Dev] @turf/distance/node_modules/tslib node module can't be installed


f1sh1918:
### Describe the Bug

This is a dev only bug!
The node module can't be installed if its already existing.

### Steps to Reproduce

1. Node modules are already installed
2. run `yarn install`
3. See error

### Expected Behavior

The node module can be installed without issues

### Additional Information

```
An unexpected error occurred: "ENOENT: no such file or directory, lstat '/Users/afischer/projects/integreat-app/shared/node_modules/@turf/distance/node_modules/tslib'".
```

steffenkleinle:
Perhaps this gets fixed by updating our dependencies already. Will do with #3097.

Issue #2932, opened 2024-09-18
Upgrade react-native to v0.75 and react to 18.3.1


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
A new react-native version got released. See the upgrade helper for necessary things to do:
https://react-native-community.github.io/upgrade-helper/?from=0.74.5&to=0.75.3&package=tuerantuer.app.integreat&name=Integreat

Also upgrade react to 18.3.1.
An upgrade of react-native-static-server is necessary for the react-native upgrade. Since a recent release it also supports older android sdks.

**Testing information**
Please test everything (the whole app) very thoroughly!

**Additional context**
:warning: Check previous upgrade PRs about possible issues.

Issue for the upgrade to v0.74: https://github.com/digitalfabrik/integreat-app/issues/2786
PR for the upgrade to v0.74: https://github.com/digitalfabrik/integreat-app/pull/2907/files
Last dependency upgrade PR: https://github.com/digitalfabrik/integreat-app/pull/2912

LeandraH:
@steffenkleinle I am stealing this PR from you because I am doing that upgrade as part of hopefully getting the newer version of react-native-static-server running. The RN upgrade might still not fix my PR but then I'll still open a separate PR for this upgrade :)

PR #3158, opened 2025-03-18
2932: Update to React-Native 75

LeandraH:
### Short Description

This PR upgrades React-Native to 0.75, and React to 18.3.1

### Proposed Changes

- Upgraded RN to 0.75
- Upgraded React to 18.3.1
- Downgraded styled components from version 6 to version 5 😢 
- Fixed some typing problems
- Deleted the endlessly growing list of duplicate HEADER_SEARCH_PATHS in the project file

### Side Effects

Everything possible

### Testing

Anything could break from the upgrade
Please also check if the error in #3116 disappears for you by going to a different branch and running `yarn install` there, it did for me.

### Resolved Issues

Fixes: #2932 
Possibly fixes: #3116 

---


native/ios/Integreat.xcodeproj/project.pbxproj
```diff
@@ -617,46 +617,6 @@
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-NativeModulesApple/React_NativeModulesApple.framework/Headers",
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-graphics/React_graphics.framework/Headers",
 					" ${PODS_CONFIGURATION_BUILD_DIR}/React-graphics/React_graphics.framework/Headers/react/renderer/graphics/platform/ios",
-					" ${PODS_CONFIGURATION_BUILD_DIR}/ReactCommon/ReactCommon.framework/Headers",
```
LeandraH, line 620: I'm not sure why Xcode regularly adds more repeats of these lines but it works without them (at least on my machine :D)

native/src/components/CalendarRangeModal.tsx
```diff
@@ -20,7 +20,7 @@ const DatePickerWrapper = styled.View`
 
 const StyledView = styled.View`
   gap: 8px;
-  justify-content: ${props => (props.theme.contentDirection === 'rtl' ? 'flex-start' : 'flex-end')};
+  justify-content: flex-end;
```
LeandraH, line 23: Not only do we not have the `contentDirection` prop in the `native` part, RN is also smart enough to already flip large parts of the displays in rtl languages.

steffenkleinle:
Could you add some information on why the downgrade of styled-components was necessary?

native/src/components/CalendarRangeModal.tsx
bahaaTuffaha: Yep .. no need for contentDirection here.

LeandraH:
> Could you please add some information on why the downgrade of styled-components was necessary?

It was your advice from [this Mattermost thread](https://chat.tuerantuer.org/digitalfabrik/pl/qzfn1zynzid7frs5gp6kfjom4r) 😁 

Basically, the combination of RN 75 and Styled-components 6 led to the type error in the native project that functions inside a styled component didn't have access to the theme type, e.g. `background-color: ${props => props.theme.colors.backgroundColor}` leading to the error `Parameter 'props' implicitly has an 'any' type`.

bahaaTuffaha:
> Basically, the combination of RN 75 and Styled-components 6 led to the type error in the native project that functions inside a styled component didn't have access to the theme type, e.g. `background-color: ${props => props.theme.colors.backgroundColor}` leading to the error `Parameter 'props' implicitly has an 'any' type`. I'll add that to the description as well, thanks!

I think there is a way to use theme in V6:
- https://github.com/styled-components/styled-components/issues/4074#issuecomment-1870027194
- Another way to do it is passing `$theme` as prop.

But it needs a lot of changes 😐️.

bahaaTuffaha (changes requested):
Works great on android api 35 ...tested most features.
but on older api (29) it doesn't show content :
Screenshot_1742475772

I compared with main branch the content is working fine.

LeandraH:
> > Basically, the combination of RN 75 and Styled-components 6 led to the type error in the native project that functions inside a styled component didn't have access to the theme type, e.g. `background-color: ${props => props.theme.colors.backgroundColor}` leading to the error `Parameter 'props' implicitly has an 'any' type`. I'll add that to the description as well, thanks!
> 
> I think there is a way to use theme in V6:
> 
>     * [Unable to style react-select component with styled-components v6 styled-components/styled-components#4074 (comment)](https://github.com/styled-components/styled-components/issues/4074#issuecomment-1870027194)
This thread is about a specific case, using [react-select](https://react-select.com/home) with styled components, which is unfortunately not what the problem here is :(
> 
>     * Another way to do it is passing `$theme` as prop.
> 
> 
> But it needs a lot of changes 😐️.

Yeah, that would be about 250 changes, for something that should be there already, especially considering that the code works in web, just not in native.

Buuuuuut, it turns out that the [latest styled-components version](https://github.com/styled-components/styled-components/releases/tag/styled-components%406.1.16) that was released three days ago fixes this problem 🥳

LeandraH:
> Works great on android api 35 ...tested most features. but on older api (29) it doesn't show content : 
> 
> I compared with main branch the content is working fine.

I obviously can't reproduce it 😭  I set up a fancy new emulator with Android 29, and I get all the content. I even tried switching the phone's language to Arabic, in case the rtl changes broke something. Do you have any more hints on how to reproduce it? Do you get any errors in the terminal?

bahaaTuffaha:
> > Works great on android api 35 ...tested most features. but on older api (29) it doesn't show content :
> > I compared with main branch the content is working fine.
> 
> I obviously can't reproduce it 😭 I set up a fancy new emulator with Android 29, and I get all the content. I even tried switching the phone's language to Arabic, in case the rtl changes broke something. Do you have any more hints on how to reproduce the missing content? Do you get any errors in the terminal?

I even changed the screen size to pixel7 pro this time and still no content: 
Not sure if its related:

Screenshot_1742492905

Screenshot_1742492844

image
Does 86x matters cuz I searched for a 64x version but there is non.

LeandraH:
I've managed to reproduce this by now. The problem seems to be in the RemoteContent, so I'm trying to fix it by upgrading the packages from Dr Pogodin (including @dr.pogodin/react-native-webview)

native/metro.config.js
```diff
@@ -2,8 +2,9 @@
 const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config')
 const path = require('path')
 
-/** @type {import('metro-config').MetroConfig} */
+/** @type {import('@react-native/metro-config').MetroConfig} */
 const defaultConfig = getDefaultConfig(__dirname)
+defaultConfig.resolver.unstable_enablePackageExports = true
```
LeandraH, line 7: This is needed for `react-navigation-header-buttons`, and it will be enabled by default in a future version of `react-native`

https://github.com/vonovak/react-navigation-header-buttons/blob/master/INSTALL.md#installation--setup

native/src/components/CustomHeaderButtons.tsx
LeandraH, line 1: I had to add color to the Icon in order to hide search and language icon on e.g. settings page.

We previously used `opacity` to hide the icons but `style` is flattened, therefore an immutable object. When tapping on an icon, an animation starts to make the tapped icon partially transparent in [@react-navigation/elements](https://github.com/react-navigation/react-navigation/blob/09d65bcdc908883c8cdd1b93fde19cec4b799ad6/packages/react-native-tab-view/src/PlatformPressable.tsx#L82). With tighter controls on the new RN version, tapping on the icon and thus changing the opacity on the immutable style object, the app crashed.

I worked around that by passing in the `color` prop separately from the `style`.

native/src/components/EventsDateFilter.tsx
LeandraH, line 1: The global `JSX` namespace is being removed, and we use `ReactElement` almost everywhere https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript

native/src/components/EventsDateFilter.tsx
```diff
@@ -60,7 +60,7 @@ const EventsDateFilter = ({
   setEndDate,
   modalOpen,
   setModalOpen,
-}: EventsDateFilterProps): JSX.Element => {
+}: EventsDateFilterProps): ReactElement => {
```
LeandraH, line 63: The global `JSX` namespace is being removed, and we use `ReactElement` almost everywhere https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript

native/src/components/Header.tsx
```diff
@@ -126,7 +127,7 @@ const Header = ({
       title={t(title)}
       iconName={iconName}
       onPress={visible ? onPress : () => undefined}
-      style={{ opacity: visible ? 1 : 0 }}
+      color={visible ? theme.colors.textColor : 'transparent'}
```
LeandraH, line 130: The `style` prop gets flattened, and becomes therefore an immutable object. When tapping on an icon, an animation starts to make the tapped icon partially transparent in [@react-navigation/elements](https://github.com/react-navigation/react-navigation/blob/09d65bcdc908883c8cdd1b93fde19cec4b799ad6/packages/react-native-tab-view/src/PlatformPressable.tsx#L82). With tighter controls on the new RN version, tapping on the icon and thus changing the opacity on the immutable style object, the app crashed.

I worked around that by passing in the `color` prop separately from the `style`.

native/src/components/MapView.tsx
LeandraH, line 1: If anybody wants to read along: https://maplibre.org/maplibre-react-native/docs/setup/migrations/v10/

native/src/components/PinchPanImage.tsx
LeandraH, line 1: I rewrote this completely, and in the process removed the bounds because I found that they make it harder to zoom in on details, any objections?

native/src/components/MapView.tsx
```diff
@@ -28,16 +39,13 @@ import MapAttribution from './MapsAttribution'
...
-MapLibreGL.setAccessToken(null)
-
 const MapContainer = styled.View`
   flex: 1;
   flex-direction: row;
   justify-content: center;
 `
 
-const StyledMap = styled(MapLibreGL.MapView)`
+const StyledMap = styled(MLMapView)`
```
steffenkleinle, line 48: 🙃 I'd prefer `MapLibreMapView` as `ML` is usually machine learning... which really confused me for a bit. Generally we try to avoid abbreviations.
```suggestion
const StyledMap = styled(MLMapView)`
```

native/src/components/RemoteContent.tsx
```diff
@@ -167,7 +167,6 @@ const RemoteContent = ({
       }}
       originWhitelist={['*']} // Needed by iOS to load the initial html
       javaScriptEnabled
-      dataDetectorTypes='none'
```
steffenkleinle, line 170: 🤔 This causes a change in behavior as per default phone numbers are detected, was this a change on purpose? https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#datadetectortypes

shared/constants/maps.ts
```diff
@@ -82,7 +82,7 @@ export const mapConfig: MapConfigProps = {
 export const defaultViewportConfig: MapViewViewport = {
   latitude: 48.366512,
   longitude: 10.894446,
-  zoom: 8,
+  zoom: 10,
```
steffenkleinle, line 85: 🤔 Why did you change this?

web/package.json
```diff
@@ -68,11 +68,12 @@
...
     "@types/dompurify": "^3.0.5",
     "@types/luxon": "^3.4.2",
-    "@types/react": "^18.2.79",
-    "@types/react-dom": "^18.2.25",
+    "@types/react": "19.0.0",
+    "@types/react-dom": "19.0.0",
     "@types/react-helmet": "^6.1.11",
     "@types/react-highlight-words": "^0.20.0",
     "@types/react-infinite-scroller": "^1.2.5",
+    "@types/styled-components": "5.1.34",
```
steffenkleinle, line 76: 🔧 This should not be necessary
```suggestion
```

web/package.json
```diff
@@ -31,23 +31,23 @@
   "dependencies": {
```
steffenkleinle, line 31: ❓ Did you only upgrade native dependencies (and those you had to in other workspaces) or basically all dependencies? Just asking whether we need separate issues to upgrade the other dependencies.

shared/package.json
```diff
@@ -13,28 +13,29 @@
...
     "sentencex": "^0.4.2"
   },
   "devDependencies": {
-    "@types/geojson": "^7946.0.10",
-    "@types/luxon": "^3.3.1",
-    "@types/md5": "^2.3.2",
-    "@types/normalize-path": "^3.0.0",
-    "@types/react": "^18.2.79",
-    "@types/seedrandom": "^3.0.5",
+    "@babel/plugin-transform-export-namespace-from": "7.25.9",
```
steffenkleinle, line 32: 🤔 Is this really necessary here?

native/src/routes/Pois.tsx
```diff
@@ -146,7 +146,7 @@ const Pois = ({ pois: allPois, cityModel, route, navigation }: PoisProps): React
       />
       <MapView
         selectFeature={selectMapFeature}
-        boundingBox={cityModel.boundingBox}
+        cityCoordinates={[cityModel.longitude, cityModel.latitude]}
```
steffenkleinle, line 149: 🤔 Why did you change this?

native/src/components/Header.tsx
steffenkleinle: 🔧 Please add a comment explaining this to the code :)

native/src/components/MapView.tsx
```diff
@@ -85,35 +93,30 @@ const MapView = ({
...
   const coordinates = selectedFeature?.geometry.coordinates
   const defaultZoom = coordinates ? normalDetailZoom : defaultViewportConfig.zoom
-  const defaultSettings: CameraSettings = {
+
+  const [cameraSettings, setCameraSettings] = useState({
     zoomLevel: zoom ?? defaultZoom,
-    centerCoordinate: coordinates,
-    bounds: coordinates ? undefined : bounds,
-  }
+    centerCoordinate: coordinates ?? cityCoordinates,
```
steffenkleinle, line 108: ❓ Changing from a bounding box to just center coordinates might lead to problems with zooming as there are regions with different sizes. Are bounding boxes not supported anymore?

native/src/components/MapView.tsx
```diff
@@ -85,35 +93,30 @@ const MapView = ({
   bottomSheetFullscreen,
   zoom,
 }: MapViewProps): ReactElement => {
-  const cameraRef = useRef<MapLibreGL.Camera>(null)
-  const mapRef = useRef<MapLibreGL.MapView>(null)
+  const cameraRef = useRef<CameraRef>(null)
```
steffenkleinle, line 96: 🔧 This is not actually used anymore and can be removed, right?

native/src/components/MapView.tsx
```diff
@@ -172,29 +175,28 @@ const MapView = ({
...
+        <ShapeSource id='location-pois' shape={embedInCollection(features)} cluster clusterRadius={clusterRadius}>
+          <CircleLayer {...clusterLayer(theme)} />
+          <SymbolLayer {...clusterCountLayer} />
+          <SymbolLayer {...markerLayer(selectedFeature)} />
+        </ShapeSource>
+        <Camera
+          {...cameraSettings}
+          followUserMode={UserTrackingMode.Follow}
+          ref={cameraRef}
+          animationDuration={2000}
```
steffenkleinle, line 196: 🔧 
```suggestion
          animationDuration={animationDuration}
```

native/src/components/PinchPanImage.tsx
```diff
@@ -1,338 +1,84 @@
...
-          this.translateY.setValue(0)
-        }
-      }
-      this.setState(state => ({ ...state, interactive: true }))
+  const pinchGesture = Gesture.Pinch()
+    .onStart(() => {
+      prevScale.value = scale.value
+    })
+    .onUpdate(e => {
+      scale.value = prevScale.value * e.scale
```
steffenkleinle, line 45: 🙃 Please avoid abbreviations here, it is not clear if `e` stands for `event` or `error` here. Also, you could get rid of the brackets.

native/src/components/PinchPanImage.tsx
```diff
@@ -1,338 +1,84 @@
...
-  render(): ReactElement | null {
-    const { uri } = this.props
-    const { imageDimensions, viewDimensions, interactive } = this.state
+  const panGesture = Gesture.Pan()
+    .minDistance(MIN_DISTANCE)
+    .onStart(() => {
+      prevTranslationX.value = translationX.value
+      prevTranslationY.value = translationY.value
+    })
+    .onUpdate(e => {
```
steffenkleinle, line 54: 🙃 Same as above.

native/src/components/base/Icon.tsx
```diff
@@ -13,6 +13,7 @@ type IconProps = {
   directionDependent?: boolean
   reverse?: boolean
   style?: { width?: number; height?: number; color?: ColorValue }
+  color?: ColorValue
```
steffenkleinle, line 16: 🤔 I don't think this is used anywhere, is it?
```suggestion
```

steffenkleinle (commented):
Terrific job, looks great! Not tested yet, will wait until comments are addressed.

native/src/components/Header.tsx
LeandraH: Sorry, I phrased it poorly, this is not a workaround but what we should have been doing for a long time. RN only started punishing us for it now.

native/package.json
```diff
@@ -44,83 +44,87 @@
     "circleci:update-config": "yarn --cwd .. circleci:update-config"
   },
   "dependencies": {
-    "@dr.pogodin/react-native-static-server": "^0.5.5",
-    "@dr.pogodin/react-native-webview": "^13.13.0",
-    "@gorhom/bottom-sheet": "^5.0.6",
-    "@maplibre/maplibre-react-native": "^9.1.0",
+    "@dr.pogodin/react-native-fs": "2.32.1",
+    "@dr.pogodin/react-native-static-server": "0.20.1",
+    "@dr.pogodin/react-native-webview": "13.15.7",
```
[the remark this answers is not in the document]
LeandraH: Yes, the maintainer updates `react-native` versions in minor upgrades (which is what `react-native` does too, I guess). But I think we can do patch releases.

native/src/@types/react-native-static-server.d.ts
```diff
@@ -2,13 +2,14 @@ declare module '@dr.pogodin/react-native-static-server' {
   type Options = {
```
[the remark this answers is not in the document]
LeandraH: Good catch, thanks!

native/src/components/CustomHeaderButtons.tsx
```diff
@@ -18,17 +18,22 @@ const StyledHeaderContainer = styled.View`
...
   if (!['language', 'search'].includes(name)) {
     throw new Error('Invalid icon name!')
   }
-  return <Icon Icon={name === 'search' ? SearchIcon : LanguageIcon} {...props} />
+  return <Icon Icon={name === 'search' ? SearchIcon : LanguageIcon} style={style} {...props} />
 }
 
 const CustomHeaderButton = (props: HeaderButtonProps) => (
-  <HeaderButton {...props} IconComponent={HeaderIcon} iconSize={23} color='black' role='button' />
+  <HeaderButton {...props} IconComponent={HeaderIcon} iconSize={23} role='button' />
```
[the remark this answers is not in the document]
LeandraH: It looks a little bigger to my eyes but an independent review from René says it works :D

native/src/components/MapView.tsx
LeandraH: Ah, good point, I don't remember anymore why I changed it, and it still works, so I'll go back to the previous version 😅

shared/constants/maps.ts
LeandraH: Came with the boundingBox changes, I'll revert it

native/src/components/PinchPanImage.tsx
LeandraH: Good catch, thanks!

native/src/components/RemoteContent.tsx
LeandraH: Yes, it was. Keeping it like this led to the app crashing in iOS. I'll see if I find a different way to disable all automatic data detection

native/src/components/base/Icon.tsx
LeandraH: It's used by the `HeaderIcon` in the `CustomHeaderButtons`. We need it for hiding the search and language icon on the header in e.g. the settings route.

shared/package.json
LeandraH: Apparently not 🤣

web/package.json
LeandraH: Yes, mainly only the native ones, I'll open an issue for web, thanks for the reminder

native/patches/react-native-tts+4.1.1.patch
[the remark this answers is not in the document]
steffenkleinle: https://github.com/digitalfabrik/integreat-app/issues/3287

native/src/components/PinchPanImage.tsx
steffenkleinle: No objections, sounds great!

steffenkleinle:
I think we should add `native/android/app/.cxx` to `.gitignore`. I have some 700+ files here now.

.gitignore
```diff
@@ -118,14 +118,17 @@ dist
 # png caches
 **/.png-cache
 
-# IntelliJ
+# Android/IntelliJ
 **/workspace.xml
 runConfigurations.xml
 **/prettier.xml
+**/.cxx
```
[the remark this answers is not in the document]
bahaaTuffaha: They appear when you switch to main 😜

steffenkleinle (changes requested):
Works mostly great, I think I still found one bug though sadly: When I switch to Arabic and open events, the app crashes for me:
```
com.facebook.react.common.JavascriptException: Error: Invalid UNTIL value: ٢٠٢٤٠٤٣٠T١٠٠٠00Z
This error is located at:
  at EventListItem (address at index.android.bundle:1:3185312)
  at RCTView (<anonymous>)
  at anonymous (address at index.android.bundle:1:356443)
```
I could not reproduce this in the latest version in the stores.

steffenkleinle:
> > Works mostly great, I think I still found one bug though sadly: When I switch to Arabic and open events, the app crashes for me:
> > ```
> > com.facebook.react.common.JavascriptException: Error: Invalid UNTIL value: ٢٠٢٤٠٤٣٠T١٠٠٠00Z
> > This error is located at:
> >   at EventListItem (address at index.android.bundle:1:3185312)
> >   at RCTView (<anonymous>)
> >   at anonymous (address at index.android.bundle:1:356443)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I could not reproduce this in the latest version in the stores.
> 
> I couldn't reproduce that in android 🤔 which location/region did you use?

🤔 Landeshauptstadt München 🙈

bahaaTuffaha:
> @bahaaTuffaha did you try on a real device or an emulator? For me this happens on a real device.

On an emulator.. I will try a real device.

LeandraH:
@steffenkleinle I couldn't reproduce the crash either (though it was also only an emulated device). Language Arabic > Go to Munich > Go to Events > Crash? Anything in the console?

steffenkleinle:
> @steffenkleinle I couldn't reproduce the crash either (though it was also only an emulated device). Language Arabic > Go to Munich > Go to Events > Crash? Anything in the console?
> 
> Edit: Oh, I see the console error. I guess it might be about a specific date causing the crash...
> 
> Second edit: It looks like the crash is about an invalid end date that begins with "20240430T1", so 30 April 2024. We have three possible events for that:
>
> But all three of those load without a problem in any environment I tested, including when I filter the event overview to just show the events for next Tuesday (30 April 2024 was a Tuesday). Is it possible that it was a one-time connectivity thing where part of the loading dropped or something? You don't usually have problems opening Arabic pages on your phone, right?

Sorry, maybe I was not clear enough. This does not seem connected to the app language but only happens when I change my device language to arabic. However, I cannot reproduce this on my emulator, only on my real device. It is definitely not a one-time connectivity thing, I could reproduce this now over multiple sessions where everything works fine with other device languages.

steffenkleinle:
> > Sorry, maybe I was not clear enough. This does not seem connected to the app language but only happens when I change my device language to arabic. However, I cannot reproduce this on my emulator, only on my real device. It is definitely not a one-time connectivity thing, I could reproduce this now over multiple sessions where everything works fine with other device languages.
> 
> Do you know where its coming from exactly? The only `until` I could find is at `ExportEventButton.tsx` !
> what if we catched the error and inserted latin/English numbers as a fallback?

The problem seems to be that on my device the locale settings (`en` per default) of luxons `DateTime.toFormat()` are not respected and the device local is used instead. The `formatDateICal` method therefore produces an arabic timestamp which is obviously not supported by the rrule library we use. We therefore now implemented a (hacky) custom workaround to produce an ICal date by transforming an ISO date. We'll check if we can find an issue upstream at luxon.

steffenkleinle (approved):
Tested on android and iOS, looks really good! Lets finally get this merged 🎉

steffenkleinle:
❓ Still not that sure if it is good that in the current pinch pan image implementation it is possible to move an image partly outside of the screen. This is not behavior I am used to from other apps, usually it always snaps back.

bahaaTuffaha (approved):
Amazing work I can feel the app is faster already. 🚀🚀
Tested on Android ✅

One issue though:
when I refresh "In the vicinity" locations while disabling location permissions the settings button at the snack-bar is not redirecting me to anything.

steffenkleinle:
Lets goooooooo! 🎉 🎉🎉🎉🎉

Issue #2934, opened 2024-09-23
Some recurring icons at events have overlapping tooltips


bahaaTuffaha:
### Steps to Reproduce

1. Go to http://localhost:9000/muenchen/en/events .
2. Hover over any recurring icon on the right.

### Expected Behavior

- One tooltip should show up.

### Actual Behavior

- Overlapping of 2 tooltips

### Additional Information
![image](https://github.com/user-attachments/assets/ecf98565-2c61-439f-9dd5-2e5bc1713fea)

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #3170, opened 2025-03-26
2934: Show only one tooltip on events' date-icon

hannaseithe:
### Short Description

When hovering over the events' date-icon we show only one tooltip

### Proposed Changes
- the problem was that inside EventListItem Component we did not ascribe a unique id to the tooltip
- in order to generate a unique id I hand in the index from the list of events to a singular item, which then can be used to generate a uniqueId
- (the alternative solution of generating a uniqueId from the events._path property I chose not to implement, since we currently do not guarantee that the path is unique (see: [#3060](https://github.com/digitalfabrik/integreat-cms/issues/3060))

### Side Effects

- ListComponent
- EventsPage
- EventsListItem

### Testing

- go to the events page: http://localhost:9000/muenchen/en/events
- hover over DateIcon of an Event Item 
     -> check that only one tooltip appears 
     -> check that the correct tooltip appears 

### Resolved Issues

Fixes: #2934 

---

### DIFF
--- a/release-notes/unreleased/2934-show-only-one-tooltip-in-events-page.yml
+++ b/release-notes/unreleased/2934-show-only-one-tooltip-in-events-page.yml
@@ -1,0 +1,6 @@
+issue_key: 2934
+show_in_stores: false
+platforms: 
+  - web
+en: The date icon of the events page now only shows one tooltip when hovering over it e.g. 'Today'
+de: -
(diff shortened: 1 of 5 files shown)


steffenkleinle (approved):
Tested on firefox, nice fix!

bahaaTuffaha (approved):
Looks good ✅ tested on Firefox and Chrome.

Issue #3155, opened 2025-03-18
Update Chat API


svenseeberg:
### Describe the Problem

The API of the Integreat chat did receive some (minor?) adjustments. The app needs to be updated to work with these changes.

### Describe Your Preferred Solution

- Instead of a `body` attribute that contains a message, the `content` attribute should be used. The `content` attribute is compatible with the OpenAI style of messages, which we decided to use. After the app has been updated, the `body` attribute will be removed from the API.
- The response to sending a message from the app to the CMS was the one message. This changed with https://github.com/digitalfabrik/integreat-cms/pull/3496. Now the full message history is returned. It has the same format as the normal GET request that is used for periodic polling. I guess this means that the same callback can be used for all requests to the chat endpoint?!

### Describe Possible Alternatives

Undo the CMS changes 😬 If the changed response format for sending a message is somehow problematic, changing the response back in the CMS is definitely possible.

### Related Issues

- https://github.com/digitalfabrik/integreat-chat/issues/256
- https://github.com/digitalfabrik/integreat-chat/issues/250

### Additional Information

N/A

svenseeberg:
Now that the package publishing also works again, the changes can be tested on webnext.integreat-app.de/testumgebung-e2e

svenseeberg:
Now also on production.

Issue #3156, opened 2025-03-18
Fix Chat Typing indicator


svenseeberg:
### Describe the Bug

The typing indicator currently has 2 (3?) minor bugs:
- It does not disappear after the "real" responses arrive (it should stay after the first welcome response appears).
- It is not aligned with the other chat messages. It sticks to the left border, while all other messages have a margin with roughly the width of the robot icon.
- It seems that it does not work at all on web next currently?

![Image](https://github.com/user-attachments/assets/2ab9a493-2edf-439b-a3a0-9462ab525cb2)

### Steps to Reproduce

1. Go to https://integreat.app/testumgebung-e2e/en
2. Send a message
3. Se that the bubble is not aligned and stays when answers appear

### Expected Behavior

It would be nice if the alignment could be fixed. The response part is not that important. If its an easy fix, then okay. But we should not spend too much time on it as the real/good solution has to be websockets.

### Related Issues

- https://github.com/digitalfabrik/integreat-chat/issues/249

### Additional Information

N/A

svenseeberg:
Sorry for coming up with this so late (and I'm totally willing to move this into another issue), but I came up with a plan how we can very easily implement a real typing indicator in the CMS. I have drafted a PR for this (https://github.com/digitalfabrik/integreat-cms/pull/3561), which we can hopefully merge soon. This will introduce another attribute in the API named `chatbot_typing`. If its true, the chat bot is currently generating an answer and we could show the typing indicator.

steffenkleinle:
> Sorry for coming up with this so late (and I'm totally willing to move this into new issue), but I came up with a plan how we can very easily implement a real typing indicator in the CMS. I have drafted a PR for this ([digitalfabrik/integreat-cms#3561](https://github.com/digitalfabrik/integreat-cms/pull/3561)), which we can hopefully merge soon. This will introduce another attribute in the API named `chatbot_typing`. If its true, the chat bot is currently generating an answer and we could show the typing indicator.

I think that is a good plan. Only edge case I can think of is the user closing and reopening the chat before we receive the response. But I think that is okay and still better than the current solution. However, does the chatbot ever not generate an answer?

svenseeberg:
> However, does the chatbot ever not generate an answer or generate more than one answer?

In case of errors it will not generate an answer. In this case the flag will time out after 120s. Otherwise it will always generate some kind of answer and never more than one.

svenseeberg:
[...]

PR #3175, opened 2025-04-01
3155: Update Chat API

lunars97:
### Short Description

Due to minor changes in integreat chat, the `body` attribute should be changed to `content` to be compatible with OpenAI style of messages and api should be adjusted so that it returns full message history

### Proposed Changes

- [x]  rename `body` attribute to `content`
- [x] refactor api endpoint to return full message history
- [x] include attribute `typing` into ChatMessages endpoint object

### Side Effects

- integreat chat api

### Resolved Issues

Fixes: #3155 
Fixes: https://github.com/digitalfabrik/integreat-chat/issues/250
Fixes: https://github.com/digitalfabrik/integreat-chat/issues/256

---


shared/api/endpoints/createSendChatMessageEndpoint.ts
```diff
@@ -24,13 +24,15 @@ export default (baseUrl: string): Endpoint<ParamsType, ChatMessageModel> =>
...
-    .withMapper(
-      (json: JsonChatMessageType): ChatMessageModel =>
-        new ChatMessageModel({
-          id: json.id,
-          content: json.content,
-          userIsAuthor: json.user_is_author,
-          automaticAnswer: json.automatic_answer,
-        }),
+    .withMapper((json: JsonChatMessagesType): ChatMessageModel[] =>
+      json.messages.map(
```
hannaseithe, line 28: As already discussed, I need a slight adaptation here, so that we return also the property chatbot_typing that is returned from the endpoint in the json object. I will use this to show the TypingIndicator see also: https://github.com/digitalfabrik/integreat-app/pull/3159#issuecomment-2780642972

shared/api/buildChatMessages.ts
```diff
@@ -0,0 +1,15 @@
+import ChatMessageModel from './models/ChatMessageModel'
```
lunars97, line 1: I am not sure whether it is okay to split this into a separate helper function...

shared/api/types.ts
```diff
@@ -81,15 +83,21 @@ export type JsonCategoryType = {
...
   user_is_author: boolean
   automatic_answer: boolean
 }
 
 export type JsonChatMessagesType = {
+  chatbot_typing: boolean
   messages: JsonChatMessageType[]
 }
 
+export type ChatMessages = {
```
lunars97, line 96: I am not sure whether it should be here...

shared/api/buildChatMessages.ts
steffenkleinle: 🔧 Very much okay to move this to a separate helper function, however, in order to align with our current naming, this should be called `mapChatMessages` and reside in `shared/api/mapping/mapChatMessages`.

shared/api/types.ts
steffenkleinle: 🔧 I would say no as these are only types for the api response. I'd just move it to `mapping/mapChatMessages` (currently `buildChatMessages`).

web/src/components/ChatController.tsx
```diff
@@ -54,7 +54,10 @@ const ChatController = ({ city, language }: ChatControllerProps): ReactElement =
     })
 
     if (data !== null) {
-      setData(messages => [...(messages ?? []), data])
+      setData({
+        typing: data.typing,
+        messages: data.messages,
+      })
```
steffenkleinle, line 60: 🔧 There should be no need for this:
```suggestion
      setData(data)
```

shared/api/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { Return as ImportedReturnType } from './endpoints/hooks/useLoadAsync'
 
 export { default as Endpoint } from './Endpoint'
+export { buildChatMessages } from './buildChatMessages'
```
steffenkleinle, line 11: 🔧 No need to export this as its never used anywhere outside of `shared`.

web/src/components/ChatController.tsx
steffenkleinle: Sure! It overrides the data we get from the `useLoadFromEndpoint` hook and persists it. Useful in this case as to avoid duplicated state.

steffenkleinle (approved):

f1sh1918 (commented):
I'm not really sure how to test that properly.
Some testing instructions would be great.
The only thing i realized (TestCMS for Munich) is that the typing indicator (...) is always shown for couple of time even i don't get a response.
But thats probably a backend issue

lunars97:
> I'm not really sure how to test that properly. Some testing instructions would be great. The only thing i realized (TestCMS for Munich) is that the typing indicator (...) is always shown for couple of time even i don't get a response. But thats probably a backend issue

Sorry for not being clear and not including it... You can check the response in network and see whether it resolves the breaking point from main 
image
image
The problem with not getting response is backend issue.

hannaseithe:
> I'm not really sure how to test that properly. Some testing instructions would be great. The only thing i realized (TestCMS for Munich) is that the typing indicator (...) is always shown for couple of time even i don't get a response. But thats probably a backend issue

Also the typing indicator functionality will be fixed by me after this PR is approved in this issue #3156 . So the behavior of the typing indicator is not in alignment with the chatbot_typing property we get from the backend now. In terms of typing indicator you can only check right now whether the ChatController receives the typing property from the backend.

shared/api/mapping/mapChatMessages.ts
```diff
@@ -0,0 +1,20 @@
...
+import { JsonChatMessagesType } from '../types'
+
+export type ChatMessages = {
+  typing: boolean
+  messages: ChatMessageModel[]
+}
+
+export const mapChatMessages = (json: JsonChatMessagesType): ChatMessages => ({
+  typing: json.chatbot_typing,
+  messages: json.messages.map(
```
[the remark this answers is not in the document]
steffenkleinle: I agree that the naming is not optimal. I was also thinking about that and did not figure out a nice enough idea to question it. Now I am thinking that perhaps just `ChatMessagesReturn` would fit? In the end, its not used outside of the endpoints in any case. I think both `Feed`and `State`do not quite reflect it either.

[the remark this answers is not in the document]
lunars97: I also agree with Steffen and would use `ChatMessagesReturn`...

hannaseithe (approved):

PR #3176, opened 2025-04-02
Decrease chat polling interval

svenseeberg:
### Short Description

With the [latest change of the Integreat CMS](https://github.com/digitalfabrik/integreat-cms/pull/3559) we introduced caching of Zammad responses. This makes the API quite fast and we can decrease polling times to improve the experience for the user.

### Proposed Changes

- Decrease chat polling interval to 8s.

### Side Effects

- More load on the CMS and more bandwidth usage. However, still only a few KB.

### Resolved Issues

Fixes: None

---


steffenkleinle (approved):
LGTM

Issue #3173, opened 2025-03-31
Adapt Layout for Link to make appointment inside POI for web


hannaseithe:
### Describe the Problem

The layout should be set up analogous to how it has been implemented here for native: #3168 

### Describe Your Preferred Solution

The appointmentLink should be always part of the OpeningHours section (if not component).
- It should not be part of the collapsible. 
- When no opening Hours are set, the Opening Hours title should say "Only with appointment" and no collapsible

### Describe Possible Alternatives

- stays the same

### Related Issues

#3148 
#3168

Issue #3148, opened 2025-03-10
Link to make an appointment at a POI missing in native


LeandraH:
### Describe the Bug

When going to e.g. https://webnext.integreat.app/testumgebung/de/locations/rathaus-augsburg in native, I don't get a link to make an appointment there (it is there in web).

### Steps to Reproduce

1. Open the native app on Testumgebung - DE
2. Go to Locations
3. Go to the location Rathaus Augsburg
4. See that there is no link to make an appointment

Note: in case the CMS has been reset when we start working on this, create a POI with a link for making an appointment and take a look at that on mobile.

### Expected Behavior

There is a link to book an appointment, like in the web version.

### Related Issues

Found in this (not all that related) issue: https://github.com/digitalfabrik/integreat-app/pull/3138#pullrequestreview-2670297338

### Additional Information

<img src="https://github.com/user-attachments/assets/06749068-e1fb-46a9-8d4f-9f3dea126804" width="300px" />

PR #3180, opened 2025-04-07
3173: Adapt layout for link to make appointment inside POI for web

lunars97:
### Short Description

Opening hours layout in poi should be adapted to include "Only with appointment" when no opening hours are set.

### Proposed Changes

- include new label "Only with appointment" and set a condition to use it only when there is no opening hours 

### Side Effects

- none

### Testing

Go to http://localhost:9000/testumgebung/de/locations/ort-mit-link-f%C3%BCr-termine and see that there is a new label 

### Resolved Issues

Fixes: #3173 

---

### DIFF
--- a/web/src/components/OpeningHours.tsx
+++ b/web/src/components/OpeningHours.tsx
@@ -1,7 +1,7 @@
 import { DateTime } from 'luxon'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
-import styled from 'styled-components'
+import styled, { useTheme } from 'styled-components'
 
 import { weekdays } from 'shared'
 import { OpeningHoursModel } from 'shared/api'
@@ -9,6 +9,7 @@ import { OpeningHoursModel } from 'shared/api'
 import { helpers } from '../constants/theme'
 import Collapsible from './Collapsible'
 import OpeningEntry from './OpeningEntry'
+import Spacer from './Spacer'
 
 const OpeningLabel = styled.span<{ $isOpen: boolean }>`
   color: ${props => (props.$isOpen ? props.theme.colors.positiveHighlight : props.theme.colors.negativeHighlight)};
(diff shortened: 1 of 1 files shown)


lunars97:
waiting for #3148 to be merged into main

web/src/components/OpeningHours.tsx
```diff
@@ -56,27 +64,34 @@ const OpeningHours = ({
...
         </OpeningLabel>
       </OpeningContainer>
     </TitleContainer>
   )
-  if (isTemporarilyClosed) {
-    return <TitleContainer>{openingHoursTitle}</TitleContainer>
+  if (isTemporarilyClosed || isOnlyWithAppointment) {
+    return (
+      <>
+        <Spacer $borderColor={theme.colors.borderColor} />
```
bahaaTuffaha, line 83: There is two spacers here:

web/src/components/OpeningHours.tsx
```diff
@@ -24,8 +24,9 @@ const Content = styled.div`
 const TitleContainer = styled.div`
   display: flex;
   flex: 1;
-  font-weight: 700;
   justify-content: space-between;
+  flex-wrap: wrap;
```
bahaaTuffaha, line 28: The space between Wegbeschreibung and the address is different from the space between Öffnungszeiten and Vorübergehend geschlossen when its wrapped.
I think you can use `gap: 12px;` to solve it 🙂.

web/src/components/__tests__/PoisDesktop.spec.tsx
```diff
@@ -55,7 +55,6 @@ describe('PoisDesktop', () => {
     expect(queryByText('pois:distanceKilometre')).toBeTruthy()
     expect(queryByText(singlePoi.location.address!)).toBeTruthy()
     expect(queryByText(singlePoi.content)).toBeTruthy()
-    expect(queryByText('pois:makeAppointment')).toBeTruthy()
```
steffenkleinle, line 58: 🤔 Why is this not truthy anymore/why did you remove it?

steffenkleinle (approved):
Thank you, well don!

web/src/components/__tests__/PoisDesktop.spec.tsx
bahaaTuffaha: Because its not at Poi anymore I moved the appointmentLink to OpeningHours component like native. 🙂

Issue #3080, opened 2025-02-12
Adjust date filter placeholder text


steffenkleinle:
### Describe the Problem

As a user I find it confusing that the default until date in the events date filter is also 01.01.1990.

### Describe Your Preferred Solution

I would want a placeholder text that transports the functionality better.
For the start date, use the *current date*.
For the end date, use the *current date + 7 days*.
Do NOT set this as default, this should still only be a placeholder!

### Describe Possible Alternatives

- current date
- date format (difficult as its localized without us knowing the exact format)

### Related Issues

Date filter was implemented in #2808 
More event filtering:
#2707 
#2112

### Additional Information

![Image](https://github.com/user-attachments/assets/61fc7981-b406-4580-8d1f-56e162ac3575)

hauf-toni:
Based on best practices, an **effective default date selection should be context-aware and intuitive for the user.** most websites avoid preselecting a specific date altogether, instead displaying placeholders such as "from" and "to" until the user interacts with the filter (see eg Airbnb, Eventim). this approach prevents misleading assumptions + provides flexibility for different use cases. when the user clicks on the filter they are typically presented with a calendar that defaults to a relevant time period such as the current day/week/month, depending on the context of the search. only after user selection is a date explicitly shown.

In our case, a practical and user-friendly default would be to **prefill the start date with today's date + set the end date to 7 days later**. this approach aligns with common patterns for event searches, where users typically look for upcoming events rather than past ones. by selecting a one-week period by default, we balance flexibility with usability — providing an immediate, useful range while still allowing users to refine their search as needed. this method also ensures that users are presented with relevant results without requiring unnecessary manual input .(See eg Eventbrite, Tripadvisor)

hannaseithe:
I am not sure whether I should set the actual default and/or just the placeholder? The issue description and the comment by @hauf-toni  seem to contradict each other.

steffenkleinle:
> I am not sure whether I should set the actual default and/or just the placeholder? The issue description and the comment by [@hauf-toni](https://github.com/hauf-toni) seem to contradict each other.

It should only be set as placeholder as follows:
start date: current date as placeholder
end date: current date + 7 days as placeholder
Nothing should be preselected.

PR #3183, opened 2025-04-08
3080: Set dateFilters placeholder to Today until a week later

hannaseithe:
### Short Description

I have adjusted the events date filter, so that it now shows >today< as the "From-Date-Placeholder" and the >date in 7 days< as the "To-Date-Placeholder"

### Proposed Changes

- when no Filter is selected yet, we show a placeholder which lets the user know the format of the date picker component. This is now set to From: TODAY and UNTIL: 7_DAYS_LATER
-  I added the property placeholderDate to the DatePicker component, which now allows us to set the placeholder. If the placeholderDate is not set, it will default to today's date
- in web (in native that was not possible), we set a light blue marker on the placeholder date (if no date is selected yet) in the calendar picker, which aligns with the placeholderDate value

### Side Effects

- 

### Testing

Check both in native and web that if the date filter is not set, that the placeholder shows as from:TODAY until:IN_A_WEEK

### Resolved Issues

Fixes: #3080

---


native/src/components/DatePicker.tsx
```diff
@@ -64,9 +64,18 @@ export type DatePickerProps = {
   error?: string
   modalOpen: boolean
   setModalOpen: (open: boolean) => void
+  placeholderDate?: DateTime
```
hannaseithe, line 67: I contemplated whether to use Date or DateTime, but since we use DateTime for another property, I thought this makes sense

web/src/components/DatePicker.tsx
```diff
@@ -143,9 +144,10 @@ const CustomDatePicker = ({ title, date, setDate, error }: CustomDatePickerProps
           onClickOutside={() => setIsCalendarOpen(false)}
           calendarClassName='calenderStyle'
           dateFormat='dd.MM.yyyy'
-          placeholderText='01.01.1990'
+          placeholderText={(placeholderDate ?? DateTime.now()).toFormat('dd.MM.yyyy')}
           onChange={(date: Date | null) => handleDateChange(date)}
           onChangeRaw={e => handleDateError(String((e?.target as HTMLInputElement).value))}
+          openToDate={(placeholderDate ?? DateTime.now()).toJSDate()}
```
hannaseithe, line 150: I added this after talking to Steffen - its basically a blue marker (until we select something). But since this is only done in web, I am not entirely sure if it makes sense

native/src/components/__tests__/DatePicker.spec.tsx
```diff
@@ -20,19 +28,24 @@ describe('DatePickerForNative', () => {
...
       modalOpen: false,
       setModalOpen,
       setDate,
       title: 'Test DatePicker',
       date: DateTime.now(),
       error: '',
+      placeholderDate: DateTime.fromISO('1999-01-09'),
     })
-
+    console.log(new Date('1999-01-09'))
```
bahaaTuffaha, line 45: You forgot to remove this line.
```suggestion
```

native/src/components/DatePicker.tsx
```diff
@@ -64,9 +64,18 @@ export type DatePickerProps = {
...
 
-const DatePicker = ({ title, date, setDate, error, modalOpen, setModalOpen }: DatePickerProps): ReactElement => {
+const DatePicker = ({
+  title,
+  date,
+  setDate,
+  error,
+  modalOpen,
+  setModalOpen,
+  placeholderDate,
```
steffenkleinle, line 77: 🙃  ~~Using a default value here also allows you to get rid of the very unspecific `dateToUse` naming.~~
```suggestion
  placeholderDate = DateTime.now(),
```

Edit: Applying the suggestion above makes this unnecessary.

native/src/components/DatePicker.tsx
```diff
@@ -75,6 +84,10 @@ const DatePicker = ({ title, date, setDate, error, modalOpen, setModalOpen }: Da
   const dayRef = useRef<TextInput>(null)
   const monthRef = useRef<TextInput>(null)
   const yearRef = useRef<TextInput>(null)
+  const dateToUse = placeholderDate ?? DateTime.now()
+  const placeholderDay = dateToUse.toFormat('dd')
+  const placeholderMonth = dateToUse.toFormat('MM')
+  const placeholderYear = dateToUse.toFormat('yyyy')
```
steffenkleinle, line 90: 🙃 
```suggestion
  const placeholderDay = placeholderDate.toFormat('dd')
  const placeholderMonth = placeholderDate.toFormat('MM')
  const placeholderYear = placeholderDate.toFormat('yyyy')
```

native/src/components/EventsDateFilter.tsx
```diff
@@ -65,6 +65,9 @@ const EventsDateFilter = ({
   const { t } = useTranslation('events')
   const currentInput = useRef<string>('from')
 
+  const today = DateTime.now()
+  const inAWeek = DateTime.now().plus({ days: 7 })
```
steffenkleinle, line 69: 🔧 I think this is simpler and also works in calendars that have different week lengths (not sure if such a calendar actually exists) .
```suggestion
  const inAWeek = DateTime.now().plus({ weeks: 1 })
```

web/src/components/__tests__/DatePicker.spec.tsx
```diff
@@ -17,19 +17,23 @@ describe('DatePicker', () => {
...
       title,
       date,
       setDate,
       error: '',
+      placeholderDate,
     })
 
     expect(getByText(title)).toBeInTheDocument()
-    expect(getByPlaceholderText('01.01.1990')).toBeInTheDocument()
+    expect(getByPlaceholderText(placeholderDate.toFormat('dd.MM.yyyy'))).toBeInTheDocument()
```
bahaaTuffaha, line 31: @steffenkleinle Is it ok to expect a variable like this or better to be hard coded like a string (for tests)?

web/src/components/DatePicker.tsx
```diff
@@ -74,6 +74,7 @@ export type CustomDatePickerProps = {
   date?: DateTime | null
   setDate: (date: DateTime | null) => void
   error?: string
+  placeholderDate?: DateTime
```
steffenkleinle, line 77: 🔧 Since we always pass a placeholder date (and it is best practice to do so), we should make this required.
```suggestion
  placeholderDate: DateTime
```

bahaaTuffaha (approved):
Nice work✨ (tested on android and web)
I left some suggestions

web/src/components/DatePicker.tsx
```diff
@@ -143,9 +144,10 @@ const CustomDatePicker = ({ title, date, setDate, error }: CustomDatePickerProps
           onClickOutside={() => setIsCalendarOpen(false)}
           calendarClassName='calenderStyle'
           dateFormat='dd.MM.yyyy'
-          placeholderText='01.01.1990'
+          placeholderText={(placeholderDate ?? DateTime.now()).toFormat('dd.MM.yyyy')}
```
steffenkleinle, line 147:
```suggestion
          placeholderText={placeholderDate.toFormat('dd.MM.yyyy')}
```

native/src/components/DatePicker.tsx
steffenkleinle, line 67: 🔧 Since we always pass a placeholder date (and it is best practice to do so), we should make this required.
```suggestion
  placeholderDate: DateTime
```

steffenkleinle (commented):
Nice work! Only a few minor suggestions :)

web/src/components/DatePicker.tsx
bahaaTuffaha: Its usually selects the current day right?

steffenkleinle: > Its usually selects the current day right?

Oh, you're right. One more reason to remove this again as this is what users are used to. Thanks! And sorry @hannaseithe for the bad suggestion here.

steffenkleinle: I think being set to today is the expected behavior so no need to get completely rid of it, just unset it imo.

native/src/components/EventsDateFilter.tsx
hannaseithe: I am very curious to see if there is another week length somewhere in the world! 😄  But I agree, Ill make the change.

steffenkleinle: Let me know if you find out 😝

native/src/components/__tests__/DatePicker.spec.tsx
```diff
@@ -98,6 +113,7 @@ describe('DatePickerForNative', () => {
       title: 'Test DatePicker',
       date: null,
       error: '',
+      placeholderDate: DateTime.fromISO('1990-01-01'),
```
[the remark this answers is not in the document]
hannaseithe: On @steffenkleinle suggestion I changed it to the prop placeholderDate to required, so I'll keep it in here

steffenkleinle (approved):
Tested on android emulator and firefox. Works as expected 🎉

Issue #3186, opened 2025-04-08
Adjust theme for high contrast mode


lunars97:
### Describe the Problem
Adjust theme to use high contrast mode colors in build-configs

### Describe Your Preferred Solution

ThemeType and ColorType should be adjusted in build-configs to be available for further usage and adjustments.

steffenkleinle:
Done in #3194

PR #3194, opened 2025-04-09
3186: Adjust theme for high contrast mode

lunars97:
### Short Description

Theme should be adjust in build-configs to accept high contrast theme on integreat

### Proposed Changes

- add high contrast theme
- add colors necessary for high contrast theme
- make adjustments in other platforms to avoid error (we are implementing for now only for integreat)

### Side Effects

- none

### Resolved Issues

Fixes: #3186 

---

### DIFF
--- a/build-configs/common/theme/colors.ts
+++ b/build-configs/common/theme/colors.ts
@@ -1,26 +1,26 @@
 export type ColorsType = {
   themeColor: string
-  themeColorLight?: string
+  themeColorLight: string
   backgroundAccentColor: string
   textColor: string
-  textSecondaryColor?: string
-  textDecorationColor?: string
-  textDisabledColor?: string
-  footerLineColor?: string
-  backgroundColor?: string
-  tunewsThemeColor?: string
-  tunewsThemeColorLight?: string
+  textSecondaryColor: string
+  textDecorationColor: string
+  textDisabledColor: string
+  footerLineColor: string
+  backgroundColor: string
+  tunewsThemeColor: string
+  tunewsThemeColorLight: string
   borderColor: string
-  positiveHighlight?: string
-  negativeHighlight?: string
-  invalidInput?: string
-  warningColor?: string
-  linkColor?: string
-  themeContrast?: string
-  ttsPlayerWarningBackground?: string
-  ttsPlayerWarningColor?: string
-  ttsPlayerBackground?: string
-  ttsPlayerPlayIconColor?: string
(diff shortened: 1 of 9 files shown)


build-configs/common/theme/colors.ts
```diff
@@ -1,26 +1,26 @@
...
-  textDisabledColor: string
-  footerLineColor: string
-  backgroundColor: string
-  tunewsThemeColor: string
-  tunewsThemeColorLight: string
+  textSecondaryColor?: string
+  textDecorationColor?: string
+  textDisabledColor?: string
+  footerLineColor?: string
+  backgroundColor?: string
```
lunars97, line 10: I made them optional, because of the error I was getting saying that those colors are absent

steffenkleinle: I think a better option would be something like
``` ts
export const commonHighContrastColors = {
  ...commonLightColors,
  backgroundAccentColor: '#20293A',
  backgroundColor: '#101217',
  textColor: '#FFFFFF',
  borderColor: '#FFFFFF',
}
```

steffenkleinle: For me this works if you adjust `build-configs/integreat/theme/colors.ts`as follows:
``` ts
export const highContrastColors: ColorsType = {
  themeColor,
  themeColorLight,
  themeContrast,
  ...commonHighContrastColors,
}
```

build-configs/malte/index.ts
```diff
@@ -19,6 +19,7 @@ const commonMalteBuildConfig: CommonBuildConfigType = {
   appIcon: 'app_icon_malte',
   notificationIcon: 'notification_icon_malte',
   lightTheme,
+  highContrastTheme: lightTheme,
```
bahaaTuffaha, line 22: :question: shouldn't it be highContrastTheme here ?

bahaaTuffaha (approved):
Looks good overall ✅
🤔 Do we need to pass the new theme to the `ThemeProvider` at `App.tsx` ?

build-configs/malte/index.ts
lunars97: This will be done in upcoming tickets where I will implement the functional part :)

lunars97:
> Looks good overall ✅ 🤔 Do we need to pass the new theme to the `ThemeProvider` at `App.tsx` ?

I will do the functional part in upcoming PR :)

Issue #2928, opened 2024-09-16
Make search more error resistant and provide more information


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
At the moment, the search in Landkreis Ostallgäu is broken. This is due to minisearch failing if two items have the same id (in our case path). While this should not happen in general and be fixed in the CMS, the app should also not just fail/load forever without any error message.

**Describe the solution you'd like**
- Provide better errors if the search fails to the users (display an error message instead of endless loading spinner)
- Provide better errors for devs (log full errors to the console)
- Make the search more error resistant by removing items with duplicated id (optional/TBD)

**Describe alternatives you've considered**
Do nothing as this is not an app issue.

**Related issues**
CMS issue: https://github.com/digitalfabrik/integreat-cms/issues/3060
Related: #3192 

**Additional context**
Event in question: http://integreat.app/ostallgaeu/de/events/handarbeits-werkstatt-mit-begegnung-und-austausch-5

Issue #3192, opened 2025-04-09
Search for Pages in Görlitz doesn't work


AnnaPerelazova:
It seems that only one municipality is affected by this bug (Landkreis Görlitz): https://integreat.app/lkgoerlitz/ (https://admin.integreat-app.de/lkgoerlitz/). In the search results, only events are displayed, but not pages. 

![Image](https://github.com/user-attachments/assets/28f7f469-d804-4e4c-ac0e-fb1b412879c0)

Thanks for looking into it!

LeandraH:
Didn't we have an issue about that as well though? I can't find it at the moment, but it might be worth fixing at our end as well, this is already the second time it happened and it's a really bad UX.

AnnaPerelazova:
@steffenkleinle thank you for checking! but why the pages with "beratung" are not shown? (e.g. https://integreat.app/lkgoerlitz/de/%C3%A4mter-beratung-und-notfallnummern/beratung-und-hilfe/migrationsberatung-und-jugendmigrationsberatung) The main problem is that is only looking for events not for pages.

steffenkleinle:
> Didn't we have an issue about that as well though? I can't find it at the moment, but it might be worth fixing at our end as well, this is already the second time it happened and it's a really bad UX.

I guess you mean this issue: https://github.com/digitalfabrik/integreat-app/issues/2928
I'll have a look.

steffenkleinle:
> [@steffenkleinle](https://github.com/steffenkleinle) thank you for checking! but why the pages with "beratung" are not shown? (e.g. https://integreat.app/lkgoerlitz/de/%C3%A4mter-beratung-und-notfallnummern/beratung-und-hilfe/migrationsberatung-und-jugendmigrationsberatung) The main problem is that is only looking for events not for pages.

I can't tell exactly why this happens, but in the end unexpected input often leads to unexpected output/behavior :/ The easiest fix would be deleting (or adjusting) the path of one of the two events with duplicated path.

AnnaPerelazova:
@steffenkleinle so it is searching for events only, because of this one duplicated path? it is not searching for pages at all, doesn't matter what are you searching for, it will show only events. Is it possible to fix this problem any soon? They are asking for update. Thx!

steffenkleinle:
There is an open PR for this waiting for review. I'll ask for fast reviews 👍 But the fastest solution would be to remove and readd the event with the duplicated path I guess.

[...]

PR #3200, opened 2025-04-15
2928: Improve error reporting and fix search with duplicates

steffenkleinle:
### Short Description

Improve error reporting for the search and fix search problems with duplicated documents.

### Proposed Changes

- Remove duplicates from documents before searching
- Report errors occurring in search
- Adjust root project code style

### Side Effects

N/A

### Testing

1. ~~Manually add a duplicated document before sanitizing in `useSearch`(L70):~~ See testing instructions by @LeandraH [here](https://github.com/digitalfabrik/integreat-app/pull/3200#pullrequestreview-2768038950).
2. Ensure that no error is thrown
3. Manually add a duplicated document after sanitizing in `useSearch`(L75):
```ts
        .addAllAsync([...documents, documents[0]!])
```
4. Ensure that the error is caught and reported (i.e. logged in dev mode)

### Resolved Issues

Fixes: #2928, #3192

---

### DIFF
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -20,11 +20,18 @@ export const prepareSearchDocuments = (
   ...(pois || []),
 ]
 
+type UseSearchReturn = {
+  data: SearchResult[]
+  error: Error | null
+  loading: boolean
+}
+
 // WARNING: This uses the document count to check whether the search documents have already been added.
 // Modifying single documents or replacing documents with a same length array will therefore NOT trigger an update
-const useSearch = (documents: SearchResult[], query: string): SearchResult[] => {
+const useSearch = (documents: SearchResult[], query: string): UseSearchReturn => {
   const [indexing, setIndexing] = useState(false)
   const [debouncedQuery, setDebouncedQuery] = useState(normalizeString(query))
+  const [error, setError] = useState<Error | null>(null)
 
   const [search] = useState(
     new MiniSearch({
@@ -55,12 +62,15 @@ const useSearch = (documents: SearchResult[], query: string): SearchResult[] => 
       search
         .addAllAsync(documents)
         .then(() => setIndexing(false))
-        .catch()
+        .catch(setError)
     }
   }, [indexing, search, documents])
 
-  // @ts-expect-error minisearch doesn't add the returned storeFields (e.g. title or path) to its typing
-  return debouncedQuery.length === 0 ? documents : search.search(debouncedQuery)
+  return {
+    data: debouncedQuery.length === 0 ? documents : (search.search(debouncedQuery) as unknown as SearchResult[]),
+    error,
+    loading: indexing,
+  }
 }
 
 export default useSearch
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -10,6 +10,17 @@ import normalizeString from '../utils/normalizeString'
 export type SearchResult = ExtendedPageModel
 const DEBOUNCED_QUERY_TIMEOUT = 250
 
+const removeDuplicatedPaths = (documents: SearchResult[]) => {
+  const paths = new Set()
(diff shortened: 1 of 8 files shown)


.idea/codeStyles/Project.xml
LeandraH, line 1: ❌  These two should go. Maybe add them to gitignore?

LeandraH (commented):
I tested it by going to http://localhost:9000/lkgoerlitz/de/search/ and setting the API to production (the duplicate item seems to not exist in dev) with `window.localStorage.setItem('api-url', 'https://cms-test.integreat-app.de')`. I'm still not getting content results :(

.idea/codeStyles/Project.xml
LeandraH: For people using IntelliJ :) Either way, these changes don't belong to this PR.

steffenkleinle: https://github.com/digitalfabrik/integreat-app/pull/3203

steffenkleinle:
> I tested it by going to http://localhost:9000/lkgoerlitz/de/search/ and setting the API to production (the duplicate item seems to not exist in dev) with `window.localStorage.setItem('api-url', 'https://cms-test.integreat-app.de')`. I'm still not getting content results :(

Sorry, I missed reverting L20 which I set to `true` for testing purposes. Could you please check again?

shared/hooks/useSearch.ts
```diff
@@ -49,18 +67,22 @@ const useSearch = (documents: SearchResult[], query: string): SearchResult[] =>
   }, [query])
 
   useEffect(() => {
-    if (!indexing && search.documentCount !== documents.length) {
+    const sanitizedDocuments = removeDuplicatedPaths(documents)
```
bahaaTuffaha, line 70: After replacing it with:
```js 
const sanitizedDocuments = removeDuplicatedPaths([...documents, documents[0]!])
```
It throws the following:
`Cannot read properties of undefined (reading 'path')`
and highlights this line as the source of the issue: 
```js
 const isNew = !paths.has(document.path)
```

steffenkleinle: I guess its better to follow @LeandraH's testing instructions [here](https://github.com/digitalfabrik/integreat-app/pull/3200#pullrequestreview-2768038950) as this is the original issue (and don't forget to set the cms url back to the test tms afterwards). However, if you would still like to test it this way, you can try this:
```ts
    const sanitizedDocuments = removeDuplicatedPaths(documents.length > 0 ? [...documents, documents[0]!] : [])
```

LeandraH (approved):
Nicely done, thank you for the quick fix!

bahaaTuffaha (approved):
Nice work 🚀
Not sure if it's related but it looks like there is a duplicate key at the list:

LauraReichert:
Hi, do you have any new information regarding this issue? 
LK Höxter contacted me, saying that the search function isn't working correct. 
In this case, its about the word "Minijob". The main info-page about this topic (https://integreat.app/kreishoexter/de/arbeit-und-bildung/arbeit/arbeitsvertrag) is not shown as a result of the search in Integreat.
If this is not the same issue, please let me know, so I can open or comment another one. Thank you!

LeandraH:
Huh, that is a weird issue, and it is indeed a different problem. The problem described in this issue finds no search results from the content pages at all, and instead only events and locations. But going to their page, I do find other content pages, so yes, a new issue would be good :)

steffenkleinle:
> Huh, that is a weird issue, and it is indeed a different problem. The problem described in this issue finds no search results from the content pages at all, and instead only events and locations. But going to their page, I do find other content pages, so yes, a new issue would be good :)

Also, I find the page for example if I search for `Probezeit` or `Arbeitsverhältnis`. It seems like words in the title are not found, e.g. it also does not work for `Unbefristet`. Can you please open an issue? Thank you!

Issue #3184, opened 2025-04-08
Add toggle button for high contrast mode for web


lunars97:
### Describe the Problem
Add toggle button for high contrast mode on web according to [figma design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=4614-1026&t=eUdYoUHdsSYF1l5N-0)
**Warning:** this would only be visible once a city is selected.
### Describe Your Preferred Solution

Add toggle button both for desktop view and mobile view according to figma design

steffenkleinle:
Done in #3201

PR #3201, opened 2025-04-15
Add implementation for toggle contrast theme

lunars97:
### Short Description

Part of #3184 implementation. This PR adds implementation logic for contrast theme toggle button function.

### Proposed Changes

- Added ContrastThemeContext.tsx to contain the main functional part and to provide state to all the components
- Made adjustments to App.tsx to supply the state changes to nested components
- Made slight adjustments to ContrastTheme.tsx component
- Created useContrastTheme.tsx Hook to be able to use easily in other components

### Side Effects
none

### Testing

Check whether everything works fine and check the browser theme detection (you can change the theme of your browser to dark and see whether the website automatically turns to contrast/dark mode).

### Resolved Issues

Fixes: #

---


steffenkleinle (changes requested):
Looks great already, very nice work 🎉

web/src/components/ContrastThemeContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+const getSystemPreference = (): boolean => {
+  const forcedColors = window.matchMedia('(forced-colors: active)')
+  const prefersContrast = window.matchMedia('(prefers-contrast: more)')
+  const prefersDark = window.matchMedia('(prefers-color-scheme: dark)')
+  return forcedColors.matches || prefersContrast.matches || prefersDark.matches
+}
+
+export const ContrastThemeProvider = ({ children }: ContrastThemeProviderProps): ReactElement => {
+  const [isContrastTheme, setIsContrastTheme] = useState<boolean>(false)
+  const [overrideEnabled, setOverrideEnabled] = useState<boolean>(false)
```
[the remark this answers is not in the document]
lunars97: I though that I might need additional state to save it in localStorage for browser detect override...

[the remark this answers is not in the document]
steffenkleinle: The browser detection does not need a state as this is nothing the user can change (and is only needed once initially if no override is set yet). Have a look at how we use `useLocalStorage` in our app at the moment. You can just pass the browser detected theme as default value.

bahaaTuffaha (approved):
Great work 🤩
Tested on Firefox and chrome and even android(web app).
There is some unreadable text and views that needs adjustments I guess for another PR.

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+import ToolbarItem from './ToolbarItem'
+import Button from './base/Button'
+import Icon from './base/Icon'
+
+const ContrastButton = styled(Button)`
+  display: flex;
+  width: 100%;
+  padding: 24px 0;
+  align-items: center;
+  cursor: pointer;
```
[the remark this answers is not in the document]
lunars97: Good catch! I will include it in this PR 3189-adjust-styles-to-align-with-high-contrast-theme I implemented misalignments here :)

web/src/components/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const ThemeContainer = ({ children, contentDirection }: ThemeContainerProps): ReactElement => {
+  const [themeType, setThemeType] = useState<ThemeType>(() => getSystemTheme())
+
+  useEffect(() => {
+    const handleSystemChange = () => {
+      const systemTheme = getSystemTheme()
+      setThemeType(prev => (prev !== systemTheme ? systemTheme : prev))
+    }
+
+    handleSystemChange()
```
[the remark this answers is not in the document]
lunars97: :thinking: I will have a look...

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ContrastIcon } from '../assets'
+import { useContrastTheme } from '../hooks/useContrastTheme'
+import useWindowDimensions from '../hooks/useWindowDimensions'
+import ToolbarItem from './ToolbarItem'
+import Button from './base/Button'
+import Icon from './base/Icon'
+
+const ContrastButton = styled(Button)`
+  display: flex;
+  width: 100%;
```
steffenkleinle, line 14: 🔧 Not needed
```suggestion
```

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+  & > span {
+    padding: 0 28px;
+    color: ${props => props.theme.colors.textColor};
+  }
+`
+
+const StyledIcon = styled(Icon)<{ $isContrastTheme: 'light' | 'contrast' }>`
+  width: 24px;
+  height: 24px;
+  transform: ${({ $isContrastTheme }) => ($isContrastTheme === 'contrast' ? 'reverse' : 'none')};
```
steffenkleinle, line 28: 🔧 This does not use the `reverse` property of the icon. Please have a look e.g. `TtsPlayer` and search for `reverse`.

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  & > span {
+    padding: 0 28px;
+    color: ${props => props.theme.colors.textColor};
+  }
+`
+
+const StyledIcon = styled(Icon)<{ $isContrastTheme: 'light' | 'contrast' }>`
+  width: 24px;
+  height: 24px;
```
steffenkleinle, line 27: 🙃 This are the default dimensions so you could just omit them:
```suggestion
```

web/src/components/base/Button.tsx
```diff
@@ -16,6 +16,7 @@ type ButtonProps = {
   label: string
   type?: 'submit' | 'button'
   disabled?: boolean
+  dir?: string
```
steffenkleinle, line 19: 🔧 Why did you choose to make this configurable? This is never set to anything else but `auto`.

web/src/hooks/useContrastTheme.tsx
```diff
@@ -0,0 +1,5 @@
+import { useContext } from 'react'
+
+import { ThemeContext, ThemeContextType } from '../components/ThemeContext'
+
+export const useContrastTheme = (): ThemeContextType => useContext(ThemeContext)
```
steffenkleinle, line 5: 🔧 This is not contrast theme specific:
```suggestion
export const useThemeContext = (): ThemeContextType => useContext(ThemeContext)
```

web/src/components/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+    contrastThemeMediaQueries.forEach(query => query.addEventListener('change', handleSystemChange))
+
+    return () => {
+      contrastThemeMediaQueries.forEach(query => query.removeEventListener('change', handleSystemChange))
+    }
+  }, [])
+
+  const toggleTheme = useCallback(() => {
+    setThemeType(prev => (prev === 'light' ? 'contrast' : 'light'))
+  }, [])
```
steffenkleinle, line 52: 🙃 If you move this inside of the `useMemo`, you can avoid the `useCallback`.

web/src/components/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const getSystemTheme = (): ThemeType => (contrastThemeMediaQueries.some(query => query.matches) ? 'contrast' : 'light')
+
+type ThemeContainerProps = {
+  children: ReactElement
+  contentDirection: UiDirectionType
+}
+
+export const ThemeContainer = ({ children, contentDirection }: ThemeContainerProps): ReactElement => {
+  const [themeType, setThemeType] = useState<ThemeType>(() => getSystemTheme())
```
steffenkleinle, line 33: 🔧 No need to wrap this function.
```suggestion
  const [themeType, setThemeType] = useState<ThemeType>(getSystemTheme)
```

web/src/components/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  useEffect(() => {
+    const handleSystemChange = () => {
+      const systemTheme = getSystemTheme()
+      setThemeType(prev => (prev !== systemTheme ? systemTheme : prev))
+    }
```
steffenkleinle, line 39: 🤔 This basically means that afterwards it is always set to the system theme, right? So you could just do:
```suggestion
    const updateTheme = () => setThemeType(getSystemTheme())
```

web/src/components/ThemeContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+      contrastThemeMediaQueries.forEach(query => query.removeEventListener('change', handleSystemChange))
+    }
+  }, [])
+
+  const toggleTheme = useCallback(() => {
+    setThemeType(prev => (prev === 'light' ? 'contrast' : 'light'))
+  }, [])
+
+  const contextValue = useMemo(() => {
+    const baseTheme = themeType === 'contrast' ? themeConfig.highContrastTheme : themeConfig.lightTheme
```
steffenkleinle, line 55: 🙃 I guess for consistency we could rename this to `contrastTheme` as well. Not necessary here though.
```suggestion
    const baseTheme = themeType === 'contrast' ? themeConfig.contrastTheme : themeConfig.lightTheme
```

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+`
+
+const ContrastThemeToggle = (): ReactElement => {
+  const { t } = useTranslation('layout')
+  const { viewportSmall } = useWindowDimensions()
+  const { toggleTheme, themeType } = useContrastTheme()
+
+  if (viewportSmall) {
+    return (
+      <ContrastButton dir='auto' label={t('contrastTheme')} onClick={toggleTheme}>
```
steffenkleinle, line 39: 🙃 `dir='auto'` is the default, so no need to pass it explicitly:
```suggestion
      <ContrastButton label={t('contrastTheme')} onClick={toggleTheme}>
```

web/src/constants/contrastThemeMediaQueries.tsx
```diff
@@ -0,0 +1,5 @@
+export const contrastThemeMediaQueries = [
```
steffenkleinle, line 1: 🙃 Sorry, by extracting this to a shared constant I did not mean that you have to put it to an own file, only to reuse it for both usages in ´ThemeContext´. So feel free to move it back to `ThemeContext`.

web/src/components/ContrastThemeToggle.tsx
lunars97: :see_no_evil: I thought there is a property called 'reverse' and visually I have not noticed it

web/src/components/ContrastThemeToggle.tsx
lunars97: width is needed so that its children could move on language direction change

web/src/components/ThemeContext.tsx
lunars97: I will rename it in another PR :+1:

web/src/constants/contrastThemeMediaQueries.tsx
lunars97: true :+1:

web/src/components/ContrastThemeToggle.tsx
steffenkleinle: True 👍

web/src/components/ContrastThemeToggle.tsx
steffenkleinle: You should use the `reverse` property as follows:
``` tsx
const StyledIcon = styled(Icon)`
  transition: transform 0.2s ease;
`
...
        <StyledIcon reverse={themeType === 'contrast'} src={ContrastIcon} />
```

web/src/components/base/Button.tsx
steffenkleinle: Please remove this property if its not needed to set this to something else than `auto`.

web/src/components/ContrastThemeToggle.tsx
steffenkleinle: Actually, you are only reversing the icon for the kebab menu at the moment, right? I think you don't do that for the toolbar item. I think you should either reverse it for both cases or not at all. I think without reversing it actually looks more "reversed" than with.

lunars97: yes, you are right. Because of color switch, it gives the feeling as if it is switching. So I would just remove it as you suggest :+1:

web/src/components/base/Button.tsx
```diff
@@ -16,6 +16,7 @@ type ButtonProps = {
   label: string
   type?: 'submit' | 'button'
   disabled?: boolean
+  dir?: 'auto'
```
steffenkleinle, line 19:
```suggestion
```

web/src/components/base/Button.tsx
```diff
@@ -28,6 +29,7 @@ const Button = ({
   tabIndex,
   className,
   type = 'button',
+  dir = 'auto',
```
steffenkleinle, line 32:
```suggestion
```

web/src/components/base/Button.tsx
```diff
@@ -38,6 +40,7 @@ const Button = ({
     aria-label={label}
     tabIndex={tabIndex}
     type={type}
+    dir={dir}
```
steffenkleinle, line 43:
```suggestion
    dir='auto'
```

steffenkleinle (approved):
🚀

web/src/components/ContrastTheme.tsx
steffenkleinle, line 1: 🔧 I think this file should be deleted as its empty

PR #3202, opened 2025-04-15
Adjust header logo for contrast theme in web

lunars97:
### Short Description

Part of #3184 implementation. This PR adjust header logo to align with contrast theme toggle. When the theme changes to high contrast then the logo should change to yellow white logo.

### Proposed Changes

- Set `currentColor` to app-logo.tsx
- set StyledIcon instead of the img

### Side Effects
- should be none

### Testing
See whether the logo changes on toggle in web

### Resolved Issues

Fixes: #

---


steffenkleinle (commented):
I think it would be better to just reuse our existing icon as it is already an svg. You can just use our own `Icon` base component and set the color there. For this to work, you have to change all occurrences of `#1E1E1C` in `app-logo.svg` to `currentColor`. If you need help with this, let me know.

web/src/components/HeaderLogo.tsx
```diff
@@ -43,12 +37,26 @@ const LogoContainer = styled.div`
   }
 `
 
+const StyledLogoIcon = styled(Icon)`
+  fill: currentcolor;
```
steffenkleinle, line 41: 🔧 Not necessary
```suggestion
```

web/src/components/HeaderLogo.tsx
```diff
@@ -43,12 +37,26 @@ const LogoContainer = styled.div`
...
+  color: ${props => props.theme.colors.textColor};
+  height: 100%;
+  width: 200px;
+  max-width: 200px;
+
+  @media ${dimensions.smallViewport} {
+    max-width: 44px;
+    max-height: 100%;
```
steffenkleinle, line 49: 🔧 Not necessary
```suggestion
```

steffenkleinle (approved):
Tested on firefox 🖼️

build-configs/integreat/assets/app-logo.svg
```diff
@@ -1,29 +1,29 @@
-<svg width="297" height="80" viewBox="0 0 297 80" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg width="297" height="80" viewBox="0 0 297 80" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
LeandraH, line 1: That fill should stay none, though :)

Issue #3189, opened 2025-04-08
Adjust styles to align with contrast theme


lunars97:
### Describe the Problem

Styles in components should align with high contrast/ dark theme to be visually distinct according to the [figma design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=4617-1554&t=PLAUdrxOtPL8L0Dm-0)

### Describe Your Preferred Solution

Adjust styles (background-color/text color and etc.) where it is needed to align with figma design

steffenkleinle:
Done in #3209

PR #3209, opened 2025-04-17
3189: Adjust components styles to align with high contrast theme

lunars97:
### Short Description
This PR made adjustments to components styles on web to align with high contrast theme figma [design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=4617-1554&t=ujM5skW5hvstKvdj-0)

### Proposed Changes

- Mostly text color and background-color of different components were adjusted to have a distinct contrast

### Side Effects

- should be none

### Testing
Check whether the contrast theme color palettes follow the figma design suggestions and have enough contrast 

### Resolved Issues

Fixes: #3189 

---

### DIFF
--- a/web/src/components/CityNotCooperatingFooter.tsx
+++ b/web/src/components/CityNotCooperatingFooter.tsx
@@ -6,6 +6,7 @@ import styled from 'styled-components'
 import { CITY_NOT_COOPERATING_ROUTE, pathnameFromRouteInformation } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
+import { helpers } from '../constants/theme'
 import Icon from './base/Icon'
 
 const FooterContainer = styled.div`
@@ -26,7 +27,7 @@ const StyledIcon = styled(Icon)`
 const Button = styled(Link)`
   background-color: ${props => props.theme.colors.themeColor};
   text-decoration: none;
-  color: ${props => (props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor)};
+  ${helpers.adaptiveThemeTextColor}
   padding: 5px 20px;
   margin: 15px;
   text-align: center;
--- a/web/src/components/NewsListItem.tsx
+++ b/web/src/components/NewsListItem.tsx
@@ -17,10 +17,7 @@ const StyledLink = styled(Link)`
 const ReadMore = styled.div<{ $type: NewsType }>`
   align-self: flex-end;
   color: ${({ theme, $type }) => {
-    if (theme.isContrastTheme) {
-      return theme.colors.themeColor
-    }
-    if ($type === LOCAL_NEWS_TYPE) {
+    if (theme.isContrastTheme || $type === LOCAL_NEWS_TYPE) {
       return theme.colors.themeColor
     }
     return theme.colors.tunewsThemeColor
--- a/web/src/components/RemoteContent.tsx
+++ b/web/src/components/RemoteContent.tsx
@@ -63,9 +63,10 @@ const RemoteContent = ({ html, centered = false, smallText = false }: RemoteCont
     const currentSandBoxRef = sandBoxRef.current
 
     const allElements = currentSandBoxRef.querySelectorAll('*')
-    allElements.forEach(el => {
-      const element = el as HTMLElement
-      element.style.removeProperty('color')
+    allElements.forEach(element => {
+      if (element instanceof HTMLElement && element.style.color === 'rgb(0, 0, 0)') {
+        element.style.removeProperty('color')
+      }
     })
 
     const anchors = currentSandBoxRef.getElementsByTagName('a')
--- a/web/src/components/Chat.tsx
+++ b/web/src/components/Chat.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 import ChatMessageModel from 'shared/api/models/ChatMessageModel'
 
 import dimensions from '../constants/dimensions'
+import { helpers } from '../constants/theme'
 import useWindowDimensions from '../hooks/useWindowDimensions'
 import ChatConversation from './ChatConversation'
 import ChatSecurityInformation from './ChatSecurityInformation'
@@ -34,7 +35,7 @@ const SubmitContainer = styled.div`
 
 const SubmitButton = styled(TextButton)`
   flex: 1;
-  color: ${props => (props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor)};
+  ${helpers.adaptiveThemeTextColor}
 `
 
 const LoadingText = styled.div`
--- a/web/src/components/ChatContentWrapper.tsx
+++ b/web/src/components/ChatContentWrapper.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement, ReactNode } from 'react'
 import styled from 'styled-components'
 
 import dimensions from '../constants/dimensions'
+import { helpers } from '../constants/theme'
 import { ChatVisibilityStatus } from './ChatContainer'
 import ChatMenu from './ChatMenu'
 
@@ -22,7 +23,7 @@ const Header = styled.div<{ $small: boolean }>`
   display: flex;
   flex-direction: ${props => (props.$small ? 'row-reverse' : 'row')};
   justify-content: space-between;
-  color: ${props => (props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor)};
+  ${helpers.adaptiveThemeTextColor}
   font-size: ${props => props.theme.fonts.hintFontSize};
   font-weight: bold;
   align-items: center;
--- a/web/src/components/ChatMenu.tsx
+++ b/web/src/components/ChatMenu.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 
 import { ArrowBackIcon, CloseIcon, MaximizeIcon, MinimizeIcon } from '../assets'
 import dimensions from '../constants/dimensions'
+import { helpers } from '../constants/theme'
 import { ChatVisibilityStatus } from './ChatContainer'
 import Button from './base/Button'
 import Icon from './base/Icon'
@@ -19,8 +20,7 @@ const StyledIcon = styled(Icon)`
   color: ${props => props.theme.colors.backgroundColor};
 
   @media ${dimensions.smallViewport} {
-    color: ${props =>
-      props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor};
+    ${helpers.adaptiveThemeTextColor}
   }
 `
 
--- a/web/src/components/Note.tsx
+++ b/web/src/components/Note.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement } from 'react'
 import styled from 'styled-components'
 
 import { NoteIcon } from '../assets'
+import { helpers } from '../constants/theme'
 import Icon from './base/Icon'
 
 const NoteContainer = styled.div`
@@ -14,7 +15,7 @@ const NoteContainer = styled.div`
 
 const NoteText = styled.span`
   font-size: ${props => props.theme.fonts.decorativeFontSizeSmall};
-  color: ${props => (props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor)};
+  ${helpers.adaptiveThemeTextColor}
 `
 
 const StyledIcon = styled(Icon)`
--- a/web/src/constants/theme.ts
+++ b/web/src/constants/theme.ts
@@ -5,6 +5,7 @@ export type HelpersType = {
   adaptiveFontSize: RuleSet
   adaptiveWidth: RuleSet
   adaptiveHeight: RuleSet
+  adaptiveThemeTextColor: RuleSet
 }
 
 export const helpers: HelpersType = {
@@ -32,5 +33,9 @@ export const helpers: HelpersType = {
       ${props => props.theme.fonts.adaptiveFontSizeSmall.value},
       ${props => props.theme.fonts.adaptiveFontSizeSmall.max}
     );
+  `,
+  adaptiveThemeTextColor: css`
+    color: ${props =>
+      props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor};
   `,
 }
--- a/web/src/routes/CityNotCooperatingPage.tsx
+++ b/web/src/routes/CityNotCooperatingPage.tsx
@@ -6,6 +6,7 @@ import GeneralFooter from '../components/GeneralFooter'
 import Layout from '../components/Layout'
 import TextButton from '../components/base/TextButton'
 import buildConfig from '../constants/buildConfig'
+import { helpers } from '../constants/theme'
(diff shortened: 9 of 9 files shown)


build-configs/common/theme/colors.ts
```diff
@@ -46,7 +46,12 @@ export const commonLightColors = {
 export const commonHighContrastColors = {
   ...commonLightColors,
   backgroundAccentColor: '#20293A',
-  backgroundColor: '#101217',
   textColor: '#FFFFFF',
+  textSecondaryColor: '#FFFFFF',
```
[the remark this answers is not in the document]
lunars97: I included it because in a lot of places text-secondary is used and I did not want to do an extra step in writing a condition, so I added it to automatically change when it is contrast...

build-configs/common/theme/colors.ts
```diff
@@ -46,7 +46,12 @@ export const commonLightColors = {
...
+  backgroundColor: '#101217',
   borderColor: '#FFFFFF',
+  positiveHighlight: '#22c253',
+  negativeHighlight: '#FF3636',
+  invalidInput: '#FF3636',
```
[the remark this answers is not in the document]
lunars97: ohh I think I missed to change it when I was testing :see_no_evil:

web/src/App.tsx
```diff
@@ -18,14 +18,49 @@ import { useContrastTheme } from './hooks/useContrastTheme'
 import safeLocalStorage, { JPAL_TRACKING_CODE_KEY } from './utils/safeLocalStorage'
 import { initSentry } from './utils/sentry'
 
-const GlobalStyle = createGlobalStyle`
+const GlobalStyle = createGlobalStyle<{ isContrastTheme: boolean }>`
```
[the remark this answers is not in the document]
lunars97: good idea!

web/src/components/Footer.tsx
```diff
@@ -38,15 +40,18 @@ const FooterContainer = styled.footer<{ $overlay: boolean }>`
...
-      <span>
-        {__VERSION_NAME__}+{__COMMIT_SHA__}
-      </span>
-    )}
-  </FooterContainer>
-)
+const Footer = ({ children, overlay = false }: FooterProps): ReactElement => {
+  const { isContrastTheme } = useContrastTheme()
+  return (
+    <FooterContainer $isContrastTheme={isContrastTheme} $overlay={overlay}>
```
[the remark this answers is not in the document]
lunars97: thank you, works better :)

web/src/components/ToolbarItem.tsx
```diff
@@ -25,8 +26,16 @@ const StyledToolbarItem = styled(Link)<{ disabled?: boolean }>`
...
-  color: ${props => (props.disabled ? props.theme.colors.textDisabledColor : props.theme.colors.textSecondaryColor)};
+const StyledIcon = styled(Icon)<{ disabled?: boolean; $isContrastTheme: boolean }>`
+  color: ${props => {
+    if (props.disabled) {
+      return props.theme.colors.textDisabledColor
+    }
+    if (props.$isContrastTheme) {
+      return props.theme.colors.textColor
+    }
+    return props.theme.colors.textSecondaryColor
```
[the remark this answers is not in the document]
lunars97: yes :see_no_evil:

build-configs/common/theme/colors.ts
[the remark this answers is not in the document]
lunars97: yes, because it is used in quite a lot of places, so I just copied the style from textColor to have a distinct contrast

web/src/components/Tile.tsx
```diff
@@ -80,9 +91,15 @@ const Tile = ({ tile }: TileProps): ReactElement => {
...
       <>
-        <ThumbnailSizer>
+        <ThumbnailSizer $isContrastTheme={isContrastTheme}>
           <Thumbnail>
-            <img alt='' ref={imageRef} />
+            <img
+              alt=''
+              ref={imageRef}
+              style={{
+                filter: isContrastTheme ? 'invert(1) sepia(0) saturate(0) brightness(20.0) contrast(1.2)' : 'none',
```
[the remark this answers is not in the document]
lunars97: I included in parent element if it is okay

web/src/components/ThemeContext.tsx
```diff
@@ -8,7 +8,7 @@ import buildConfig from '../constants/buildConfig'
 type ThemeType = 'light' | 'contrast'
 
 export type ThemeContextType = {
-  theme: ReturnType<typeof buildConfig>['lightTheme'] | ReturnType<typeof buildConfig>['highContrastTheme']
+  theme: ReturnType<typeof buildConfig>['lightTheme'] | ReturnType<typeof buildConfig>['contrastTheme']
```
steffenkleinle, line 11: 🔧 The correct type here would be `ThemeType`. However, in order to use it, you need to rename `ThemeType` above in L8 to e.g. `ThemeName` or `ThemeKey` and import `ThemeType` from `build-configs`.

steffenkleinle (commented):
I found one more issue: The bottom action sheet handler is still white:
image

web/src/components/ThemeContext.tsx
lunars97: Good idea! I placed it in an extra file in build-configs...

steffenkleinle (approved):
Not tested again

web/src/components/CityNotCooperatingFooter.tsx
```diff
@@ -26,7 +26,7 @@ const StyledIcon = styled(Icon)`
 const Button = styled(Link)`
   background-color: ${props => props.theme.colors.themeColor};
   text-decoration: none;
-  color: ${props => props.theme.colors.textColor};
+  color: ${props => (props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textColor)};
```
LeandraH, line 29: ❓ This looks confusing to me, why would the color of the button text be a background color? Can we give that color a different name?

web/src/components/CategoryListItem.tsx
```diff
@@ -23,6 +23,7 @@ const CategoryThumbnail = styled.img`
   padding: 0 5px;
   flex-shrink: 0;
   object-fit: contain;
+  filter: ${props => (props.theme.isContrastTheme ? 'brightness(0) invert(1)' : 'none')};
```
LeandraH, line 26: ❓ Do we need the `brightness(0)` as well? Looks to me like `invert(1)` should already be enough?

web/src/components/Tile.tsx
```diff
@@ -51,6 +60,11 @@ const TileContainer = styled.div`
     cursor: pointer;
   }
 
+  & img {
+    filter: ${props =>
+      props.theme.isContrastTheme ? 'invert(1) sepia(0) saturate(0) brightness(20.0) contrast(1.2)' : 'none'};
```
LeandraH, line 65: image

I found the tiles at München to be a bit irregular, I got a more even result with `filter: invert(1) saturate(0) brightness(7)`

web/src/components/Tile.tsx
```diff
@@ -28,6 +28,15 @@ const ThumbnailSizer = styled.div`
   width: 150px;
   max-width: 33.3vw;
   margin: 0 auto;
+
+  & div:hover {
+    ${props =>
+      props.theme.isContrastTheme &&
+      `
+        outline: 8px solid ${props.theme.colors.themeColor};
```
LeandraH, line 36: 😍 That's a really nice touch!

web/src/components/__tests__/SearchListItem.spec.tsx
```diff
@@ -33,10 +34,10 @@ describe('SearchListItem', () => {
     )
 
     expect(getByLabelText(excerpt)).toBeTruthy()
-    expect(getByText(excerptBeforeQuery)).not.toHaveProperty('style', expect.objectContaining(highlightStyle))
-    expect(queryAllByText(query)[0]).toHaveProperty('style', expect.objectContaining(highlightStyle))
-    expect(queryAllByText(query)[1]).toHaveProperty('style', expect.objectContaining(highlightStyle))
-    expect(getByText(excerptAfterQuery)).not.toHaveProperty('style', expect.objectContaining(highlightStyle))
+    expect(getByText(excerptBeforeQuery)).not.toHaveStyle(highlightStyle)
```
LeandraH, line 37: ❓ If we're using the `_value`, can we rename it to `value`?

web/src/components/RemoteContentSandBox.ts
LeandraH, line 1: 🤔 We seem to have an issue here with municipalities manually giving text the color black like in the lower part here: http://localhost:9000/muenchen/de/sprache/deutschkurse/integrationskurse

I'm not sure if that's a problem for us to fix, though

LeandraH (approved):
This looks very cool already, nice job! Tested in Firefox

LeandraH:
Ah, one more thing, the space below the bottom sheet handler is still white
image

web/src/components/CityNotCooperatingFooter.tsx
steffenkleinle: Hmm, the problem here is basically that the text color for the contrast theme (white) has too little contrast on the theme color (yellow). Therefore, we set the text color to the background color here. This is basically only a workaround which should be history with the switch to the new design system. I therefore think that the effort to rename this would be too much.

steffenkleinle:
> Ah, one more thing, the space below the bottom sheet handler is still white

Hmm, I guess there should be space between the content and the toolbar such that the toolbar sticks to the bottom 🤔 Nothing introduced in this PR though and should therefore probably be fixed on main.

web/src/components/CategoryListItem.tsx
lunars97: good catch! :)

web/src/components/Tile.tsx
lunars97: looks way better :heart_hands:

bahaaTuffaha:
The text "IntegreatTestCms Chat Support" may need adjustment and the data protection statement popup background is white.

web/src/components/RemoteContent.tsx
```diff
@@ -61,6 +63,13 @@ const RemoteContent = ({ html, centered = false, smallText = false }: RemoteCont
       return
     }
     const currentSandBoxRef = sandBoxRef.current
+
+    const allElements = currentSandBoxRef.querySelectorAll('*')
+    allElements.forEach(el => {
+      const element = el as HTMLElement
+      element.style.removeProperty('color')
```
lunars97, line 70: I think this is the only way to remove the color from remote content :thinking:

web/src/components/NewsListItem.tsx
```diff
@@ -16,7 +16,15 @@ const StyledLink = styled(Link)`
...
 const ReadMore = styled.div<{ $type: NewsType }>`
   align-self: flex-end;
-  color: ${({ theme, $type }) => ($type === LOCAL_NEWS_TYPE ? theme.colors.themeColor : theme.colors.tunewsThemeColor)};
+  color: ${({ theme, $type }) => {
+    if (theme.isContrastTheme) {
+      return theme.colors.themeColor
+    }
+    if ($type === LOCAL_NEWS_TYPE) {
+      return theme.colors.themeColor
+    }
```
steffenkleinle, line 25:
```suggestion
    if (theme.isContrastTheme || $type === LOCAL_NEWS_TYPE) {
      return theme.colors.themeColor
    }
```

web/src/components/ContrastTheme.tsx
steffenkleinle, line 1: 🔧 Please delete the empty file

web/src/components/CityNotCooperatingFooter.tsx
steffenkleinle: @lunars97 using a helper as proposed in another comment by me for this would clarify and make this more readable imo. Still optional though.

web/src/components/RemoteContentSandBox.ts
```diff
@@ -15,6 +15,7 @@ const RemoteContentSandBox = styled.div<{ $centered: boolean; $smallText: boolea
     img {
     max-width: 100%;
     max-height: 100%;
+    filter: ${props => props.theme.isContrastTheme && 'invert(1)'};
```
steffenkleinle, line 18: ❓ I am actually not sure if we should do this for all images. I think its okay that standard images have not that good contrast but it would feel weird for them to be inverted imo.

web/src/components/RemoteContentSandBox.ts
```diff
@@ -171,6 +174,8 @@ const RemoteContentSandBox = styled.div<{ $centered: boolean; $smallText: boolea
     }
 
     img {
+      filter: ${props => props.theme.isContrastTheme && 'invert(1)'};
```
steffenkleinle, line 177: ❓ Same as above.

web/src/components/RemoteContent.tsx
steffenkleinle: I think this is a good approach 👍 However I think we should only remove this for text set to black and not in general.
See other suggestion.

web/src/components/RemoteContent.tsx
```diff
@@ -61,6 +61,13 @@ const RemoteContent = ({ html, centered = false, smallText = false }: RemoteCont
...
```
steffenkleinle, line 68: 🤔 What do you think of this solution? This avoids removing all colors which might be quite a big change for some regions.
```suggestion
      if (element instanceof HTMLElement && element.style.color === 'rgb(0, 0, 0)') {
        element.style.removeProperty('color')
      }
```

web/src/components/RemoteContentSandBox.ts
lunars97: I think it implies on all the possible affected formats of image

web/src/components/RemoteContent.tsx
lunars97: maybe we could change it only when it is contrast theme. Otherwise there might be some colors which might have less contrast and totally get blended with the background... :thinking:

lunars97: in the case above there is no check for contrast theme, so it will be applied for both of the themes. But I can implement your suggestion with checking theme type and it is better solution :+1:

web/src/components/NewsListItem.tsx
lunars97: :monkey: I missed it, they have the same color

web/src/components/RemoteContent.tsx
steffenkleinle: Hmm, not sure if the check for the theme is actually necessary if we only do it for black anyway? In the case of the light theme there should not be a difference if we remove black colors.

web/src/components/RemoteContentSandBox.ts
steffenkleinle: I think we should discuss this in integreat-general then 🤔 But the best option would probably to only use svgs for icons and not apply this for anything else.

web/src/components/RemoteContentSandBox.ts
lunars97: we will leave it as it is right now I guess to not block the PR?

LeandraH (approved):
Looks great! Not tested again

Issue #3144, opened 2025-03-06
Add app-toolbelt


steffenkleinle:
### Describe the Problem

Add app-toolbelt as dependency and make sure its possible to its scripts easily.

### Describe Your Preferred Solution

~~Please do some research on what is the best way to include this! It might be best to include it the same way as we do it for the entitlementcard, but I remember it being relatively complicated. Check back with the team on your findings and thoughts :)~~
Edit: For the entitlementcard we switched from globally to locally installing the app-toolbelt here: https://github.com/digitalfabrik/entitlementcard/pull/2023
Seems like this is the best approach atm.

Add app-toolbelt as dependency and make sure its possible to its scripts easily:
- Add app-toolbelt to integreat-app
- ~~Adjust installation documentation~~: Should not be necessary with a local installation as in https://github.com/digitalfabrik/entitlementcard/pull/2023
- Make sure it runs in the CI

❗ This is NOT about changing/replacing any scripts yet.

### Describe Possible Alternatives

N/A

### Related Issues

Blocked by: https://github.com/digitalfabrik/app-toolbelt/issues/19
Blocks: #3051 
Part of: #2644 

### Additional Information

app-toolbelt repo: https://github.com/digitalfabrik/app-toolbelt

steffenkleinle:
See https://github.com/digitalfabrik/entitlementcard/pull/2023 on how to add the app toolbelt locally. I think that is the best approach. @bahaaTuffaha or did you find out something else in the meantime?

bahaaTuffaha:
This is already included at #3051

Issue #3051, opened 2025-01-27
Replace scripts in integreat-app/tools with app-toolbelt


bahaaTuffaha:
### Describe the Problem

Remove the scripts in the /tools directory and replace their usages in the CI (and everywhere else) with the scripts from the app-toolbelt added in #3144.

### Describe Your Preferred Solution

- Remove scripts in the tools folder
- Replace usages in the CI with the app-toolbelt
- TBD: Delete tools folder afterwards

❗ Make sure that all scripts are actually replaced

### Describe Possible Alternatives

N/A

### Related Issues

Blocked by: #3144 
Part of: #2644 

### Additional Information

app-toolbelt repo: https://github.com/digitalfabrik/app-toolbelt

PR #3210, opened 2025-04-17
3144: Add app-toolbelt to tools dev dependencies

bahaaTuffaha:
### Short Description

Add app-toolbelt as dependency

### Proposed Changes

- Just added app-toolbelt as devDependencies
- This is already included in #3051

### Resolved Issues

Fixes: #3144

---


steffenkleinle (approved):

steffenkleinle:
Actually, since this is a one-liner, I think we should close this and just do it together in #3211 together (or rather: just close this as this is already included there).

Issue #3187, opened 2025-04-08
Save contrast mode in localStorage for web


lunars97:
### Describe the Problem
Contrast mode should be saved in local storage to be able to use the theme consistently during the interaction with the website

### Describe Your Preferred Solution

Add contrast mode to localStorage

steffenkleinle:
Fixed in #3212

PR #3212, opened 2025-04-17
3187: Save contrast theme in localStorage on web

lunars97:
### Short Description

This PR saves the state of the contrast theme in LocalStorage. 

### Proposed Changes

- used LocalStorage Hook to save the state ('light' or 'contrast')

### Side Effects

- none

### Testing
Go to http://localhost:9000/testumgebung/de and see state being persistent. You can also try to enable the dark/contrast theme on your browser and check the page whether it immediately detects the enabled theme on your browser and keeps it persistent throughout the interaction.

### Resolved Issues

Fixes: #3187 

---

### DIFF
--- a/web/src/components/ThemeContext.tsx
+++ b/web/src/components/ThemeContext.tsx
@@ -1,10 +1,11 @@
-import React, { createContext, ReactElement, useEffect, useMemo, useState } from 'react'
+import React, { createContext, ReactElement, useMemo } from 'react'
 import { ThemeProvider } from 'styled-components'
 
 import { ThemeType, ThemeKey } from 'build-configs'
 import { UiDirectionType } from 'translations'
 
 import buildConfig from '../constants/buildConfig'
+import useLocalStorage from '../hooks/useLocalStorage'
 
 export type ThemeContextType = {
   theme: ThemeType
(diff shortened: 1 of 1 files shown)


web/src/components/ThemeContext.tsx
```diff
@@ -30,12 +31,18 @@ type ThemeContainerProps = {
 }
 
 export const ThemeContainer = ({ children, contentDirection }: ThemeContainerProps): ReactElement => {
-  const [themeType, setThemeType] = useState<ThemeType>(() => getSystemTheme())
+  const { value: themeType, updateLocalStorageItem: setThemeType } = useLocalStorage<ThemeType>({
+    key: 'themeType',
```
steffenkleinle, line 35: 🙃 I think `theme` would be enough here. However, `themeType` is fine as well.
```suggestion
    key: 'theme',
```

web/src/components/ThemeContext.tsx
```diff
@@ -30,12 +31,18 @@ type ThemeContainerProps = {
...
 
   useEffect(() => {
     const handleSystemChange = () => {
-      const systemTheme = getSystemTheme()
-      setThemeType(prev => (prev !== systemTheme ? systemTheme : prev))
+      const storedTheme = localStorage.getItem('themeType')
+      if (!storedTheme || storedTheme === 'null') {
+        const currentTheme = getSystemTheme()
+        setThemeType(currentTheme)
+      }
```
steffenkleinle, line 45: 🔧 This is never executed as `useLocalStorage` directly persists the value to the local storage, so it is always set when you check for `storedTheme`. Removing this whole `useEffect` therefore still has the same behavior.
If you want to automatically update the theme if the browser settings change, just remove the `if`-clause.

steffenkleinle:
Tested on firefox and works perfectly 👍

web/src/components/ThemeContext.tsx
lunars97: I refactored it so the toggle state is saved as well

steffenkleinle: My message from above is still true. This will never be executed and can therefore be removed. 🤔

lunars97: no, this is needed to keep the state after refresh, without it, it will always go back to the system theme

steffenkleinle: I cannot reproduce this. For me this works just fine after deleting lines L43 to L58 :/ Can you please test again?

lunars97:
> Is it intended that the default theme type is contrast? @lunars97 I would tend to say `light` is the better choice

As Steffen stated above, if a user enables browser theme, then a user's browser theme should be applied automatically...

web/src/components/ThemeContext.tsx
lunars97: you are right, I think I missed something when I was testing it :see_no_evil:

Issue #3182, opened 2025-04-07
Chat input field loosing focus when polling


svenseeberg:
### Describe the Problem

The app polls the CMS every 8s for new chat messages, even for not yet started chats. For empty chats, the response takes a bit longer and the chat window is not usable during that time. At the same time the textbox is loosing focus after each request. This only happens for empty chats, existing chats are not a problem.

### Describe Your Preferred Solution

Polling should start after the first message has been sent. That means we poll once the chat window opens. If no chat history exists, polling should not start. If a chat history exists, we start polling. If the first message is sent, the app immediately gets a response with the chat history and can activate polling.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

Priority is not that high right now.

Tracking in https://github.com/digitalfabrik/integreat-chat/issues/319

svenseeberg:
> Priority is not that high right now.

Well it might be a bit more important, as it is currently very tedious to start a new chat, especially when using a smartphone :S

hannaseithe:
When we pull from the backend too early (before the chat is started) the backend returns a response of `{error: "Chat not found"}` (which is correct according to @svenseeberg since the chat hasnt started at that point). Which in return triggers our condition of `isLoading={chatMessagesReturn === null && (loading || sendingStatus === 'sending')}` in ChatController.ts, which then shows the loadingscreen for the chat

![Image](https://github.com/user-attachments/assets/0427ed7b-91be-4a0f-ad87-c036635ccc6f)

PR #3214, opened 2025-04-18
3182: Prevent polling before sending any messages

bahaaTuffaha:
### Short Description
The app polls the CMS every 8s for new chat messages, even for not yet started chats. For empty chats, the response takes a bit longer and the chat window is not usable during that time. At the same time the textbox is loosing focus after each request. This only happens for empty chats, existing chats are not a problem.

### Proposed Changes

~~- I did an extra step and moved the `LoadingContainer` to just replace the `StyledChatConversation` component while loading so it won't interrupt the user while typing the message at the input field and I disabled sending while it loads.~~
- For `ChatController` I checked for number of messages to enable refreshing interval.
- Moved the `isBrowserTabActive` condition inside a new function to stop refreshing while browser is inactive even when the interval is triggered.
### Side Effects

- None other `chatController.tsx` and `Chat.tsx` are affected.

### Testing

- Replace `refreshMessages()` with `console.log("Refreshed")`  to know when it refreshes.
~~- You can see how loading covers the chat only by passing true for the ` isLoading` prop at chat component.~~

### Resolved Issues

Fixes: #3182 
Fixes: https://github.com/digitalfabrik/integreat-chat/issues/319

---


web/src/components/ChatController.tsx
```diff
@@ -37,12 +37,18 @@ const ChatController = ({ city, language }: ChatControllerProps): ReactElement =
...
-    if (isBrowserTabActive) {
-      const pollMessageInterval = setInterval(refreshMessages, POLLING_INTERVAL)
+    const messageCount = chatMessages?.messages.length ?? 0
+    const RefreshMethod = () => {
+      if (isBrowserTabActive) {
+        refreshMessages()
+      }
+    }
+    if (messageCount > 0) {
+      const pollMessageInterval = setInterval(RefreshMethod, POLLING_INTERVAL)
```
steffenkleinle, line 47: 🤔 Why not just
```suggestion
    const messageCount = chatMessages?.messages.length ?? 0
    if (isBrowserTabActive && messageCount > 0) {
      const pollMessageInterval = setInterval(refreshMessages, POLLING_INTERVAL)
```

web/src/components/Chat.tsx
```diff
@@ -72,28 +72,27 @@ const Chat = ({ messages, submitMessage, hasError, isLoading }: ChatProps): Reac
...
+          <StyledLoadingSpinner />
+          <LoadingText>{t('loadingText')}</LoadingText>
+        </LoadingContainer>
+      ) : (
+        <StyledChatConversation
+          $height={deviceHeight - chatInputContainerHeight}
+          messages={messages}
+          hasError={hasError}
+        />
+      )}
```
steffenkleinle, line 95: ❓ Why did you change this? This leads to jumpy behavior as the chat input is already shown while loading and is only pushed down once it has finished loading.

web/src/components/Chat.tsx
```diff
@@ -107,7 +106,11 @@ const Chat = ({ messages, submitMessage, hasError, isLoading }: ChatProps): Reac
           />
         </InputSection>
         <SubmitContainer>
-          <SubmitButton disabled={textInput.length === 0} onClick={onSubmit} text={t('sendButton')} />
+          <SubmitButton
+            disabled={textInput.length === 0 || hasError || isLoading}
```
steffenkleinle, line 110: 🔧 This logic is duplicated, please extract to a constant:
```ts
  const submitDisabled = textInput.trim().length === 0 || hasError || isLoading
  const submitOnEnter = (event: KeyboardEvent) => {
    const isEnterAllowed = event.key === 'Enter' && !event.shiftKey && !submitDisabled
    ...
```
```suggestion
            disabled={submitDisabled}
```

release-notes/unreleased/3182-prevent-polling-before-first-message.yml
```diff
@@ -0,0 +1,5 @@
+issue_key: 3182
+show_in_stores: false
+platforms: # relevant platforms, possible values: web, android and ios
+  - web
+en: Improvements to the integreat Chat Support.
```
steffenkleinle, line 5: 🤔 I think this release note is redundant since it is not shown to our users and does not provide any information to devs or other teams as it is really unspecific. I think it can just be removed.
Generally, `Integreat` should always be uppercased and the feature is just called `Integreat Chat` without `Support`.

steffenkleinle (changes requested):

web/src/components/ChatController.tsx
bahaaTuffaha: If the interval got triggered you can't stop 'refreshMessages()' from running when isBrowserTabActive is not active 🤔

Unless we clear the interval with return...

web/src/components/ChatController.tsx
```diff
@@ -37,12 +37,13 @@ const ChatController = ({ city, language }: ChatControllerProps): ReactElement =
   const isBrowserTabActive = useIsTabActive()
 
   useEffect(() => {
-    if (isBrowserTabActive) {
-      const pollMessageInterval = setInterval(refreshMessages, POLLING_INTERVAL)
-      return () => clearInterval(pollMessageInterval)
+    const messageCount = chatMessagesReturn?.messages.length ?? 0
+    if (!isBrowserTabActive || messageCount === 0) {
+      return
```
bahaaTuffaha, line 42: This return should invoke the clean up for useEffect.

web/src/components/Chat.tsx
steffenkleinle: At least the current implementation makes the UI jump, so either revert it or try to fix it :)

web/src/components/ChatSecurityInformation.tsx
```diff
@@ -21,8 +23,8 @@ const SecurityIconContainer = styled.button`
 `
 
 const SecurityIcon = styled(Icon)`
-  width: 32px;
-  height: 32px;
+  width: 100%;
+  height: 100%;
```
steffenkleinle, line 27: 🤔 Why did you change this?

steffenkleinle (commented):
❓ Why did you revert the change that you check whether the user actually typed a message/the message is empty? I think that was a good addition.

web/src/components/ChatSecurityInformation.tsx
bahaaTuffaha: You can download and slow down the video.

https://github.com/user-attachments/assets/4ee4ec1f-f826-4a33-8938-24b344b3e88b

svenseeberg:
Is there a chance to have this merged and deployed before the 14th of May? This fix would be extremely helpful for our next planned tests :)

steffenkleinle (approved):
Works as expected, thanks!

svenseeberg:
For what its worth: in my tests it works as well :D

LeandraH (approved):
Nice job! Tested in Firefox

PR #3219, opened 2025-04-23
Suggestion for improving debounce on search page

bahaaTuffaha:
### Short Description

Suggestion for improving debounce on search page at #3137 

- Used a state called `draftSearchText` to keep the search seperate from SearchParams and filtering.
- This will keep delaying while you type so it wont lag while typing.
- Used useEffect instead of `handleFilterTextChanged` to not change the url every change at the SearchInput and depends on debouncedQuery.
---

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -26,3 +26,5 @@ If we verify a reported security vulnerability, our policy is:
 
 - A security advisory will be released on the project website detailing the
   vulnerability, as well as recommendations for end-users to protect themselves.
+
+For more information, please read our [full policy](https://security.tuerantuer.org/policy.txt).
--- a/build-configs/aschaffenburg/index.ts
+++ b/build-configs/aschaffenburg/index.ts
@@ -32,7 +32,7 @@ const commonAschaffenburgBuildConfig: CommonBuildConfigType = {
     floss: false,
     pois: false,
     newsStream: true,
-    pushNotifications: true,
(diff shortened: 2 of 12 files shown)


steffenkleinle:
@bahaaTuffaha can you please explain what you did here/what it does improve?

bahaaTuffaha:
> @bahaaTuffaha can you please explain what you did here/what it does improve?

I updated the description.

steffenkleinle (commented):
Nice, this actually feels a lot snappier as the text in the search input is directly updated. Does this also work without the `useEffect`, i.e. by direclty updating it in `onFilterTextChange`?
Also, please use the `useSearchParams` api instead of manually navigating (see updates in #3137).

bahaaTuffaha:
> Does this also work without the `useEffect`, i.e. by direclty updating it in `onFilterTextChange`?

I think this useEffect is needed because it listens to debouncedQuery to update url params.

bahaaTuffaha:
I did some tests with and without `memo` : its snappier on long text using memo

https://github.com/user-attachments/assets/13e03bc7-9b9a-48e4-86b8-2d5b5c3d1e2a

steffenkleinle:
I updated #3137 with a few of your suggestions, can you check again there please if that solves your issues?

steffenkleinle:
Closed in favor of #3137

Issue #3154, opened 2025-03-17
Check languages supported for tts


steffenkleinle:
### Describe the Problem

For native, we currently have an array of three languages which are unsupported:
https://github.com/digitalfabrik/integreat-app/blob/ade679252b4c9591406f701c4e478204f5e5284a/native/src/components/TtsContainer.tsx#L15

This list might not be complete and might differ between android and ios.

### Describe Your Preferred Solution

We should check every language we support on whether it is supported or not. For supported languages tts should be enabled. If there are differences between android and ios, the language should only be enabled on the corresponding OS.

### Describe Possible Alternatives

N/A

### Related Issues

Related to: #2916, #2917, #2918

### Additional Information

https://cloud.google.com/text-to-speech/docs/list-voices-and-types#list_of_all_supported_languages

PR #3221, opened 2025-04-24
3154: Check supported languages for TTS

bahaaTuffaha:
### Short Description

We should check every language we support on whether it is supported or not. For supported languages tts should be enabled. If there are differences between android and ios, the language should only be enabled on the corresponding OS.

### Proposed Changes

- Checking Tts.voices() for available languages and return true if current language is supported.
- Used this useEffect to just check for language support only when I change language.
- Added `enabled` flag to `play` function's if statement to prevent users from playing unsupported language while the tts is showed up from previous supported language.

Note: I didn't test this on iOS.

### Side Effects

None.

### Testing

- Switch to language that is not supported like Persian or Kurmanji.
- Read-aloud button should not appear on these unsupported languages.

### Resolved Issues

Fixes: #3154 

---

### DIFF
--- a/native/src/components/TtsContainer.tsx
+++ b/native/src/components/TtsContainer.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, ReactElement, useCallback, useContext, useMemo, useState } from 'react'
+import React, { createContext, ReactElement, useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import Tts, { Options } from 'react-native-tts'
 
@@ -12,8 +12,6 @@ import useSnackbar from '../hooks/useSnackbar'
 import { reportError } from '../utils/sentry'
 import TtsPlayer from './TtsPlayer'
 
-// For a list of available languages see https://cloud.google.com/text-to-speech/docs/list-voices-and-types#list_of_all_supported_languages
-const TTS_UNSUPPORTED_LANGUAGES = ['fa', 'ka', 'kmr']
 const TTS_OPTIONS: Options = {
   androidParams: {
     KEY_PARAM_PAN: 0,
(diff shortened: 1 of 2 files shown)


native/src/components/TtsContainer.tsx
```diff
@@ -50,12 +48,35 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
   const [sentences, setSentences] = useState<string[]>([])
+  const [isLanguageSupported, setIsLanguageSupported] = useState(true)
   const { languageCode } = useContext(AppContext)
   const { t } = useTranslation('layout')
   const showSnackbar = useSnackbar()
   const title = sentences[0] || t('nothingToRead')
   const shortTitle = truncate(title, { maxChars: TTS_MAX_TITLE_DISPLAY_CHARS })
-  const enabled = buildConfig().featureFlags.tts && !TTS_UNSUPPORTED_LANGUAGES.includes(languageCode)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
steffenkleinle: The following works for me:
``` ts
  const { data: voices } = useLoadAsync(Tts.voices)
  const isLanguageSupported = voices && voices.some(({ language }) => language.split('-')[0] === languageCode)
  const enabled = buildConfig().featureFlags.tts

  const showTtsPlayer = useCallback((): void => {
    if (!enabled || visible) {
      return
    }
    if (!isLanguageSupported) {
      showSnackbar({ text: t('languageNotSupported') })
      return
    }

```

native/src/components/TtsContainer.tsx
```diff
@@ -110,7 +131,7 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
     async (index = sentenceIndex) => {
       const safeIndex = Math.max(0, index)
       const sentence = sentences[safeIndex]
-      if (sentence) {
+      if (sentence && enabled) {
```
steffenkleinle, line 134: 🔧 This should not be needed if other suggestions are applied.

bahaaTuffaha: This is needed even with applied changes:
`if (sentence && isLanguageSupported) {` 

Changing language while the player is still open will let the user play even if not supported so I added this flag.

steffenkleinle: The player should be closed if the new language is not supported. Otherwise there is no feedback to the user why the tts feature is disabled.

native/src/components/TtsContainer.tsx
```diff
@@ -110,6 +115,11 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
     async (index = sentenceIndex) => {
       const safeIndex = Math.max(0, index)
       const sentence = sentences[safeIndex]
+      if (!isLanguageSupported) {
+        close()
+        showSnackbar({ text: t('languageNotSupported') })
+        return
+      }
```
steffenkleinle, line 122: 🤔 I think this is the wrong place for this, in this case a `useEffect` with the only dependency `isLanguageSupported` feels like the better choice. At least it does not belong into the `play` function (the play function should not be called in the first place if the language is not supported).
🔧 I think showing a snackbar in this case is confusing to users as the context is missing and a general "your language is not supported" is not true in general. Therefore I would just silently close the player.

bahaaTuffaha: To test this please:
- Switch to the English or German language (supported language).
- Open read-aloud function from kebab menu.
- Switch to unsupported language like Persian.
- here you can notice that the player is open and when you press play it will close() then show a snack bar in this case.

bahaaTuffaha: I had to add visible and close to the dependency because eslint is complaining and visible is used here to prevent useEffect from calling `close()` at the first run and mess with how many times `stop()` has been called at `TtsContainer.spec.tsx`.

https://github.com/digitalfabrik/integreat-app/blob/fd7276a131844e92269d3a83758596600135ef6c/native/src/components/TtsContainer.tsx#L144-L148

steffenkleinle (approved):
Nice solution, works as expected 🎉 I think we can then finally release the tts feature 🚀

f1sh1918 (approved):
Tested on ios. Looks good 👍

Issue #3240, opened 2025-04-29
Directly open chat via query param


steffenkleinle:
### Describe the Problem

In order to make it easier for users to use the chat, it would be nice to directly open the chat via URL.

### Describe Your Preferred Solution

Add a query param `?chat=true` which opens up the chat.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3242, opened 2025-04-29
3240: Open chat via query param

steffenkleinle:
### Short Description

Allow directly opening the chat via query param.

### Proposed Changes

- Automatically open the chat if the query param `?chat=true` is set
- Remove the query param after initial rendering

### Side Effects

N/A

### Testing

Test that the chatbot opens automatically if appending `?chat=true` to any url, e.g.
localhost:9000/testumgebung/en?chat=true
http://localhost:9000/testumgebung/en/events?chat=true
Check that other query params are not removed:
http://localhost:9000/testumgebung/en/events?chat=true&asdf=test

### Resolved Issues

Fixes: #3240

---


web/src/components/ChatContainer.tsx
```diff
@@ -1,7 +1,10 @@
-import React, { ReactElement, useContext, useState } from 'react'
+import React, { ReactElement, useContext, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { useSearchParams } from 'react-router-dom'
 import styled from 'styled-components'
 
+import { CHAT_QUERY_KEY, parseQueryParams } from 'shared/dist'
```
bahaaTuffaha, line 6:
```
WARNING in ./components/ChatContainer.tsx 81:28-42
export 'CHAT_QUERY_KEY' (imported as 'CHAT_QUERY_KEY') was not found in 'shared/dist' 
```

```suggestion
import { CHAT_QUERY_KEY, parseQueryParams } from 'shared'
```

bahaaTuffaha (approved):
There is a small issue but its good overall.
Tested on chrome and Firefox ✅
Tested with other params like query it looks good ✅

bahaaTuffaha:
And you need to fix `ChatContainer.spec.tsx` 'should close chat if close button was clicked' test is complaining.
I think it needs `MemoryRouter` .

shared/routes/query.ts
LeandraH, line 1: 🔧 I'd probably also add a small test :)

web/src/components/ChatContainer.tsx
```diff
@@ -61,16 +64,25 @@ type ChatContainerProps = {
 }
 
 const ChatContainer = ({ city, language }: ChatContainerProps): ReactElement => {
-  const { t } = useTranslation('chat')
-  const [chatVisible, setChatVisible] = useState(false)
+  const [searchParams, setSearchParams] = useSearchParams()
```
LeandraH, line 67: 🙃 I know they're technically called search params but can we call the state here query params instead, just for future me? So, `const [queryParams, setQueryParams] = useSearchParams()`

Issue #3185, opened 2025-04-08
Add toggle button for high contrast mode for native


lunars97:
### Describe the Problem
Add toggle button for high contrast mode for native app. Here is [the figma design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=4617-1602&t=eUdYoUHdsSYF1l5N-0)
### Describe Your Preferred Solution
Add toggle button for it on app side menu

connium:
Hint for testing: restart the bundler

steffenkleinle:
Done in #3185

PR #3215, opened 2025-04-21
3185: Add toggle implementation for contrast mode for native

lunars97:
### Short Description

This PR adds toggle implementation for contrast theme. The contrast theme toggle function is located in settings according to the figma design.

### Proposed Changes

- ThemeContext.tsx was implemented and added to App.tsx
- useThemeContext.tsx hook was added to further use it in Header.tsx as a settings toggle function

### Side Effects

-  none

### Testing

- Run the native app
- See the toggle function in settings
- Toggle the `Contrast mode` and see that the theme changes to contrast/dark

### Resolved Issues

Fixes: #3185 

---

PR #3245, opened 2025-04-30
3185: Add toggle implementation for contrast mode for native

steffenkleinle:
### Short Description

This PR adds toggle implementation for contrast theme. The contrast theme toggle function is located in settings according to the figma design.

### Proposed Changes

- ThemeContext.tsx was implemented and added to App.tsx
- useAppContext was used to save the theme state 
- themeContrast section was added to createSettingsSection.tsx
- description section for contrastTheme was added to settings

### Side Effects

-  none

### Testing

- Run the native app
- See the toggle function in settings
- Toggle the `Contrast mode` and see that the theme changes to contrast/dark

### Resolved Issues

Fixes: #3185

Reopened from https://github.com/digitalfabrik/integreat-app/pull/3215 to split web and native implementations.
Credit to @lunars97.

---


native/src/components/Header.tsx
```diff
@@ -196,6 +199,7 @@ const Header = ({
           ? [renderOverflowItem(HeaderButtonTitle.Location, () => navigation.navigate(LANDING_ROUTE))]
           : []),
         renderOverflowItem(HeaderButtonTitle.Settings, () => navigation.navigate(SETTINGS_ROUTE)),
+        ...[renderOverflowItem(t(HeaderButtonTitle.ContrastTheme), toggleTheme)],
```
steffenkleinle, line 202: ❓ Why did you choose to move this to the header overflow menu? I think this should go to the app settings.

lunars97: It was the previous version, so I have not yet implemented it

native/src/contexts/__tests__/AppContextProvider.spec.tsx
```diff
@@ -96,6 +97,7 @@ describe('AppContextProvider', () => {
       jpalTrackingCode: '123456',
       jpalSignals: [],
       externalSourcePermissions: { 'https://vimeo.com': true },
+      selectedTheme: 'light' as ThemeKey,
```
steffenkleinle, line 100: 🔧 We want to avoid typecasts whenever possible. This becomes redundant if you change L88 to:
``` ts
    const settings: SettingsType = {
```

native/src/routes/Settings.tsx
```diff
@@ -21,7 +21,7 @@ type SettingsProps = {
 const Settings = ({ navigation }: SettingsProps): ReactElement => {
   const appContext = useCityAppContext()
   const showSnackbar = useSnackbar()
-  const { t } = useTranslation('settings')
+  const { t } = useTranslation<['settings', 'layout']>(['settings', 'layout'])
```
steffenkleinle, line 24: 🙃 Is the type annotation actually necessary?
```suggestion
  const { t } = useTranslation(['settings', 'layout'])
```
Edit: If you do `t('layout:my-key')` anyway, the previous `useTranslation('settings')` would be enough. You only need to pass the additional namespace if you want to use `layout:my-key` like this: `t('my-key')`. I think in this case the better choice would be to do it explicitly, so you can just revert this here:
```suggestion
  const { t } = useTranslation('settings')
```

translations/translations.json
```diff
@@ -7054,6 +7054,7 @@
       "settings": "Einstellungen",
       "copyUrl": "Link kopieren",
       "contrastTheme": "Kontrast Modus",
+      "contrastThemeDescription": "Mit dem Kontrastmodus wird die Lesbarkeit Ihrer Anwendung verbessert",
```
steffenkleinle, line 7057: 🔧 We use `du` everywhere in the app. Perhaps the following would be even simpler though:
```suggestion
      "contrastThemeDescription": "Mit dem Kontrastmodus wird die Lesbarkeit verbessert",
```

translations/translations.json
```diff
@@ -7093,6 +7094,7 @@
       "settings": "ሴቲንጎች",
       "copyUrl": "አገናኝን ቅዳ",
       "contrastTheme": "የንፅፅር ሁነታ",
+      "contrastThemeDescription": "የንፅፅር ሁነታ የመተግበሪያዎን ተነባቢነት ያሻሽላል።",
```
steffenkleinle, line 7097: ❓ Where do you get all these translations from? We usually don't want to add them ourselves except for our mother languages (+ english). Instead, these should be translated by professional translators later on.

steffenkleinle:
Looks good otherwise, I sadly can't approve as this is my own PR :D Tested on android.

translations/translations.json
lunars97: got you :+1:

translations/translations.json
steffenkleinle: Sorry for the effort you put into this and that I did not spot this with the contrast theme. But then please delete the translations again and also those for `contrastTheme` :/

lunars97: yes, I think it is really important to include what you wrote above otherwise it is not much clear..

lunars97: I think it makes it clear, thank you :+1:

steffenkleinle:
I think we are good to merge this :)

LeandraH:
I'm still getting the error, even with a fresh `yarn start` and cleaned build folder in Xcode :(

steffenkleinle:
> I'm still getting the error, even with a fresh `yarn start` and cleaned build folder in Xcode :(

I can't reproduce this, could you try 

> I'm still getting the error, even with a fresh `yarn start` and cleaned build folder in Xcode :(

I cannot reproduce this :/ Can you try running `yarn ts:check` and then restarting the bundler argain? I think this also works for @bahaaTuffaha and @lunars97 correctly, right? I therefore suspect an issue in your setup.

bahaaTuffaha:
I can't reproduce the issue on Android...

PR #3250, opened 2025-05-01
1021: Enable contrast mode for malte, aschaffenburg and obdach

steffenkleinle:
### Short Description

Enable contrast mode for malte, aschaffenburg and obdach.

### Proposed Changes

- Set contrast mode colors for malte, aschaffenburg and obdach
- Set malte logo text color to text color
- Use icon instead of logo for aschaffenburg (since the text is otherwise doubled and not readable in contrast mode)

### Side Effects

- The text color of the malte logo is now slightly different in light theme (`#000000` instead of `#1D1D1B`)
- The aschaffenburg logo is replaced with the icon also for desktop

### Testing

Check that both themes look good for all build configs.

### Resolved Issues

Fixes: N/A

---


lunars97 (approved):
Thank you for your PR! :yellow_heart:

PR #3251, opened 2025-05-01
1021: Only invert svg colors

steffenkleinle:
### Short Description

Currently the color inversion filter is applied on all images which leads to inverted "normal" images.
Therefore only apply the filter on svgs.

### Proposed Changes

- Apply color inversion only on svgs

### Side Effects

- There might be edge cases of svgs that should not be inverted and other images that should be inverted

### Testing

Check icons and images, e.g
http://localhost:9000/testumgebung/en/welcome/welcome-to-city/integreat-in-sign-language-video
http://localhost:9000/testumgebung/en/welcome/important-offices/jobcenter

### Resolved Issues

Fixes: N/A

---


LeandraH (commented):
Nice! I think only inverting the SVGs does most of the trick, but there's this annoying yellow icon...

image

This is on http://localhost:9000/muenchen/en/welcome-to-our-website

I'm also not sure what to do about a Ukraine flag:

image

http://localhost:9000/augsburg/en/current-topics

steffenkleinle:
> Nice! I think only inverting the SVGs does most of the trick, but there's this annoying yellow icon...
> image
> 
> This is on http://localhost:9000/muenchen/en/welcome-to-our-website
> 
> I'm also not sure what to do about a Ukraine flag:
> image
> 
> http://localhost:9000/augsburg/en/current-topics

I don't think there is really a solution for this at the moment. Since this PR improves the situation nevertheless (as at least pngs/jpgs are not miscolored anymore), I still think this is an improvement and should be merged.

Short term solution: Replace svgs that should not be recolored with pngs in the cms
Mid term solution: Add `fill="currentColor"` to all paths in the svgs that should be recolored and use the color attribute instead of a filter to recolor the svgs.

Issue #3207, opened 2025-04-16
Wrap poi contact section in a collapsible


steffenkleinle:
### Describe the Problem

We currently allow users to show/hide individual contacts in poi details. While this allows for a little better overview, it is tedious to collapse every single contact and still takes more space than it would need to.

### Describe Your Preferred Solution

It should be possible to show/hide the whole contact section in poi details using our `Collapsible` component with the title `Contact`.

### Describe Possible Alternatives

N/A

### Related Issues

Multiple contacts were implemented in https://github.com/digitalfabrik/integreat-app/issues/3121.
More info (opening hours) will be added in #3162.

### Additional Information

It is planned to add even more information to contacts (opening hours, #3162).

See e.g. http://localhost:9000/testumgebung/en/locations/lr
<img src="https://github.com/user-attachments/assets/4366625f-0cdc-4b94-9e72-591914110196" width="150px" />

PR #3253, opened 2025-05-02
3207: Wrap POI in collapsible

bahaaTuffaha:
### Short Description

We currently allow users to show/hide individual contacts in poi details. While this allows for a little better overview, it is tedious to collapse every single contact and still takes more space than it would need to.

### Proposed Changes

- Used collapsible component but noticed there is no difference between higher level collapsible and its children so I passed a prop called `isParent` and increased the size of the font.
- Added new `adaptiveMediumFontSize` to helpers for web to increase the size when isParent is true.
- Increased padding at 'CollapseHeaderWrapper'.

### Side Effects

- The whole POI details could be affected.

### Testing

- For web and native: go to map (poi) http://localhost:9000/testumgebung/en/locations
- Check all different locations that has different details.

### Resolved Issues

Fixes: #3207 

---


bahaaTuffaha:
What do you think? Is there a better way to do this? @hauf-toni 
-  I moved the address out of contacts collapsible.
- Increased the size of the label when I use collapsible as parent.

steffenkleinle (changes requested):
Thanks! This already looks pretty good. Two things though:
- The `contacts` collapsible should only contain the contacts, not e.g. the content (which is currently included on both web and native)
- The `contacts` title is way too big imo. I think the better approach would be to keep the same font size, but make the children (i.e. the names of the contacts in this case) non-bold, see attached screenshot
image

hauf-toni:
> Thanks! This already looks pretty good. Two things though:
> 
> * The `contacts` collapsible should only contain the contacts, not e.g. the content (which is currently included on both web and native)
> * The `contacts` title is way too big imo. I think the better approach would be to keep the same font size, but make the children (i.e. the names of the contacts in this case) non-bold, see attached screenshot
> 
> image

looks great! one thing: I think it makes sense to only have a collapsible for "contact" and not for every contact itself. in most cases, we won't have a lot of contacts here and users should be able to see all the contacts with one click instead of having to open "contacts" and then having to open each individual contact as well.

web/src/components/Collapsible.tsx
```diff
@@ -7,10 +7,10 @@ import { helpers } from '../constants/theme'
 import Button from './base/Button'
 import Icon from './base/Icon'
 
-const Container = styled.div`
+const Container = styled.div<{ $isGapEnabled: boolean }>`
```
steffenkleinle, line 10: 🔧 I think this is not the best solution sadly :/ Introducing a separate prop for one style only is quite complicated. Instead, it would be better to just revert the gap here and overwrite the style where you use the collapsible and do not want the gap. You can achieve this as follows:
- Add a new prop `className` and pass it to this styled component `Container` in L57
-  Overwrite the style in web/src/PoiDetails as follows:
``` ts
const StyledCollapsible = styled(Collapsible)`
  gap: 0;
`
```

steffenkleinle (approved):
Nice work!

LeandraH (approved):
Nicely done! Tested in Firefox and emulated iOS

Issue #3195, opened 2025-04-09
Broken Toolbar At POI


bahaaTuffaha:
### Describe the Bug

For some languages the toolbar's container not fitting the ToolbarItems.

### Steps to Reproduce

1. Go to [http://localhost:9000/testumgebung/de/locations](http://localhost:9000/testumgebung/de/locations)
2. Look at the toolbar.

### Expected Behavior

It shouldn't overflow the toolbar container.

### Additional Information

https://github.com/user-attachments/assets/c3c0a8e8-27df-45ab-ab58-f2a5541156de

bahaaTuffaha:
> I can't reproduce this in Firefox or Chrome :(

what language did use? you can try dutch or español.

LeandraH:
German, like in the example. I also couldn't reproduce it in Spanish or any other language that I tried (we don't have Dutch, but you probably meant Deutsch, which is the German word for German?) Maybe it depends on the OS? In which case, @steffenkleinle might be able to reproduce the issue?

steffenkleinle:
> German, like in the example. I also couldn't reproduce it in Spanish or any other language that I tried (we don't have Dutch, but you probably meant Deutsch, which is the German word for German?) Maybe it depends on the OS? In which case, [@steffenkleinle](https://github.com/steffenkleinle) might be able to reproduce the issue?

Yes, I can reproduce this (both on macos and linux). 

![Image](https://github.com/user-attachments/assets/1d7e8bee-9e31-4431-a1f0-00950a3b7220)

LeandraH:
Sorry, turns out, I can reproduce this too, but only on local, not on production.

steffenkleinle:
> Sorry, turns out, I can reproduce this too, but only on local, not on production.

The tts is feature is not yet enabled on production.

PR #3258, opened 2025-05-06
3195: Fix displaying of toolbar and toolbar items

steffenkleinle:
### Short Description

The toolbar got very crowded and broke things in a few places. This PR improves this by applying multiple changes.

### Proposed Changes

- Change the order of toolbar items (pdf > feedback > share > tts > high contrast)
- Move copy url to share toolbar item
- Set a `max-width` and allow text wrapping for toolbar items
- Display less toolbar items for desktop pois to allow fitting in one row (hide the contrast mode for pois and tts for pois list)
- Remove/reduce aside left spacing to avoid overlap
- Fix contrast of footer on pois desktop
- Fix overlap of footer and zoom buttons on pois desktop

### Side Effects

The contrast mode can't be toggled anymore from pois web.

### Testing

Check the toolbar items in different languages/pages/screen sizes, e.g.:
- http://localhost:9000/testumgebung/de
- http://localhost:9000/testumgebung/mk/locations

### Resolved Issues

Fixes: #3195

---

### DIFF
--- a/web/src/styles/MapView.css
+++ b/web/src/styles/MapView.css
@@ -15,4 +15,5 @@
 .maplibregl-ctrl-top-left,
 .maplibregl-ctrl-top-right {
   z-index: 0;
+  margin-bottom: 48px;
 }
--- a/web/src/components/CityContentToolbar.tsx
+++ b/web/src/components/CityContentToolbar.tsx
@@ -1,7 +1,7 @@
-import React, { memo, ReactNode, useContext, useState } from 'react'
+import React, { memo, ReactNode, useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { CopyIcon, DoneIcon, ReadAloudIcon } from '../assets'
+import { ReadAloudIcon } from '../assets'
 import useWindowDimensions from '../hooks/useWindowDimensions'
 import { RouteType } from '../routes'
 import ContrastThemeToggle from './ContrastThemeToggle'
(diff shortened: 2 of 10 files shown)


web/src/styles/MapView.css
```diff
@@ -15,4 +15,5 @@
 .maplibregl-ctrl-top-left,
 .maplibregl-ctrl-top-right {
   z-index: 0;
+  margin-bottom: 48px;
```
f1sh1918, line 18: ❓ can you just give me hint why you added this margin

steffenkleinle:
> Looks good except of two issue.
> 
> First one is minor The attribution on the map can't be read in dark mode

Good catch!

> Second one should be fixed. The changing of the order resulted in sharing options can't be read

Thanks, do you think it makes sense to just make the popup open vertically instead of horizontally on mobile?

web/src/styles/MapView.css
steffenkleinle: This fixes the overlap of footer and zoom buttons on pois desktop with smaller (but not yet mobile) screen widths, see e.g. here:
image

f1sh1918:
> > Looks good except of two issue.
> > First one is minor The attribution on the map can't be read in dark mode
> 
> I can't reproduce this, can you share a little bigger section of the page? Since its on the bottom, it looks more like its the footer?
> 
> > Second one should be fixed. The changing of the order resulted in sharing options can't be read
> 
> Thanks, do you think it makes sense to just make the popup open vertically instead of horizontally on mobile?

i think there was an issue before when i tried to do it vertically but you can give it a try

Its just the attribution of the map that can't be read because the overlay is gray and the font color is white. Therefore the overlay should be more darken

f1sh1918 (approved):
Tested on chrome. LGTM 👍

Issue #3220, opened 2025-04-24
Push poi detail toolbar to bottom


steffenkleinle:
### Describe the Problem

Currently the toolbar for poi details is shown directly below the poi detail/content with additional space below.
This is especially visible with the contrast mode being implemented (see https://github.com/digitalfabrik/integreat-app/pull/3209#issuecomment-2823542453).

### Describe Your Preferred Solution

The toolbar should stick to the bottom with additional space between.

### Describe Possible Alternatives

Just color the background anyway.

### Related Issues

N/A

### Additional Information

http://localhost:9000/testumgebung/en/locations/berlin
https://private-user-images.githubusercontent.com/18513943/436445485-fbab57db-3558-4440-b99c-563c8b1b2688.png

PR #3259, opened 2025-05-07
3220: Push toolbar to bottom in bottom action sheet

steffenkleinle:
### Short Description

Push toolbar to bottom in bottom action sheet to avoid

### Proposed Changes

- Fix the height of the bottom action sheet
- Push the toolbar down
- Remove css for bottom sheet in favor of global style
- Use global styles also for light theme

### Side Effects

~~Scrolling to top/bottom still produces some white background. Was not able to figure out how to prevent it (without breaking things in the light mode). Perhaps someone has ideas?~~

### Testing

Go to http://localhost:9000/testumgebung/fi/locations/berliini, activate the dark mode and check that there is no white space below the toolbar and the toolbar sticks to the bottom.

### Resolved Issues

Fixes: #3220

---


f1sh1918 (approved):
hey @steffenkleinle,
i think your solution looks good. An alternative would be just to style container via GlobalStyle.tsx

this would result to this
image

You could just add this to global styles
```
          [data-rsbs-scroll='true'] {
          background-color: ${props.theme.colors.backgroundAccentColor};
        }
```

bahaaTuffaha (commented):
It doesn't close completely 😐️

bahaaTuffaha (approved):
Nice work ✅

PR #3260, opened 2025-05-07
Limit search results for performance

bahaaTuffaha:
### Short Description

#3137 Still a bit laggy due rendering a large list with `.map()`.

### Proposed Changes

- Limits number of search result to an amount that can snapper on search and doesn't affect the searching experience.
- Adjusted SearchListItem and SearchFeedback to use debouncedQuery.

~Note: There is another option is to use React-window or React-virtualized that could render the whole list without issues but it needs a fixed sized height.~

### Side Effects

None.

### Testing

- Go to http://localhost:9000/muenchen/en/search
- Test if its snapper typing on the search without lags.

### DIFF
--- a/native/src/routes/SearchModal.tsx
+++ b/native/src/routes/SearchModal.tsx
@@ -3,7 +3,15 @@ import { useTranslation } from 'react-i18next'
 import { KeyboardAvoidingView, Platform } from 'react-native'
 import styled from 'styled-components/native'
 
-import { parseHTML, SEARCH_FINISHED_SIGNAL_NAME, SEARCH_ROUTE, SearchResult, useSearch, useDebounce } from 'shared'
+import {
+  parseHTML,
+  SEARCH_FINISHED_SIGNAL_NAME,
+  SEARCH_ROUTE,
+  SearchResult,
+  useSearch,
+  useDebounce,
+  MAX_SEARCH_RESULTS,
+} from 'shared'
 
 import FeedbackContainer from '../components/FeedbackContainer'
 import List from '../components/List'
@@ -54,9 +62,10 @@ const SearchModal = ({
   const contentLanguageReturn = useSearch(documents, debouncedQuery)
   const fallbackLanguageReturn = useSearch(fallbackLanguageDocuments, debouncedQuery)
   const searchResults = contentLanguageReturn.data.concat(fallbackLanguageReturn.data)
+  const LimitedResults = searchResults.slice(0, MAX_SEARCH_RESULTS)
 
   useReportError(contentLanguageReturn.error ?? fallbackLanguageReturn.error)
-  useAnnounceSearchResultsIOS(searchResults)
+  useAnnounceSearchResultsIOS(LimitedResults)
 
   const onClose = (): void => {
     sendTrackingSignal({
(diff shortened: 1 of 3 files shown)


native/src/routes/SearchModal.tsx
```diff
@@ -54,9 +62,10 @@ const SearchModal = ({
   const contentLanguageReturn = useSearch(documents, debouncedQuery)
   const fallbackLanguageReturn = useSearch(fallbackLanguageDocuments, debouncedQuery)
   const searchResults = contentLanguageReturn.data.concat(fallbackLanguageReturn.data)
+  const LimitedResults = searchResults.slice(0, MAX_SEARCH_RESULTS)
```
steffenkleinle, line 65: 🔧 This should not be uppercase

web/src/routes/SearchPage.tsx
```diff
@@ -98,18 +106,20 @@ const SearchPage = ({ city, cityCode, languageCode }: CityRouteProps): ReactElem
     if (loading) {
       return <LoadingSpinner />
     }
+    const LimitedResults = results.slice(0, MAX_SEARCH_RESULTS)
```
steffenkleinle, line 109: 🔧 Same as above

steffenkleinle (commented):
Is there any reason why you decided to do the slicing twice in web and native separately? Wouldn't it be easier to just directly slice the results in `useSearch` in shared? That way, there is no need for duplication and no need for separate variables. If we ever need a different amount of search results depending on web/native, we can just introduce a new param then.

bahaaTuffaha:
> Is there any reason why you decided to do the slicing twice in web and native separately? Wouldn't it be easier to just directly slice the results in `useSearch` in shared? That way, there is no need for duplication and no need for separate variables. If we ever need a different amount of search results depending on web/native, we can just introduce a new param then.

useSearch is a good idea but there is fallbackLanguageReturn will concatenate with contentLanguageReturn and give double the amount so we have 2 options maybe:
1- slice separately as I did.
2- slice at `useSearch` and reduce the amount to 37 or 38.

native/src/routes/SearchModal.tsx
```diff
@@ -102,7 +111,7 @@ const SearchModal = ({
                   routeType={SEARCH_ROUTE}
                   language={languageCode}
                   cityCode={cityCode}
-                  noResults={searchResults.length === 0}
+                  noResults={limitedResults.length === 0}
                   query={query}
```
steffenkleinle, line 115: 🤔 Should we use `debouncedQuery` here as well?

native/src/routes/SearchModal.tsx
```diff
@@ -54,9 +62,10 @@ const SearchModal = ({
   const contentLanguageReturn = useSearch(documents, debouncedQuery)
   const fallbackLanguageReturn = useSearch(fallbackLanguageDocuments, debouncedQuery)
   const searchResults = contentLanguageReturn.data.concat(fallbackLanguageReturn.data)
+  const limitedResults = searchResults.slice(0, MAX_SEARCH_RESULTS)
```
steffenkleinle, line 65: 🙃 `limited` sounds a little like they are of insufficient quality. Not sure how to name it better. We could also just do the following though, not sure if thats better though. Same for web.
```suggestion
  const searchResults = contentLanguageReturn.data.concat(fallbackLanguageReturn.data).slice(0, MAX_SEARCH_RESULTS)
```

steffenkleinle (approved):
Way smoother now, thanks! Great idea 🏎️

native/src/routes/SearchModal.tsx
steffenkleinle: I think if it makes things slower on web it should be the same here, right? In general I think its good to have consistent behavior anyway.

bahaaTuffaha: Wait its used already:
https://github.com/digitalfabrik/integreat-app/blob/2323f7bc781cd160bbe8815c0396e4313776e1eb/native/src/routes/SearchModal.tsx#L61-L63

steffenkleinle: yes, but not for the highlight/feedback 🤔

LeandraH (approved):
Good idea! I'm not sure I like that we're effectively lying to our users by telling them there are "only" 75 search results ever, but I don't think anybody really goes looking for results that long.

Tested on Firefox

Issue #3262, opened 2025-05-09
Unknown workspace "tools" error when starting native


bahaaTuffaha:
### Describe the Bug

An error `Unknown workspace "tools"` occurs when I run `yarn start` at native.

```bash
yarn run v1.22.22
$ yarn start:integreat-test-cms
$ node ./run packager integreat-test-cms
error Unknown workspace "tools".
node:child_process:966
    throw err;
    ^

Error: Command failed: yarn workspace --silent tools --silent app-toolbelt v0 build-config to-bash integreat-test-cms common
error Unknown workspace "tools".

    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    
...
```

### Steps to Reproduce

1. Start native.
2. Look at the error.

### Related Issues

Related to https://github.com/digitalfabrik/integreat-app/pull/3211

PR #3263, opened 2025-05-09
3262: Add tools workspace to the monorepo

bahaaTuffaha:
### Short Description

An error `Unknown workspace "tools"` occurs when I run yarn start at native.

### Proposed Changes

- Moved app-toolbelt package to root package.json.
- Removed `/tools`
- Removed `restore_yarn_tools_cache.yml`
- Changed `--cwd tools app-toolbelt` to ``--cwd . app-toolbelt`
- Changed `working_directory: tools` to `working_directory: .`
- Changed `yarn workspace --silent tools --silent app-toolbelt` to `yarn --silent app-toolbelt`

### Side Effects

- None.

### Testing

- Check if native runs without issues for Android and iOS.
- Check that CircleCi jobs are working correctly.

### Resolved Issues

Fixes: #3262 

---


steffenkleinle (changes requested):
As mentioned in the call, this will significantly increase CI runtime as we then need to install all node modules instead of just the tools whenever we want to use the app-toolbelt (2 minutes vs 12 seconds). Therefore, I think that this is not a good solution.
An alternative could be (but I have not tested it):
- Also add the app-toolbelt as dependency to the root package.json and use that for all build-config scripts that now use `yarn workspace tools`.
- Keep the tools directory separate from the workspace and use that for all standalone scripts not related to the build config (release, version bump, ...)

steffenkleinle (changes requested):
Please read my comment again, this is very different from what I proposed. The tools dir/node modules should not be deleted. Instead, app-toolbelt should only be added additionally to the root.

.circleci/src/jobs/bump_version.yml
```diff
@@ -13,18 +13,18 @@ steps:
   - run:
       name: Calculate next version name
       command: echo "export NEW_VERSION_NAME=$(yarn --silent app-toolbelt v0 version calc | jq .versionName)" >> ${BASH_ENV}
-      working_directory: tools
+      working_directory: .
```
steffenkleinle, line 16: ❌ Please revert all these changes in the CI. It should continue to use the tools app-toolbelt.

.circleci/src/jobs/deliver_android.yml
```diff
@@ -25,7 +25,7 @@ steps:
       directory: native/android
   - run:
       name: Prepare Play Store metadata
-      command: yarn --cwd tools app-toolbelt v0 release-notes prepare-metadata << parameters.build_config_name >> playstore
+      command: yarn --cwd . app-toolbelt v0 release-notes prepare-metadata << parameters.build_config_name >> playstore
```
steffenkleinle, line 28: ❌ Please revert. I think L21 to restore the root node modules might be deleted now as well if you adjust `util.rb` like this:
``` ruby
yarn --cwd ../tools --silent app-toolbelt v0 build-config to-json #{build_config_name} #{platform}
```

steffenkleinle (changes requested):
I just realized that the problem is/can be solved quite simple actually:
Only change the build config commands in `run` to:
``` bash
yarn --cwd tools --silent app-toolbelt v0 build-config ...
```
Same for all other commands that do `yarn workspace [--silent] tools`
Everything else can most likely be reverted.

bahaaTuffaha:
> I just realized that the problem is/can be solved quite simple actually: Only change the build config commands in `run` to:
> 
> ```shell
> yarn --cwd tools --silent app-toolbelt v0 build-config ...
> ```
> 
> Same for all other commands that do `yarn workspace [--silent] tools`. Everything else can most likely be reverted.

so I just have to revert everything and replace `yarn --cwd` with `yarn workspace` and keep the app-toolbelt package at tools?

steffenkleinle:
I created a new PR here which I tested both locally and in the CI: https://github.com/digitalfabrik/integreat-app/pull/3270

bahaaTuffaha:
Closed due to better implementation at: https://github.com/digitalfabrik/integreat-app/pull/3270

Issue #3152, opened 2025-03-12
Rewrite tests for SearchListItem


steffenkleinle:
### Describe the Problem

We currently have two disabled tests for SearchListItem.

### Describe Your Preferred Solution

Rewrite the tests for SearchListItem such that it is correctly tested and no tests are disabled.

### Describe Possible Alternatives

Delete the tests.

### Related Issues

N/A

### Additional Information

https://github.com/digitalfabrik/integreat-app/blob/ade679252b4c9591406f701c4e478204f5e5284a/native/src/components/__tests__/SearchListItem.spec.tsx#L68
https://github.com/digitalfabrik/integreat-app/blob/ade679252b4c9591406f701c4e478204f5e5284a/native/src/components/__tests__/SearchListItem.spec.tsx#L81

PR #3264, opened 2025-05-11
3152: Rewrite tests for SearchListItem

bahaaTuffaha:
### Short Description

We currently have two disabled tests for SearchListItem.
Rewrite the tests for SearchListItem such that it is correctly tested and no tests are disabled.

### Proposed Changes

- For "should show excerpt around query if only match in content" at native the title always bold so I changed the `assertHighlighting` to true.
- I added new test to check  for the tracking signal for native.
- I added new test for checking if thumbnail is rendered when provided for both web and native.

### Side Effects

None.

### Testing

- Run `yarn test` for native and web.

### Resolved Issues

Fixes: #3152

---


web/src/components/SearchListItem.tsx
```diff
@@ -71,7 +71,7 @@ const SearchListItem = ({ title, contentWithoutHtml, query, path, thumbnail }: S
       <StyledLink to={path}>
         <CategoryItemContainer dir='auto'>
           <CategoryTitleContainer>
-            {!!thumbnail && <CategoryThumbnail alt='' src={thumbnail} />}
+            {!!thumbnail && <CategoryThumbnail alt='thumbnail' src={thumbnail} />}
```
steffenkleinle, line 74: 🔧 I think this is not something we should do, this worsens a11y. Please revert.

steffenkleinle (approved):

native/src/components/__tests__/SearchListItem.spec.tsx
```diff
@@ -65,25 +79,55 @@ describe('SearchListItem', () => {
...
-  it.skip('should show excerpt around query if only match in content', () => {
+  it('should show excerpt around query if only match in content', () => {
     const excerptBeforeQuery = 'This is a'
     const query = 'sample'
     const excerptAfterQuery = 'page'
 
     const { getByText } = renderSearchListItem(query)
 
-    assertHighlighting(getByText(category.title), false)
+    assertHighlighting(getByText(category.title), true) // Titles are bold all the time from HighlighterCategoryTitle
```
LeandraH, line 89: 💭 Maybe we should remove the assertions for the title being bold? It doesn't seem like something that needs to be tested

LeandraH (commented):
Awesome, I'm so happy that someone finally fixed this, thank you!

However, one of them is still failing for me, this is after running `yarn test` in the root:

image

LeandraH (approved):

steffenkleinle:
@bahaaTuffaha can this be merged?

Issue #3233, opened 2025-04-29
Setup material ui


steffenkleinle:
### Describe the Problem

The new design system uses material design v3 which we need to implement.

### Describe Your Preferred Solution

Install and setup mui (material ui) for web: https://mui.com/material-ui/getting-started/installation/
Since styled-components is deprecated, use the default with emotion instead.
⚠ Fonts will be configured in a separate issue as we do not use the default roboto:

### Describe Possible Alternatives

Use another material design library.

### Related Issues

N/A

### Additional Information

N/A

steffenkleinle:
Done in #3265

Issue #3234, opened 2025-04-29
Configure color palette


steffenkleinle:
### Describe the Problem

Our new design system uses a custom color scheme/palette for integreat, malte and aschaffenburg that is different from material ui's standard color palette: [Figma](https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=0-1&p=f&t=9HPV7E2352PxLKbD-0)

### Describe Your Preferred Solution

Configure material ui on web to use the corresponding color palette.
Share as much as possible between the different build configs.
Make sure that the colors for the contrast mode also work correctly.

### Describe Possible Alternatives

N/A

### Related Issues

Blocked by: #3233 

### Additional Information

https://mui.com/material-ui/customization/color/
https://mui.com/material-ui/customization/palette/
https://mui.com/material-ui/customization/dark-mode/

steffenkleinle:
Done in #3265

PR #3265, opened 2025-05-12
3233/3234: Setup mui and color palette

steffenkleinle:
### Short Description

This PR deprecates our old themes, adds material ui and its peer dependencies for web and sets up the new color palette for all build configs.
This should not be a breaking change.

### Proposed Changes

- Deprecate the old theme/colors by renaming to legacy
- Add mui to web
- Add the new color palette for all build configs (a few values have to be clarified with UI/UX still)

### Side Effects

I added a test button. Might be removed at a later point.

### Testing

See the material UI button.
Feel free to check it out and play around with material ui.

### Resolved Issues

Fixes: #3233, #3234

---

### DIFF
--- a/build-configs/ThemeType.ts
+++ b/build-configs/ThemeType.ts
@@ -1,0 +1,40 @@
+export type SimplePaletteColor = {
+  main: string
+  light?: string
+}
+
+export type PaletteColor = SimplePaletteColor & {
+  light: string
+  dark: string
+  contrastText: string
+}
+
+export type CommonColors = {
+  primary: PaletteColor
+  tertiary: PaletteColor
+  error: PaletteColor
+  warning: SimplePaletteColor
+  success: SimplePaletteColor
+  info?: SimplePaletteColor
+  tunews: SimplePaletteColor
+  divider: string
+  link: string
+}
+
+export type CommonColorPalette = CommonColors & {
+  mode: 'light' | 'dark'
+  surface: PaletteColor
+  text: {
+    primary: string
+    secondary: string
+    disabled: string
+  }
+}
+
+export type ThemeColorPalette = CommonColorPalette & {
+  secondary: PaletteColor
+}
+
+export type Theme = {
+  palette: ThemeColorPalette
+}
--- a/web/package.json
+++ b/web/package.json
@@ -29,8 +29,12 @@
     "postinstall": "npm-license-crawler -onlyDirectDependencies -json ./src/assets/licenses.json"
   },
   "dependencies": {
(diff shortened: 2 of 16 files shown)


build-configs/ThemeType.ts
```diff
@@ -1,7 +1,40 @@
-import { ColorsType } from './common/theme/colors'
-import { FontsType } from './common/theme/fonts'
+export type SimplePaletteColor = {
```
steffenkleinle, line 1: I decided to recreate the types myself to enable cross platform compatibility (as mui is not available for native). I am not totally convinced this is the right way to go, but seemed like the best approach for now.

web/package.json
```diff
@@ -29,8 +29,12 @@
     "postinstall": "npm-license-crawler -onlyDirectDependencies -json ./src/assets/licenses.json"
   },
   "dependencies": {
+    "@emotion/react": "^11.14.0",
+    "@emotion/styled": "^11.14.0",
```
bahaaTuffaha, line 33: Do we need emotion here? or this is a preparation for migrating to emotion too?

web/src/App.tsx
```diff
@@ -32,19 +36,31 @@ const App = (): ReactElement => {
...
-        </>
-      </I18nProvider>
+      <ThemeProvider
+        theme={createTheme({
+          colorSchemes: {
+            light: buildConfig().lightTheme,
+            dark: buildConfig().darkTheme,
+          },
+        })}>
+        <Button startIcon={<RocketLaunch color='secondary' />}>Test</Button>
```
bahaaTuffaha, line 51: Do we need that?

bahaaTuffaha (approved):
LGTM ✅ (left some observations)

web/src/App.tsx
steffenkleinle: We do not need this, its more a proof of concept that mui works as expected. I'll remove it later.

web/package.json
steffenkleinle: Mui uses emotion as styling library per default, see also https://mui.com/material-ui/getting-started/installation/#default-installation. Since support for styled components is legacy and the whole library is deprecated, I think it makes lots of sense to migrate to emotion during the style system implementation.

LeandraH (approved):
🚀 This looks great!

build-configs/ThemeType.ts
steffenkleinle: True, but installing that library does not work/does not make sense for web (and vice versa with mui). This is a shared workspace so installing a web only/native only solution does not work.

Issue #3266, opened 2025-05-12
Search function in Integreat is not showing all of the results


LauraReichert:
### Short description

LK Höxter has problems with the search function in Integreat.
If user search for the word "Minijob" via the magnifier, the main page (https://integreat.app/kreishoexter/de/arbeit-und-bildung/arbeit/arbeitsvertrag) is not shown as a result.
The word "Minijob" is used as a headline/ title. It seems like words in the title are not found, e.g. it also does not work for "unbefristet".

### Steps to Reproduce

1. Go to https://integreat.app/kreishoexter/de
2. Click on the magnifier to search for a word
3. Type in "minijob"
4. The main content page regarding "Minijobs" (https://integreat.app/kreishoexter/de/arbeit-und-bildung/arbeit/arbeitsvertrag) is not shown

### Expected Behavior

All pages with the word "minijob" are shown (regardless of whether it is a headline or a normal text)

LauraReichert:
@steffenkleinle is the status of the issue completed/ done and it should work again?
Just checked it out and I still don't get the right results for "minijob"..

steffenkleinle:
> [@steffenkleinle](https://github.com/steffenkleinle) is the status of the issue completed/ done and it should work again? Just checked it out and I still don't get the right results for "minijob"..

This is done, but not released yet. I just triggered a new delivery/release to beta. Once this is tested, we can release to production.

steffenkleinle:
@LauraReichert this should now be fixed, see https://integreat.app/kreishoexter/de/search?query=minijob.

PR #3268, opened 2025-05-13
3266: Fix missing search results due to problems with html parsing

steffenkleinle:
### Short Description

Fix missing search results by improving html parsing and normalizing the complete content.
`processTerm` only normalizes one chunk of the content a time.

### Proposed Changes

- Normalize/parse the whole content
- Fix parsing of `<br>` and `<p>` tags

### Side Effects

N/A

### Testing

See issue description. Verify that the shown excerpt works as expected.

### Resolved Issues

Fixes: #3266 

---


shared/hooks/useSearch.ts
```diff
@@ -47,7 +50,8 @@ const useSearch = (documents: SearchResult[], query: string): UseSearchReturn =>
     new MiniSearch({
       idField: 'path',
       fields: ['title', 'content'],
-      storeFields: ['title', 'content', 'path', 'thumbnail'],
```
bahaaTuffaha, line 50: Does `storeFields` optimize performance in some way?

bahaaTuffaha (approved):
Nice work ✅ tested on Android and web.

shared/hooks/useSearch.ts
```diff
@@ -69,8 +73,10 @@ const useSearch = (documents: SearchResult[], query: string): UseSearchReturn =>
     }
   }, [indexing, search, documents])
 
+  const results: string[] = search.search(normalizedQuery).map(result => result.id)
+
   return {
-    data: normalizedQuery.length === 0 ? documents : (search.search(normalizedQuery) as unknown as SearchResult[]),
+    data: normalizedQuery.length === 0 ? documents : documents.filter(document => results.includes(document.path)),
```
LeandraH, line 79: 🤔 It seems a little extra work to me to map the search results to their IDs and then filter the documents for the ones that have those IDs instead of immediately using the search results. What's the advantage here?

LeandraH (approved):
Nice, thanks for the quick fix! Tested in Firefox

Also, oops, we really should have caught that one before 😅

shared/hooks/useSearch.ts
steffenkleinle: We lowercase/normalize the content of the search results above. Using those directly therefore leads to the excerpt of search results being displayed lowercased/normalized.

LeandraH: Ah, of course, thanks for the explanation :)

Issue #3043, opened 2025-01-22
Improve release notes


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
Improve our release notes the same way we do in the entitlementcard project.

**Describe the solution you'd like**
Github release notes should be nicely formatted, include the contributors as well as the corresponding PRs/issues.
Keep in mind that release notes to the stores have to continue to work while being understandable to all users.

**Describe alternatives you've considered**
Keep everything as it is.

**Additional context**
Look into using the Toolbelt function for this:
Blocked by #2644

LeandraH:
Linked to #2644

PR #3283, opened 2025-05-21
3043: Improve release notes and fix release notify

steffenkleinle:
### Short Description

Use a single notification job and create a single github release with automatic release notes from github (for all platforms).

### Proposed Changes

- Fix the release creation by only passing `--production-release` flag if true (commander seems to not like passing `false`)
- Create only a single github release for all platforms (or web and native for platform specific deliveries)
- Only notify once in mattermost for deliveries
- Move release notes also for native only deliveries (as they are only used for native)
- Only notify once in mattermost for promotions

### Side Effects

- Web release notes are not used anymore

### Resolved Issues

Fixes: #3043

---


LeandraH (approved):
Nice!

bahaaTuffaha (approved):
So you combined them into notify_release.yml 🤔?
Nice work ✅

PR #3285, opened 2025-05-22
3199: Simplify styling

steffenkleinle:
### Short Description

This proposal actually aligns the search input and the description below it without the need for any calculations with widths. Also, it increases usability on small devices by removing the excessive padding.


bahaaTuffaha:
Thank you !
I will merge it to my branch and adjust some small things like the padding 🙂

Issue #3289, opened 2025-05-23
Fix high contrast mode bugs


bahaaTuffaha:
### Describe the Bug

- The close (x) button at TtsHelpModal is white when high contrast mode is activated.
- The Data protection popup at the Integreat-chat is in dark background when high contrast mode is disabled.

### Steps to Reproduce

1. Go to any page that has content and click on Read aloud feature (make sure the browser isn't supported or choose unsupported language to get the `TtsHelpModal` ) .
2.  Data protection popup its located at Integreat-chat (lock icon/button) click on it while high contrast mode is disabled.

### Additional Information

<img src="https://github.com/user-attachments/assets/d1a661bf-43ac-45b6-90ff-75f2a0b66f44" width="auto" />

<img src="https://github.com/user-attachments/assets/14b30191-6576-4c2f-b8b6-88d0a05ee4d0" width="300px" />

<img src="https://github.com/user-attachments/assets/bda6e6c1-714f-48b1-82c3-c46b0ec72087" width="300px" />

PR #3295, opened 2025-05-28
3289: Fix high contrast mode bugs

bahaaTuffaha:
### Short Description

- The close (x) button at TtsHelpModal is white when high contrast mode is activated.
- The Data protection popup at the Integreat-chat is in dark background when high contrast mode is disabled.

### Proposed Changes

- Due to the background of the `TtsHelpModal` being light-yellow even in contrast mode, changing the color of the back button at `ModalContent` will affect another component `PoiFilters`'s back button that has a dark background so I had to apply the theme color to the `TtsHelpModal`.
- Fixed the backgroundColor for InformationTooltipContainer at `ChatSecurityInformation.tsx`.
- I was bothered by the Data protection popup still open even if clicked outside of the div so I added a simple handler called `onMouseLeave` to close this popup when the mouse leaves it.

### Side Effects

-  `ModalContent.tsx`, `TtsHelpModal.tsx`

### Testing

- Go to any 3rd leaf content and open the tts/read aloud (make sure you are on unsupported browser or missing voices).
- For the Data protection popup you need to open the integreat-chat then click on the lock button.
- Test that `ModalContent`'s title is behaving correctly on contrast mode and light mode.

### Resolved Issues

Fixes: #3289 

---


web/src/components/ChatSecurityInformation.tsx
```diff
@@ -50,7 +49,12 @@ const ChatSecurityInformation = (): ReactElement => {
 
   return (
     <SecurityInformationContainer>
-      {securityInformationVisible && <InformationTooltipContainer>{t('dataSecurity')}</InformationTooltipContainer>}
+      {securityInformationVisible && (
+        // eslint-disable-next-line styled-components-a11y/no-static-element-interactions
+        <InformationTooltipContainer onMouseLeave={() => setSecurityInformationVisible(false)}>
+          {t('dataSecurity')}
+        </InformationTooltipContainer>
```
steffenkleinle, line 56: ❌  I am not sure whether this is a good change. I feel like this is unexpected behavior for users. Have you thought of using our `useOnClickOutside` functionality instead? I feel like this solves the problem pretty elegant. Compare to e.g. `HeaderActionItemDropdown`.

steffenkleinle (commented):
Tested on firefox, looks good. I think the data protection information behavior should still be improved though.

steffenkleinle (approved):
Tested on firefox

web/src/components/ChatSecurityInformation.tsx
steffenkleinle: https://github.com/digitalfabrik/integreat-app/pull/3320

PR #3300, opened 2025-06-03
2499: Reduce LCP time by 20 percent

bahaaTuffaha:
### Short Description

Google still thinks our mobile performance is too slow, i.e. we have a LCP longer than 2.5 on most sites on mobile. Therefore we should try to speed up our web app more.

### Proposed Changes

- Reduced LCP time by around 20% by using TerserPlugin's minimizer.

### Side Effects

- It will compress everything at production so functions will be harder to read while inspecting JavaScript files.

### Testing
- Before testing at `webpack.config.ts` change `minimize: !devServer,` to  `minimize:true` to work on dev.
- `yarn build`
- Open chrome -> f12 then look for lighthouse from the tabs above.
- Most my tests were done on the following settings: Navigation(default), device:Mobile, all categories selected.
- Check LCP time before applying these changes and after. 

### Resolved Issues

Partly fixes: #2499 

---


web/tools/webpack.config.ts
```diff
@@ -189,6 +191,17 @@ const createConfig = (
...
       usedExports: true,
       runtimeChunk: 'single',
+      minimize: !devServer,
+      minimizer: [
+        new TerserPlugin({
+          parallel: true,
+          terserOptions: {
+            compress: {
+              drop_console: true,
+            },
```
bahaaTuffaha, line 201: This will drop console logs from production.

web/tools/webpack.config.ts
```diff
@@ -13,6 +13,8 @@ import 'webpack-dev-server'
 import loadBuildConfig, { WEB } from 'build-configs'
 import { WebBuildConfigType } from 'build-configs/BuildConfigType'
 
+const TerserPlugin = require('terser-webpack-plugin')
```
steffenkleinle, line 16: 🔧 Please use `import` over `require` here. You probably need to add `terser-webpack-plugin` to `devDependencies` for this.

steffenkleinle:
> Before testing at webpack.config.ts change minimize: !devServer, to minimize:true to work on dev.

Why is this necessary? With `yarn build` `devServer` should be `false` already.

steffenkleinle (approved):
Nice improvement!

steffenkleinle (approved):

f1sh1918:
hm what server do you use? 
I use `http-server` if i server the build i can not reload the page and getting 404.
To run lighthouse i need this feature. Any idea how to fix this @bahaaTuffaha

steffenkleinle:
> > > hm what server do you use? I use `http-server` if i server the build i can not reload the page and getting 404. To run lighthouse i need this feature. Any idea how to fix this @bahaaTuffaha
> > 
> > 
> > You can just test this with `yarn start` and run lighthouse there, no need to actually build it.
> 
> hm i don't think then the resulting lighthouse values are a real indicator tbh but okay

We can just test this on webenxt/beta when its delivered there. If it cuts 3s in dev, why do you think its not a real indicator (even if the actual gain on production might be smaller)?

f1sh1918 (approved):

PR #3302, opened 2025-06-03
3246: Replace text buttons with mui button

steffenkleinle:
### Short Description

Replace text buttons with mui buttons.

### Proposed Changes

- Replace all occurrences of `TextButton` with mui `Button`
- Add icons to some buttons (send, copy, download, done)
- Add `Spacing` component
- Fix theme types

### Side Effects

-

### Resolved Issues

Fixes: #

---


web/src/components/CityNotCooperatingFooter.tsx
```diff
@@ -54,8 +45,10 @@ const CityNotCooperatingFooter = ({ languageCode }: CityNotCooperatingFooterProp
     <FooterContainer>
       <StyledIcon src={CityNotCooperatingIcon} />
       <Question>{t('cityNotFound')}</Question>
-      <Button to={pathnameFromRouteInformation({ route: CITY_NOT_COOPERATING_ROUTE, ...{ languageCode } })}>
-        {t('suggestToRegion', { appName: buildConfig().appName })}
+      <Button variant='outlined'>
+        <Link to={pathnameFromRouteInformation({ route: CITY_NOT_COOPERATING_ROUTE, ...{ languageCode } })}>
+          {t('suggestToRegion', { appName: buildConfig().appName })}
+        </Link>
       </Button>
```
steffenkleinle, line 52: 🤔 Not sure if this is the right approach to show a link in button optics. Let me know if you have objections or better ideas.

build-configs/ThemeType.ts
```diff
@@ -9,6 +9,20 @@ export type PaletteColor = SimplePaletteColor & {
   contrastText: string
 }
 
+export type ActionColor = {
```
f1sh1918, line 12: ❓ where do you have also these action states? Is the source from figma?
What is `activatedOpacity`?

f1sh1918 (approved):
Some mostly optional adjustments and improvements but everything works and looks good so far 👍

web/src/components/CityNotCooperatingFooter.tsx
steffenkleinle: Not really sadly, with `component="button"` we need to supply an `onClick` function instead of an `href`.

web/src/components/ContrastThemeToggle.tsx
```diff
@@ -1,9 +1,9 @@
+import { useTheme } from '@emotion/react'
```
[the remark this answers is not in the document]
steffenkleinle: Will do this in a separate PR, same for `styled`.

bahaaTuffaha (changes requested):
Not able to open the security button

steffenkleinle:
@LeandraH @bahaaTuffaha I think I fixed all issues with rtl layouts now. A lot of things are automatically inverted by mui/emotion for rtl layouts, so I had to remove a few occurrences where we do that manually (as that would invert it twice).

This also fixes the position of the toolbar on rtl layouts which seems to be broken for quite a while already also on main.

steffenkleinle:
> Nice work ... great refactoring ✨. Tested on Chrome and some parts on Firefox.✅
> 
> Two points could be done on another PR: 1- The calendar's date selection on light looks too dark (hard to read) and the color looks out of place when the red color is the main theme for ex. at Malte.

Fixed, nice catch!

> 2- If it's in Deutsch all the time we can keep them ltr (checkbox and the tip).

This is just because of missing app translations, should be fixed with #3353.

PR #3303, opened 2025-06-03
Adjust export event button text

steffenkleinle:
### Short Description

This PR changes the export event button text from `Export as iCal` to just `Export`. The former assumes technical knowledge that users know what `iCal` is and might be confusing for those that don't. I checked the naming in calendar apps (google calendar, nextcloud calendar) and this aligns with that naming.

### Proposed Changes

- Change button text from `Export as iCal` to `Export`

### Side Effects

N/A

### Testing

http://localhost:9000/testumgebung/de/events/lauras-test-treffen

### Resolved Issues

Fixes: N/A

---


bahaaTuffaha (approved):
It looks so cleannnn ✨
This is a contrast issue found:
The radio button is white in the two circles
image

2 - Does the button color have enough contrast from the background? @hauf-toni ?

steffenkleinle:
> It looks so cleannnn ✨ This is a contrast issue found: The radio button is white in the two circles image
> 
> 2 - Does the button color have enough contrast from the background? @hauf-toni ?

It definitely does not have enough contrast, good find :D See #3307 though.

hauf-toni:
> It looks so cleannnn ✨ This is a contrast issue found: The radio button is white in the two circles image
> 
> 2 - Does the button color have enough contrast from the background? @hauf-toni ?

The button color doesn’t stand out enough from the background due to the thin black outline. I recommend adjusting it to ensure a minimum contrast ratio of 4.5:1 for accessibility.

Issue #3307, opened 2025-06-04
Improve visibility of radio buttons for darkTheme


f1sh1918:
### Describe the Problem

Currently in the malte help form the active radio box selection is not really good highlighted (white dot) inside white box

Can be tested here: http://localhost:9000/testumgebung/de/gewaltschutz-kontakformular

### Describe Your Preferred Solution

Highlight selection with a different color

### Additional Information

<img width="735" alt="Image" src="https://github.com/user-attachments/assets/cd0b8e08-6ef4-4e16-a4ce-cd166c57a577" />

steffenkleinle:
Should be fixed on main, so no mui components there :) Just change the colors for now.

Issue #3316, opened 2025-06-04
Replace radio group with mui radio group


steffenkleinle:
### Describe the Problem

Our radio group component uses our legacy custom design. This is only for web.

### Describe Your Preferred Solution

- Replace our custom radio group component with mui radio group
- Ensure all states are implemented

### Describe Possible Alternatives

N/A

### Related Issues

Part of: #3231

### Additional Information

https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=512-2144&p=f&t=FuDFR0aDLWOPGiwB-0
https://mui.com/material-ui/react-radio-button/

steffenkleinle:
Done in #3357

PR #3317, opened 2025-06-04
3307: Improve visibility of radio buttons

bahaaTuffaha:
### Short Description

Currently in the malte help form the active radio box selection is not really good highlighted (white dot) inside white box.
Can be tested here: http://localhost:9000/testumgebung/de/gewaltschutz-kontakformular

### Proposed Changes

- Just used `isContrastTheme` to switch between theme color (for contrast mode) and default color.

### Side Effects

None.

### Testing

- Go to any recurring event and click on export as iCal or go to http://localhost:9000/testumgebung/de/gewaltschutz-kontakformular at malte.

### Resolved Issues

Fixes: #3307 

---

### DIFF
--- a/web/src/components/base/RadioGroup.tsx
+++ b/web/src/components/base/RadioGroup.tsx
@@ -20,7 +20,8 @@ const Radio = styled.input`
   width: 1.3em;
   height: 1.3em;
   align-self: center;
-  accent-color: ${props => props.theme.colors.textSecondaryColor};
+  accent-color: ${props =>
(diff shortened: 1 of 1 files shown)


steffenkleinle (approved):
Tested on firefox

steffenkleinle:
> hm will this be part of new design system not to use the theme color. I think red indicator color for selection is a no go tbh because this is an error color. i would use some blue primary color or sth similar

This is only a temporary fix for the old styles. The component will be replaced anyway in #3316 in the new design system. Since we use the theme color everywhere atm, I don't see the point in changing this for a single component that will be removed anyway later on.

PR #3324, opened 2025-06-06
Only create a single tag per delivery

steffenkleinle:
### Short Description

Currently all tags (all, web, native, android, ios) are created for every delivery since we don't pass the platforms parameter.

### Proposed Changes

- Pass the corresponding platform to the git release script

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---


lunars97 (approved):
:+1:

Issue #3328, opened 2025-06-11
Back navigation is not working a second time


steffenkleinle:
### Describe the Bug

The back navigation of the browser is only working the first time. The second time nothing happens.

### Steps to Reproduce

1. Go to https://integreat.app/testumgebung/de/willkommen/bei-uns/frag-integreat
2. Click on `Was ist "Frag Integreat"?`
3. Navigate back using the browser
4. Click on `Was ist "Frag Integreat"?`
5. Navigate back using the browser
6. Nothing happens

### Expected Behavior

The back navigation should always work.

### Related Issues

N/A

### Additional Information

N/A

lunars97:
I think this issue is related to ChatContainer.tsx...

steffenkleinle:
> I think this issue is related to ChatContainer.tsx. query parameter's behaviour...

Bingo! 🕵‍♀ ❤

PR #3330, opened 2025-06-13
3328: Only update query params if chat is open

steffenkleinle:
### Short Description

Only update query params if automatic chat opening is set. This avoids that the web app is performing an unnecessary navigation which breaks the back navigation (due to the route B being updated to route A during back navigation).

### Proposed Changes

- Only update query params if automatic chat opening is set 

### Side Effects

N/A

### Testing

Go to testumgebung and try to navigate back more than once.

### Resolved Issues

Fixes: #3328

---


LeandraH (approved):
Nice! Tested in Firefox

Issue #3208, opened 2025-04-16
Avoid showing complete website url for poi contacts


steffenkleinle:
### Describe the Problem

We currently show the whole url for poi contacts, which can be quite lengthy and reduce readability.

### Describe Your Preferred Solution

- Use only the link text `Website` instead of displaying the whole url
- show the external link icon next to it. (As it i used in the directions section for google maps)
- Open the external link in a new tab

### Describe Possible Alternatives

Only show the url domain and display the external link icon next to it.

### Related Issues

Implemented in #3121.

### Additional Information

See e.g. on
### Local dev
http://localhost:9000/testumgebung/en/locations/bildungsberatung-im-landratsamt-augsburg http://localhost:9000/augsburg/en/locations/dreizehn
### Staging
https://webnext.integreat.app/testumgebung/en/locations/bildungsberatung-im-landratsamt-augsburg https://webnext.integreat.app/augsburg/en/locations/dreizehn
<img src="https://github.com/user-attachments/assets/5c7c5df4-2949-4ed5-96bc-df6a1049f89f" width="150px" />

PR #3332, opened 2025-06-15
3208: Hide URL on POI contacts

LeandraH:
### Short Description

This PR hides the URL in contacts on POIs; the URL is still there as the link destination. I also changed the color of the contact items to be link-colored because I think that is better messaging to the users that something happens when they tap / click on the items

### Proposed Changes

- Instead of a URL, the contact in a POI now shows image
- The items to send an email, call someone, or open the website from the POI contacts are blue.

### Side Effects

- If you just want to copy the URL of a POI contact, it's a little bit more work now.

### Testing

In native and in web, open any POI with a contact with a link in it, e.g. http://localhost:9000/augsburg/en/locations/dreizehn

### Resolved Issues

Fixes: #3208 

---


bahaaTuffaha (approved):
🙃 The icon should be flipped horizontally like the one for googleMaps does.
image

Tested on Firefox, brave and android.

PR #3344, opened 2025-06-23
3246: Fix content direction

LeandraH:
Just a small fix for the content direction in the MUI-text-button branch, now the it's only `rtl` when the language is actually a `rtl` one


bahaaTuffaha (commented):
Nice quick fix that solved most of the text alignment issues.

Now the RTL is having an issue with date filter's calendar buttons:

image

bahaaTuffaha (approved):
Tested on Firefox and Chrome ✅

steffenkleinle:
Thank you @LeandraH! I figured out why this happens, but I wasn't sure what to do about it. This seems like a great solution ❤️

Issue #3345, opened 2025-06-23
Button to open date filters is blue in iOS browser


LeandraH:
### Describe the Bug

When opening the app in my iPhone's browser, the button to open the date filters with is blue instead of black.

### Steps to Reproduce

1. Go to 'https://integreat.app/testumgebung/de/events' from an iPhone browser
2. See that the button to open the date filter view is blue
3. Tap on that button
4. See that the button to close the date filter view is blue

### Expected Behavior

The buttons to open and close the date filter view should be black

### Additional Information

<img src="https://github.com/user-attachments/assets/40fa364a-723d-4f30-a843-66bf5d5602b1" width="300px" />
<img src="https://github.com/user-attachments/assets/0870349c-3209-4097-b5a8-7b2a16cc291f" width="300px" />

bahaaTuffaha:
Couldn't reproduce the issue here... maybe its an iOS or safari problem

LeandraH:
Nice, thank you for your feedback! I think you checked this in the production environment (integreat.app), right? The fix isn't in production yet, but you should be able to see it in the beta environment (webnext.integreat.app).

kalairubachandrakumaran-oss:
@LeandraH 
Yes, that's correct. I tested it on the production environment with location = "Testumgebung" (same link which was mentioned in steps to reproduce)
-> I've now checked on the webnext.integreat.app and it looks fixed (/)

<img width="300" height="650" alt="Image" src="https://github.com/user-attachments/assets/a345bfa2-ade5-4f1a-97b1-35fb9e360b45" />

PR #3359, opened 2025-07-01
3345: Fix color of date filter toggle in browser in iOS

LeandraH:
### Short Description

The button to toggle the date filter in mobile Safari was blue, with this it's black again.

### Proposed Changes

- Using the `theme.colors.textColor` even when we're not in high contrast mode

### Side Effects

- None

### Testing

Use an iPhone and check in its browser that the color is correct

### Resolved Issues

Fixes: #3345 

---


bahaaTuffaha (approved):
Tested on Android through Chrome and Firefox ✅

Issue #3315, opened 2025-06-04
Replace our checkbox component with mui checkbox


steffenkleinle:
### Describe the Problem

Our checkbox component uses our legacy custom design.

### Describe Your Preferred Solution

- Replace our custom checkbox component with mui checkbox
- Ensure all states are implemented

### Describe Possible Alternatives

N/A

### Related Issues

Part of: #3231

### Additional Information

https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=69-132&p=f&t=FuDFR0aDLWOPGiwB-0
https://mui.com/material-ui/react-checkbox/

LeandraH:
Done in #3360

PR #3360, opened 2025-07-01
3315: MUI checkbox

LeandraH:
### Short Description

This PR replaces our own checkbox with the one from MUI

### Proposed Changes

- Uses the MUI checkbox
- Removes the id prop since we don't need that anymore

### Side Effects

- None

### Testing

Places where the checkbox is used:
- Page for consenting to external sources (http://localhost:9000/consent)
- Overlay to filter POIs
- Feedback
- Malte help form (for that, you need to use yarn start:malte, set the api-url in the local storage to https://cms-test.malteapp.de, and go to http://localhost:9000/testumgebung/de/kontaktformular)

### Resolved Issues

Fixes: #3315 

---


web/src/components/base/Checkbox.tsx
```diff
@@ -1,39 +1,19 @@
 import styled from '@emotion/styled'
+import { Checkbox as MUICheckbox, FormControlLabel } from '@mui/material'
```
steffenkleinle, line 2: 🔧 Same as in the other PR.
```suggestion
import { Checkbox as MuiCheckbox, FormControlLabel } from '@mui/material'
```

web/src/components/base/Checkbox.tsx
```diff
@@ -1,39 +1,19 @@
...
-  min-width: 20px;
-  align-self: center;
-`
-
-const StyledLabel = styled.label`
-  color: ${props => props.theme.colors.textColor};
-  padding: 4px;
-  cursor: pointer;
+const StyledLabel = styled(FormControlLabel)`
+  font-weight: 400;
```
steffenkleinle, line 6: 🤔 Isn't a font-weight of 400 the default?

steffenkleinle (approved):
Tested on firefox

bahaaTuffaha:
This is malte:
@hauf-toni is this blue theme will also be applied to malte and if so in this case do I use a blue color for toggle button at #3361?

steffenkleinle:
> This is malte: @hauf-toni is this blue theme will also be applied to malte ? and if so in this case do I use a blue color for toggle button at #3361?

AFAIK we do not really want to use the theme colors anymore, so no more integreat yellow or malte red due to contrast as primary color, only as accent in some cases (e.g. on hover...). So I think that the toggle buttons should also use blue, yes.

bahaaTuffaha (approved):
Tested on Firefox and Chrome ✅

web/src/components/base/Checkbox.tsx
LeandraH: Good point, I'll already use the Typography here :)

Issue #3313, opened 2025-06-04
Implement ToggleButton with mui components


steffenkleinle:
### Describe the Problem

Our toggle button component uses our legacy custom design.

### Describe Your Preferred Solution

- Use mui components to implement the toggle button
- Ensure all states are implemented

### Describe Possible Alternatives

N/A

### Related Issues

Part of: #3231

### Additional Information

https://mui.com/material-ui/react-button/

PR #3361, opened 2025-07-02
3313: Mui toggle button

bahaaTuffaha:
### Short Description

Our toggle button component uses our legacy custom design.

### Proposed Changes

- Changed `ToggleButton` to now use Mui's ToggleButton.
- I adjusted `ToggleButton` in a way that we can use it a standalone with text and onclick or with `ToggleButtonGroup` in this case you have to use value and insted of onClick you can use onChange.
- for `*Mui-selected:hover` this prevents the hover effect specially for mobile screens that will leave the button in white until you click outside of the button.
- I noticed that StyledIcon have a different size at PoiFilters so I created 2 sizes 'small' and 'medium' icon size.

### Side Effects

- None except PoiFilters and FeedbackButtons.

### Testing

- Test PoiFilters
- Test FeedbackButtons
- Test if button behave correctly even on mobile screen.

### Resolved Issues

Fixes: #3313

---


web/src/components/PoiFilters.tsx
```diff
@@ -88,6 +93,15 @@ const PoiFilters = ({
...
   const { t } = useTranslation('pois')
 
+  const handleFilterChange = (event: React.MouseEvent<HTMLElement>, newValue: string | null) => {
+    if (newValue === null) {
+      setSelectedPoiCategory(null)
+    } else {
+      const category = poiCategories.find(cat => cat.id.toString() === newValue)
+      setSelectedPoiCategory(category ?? null)
+    }
+  }
```
steffenkleinle, line 103: 🔧 This can be simplified with only one call to `setSelectedPoiCategory`, could you try to do that please?

web/src/components/PoiFilters.tsx
```diff
@@ -49,11 +50,15 @@ const SortingHint = styled.div`
...
-const TileRow = styled.div<{ itemCount: number }>`
+const TileRow = styled(ToggleButtonGroup)<{ itemCount: number }>`
   display: grid;
   gap: 24px ${tileColumnGap}px;
   justify-content: center;
   grid-template-columns: repeat(${props => props.itemCount}, ${toggleButtonWidth}px);
+
+  & [class*='MuiToggleButton-root'] {
+    border-radius: 18px;
+  }
```
steffenkleinle, line 61: 🤔 I think we should avoid modifying mui styles in single components wherever possible and instead do that globally if necessary.

web/src/components/FeedbackButtons.tsx
```diff
@@ -19,23 +20,32 @@ type FeedbackButtonsProps = {
 
 const FeedbackButtons = ({ isPositive, onRatingChange }: FeedbackButtonsProps): ReactElement => {
   const { t } = useTranslation('feedback')
+
+  const handleChange = (event: React.MouseEvent<HTMLElement>, newValue: string | null) => {
+    if (newValue === t('useful')) {
```
steffenkleinle, line 25: 🔧 I think it would be better to set an explicit value instead of falling back to the text property.

web/src/components/FeedbackButtons.tsx
```diff
@@ -19,23 +20,32 @@ type FeedbackButtonsProps = {
...
+  }
+
+  const currentValue = () => {
+    if (isPositive === null) {
+      return null
+    }
+    if (isPositive) {
+      return t('useful')
+    }
+    return t('notUseful')
```
steffenkleinle, line 39: 🙃 
```suggestion
    return t(isPositive ? 'useful' : 'notUseful')
```

web/src/components/FeedbackButtons.tsx
```diff
@@ -19,23 +20,32 @@ type FeedbackButtonsProps = {
 
 const FeedbackButtons = ({ isPositive, onRatingChange }: FeedbackButtonsProps): ReactElement => {
   const { t } = useTranslation('feedback')
+
+  const handleChange = (event: React.MouseEvent<HTMLElement>, newValue: string | null) => {
```
steffenkleinle, line 24: 💭 Hmm, I think if we'd use a string union instead of a boolean for the rating the logic here would be a lot easier/better understandable.
Since this was not changed in your PR, no need to change this here though.
If you do, I am thinking something of the following:
``` ts
// FeedbackContainer.tsx
type Rating = 'positive' | 'negative'
const [rating, setRating] = useState<Rating | null>(null)

// FeedbackButtons
  const handleChange = (event: React.MouseEvent<HTMLElement>, newValue: Rating | null) =>
    setRating(newValue === rating ? null : newValue)

  return (
    <>
      <div>{t('description')}</div>
      <StyledToggleButtonGroup value={rating} onChange={handleChange} exclusive >
        <ToggleButton icon={SentimentSatisfiedOutlinedIcon} text={t('useful')} value='positive' />
        <ToggleButton icon={SentimentDissatisfiedOutlinedIcon} text={t('notUseful')} value='negative' />
      </StyledToggleButtonGroup>
    </>
  )
```

web/src/components/FeedbackButtons.tsx
bahaaTuffaha: I was trying to stay away from nested ternary.

steffenkleinle: This is not a nested ternary, the null case is already handled before.

web/src/components/PoiFilters.tsx
```diff
@@ -108,15 +122,13 @@ const PoiFilters = ({
...
+            onChange={handleFilterChange}>
             {poiCategories.map(it => (
-              <ToggleButton
-                key={it.id}
-                text={it.name}
-                active={it.id === selectedPoiCategory?.id}
-                onClick={() => setSelectedPoiCategory(it.id === selectedPoiCategory?.id ? null : it)}
-                icon={it.icon}
-              />
+              <ToggleButton iconSize='medium' key={it.id} value={it.id.toString()} text={it.name} icon={it.icon} />
```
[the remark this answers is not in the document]
bahaaTuffaha: id is a number... you can test it without `.toString()` the Poi filter won't get selected.

web/src/components/PoiFilters.tsx
bahaaTuffaha: This is not the default behavior so if I moved it to global it will affect also the feedback buttons.

bahaaTuffaha:
> Nice work, I really like the new rating type over the boolean, its a lot more explicit imo. We should also change this in native, but we can do that in a separate PR if you want.
> 
> One more thing: I think the contrast especially for the poi filters between selected and deselected is really not great, looks very similar imo:
> 
> Image

Thats the default what do you suggest? @hauf-toni

web/src/components/base/ToggleButton.tsx
```diff
@@ -1,31 +1,25 @@
...
 
-const StyledButton = styled(Button)<{ active: boolean | null }>`
-  box-shadow:
-    0 1px 2px rgb(0 0 0 / 25%),
-    0 1px 4px 1px rgb(0 0 0 / 15%);
-  border-radius: 18px;
+const StyledButton = styled(MuiToggleButton)`
+  display: flex;
+  flex-direction: column;
+  border-color: ${props => props.theme.colors.textDisabledColor} !important;
```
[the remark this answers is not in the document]
bahaaTuffaha: Yes I tested it without `!important` it didn't do anything. 🤔 I could try with sx directly.

web/src/components/ThemeContainer.tsx
```diff
@@ -35,6 +35,19 @@ const createTheme = (
...
     },
+    components: {
+      MuiToggleButton: {
+        styleOverrides: {
+          root: {
+            '&:hover': {
+              backgroundColor: 'transparent',
+              color: buildConfig().legacyContrastTheme.colors.linkColor,
+            },
+            borderRadius: '18px !important',
```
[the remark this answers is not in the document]
bahaaTuffaha: It won't affect the border without it ... you can test it out .

web/src/components/__tests__/CityNotCooperatingFooter.spec.tsx
```diff
@@ -11,6 +11,12 @@ jest.mock('../../constants/buildConfig', () =>
       cityNotCooperating: true,
     },
     legacyLightTheme,
+    legacyContrastTheme: {
+      colors: {
+        linkColor: '#0000FF',
+        backgroundAccentColor: '#FFFFFF',
+      },
+    },
```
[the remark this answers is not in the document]
bahaaTuffaha: Oh it needs just `legacyContrastTheme` form 'build-configs/integreat/theme' 😅

web/src/components/ThemeContainer.tsx
```diff
@@ -35,6 +35,19 @@ const createTheme = (
       light: buildConfig().lightTheme,
       dark: buildConfig().darkTheme,
     },
+    components: {
+      MuiToggleButton: {
+        styleOverrides: {
+          root: {
+            '&:hover': {
+              backgroundColor: 'transparent',
+              color: buildConfig().legacyContrastTheme.colors.linkColor,
```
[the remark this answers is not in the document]
bahaaTuffaha: https://github.com/user-attachments/assets/8115257f-4aa1-44b4-bece-518ceac447e3

here using the new palette:

```js
  backgroundColor:
              themeType === 'contrast'
                ? buildConfig().darkTheme.palette.surface.light
                : buildConfig().lightTheme.palette.surface.light,
            borderColor: `${buildConfig().lightTheme.palette.divider} !important`,
```

web/src/components/base/ToggleButton.tsx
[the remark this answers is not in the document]
steffenkleinle: Do we even have to set the border color? I think we should use material ui's defaults instead 🤔

web/src/components/ThemeContainer.tsx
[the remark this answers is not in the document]
steffenkleinle: Oh well :/ But if you set it as a style in ToggleButton.tsx, the `!important` is not necessary? If so, I'd perhaps actually prefer that after all, sorry for the mixed reviews here. This would also make sense in case we use this toggle button in the way intended by mui, e.g. for the events filters.

web/src/components/base/ToggleButton.tsx
```diff
@@ -1,53 +1,46 @@
...
 import Icon from './Icon'
 
 export const toggleButtonWidth = 100
 
-const StyledIcon = styled(Icon)`
-  width: 40px;
-  height: 40px;
-  color: ${props =>
-    props.theme.isContrastTheme ? props.theme.colors.backgroundColor : props.theme.colors.textSecondaryColor};
+const StyledIcon = styled(Icon)<{ iconSize: 'small' | 'medium' }>`
```
steffenkleinle, line 10: ❓ @hauf-toni would it make sense to unify the icon size here? Currently it looks like this:
image
With same-size icons compared to the poi filters, it would like this:
image
For comparison, the poi filters look like this:
image

web/src/components/ThemeContainer.tsx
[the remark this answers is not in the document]
hauf-toni: looks great, thanks for your suggestion.

web/src/components/FeedbackButtons.tsx
```diff
@@ -29,8 +29,18 @@ const FeedbackButtons = ({ rating, setRating }: FeedbackButtonsProps): ReactElem
     <>
       <div>{t('description')}</div>
       <StyledToggleButtonGroup exclusive value={rating} onChange={handleChange}>
-        <ToggleButton value={RATING_POSITIVE} icon={SentimentSatisfiedOutlinedIcon} text={t('useful')} />
-        <ToggleButton value={RATING_NEGATIVE} icon={SentimentDissatisfiedOutlinedIcon} text={t('notUseful')} />
+        <ToggleButton
+          iconSize='medium'
```
steffenkleinle, line 33: 🔧 Since we unified the icon size, there is no more `small` and therefore no more need to have this option/prop. Please remove it.

web/src/components/__tests__/CityNotCooperatingFooter.spec.tsx
```diff
@@ -11,7 +11,6 @@ jest.mock('../../constants/buildConfig', () =>
       cityNotCooperating: true,
     },
     legacyLightTheme,
-    lightTheme,
```
steffenkleinle: 🤔 Why did you remove this/had to remove this?

web/src/components/base/ToggleButton.tsx
```diff
@@ -35,7 +36,7 @@ type ToggleButtonProps = {
 const ToggleButton = ({ text, onClick, className, value, ...props }: ToggleButtonProps): ReactElement => (
   <StyledButton
     color='primary'
-    sx={{ textTransform: 'inherit' }}
+    sx={{ textTransform: 'inherit', borderRadius: '18px !important' }}
```
steffenkleinle, line 39: 🤔 Why is the `textTransform: 'inherit'` necessary? I can't really spot a difference 🙈

web/src/components/base/ToggleButton.tsx
```diff
@@ -20,6 +20,7 @@ const StyledButton = styled(MuiToggleButton)`
   height: 100px;
   padding: 8px;
   text-align: center;
+  border-color: ${props => props.theme.colors.textDisabledColor} !important;
```
steffenkleinle, line 23: 🔧 I found another solution both for the border color as well as for the border radius without both setting a custom color and using `!important`:
Add the following in `FeedbackButtons.tsx` to the styling of `StyledToggleButtonGroup` and in `PoiFilters.tsx` to `TileRow`:
``` tsx
  .${toggleButtonGroupClasses.grouped} {
    border-radius: 18px;
  }
  
  .${toggleButtonGroupClasses.middleButton}, .${toggleButtonGroupClasses.lastButton} {
    border-left: 1px solid ${props => (props.theme.vars || props.theme).palette.divider};
  }
```

We could also create a separate file `SpacedToggleButtonGroup` to avoid duplicating the style.

Inspiration taken from https://mui.com/material-ui/react-toggle-button/#spacing.

steffenkleinle:
One more thing: I think it looks a lot better if we remove (i.e. overwrite the style, not actually remove it) the `margin-top: 8px` from `StyledSmallViewTip` for the toggle button as there is currently a big space between the icon and the text. Perhaps it does actually not make that much sense to share the styling at all and just add a own style for it.

web/src/components/__tests__/CityNotCooperatingFooter.spec.tsx
bahaaTuffaha: because I'm not using `lightTheme` at `ThemeContainer.tsx` anymore.

lunars97 (approved):
Really good work! LGTM :+1:

PR #3367, opened 2025-07-07
Apply font family in Mui

LeandraH:
### Short Description

I realised the font family isn't applied in the theme, so I did that here. This is mostly so that I don't accidentally add unnecessary styling like I did in the Checkbox.

### Proposed Changes

- Added the font family to the theme
- Removed unnecessary styling that I had previously added to get the correct font family

### Side Effects

- Most components should now have Noto as a font

### Testing

Take a look at a few components that have text

### Resolved Issues

Fixes: -

---


lunars97 (approved):
LGTM :green_heart: Tested on Chrome

Issue #3311, opened 2025-06-04
Replace ChipButton with mui chip


steffenkleinle:
### Describe the Problem

Our ChipButton component uses our legacy custom design.

### Describe Your Preferred Solution

- Replace our chip button component with mui chip button
- Ensure all states are implemented

### Describe Possible Alternatives

N/A

### Related Issues

Part of: #3231

### Additional Information

https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-740&p=f&t=FuDFR0aDLWOPGiwB-0
https://mui.com/material-ui/react-chip/

PR #3371, opened 2025-07-14
3311: Mui chip button

LeandraH:
### Short Description

This PR replaces our own chip button with the one from the design system: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-740&p=f&m=dev

### Proposed Changes

- Added the neutral color palette to our styles because it's used for the chip buttons (https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=0-1&p=f&m=dev)
- Replaced the chip button with the one from Mui: https://mui.com/material-ui/react-chip/
- Added some chips to the header for easier testing that I will remove before merging

### Side Effects

- The chip button on the map doesn't look great, but that will probably look different in the future anyway: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3048-15002&m=dev

### Testing

Take a look at the chip buttons in the header, maybe add or remove a bit.

### Resolved Issues

Fixes: #3311 

---

### DIFF
--- a/web/src/components/base/ChipButton.tsx
+++ b/web/src/components/base/ChipButton.tsx
@@ -1,46 +1,126 @@
 import styled from '@emotion/styled'
-import CloseIcon from '@mui/icons-material/Close'
-import { SvgIconProps } from '@mui/material'
-import React, { ElementType, ReactElement } from 'react'
+import { Chip, ChipOwnProps, Theme, useTheme } from '@mui/material'
+import React, { ReactElement } from 'react'
 
-import Button from './Button'
 import Icon from './Icon'
 
-const StyledButton = styled(Button)`
-  display: flex;
-  height: 30px;
-  padding: 4px 8px;
-  align-items: center;
-  margin: 0 2px;
-  border-radius: 20px;
-  gap: 4px;
-  background-color: ${props => props.theme.colors.backgroundColor};
-  color: ${props => props.theme.colors.textSecondaryColor};
-  font-family: ${props => props.theme.fonts.web.contentFont};
-  font-size: 0.875rem;
+const sharedIconMargins = `
+  .MuiChip-icon {
+    margin-inline-start: 12px;
+  }
+
+  .MuiChip-deleteIcon {
+    margin-inline-end: 12px;
+  }
 `
 
-const StyledIcon = styled(Icon)`
-  color: ${props => props.theme.colors.textSecondaryColor};
-  height: 16px;
-  width: 16px;
+const ExtraSmallChip = styled(Chip)`
+  ${sharedIconMargins}
+  height: 20px;
+  font-size: 10px;
+
+  .MuiChip-icon,
+  .MuiChip-deleteIcon {
+    height: 10px;
+    width: 10px;
+  }
 `
 
-type ChipButtonProps = {
-  text: string
-  icon: string | ElementType<SvgIconProps>
-  onClick: () => void
-  label?: string
-  closeButton?: boolean
-  className?: string
+const SmallChip = styled(Chip)`
+  ${sharedIconMargins}
+  height: 24px;
+  font-size: 12px;
+
+  .MuiChip-icon,
+  .MuiChip-deleteIcon {
+    height: 12px;
+    width: 12px;
+  }
+`
+
+const MediumChip = styled(Chip)`
+  ${sharedIconMargins}
+  font-size: 14px;
+
+  .MuiChip-icon,
+  .MuiChip-deleteIcon {
+    height: 14px;
+    width: 14px;
+  }
+`
+
+const getIconElement = (IconProp: undefined | string | ReactElement): ReactElement | undefined => {
+  if (typeof IconProp === 'string') {
+    return <Icon src={IconProp} />
+  }
+  return IconProp
 }
 
-const ChipButton = ({ text, onClick, label, className, ...props }: ChipButtonProps): ReactElement => (
-  <StyledButton label={label ?? text} onClick={onClick} className={className}>
-    <StyledIcon src={props.icon} />
-    <div>{text}</div>
-    {props.closeButton && <StyledIcon src={CloseIcon} />}
-  </StyledButton>
-)
+const mapPropsToMuiChipProps = (props: ChipButtonProps, theme: Theme): ChipOwnProps => {
+  const { variant, icon, size, ...rest } = props
+  const iconElement = getIconElement(icon)
+  if (variant === 'outlined') {
+    return {
+      variant: 'outlined',
+      icon: iconElement,
+      sx: {
+        '.MuiChip-label': {
+          color: theme.palette.neutral[900],
+        },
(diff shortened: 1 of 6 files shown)


LeandraH:
> * Do we need to added the dark palette from MUI or wating to be merged from other PRs?

Good question, @hauf-toni should the chips in the new design system have different colors depending on whether the app is in light or dark mode?

hauf-toni:
> > * Do we need to added the dark palette from MUI or wating to be merged from other PRs?
> 
> Good question, @hauf-toni should the chips in the new design system have different colors depending on whether the app is in light or dark mode?

Yes, definitely. Should we create an overview with the corresponding colors?

LeandraH:
> > > * Do we need to added the dark palette from MUI or wating to be merged from other PRs?
> > 
> > 
> > Good question, @hauf-toni should the chips in the new design system have different colors depending on whether the app is in light or dark mode?
> 
> Yes, definitely. Should we create an overview with the corresponding colors?

That would be great :)

bahaaTuffaha:
> > > * Do we need to added the dark palette from MUI or wating to be merged from other PRs?
> > 
> > 
> > Good question, @hauf-toni should the chips in the new design system have different colors depending on whether the app is in light or dark mode?
> 
> Yes, definitely. Should we create an overview with the corresponding colors?

What about the default dark Mui palette that I used for the other components?

hauf-toni:
> > > > * Do we need to added the dark palette from MUI or wating to be merged from other PRs?
> > > 
> > > 
> > > Good question, @hauf-toni should the chips in the new design system have different colors depending on whether the app is in light or dark mode?
> > 
> > 
> > Yes, definitely. Should we create an overview with the corresponding colors?
> 
> What about the default dark Mui palette that I used for the other components?

That would be even better. Could you share your resource in this thread so I can create an overview based an it?

hauf-toni:
> Thanks for all the effort put into this! ❤️ 💭 I am sadly not sure if this solution is one we should go with. This includes lots of changes and adjustments of the mui default, both for colors as well as sizes and paddings. Most of these changes only lead to very minor differences in look and feel compared to the mui default, see screenshots attached below. Obviously, these changes are configured like this in the design system in figma, but I think we should try to stick to the mui defaults wherever possible, especially if the differences are that small.
> 
> Furthermore, we currently only need one chip size (poi filters), so I think we should not implement multiple different sizes (even if proposed in the design system).
> 
> Comparison: Mui defaults (custom icon sizes/margins): image
> 
> Mui defaults (mui default icon sizes/margins and fonts for default chip): image
> 
> Current implementation: image
> 
> @hauf-toni what is your opinion on this? In the end, @LeandraH implemented it exactly like it was proposed in the design system.

I agree with you @steffenkleinle – we should go with the default MUI implementation. I am a bit confused why the design in figma differed from them in the first place since we only changed color + typeface when adjusting these chips from the original file. sorry that this lead to extra work for you @LeandraH , that was not our intention.

web/src/components/base/ChipButton.tsx
```diff
@@ -1,46 +1,96 @@
-import styled from '@emotion/styled'
-import CloseIcon from '@mui/icons-material/Close'
-import { SvgIconProps } from '@mui/material/SvgIcon'
-import React, { ElementType, ReactElement } from 'react'
+import { Chip, ChipOwnProps, Theme, useTheme } from '@mui/material/SvgIcon'
```
bahaaTuffaha, line 1: SvgIcon is jack of all trades 😂
Image

web/src/components/base/ChipButton.tsx
bahaaTuffaha, line 1: 💭  How about we use styled-components for all these colors and pass `variant` as prop?

web/src/components/base/ChipButton.tsx
```diff
@@ -1,46 +1,96 @@
...
+    return {
+      variant: 'outlined',
+      icon: iconElement,
+      sx: isContrastMode
+        ? {
+            '.MuiChip-label, .MuiChip-icon, .MuiChip-deleteIcon': {
+              color: theme.palette.neutral[50],
+            },
+            borderColor: theme.palette.neutral[50],
+            backgroundColor: '#000',
```
bahaaTuffaha, line 36: Hard coded color detected. 🕵️

bahaaTuffaha (commented):
In contrast mode the first 4 chips they nearly look the same? is that expected?
Image

LeandraH:
> In contrast mode the first 4 chips they nearly look the same? is that expected? Image

Yes, the design for them is here: https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=5199-4016&m=dev

web/src/components/base/ChipButton.tsx
LeandraH: Unfortunately, only two of these exist as variants, whereas our variant `primary` doesn't exist as such in Mui, so it's added in as a color.

web/src/components/ThemeContainer.tsx
```diff
@@ -51,6 +51,8 @@ const createTheme = (
       },
       typography: buildConfig().typography,
       palette: isContrast ? buildConfig().darkTheme.palette : buildConfig().lightTheme.palette,
+      neutral: buildConfig().lightTheme.palette.neutral,
+      quartary: buildConfig().lightTheme.palette.quartary,
```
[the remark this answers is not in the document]
LeandraH: About using the light themes palette: It's the same colors in light mode and dark mode. But I agree that it shouldn't be here in the first place :)

web/src/components/base/ChipButton.tsx
```diff
@@ -1,46 +1,97 @@
...
-)
+const mapPropsToMuiChipProps = (props: ChipButtonProps, theme: Theme): ChipOwnProps => {
+  const { variant, icon, ...rest } = props
+  const iconElement = getIconElement(icon)
+  const isContrastMode = theme.palette.mode === 'dark'
+  if (variant === 'outlined') {
+    return {
+      variant: 'outlined',
+      icon: iconElement,
+      sx: isContrastMode
```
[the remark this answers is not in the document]
steffenkleinle: No, I definitely do not mean that :D I meant that you should just be able to use a color from the palette without having to worry if you are in dark or light mode. If the corresponding color differs between light and dark mode, it should be configured differently in the light and dark theme palettes (as done with `tertiary`). Compare https://github.com/digitalfabrik/integreat-app/blob/1187a639384abdb8b35d5165701d500ffece4229/build-configs/common/theme/colors.ts#L37 for light palette and https://github.com/digitalfabrik/integreat-app/blob/1187a639384abdb8b35d5165701d500ffece4229/build-configs/common/theme/colors.ts#L65 for the dark palette.

PR #3372, opened 2025-07-15
2885: New fields in POIs

### LINKED ISSUE #2885: Extend POI with organization and barrier-free fields
In cms there were new fields added to POIS: organization and barrier-free:
https://github.com/digitalfabrik/integreat-cms/issues/2037

Display these fields in the app.

LeandraH:
### Short Description

This PR adds chips with the organization name and the accessibility to the POIs. Unfortunately, a decent amount of this work will need to be touched again when we merge the design system.

### Proposed Changes

- Added icon for not accessible
- Added chip designs to the POI details (there's some overlap with our existing `ChipButton` but it wasn't made for chips that aren't buttons. The chips in Mui are made for that, so I decided to just vaguely copy the styling into the `PoiDetails` instead of doing a big rewrite of our `ChipButton` that's on its way out anyway)
- Figured out why the resizing of the category icon didn't work, it's because `SvgCssUri` doesn't use the `width` and `height` in the `style` prop, but only those that are passed in separately as their own props, fixed that in `SimpleImage`
- Removed the category from the top of the POIs in native since it's shown in the chips lower down

### Side Effects

- Should be none

### Testing

Check on a couple of POIs, see the new chips, I mostly used LR in Testumgebung DE

### Resolved Issues

Fixes #2885 

---


web/src/components/PoiDetails.tsx
```diff
@@ -6,7 +6,13 @@ import { useTranslation } from 'react-i18next'
 import { getExternalMapsLink } from 'shared'
 import { PoiModel } from 'shared/api'
 
-import { ExternalLinkIcon, LocationIcon, PoiThumbnailPlaceholderLarge } from '../assets'
+import {
+  AccessibleIcon,
+  ExternalLinkIcon,
+  LocationIcon,
+  NotAccessibleIcon,
```
bahaaTuffaha, line 13: You can you the corresponding icon from mui for web.
For example: https://mui.com/material-ui/material-icons/?selected=Accessible

bahaaTuffaha (approved):
Nice work
Tested on web and native ✅
I left some comments.

web/src/components/PoiDetails.tsx
LeandraH: I used Mui for the not accessible one, but since we already have an accessible one, no need to change it, especially since we'll be overwriting it with the Mui ones soon :)

native/src/components/PoiDetails.tsx
```diff
@@ -61,15 +76,38 @@ const PoiDetails = ({ poi, language, distance }: PoiDetailsProps): ReactElement
...
       {distance !== null && (
         <StyledDistance>{t('distanceKilometre', { distance: distance.toFixed(1) })}</StyledDistance>
       )}
-      <StyledCategory>{category.name}</StyledCategory>
       <Thumbnail source={thumbnail} resizeMode='cover' />
       <HorizontalLine />
+      <ChipsContainer>
+        {poi.barrierFree !== null && <Chip>{barrierFreeChip}</Chip>}
+        <Chip>
+          <Text>{poi.organization?.name}</Text>
```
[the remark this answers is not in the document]
LeandraH: Ouch, good catch, thank you!

web/src/components/PoiDetails.tsx
LeandraH: In the PR for the new design system, not in main :)

steffenkleinle (changes requested):
Nice work! Looks great already, but I think the icons are a little too small. Sorry for not catching that in the design review.

web/src/components/__tests__/PoiDetails.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { renderWithRouterAndTheme } from '../../testing/render'
+import PoiDetails from '../PoiDetails'
+
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string, params?: { distance: string }) => (params ? `${key}: ${params.distance}` : key),
+  }),
+}))
+
+describe('PoiDetails (web)', () => {
```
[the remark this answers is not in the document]
LeandraH: Yup, thanks!

native/src/utils/DatabaseConnector.ts
```diff
@@ -410,7 +409,7 @@ class DatabaseConnector {
           ? {
               name: category.organization.name,
               logo: category.organization.logo,
-              url: category.organization.url,
+              website: category.organization.url,
```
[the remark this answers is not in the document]
LeandraH: Boo, you're right, that doesn't work. I changed it because we are getting `website` from the API. I guess that is a good reason to keep the storing and mapping separately 😬

shared/api/types.ts
```diff
@@ -119,6 +121,8 @@ export type JsonPoiType = {
   appointment_url: string | null
   category: JsonPoiCategoryType
   contacts: JsonContactType[]
+  organization: OrganizationType
+  barrier_free: boolean | null
```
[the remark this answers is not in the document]
steffenkleinle: Makes sense, thanks for explaining. @digitalfabrik/cms-team how do you generate your docs? Apparently there is a mistake here.

translations/translations.json
```diff
@@ -1132,7 +1132,9 @@
       "back": "Zurück",
       "ok": "Ok",
       "minimize": "Minimieren",
-      "privacyPolicy": "Ich habe die <1>Datenschutzerklärung</1> gelesen und stimme der Verarbeitung und Speicherung meiner Daten zu."
+      "privacyPolicy": "Ich habe die <1>Datenschutzerklärung</1> gelesen und stimme der Verarbeitung und Speicherung meiner Daten zu.",
+      "accessible": "Barrierefrei",
+      "notAccessible": "Nicht barrierefrei"
```
steffenkleinle, line 1137: I guess we missed it here to add the constants for English as well, we should take care in the future to always also add new translations in English.

LeandraH: Good point, thank you!

Issue #3364, opened 2025-07-02
Improve the bottom sheet at Poi for Android


bahaaTuffaha:
### Describe the Bug

when you pull the bottom sheet handle at Poi while using gesture system navigation (Android) its hard to pull without triggering the system nav.

### Steps to Reproduce

1. Change system nav to gestures.
2. Open the native app.
3. Go to Map/Poi
4. Close the bottom sheet and try to open it again.

### Expected Behavior

- It should not conflict with system nav.

### Related Issues

N/A

### Additional Information

https://github.com/user-attachments/assets/38db0b7c-8eab-4843-953a-25804afb1aaf

PR #3386, opened 2025-07-15
3364: Improve bottom sheet at POI

bahaaTuffaha:
### Short Description
when you pull the bottom sheet handle at Poi while using gesture system navigation (Android) its hard to pull without triggering the system nav.

### Proposed Changes

- Used useSafeAreaInsets to know the bottom safe area at `bottomInset`.
- After moving the bottom sheet above the safe area it will leave an empty space so I added `<BottomSpace/>` to fill the gap.

### Side Effects

None.

### Testing

- You need to switch system navigation to gesture instead of 3 buttons (you can test both).
- Go to POI (Map) and test the bottom sheet.

### Resolved Issues

Fixes: #3364 

---

### DIFF
--- a/native/src/components/PoisBottomSheet.tsx
+++ b/native/src/components/PoisBottomSheet.tsx
@@ -30,7 +30,7 @@ const BottomSheetContent = styled.View`
 
 const BottomSpace = styled.View<{ bottomInset: number }>`
   width: 100%;
-  height: ${props => props.bottomInset + 'px'};
+  height: ${props => `${props.bottomInset}px`};
   background-color: ${props => props.theme.colors.backgroundColor};
 `
 
--- a/native/src/components/PoisBottomSheet.tsx
+++ b/native/src/components/PoisBottomSheet.tsx
@@ -6,6 +6,7 @@ import BottomSheet, {
 import React, { memo, ReactElement, Ref } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Platform } from 'react-native'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import styled, { useTheme } from 'styled-components/native'
 
 import { LocationType } from 'shared'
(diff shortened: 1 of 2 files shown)


LeandraH (commented):
Great idea but I think this fix should only be on Android (is it even a problem on all Android devices?). I just tested it on my iPhone and it adds a needlessly large white bar at the bottom. Or we could only add the bar when the BottomSheet is at the bottom?

lunars97:
I would say that now it definitely works better on Android :+1:

native/src/App.tsx
```diff
@@ -80,11 +79,14 @@ export const NavigationContainerWithTheme = ({ onStateChange }: NavigationContai
   }
 
   return (
-    <NavigationContainer onStateChange={onStateChange} theme={navigationTheme} linking={linking}>
-      <HeaderButtonsProvider stackType='native'>
-        <Navigator />
-      </HeaderButtonsProvider>
-    </NavigationContainer>
+    <SafeAreaView style={{ flex: 1, backgroundColor: theme.colors.backgroundAccentColor }} edges={['bottom']}>
```
steffenkleinle, line 82: 🤔 I am actually not sure if we should not rather use the standard background color such that it blends together. Not really sure though.

steffenkleinle (approved):
Tested on Android emulator, works as expected.

native/src/App.tsx
steffenkleinle: Could you check if you find anything official, e.g. in the material design guidelines?

bahaaTuffaha: I couldn't find anything about the SafeArea and how it looks with the contrast theme.

native/src/App.tsx
```diff
@@ -119,12 +121,7 @@ const App = (): ReactElement => {
               <SafeAreaProvider>
                 <SnackbarContainer>
                   <TtsContainer>
-                    <>
-                      <StatusBar />
-                      <IOSSafeAreaView>
-                        <NavigationContainerWithTheme onStateChange={onStateChange} />
-                      </IOSSafeAreaView>
-                    </>
+                    <NavigationContainerWithTheme onStateChange={onStateChange} />
```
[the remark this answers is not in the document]
steffenkleinle: It just got moved, not removed. See L83.

LeandraH:
Just tested on iOS, it looks great! I'll merge this and then start a native delivery :)

Issue #3284, opened 2025-05-21
Check if we can replace rn-blob-util with rn-fs


LeandraH:
### Describe the Problem

With upgrading `@dr.pogodin/react-native-static-server`, we needed to add the dependency `@dr.pogodin/react-native-fs`. We should check if we can replace `react-native-blob-util` with `@dr.pogodin/react-native-fs` to have one less dependency.

### Describe Your Preferred Solution

Replace `react-native-blob-util` with `@dr.pogodin/react-native-fs` and remove `react-native-blob-util`.

### Describe Possible Alternatives

We can keep it.

### Related Issues

#2932 
~~Blocked by:~~ #2389

This might fix: #3340, #3342

steffenkleinle:
As discussed in #3284: react-native-fs seems to be noticeably slower than react-native-blob-util. Furthermore do we still need react-native-blob-util for react-native-pdf, so there is no benefit in switching atm.

PR #3393, opened 2025-07-17
3284: Replace react-native-blob-util

bahaaTuffaha:
### Short Description

We should check if we can replace react-native-blob-util with @dr.pogodin/react-native-fs to have one less dependency.

### Proposed Changes

- Replaced `BlobUtil.fs` with `RNFS` and what is equivalent at `react-native-fs`.
- Replaced the old mock `react-native-blob-util.ts` with the new one `react-native-fs.ts`.
- writeFile needs to check if there is a directory to write file to it otherwise it wont work.
- Found an issue while fixing the tests that says
```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```
So I found that it's coming from sentry. to solve it I had to remove `...jest.requireActual('@sentry/react-native'),`at `sentry.spec.ts` and added a simple mock at `jest.setup.ts` this will save about 1 or 1.5 second when you do `yarn test` .
- :warning: I can't remove  `react-native-blob-util.ts` from package.json because it's used by `react-native-pdf` check this: https://www.npmjs.com/package/react-native-pdf.

### Side Effects

- Anything could be affected.

### Testing

- Test the app thoroughly.
- Check 3rd leaf content is loading.
- Check if content can be accessed offline.

### Resolved Issues

Fixes: #3284 

---

### DIFF
--- a/native/src/utils/DatabaseConnector.ts
+++ b/native/src/utils/DatabaseConnector.ts
@@ -1,7 +1,7 @@
+import * as RNFS from '@dr.pogodin/react-native-fs'
 import { BBox } from 'geojson'
 import { map, mapValues } from 'lodash'
 import { DateTime } from 'luxon'
-import BlobUtil from 'react-native-blob-util'
 import { rrulestr } from 'rrule'
 
 import {
@@ -36,7 +36,7 @@ export const CONTENT_VERSION = 'v9'
 export const RESOURCE_CACHE_VERSION = 'v1'
 
 // Our pdf view can only load from DocumentDir. Therefore we need to use that
-export const CACHE_DIR_PATH = BlobUtil.fs.dirs.DocumentDir
+export const CACHE_DIR_PATH = RNFS.DocumentDirectoryPath
 export const UNVERSIONED_CONTENT_DIR_PATH = `${CACHE_DIR_PATH}/content`
 // Offline saved content like categories, events and pois
 export const CONTENT_DIR_PATH = `${UNVERSIONED_CONTENT_DIR_PATH}/${CONTENT_VERSION}`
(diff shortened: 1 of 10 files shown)


native/src/utils/DatabaseConnector.ts
```diff
@@ -226,19 +226,27 @@ class DatabaseConnector {
...
+      .catch(() => false)
+    const baseContentDirExists = await RNFS.stat(UNVERSIONED_CONTENT_DIR_PATH)
+      .then(stat => stat.isDirectory())
+      .catch(() => false)
+    const resourceCacheDirExists = await RNFS.stat(RESOURCE_CACHE_DIR_PATH)
+      .then(stat => stat.isDirectory())
+      .catch(() => false)
+    const baseResourceCacheDirExists = await RNFS.stat(UNVERSIONED_RESOURCE_CACHE_DIR_PATH)
+      .then(stat => stat.isDirectory())
+      .catch(() => false)
```
steffenkleinle, line 240: 🤔 Why not just use `exists()`?
```suggestion
    const contentDirExists = await RNFS.exists(CONTENT_DIR_PATH)
    const baseContentDirExists = await RNFS.exists(UNVERSIONED_CONTENT_DIR_PATH)
    const resourceCacheDirExists = await RNFS.exists(RESOURCE_CACHE_DIR_PATH)
    const baseResourceCacheDirExists = await RNFS.exists(UNVERSIONED_RESOURCE_CACHE_DIR_PATH)
```

native/src/utils/DatabaseConnector.ts
```diff
@@ -881,7 +889,13 @@ class DatabaseConnector {
   }
 
   async writeFile(path: string, data: string): Promise<void> {
-    return BlobUtil.fs.writeFile(path, data, 'utf8')
+    // Create directory structure if it doesn't exist
+    const dirname = path.substring(0, path.lastIndexOf('/'))
+    const dirExists = await RNFS.exists(dirname)
+    if (!dirExists) {
+      await RNFS.mkdir(dirname, { NSURLIsExcludedFromBackupKey: true })
+    }
```
steffenkleinle, line 897: 🔧 `mkdir` does not fail if the directory already exists ([docs](https://github.com/birdofpreyru/react-native-fs?tab=readme-ov-file#mkdir)), so you can just do:

```suggestion
    const dirname = path.substring(0, path.lastIndexOf('/'))
    await RNFS.mkdir(dirname, { NSURLIsExcludedFromBackupKey: true })
```

steffenkleinle (commented):
Looks good already :) I think its okay to keep react-native-blob-util also as dependency for now and continue with this PR, react-native-fs seems better maintained/more stable at the moment.

native/src/utils/DatabaseConnector.ts
```diff
@@ -1,7 +1,7 @@
+import * as RNFS from '@dr.pogodin/react-native-fs'
```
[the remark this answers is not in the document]
bahaaTuffaha: I read the first part 😞
```
// Instead, this fork should be imported like this:
import * as RNFS from '@dr.pogodin/react-native-fs';
```

bahaaTuffaha:
@steffenkleinle I noticed a loading screen for each content page I click on even if the device was disconnected from the internet.
Any idea?

bahaaTuffaha:
I compared the profiling of this PR (top window) and main (the window below) by going to categories => "welcome to our website" => "About integreat" then I stopped the profiler.
It look like the main (react-native-blob-util) is a bit faster.
image

Issue #3388, opened 2025-07-16
No languages supported for tts


bahaaTuffaha:
### Describe the Bug

- Tts keeps saying "Sprache wird nicht unterstützt" for every language (was reported from real devices) .
- Update: this could be related to voices list at `TtsContainer.tsx`
```js
  const { data: voices } = useLoadAsync(Tts.voices)
  const isLanguageSupported =
    voices && voices.some(({ language }) => getGenericLanguageCode(language) === getGenericLanguageCode(languageCode))
```
### Recommended approach

To report to sentry when this issue occurs to know how often this is faulty.

### Steps to Reproduce

1. Go to any page (3rd leaf content).
2. Click on 3 dot kebab menu.
3. Select read-aloud.

### Expected Behavior

To work as expected.

### Related Issues

~~It could be related to this : #3287 but not sure because its not published yet to production.~~

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/mqkxo44rx3d35eqm1qqpe97g7r

PR #3400, opened 2025-07-22
3388: Add logging and retry to load voices for tts if no voices available

steffenkleinle:
### Short Description

This PR adds logging to sentry to help debug cases in which no languages were returned by tts voices().
It also implements retrying if no voices are found or there is an error.

### Proposed Changes

- Add logging for errors as well as empty languages array
- Allow to pass extra data to sentry logs
- Retry loading of tts languages if there is an error or empty languages up to two times

### Side Effects

Hopefully none.

### Testing

Check that the tts still works as expected. If you want, you can test by e.g. setting voices to an empty array.

### Resolved Issues

Fixes: #3388 

---


native/src/components/TtsContainer.tsx
```diff
@@ -54,11 +56,22 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
+  const { data: voices, error, refresh, loading } = useLoadAsync(Tts.voices)
   const isLanguageSupported =
     voices && voices.some(({ language }) => getGenericLanguageCode(language) === getGenericLanguageCode(languageCode))
   const enabled = buildConfig().featureFlags.tts
 
+  useReportError(error)
+
+  useEffect(() => {
+    const voicesMissing = voices?.length === 0 || error !== null
+    if (!loading && voicesMissing && voicesRetries < 2) {
```
LeandraH, line 68: 💭 This looks a bit arbitrary and confusing to me, why do we give it exactly two retries? It seems to me like this is hopefully a temporary thing until we figure out what the actual problem is? Or am I misunderstanding that? Maybe we can document that?

LeandraH (approved):
Nice idea, thanks! Tested in emulated iOS, hope this helps 🤞

bahaaTuffaha (approved):
Nice work 
Tested on android ✅

PR #3407, opened 2025-07-23
Update ReleaseNoteTemplate to remove web platform

LeandraH:
### Short Description

This just updates our template for the release notes to not include web as a platform anymore since we don't create release notes for them anymore. I also took the opportunity to remove `#` from the example GitHub issue (`1234` instead of `#1234`) that we don't use because it would turn the issue key into a comment.


LeandraH:
You already did :) https://github.com/digitalfabrik/integreat-app/blob/main/docs/contributing.md#when-to-write-release-notes

lunars97 (approved):
Thank you :green_heart:

PR #3411, opened 2025-07-24
Exclude iOS folder from IntelliJ search

steffenkleinle:
### Short Description

Search results from the folder cluttered the search in IntelliJ, therefore mark it excluded.

### Proposed Changes

- Mark folder `native/ios/DerivedData` as excluded in IntelliJ

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---


steffenkleinle:
> Not tested. Also, still not sure about keeping IntelliJ files in the code base since that is only interesting for the people who use IntelliJ 🙃

Since it is not hurting anyone to have those in the codebase but makes life a lot better for IntelliJ users, I'd very much like to keep them there. 🙈

bahaaTuffaha (approved):
Not tested.

Issue #3405, opened 2025-07-23
Use default mui imports


steffenkleinle:
### Describe the Problem

Mui provides two different options for imports:
Named imports (aka barrel imports) from `@mui/material`/`@mui/material-icons` and default imports (e.g. `import Button from '@mui/material/Button'`.
While for production the bundle size should be the same (if optimized by the bundler with tree shaking), the performance is possibly significantly slower on development, see https://mui.com/material-ui/guides/minimizing-bundle-size. Accordingly, mui recommends to always use default imports.

### Describe Your Preferred Solution

- Replace all existing barrel imports with default imports
- Add an eslint rule to prevent future barrel imports ([docs](https://mui.com/material-ui/guides/minimizing-bundle-size/#enforce-best-practices-with-eslint))
- If possible, configure IntelliJ and VSCode to automatically use default imports

### Describe Possible Alternatives

Just don't care and use whichever.

### Related Issues

N/A

### Additional Information

N/A

PR #3414, opened 2025-07-24
3405: Force default mui imports and restrict shared imports

steffenkleinle:
### Short Description

In order to respect muis best practice imports and keep it consistent, this PR forbids imports directly from `@mui/icons-material` and `@mui/material`.

At the same time, I also forbid imports from `shared` from anything else but `shared`, `shared/api` and `shared/api/testing/endpoints`.

### Proposed Changes

- Add eslint rule to forbid imports as described above
- Fix all wrong imports

### Side Effects

The proposed [eslint rule config](https://mui.com/material-ui/guides/minimizing-bundle-size/#enforce-best-practices-with-eslint) with regex did not work. Probably it was only added in a later version. I managed to work around it though.

### Testing

N/A

### Resolved Issues

Fixes: #3405

---


.vscode/settings.json
steffenkleinle, line 1: @LeandraH @bahaaTuffaha is this file still up-to-date?

LeandraH:
To be honest, I have no idea about the settings files for VS Code, I use my own. I wouldn't necessarily want to use a shared one either since a lot of things in there are personal preferences like automatic stashing or how to display SVG files.

web/.vscode/extensions.json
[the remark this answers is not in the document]
bahaaTuffaha: Yes I think so.

.vscode/settings.json
bahaaTuffaha: why `local-history.path` hard coded like that ?

For these lines the IDE says : Unknown Configuration Setting (they could be deprecated).
```
  "yarn.packageJson": "package.json",
  "prettier.eslintIntegration": true,
  "jest.pathToConfig": "jest.config.js",
  "local-history.path": "C:\\Projects\\Integreat\\intregreat-app backup",
  ```

steffenkleinle:
> To be honest, I have no idea about the settings files for VS Code, I use my own. I wouldn't necessarily want to use a shared one either since a lot of things in there are personal preferences like automatic stashing or how to display SVG files.

I don't mind, I can also just delete these settings if you'd prefer that.

LeandraH:
> > To be honest, I have no idea about the settings files for VS Code, I use my own. I wouldn't necessarily want to use a shared one either since a lot of things in there are personal preferences like automatic stashing or how to display SVG files.
> 
> I don't mind, I can also just delete these settings if you'd prefer that.

I guess that might be easier than maintaining them if nobody uses them. @bahaaTuffaha do you use them?

LeandraH (approved):
Looks great, thank you! Tested in Firefox

bahaaTuffaha:
> > > To be honest, I have no idea about the settings files for VS Code, I use my own. I wouldn't necessarily want to use a shared one either since a lot of things in there are personal preferences like automatic stashing or how to display SVG files.
> > 
> > 
> > I don't mind, I can also just delete these settings if you'd prefer that.
> 
> I guess that might be easier than maintaining them if nobody uses them. @bahaaTuffaha do you use them?

I don't use local-history among them.

bahaaTuffaha (approved):
Thank you!
Tested  ✅
image

PR #3415, opened 2025-07-24
3410: Force sound on iOS in silent mode

### LINKED ISSUE #3410: Force tts sound for iOS silent mode
### Describe the Problem

When using TTS an iPhone in silent mode (using the hardware slider button), no sound is played. This is very confusing for users as its hard to draw the connection/there is no indication why no sound is played.

We should force playing sound as implemented [here](https://github.com/digitalfabrik/lunes-app/pull/1100).

Show a warning to the user instead (see #3391).

### Related Issues

Duplicates: #3391 
Lunes Issue: https://github.com/digitalfabrik/lunes-app/issues/1098

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/rmpupwtkbinwzfomcoksbh4jiy
In Lunes, we rejected the idea of forcing sound in favor of showing a warning to users. However, the situation there is a little different as there sounds is played automatically in some cases, in which the silent mode should not be ignored.
In our case, however, sound is only played by voluntary decisions of the users to play sound, in which case we can simply ignore the silent mode.

LeandraH:
### Short Description

This PR fixes the issue of TTS being overridden by silent mode in iOS

### Proposed Changes

- This sets the category of our audio to playback, which ensures that silent mode is ignored.

### Side Effects

- Should be none

### Testing

This needs to be tested on a real iOS device! Set the phone to silent, using the switch on the side. Then go to a leaf category and use TTS.

### Resolved Issues

Fixes #3410 

---


native/src/components/TtsContainer.tsx
```diff
@@ -60,9 +61,13 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
-    await Tts.getInitStatus().catch(async error =>
-      error.code === 'no_engine' ? Tts.requestInstallEngine() : undefined,
-    )
+    await Tts.getInitStatus()
+      .catch(async error => (error.code === 'no_engine' ? Tts.requestInstallEngine() : undefined))
+      .then(() => {
+        if (Platform.OS === 'ios') {
+          Tts.setIgnoreSilentSwitch('ignore')
+        }
+      })
```
steffenkleinle, line 70: 🔧 Can we switch `catch` and `then` such that `then` comes first?

LeandraH: Hmm, I think the catch should be before the then, since the error catching is pretty specific, and if there is an error in initializing TTS, there is no need to add any settings to it. But it's not something I feel strongly about, if you do, I'll happily change it.

The platform check shouldn't be necessary (I didn't actually test it) but I find it clearer to say that something that only affects iOS is only executed on iOS.

native/ios/Integreat/AppDelegate.swift
LeandraH, line 1: These changes are purely stylistic.

native/src/components/TtsContainer.tsx
steffenkleinle: > Hmm, I think the catch should be before the then, since the error catching is pretty specific, and if there is an error in initializing TTS, there is no need to add any settings to it. But it's not something I feel strongly about, if you do, I'll happily change it.

Then please move it out of the then block and use a separate `await` if its not related. I know that there was a mix with async/await and then/catch before already.

> The platform check shouldn't be necessary (I didn't actually test it) but I find it clearer to say that something that only affects iOS is only executed on iOS.

I am not sure it's a good idea to basically write code/add additional logic for basically documentation purposes. Not really sure myself though, so fine for me.

steffenkleinle (approved):
~~Not tested on iOS as I don't have a real device~~ Tested on real iOS and Android still works as expected.

bahaaTuffaha (approved):
Tested on Android ✅

I noticed that the player will stop when the screen dims. Is it possible to prevent the screen dimming while its reading.🤔

LeandraH:
> I noticed that the player will stop when the screen dims. Is it possible to prevent the screen dimming while its reading.🤔

Good question, maybe open a ticket about it?

Docmale:
Is this PR wrong in a Web-Release?

Docmale:
https://github.com/digitalfabrik/integreat-app/releases/tag/2025.8.1-web

LeandraH:
Hmm, let's keep an eye on whether that happens again for now? If it does, we should probably open a bug ticket for it

Issue #3373, opened 2025-07-15
Use Mui Divider


LeandraH:
### Describe the Problem

We currently have a few places where we make a divider line (e.g. the Toolbar and the ConsentSection). Let's instead use the one that Mui offers out of the box: https://mui.com/material-ui/react-divider/
Design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-726&p=f&m=dev

### Describe Your Preferred Solution

Use the Mui divider everywhere that has a manually written one right now.

### Related Issues

Part of #3231

PR #3417, opened 2025-07-25
3373: Mui divider

bahaaTuffaha:
### Short Description
We currently have a few places where we make a divider line (e.g. the Toolbar and the ConsentSection). Let's instead use the one that Mui offers out of the box

### Proposed Changes

- Replaced the following with Mui divider :
   - hr components.
   - Any `border-bottom` that used for creating a divider except the ones at `RemoteContentSandBox.tsx` and `SharingPopup.tsx`.
   - At Spacer.tsx.
- In some cases I had to wrap the whole components with fragment. (the divider is allergic to flex containers).
- At `CityContentHeader.tsx` I used `KebabItemDivider` it could sound unnecessary instead of remapping it again with divider.

### Side Effects

- This could affect alot of lists and components check test section for what to test 👇️

### Testing
- Test the following components/pages:
   - Poi lists.
   - toolbar.
   - Contact details.
   - Consent section.
   - News
   - City selector page/list.
   - Nearby cities.
   - CityNotCooperatingFooter.
   - CityContentFooter.
   - Lists.
 You can open https://integreat.app to make comparisons and don't forget to check mobile screens.

### Resolved Issues

Fixes: #3373

---

### DIFF
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import { Divider } from '@mui/material'
+import Divider from '@mui/material/Divider'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
 
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -15,7 +15,7 @@ const SidebarFooterContainer = styled.div`
   margin-top: -10px; /* to counteract the padding-top of the normal footer */
   padding: 0 27px;
 
-  > *:not(hr) {
+  > a {
     color: ${props => props.theme.colors.textColor};
     padding: 16px 0;
     display: flex;
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -1,10 +1,12 @@
 import styled from '@emotion/styled'
+import { Divider } from '@mui/material'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { DISCLAIMER_ROUTE, LICENSES_ROUTE, pathnameFromRouteInformation } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
+import useWindowDimensions from '../hooks/useWindowDimensions'
 import Footer from './Footer'
 import Link from './base/Link'
 
@@ -13,17 +15,12 @@ const SidebarFooterContainer = styled.div`
   margin-top: -10px; /* to counteract the padding-top of the normal footer */
   padding: 0 27px;
 
-  > * {
+  > *:not(hr) {
     color: ${props => props.theme.colors.textColor};
     padding: 16px 0;
     display: flex;
     align-items: center;
     justify-content: center;
-    border-bottom: 1px solid ${props => props.theme.colors.footerLineColor};
-
-    &:last-child {
-      border-bottom: none;
-    }
   }
 `
 
@@ -34,6 +31,7 @@ type CityContentFooterProps = {
 }
 
 const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFooterProps): ReactElement => {
+  const { viewportSmall } = useWindowDimensions()
   const { aboutUrls, privacyUrls, accessibilityUrls } = buildConfig()
   const { t } = useTranslation(['layout', 'settings'])
   const aboutUrl = aboutUrls[language] || aboutUrls.default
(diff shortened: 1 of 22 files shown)


web/src/components/Spacer.tsx
```diff
@@ -1,7 +1,8 @@
 import styled from '@emotion/styled'
+import { Divider } from '@mui/material'
 
-const Spacer = styled.hr<{ borderColor: string }>`
+const Spacer = styled(Divider)`
```
steffenkleinle, line 4: 🔧 I think export does not warrant its own file anymore, I think we should just apply the margin separately if needed and delete this file.

web/src/components/ConsentSection.tsx
```diff
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled'
+import { Divider } from '@mui/material'
```
steffenkleinle, line 2: 🔧 You'll have to change all imports with #3414 to
``` suggestion
import Divider from '@mui/material/Divider'
```

web/src/components/CityContentHeader.tsx
```diff
@@ -85,17 +93,20 @@ const CityContentHeader = ({
...
+    </KebabItemDivider>,
+    <KebabItemDivider key='language'>
+      <LanguageSelector
+        languageChangePaths={languageChangePaths}
+        isHeaderActionItem
+        languageCode={languageCode}
+        inKebabMenu
+        closeSidebar={() => setIsSidebarOpen(false)}
+      />
+    </KebabItemDivider>,
```
steffenkleinle, line 109: 🔧 I think this is the wrong place to add dividers for the kebab menu. Instead, this should be handled in `KebabMenu.tsx`. Perhaps it also makes sense to add utility methods to add dividers to a list of elements?

steffenkleinle (commented):
Nice work!
What do you think of a utility method like this?
``` tsx
type WithDividerProps = {
  children: ReactElement
}

const WithDivider = ({ children }: WithDividerProps) => (
  <>
    {children}
    <Divider />
  </>
)
```

web/src/components/Spacer.tsx
bahaaTuffaha: This makes sense ... Do you wanna call it `WithDivider` ?

steffenkleinle: > Do you wanna call it WithDivider ?

Those are unrelated things. Most of the usages of `Spacer` you should be able to just replace with `Divider`. In some cases you might need to apply a margin.

web/src/components/Contact.tsx
```diff
@@ -59,7 +57,7 @@ const Contact = ({
       {!!email && (
         <ContactItem iconSource={MailOutlinedIcon} iconAlt={t('eMail')} link={`mailto:${email}`} content={email} />
       )}
-      {!isLastContact && <Spacer borderColor={theme.colors.borderColor} />}
+      {!isLastContact && <Divider style={{ margin: '12px 0' }} />}
```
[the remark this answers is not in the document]
bahaaTuffaha: I went for style because it's simple non-theme-dependent like margin.

steffenkleinle:
> > Sorry for the late idea but I think the correct way to continue here would be for most cases to use the mui lists: https://mui.com/material-ui/react-list/#introduction
> > There you can just set a list divider, no need for any manual logic when (not) to display a divider 🤔 What do you think?
> 
> Your idea is to replace `.map()` where divider is used with mui list? I will try it 🙂

Exactly. I think in general we should use mui list for all our lists, not just the ones with dividers. However, this might get too big for this PR so we can also just keep this as it is for now.

bahaaTuffaha:
> > > Sorry for the late idea but I think the correct way to continue here would be for most cases to use the mui lists: https://mui.com/material-ui/react-list/#introduction
> > > There you can just set a list divider, no need for any manual logic when (not) to display a divider 🤔 What do you think?
> > 
> > 
> > Your idea is to replace `.map()` where divider is used with mui list? I will try it 🙂
> 
> Exactly. I think in general we should use mui list for all our lists, not just the ones with dividers. However, this might get too big for this PR so we can also just keep this as it is for now.

There is 5 tasks that involves with lists do I leave a note in all of them or that obvious?
image

web/src/components/Contact.tsx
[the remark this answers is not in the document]
LeandraH: Though I'd expect to then be able to use a styled component instead?

web/src/components/KebabMenu.tsx
```diff
@@ -114,7 +115,14 @@ const KebabMenu = ({ items, show, setShow, Footer }: KebabMenuProps): ReactEleme
...
           </Heading>
-          <Content>{items}</Content>
+          <Content>
+            {items.map((item, index) => (
+              <React.Fragment key={`menu-item-${index + 1}`}>
+                {item}
+                {index < items.length - 1 && <Divider />}
+              </React.Fragment>
+            ))}
+          </Content>
```
LeandraH, line 125: 💭 Same here, maybe we can use a Mui list here?

Also, we do previously have a bottom border on the last element as well.

web/src/components/NearbyCities.tsx
```diff
@@ -46,7 +47,12 @@ const NearbyCities = ({ cities, language, filterText, stickyTop }: NearbyCitiesP
     <div>
       <CityListParent stickyTop={stickyTop}>{t('nearbyCities')}</CityListParent>
       {nearbyCities.length > 0 ? (
-        nearbyCities.map(city => <CityEntry key={city.code} city={city} language={language} filterText={filterText} />)
+        nearbyCities.map(city => (
+          <React.Fragment key={city.code}>
+            <Divider />
+            <CityEntry key={city.code} city={city} language={language} filterText={filterText} />
+          </React.Fragment>
+        ))
```
LeandraH, line 55: 💭 Another candidate for a Mui list?

web/src/components/PoisDesktop.tsx
LeandraH, line 1: ❓ I think the bottom border under the ListHeader hasn't been replaced or am I missing something?

LeandraH (commented):
That looks really good already! I just have a few small comments :)

web/src/components/KebabMenu.tsx
bahaaTuffaha: check production... there is no divider on the last element.
image

web/src/components/NearbyCities.tsx
LeandraH: Sure looks like that's okay :)

I'll open more tickets for the .maps

web/src/components/CityContentFooter.tsx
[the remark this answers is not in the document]
LeandraH: Good point, I guess I'll create some more tickets

PR #3420, opened 2025-07-28
Fix delivery by force clearing the yarn cache

LeandraH:
### Short Description

The delivery is failing and my guess is that it's because of some caching issue, especially since this is the first delivery since the RN-upgrade was merged.

### Proposed Changes

- Force clean the yarn cache

### Side Effects

- Should be none

### Testing

Run another delivery after this is merged

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,14 +163,14 @@ commands:
         steps:
             - restore_cache:
                 keys:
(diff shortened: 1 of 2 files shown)


steffenkleinle:
I think the problem was something else (or multiple other things). Anyway, I got it working without resetting the cache, see #3422.

Issue #3423, opened 2025-07-29
Fix safe area for absolute positioned components


steffenkleinle:
### Describe the Bug

For example the snackbar as absolute positioned component partly vanishes behind the safe area view at the bottom.

### Steps to Reproduce

1. Go to a city
2. Click on three dots menu
3. Click on `Read aloud`
4. See snackbar behind bottom navigation

### Expected Behavior

The snackbar is displayed above the bottom navigation.

### Related Issues

Introduced in: #3364 

### Additional Information

This bug was never on production, only on current main.

PR #3425, opened 2025-07-29
3423: Fix safe area for absolute elements

steffenkleinle:
### Short Description

Fix safe area for absolute elements.

### Proposed Changes

- Fix the safe area view for snackbar
- Fix the safe area view for tts player

### Side Effects

N/A

### Testing

Check the tts player and the snackbar on different devices.

### Resolved Issues

Fixes: #3423 

---

### DIFF
--- a/native/src/components/TtsPlayer.tsx
+++ b/native/src/components/TtsPlayer.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import styled, { css } from 'styled-components/native'
 
 import { CloseIcon, PauseIcon, PlaybackIcon, PlayIcon } from '../assets'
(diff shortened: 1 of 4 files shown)


native/src/components/TtsPlayer.tsx
```diff
@@ -26,9 +28,11 @@ const StyledTtsPlayer = styled.View<{ $isPlaying: boolean }>`
   align-items: center;
   align-self: center;
   position: absolute;
-  bottom: 5px;
-  min-height: 93px;
+  bottom: 0;
+  min-height: 112px;
   gap: ${props => (props.$isPlaying ? '0px;' : '20px')};
+  margin-bottom: ${props => props.bottom + extraBottomMargin}px;
+  padding: 8px;
```
bahaaTuffaha, line 35: 🤔 what do you think of this?
```suggestion
  bottom: ${props => props.bottom + extraBottomMargin}px;
  min-height: 112px;
  gap: ${props => (props.$isPlaying ? '0px;' : '20px')};
  padding: 8px;
```

steffenkleinle:
> Tested on android both 3-button nav and gesture nav ✅ For snackbar it took the whole area is that expected?

I would say yes as the alternative of the snackbar moving behind the bottom navigation looks weird imo. Feel free to try it without setting the background color in `SnackbarContainer`.

f1sh1918 (approved):
Tested on android and ios. Good work 👍

Issue #2919, opened 2024-09-11
Add text highlighting for TTS


ztefanie:
**Is your feature request related to a problem? Please describe.**
As a user with low reading skills, I want to see which part the tts feature is currently reading out.

**Describe the solution you'd like**
Implement a text highlight, e.g. a light yellow background, of the text part that is currently read out with the tts feature.
Add this for web, android and ios.

**Describe alternatives you've considered**
Close this issue if it is too much effort.

f1sh1918:
@hauf-toni highlighting sentence would be that difficult which our approach. Highlighting words yes because we read sentence per sentence because that sounds natural with all the stops. If we read word per word we may have to introduce manual "pauses"

Just here some example how we could implement it

https://github.com/user-attachments/assets/0f258d47-75a2-4feb-8bc8-83899e0c7d2e

f1sh1918:
We discussed that a proper implementation is too much effort and the current approach is too error prone.
Even there is no priority from the users to implement this soonish

PR #3429, opened 2025-07-30
2919: Tts text highlighting

f1sh1918:
…o decode html entities in page content

### Short Description

Since its easier for the user to follow the text, we may highlight the current sentence that will be read

### Proposed Changes

- expose the current sentence index from TTSContainer
- replace the currently read sentence and wrap it with `<mark>`

Known issues:
- if html was replaced  within a sentence it will not work, since the sentence cannot be found because its including html tags and the sentences from tts are all html parsed and don't include tags
- if a sentence appears two times it will jump back to the first occurrence  

### Side Effects

-

### Testing

 1. Go to a `Page` http://localhost:9000/augsburg/de/news/tu-news/1925
 2. Open TTS Player and play

### Resolved Issues

Fixes: #2919

---

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -41,12 +41,14 @@
     "dompurify": "^3.2.4",
     "easy-speech": "^2.4.0",
     "focus-trap-react": "^10.2.3",
+    "html-entities": "^2.6.0",
     "i18next": "^25.0.1",
     "luxon": "^3.4.2",
     "mapbox-gl": "npm:empty-npm-package@1.0.0",
     "maplibre-gl": "^4.6.0",
     "normalize.css": "^8.0.1",
     "react": "19.0.0",
+    "react-datepicker": "^7.5.0",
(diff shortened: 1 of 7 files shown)


hauf-toni:
Just a quick note on the fast-forward/rewind function as discussed with @steffenkleinle : In the future once this has been implemented we should **not** automatically play content in the tts function, but rather highlight the section to which the user has jumped and let them press “Play” themselves (in line with best practice). However, since text highlighting has not yet been implemented, I suggest that we let playback start automatically when jumping forward or backward for the time being. this way users will continue to receive clear feedback about where they are. Once highlighting is implemented, we can switch to pure sentence display (without automatic playback) for better control and accessibility.

let me know if you need further clarification or would like to review the feature again once highlighting is ready.

f1sh1918:
will not be implemented

Issue #3427, opened 2025-07-30
Kebab menu not sticking to bottom


steffenkleinle:
### Describe the Bug

If I scroll down before I open the kebab menu, the kebab menu is scrolled up as well.

### Steps to Reproduce

⚠️ This happens only on the design system web branch, its not a bug on main/production.

0. Use a small screen
1. Go to https://future.integreat.app/testumgebung/de
2. Scroll down
3. Scroll a little bit up for the header to appear
4. Click on the kebab menu
5. See the kebab menu not being fully visible

### Expected Behavior

The kebab menu being fully visible all the time.

### Related Issues

N/A

### Additional Information

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/512743e1-12e8-41f2-8088-7f5698eea45a" />

steffenkleinle:
Done in #3427

PR #3430, opened 2025-07-30
3427: Fix kebab menu

bahaaTuffaha:
### Short Description

This is a fix for kebab menu not showing up correctly on the screen when opened on smallViewport (mobile) after scrolling.

### Proposed Changes

- just changed the `position` for `Overlay` and `List` to "fixed" instead of "absolute".

### Side Effects

- Hope none.

### Testing

1. Use a small screen.
2. Go to http://localhost:9000/testumgebung/de
3. Scroll down.
4. Scroll a little bit up for the header to appear.
5. Click on the kebab menu.

### Resolved Issues

Fixes: #3427 

---

### DIFF
--- a/web/src/components/KebabMenu.tsx
+++ b/web/src/components/KebabMenu.tsx
@@ -24,7 +24,7 @@ const ToggleContainer = styled.div`
 
 const List = styled.div`
   font-family: ${props => props.theme.fonts.web.decorativeFont};
-  position: absolute;
+  position: fixed;
   top: 0;
   width: 80vw;
   height: 100vh;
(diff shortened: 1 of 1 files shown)


steffenkleinle (approved):
Tested on firefox, nice fix. Do you have any clue why this works on main?

bahaaTuffaha:
> Tested on firefox, nice fix. Do you have any clue why this works on main?

I have no clue.
I think its possible to remove the scrollY logic? or will it affect something else?

LeandraH (approved):
Nice, thank you! It works because the absolute positioning puts the element in relation to its parent element whereas fixed puts the element relative to the viewport.

steffenkleinle:
> @LeandraH @steffenkleinle Do we need the scrollY logic?

Seems like, see https://github.com/digitalfabrik/integreat-app/pull/3050. Not checked though.

Issue #3446, opened 2025-08-04
Fix native tests about worker process has failed to exit gracefully


bahaaTuffaha:
### Describe the Bug

when you do `yarn test` on native it will say at the end:
```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```
So i checked for open handlers using`yarn jest --detectOpenHandles` and all the fingers are pointing towards sentry.

### Steps to Reproduce

1. run `yarn test` and `yarn jest --detectOpenHandles` .

### Expected Behavior

No warnings.

### Related Issues

None.

### Additional Information

N/A

PR #3447, opened 2025-08-04
3446: Fix sentry tests

bahaaTuffaha:
### Short Description
When you do `yarn test` it will say at the end `A worker process has failed to exit gracefully and has been force exited...` so the issue was sentry's mock.

### Proposed Changes

- Moved sentry mock to jest setup file.
- Removed requireActual() no need for it.

### Side Effects

None.

### Testing

Test `yarn test`

### Resolved Issues

Fixes: #3446

---


steffenkleinle (approved):
Thanks, looks good! No more warnings here :)

Issue #3377, opened 2025-07-15
Replace our DatePicker with the Mui one


LeandraH:
### Describe the Problem

Let's use the Mui Date Picker!
Mui Date Picker: https://mui.com/x/react-date-pickers/
Design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-757&p=f&m=dev

### Describe Your Preferred Solution

Rewrite our DatePicker component to use Mui
We can probably remove the react-datepicker package as well as its CSS code

### Related Issues

Part of #3231

bahaaTuffaha:
Aren't we using the format dd.MM.yyyy, or has it changed?

<img width="590" height="154" alt="Image" src="https://github.com/user-attachments/assets/dc757b38-fa8b-4485-83e5-1895d07a96bb" />


To make it custom check this out https://mui.com/x/react-date-pickers/adapters-locale/#with-luxon
But it require extra package 🫣

steffenkleinle:
> Aren't we using the format dd.MM.yyyy, or has it changed?
[...]

PR #3449, opened 2025-08-05
3377: Mui Datepicker

bahaaTuffaha:
### Short Description
We need to replace `react-datepicker` with Mui date picker.

### Proposed Changes

- I installed `@mui/x-date-pickers` to be able to customize the datepicker.
- Now the datepicker is using `LocalizationProvider` from Mui, which will allow the datepicker to change localization and date format depending on the languageCode.
- Due to this error `Invalid DateTime: unparsable` when clearing one of the date sections like YYYY so I had to create `CustomAdapterLuxon` to override some methods like `parse` and `getInvalidDate` .

:warning:  Issues I faced with this implementation:
1. out of range error I can't set the `datePickerError` when this error is catched inside CustomAdapterLuxon and also not caught at  `onError={reason =>` either.
~~2. fireEvent.change not working so I removed the "handles date change correctly" test from `DatePicker.spec.tsx`.~~

### Side Effects

This will effect the events page.

### Testing

- Test how the filter works with invalid date , empty, and start > end.
- Test the event page on desktop and mobile screens.

### Resolved Issues

Fixes: #3377 

---


web/src/components/DatePicker.tsx
```diff
@@ -75,16 +41,22 @@ const isValidJsDate = (date: Date | null): boolean => {
...
+export class CustomAdapterLuxon extends AdapterLuxon {
+  parse = (value: string, format: string): DateTime | null => {
+    try {
+      return DateTime.fromFormat(value, format)
+    } catch (_) {
+      return null
+    }
   }
+
+  getInvalidDate = (): DateTime => DateTime.now()
```
steffenkleinle, line 58: ❓ Is the implementation of `getInvalidDate` also necessary? It seems to work fine for me without.

web/src/components/DatePicker.tsx
```diff
@@ -132,34 +89,35 @@ const CustomDatePicker = ({
...
-          open={isCalendarOpen}
-          value={value}
-          popperPlacement='bottom'
-          selected={isValidJsDate(date?.toJSDate() ?? null) ? date?.toJSDate() : null}
-          onSelect={e => {
-            setValue(DateTime.fromJSDate(e ?? new Date()).toFormat('dd.MM.yyyy'))
-            setDatePickerError('')
+      <LocalizationProvider dateAdapter={CustomAdapterLuxon} adapterLocale={languageCode}>
+        <DatePicker
+          value={date}
```
steffenkleinle, line 94: ❓ Should we enable `disablePast`? What do you think?

web/src/components/DatePicker.tsx
```diff
@@ -75,16 +41,22 @@ const isValidJsDate = (date: Date | null): boolean => {
...
-    const checkDate = DateTime.fromJSDate(date)
-    const maxValidYear = 2500
-    // Added this operator to prevent crashing rrule when typing random number
-    return checkDate.year <= maxValidYear
-  } catch (_) {
-    return false
+  const checkDate = DateTime.fromJSDate(date)
+  const maxValidYear = 2500
+  return checkDate.year <= maxValidYear
+}
```
steffenkleinle, line 47: ❓ Is the `isValidJsDate` still necessary? I think we already get a luxon `DateTime` so imo we can just set the date directly 🤔

web/src/components/DatePicker.tsx
```diff
@@ -132,34 +89,35 @@ const CustomDatePicker = ({
...
-          name={calendarLabel}
-          isCalendarOpen={isCalendarOpen}
-          onClick={() => setIsCalendarOpen(true)}>
-          <EventIcon />
-        </StyledIconButton>
-      </StyledInputWrapper>
+      </LocalizationProvider>
       <StyledTitle>{title}</StyledTitle>
-      {!!(error || datePickerError) && <StyledError>{error || datePickerError}</StyledError>}
+      {error || datePickerError ? <StyledError>{error || datePickerError}</StyledError> : null}
```
steffenkleinle, line 120: 🔧 Should we use the built-in/suggested way to display an error instead?
https://mui.com/x/react-date-pickers/validation/#show-the-error

web/src/components/DatePicker.tsx
```diff
@@ -132,34 +89,35 @@ const CustomDatePicker = ({
...
+              placeholder: placeholderDate
+                .setLocale(languageCode)
+                .toLocaleString({ month: '2-digit', day: '2-digit', year: 'numeric' }),
+            },
+            openPickerButton: {
+              'aria-label': calendarLabel,
+            },
+          }}
+          onError={reason => {
+            if (reason) {
```
steffenkleinle, line 113: 🔧 Why is this if necessary? I think `reason` is always truthy.

web/src/components/DatePicker.tsx
```diff
@@ -132,34 +89,35 @@ const CustomDatePicker = ({
...
+      <LocalizationProvider dateAdapter={CustomAdapterLuxon} adapterLocale={languageCode}>
+        <DatePicker
+          value={date}
+          onChange={newValue => {
+            if (newValue) {
+              handleDateChange(newValue.toJSDate())
+              setValue(newValue.toFormat('dd.MM.yyyy'))
+              setDatePickerError('')
+            }
+          }}
```
steffenkleinle, line 101: 🔧 Is this custom logic actually necessary? It seems to me like this works just fine as well:
```suggestion
          onChange={setDate}
```

web/src/components/__tests__/DatePicker.spec.tsx
```diff
@@ -41,27 +52,6 @@ describe('DatePicker', () => {
...
-      placeholderDate,
-      calendarLabel: 'calendar',
-    })
-
-    const input = getByPlaceholderText(DateTime.fromJSDate(today).toFormat('dd.MM.yyyy'))
-
-    fireEvent.change(input, { target: { value: newValue } })
-
-    expect(setDate).toHaveBeenCalledWith(newValue)
-  })
```
steffenkleinle, line 63: ❌ Please readd the test. The following worked for me:
``` tsx
  it('handles date change correctly', () => {
    const newValue = DateTime.now().plus({ days: 1 })
    const today = new Date()
    const placeholderDate = DateTime.now()

    const { getByPlaceholderText, getByLabelText, getByText } = renderCustomDatePicker({
      title: 'From Date',
      date: DateTime.local(),
      setDate,
      error: '',
      placeholderDate,
      calendarLabel: 'calendar',
    })
    fireEvent.click(getByLabelText('Choose date, selected date is 6. Aug. 2025'))
    fireEvent.click(getByText(newValue.day))
    const input = getByPlaceholderText(DateTime.fromJSDate(today).toFormat('dd.MM.yyyy'))

    fireEvent.change(input, { target: { value: newValue } })

    expect(setDate).toHaveBeenCalledWith(newValue)
  })
```

steffenkleinle (changes requested):
Nice work on this, thank you! I think we can make a little more use of the utilities mui provides here. Also the test should be readded and fixed :)

web/src/components/__tests__/DatePicker.spec.tsx
bahaaTuffaha: so we are changing the date from UI too?
The 'Choose date, selected date is 6. Aug. 2025' is static and tomorrow it wont work I think.

steffenkleinle: True, guess we need something like `getByLabelText('Choose date, { exact: false })` then.

web/src/components/DatePicker.tsx
steffenkleinle: Can we use `maxDate` instead? https://mui.com/x/react-date-pickers/date-picker/#order-of-years

bahaaTuffaha: Nice find thanks.

web/src/components/DatePicker.tsx
bahaaTuffaha: it only works for the calendar not for the input field.

web/src/components/DatePicker.tsx
```diff
@@ -95,71 +57,49 @@ const CustomDatePicker = ({
...
-      <StyledInputWrapper>
-        <StyledInput
-          open={isCalendarOpen}
-          value={value}
-          popperPlacement='bottom'
-          selected={isValidJsDate(date?.toJSDate() ?? null) ? date?.toJSDate() : null}
-          onSelect={e => {
-            setValue(DateTime.fromJSDate(e ?? new Date()).toFormat('dd.MM.yyyy'))
-            setDatePickerError('')
+      <LocalizationProvider dateAdapter={CustomAdapterLuxon} adapterLocale={languageCode}>
```
steffenkleinle, line 79: ❓ Does it make sense to wrap every date picker separately in this `LocalizationProvider` instead of adding this around the whole app?

web/src/components/__tests__/DatePicker.spec.tsx
```diff
@@ -1,25 +1,37 @@
...
 
+const mockedUseCityContentParams = useCityContentParams as jest.MockedFunction<typeof useCityContentParams>
 describe('DatePicker', () => {
   const setDate = jest.fn()
+  beforeEach(() => {
+    mockedUseCityContentParams.mockReturnValue({
+      cityCode: 'city',
+      languageCode: 'de',
+    })
+  })
```
steffenkleinle, line 21: 🙃 Instead of mocking this, you could also use `renderRoute` instead of `renderWithTheme` here with a city content route (e.g. `/augsburg/de/events`). Compare e.g. `ChatContainer.spec.tsx`.

web/src/components/DatePicker.tsx
steffenkleinle: Oh okay, seems like the typing is wrong. I guess it would still be better practice to just save the error/reason in the state and then decide separately what to do with the specific error:
``` tsx
const [validationError, setValidationError] = React.useState<DateValidationError | null>(null)
const errorMessage = error ?? (validationError ? t('invalidDate') : null)

<DatePicker onError={setError} ... />
```

web/src/components/DatePicker.tsx
```diff
@@ -1,67 +1,30 @@
...
-
-const StyledIconButton = styled(IconButton)<{ isCalendarOpen: boolean }>`
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  position: absolute;
-  right: 16px;
-  align-self: center;
+const StyledError = styled.span`
+  color: ${props => props.theme.colors.invalidInput};
```
steffenkleinle, line 18: 🔧 Please use the new palette:
```suggestion
  color: ${props => props.theme.palette.error.main};
```

web/src/components/DatePicker.tsx
```diff
@@ -95,71 +57,49 @@ const CustomDatePicker = ({
   calendarLabel,
 }: CustomDatePickerProps): ReactElement => {
   const { t } = useTranslation('events')
-  const [isCalendarOpen, setIsCalendarOpen] = useState(false)
   const [datePickerError, setDatePickerError] = useState('')
-  const [value, setValue] = useState('')
-  const maxInputLength = 10
+  const { languageCode } = useCityContentParams()
+  const isError = error || datePickerError
```
steffenkleinle, line 62: ❌ It seems like the error message disappears if I type another character after the error message is first displayed. Is this on purpose?

web/src/components/DatePicker.tsx
bahaaTuffaha: Oh this is a provider already so I think yes. I will move it to `app.tsx`

web/src/components/DatePicker.tsx
bahaaTuffaha: `Property 'palette' does not exist on type 'DefaultTheme'.ts`

web/src/components/DatePicker.tsx
steffenkleinle: Hmm yes, seems like it works now for me as well 🙈

web/src/components/DatePicker.tsx
```diff
@@ -57,49 +42,50 @@ const CustomDatePicker = ({
   calendarLabel,
 }: CustomDatePickerProps): ReactElement => {
   const { t } = useTranslation('events')
-  const [datePickerError, setDatePickerError] = useState('')
+  const [validationError, setValidationError] = useState<DateValidationError | null>(null)
   const { languageCode } = useCityContentParams()
-  const isError = error || datePickerError
+  const defaultMaxDate = '2099-12-31'
+  const defaultMinDate = '1900-01-01'
```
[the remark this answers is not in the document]
steffenkleinle: Not sure how you mean that, for me it works for other dates as well:
image
(This is without custom validation/onChange logic and only uses min/max date).

web/src/components/DatePicker.tsx
```diff
@@ -57,49 +42,50 @@ const CustomDatePicker = ({
...
+        setValidationError(null)
+      }
     }
   }
 
   useEffect(() => {
     if (date == null) {
-      setDatePickerError('')
+      setValidationError(null)
     }
```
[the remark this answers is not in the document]
bahaaTuffaha: I remember it was serving a purpose ... now its useless 🥲

bahaaTuffaha:
Note: English will use American date unless we change adapterLocale for en to `en-gb`.

steffenkleinle (approved):
Great work, tested on Firefox 📆

LeandraH (commented):
Hmm, it looks pretty good already but I don't like that when clicking on the empty field, we get "DD.MM.YYYY" as a placeholder. It looks like we could get it to show the placeholder in the correct language by adding a `localeText` as described [here](https://mui.com/x/react-date-pickers/adapters-locale/#custom-field-placeholder), but that looks like we either need to map the available languages to imports or maintain new translations ourselves. Maybe we can remove the placeholder instead, to be more like it is in the current implementation?

bahaaTuffaha:
> Hmm, it looks pretty good already but I don't like that when clicking on the empty field, we get "DD.MM.YYYY" as a placeholder. It looks like we could get it to show the placeholder in the correct language by adding a `localeText` as described [here](https://mui.com/x/react-date-pickers/adapters-locale/#custom-field-placeholder), but that looks like we either need to map the available languages to imports or maintain new translations ourselves. Maybe we can remove the placeholder instead, to be more like it is in the current implementation?

what? the placeholder gets changed if you changed the language. This is English:
image

bahaaTuffaha:
The only placeholder that looks bad is the Arabic one:
image

Maybe because RTL layout is affecting it?

LeandraH:
image
This is the placeholder on the Geman page. It should be `TT.MM.JJJJ`. When I switch it to e.g. French, the placeholder stays `DD/MM/YYYY`, so the structure is correct but the abbreviations aren't translated. And no idea what's going on with the Arabic one, that's really pretty bad. If we can't get them translated properly, maybe we can hide them?

bahaaTuffaha:
> maybe we can hide them?

I will try...

web/src/utils/muiDatePickerLocales.ts
LeandraH, line 1: Oooooh, that's a cool idea! But it would mean having to maintain a second place for importing each language, which we probably shouldn't do.

How about using `defaultValue={placeholderDate}` in the `StyledDatePicker` without the `placeholder` prop? That removes the entire "DD.MM.YYYY" thing and puts the dates in the correct formatting in.

lunars97: But there is also an issue for people who get confused when they see just numbers without understanding whether they need to start with month or day :thinking:

LeandraH: Hmm, that's true. @hauf-toni Do you think we can show the date picker fields with the current date and one a week in the future in the date filters for events (basically as it is in the current implementation), or do we need to have a placeholder that is something like "TT.MM.JJJJ" in German?

web/src/utils/muiDatePickerLocales.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Get MUI date picker locale text for the given language code
+ * @param languageCode - The language code (e.g., 'de', 'fr', 'es')
+ * @returns MUI locale text object or undefined for unsupported languages
+ */
+export const getDatePickerLocaleText = (languageCode: string): Record<string, unknown> | undefined => {
+  if (languageCode in LOCALE_MAP) {
+    return LOCALE_MAP[languageCode as SupportedLanguageCode].components.MuiLocalizationProvider.defaultProps.localeText
+  }
+  return undefined
+}
```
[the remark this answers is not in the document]
steffenkleinle: Hmm, maybe I misunderstood the other conversation then?
> I think displaying the date picker fields as you suggested (the current implementation) works well, since users will recognize the date format this way. It makes it clear which dates need to be entered.

What is meant by `date picker fields` then? I assumed its e.g. `TT.MM.JJJJ`.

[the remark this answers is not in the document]
LeandraH: No, it's actual dates, as it is in production right now
image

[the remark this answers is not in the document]
LeandraH: Toni quotes my message from above hers, I used the word "current implementation" to mean production:

> Do you think we can show the date picker fields with the current date and one a week in the future in the date filters for events (basically as it is in the current implementation), or do we need to have a placeholder that is something like "TT.MM.JJJJ" in German?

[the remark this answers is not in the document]
steffenkleinle: Works for me, thanks for clarifying. I am not happy with the ambiguity for e.g. 01.01 or 02.02 but also not happy with the necessary workaround, so fine for me.

web/src/components/DatePicker.tsx
```diff
@@ -45,27 +47,33 @@ const CustomDatePicker = ({
   const [validationError, setValidationError] = useState<DateValidationError | null>(null)
   const { languageCode } = useCityContentParams()
   const errorMessage = error ?? (validationError ? t('invalidDate') : null)
+  const muiLocaleText = getDatePickerLocaleText(languageCode)
+  const currentTheme = useTheme()
+  const ltrTheme = { ...currentTheme, direction: 'ltr' }
```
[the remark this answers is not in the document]
bahaaTuffaha: This is a good idea :+1:

web/src/utils/muiDatePickerLocales.ts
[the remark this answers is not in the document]
bahaaTuffaha: Even if we used the placeholder using `defaultValue={placeholderDate}` the YYYY will show up if you backspaced the year field for example.

Issue #3451, opened 2025-08-05
Improve the contrast for PoiChip's border at web


bahaaTuffaha:
### Describe the Problem

The border for PoiChips is in black while the contrast mode is activated.
Thanks to @Docmale who found the issue.
### Describe Your Preferred Solution

- Change the border color to be similar to the label.

### Describe Possible Alternatives

N/A.

### Related Issues

N/A.

### Additional Information

<img width="332" height="743" alt="Image" src="https://github.com/user-attachments/assets/cecfce84-cd1e-49c3-9efb-524048284a12" />

PR #3452, opened 2025-08-05
3451: Change Poi chip border color

bahaaTuffaha:
### Short Description

The border for `PoiChips` is in black while the contrast mode is activated.

### Proposed Changes

- Just change the border color to `textSecondaryColor` like we did at native. 

### Side Effects

None.

### Testing

- Go to map (Poi) and go to any location.
- Check that the border looks good when the contrast mode is enabled.

### Resolved Issues

Fixes: #3451 

---


LeandraH (approved):
Nice catch! Tested on Firefox

Issue #3416, opened 2025-07-25
Prevent Bottom Sheet from getting under the Headroom


bahaaTuffaha:
### Describe the Bug

This problem located at `3231-design-system-web`  about the bottom sheet getting under the headroom on smallViewport because I added `zIndex={10}` to solve mui label text being above the headroom at http://localhost:9000/testumgebung/de/kontaktformular. 

### Steps to Reproduce

1. Checkout to `3231-design-system-web`
2. go to Map.
3. Pull the bottom sheet to the top.

### Expected Behavior

To cover the full screen as it was before.

### Related Issues

The problem came from #3358

### Additional Information

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/73c918bb-3ad3-4401-8fc0-4f8daf0b79e6" />

steffenkleinle:
Done in #3457

PR #3457, opened 2025-08-06
3416: Headroom z-index fix

bahaaTuffaha:
### Short Description

This problem located at 3231-design-system-web about the bottom sheet getting under the headroom on smallViewport.

### Proposed Changes

- Removed the z-index at the headroom.
- Changed the MuiInputLabel zindex to 0.

### Side Effects

Hope none.

### Testing

- Go to http://localhost:9000/testumgebung/de/kontaktformular (Malte).
- scroll down and look if the input labels are above the headroom (header).
- Test the POI bottom sheet.

### Resolved Issues

Fixes: #3416

---

### DIFF
--- a/web/src/components/base/Input.tsx
+++ b/web/src/components/base/Input.tsx
@@ -1,5 +1,12 @@
+import styled from '@emotion/styled'
 import TextField from '@mui/material/TextField'
 import React, { ChangeEvent, ReactElement } from 'react'
+
+const StyledTextField = styled(TextField)`
+  .MuiInputLabel-root {
+    z-index: 0;
+  }
+`
 
 export type InputProps = {
   id: string
@@ -42,7 +49,7 @@ const Input = ({
   }
 
   return (
-    <TextField
+    <StyledTextField
       id={id}
       onClick={onClick}
       onChange={onInputChange}
--- a/web/src/components/base/Input.tsx
+++ b/web/src/components/base/Input.tsx
@@ -1,9 +1,10 @@
 import styled from '@emotion/styled'
+import { inputLabelClasses } from '@mui/material/InputLabel'
 import TextField from '@mui/material/TextField'
 import React, { ChangeEvent, ReactElement } from 'react'
 
 const StyledTextField = styled(TextField)`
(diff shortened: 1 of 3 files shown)


web/src/components/base/Input.tsx
```diff
@@ -1,6 +1,13 @@
+import styled from '@emotion/styled'
 import TextField from '@mui/material/TextField'
 import React, { ChangeEvent, ReactElement } from 'react'
 
+const StyledTextField = styled(TextField)`
+  .MuiInputLabel-root {
+    z-index: 0;
+  }
```
steffenkleinle, line 8: 🙃 Could we change this to
```suggestion
  .${textFieldClasses.root} {
    z-index: 0;
  }
```

steffenkleinle (approved):
Thanks for the fix, lets see if this leads to problems in other places :)

web/src/components/base/Input.tsx
steffenkleinle: My bad, sorry. Copy pasting is not always the best idea 🤦‍♂️

steffenkleinle:
> Found one small issue with this adjustment that the autocomplete suggestion will cover the label partially.

Good catch, that should still be fixed. z-index can be a pain if nested/lots of different things that can be on top/below each other...

web/src/components/base/Input.tsx
```diff
@@ -1,6 +1,17 @@
...
 import TextField from '@mui/material/TextField'
 import React, { ChangeEvent, ReactElement } from 'react'
 
+const StyledTextField = styled(TextField)`
+  .${inputLabelClasses.root} {
+    z-index: 0;
+  }
+
+  input:autofill {
+    transition-delay: calc(infinity * 1s);
```
LeandraH, line 12: ❌ That is a pretty hacky way of not showing a background color. I'm sure you can come up with a more elegant (and understandable) solution :)

LeandraH (commented):
Looks pretty good already, except for that autofill hack :)

LeandraH (approved):
Nice!

Issue #3213, opened 2025-04-18
Show mobile phone number in poi details


steffenkleinle:
### Describe the Problem

The CMS added a new property `mobile_number` to poi contacts. Display the number in the corresponding contact.

### Describe Your Preferred Solution

Add the field to the modal and display the number in the UI.

Example: Two of the contacts of this poi have a mobile phone number set that should be displayed: https://webnext.integreat.app/testumgebung/en/locations/lr 

### Describe Possible Alternatives

N/A

### Related Issues

CMS issue: https://github.com/digitalfabrik/integreat-cms/issues/3497

### Additional Information

API docs: https://digitalfabrik.github.io/integreat-cms/api-docs.html#id12

FreshtaM:
I've added support for displaying the new mobile_number field in POI contact details. The mobile number now shows up in the contact modal if it's set. You can see it in action with the example POI you linked. Let me know if anything needs adjusting.
 Commit url : 
https://github.com/digitalfabrik/integreat-app/commit/42466140e4fc154a497f9e20bf9dbe37bc07d288

PR #3463, opened 2025-08-09
3213: Show mobile phone number in poi details

FreshtaM:
Display the new  mobile_number  field for POI contacts in the Proposed Changes

* Added support for the new `mobile_number` field from the CMS.
* Display the mobile number in the POI contact modal if it Side Effects

* None expected, but this touches the POI contact modal, so worth double-checking other contact fields render as expected.

Testing

* Check that POI contacts with a `mobile_number` have it correctly displayed in the modal.
* Confirm nothing breaks for POIs without a mobile number.
* Test both German and English language views for consistency.

 Resolved Issues: #3213


web/src/assets/index.ts
```diff
@@ -71,6 +72,7 @@ import PoiThumbnailPlaceholderLarge from '../../../assets/poi-thumbnail-placehol
 import PoiThumbnailPlaceholder from '../../../assets/poi-thumbnail-placeholder.jpg'
 
 export {
+  mobileIcon,
```
steffenkleinle, line 75: 🔧 Please stick to the naming conventions here, so UpperCamelCase instead of just camelCase. Also, would be nice to keep this alphabetically sorted.
```suggestion
  MobileIcon,
```

assets/icons/iphone.png
steffenkleinle, line 1: ❌ Could you please use an svg instead of a png here? Otherwise, the icon won't be displayed.
🔧 Please name the icon `mobile.svg` to stay consistent :)

web/src/components/Contact.tsx
```diff
@@ -40,12 +40,12 @@ const Contact = ({
...
-      {!!mobilePhoneNumber && (
+      {!!mobileNumber && (
         <ContactItem
-          iconSource={PhoneIcon}
+          iconSource={mobileIcon}
           iconAlt={t('mobilePhone')}
-          link={`tel:${mobilePhoneNumber}`}
-          content={mobilePhoneNumber}
+          link={`tel:${mobileNumber.replace(/\s/g, '')}`}
+          content={mobileNumber.replace(/\s/g, '')}
```
steffenkleinle, line 48: ❓ Is it actually necessary to replace whitespaces here? I think it should work with whitespaces as well, right?

steffenkleinle (changes requested):
Looks very good already, thank you for your PR! 🎉 
It seems like there are still two problems though:
- The mobile icon is currently not displayed on web, should be fixed by using an svg instead of an png
- This does not work yet for native as we are still using the old `mobilePhoneNumber` there instead of the new `mobileNumber`

Furthermore, there are still a few tests/checks failing :)
Let me know if you need any help with fixing those problems! Thank you 🙏

assets/icons/MobileIcon.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="512" height="512" viewBox="0 0 24 24"><path fill="currentColor" d="M16.73 2.065H7.27a2.386 2.386 0 0 0-2.24 2.5v14.87a2.386 2.386 0 0 0 2.24 2.5h9.46a2.386 2.386 0 0 0 2.24-2.5V4.565a2.386 2.386 0 0 0-2.24-2.5Zm1.24 17.37a1.391 1.391 0 0 1-1.24 1.5H7.27a1.391 1.391 0 0 1-1.24-1.5V4.565a1.391 1.391 0 0 1 1.24-1.5H8.8v.51a1 1 0 0 0 1 1h4.4a1 1 0 0 0 1-1v-.51h1.53a1.391 1.391 0 0 1 1.24 1.5Z"/><path fill="currentColor" d="M10 18.934h4a.5.5 0 0 0 0-1h-4a.5.5 0 0 0 0 1Z"/></svg>
```
steffenkleinle, line 1: 🔧 Please adjust this to a size of 24x24 for consistency :)
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M16.73 2.065H7.27a2.386 2.386 0 0 0-2.24 2.5v14.87a2.386 2.386 0 0 0 2.24 2.5h9.46a2.386 2.386 0 0 0 2.24-2.5V4.565a2.386 2.386 0 0 0-2.24-2.5Zm1.24 17.37a1.391 1.391 0 0 1-1.24 1.5H7.27a1.391 1.391 0 0 1-1.24-1.5V4.565a1.391 1.391 0 0 1 1.24-1.5H8.8v.51a1 1 0 0 0 1 1h4.4a1 1 0 0 0 1-1v-.51h1.53a1.391 1.391 0 0 1 1.24 1.5Z"/><path fill="currentColor" d="M10 18.934h4a.5.5 0 0 0 0-1h-4a.5.5 0 0 0 0 1Z"/></svg>
```

web/src/components/Contact.tsx
steffenkleinle: I just tried this and it works just fine without the whitespace removal. Since we usually try to limit code to the necessary, I'd prefer to remove this whitespace removal if there is no reason for it? Thank you.

FreshtaM: ok that is good but actually I  tried to remove whitespace  from content it also worked

bahaaTuffaha (commented):
Nice work 🎉
Tested on web and Android ✅

You need to fix `Contact.spec.tsx` there is some `mobilePhoneNumber` that needs replacement.

you can always check these commands before doing commits:
```
yarn test

yarn lint

yarn ts:check

yarn prettier:write
```

native/src/components/Contact.tsx
```diff
@@ -45,12 +45,12 @@ const Contact = ({
           Icon={PhoneIcon}
         />
       )}
-      {!!mobilePhoneNumber && (
+      {!!mobileNumber && (
```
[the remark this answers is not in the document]
FreshtaM: thanks for your suggestion ,  I updated the conditional check in Contact.tsx to use mobileNumber != null instead of !!mobileNumber as per the ESLint recommendation. This makes the logic more explicit and avoids potential issues with falsy values like empty strings or zero. Let me know if you spot anything else!

web/src/components/Contact.tsx
```diff
@@ -40,12 +40,12 @@ const Contact = ({
...
       )}
-      {!!mobilePhoneNumber && (
+      {!!mobileNumber && (
         <ContactItem
-          iconSource={PhoneIcon}
+          iconSource={MobileIcon}
           iconAlt={t('mobilePhone')}
-          link={`tel:${mobilePhoneNumber}`}
-          content={mobilePhoneNumber}
+          link={`tel:${mobileNumber.replace(/\s/g, '')}`}
```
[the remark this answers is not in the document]
bahaaTuffaha: Yeah it sounds like a good idea to strip spaces from `mobileNumber`  at both web and native.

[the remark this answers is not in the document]
lunars97: I think it is better to go without space remover as @steffenkleinle suggested above, so i removed it

native/src/components/Contact.tsx
```diff
@@ -45,12 +45,12 @@ const Contact = ({
           Icon={PhoneIcon}
         />
       )}
-      {!!mobilePhoneNumber && (
+      {mobileNumber != null && (
         <PoiDetailRow
-          externalUrl={`tel:${mobilePhoneNumber}`}
+          externalUrl={`tel:${mobileNumber.replace(/\s/g, '')}`}
```
steffenkleinle, line 50: 🔧 Not needed
```suggestion
          externalUrl={`tel:${mobileNumber}`}
```

steffenkleinle (approved):
Thanks :)

lunars97:
Thank you a lot for your contribution @FreshtaM You did a really nice work here :yellow_heart: As we discussed it, I will now merge it into main :)

Issue #3261, opened 2025-05-08
Update Firebase to version 22


LeandraH:
### Describe the Problem

There is a major version update of Firebase out: https://rnfirebase.io/migrating-to-v22
This probably requires a rewrite of our `PushNotificationsManager`, and it might fix some of our ongoing problems with notifications.

### Describe Your Preferred Solution

Update the Firebase packages to the latest versions.

### Related Issues

PN bugs: [#3171](https://github.com/digitalfabrik/integreat-app/issues/3171), [#3094](https://github.com/digitalfabrik/integreat-app/issues/3094)

steffenkleinle:
We also need to have a look at https://notifee.app/react-native/docs/integrations/fcm as the setup differs quite a bit since we are using notifee.

Issue #3171, opened 2025-03-26
Issue with Push Notifications on iPhone


nassabay:
### Describe the Bug

The district of Viersen has informed us as follows:

"Sie haben eine Push-Nachricht erstellt mit Modus: benutze Standard-Sprache, wenn keine Übersetzung verfügbar. Es erscheint bei der Nachricht in den Sprachen aber nichts."

I was able to reproduce the following in the test environment:

**Android:**
- Device: Samsung Galaxy M32
- One UI Version: 5.1
Result: Push notification received. The German message was visible in other languages. Everything worked correctly.

**iPhone:**

- iOS Version: 18.3.2
- Model: iPhone SE
Result: No push notification received. No information, nothing at all.

### Related Issues

Related/duplicates: #3094, #3261

AnnaPerelazova:
we have another issue on iOS 18.3 with push notifications so I think those are related. It is from the following municipality: https://integreat.app/altmarkkreis-salzwedel/de. They tested push notifications on the IPhone. When clicking on a received push notification, they were redirected to the municipality’s first page, not to the push notification.

f1sh1918:
@nassabay can you please update to recent iOS 18.4 and recheck the issue. If we use recent XCode version (for building the app) we can only deploy and test on recent iOS version

nassabay:
@f1sh1918 here is the result:

Device: Samsung Galaxy M32
One UI Version: 5.1
Result: Push notification received. The message in Spanish was visible in that language. Everything worked correctly. 

iPhone: iOS Version: 18.3.2 Work-Phone
Model: iPhone SE
Result: Push notification was received. Everything was as it should be. Which is strange because in the first test I didn't receive any message. 

iPhone: iOS Version: 18.4 Private-Phone
Model: iPhone SE
Result: No push notification received. No information, nothing at all.

PR #3469, opened 2025-08-13
3261, 3094, 3171, 3488: Upgrade and fix push notifications

steffenkleinle:
### Short Description

Upgrade react-native-firebase and correctly implement react-native-firebase and notifee listeners.

### Proposed Changes

- Upgrade react-native-firebase
- Correctly implement all listeners
- Refactor to simplify usage
- Bump app-toolbelt to v0.6.0 (requires https://github.com/digitalfabrik/app-toolbelt/pull/42) to be able to write plists
- Write required GoogleService-Info.plist in pre-actions and add to xcode (https://rnfirebase.io/#generating-ios-credentials)
- Correctly initialize react-native-firebase (https://rnfirebase.io/#configure-firebase-with-ios-credentials-react-native-077)
- Remove unused google service properties from build configs

### Side Effects

- I had to delete and recreate the Podfile.lock. Should be fine though.

### Testing

Test the display and opening of the news on press on both Android and iOS for all three app states each (foreground, background, quit).

### Resolved Issues

Fixes: #3261, #3094, #3171, #3488

---


native/src/utils/PushNotificationsManager.ts
```diff
@@ -60,8 +55,8 @@ export const unsubscribeNews = async (city: string, language: string): Promise<v
   const topic = newsTopic(city, language)
 
   try {
-    const messaging = await importFirebaseMessaging()
-    await messaging().unsubscribeFromTopic(topic)
+    const { getMessaging, unsubscribeFromTopic } = await import('@react-native-firebase/messaging')
+    await unsubscribeFromTopic(getMessaging(), topic)
```
LeandraH, line 59: 💭 Shouldn't the log also be within the try?

steffenkleinle:
> I'm unfortunately not getting any push notifications on emulated or real iOS but I do get the following error when opening the app on either: Image

Should be fixed now, seems like we were missing crucial steps from the react-native-firebase configuration (https://rnfirebase.io/#generating-ios-credentials and https://rnfirebase.io/#configure-firebase-with-ios-credentials-react-native-077) that got lost in the react-native upgrade (#3158).

kalairubachandrakumaran-oss:
Tested on Android with BrowserStack, following results:
- When the app is closed or running in the background, you don’t see the push notification text itself. Instead, you only see the Integreat icon and an indication that something arrived in the phone’s header area
- But when the Integreat app is open, you can see the push notification directly (compare video)
https://github.com/user-attachments/assets/9ee97b32-94bb-40d2-9fb8-c2b47d3a237a

- On iOS, tested with a real device, everything works correctly 

Big thanks to @LeandraH  for taking the time to test this with me and for explaining everything :)

Issue #3384, opened 2025-07-15
Rewrite the city list to use Mui


LeandraH:
### Describe the Problem

Let's use Mui for the city list! And also for the list of nearby cities!
New design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3046-11401&m=dev
Mui list: https://mui.com/material-ui/react-list/

You may also use a list Divider check this discussion: https://github.com/digitalfabrik/integreat-app/pull/3417#pullrequestreview-3070260672

### Describe Your Preferred Solution

Rewrite the city list and the nearby city list to use the Mui list components (in particular, the sticky subheader should work nicely for the letters staying on as subheaders) and the new design, of course!

### Related Issues

Part of #3231

steffenkleinle:
Done in #3384

PR #3475, opened 2025-08-16
3384: Rewrite city list with mui components

steffenkleinle:
### Short Description

Rewrite the city list with mui components.

### Proposed Changes

- Use the different mui components to implement the city list, list items, list texts, list buttons, ...
- Add shared class CityListGroup
- Add helpers for dividers

### Side Effects

N/A

### Testing

Test the city list and try different queries, location settings, screen sizes.

### Resolved Issues

Fixes: #3384

---

### DIFF
--- a/web/src/components/CityListGroup.tsx
+++ b/web/src/components/CityListGroup.tsx
@@ -1,0 +1,48 @@
+import shouldForwardProp from '@emotion/is-prop-valid'
+import ListItem from '@mui/material/ListItem'
+import ListSubheader from '@mui/material/ListSubheader'
+import { styled } from '@mui/material/styles'
+import React, { ReactElement } from 'react'
+
+import { CityModel } from 'shared/api'
+
+import CityEntry from './CityEntry'
+import List from './base/List'
+
+export const StyledListSubheader = styled(ListSubheader, { shouldForwardProp })<{ stickyTop: number }>`
+  top: ${props => props.stickyTop}px;
+  transition: top 0.2s ease-out;
+`
+
+const StyledListItem = styled(ListItem)`
+  flex-direction: column;
+`
+
+type CityListGroupProps = {
+  title: string
+  cities: CityModel[]
+  stickyTop: number
+  languageCode: string
+  filterText: string
+  NoItemsMessage?: string | ReactElement
+}
+
+const CityListGroup = ({
+  title,
+  cities,
+  stickyTop,
+  filterText,
+  languageCode,
+  NoItemsMessage = 'search:nothingFound',
+}: CityListGroupProps): ReactElement => (
+  <StyledListItem key={title} alignItems='flex-start'>
+    <List
+      Header={<StyledListSubheader stickyTop={stickyTop}>{title}</StyledListSubheader>}
+      items={cities}
+      renderItem={city => <CityEntry key={city.code} city={city} language={languageCode} filterText={filterText} />}
+      NoItemsMessage={NoItemsMessage}
+    />
+  </StyledListItem>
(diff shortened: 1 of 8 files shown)


web/src/utils/index.tsx
```diff
@@ -0,0 +1,7 @@
+import Divider from '@mui/material/Divider'
+import React, { ReactElement } from 'react'
+
+export const join = <T, U>(items: T[], separator: (index: number) => U): (T | U)[] =>
+  items.flatMap((item, index) => [item, separator(index)]).slice(0, -1)
+
+export const withDividers = (items: ReactElement[]): ReactElement[] => join(items, index => <Divider key={index} />)
```
lunars97, line 7: there is a suggestion from the documentation, I think it is better to implement as they are suggesting `If you're using the Divider to wrap other elements, such as text or chips, we recommend changing its rendered element to a plain <div> using the component prop, and setting role="presentation". This ensures that it's not announced by screen readers while still preserving the semantics of the elements inside it.`

lunars97 (approved):
Looks really nice! Thank you :fire:

web/src/components/CityListGroup.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const CityListGroup = ({
+  title,
+  cities,
+  stickyTop,
+  filterText,
+  languageCode,
+  NoItemsMessage = 'search:nothingFound',
+}: CityListGroupProps): ReactElement => (
+  <StyledListItem key={title} alignItems='flex-start'>
```
LeandraH, line 52: ❓ I find this naming confusing, I would expect a `StyledListItem` to be an item in the list.

LeandraH (commented):
Nice, this looks pretty good already, but the spacings don't seem to match those in the design, and frankly look a little weird to me. The line before a new letter is longer than the other ones, and the single letters are also not aligned with the city names. The single letters also don't have a line under them. And the space given to each city name seems too high to me, it looks weird. According to the design, they should each be 56px but the ones here are 68px. This is all in Firefox.

Image

web/src/components/base/List.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const List = <T,>({ Header, items, renderItem, NoItemsMessage, className }: ListProps<T>): ReactElement => {
+  if (items.length === 0) {
+    return typeof NoItemsMessage === 'string' ? <Failure errorMessage={NoItemsMessage} /> : NoItemsMessage
+  }
+  return (
+    <StyledMuiList className={className}>
+      {Header}
+      {withDividers(items.map(renderItem))}
```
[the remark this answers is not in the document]
steffenkleinle: What built-in divider do you mean? This helper function uses the mui Divider under the hood 🤔

steffenkleinle:
@LeandraH I did a few adjustments, most importantly I now use one list as it is supposed to be over many sublists for each letter. This cleans up the spacing and sizing quite a bit already.
One main points remain:
Due to the fact that each list item (i.e. city) is a link/pressable, additional padding is necessary, otherwise the on press ripple would look really weird. I also kind of like it like this to create some more visual distinction between the sub headers (A, B, C, ...) and the actual list items. Obviously not implemented like this in the design.

I also removed the dividers between the cities and only kept the ones between the groups, I think the whole list looks a lot cleaner like this 🤔 

Please let me know what you think (also cc @hauf-toni).
It would currently look like this:
image

LeandraH (approved):
Nice, looks good to me, but I would appreciate another look at it from @hauf-toni as well :)

Issue #3376, opened 2025-07-15
Rewrite Breadcrumbs to use the ones from Mui


LeandraH:
### Describe the Problem

Let's replace our Breadcrumbs component with the one from Mui!
Design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-734&p=f&m=dev
Mui Breadcrumbs: https://mui.com/material-ui/react-breadcrumbs/

### Describe Your Preferred Solution

Replace our Breadcrumbs component with the one from Mui
See if we still need the JSON-LD

### Related Issues

Part of #3231

PR #3476, opened 2025-08-18
3376: Mui Breadcrumbs

bahaaTuffaha:
### Short Description

Let's replace our Breadcrumbs component with the one from Mui and showing the current path.

### Proposed Changes

- Replaced the old implementation with Mui Breadcrumbs.
- Now `BreadCrumb` uses span instead of li becouse of this error : `<li> cannot be a descendant of <li>.
This will cause a hydration error.` and just added `display: list-item;` to mimic the list behavior.
- Created the function `getLastBreadcrumbs` that takes the `ancestorBreadcrumbs` and `currentBreadcrumb` to add the Ellipsis and slice the last 2 items (current path included).
- To ensure that the home icon and the separators are not getting hidden by the overflow I added overflow: visible to even items in the breadcrumb and the first one (Home icon).
- I kept the same shrink logic at `BreadCrumb` I also added more shrinking only for current path (for small screens) when its more than 2 items to give more readablity to the previous path.
- Added more small tests.

### Side Effects

- None.

### Testing

- Test the `BreadCrumbs` with 1 , 2 ,3 ... levels.
- Test it on pages with long titles.
- Test it on small screens.

### Resolved Issues

Fixes: #3376 

---


web/src/components/Breadcrumb.tsx
```diff
@@ -14,35 +16,26 @@ const ListItem = styled.li<{ shrink: boolean }>`
...
  * Displays breadcrumbs (Links) for lower category levels
  */
-const Breadcrumb = ({ children, shrink }: BreadcrumbProps): ReactElement => (
-  <ListItem shrink={shrink}>
-    <Separator aria-hidden />
-    {children}
-  </ListItem>
+const Breadcrumb = ({ title, to, shrink, isCurrent }: BreadcrumbProps): ReactElement => (
+  <Link css={helpers.removeLinkHighlighting} color='inherit' href={to}>
+    <StyledTitle shrink={shrink} isCurrent={isCurrent}>
```
lunars97, line 35: I wonder can't we use Typography from MUI here instead of span :thinking:

web/src/components/Breadcrumbs.tsx
```diff
@@ -45,42 +51,92 @@ const StyledIcon = styled(Icon)`
...
+            )
+          }
+
+          if (breadcrumb.title === '...') {
+            return <StyledEllipsis key='ellipsis'>...</StyledEllipsis>
+          }
+
+          return (
+            <Breadcrumb
+              title={breadcrumb.title}
```
lunars97, line 130: what if we use collapsed breadcrumbs to avoid slicing and shrinking them like it is shown [here](https://mui.com/material-ui/react-breadcrumbs/#collapsed-breadcrumbs) :thinking:

bahaaTuffaha: Hmm this is a good idea but still if the user unfold it it will take the same same space...
Unless we want to enable that for just for desktop.

web/src/components/Breadcrumbs.tsx
```diff
@@ -45,42 +51,92 @@ const StyledIcon = styled(Icon)`
...
   return (
-    <Wrapper>
+    <StyledBox>
       <JsonLdBreadcrumbs breadcrumbs={jsonLdBreadcrumbs} />
-      <OrderedList>
-        {ancestorBreadcrumbs.map((breadcrumb, index) =>
-          ancestorBreadcrumbs.length > 1 && index === 0 ? (
-            <li key={breadcrumb.pathname}>
-              <StyledLink to={breadcrumb.pathname}>
+      <StyledMuiBreadcrumbs length={ancestorBreadcrumbs.length} aria-label='breadcrumb' separator={<Separator />}>
```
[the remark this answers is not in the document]
bahaaTuffaha: The icon takes more space in comparison. If you added a border around this icon you can see that space.

web/src/components/Breadcrumbs.tsx
lunars97: you are right here :thinking:

lunars97:
> > Thank you for your PR! Everything looks good so far 👍 except for separator being hidden in smaller viewports (See screenshot below)
> 
> Hmm it should look like this: (I tested it on Chrome and Firefox) image

hmm in my case, it is still showing a little bit sliced view when it is less than 500px

web/src/components/Breadcrumb.tsx
bahaaTuffaha: Hmm I overlooked this comment .. I will try it 👍️

steffenkleinle:
> > > > Thank you for your PR! Everything looks good so far 👍 except for separator being hidden in smaller viewports (See screenshot below)
> > > 
> > > 
> > > Hmm it should look like this: (I tested it on Chrome and Firefox) image
> > 
> > 
> > hmm in my case, it is still showing a little bit sliced view when it is less than 500px
> 
> seems like it does not have any sliced view anymore 👍 , but shouldn't the ellipse be clickable? 🤔 On desktop view it is possible, on mobile view there is a text cursor when I hover over it and no path...

Agreed, I think it should be clickable.

steffenkleinle:
> > > > Agreed, I think it should be clickable.
> > > 
> > > 
> > > For mobile? 👀
> > 
> > 
> > Yes, or is there a reason why it shouldn't be?
> 
> No, just it will get crammed up in one line even with ellipse.

But being clickable/not clickable doesn't change that, does it?

steffenkleinle:
> > But being clickable/not clickable doesn't change that, does it?
> 
> I thought clickable means collapsed-breadcrumbs... https://mui.com/material-ui/react-breadcrumbs/#collapsed-breadcrumbs

No, I think collapsed doesn't really help. I just meant that pressing on a breadcrumb should always also navigate to that parent category, no matter if the full text or only an ellipse is shown.

bahaaTuffaha:
> No, I think collapsed doesn't really help. I just meant that pressing on a breadcrumb should always also navigate to that parent category, no matter if the full text or only an ellipse is shown.

that's what it do ... did you test it?

web/src/components/Breadcrumbs.tsx
```diff
@@ -45,42 +65,109 @@ const StyledIcon = styled(Icon)`
...
+
+  if (allBreadcrumbs.length <= breadCrumbsLimit) {
+    return allBreadcrumbs
+  }
+
+  const home = allBreadcrumbs[0] as BreadcrumbModel
+  const rest = allBreadcrumbs.slice(1)
+  const ellipsis = new BreadcrumbModel({
+    title: '...',
+    pathname: '',
```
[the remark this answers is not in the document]
bahaaTuffaha: Because ellipsis just a span but now I will change it to use the rest...

web/src/utils/getBreadcrumbs.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const home = allBreadcrumbs[0] as BreadcrumbModel
+  const rest = allBreadcrumbs.slice(1)
+  const ellipsisPathname = returnAllBreadcrumbs ? '' : (rest[0]?.pathname ?? '/')
+
+  const ellipsis = new BreadcrumbModel({
+    title: '...',
+    pathname: ellipsisPathname,
+    node: React.createElement(StyledEllipsis, null, '...'),
```
bahaaTuffaha, line 30: I think @steffenkleinle won't like this but this is the way I found to set the node without turning the file into tsx . 😅

web/src/components/Breadcrumbs.tsx
```diff
@@ -1,40 +1,60 @@
...
-  padding: 0;
+const StyledMuiBreadcrumbs = styled(MuiBreadcrumbs)<{ length: number }>`
+  & ol {
+    flex-wrap: nowrap;
+    overflow: hidden;
+    align-items: flex-start;
+  }
+
+  & li {
+    overflow: hidden;
```
lunars97, line 30:
```suggestion
& a {
          text-decoration: none;
      }
```

lunars97 (changes requested):
Looks really good :+1: except for underlined link style. See my suggestion :hammer:

web/src/components/Breadcrumbs.tsx
steffenkleinle: Please rather use our own custom `Link` component :) It takes care of all the internal/external link stuff and removes link highlighting.

steffenkleinle (commented):
Why did you decide to include the current page in the breadcrumbs? To me, that is
both a little confusing (twice the same text, link/button highlighting but nothing happens on click) and, more importantly, reduces the available space for actually necessary/helpful breadcrumbs/parent categories a lot, especially on small devices.

hauf-toni:
Based on todays research, here is what I found out:
ℹ️ _read if interested or skip below to the really important content if not_:

- breadcrumbs should be used for applications with deep, nested navigation (like Integreat)
- they should be complementary to the main navigation, not a replacement
- labels should mirror actual page titles, without jargon or abbreviations
- the last breadcrumb should represent the current page without repeating the H1 and should not be a link
- all breadcrumbs except the last should be clickable
- for long trails, collapse to show only 1–2 previous levels plus the current page (eg, home > … > sneakers)
- on mobile a back button takes priority, so breadcrumbs should be minimal (e.g., Home > .... > current page)
- for narrow screens, allow horizontal scrolling of breadcrumbs instead of wrapping them into multiple lines.

**🔧  take-aways for the current implementation as seen above:**

1. 
- too many separators (> > >),  indicates collapsed levels, but they are visually noisy + unclear
– best practice would be to collapse to a single ellipsis (Home > … > Current Page) (no repetition of the page name itself!) instead of repeating separators
- truncation without clarity (both “Leben in Deu...” and “Gleichstel...” are truncated which makes orientation hard) should be avoided.

2. 
– the screenshot shows the current page (Gleichstellung...) as a link
– best practice is to make the last breadcrumb static instead of a link

Let me know if you have any further questions.

steffenkleinle (changes requested):
Nice work already, but I think this still needs some more work and polishing:
- There are still tons of custom styling here, please use the corresponding mui components (e.g. `Button` with `component={Link}` instead of custom styling the links
- I also think that the ts code could still be cleaned up/simplified a bit, could you give it a try please? Thanks :)
- The last breadcrumb should
  a) not be highlighted like a link/something clickable in blue
  b) ~~not have the actual title of the page but instead something like `Current page`~~
- What do you think of using a mui menu for the collapsed item(s) like nextcloud does?

image

image

lunars97:
> Nice work already, but I think this still needs some more work and polishing:
> 
> * There are still tons of custom styling here, please use the corresponding mui components (e.g. `Button` with `component={Link}` instead of custom styling the links
> * I also think that the ts code could still be cleaned up/simplified a bit, could you give it a try please? Thanks :)
> * The last breadcrumb should
>   a) not be highlighted like a link/something clickable in blue
>   b) not have the actual title of the page but instead something like `Current page`
> * What do you think of using a mui menu for the collapsed item(s) like nextcloud does?
> 
> image image



> Nice work already, but I think this still needs some more work and polishing:
> 
> * There are still tons of custom styling here, please use the corresponding mui components (e.g. `Button` with `component={Link}` instead of custom styling the links
> * I also think that the ts code could still be cleaned up/simplified a bit, could you give it a try please? Thanks :)
> * The last breadcrumb should
>   a) not be highlighted like a link/something clickable in blue
>   b) not have the actual title of the page but instead something like `Current page`
> * What do you think of using a mui menu for the collapsed item(s) like nextcloud does?
> 
> image image

I do agree with point **a**, but not really agree with point **b** :) Have a look at this example from WAI https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/

bahaaTuffaha:
> Nice work already, but I think this still needs some more work and polishing:
> 
> * There are still tons of custom styling here, please use the corresponding mui components (e.g. `Button` with `component={Link}` instead of custom styling the links
> * I also think that the ts code could still be cleaned up/simplified a bit, could you give it a try please? Thanks :)
> * The last breadcrumb should
>   a) not be highlighted like a link/something clickable in blue
>   b) not have the actual title of the page but instead something like `Current page`
> * What do you think of using a mui menu for the collapsed item(s) like nextcloud does?

I agree with @lunars97 plus point a is against what we have at the design:
https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-734&p=f&t=407apDh8Gf1QuVf1-0

hauf-toni:
> New changes: now mobile and desktop will behave the same for the collapsed breadcrumbs: @hauf-toni
> 
>  2025-10-01.16-56-40.mp4

That looks great! Thank you for your effort

steffenkleinle (approved):

lunars97:
Looks really good :+1: Small remark: the last element should be also a link and should have an attribute `aria-current="page"`

web/src/components/Breadcrumb.tsx
```diff
@@ -1,48 +1,40 @@
...
-  <ListItem shrink={shrink}>
-    <Separator aria-hidden />
-    {children}
-  </ListItem>
-)
+const Breadcrumb = ({ title, to }: BreadcrumbProps): ReactElement => {
+  const current = to === normalizePath(useLocation().pathname)
+
+  if (current) {
+    return <StyledTypography aria-current='page'>{title} </StyledTypography>
```
lunars97, line 30:
```suggestion
    return <Link to={to} aria-current='page'>{title}</Link>
```

bahaaTuffaha: 1- Even if we show just 1 breadcrumb for small screens the title could be too long:
http://localhost:9000/muenchen/de/beratung-und-hilfe-4/beratung-und-hilfe-allgemein/formularservice-hilfe-beim-ausf%C3%BCllen-von-formularen-und-antr%C3%A4gen

image


2- Did we agreed to keep the last breadcrumb non-Clickable ?

Issue #3439, opened 2025-08-01
Add translations for Indonesian


steffenkleinle:
### Describe the Problem

As a user I want to see the app translated in my native language.

### Describe Your Preferred Solution

Add translations for Indonesian ~~and Hausa~~.

### Describe Possible Alternatives

N/A

### Related Issues

Support added in: #3438

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/axf4wf1cdjbiug5w484a34z3iy

LeandraH:
We're not implementing Hausa for now: https://chat.tuerantuer.org/digitalfabrik/pl/u1bchwerwpr5uqz9h7rfp545ra

PR #3481, opened 2025-08-20
3439: Indonesian translations

LeandraH:
### Short Description

Adding the Indonesian translations and removing Hausa since the municipality decided against it (https://chat.tuerantuer.org/digitalfabrik/pl/u1bchwerwpr5uqz9h7rfp545ra)

### Resolved Issues

Fixes: #3439 

---

### DIFF
--- a/translations/src/config.ts
+++ b/translations/src/config.ts
@@ -156,9 +156,6 @@ class Config {
     id: {
       rtl: false,
     },
-    ha: {
-      rtl: false,
-    },
   }
 
   // Fallbacks for unnormalized language codes from our backend
--- a/translations/translations.json
+++ b/translations/translations.json
@@ -105,6 +105,13 @@
       "organizationMoreInformation": "További <1>{{organization}}</1>-információkat <3>{{domain}}</3> oldalon találhat.",
       "disabledPdf": "PDF-export ezen a nyelven jelenleg nem lehetséges."
     },
+    "id": {
+      "createPdf": "Buat fail PDF",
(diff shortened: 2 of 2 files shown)


lunars97 (approved):
Thank you for your PR :yellow_heart: Not tested.

bahaaTuffaha (approved):
I couldn't find the language at Testumgebung to test it.
I think it needs a release note.

LeandraH:
> I couldn't find the language at Testumgebung to test it. 
The languages that are available in each municipality are set in the CMS. So to test the language in Testumgebung, you have to first enable it in the CMS.

> I think it needs a release note.
We already had a release note for the first part of the ticket, I don't think we need to say a second time that Indonesian is really available now (if a municipality enables it 😅 )

steffenkleinle:
> > I think it needs a release note.
> 
> We already had a release note for the first part of the ticket, I don't think we need to say a second time that Indonesian is really available now (if a municipality enables it 😅 )

I think its fine without a release note, but in the past I always added two release notes because they are actually different things:
- PR 1: Add support for Indonesian for municipalities (and therefore for content in Indonesian)
- PR 2: Add app translations in Indonesian (make the app actually available in Indonesian)

steffenkleinle:
> The language can be selected, but there is no content, and regarding the CMS

This issue is only about app translations, so the all the UI texts (e.g. menu items, messages, ...). Therefore it is not necessary that there is content in order to test this issue. Wouldn't hurt though obviously.

>I think the translation is missing. I need someone who can show me how to do this?

I think you should talk to @digitalfabrik/service-team in this regard :)

Issue #3310, opened 2025-06-04
Use imports from mui for styled and useTheme


steffenkleinle:
### Describe the Problem

mui provides its own utilities for `styled` and `useTheme` which should be used over those from emotion.

### Describe Your Preferred Solution

- Change all imports from `import styled from '@emotion/styled'` to `import { styled } from '@mui/material/styles'`
- Change all imports from `import { useTheme } from '@emotion/react'` to `import { useTheme } from '@mui/material/styles'`

### Describe Possible Alternatives

N/A

### Related Issues

Discussed in #3302.

### Additional Information

N/A

PR #3482, opened 2025-08-20
3310: Use MUI imports for styled and useTheme

lunars97:
### Short Description

This PR replaces all the occurences of styled and useTheme from '@emotion/styled' with '@mui/material/styles' as well as adjusts `styled`

### Proposed Changes

- Replace '@emotion/styled' with '@mui/material/styles' 

### Side Effects

- none

### Testing

Test the web app and see that it does not breaks anything :muscle: 

### Resolved Issues

Fixes: #3310 

---

### DIFF
--- a/web/src/@types/@emotion/react.d.ts
+++ b/web/src/@types/@emotion/react.d.ts
@@ -1,6 +1,0 @@
-import '@mui/material/styles'
-
-declare module '@mui/material/styles' {
-  // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
-  export interface Theme {}
-}
--- a/web/src/@types/@emotion/react.d.ts
+++ b/web/src/@types/@emotion/react.d.ts
@@ -1,7 +1,6 @@
-import '@emotion/react'
-import { Theme as MuiTheme } from '@mui/material/styles'
+import '@mui/material/styles'
 
-declare module '@emotion/react' {
+declare module '@mui/material/styles' {
   // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
-  export interface Theme extends MuiTheme {}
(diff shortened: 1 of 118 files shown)


LeandraH (approved):
Wooops

bahaaTuffaha (changes requested):
1) 🔧 Something wrong happened to the breadcrumbs:
Image

2) ❓ Do we need the package `@emotion/styled` ?
3) 🔧  I think we can remove `web/src/@types/@emotion/react.d.ts ` because `declare module '@mui/material/styles' {` is already declared at `web/src/@types/mui-theme.d.ts`.

lunars97:
> 1. 🔧 Something wrong happened to the breadcrumbs:
> 
> Image
> 2. ❓ Do we need the package `@emotion/styled` ?
> 3. 🔧  I think we can remove `web/src/@types/@emotion/react.d.ts ` because `declare module '@mui/material/styles' {` is already declared at `web/src/@types/mui-theme.d.ts`.

oops :see_no_evil: I will have a look :+1: Thank you for your review!

lunars97:
> 1. 🔧 Something wrong happened to the breadcrumbs:
> 
> Image
> 2. ❓ Do we need the package `@emotion/styled` ?
> 3. 🔧  I think we can remove `web/src/@types/@emotion/react.d.ts ` because `declare module '@mui/material/styles' {` is already declared at `web/src/@types/mui-theme.d.ts`.

yes, about the package `@emotion/styled`, it is currently used by `@integreat-app/react-sticky-headroom`, if I remove it from web package then there is an error saying that there is no `@emotion/styled`, can we also replace it in `@integreat-app/react-sticky-headroom`? :thought_balloon:

Issue #3399, opened 2025-07-22
Consistent wording for imprint


AnnaPerelazova:
### Describe the Problem

In the web app, the element is labeled "Impressum und Kontakt’" (see web screenshot), while in the app and CMS, it’s labeled "Impressum" only (see app and CMS screenshots)

### Describe Your Preferred Solution

~~Ensure consistent wording in the web, app and cms **"Impressum und Kontakt"** (not just "Impressum")~~
As discussed in https://github.com/digitalfabrik/integreat-cms/issues/3801 we want to only use `Impressum` everywhere. Correspondingly, change the text for web from `Impressum und Kontakt` to just `Impressum`.
This ensures consistent and concise wording.

Make sure that:
- You use our internalization (i18n) library [react-i18next](https://react.i18next.com) for the updated text `Impressum`
- Use an existing translation if possible
- If applicable: Check if the translations for `Impressum and Kontakt` is now unused and remove it in that case

### Describe Possible Alternatives

N/A

### Related Issues

CMS issue: https://github.com/digitalfabrik/integreat-cms/issues/3801

### Additional Information
**Web:**
<img width="1270" height="330" alt="Image" src="https://github.com/user-attachments/assets/69f42f7b-e591-4828-9eee-5501b7f6e1ae" />

**Native:**
<img src="https://github.com/user-attachments/assets/4939d3dd-0fb0-467a-82b8-1a0dd7f48a4f" width="300px" />

hauf-toni:
see my thoughts on this [in this issue thread](https://github.com/digitalfabrik/integreat-cms/issues/3801#issuecomment-3132978576) 🧵

PR #3485, opened 2025-08-22
Consistent imprint wording

michellegraf:
### Short Description
refactoring(wording imprint): consistent wording for imprint implemented

### Proposed Changes
- removed contact from imprint translation
- corrected disclaimer translation for Paschtu
- added {{appName}} to NewGreek (el) pageTitle translation since it was missing while being used in any other place

### Side Effects
- no

### Testing
- Check all places where only "imprint" appears instead of "imprint and contact" (42 languages in total).

### Resolved Issues
- Issue: #3399 consistent wording for imprint

### Notes
- Does the property key "imprintAndContact" now also need to be changed? Since "AndContact" no longer fits.

### DIFF
--- a/translations/translations.json
+++ b/translations/translations.json
@@ -2583,130 +2583,130 @@
   },
   "disclaimer": {
     "de": {
-      "pageTitle": "Impressum und Kontakt"
+      "pageTitle": "Impressum"
     },
     "am": {
-      "pageTitle": "ዕትም እና የግንኙነት መረጃ"
+      "pageTitle": "ዕትም"
     },
     "ar": {
-      "pageTitle": "هيئة التحرير وجهة الاتصال"
+      "pageTitle": "هيئة التحرير"
     },
     "bg": {
-      "pageTitle": "Импресум и контакт"
+      "pageTitle": "Импресум"
     },
     "ckb": {
-      "pageTitle": "زانیارییەکانی ناوەندی بڵاوکەرەوە و زانیارییەکانی پەیوەندی"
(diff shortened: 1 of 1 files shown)


michellegraf:
> Congratulations on your first PR and thank you for taking it! 🔥 💚 Answer to your question: I think we also need to rename the property key and replace it in other components where it is used to avoid confusion 👍

Thanks a lot for your feedback and support! 💚 That makes sense - I'll rename the property key and make sure to replace it in all other components where it is used. I will do the update on Wednesday! Thank you 👍

steffenkleinle:
> > Terrific work, thanks for your work on this! Out of curiosity, how did you adjust the translations in all languages?
> 
> Thank you for your feedback!! Regarding your question: I searched in the translations.json file to see if "Impressum" was already translated somewhere, so that I wouldn’t introduce a new translation for the same term. The value of the property key "disclaimer" was always translated as "Impressum" in the respective language, so I followed that translation and reused it. Only in two cases – in Somali ("so") and Tigrinya ("ti") – the values of "disclaimer" and "imprintAndContact" differed slightly. For those I double-checked with ChatGPT and Google to decide which translation was more accurate.

Perfect solution and very thorough!

steffenkleinle (approved):
Great work, thank you!

LeandraH (approved):
Nicely done, thank you very much! Tested on Firefox

kalairubachandrakumaran-oss:
Tested:
I checked a few languages spot-wise and it looks fine, for example for Munich. If I have to check all of them, please let me know.

LeandraH:
I would say spot-checking is perfect :)

Issue #3489, opened 2025-08-26
Chinese app translations not used


steffenkleinle:
### Describe the Bug

It seems like our Chinese app translations are never used in the web app. Instead, we always use the German fallback.
Same for the native app if setting the OS language to Chinese.

### Steps to Reproduce

1. Go to https://webnext.integreat.app/oberhausen/zh
2. See that all app constants are in German (e.g. Lokale Informationen)

### Expected Behavior

The Chinese translations should be used.

### Related Issues

N/A

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/5g79bechgtbttdseojxces8scw

This is due to the fact that we expect `ch-zn` as language slug while the CMS delivers `zh` only. Adding a fallback should fix the problem.

PR #3490, opened 2025-08-26
3489: Add chinese fallback to show chinese app translations

steffenkleinle:
### Short Description

Add chinese fallback to show chinese app translations.

### Proposed Changes

- Add fallback from `zh` to `zh-CN`
- Add second fallback to the default language for fallback languages (translations only)

### Side Effects

- ~~It seems like using a language slug fallback (i.e. from `zh` to `zh-CN`) breaks the fallback of missing translations to the default language (German), see e.g. http://localhost:9000/muenchen/zh and open the chat. `privacyPolicyInformation` is just displayed as is.~~ Fixed this by adding the default language as second fallback to all fallback languages.

### Testing

Check http://localhost:9000/muenchen/zh and http://localhost:9000/oberhausen/zh and assure that the app is translated to Chinese.

### Resolved Issues

Fixes: #3489

---

### DIFF
--- a/native/src/components/I18nProvider.tsx
+++ b/native/src/components/I18nProvider.tsx
@@ -24,7 +24,7 @@ const I18nProvider = ({ children }: I18nProviderProps): ReactElement | null => {
       const i18nextInstance = i18next.createInstance().use(NativeLanguageDetector)
       await i18nextInstance.init({
         resources,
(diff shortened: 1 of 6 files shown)


LeandraH (approved):
Nice, tested in Firefox

Issue #3483, opened 2025-08-21
All-day event on a day with the time change is shown as lasting two days


LeandraH:
### Describe the Bug

When setting up a monthly recurring all-day event on a day where the time zone switches to summer time (e.g. 29 March 2026), that day shows up as a two-day long event.

### Steps to Reproduce

1. Set up a monthly recurring all-day event on the last Sunday of each month (if the CMS hasn't been reset since opening this ticket, here's one: http://localhost:9000/testumgebung/de/events/monthly-recurring-to-test-time-zone-change)
2. Check on the March 2026 date
3. See that it says "29. März 2026 - 30. März 2026"

### Expected Behavior

It should say "29. März 2026"

### Related Issues

Originally reported in the CMS repo: https://github.com/digitalfabrik/integreat-cms/issues/3224

### Additional Information

<img width="286" height="343" alt="Image" src="https://github.com/user-attachments/assets/33f92a8f-c837-49ac-873d-8c948dee44c5" />

PR #3492, opened 2025-08-26
3483: Fix DST issues for all day recurring events

lunars97:
### Short Description

This PR fixes a bug where all-day events that occur on DST transition days were incorrectly displayed as spanning two days. The issue affected recurring events, particularly when crossing summer winter time changes, causing to show incorrect date ranges.

### Proposed Changes

- Adjusted DateModel.tsx to fix dst issue on recurrence function by localizing first the `actualDate` and then then normalizing back to UTC
- Removed offset because it is no more necessary
- Adjusted tests file

### Side Effects

- i hope none :see_no_evil: 

### Testing

- Go to [Testumgebung](http://localhost:9000/testumgebung/de/events/test-winter-and-summer-time-change) and see that 29 of March is shown without spanning into two days
- Check other events to make sure it does not break anything

### Resolved Issues

Fixes: #3483 

---

### DIFF
--- a/shared/api/models/DateModel.ts
+++ b/shared/api/models/DateModel.ts
@@ -82,7 +82,7 @@ class DateModel {
     return localRecurrenceRule
       .between(minDate, maxDate, true, (_, index) => index < count)
       .map(offsetDate => {
-        const actualDate = DateTime.fromJSDate(offsetDate, { zone: 'local' }).toUTC()
+        const actualDate = DateTime.fromJSDate(offsetDate).toUTC()
         return new DateModel({
           allDay: this.allDay,
           startDate: actualDate,
--- a/shared/api/models/__tests__/DateModel.spec.ts
+++ b/shared/api/models/__tests__/DateModel.spec.ts
@@ -154,8 +154,8 @@ describe('DateModel', () => {
     it('should return exactly count recurrences', () => {
       const recurrenceRule = rrulestr('DTSTART:20230414T050000\nRRULE:FREQ=WEEKLY;BYDAY=MO')
       const date = new DateModel({
-        startDate: DateTime.fromISO('2017-09-27T19:30:00'),
-        endDate: DateTime.fromISO('2017-09-28T21:30:00'),
+        startDate: DateTime.fromISO('2017-09-27T19:30:00+02:00'),
+        endDate: DateTime.fromISO('2017-09-28T21:30:00+02:00'),
         allDay: false,
         recurrenceRule,
       })


shared/api/models/DateModel.ts
```diff
@@ -90,9 +96,18 @@ class DateModel {
     return localRecurrenceRule
       .between(minDate, maxDate, true, (_, index) => index < count)
       .map(offsetDate => {
-        const actualDate = DateTime.fromJSDate(offsetDate).plus({ minutes: offsetDate.getTimezoneOffset() })
+        const actualDate = DateTime.fromJSDate(offsetDate, { zone: 'local' }).toUTC()
```
lunars97, line 99: seems like `getTimezoneOffset` is causing the issue, can't we just set the zone to `local` and convert it to UTC. Just changing this line seems like fixes the issue :thinking: We could also adjust the tests accordingly...

LeandraH: @michellegraf and I just had a look at the tests and couldn't find a spot where the time differed, but we also didn't manage to check all of them. It might work because we're all in the German time zone. Maybe adapt the tests and then run them with a different time zone on your laptop?

shared/api/models/__tests__/DateModel.spec.ts
```diff
@@ -548,25 +560,160 @@ describe('DateModel', () => {
...
-          endDate: DateTime.fromISO('2026-12-06T23:59:00.000+01:00'),
+          startDate: '2028-12-06T00:00:00.000Z',
+          endDate: '2028-12-06T23:59:00.000Z',
+          recurrenceRule,
+          offset: 0,
+        },
+      ])
+    })
+
+    it('should not jump for spring transition for timed and all-day events', () => {
```
LeandraH, line 588: 😍

shared/api/models/DateModel.ts
```diff
@@ -28,14 +28,20 @@ class DateModel {
...
-  }
 
+    if (allDay) {
+      this._startDate = startDate.startOf('day')
+      this._endDate = endDate.startOf('day').set({ hour: 23, minute: 59, second: 0 })
+      this._duration = this._endDate.diff(this._startDate)
+    } else {
+      this._startDate = startDate
+      this._endDate = endDate
+      this._duration = endDate.diff(startDate)
```
bahaaTuffaha, line 42: For consistency:
```suggestion
      this._duration = this._endDate.diff(this._startDate)
```

shared/api/models/DateModel.ts
```diff
@@ -90,13 +96,22 @@ class DateModel {
...
+        const actualDate = DateTime.fromJSDate(offsetDate, { zone: 'local' }).toUTC()
+        if (this.allDay) {
+          return new DateModel({
+            allDay: true,
+            startDate: actualDate.startOf('day'),
+            endDate: actualDate.startOf('day').set({ hour: 23, minute: 59, second: 0 }),
+            recurrenceRule: this.recurrenceRule,
+            offset: actualDate.offset,
+          })
+        }
```
bahaaTuffaha, line 108: 💭 How about adding a comment explaining these changes.

shared/api/models/DateModel.ts
```diff
@@ -90,13 +96,22 @@ class DateModel {
...
         return new DateModel({
-          allDay: this.allDay,
+          allDay: false,
```
bahaaTuffaha, line 110: ❓ why this is false?

bahaaTuffaha (approved):
Great work!
Tested on chrome and from different time zone ✅

shared/api/models/__tests__/DateModel.spec.ts
```diff
@@ -8,6 +8,14 @@ describe('DateModel', () => {
   const locales = ['de', 'en', 'fr', 'ar', 'fa', 'ru']
 
   const normalizeWhitespaces = (str: string) => str.replace(/\s+/g, ' ')
+  const toUTCSpans = (models: DateModel[]) =>
+    models.map(m => ({
+      allDay: m.allDay,
+      recurrenceRule: m.recurrenceRule ?? null,
+      startDate: m.startDate.toUTC().toISO(),
+      endDate: m.endDate.toUTC().toISO(),
+      offset: m.offset,
```
steffenkleinle, line 17: 🙃 
```suggestion
  const toUTCSpans = (dates: DateModel[]) =>
    dates.map(date => ({
      allDay: date.allDay,
      recurrenceRule: date.recurrenceRule ?? null,
      startDate: date.startDate.toUTC().toISO(),
      endDate: date.endDate.toUTC().toISO(),
      offset: date.offset,
```

shared/api/models/__tests__/DateModel.spec.ts
```diff
@@ -147,41 +155,41 @@ describe('DateModel', () => {
     it('should return exactly count recurrences', () => {
       const recurrenceRule = rrulestr('DTSTART:20230414T050000\nRRULE:FREQ=WEEKLY;BYDAY=MO')
       const date = new DateModel({
-        startDate: DateTime.fromISO('2017-09-27T19:30:00+02:00'),
-        endDate: DateTime.fromISO('2017-09-28T21:30:00+02:00'),
+        startDate: DateTime.fromISO('2017-09-27T19:30:00'),
+        endDate: DateTime.fromISO('2017-09-28T21:30:00'),
```
steffenkleinle, line 159: ❓ Why did you change this? From the CMS we always get dates with a timezone offset like before 🤔

shared/api/models/__tests__/DateModel.spec.ts
```diff
@@ -411,28 +420,28 @@ describe('DateModel', () => {
...
+        {
           allDay: false,
           recurrenceRule,
-          startDate: DateTime.fromISO('2024-10-14T10:30:00.000+02:00'),
-          endDate: DateTime.fromISO('2024-10-14T12:00:00.000+02:00'),
-          offset: 120,
-        }),
+          startDate: '2024-10-14T10:30:00.000Z',
+          endDate: '2024-10-14T12:00:00.000Z',
+          offset: 0,
```
steffenkleinle, line 443: ❓ Could you perhaps explain these changes? It seems like the times are now off by two hours here compared to before (before: 10:30 in local time/UTC + 2, now: 10:30 in UTC). Or was the offset taken into account before and therefore basically also made it 10:30 in UTC?

shared/api/models/DateModel.ts
```diff
@@ -28,14 +28,20 @@ class DateModel {
     recurrenceRule: RRuleType | null
     offset?: number
   }) {
+    this._allDay = allDay
     this._recurrenceRule = recurrenceRule
     this._offset = offset ?? startDate.offset
```
steffenkleinle, line 33: ❓ Is `offset` still necessary? I am asking since its `0` in all tests.

steffenkleinle (commented):
Tested on firefox, seems to work great! Nice solution and thanks for looking into this 🎉 Also checked with different time zones :)

shared/api/models/__tests__/DateModel.spec.ts
steffenkleinle: I just tested this without these changes and the tests all succeed. I'd therefore prefer that you revert this change to the DateModel params to correctly reflect the dates we get from the CMS.

steffenkleinle (approved):

PR #3493, opened 2025-08-26
Fix app-toolbelt prepare-metadata usages

steffenkleinle:
### Short Description

With app-toolbelt v0.5.7 a breaking change to prepare-metadata was introduced.
This PR fixes the usage of prepare-metadata.

### Proposed Changes

- Fix app-toolbelt prepare-metadata usages

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -515,7 +515,7 @@ jobs:
             - restore_ruby_cache:
                 directory: native/android
             - run:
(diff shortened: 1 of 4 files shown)


david-venhoff (commented):
Looks mostly good to me, just not sure if the parser accepts the `--metadata` argument or if it only takes it as a positional argument

PR #3499, opened 2025-08-29
3341: Add TTS intent to the Android manifest

LeandraH:
### Short Description

It looks like the intent might be needed in Android API 30 and higher on some Android devices. It's not mentioned in the documentation of `react-native-tts` but there are some mentions of it on Stackoverflow: https://stackoverflow.com/questions/74928642/why-is-android-11-mandating-intent-action-tts-service

And it fixed TTS on David's personal Android device for Lunes: https://github.com/digitalfabrik/lunes-app/pull/1135#discussion_r2291577677

### Proposed Changes

- Add an intent for TTS to the Android Manifest

### Side Effects

Hopefully none

### Testing

Test TTS on an Android device, ideally a real one.

### Resolved Issues

Fixes: #3341 

---

### DIFF
--- a/native/android/app/src/main/AndroidManifest.xml
+++ b/native/android/app/src/main/AndroidManifest.xml
@@ -74,5 +74,8 @@
         <action android:name="android.intent.action.VIEW" />
         <data android:scheme="geo"/>
     </intent>
+    <intent>
+        <action android:name="android.intent.action.TTS_SERVICE" />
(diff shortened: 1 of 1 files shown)


f1sh1918:
I can definitly say that it's a random error that doesn't happen every time on android 11 real device. Because I have one. I would try to increase timeouts and retries

Issue #3459, opened 2025-08-06
Rewrite the CityContentFooter to use a Mui list


LeandraH:
### Describe the Problem

We currently have a list of items in the CityContentFooter, let's use a Mui list with a divider in there instead!

Found here: https://github.com/digitalfabrik/integreat-app/pull/3417#discussion_r2247790309

### Describe Your Preferred Solution

The CityContentFooter to have a nice Mui list with a divider set by the Mui list

### Related Issues

Part of #3231

PR #3505, opened 2025-09-05
3459: Mui list for cityContentFooter

bahaaTuffaha:
### Short Description

We currently have a list of items in the CityContentFooter, let's use a Mui list with a divider in there instead!

### Proposed Changes

- Used `ListItem`, `ListItemButton` and `ListItemText` for CityContentFooter.
- Changed the background color of `FooterContainer` to themeColorLight.
- Adjusted the base List to have horizontal list option.

### Side Effects

It could affect POI because its been used there.

### Testing

- Test the footer at categories page.
- Test the footer at POI Page.
- Test how it behaves on small devices.
- check aschaffenburg , obdach  and malte also.

### Resolved Issues

Fixes: #3459 

---

### DIFF
--- a/web/src/components/CityContentFooter.tsx
+++ b/web/src/components/CityContentFooter.tsx
@@ -25,7 +25,7 @@ const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFoote
   const linkItems = getFooterLinks({ city, language })
   const FooterContent = (
     <StyledList
-      items={linkItems.map(item => (
+      Items={linkItems.map(item => (
         <FooterLink key={item.to} to={item.to} text={item.text} mode={mode} />
       ))}
       horizontal={mode !== 'sidebar'}
--- a/web/src/components/GeneralFooter.tsx
+++ b/web/src/components/GeneralFooter.tsx
@@ -26,7 +26,7 @@ const GeneralFooter = ({ language }: GeneralFooterProps): ReactElement => {
   return (
     <Footer>
       <StyledList
-        items={linkItems.map(item => (
+        Items={linkItems.map(item => (
           <FooterLink key={item.to} to={item.to} text={item.text} />
         ))}
         horizontal={!viewportSmall}


lunars97 (approved):
LGTM except for the color I mentioned above should be changed and I think footer on Poi list looks nice :+1: , so maybe we could keep it like this or change it later if there will be a need

steffenkleinle (changes requested):
The design of the footer for http://localhost:9000/testumgebung/en is quite different than http://localhost:9000/landing/en. I think we should have a consistent design here.

web/src/components/Footer.tsx
```diff
@@ -43,9 +60,9 @@ const Footer = ({ children, overlay = false }: FooterProps): ReactElement => (
   <FooterContainer overlay={overlay}>
     {children}
     {buildConfig().featureFlags.developerFriendly && (
-      <span>
+      <StyledDevText variant='body2'>
```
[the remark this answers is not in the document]
bahaaTuffaha: It looks terrible for the developers 🫣
image

web/src/hooks/useFooterLinks.ts
```diff
@@ -0,0 +1,47 @@
...
+        route: LICENSES_ROUTE,
+      })
+    : RoutePatterns[LICENSES_ROUTE]
+
+  return [
+    { to: disclaimerPath, text: t('imprint') },
+    { to: aboutUrl, text: t('settings:about', { appName: buildConfig().appName }) },
+    { to: privacyUrl, text: t('privacy') },
+    { to: licensesPath, text: t('settings:openSourceLicenses') },
+    ...(accessibilityUrl ? [{ to: accessibilityUrl, text: t('accessibility') }] : []),
```
steffenkleinle, line 44: 🔧 If you just return the key (e.g. `'imprint'` instead of `t('imprint')`), you don't need to make this a hook because of `useTranslation`. Translating when you actually render the links is enough.

Alternatively, you could also directly render the links in this component and rename it to `FooterLinks`.

web/src/components/base/List.tsx
```diff
@@ -1,20 +1,30 @@
...
+const List = ({ items, NoItemsMessage, className, horizontal = false }: ListProps): ReactElement => {
   if (items.length === 0) {
     return typeof NoItemsMessage === 'string' ? <Failure errorMessage={NoItemsMessage} /> : NoItemsMessage
   }
-  return <MuiList className={className}>{withDividers(items)}</MuiList>
+  return horizontal ? (
+    <HorizontalList className={className}>{items}</HorizontalList>
+  ) : (
+    <MuiList className={className}>{withDividers(items)}</MuiList>
+  )
```
steffenkleinle, line 27: ❌ Sorry, I think my comment was not quite clear. What I meant is that there should be no conditional in this component at all. Instead, you should just use this component for normal vertical lists. For the footer, you should have the conditional directly in `Footer.tsx` and use the mui list directly there.

steffenkleinle (changes requested):
I think there should be some minimal padding above the links:
image
Perhaps this needs to be checked with @hauf-toni as well though.

web/src/components/CityContentFooter.tsx
```diff
@@ -30,49 +34,43 @@ type CityContentFooterProps = {
...
+          <ListItemText
+            primary={
+              <Typography variant='body2' textAlign='center'>
+                {text}
+              </Typography>
+            }
+          />
+        </ListItemButton>
+      </StyledListItem>
+    )
```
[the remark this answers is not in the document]
bahaaTuffaha: Sure

web/src/hooks/useFooterLinks.ts
bahaaTuffaha: So do you want me to turn into normal function and move it to utils?

web/src/components/FooterLinksList.tsx
```diff
@@ -0,0 +1,47 @@
...
+const FooterLinksList = ({ linkItems }: FooterLinksListProps): ReactElement[] => {
+  const { t } = useTranslation(['layout', 'settings'])
+
+  return linkItems.map(item => (
+    <StyledListItem key={item.to} disablePadding>
+      <ListItemButton component={Link} to={item.to}>
+        <ListItemText
+          primary={
+            <Typography variant='body2' textAlign='center'>
+              {t(item.text, item.prefix ? { appName: item.prefix } : undefined)}
```
steffenkleinle, line 38: 🤔 What do you think of just removing the app name from the translation? On most other websites, this footer item is only called `About` without the app name in it again. This way, we could a) get rid of this complicated logic here and b) safe some space to avoid content being cut on medium screen sizes.

web/src/components/FooterLinksList.tsx
```diff
@@ -0,0 +1,47 @@
...
+  to: string
+  text: string
+  prefix?: string
+}
+
+type FooterLinksListProps = {
+  linkItems: FooterLinkItem[]
+}
+
+const FooterLinksList = ({ linkItems }: FooterLinksListProps): ReactElement[] => {
```
steffenkleinle, line 29: 🙃 Personally, I think its weird to have a component called `Lists` that only maps items without actually wrapping it in a list. I'd therefore prefer to just have a `FooterLink` component instead and do the mapping in the place where we actually use the items? This way, it would also be easier to apply styling with styled components and the `className` prop, e.g. to set the height of the buttons.
```suggestion
const FooterLink = ({ to, label }: FooterLinkProps): ReactElement => {
```

web/src/components/CityContentFooter.tsx
```diff
@@ -34,43 +21,17 @@ type CityContentFooterProps = {
...
-    )
-  })
-
 const CityContentFooter = ({ city, language, mode = 'normal' }: CityContentFooterProps): ReactElement => {
-  const linkItems = useFooterLinks({ city, language })
-
-  const LinksList = <StyledList NoItemsMessage='' items={linkListItems(linkItems)} horizontal={mode !== 'sidebar'} />
+  const linkItems = getFooterLinks({ city, language })
+  const FooterContent = (
+    <StyledList NoItemsMessage='' items={FooterLinksList({ linkItems })} horizontal={mode !== 'sidebar'} />
```
steffenkleinle, line 27: 🔧 Same as before :)

steffenkleinle (changes requested):
❌ On medium devices, some footer content is cut off: 
image

web/src/components/OverlayFooter.tsx
```diff
@@ -0,0 +1,37 @@
...
+    sx={{
+      '& .MuiTypography-root': {
+        font: '12px/20px "Helvetica Neue", Arial, Helvetica, sans-serif',
+      },
+      '& .MuiListItem-root': {
+        '& a': {
+          height: 30,
+        },
+      },
+    }}>
```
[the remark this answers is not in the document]
bahaaTuffaha: > ❓ Why do we use a different font family here?

I copied the one we use at the production (when you inspect the footer).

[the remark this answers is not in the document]
steffenkleinle: > It takes too much space I will try again anyway

It is okay to use a smaller font size, but we should use the corresponding typography variant instead of hardcoding it here.

> And moved it to sx just to get around Expected class selector ".MuiTypography-root" to be kebab-case  selector-class-pattern

I can't reproduce that, but perhaps you could also fix this by using the exported classes instead, i.e. `typographyClasses`.

web/src/components/GeneralFooter.tsx
```diff
@@ -1,39 +1,31 @@
...
   return (
     <Footer>
-      {viewportSmall ? (
-        <SidebarFooterContainer>
-          <StyledList NoItemsMessage='' items={linkListItems(linkItems)} horizontal={false} />
-        </SidebarFooterContainer>
-      ) : (
-        <StyledList NoItemsMessage='' items={linkListItems(linkItems)} horizontal />
-      )}
+      <StyledList NoItemsMessage='' items={FooterLinksList({ linkItems })} horizontal={viewportSmall === false} />
```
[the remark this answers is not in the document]
bahaaTuffaha: `items: ReactElement[]`  expects array of elements so this will say:
```js
Type 'ReactElement<any, any>' is missing the following properties from type 'ReactElement<unknown, string | JSXElementConstructor<any>>[]': length, pop, push, concat, and 28 more.ts(2740)
List.tsx(8, 3): The expected type comes from property 'items' which is declared here on type 'IntrinsicAttributes & { css?: Interpolation<Theme>; } & ListProps & MUIStyledCommonProps<Theme> & { ...; }'
```

web/src/components/FooterLinksList.tsx
bahaaTuffaha: I think `About us` is a good Idea becouse it directs to https://integreat-app.de/en/about-us/ 👍️

web/src/components/GeneralFooter.tsx
```diff
@@ -25,7 +25,12 @@ const GeneralFooter = ({ language }: GeneralFooterProps): ReactElement => {
 
   return (
     <Footer>
-      <StyledList NoItemsMessage='' items={FooterLinksList({ linkItems })} horizontal={viewportSmall === false} />
+      <StyledList
+        items={linkItems.map(item => (
+          <FooterLink key={item.to} to={item.to} text={item.text} />
+        ))}
+        horizontal={viewportSmall === false}
```
steffenkleinle, line 32: 🙃 
```suggestion
        horizontal={!viewportSmall}
```

web/src/components/OverlayFooter.tsx
```diff
@@ -8,30 +9,21 @@ type OverlayFooterProps = {
...
   display: flex;
-  white-space: nowrap;
   padding: ${props => props.theme.spacing(2)};
-  background-color: rgb(255 255 255 / 50%);
+  background-color: ${props => props.theme.palette.surface.main}80;
+  border-radius: 0;
 
   ${props => props.theme.breakpoints.up('md')} {
     padding: 0 ${props => props.theme.spacing(1)};
   }
```
steffenkleinle, line 18: 🤔 Since there is no overlay footer on screen sizes smaller then `md`, this condition is unnecessary.
🙃 I think it looks better without the horizontal 8px padding which would also give us more space still :)
```suggestion
  padding: 0;
```

web/src/components/OverlayFooter.tsx
```diff
@@ -8,30 +9,21 @@ type OverlayFooterProps = {
...
+
+  & .${listItemClasses.root} a {
+    height: 40px;
```
steffenkleinle, line 21: 🙃 What do you think of reducing the padding here to allow for more space?
```suggestion
    height: 40px;
    padding: 8px;
```
Otherwise it looks like this:
image

web/src/components/CityContentFooter.tsx
```diff
@@ -31,48 +22,21 @@ type CityContentFooterProps = {
...
-    return (
-      <Footer>
-        <SidebarFooterContainer>{Links}</SidebarFooterContainer>
-      </Footer>
-    )
+    return <Footer>{FooterContent}</Footer>
   }
 
-  return <Footer overlay={mode === 'overlay'}>{Links}</Footer>
+  return mode === 'overlay' ? <OverlayFooter>{FooterContent}</OverlayFooter> : <Footer>{FooterContent}</Footer>
```
steffenkleinle, line 39: 🔧 This can be simplified, for modes `sidebar` and `normal` exactly the same thing is rendered. The check for `mode === 'sidebar'` can therefore be removed here (above its still necessary though).
``` tsx
  return mode === 'overlay' ? <OverlayFooter>{FooterContent}</OverlayFooter> : <Footer>{FooterContent}</Footer>
```

steffenkleinle (changes requested):
❌ The contrast with dark theme in the sidebar is sadly not great yet:
image

build-configs/aschaffenburg/theme/index.ts
```diff
@@ -7,9 +7,9 @@ import aschaffenburgFonts from './fonts'
 const customColors = {
   secondary: {
     // TODO we do not yet have light/dark colors for aschaffenburg
-    light: '#009684',
+    light: '#33A695',
     main: '#009684',
-    dark: '#009684',
+    dark: '#007A6B',
```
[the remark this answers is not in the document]
steffenkleinle: @hauf-toni could you please check these? Same for malte and aschaffenburg :)

hauf-toni:
> I think there should be some minimal padding above the links: image Perhaps this needs to be checked with @hauf-toni as well though.

yes, lgtm

web/src/components/OverlayFooter.tsx
bahaaTuffaha: This is perfect 😁

web/src/components/Footer.tsx
```diff
@@ -16,11 +16,13 @@ const FooterContainer = styled('footer')`
...
   background-color: ${props => props.theme.legacy.colors.backgroundAccentColor};
   padding-bottom: ${props => props.theme.spacing(2)};
+  color: ${props => props.theme.palette.text.primary};
 
   ${props => props.theme.breakpoints.up('md')} {
     background-color: ${props =>
       props.theme.isContrastTheme ? props.theme.palette.secondary.dark : props.theme.palette.secondary.light};
-    padding: 0 4px;
+    padding: 8px;
+    color: inherit;
```
steffenkleinle, line 25: 🤔 Is this actually necessary? I don't see a difference

build-configs/aschaffenburg/theme/index.ts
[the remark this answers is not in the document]
steffenkleinle: Just keep those colors for now, there is still the todo :)

web/src/components/Footer.tsx
```diff
@@ -1,36 +1,28 @@
...
-    content: '';
+  align-items: center;
+  gap: ${props => props.theme.spacing(2)};
+  background-color: ${props => props.theme.legacy.colors.backgroundAccentColor};
+  padding-bottom: ${props => props.theme.spacing(2)};
+  color: ${props => props.theme.palette.text.primary};
+
+  ${props => props.theme.breakpoints.up('md')} {
+    background-color: ${props =>
+      props.theme.isContrastTheme ? props.theme.palette.secondary.dark : props.theme.palette.secondary.light};
```
steffenkleinle, line 23: 💭 I think we should change this to tertiary as discussed in https://github.com/digitalfabrik/integreat-app/pull/3505#discussion_r2329545803.
```suggestion
    background-color: ${props =>
      props.theme.isContrastTheme ? props.theme.palette.tertiary.dark : props.theme.palette.tertiary.light};
```

web/src/components/Footer.tsx
```diff
@@ -1,36 +1,28 @@
...
-  & > *::after {
-    padding-inline-end: 10px;
-    content: '';
-  }
-
-  & > *:last-child::after {
-    content: '';
+  align-items: center;
+  gap: ${props => props.theme.spacing(2)};
+  background-color: ${props => props.theme.legacy.colors.backgroundAccentColor};
```
steffenkleinle, line 17: 🔧 I think we should use the same color for both small and big screen sizes. We should not use a legacy color anymore anyway.
```suggestion
  background-color: ${props =>
    props.theme.isContrastTheme ? props.theme.palette.tertiary.dark : props.theme.palette.tertiary.light};
```

steffenkleinle (commented):
Great work and refactoring, only the color missing :)

web/src/components/Footer.tsx
bahaaTuffaha: I can see a difference... the text color got more contrast while contrast mode is enabled (without it its off white)

steffenkleinle (approved):
🎉

Issue #3380, opened 2025-07-15
Rewrite Category lists to use Mui and new design


LeandraH:
### Describe the Problem

Let's rewrite the lists in our 2nd- and 3rd-level categories! We should use the Mui List components, and follow the new design.
New design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=14-751&p=f&m=dev
New design of the whole page: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3074-9734&m=dev
Mui List: https://mui.com/material-ui/react-list/

### Describe Your Preferred Solution

Rewrite our lists in the categories to use Mui lists and to follow the new design. Please make sure that it fits both for the mobile web and the desktop web version.

### Related Issues

Part of #3231

PR #3516, opened 2025-09-09
3380: Mui Category lists

bahaaTuffaha:
### Short Description

Rewrote the lists in our 2nd- and 3rd-level categories.

### Proposed Changes

- Used `ListItem`, `ListItemButton` and `ListItemText` for CategoryList.
- For rtl langauges it already fallback to `Noto Sans Arabic` ex. arabic  (You can check that from Firefox the one that has underline). so I just increased it's weight to 700 because we have only 400 and 700.

### Side Effects

None.

### Testing

- Go to http://localhost:9000/muenchen/de/willkommen or any 2nd level categories.
- Check that the sub icon works if its available.
- Check how it looks on different languages.
- Check 3rd level also.
- Check how it behaves on small screens.

### Resolved Issues

Fixes: #3380 

---


steffenkleinle (changes requested):
Great work!
The indentation of sub category items title is inconsistent if there are only some icons, I think this looks really weird (same for categories actually):
image

Furthermore, if there are no sub category icons, it is not that easy to spot which are categories and which are the sub categories:
image
Perhaps we should remove the dividers between sub category items or add some more indendation @hauf-toni?
image

web/src/components/CategoryListItem.tsx
```diff
@@ -1,61 +1,61 @@
...
-  margin: 0 auto;
-  width: inherit;
-
-  &:hover {
-    color: inherit;
-    text-decoration: inherit;
-    transition: background-color 0.5s ease;
-    background-color: ${props => props.theme.legacy.colors.backgroundAccentColor};
+  [dir='rtl'] & {
+    font-weight: 400;
```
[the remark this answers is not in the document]
bahaaTuffaha: It didn't work with changing the font-weight alone in the second one.

bahaaTuffaha:
> The indentation of sub category items title is inconsistent if there are only some icons, I think this looks really weird (same for categories actually):

Less padding for subcat that has a icon?

web/src/components/CategoriesContent.tsx
```diff
@@ -13,9 +14,7 @@ import OrganizationContentInfo from './OrganizationContentInfo'
 import Page from './Page'
 import Tiles from './Tiles'
 
-const List = styled('ul')`
-  list-style-type: none;
-
+const StyledMuiList = styled(MuiList)`
   & a {
     ${helpers.removeLinkHighlighting}
   }
```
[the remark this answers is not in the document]
bahaaTuffaha: It was in blue link before. 🥲
nice 👍️

web/src/components/CategoriesContent.tsx
```diff
@@ -69,11 +68,11 @@ const CategoriesContent = ({
       content={categoryModel.content}
       lastUpdate={categoryModel.lastUpdate}
       Footer={
-        <List>
+        <StyledMuiList>
```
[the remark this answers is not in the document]
bahaaTuffaha: Yes we can.

bahaaTuffaha:
> What do you think @bahaaTuffaha @steffenkleinle

We can also use the inset variant for subcategories https://mui.com/material-ui/react-divider/#variants
for ex.:

image

1) should we make the icon smaller?
2) So this is bold (700) instead of 500 for main category?
3) what about subcategory Icons:
> The indentation of sub category items title is inconsistent if there are only some icons, I think this looks really weird

web/src/components/CategoryListItem.tsx
```diff
@@ -65,25 +42,39 @@ type CategoryListItemProps = {
...
+          <ListItemAvatar>
+            <Avatar src={category.thumbnail} variant='square' />
+          </ListItemAvatar>
+        )}
+        <ListItemText primary={<StyledTypography variant='title2'>{category.title}</StyledTypography>} />
+      </ListItemButton>
+      {SubCategories.length > 0 && (
+        <>
+          <Divider />
+          <StyledList isInset items={SubCategories} />
```
steffenkleinle, line 74: 🔧 Instead of adding a padding on the button and adding a custom `isInset` property, just add a padding to the whole list:
```suggestion
          <Stack paddingInlineStart={7}>
            <StyledList items={SubCategories} />
          </Stack>
```

web/src/components/CategoryListItem.tsx
```diff
@@ -65,25 +42,39 @@ type CategoryListItemProps = {
...
+  const SubCategories = subCategories.map(subCategory => {
+    const { path, thumbnail, title } = subCategory
+    return (
+      <SubCategoryListItem key={path}>
+        <StyledSubCategoryListItemButton component={Link} to={path}>
+          {!!thumbnail && (
+            <ListItemAvatar>
+              <Avatar src={thumbnail} variant='square' />
+            </ListItemAvatar>
+          )}
```
steffenkleinle, line 54: 🔧 As discussed with @hauf-toni and @laravoelk: We should remove 3rd level icons here.
```suggestion
```

web/src/components/CategoryListItem.tsx
```diff
@@ -65,25 +42,39 @@ type CategoryListItemProps = {
...
-      <SubCategoriesContainer>{SubCategories}</SubCategoriesContainer>
-    </Row>
+    <StyledListItem disablePadding>
+      <ListItemButton component={Link} to={category.path} dir='auto'>
+        {!!category.thumbnail && (
+          <ListItemAvatar>
+            <Avatar src={category.thumbnail} variant='square' />
+          </ListItemAvatar>
+        )}
+        <ListItemText primary={<StyledTypography variant='title2'>{category.title}</StyledTypography>} />
```
steffenkleinle, line 69: 🙃 
```suggestion
        <ListItemText primary={<Typography variant='title2' fontWeight='bold'>{category.title}</Typography>} />
```

web/src/components/CategoryListItem.tsx
```diff
@@ -65,25 +42,39 @@ type CategoryListItemProps = {
...
-    <Row>
-      <StyledLink dir='auto' to={category.path}>
-        {!!category.thumbnail && <CategoryThumbnail alt='' src={category.thumbnail} />}
-        <CategoryItemCaption>{category.title}</CategoryItemCaption>
-      </StyledLink>
-      <Divider />
-      <SubCategoriesContainer>{SubCategories}</SubCategoriesContainer>
-    </Row>
+    <StyledListItem disablePadding>
+      <ListItemButton component={Link} to={category.path} dir='auto'>
```
steffenkleinle, line 63: ❓ Is this direction still necessary?
```suggestion
      <ListItemButton component={Link} to={category.path}>
```

web/src/components/CategoryListItem.tsx
```diff
@@ -1,62 +1,39 @@
...
-  font-weight: bold;
-  text-decoration: inherit;
-  height: 100%;
-  min-width: 1px; /* needed to enable line breaks for too long words, exact value doesn't matter */
-  flex-grow: 1;
-  word-wrap: break-word;
-`
+const StyledSubCategoryListItemButton = styled(ListItemButton)`
+  padding-left: 72px;
+  min-height: 56px;
```
steffenkleinle, line 31: ❓ Personal opinion, I think it looks better if the sub category list items are a little bit shorter without the min-height and it takes less space. Also, the distinction between categories and sub categories gets bigger through this.
```suggestion
```

web/src/components/CategoryListItem.tsx
```diff
@@ -1,62 +1,39 @@
...
 
-const CategoryThumbnail = styled('img')`
-  width: 30px;
-  height: 30px;
-  padding: 0 5px;
-  flex-shrink: 0;
-  object-fit: contain;
-  filter: ${props => (props.theme.isContrastTheme ? 'invert(1)' : 'none')};
+const SubCategoryListItem = styled(ListItem)`
+  padding: 0;
```
steffenkleinle, line 26: 🔧 Please use `disablePadding` instead on the `ListItem`.

web/src/components/ThemeContainer.tsx
```diff
@@ -79,6 +79,16 @@ const createTheme = (
...
+      MuiAvatar: {
+        styleOverrides: {
+          root: {
+            backgroundColor: 'transparent',
+            '.MuiAvatar-fallback': {
+              display: 'none',
+            },
+          },
+        },
+      },
```
steffenkleinle, line 91: 🤔 Probably we are actually abusing the mui avatar for list item icons, there is also a `ListItemIcon` component, could you perhaps try to use that instead? No need though as we did the same before.
https://mui.com/material-ui/api/list-item-icon/

web/src/components/CategoryListItem.tsx
bahaaTuffaha: :thinking:  is there any advantage from using `Stack` here?

steffenkleinle: You can also add it as a style to `StyledList`, doesnt matter. I'd just prefer that over passing it thorugh to the items and the divider.

web/src/components/CategoryListItem.tsx
bahaaTuffaha: Nice catch ... no need for it. and fixes the icon alignment for RTL 👍️

web/src/components/ThemeContainer.tsx
bahaaTuffaha: If the icon is missing it will look like a missing profile image 😅. so I just hide it on fallback.
image

steffenkleinle: The difference is that for `ListItemIcon` you don't need to override the default of displaying a fallback profile image which is definitely not the "right" thing to do 😅  The `ListItemAvatar` is intended for profile pictures which is not something we use. Therefore please use `ListItemIcon`. We should also adjust all other usages of `ListItemAvatar` but we can do that separately.

web/src/components/CategoryListItem.tsx
steffenkleinle: @hauf-toni what do you think:
<p float="left">
image
image
</p>

bahaaTuffaha: hmm  I can see there is padding on the list other than the height difference...

steffenkleinle: Yes, without a min-height this would also need to be removed again as you did before.

bahaaTuffaha: on small screens without `min-height: 56px;` we will end up with 3 different sizes instead of 2
;ist

steffenkleinle: Personally I think its fine that longer items just take more space.

lunars97 (approved):
LGTM! Thank you for your effort 💚 I have one small change suggestion above :)

Issue #3381, opened 2025-07-15
Rewrite POIs list to use Mui


LeandraH:
### Describe the Problem

Let's rewrite the POIs list to use the Mui List components and the new design!
Design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3048-12029&m=dev
New design of the whole page: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3048-1950&m=dev
Mui List: https://mui.com/material-ui/react-list/

### Describe Your Preferred Solution

Rewrite our POIs list to use Mui lists and follow the new design. Please make sure that it works well in all screen sizes.

### Related Issues

Part of #3231

Issue #3508, opened 2025-09-08
Use mui typography components for texts


steffenkleinle:
### Describe the Problem

We currently still have lots of usages of the legacy fonts and other ways to render texts. Examples are for example the captions.

### Describe Your Preferred Solution

We should consistently use the mui `Typography` component with the correct variant applied.
Check the docs for the possible options and props: https://mui.com/material-ui/react-typography/

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3526, opened 2025-09-14
3381: Rewrite poi list to use mui

lunars97:
### Short Description

Rewrite poi list to use MUI list

### Proposed Changes

-

### Side Effects

-

### Resolved Issues

Fixes: #3381 

---


web/src/components/PoiListItem.tsx
```diff
@@ -74,16 +52,31 @@ const PoiListItem = ({ poi, distance, selectPoi }: PoiListItemProps): ReactEleme
...
-      </ListItemContainer>
+      <ListItem disablePadding>
+        <StyledListItemButton onClick={selectPoi} id={slug} to={slug} component={Link} aria-label={title}>
+          <StyledListItemAvatar>
+            <Avatar src={thumbnail || PoiThumbnailPlaceholder} alt='' variant='square' />
+          </StyledListItemAvatar>
+          <StyledListItemText
+            slotProps={{
+              secondary: {
+                component: 'div',
```
lunars97, line 63: added because of the console error `In HTML, <p> cannot be a descendant of <p>.
This will cause a hydration error.`

web/src/components/PoisDesktop.tsx
```diff
@@ -126,7 +122,9 @@ const PoisDesktop = ({
         {canDeselect ? (
           <GoBack goBack={deselect} text={t('detailsHeader')} />
         ) : (
-          <ListHeader>{t('listTitle')}</ListHeader>
+          <StyledTypography component='h1' variant='h3'>
```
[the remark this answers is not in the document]
lunars97: I used it because this is the main heading here and according to sematic structure, it should be `h1`. I also have the question regarding the Typography variants, why do we have `h1`, `h2`, `h3`  variants? Do we intend to use them because I think they have really big font sizes and those font sizes have not been used so far. It would be good to use them for headings and subheadings. What do you think @steffenkleinle @LeandraH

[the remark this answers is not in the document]
steffenkleinle: I don't know about html semantics that well, but I am not sure if `Near you` is even our main headline here? For me its more just a label/hint with some additional info on the list, but this might be completely wrong as well. `h1`, `h2` and `h3` variants should be used for those captions, yes. The issue to do this is #3508.
However, we currently don't really have any h2/h3 headlines on the app side but only in the content from the CMS.

web/src/components/PoiListItem.tsx
```diff
@@ -1,65 +1,42 @@
...
+  gap: ${props => props.theme.spacing(2)};
+` as typeof ListItemButton
 
-  &:first-of-type {
-    padding-top: 0;
+const StyledListItemAvatar = styled(ListItemAvatar)`
+  [class*='MuiAvatar-root'] {
+    margin-top: 8px;
+    width: 94px;
+    height: 94px;
```
bahaaTuffaha, line 26: How about changing this to 64:
The design says `56px` I feel that the ratio of the design isn't right.
so I picked something slightly above: 
```suggestion
    width: 64px;
    height: 64px;
```
And no need to `margin-top: 8px;`

web/src/components/PoiListItem.tsx
```diff
@@ -74,16 +51,27 @@ const PoiListItem = ({ poi, distance, selectPoi }: PoiListItemProps): ReactEleme
...
-      </ListItemContainer>
+      <ListItem disablePadding>
+        <StyledListItemButton onClick={selectPoi} id={slug} to={slug} component={Link} aria-label={title}>
+          <StyledListItemAvatar>
+            <Avatar src={thumbnail || PoiThumbnailPlaceholder} alt='' variant='square' />
+          </StyledListItemAvatar>
+          <StyledListItemText
+            disableTypography
+            primary={
+              <StyledTypography variant='title2' component='h2'>
```
bahaaTuffaha, line 62: What matches the design is `body2` (fontSize: 14, fontWeight: 400)
```suggestion
              <StyledTypography variant='body2' component='h2'>
```

web/src/components/PoiListItem.tsx
```diff
@@ -74,16 +51,27 @@ const PoiListItem = ({ poi, distance, selectPoi }: PoiListItemProps): ReactEleme
...
+                {title}
+              </StyledTypography>
+            }
+            secondary={
+              <StyledTypography variant='body1' component='div'>
```
bahaaTuffaha, line 67: What matches the design here is `body3` (fontSize: 12, fontWeight: 400)
You have to adjust the color to slightly gray like the design.
```suggestion
              <StyledTypography variant='body3' component='div'>
```

web/src/components/PoiListItem.tsx
```diff
@@ -1,65 +1,42 @@
...
 
-const ListItemContainer = styled('ul')`
-  font-family: ${props => props.theme.legacy.fonts.web.contentFont};
-  display: flex;
-  padding: clamp(10px, 1vh, 20px) 0;
-  cursor: pointer;
+const StyledListItemButton = styled(ListItemButton)`
+  align-items: flex-start;
+  gap: ${props => props.theme.spacing(2)};
+` as typeof ListItemButton
```
bahaaTuffaha, line 20: Add Padding 12 Top-Bottom and 16 for Left-Right
  `min-height` to keep the visual weight consistent 🙂.
  ```js
  padding: 12px 16px;
  min-height: 80px;
  ```

bahaaTuffaha (changes requested):
Here's my changes to make the visual weight matches the design as much as possible.

web/src/components/PoiListItem.tsx
lunars97: hmm :thinking: should we stricktly follow the design? for me this font seems too small...

web/src/components/PoiListItem.tsx
bahaaTuffaha: image

web/src/components/PoiListItem.tsx
lunars97: I would say it looks better when they are on the same level and it would be better to keep the margin there :thinking:

web/src/components/PoiListItem.tsx
lunars97: this also looks good, so i would also go for it :+1:

web/src/components/List.tsx
steffenkleinle, line 1: ❌ This is our old legacy list component. Please use the one in the `base` folder instead. Sorry for the confusion, we should delete this file asap.

web/src/components/PoiListItem.tsx
```diff
@@ -73,19 +46,24 @@ const PoiListItem = ({ poi, distance, selectPoi }: PoiListItemProps): ReactEleme
...
+          <StyledAvatar src={thumbnail || PoiThumbnailPlaceholder} alt='' variant='square' />
+        </ListItemAvatar>
+        <StyledListItemText
+          slotProps={{ primary: { component: 'h2' }, secondary: { component: 'div' } }}
+          primary={title}
+          secondary={
+            <>
+              {' '}
+              {distance !== null && <p>{t('distanceKilometre', { distance: distance.toFixed(1) })}</p>}{' '}
+              <p>{category.name}</p>{' '}
```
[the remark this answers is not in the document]
lunars97: hmm I think it was generated by linting ...

Issue #3511, opened 2025-09-08
Use mui drawer to replace sidebar


steffenkleinle:
### Describe the Problem

We currently implement a custom sidebar component while mui provides the drawer as sidebar solution:
https://mui.com/material-ui/react-drawer/

### Describe Your Preferred Solution

Replace our custom sidebar with a mui drawer.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

https://mui.com/material-ui/react-drawer/

Issue #3507, opened 2025-09-08
Merge and migrate SidebarActionItem and ContrastToggle to mui


steffenkleinle:
### Describe the Problem

We currently have to very similar components with different names: SidebarActionItem and ContrastToggle

### Describe Your Preferred Solution

Delete `ContrastToggle` and use `SidebarActionItem` instead.
Use mui components for `SidebarActionItem`.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

Issue #3486, opened 2025-08-25
Implement mui bottom navigation


steffenkleinle:
### Describe the Problem

On mobile devices it is considered best practice to have a bottom navigation instead of tabs for the main navigation routes.

### Describe Your Preferred Solution

Implement bottom navigation on mobile screens (sizes tbd) using https://mui.com/material-ui/react-bottom-navigation.

### Describe Possible Alternatives

N/A

### Related Issues

Discussed in https://github.com/digitalfabrik/integreat-app/issues/3375

### Additional Information

https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=2831-17369&p=f&m=dev

steffenkleinle:
@hauf-toni this is how it would look with the mui defaults:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/48b1a929-2133-488c-8307-8587b43c7db1" />

I think it looks fairly good, not sure what to do with the text `local information` as it is too long for the navigation item. Should we reduce the text size/paddings? Or can we come up with an alternative text? We definitely had this discussion in the past though and did not find a better term, but perhaps there is now? If we decide to go with another term, we should definitely check that with the service team and make sure its consistent everywhere (native, web, cms).

steffenkleinle:
> > [@hauf-toni](https://github.com/hauf-toni) this is how it would look with the mui defaults:
> 
> > I think it looks fairly good, not sure what to do with the text `local information` as it is too long for the navigation item. Should we reduce the text size/paddings? Or can we come up with an alternative text? We definitely had this discussion in the past though and did not find a better term, but perhaps there is now? If we decide to go with another term, we should definitely check that with the service team and make sure its consistent everywhere (native, web, cms).
> 
> Hmm, tough question. Overall I'm very happy how the screendesign looks, so great job on that! Concerning font + padding: reducing that should be a last resort as readability and touch targets are more important than squeezing text in and it would not benefit the overall look of the bottom nav bar.

Agreed, also the touch target would not be reduced, still the whole icon would respond to touches. I also agree that readability is more important and that it would definitely not improve the looks.

[...]

PR #3528, opened 2025-09-17
3511: Use mui drawer for sidebar

steffenkleinle:
### Short Description

This PR removes our custom portal sidebar implementation and replaces it with a mui drawer.

### Proposed Changes

- Use mui drawer to replace our custom sidebar implementation
- Hide navigation tabs on small screens
- Use mui components for SidebarActionItems
- Delete redundant ContrastToggle

### Side Effects

- Navigation items on small screens will only be readded with #3486

### Testing

Use a small device and checkout the sidebar and the language selection.
Make sure all sidebar actions work and it looks good with both a light and dark theme.
Make sure the high contrast mode toolbar item works on big screens as well.

### Resolved Issues

Fixes: #3511, #3507

---


web/src/components/Sidebar.tsx
```diff
@@ -1,124 +1,60 @@
...
-              <StyledIconButton onClick={() => setShow(false)} aria-label={t('sideBarCloseAriaLabel')}>
-                <CloseIcon />
-              </StyledIconButton>
-            </ActionBar>
-          </Heading>
-          <Content headerHeight={headerHeight}>{children}</Content>
-          {Footer}
-        </SidebarContainer>
-      </Portal>
+      <Drawer open={open} onClose={() => setOpen(false)} container={drawerContainer} anchor='right'>
```
LeandraH, line 47: 💭 Maybe a comment here that the `anchor='right'` is flipped automatically? Then again, might not be necessary. I'm not sure 😅

LeandraH (approved):
Nice, thank you! Tested in Firefox. But I think it would look better if the items in the sidebar were full-width or at least centered

Image

steffenkleinle:
> Nice, thank you! Tested in Firefox. But I think it would look better if the items in the sidebar were full-width or at least centered

Good catch, should be fixed.

web/src/components/Sidebar.tsx
steffenkleinle: Since mui takes care of a lot of direction specific things, I think this is to be expected and we would need to add this kind of comment in a few places. I'd therefore not put one.

web/src/components/SidebarActionItem.tsx
```diff
@@ -1,37 +1,26 @@
...
-const SidebarActionItem = ({ text, iconSrc }: SidebarActionItemProps): ReactElement => (
-  <Container aria-label={text} dir='auto'>
-    <StyledIcon src={iconSrc} />
-    <span>{text}</span>
-  </Container>
+const SidebarActionItem = ({ text, icon, to, onClick }: SidebarActionItemProps): ReactElement => (
+  <ListItem disablePadding>
+    <ListItemButton dir='auto' component={to ? Link : ListItemButton} to={to} onClick={onClick}>
+      <ListItemAvatar>
+        <Avatar>{icon}</Avatar>
```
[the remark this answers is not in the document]
steffenkleinle: Nice!

steffenkleinle:
> I found a couple of small issues:
> 
>     1. Language selector's close button is moving with the list.
> 
>     2. Scrolling to top after closing the sidebar.
> 
> 
> 2025-09-17.21-25-49.mp4

Thanks for the careful testing, nice finds! I pushed fixes for hopefully all the problems, please test again :)

bahaaTuffaha (approved):
Nice work... I like the animation on this Drawer.
Tested on Chrome and Firefox ✅

One thing that occurs if I went from desktop screen to mobile then closed the sidebar:
image

PR #3530, opened 2025-09-17
3508: Use mui typography for texts

lunars97:
### Short Description

Rewrite all text occurences with MUI Typography component

### Proposed Changes

- Replaced legacy fonts with MUI Typography component and its correct variants
- Adjusted font colors where necessary, introduced new color `neutral` in TextType (see the [discussion])(https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3063-8006&m=dev) 
- Removed `Local Information` headline from the main page as it is shown in [figma](https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3046-12275&m=dev)

### Side Effects
-should be none

### Testing
Check font style of texts in all visible components :sparkles: 

### Resolved Issues

Fixes: #3508 

---


web/src/components/Contact.tsx
```diff
@@ -32,11 +36,17 @@ const Contact = ({
...
+          <StyledTypography variant='label1' component='h2'>
+            {t('contactInformation')}
+          </StyledTypography>
+        )}
+      </StyledContactHeader>
       {!!website && (
         <ContactItem
           iconSource={PublicOutlinedIcon}
-          iconAlt={t('website')}
+          iconAlt=''
```
LeandraH, line 49: Thanks for always keeping an eye out for the a11y tags! But I'm actually not sure if this one needs to go, I would possibly find it useful to know that the thing after the icon is a website.

web/src/components/PageDetail.tsx
```diff
@@ -15,13 +16,17 @@ type PageDetailProps = {
 
 const PageDetail = ({ identifier, information, path }: PageDetailProps): ReactElement => (
   <div>
-    <Identifier>{identifier}: </Identifier>
+    <StyledTypography variant='label1' component='span'>
+      {identifier}:{' '}
```
LeandraH, line 20: 😅

web/src/components/PoiDetails.tsx
```diff
@@ -53,8 +52,8 @@ const Thumbnail = styled('img')`
   }
 `
 
-const Distance = styled('div')`
-  ${helpers.adaptiveFontSize};
+export const StyledTypography = styled(Typography)<TypographyProps>`
```
LeandraH, line 55: ❓ Does this need to be exported?

LeandraH (commented):
This looks great, that must have been so much work to find all these occurrences! One question though, should the outcome of this ticket be that the new font is used everywhere? Because it's currently not used in e.g. the chat button or the footer.

Also, I wonder if we should add the neutral color somewhere in the base Typography so that we don't have to add it manually everywhere.

web/src/components/PoiDetails.tsx
lunars97: I reused it in another component, but it could be a bad practice ...

Issue #3529, opened 2025-09-17
Move toolbar to sidebar on small screens


steffenkleinle:
### Describe the Problem

Since we decided to implement a bottom navigation toolbar (#3486) on small screens, the bottom of the screen gets pretty crowded and a footer does not really fit there anymore.
It gets further crowded with the chat button and the tts player.

### Describe Your Preferred Solution

We should therefore move the toolbar to the sidebar on small screens.
Discuss which items still make sense there/are not necessary on small screens (e.g. share).

### Describe Possible Alternatives

N/A

### Related Issues

Blocks: #3486 

### Additional Information

N/A

Issue #3513, opened 2025-09-08
Use mui dialog for the dialogs


steffenkleinle:
### Describe the Problem

We currently implement a custom modal/dialog for our feedback form, the chat and the tts help modal.

### Describe Your Preferred Solution

Instead of implementing our own dialog, use the mui dialog.
The mui dialog allows to be full screen on small sizes only:
https://mui.com/material-ui/react-dialog/#responsive-full-screen

Using the mui dialog hopefully also fixes the issue of opening the feedback from the sidebar. Otherwise, fix it here or create a new issue for it.

### Describe Possible Alternatives

N/A

### Related Issues

Blocked by: #3365 

### Additional Information

https://mui.com/material-ui/react-dialog/

PR #3531, opened 2025-09-18
3529: Move toolbar to sidebar on small screens

steffenkleinle:
### Short Description

This PR removes the toolbar on small screens and moves all its content to the sidebar.

### Proposed Changes

- Only show the toolbar on large screens
- Show all toolbar items in the sidebar (except sharing as it is easy to share using the OS/browser)
- Rewrite the toolbar and toolbar items using mui components
- Improve padding between toolbar and content on large screens
- Delete CategoriesToolbar and move logic to PdfToolbarItem and CityContentToolbar
- Refactor CityContentToolbar and remove now redundant props as there is no toolbar on small viewports anymore

### Side Effects

- The toolbar in the pois sidebar still needs some work: #3532 
- The feedback does not open correctly yet: #3513 
- Location change has been removed from the sidebar
- I used a new visual element ([card](https://mui.com/material-ui/react-card/)) for the feedback in the sidebar

### Testing

Test the sidebar and toolbar on different routes and screen sizes and make sure everything works and looks as expected.

### Resolved Issues

Fixes: #3529

---

### DIFF
--- a/web/src/components/Sidebar.tsx
+++ b/web/src/components/Sidebar.tsx
@@ -25,6 +25,9 @@ const StyledDrawer = styled(Drawer)`
 const Header = styled(Paper)`
   position: fixed;
   width: inherit;
+
+  /* Position header above sidebar content */
+  z-index: 1;
 `
 
 type SidebarProps = {


LeandraH (approved):
Nice, thank you! Tested on Firefox

I have a couple of small things:

1. I don't love the paper for the feedback buttons, I would expect them to either have a more visible separation from the background (maybe with a bigger shadow or a different background color?) or to not be separate from the background.
Image

2. That sidebar is very different from the design, I assume you've cleared that up with Toni already?

Image

3. One of my favorite bloggers just published a post about reserving space for the scrollbar so that you don't get that jumping like we do now on longer pages when opening the sidebar: https://gomakethings.com/prevent-scrollbar-jump/ I haven't tested that yet but maybe it helps here? But also, the jumping is not exactly a new issue 😅

LeandraH (changes requested):
Actually, sorry, I just noticed that you can't give feedback from the Sidebar. Or at least I can't. When I click on either button, the screen doesn't change, and when I then close the sidebar, the feedback shortly shows up but then disappears before I can interact with it.

https://github.com/user-attachments/assets/3ad563a6-4c8a-4551-8694-6b454ef18532

LeandraH:
> Can you elaborate a little more what jumping issue you mean? I am not sure I know what problem you are talking about :D Or at least I don't spot any jumping 🙈

When you are on a page with a scroll bar, that scroll bar disappears when you open the sidebar. In our case, you mostly only notice it because the icon to close the sidebar is not at the same place as the icon to open it.

https://github.com/user-attachments/assets/fa5f7358-9610-44ee-9edd-efb307dd2826

LeandraH (approved):
Woops, just saw that you already opened a ticket for it :D

Issue #3023, opened 2024-12-09
Implement Fallback (Sprachauswahl) if user's device language is not available 


hauf-toni:
### Describe the Problem
Currently, if a user's device language is unavailable in the app, the content is displayed in german. Defaulting to a specific language could exclude or confuse users who are not proficient in this language. 

Providing a language selection screen ensures inclusivity and allows users to make their own choice. This approach improves inclusivity for users from diverse linguistic backgrounds, reduces frustration and aligns with best practices.

### Proposed Solution
When a user's device language is not supported, the language selection should be triggered. Once the user selects a language, their choice should be saved locally (if possible) to avoid showing the screen again unnecessarily.

Details:
- Web:
  1. The language selection in the header should be visible on ALL routes, including city selection, main disclaimer, ...
  2. ~~[Optional] The language not available dialog should be displayed on all routes, including city selection, main disclaimer, ...~~: Too much effort for too little effect, the language selection is easily available now.
  3. The language not available message should be adjusted to `The currently selected language is not available in this region. Please select another language.`
- Native: Everything will stay as it is

### Alternatives
Keep German as Fallback/Use English instead.

### Additional Context
https://github.com/digitalfabrik/integreat-app/issues/3023#issuecomment-3298612862

steffenkleinle:
@hauf-toni this is both for native and web, right?
On web, we already have the language selection once you navigate to a city (i.e. we dont have it on the landing/city selection page, the disclaimer and so on).
- Possible problem on web: If you navigate to the landing page with a not supported language, you are shown a language selection. If you then select a language and then select to a city where this language is not available, you will be presented with another language selection. I can imagine that this would be confusing/frustrating for users
- Possible problem on native: The UI language does not necessarily equal the content language. We currently have a language selection for the content language upon navigation to a city that does not support the currently selected language. It seems like the issue is more about the UI/app language. Adding separate language selection for that could be confusing/frustrating as well.
What do you think?

hauf-toni:
Thanks for your thorough insights and highlighting these challenges @steffenkleinle this would actually be for both native and web (although for native I don't think it would be a problem for more than once, right?). 

A: This is indeed quite tricky. I think it would be useful to provide a bit more context for the second language selection, e.g. “The language you have selected is currently not available. Please select another language.” Could we follow this flow and show it to the user?

B: Could you explain why we currently have the option of displaying two different languages at the same time, e.g. headlines in farsi and text in german?

[...]

PR #3535, opened 2025-09-20
3023: Improve language selection

steffenkleinle:
### Short Description

This PR improves the message if a language is not available.
Furthermore, it makes all non-city-content routes language specific as well (i.e. main disclaimer, landing, consent, not cooperating, not found) and adds an option to select another language.

### Proposed Changes

- Improve language not available message and remove duplicated translations
- Show header for landing page
- Make all routes language specific and show language header item
- Add sidebar on all routes (to be able to switch to high contrast mode)
- Increase logo size to 48px on small screens

### Side Effects

- The current solution to get all available languages is quite hacky. See comment below.

### Testing

Check all non-city-content routes and change the language. Check the language not available page (e.g. http://localhost:9000/testumgebung/adsf). Check the sidebar and the header on landing and small screens.

### Resolved Issues

Fixes: #3023

---


steffenkleinle:
We have a small problem here: There seems to be currently no way to query all available languages from the CMS and in the app code we don't have any language names, only the slugs. In order to be able to show the language change menu also on non-city routes, I had to implement a quite hacky solution to check all languages of all cities:
``` ts
  const { data: cities } = useLoadFromEndpoint(createCitiesEndpoint, cmsApiBaseUrl, undefined)
  const languages = [
    ...new Map(cities?.flatMap(city => (city.live ? city.languages : [])).map(item => [item.code, item])).values(),
  ]
```
However, this is not very elegant and inefficient. Any better ideas? @digitalfabrik/cms-team is there perhaps already a languages endpoint that is just not yet in the docs/how much effort would it be to add an endpoint for all currently enabled languages?

bahaaTuffaha (approved):
Tested on Chrome ✅️

I noticed all of the languages are enabled at the action button.
for example:
http://localhost:9000/tuebingen/ar
is that ok?
Image

steffenkleinle:
> Tested on Chrome ✅️
> 
> I noticed all of the languages are enabled at the action button. for example: http://localhost:9000/tuebingen/ar is that ok? Image

Good find, fixed! Now it only shows the actually available languages in that city.

Issue #3525, opened 2025-09-12
Fix language direction when English word used in middle of Arabic sentence


bahaaTuffaha:
### Describe the Bug

If you hover above `Create a PDF` the tooltip will show switched sentences around the word PDF due to different direction.
In Arabic it should be:
<p>اصدار ملف PDF غير متاح حاليا في هذه اللغة.<p>

### Steps to Reproduce

1. Go to http://localhost:9000/muenchen/en (new design ticket or main)
2. Switch to Arabic.
3. hover above `Create a PDF`.

### Expected Behavior

Texts that are Arabic or in other RTL languages should also be dispalyed RTL.
We could try to set the `direction` prop of `Typography` to `auto` as default in `ThemeContainer`/`theme`.

### Additional Information

<img width="497" height="466" alt="Image" src="https://github.com/user-attachments/assets/a89eb6a3-9e2a-4dff-a1f7-6514c67d133d" />

PR #3542, opened 2025-09-23
3525: Fix direction of mixed rtl/ltr texts

steffenkleinle:
### Short Description

This PR applies `dir='auto'` for all typography and tooltips.

### Proposed Changes

- Set `dir='auto'` as default prop for mui `Typography` and `Tooltip`

### Side Effects

N/A

### Testing

See issue.

### Resolved Issues

Fixes: #3525

---

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -34,7 +34,6 @@ const Row = styled('div')`
   flex-wrap: wrap;
   overflow-x: auto;
   padding: 0 16px;
-  gap: 8px;
 
   ${props => props.theme.breakpoints.down('md')} {
     padding: 0 8px;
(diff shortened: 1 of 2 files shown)


web/src/components/Header.tsx
```diff
@@ -70,8 +71,10 @@ export const Header = ({ actionItems = [], logoHref, cityName, language, TabBar
       <Paper>
         <HeaderContainer ref={ref}>
           <Row>
-            <HeaderLogo link={logoHref} />
-            {!!cityName && <HeaderTitle title={cityName} landingPath={landingPath} />}
+            <Stack direction='row-reverse' gap={1}>
```
steffenkleinle, line 74: Can anyone tell me why I need to set `row-reverse` here? 😨  I don't get it 🙈

lunars97: shouldn't it be used within conditional, if it is rtl then row-reverse? :thinking:

steffenkleinle: In theory it should not be necessary at all since this is usually automatically handled by mui. But its actually necessary for both ltr and rtl.

bahaaTuffaha: I think it has to do with the order at the `HeaderLogo` but your solution is ok.

bahaaTuffaha (approved):
I found another issue 🙂:
The image should be on the right side.

Image

Issue #3532, opened 2025-09-18
Move toolbar from poi panel


steffenkleinle:
### Describe the Problem

There is only limited space in the poi panel such that the toolbar items currently overflow. We therefore need a solution for the toolbar in the poi panel (desktop).
We already limit the amount of items shown (we hide the tts and contrast mode toolbar items) which is also not optimal.

### Describe Your Preferred Solution

We should think of a scalable solution for toolbar items for the pois desktop.
Options:
- We could implement a sidebar like solution (with a header in the poi panel), similar to the CityContentSidebar on small devices
- We could use a mui menu: https://mui.com/material-ui/react-menu/
- ...

### Describe Possible Alternatives

Keep it as it is and just optimize the styling/prevent items from overflowing.

### Related Issues

Overflowing partly introduced in: #3531

### Additional Information

<img width="866" height="400" alt="Image" src="https://github.com/user-attachments/assets/eb7151b3-b307-4926-a66b-78d981f16828" />

steffenkleinle:
@hauf-toni do you have any opinion or other idea here?

hauf-toni:
This option is much better imo since users can connect the content and the options in the sidebar directly: 

<img width="514" height="313" alt="Image" src="https://github.com/user-attachments/assets/e7f9181e-a635-4137-8e15-5ce8cb2531b1" />

Thank you for your proposal, I would approve this solution. If you would like a design on how the options are displayed once the sidebar is clicked by the user, please let me know.

PR #3546, opened 2025-09-24
3532: Move poi toolbar from poi panel to sidebar

steffenkleinle:
### Short Description

This PR adds a sidebar menu to the poi panel and moves the toolbar there.

### Proposed Changes

- Add poi sidebar menu with the toolbar to the panel header
- Remove back button on poi mobile bottom sheet
- Position thumbnail below general information
- Rename and merge some translations

### Side Effects

- Back button for the poi bottom sheet on mobile was removed

### Testing

Check the pois on both desktop and mobile and for both the list and detail view. Check the sidebar.

### Resolved Issues

Fixes: #3532

---


web/src/components/BottomActionSheet.tsx
```diff
@@ -76,7 +69,6 @@ const BottomActionSheet = ({
       scrollLocking={false}
       blocking={false}
       onSpringStart={initializeScrollElement}
-      onSpringEnd={() => setBottomActionSheetHeight(bottomSheetRef.current?.height ?? 0)}
```
LeandraH, line 79: ❓ Why did you remove this?

web/src/components/PoiPanelHeader.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useTranslation } from 'react-i18next'
+
+import CityContentSidebar from './CityContentSidebar'
+
+type PoiPanelHeaderProps = {
+  goBack: (() => void) | null
+}
+
+const PoiPanelHeader = ({ goBack }: PoiPanelHeaderProps): ReactElement => {
+  const { t } = useTranslation('pois')
```
LeandraH, line 15: ❌ That should be 'common'

LeandraH (approved):
Nice! Tested in Firefox

lunars97:
should `share` button with its options also be there? :thinking:

steffenkleinle:
> should `share` button with its options also be there? 🤔

Probably yes. But since we probably need to switch from the sidebar to the menu from the design system, I would not put more work into this right now. I'll create a separate issue for this. Good catch!

lunars97:
Everything looks really good except for the color of back button on mobile device in dark mode when a user is in poi details
image
Could you also fix it?

Issue #3547, opened 2025-09-24
Add welcome message to landing page


steffenkleinle:
### Describe the Problem

The design adds a welcome message to the landing page.

### Describe Your Preferred Solution

Add the welcome title and text to the landing page and fix all visual problems introduced in #3535.

### Describe Possible Alternatives

N/A

### Related Issues

Related: #3023 

### Additional Information

https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=3046-11401&t=bVb8C7wU44Kl66H9-0

PR #3549, opened 2025-09-24
3547: Add welcome to landing page

steffenkleinle:
### Short Description

This PR adds a welcome message and title to the landing page and fixes some visual issues.

### Proposed Changes

- Add welcome texts to landing page
- Remove headroom from landing search box
- Adjust search box spacing and refactor

### Side Effects

- The search box on the landing page does does not reappear when scrolling up

### Testing

Check the landing page and the search page for both mobile and desktop. 

### Resolved Issues

Fixes: #3547

---


LeandraH (commented):
When I open the landing page, the search box is active, that's not good for a11y. It also makes this search bar kind of aggressively jump into my face, especially because the bar is really long. I think the search bar should be about as long as a long city name, more like in the design.

Image

steffenkleinle:
> Nice, thank you!
> 
> One more thing (sorry!), what do you think about having the placeholder in the search bar be something like "Augsburg" to make it clear that you're supposed to find a place here and not content information?

No worries! Personally I don't mind, could be a good idea. Could you perhaps check that with Toni and do this in a separate PR if necessary? 🙏

steffenkleinle:
> > Nice, thank you!
> > One more thing (sorry!), what do you think about having the placeholder in the search bar be something like "Augsburg" to make it clear that you're supposed to find a place here and not content information?
> 
> No worries! Personally I don't mind, could be a good idea. Could you perhaps check that with Toni and do this in a separate PR if necessary? 🙏

Maybe we should go with `Landkreis München` though if we go for this as there are definitely lots of users that don't know Augsburg and therefore dont know what it is/means.

LeandraH:
> > When I open the landing page, the search box is active, that's not good for a11y. It also makes this bright blue search bar kind of aggressively jump into my face, especially because the bar is really long. I think the search bar should be about as long as a long city name, more like in the design.
> 
> TBH I like the autofocus when I go to the locations page I just type instead of scrolling or using tab to reach it.

That makes sense but you also know what to expect when opening the page, our users, especially new ones, probably don't :)

Issue #3524, opened 2025-09-12
Fix TTS slow-down on unsupported browsers


bahaaTuffaha:
### Describe the Bug

when I click on Read aloud at any content page it takes about 5 seconds to show the `ttsHelperModal` on unsupported browsers like: the base Chromium web browser and Brave.

while it takes 5 seconds if you clicked on other page like news or map it will crash;
```
EasySpeech: not initialized. Run EasySpeech.init() first
```

### Steps to Reproduce

1. Use the base Chromium.
2. Go to https://integreat.app/muenchen/en/welcome-2/new-in-munich-first-steps for ex.
3. Click on Read aloud.
4. Click on map before the modal showing up.

### Expected Behavior

The error should be handled and the `ttsHelperModal` should show up immediately.
### Additional Information

https://github.com/user-attachments/assets/d647c2d4-0712-427d-8510-f34a875f4cf4

bahaaTuffaha:
🤦‍♂️ Ah.. that's why it was `EasySpeech.init({ maxTimeout: 500, interval: 250 })` before changing it to `maxTimeout:5000`

PR #3552, opened 2025-09-25
3524: Fix Easy speech not initialized state

bahaaTuffaha:
### Short Description

When I click on Read aloud at any content page it takes about 5 seconds and if you clicked on other page like news or map it will crash.

### Proposed Changes

- Created a state called `isInitializing` to show a backdrop with spinner while initializing.
- Lowered the `TTS_TIMEOUT` to 2 seconds.
- Changed  the color of BookIcon at the `TtsHelperModal` to the correct one and replaced legacy text color with palette one.

### Side Effects

- None.

### Testing
- For this test you need unsupported browser like brave or base chromium.
- Test Read aloud on any content page.
- Test it on small screens.

### Resolved Issues

Fixes: #3524

---

### DIFF
--- a/web/src/components/TtsContainer.tsx
+++ b/web/src/components/TtsContainer.tsx
@@ -1,14 +1,21 @@
+import Backdrop from '@mui/material/Backdrop'
+import { styled } from '@mui/material/styles'
 import EasySpeech from 'easy-speech'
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
Looks good! Unfortunately, TTS worked on all my installed browsers (even Chromium 🤷)

PR #3553, opened 2025-09-26
Fix dimensions

steffenkleinle:
### Short Description

The boolean dimensions where configured wrongly.

### Proposed Changes

- Set the correct dimensions
- Adjust text size of bottom navigation on xsmall screens

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/web/src/__mocks__/useDimensions.ts
+++ b/web/src/__mocks__/useDimensions.ts
@@ -25,8 +25,9 @@ export const mockDimensions: Dimensions = {
 
   mobile: true,
   desktop: false,
+  xsmall: true,
   small: false,
-  medium: true,
+  medium: false,
   large: false,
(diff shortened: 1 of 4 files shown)


hauf-toni:
I recommend implementing a breakpoint at which the layout switches from using icons accompanied by labels to icons only. on very small screens the combination of icons and text often leads to cramped layouts and potential issues. by switching to icons only we maintain clarity @steffenkleinle

bahaaTuffaha (approved):
Looks good ✅️

LeandraH (approved):
Looks fine to me, but I'm honestly a bit confused about why we have these additional functions instead of using [Mui media queries](https://mui.com/material-ui/react-use-media-query/)?

Issue #3560, opened 2025-09-29
Use camelCase in DatabaseConnector


LeandraH:
### Describe the Problem

Due to historical reasons, we have so far been using snake_case in the `DatabaseConnector`. We don't need to do that anymore, so we can use camelCase now.

### Describe Your Preferred Solution

Use camelCase for all the keys in the `DatabaseConnector`

### Describe Possible Alternatives

We could leave it.

### Related Issues

Discussed in https://github.com/digitalfabrik/integreat-app/pull/3541#discussion_r2387688077

PR #3561, opened 2025-09-29
3560: camelCase in DatabaseConnector

LeandraH:
### Short Description

Renamed the keys using snake_case to use camelCase

### Proposed Changes

- Renamed the keys using snake_case to use camelCase
- Increased cache_version

### Side Effects

- Should be none

### Testing

Check that the app still runs in native

### Resolved Issues

Fixes: #3560 

---


lunars97 (approved):
Thank you 💛

Issue #3543, opened 2025-09-23
Implement MUI Snackbar


lunars97:
### Describe the Problem

Use MUI Snackbar https://mui.com/material-ui/react-snackbar/ to notify/update a user for ex. `Alert` MUI component which is used in Feedback and in Malte Form (see the usage for `Alert` https://mui.com/material-ui/react-snackbar/#use-with-alerts)

### Describe Your Preferred Solution

Use `Snackbar` for `Alert` components
See whether it can be used somewhere else as well 🤔 

### Describe Possible Alternatives

N/A

### Related Issues

#3231 
See `Alert` implementation for Feedback and Malte Form #3409

PR #3563, opened 2025-09-30
3543: Implement mui snackbar

lunars97:
### Short Description

Mui `Snackbar` has been implemented for `Alert` component in `FeedbackToolbaritems`, `MalteHelpForm` and `FeedbackContainer`

### Proposed Changes

- Wrap `Alert` in Snackbar
- Created reusable Snackbar component
- Moved `SendingStatusType` to `shared/constants` 
- Feedback dialog closes if it was successful and then appears the successfull Snackbar, if there was an error then it does not close and the snackbar with error appears
- Adjusted tests

### Side Effects

- Users will not be navigated to additional `thank you` page. Instead snackbar will appear with the related message

### Testing

- Go to Feedback Form both in toolbar and in not found search result as well as MalteHelpForm (for that, run yarn start:malte, set the `api-url` in the local storage to `https://cms-test.malteapp.de`, and go to http://localhost:9000/testumgebung/de/kontaktformular)
- Send the form
- See the snackbar

### Resolved Issues

Fixes: #3543 

---


web/src/components/Sidebar.tsx
```diff
@@ -54,14 +54,16 @@ const Sidebar = ({ children, open, setOpen, Footer, OpenButton }: SidebarProps):
         </IconButton>
       )}
       <StyledDrawer
+        variant='temporary'
```
lunars97, line 57: added just for testing reasons, it will be fixed in #3513

web/src/components/Feedback.tsx
```diff
@@ -85,63 +87,72 @@ const Feedback = ({
...
+        <PrivacyFormControl error={showErrors && !privacyPolicyAccepted} required>
+          <PrivacyCheckbox language={language} checked={privacyPolicyAccepted} setChecked={setPrivacyPolicyAccepted} />
+          {showErrors && !privacyPolicyAccepted && <FormHelperText>{t('common:notePrivacyPolicy')}</FormHelperText>}
+        </PrivacyFormControl>
+        <Button onClick={handleSubmit} variant='contained' startIcon={<SendIcon />}>
+          {t('send')}
+        </Button>
+      </Container>
+      <Snackbar open={snackbarOpen} onClose={handleClose}>
+        <Alert onClose={handleClose} severity={sendingStatus === 'failed' ? 'error' : 'success'} variant='filled'>
```
lunars97, line 151: used `filled`variant to have enough contrast with the background

steffenkleinle:
> Do we expect to close the Feedback modal once we send the form and before Snackbar appears? 💭

Yes, I'd say it would be instead of the feedback thanks message.

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const StyledAlert = styled(Alert)`
+  display: flex;
+  align-items: center;
+
+  '& .muialert-action': {
+    margin: 0;
+    padding: 0;
+    align-self: center;
+  }
```
bahaaTuffaha, line 30: It passed the `yarn lint && yarn ts:check` without issues but the IDE says the following:
Image

🤔  Does it matter if it was small m and a at `.muialert-action`
Image

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Link } from 'react-router-dom'
+
+import { ExtendedSendingStatusType, SNACKBAR_AUTO_HIDE_DURATION } from 'shared'
+
+import useDimensions from '../../hooks/useDimensions'
+
+const StyledMuiSnackbar = styled(MUISnackbar)(({ theme }) => ({
+  [theme.breakpoints.down('lg')]: {
+    marginBottom: 50,
+  },
```
bahaaTuffaha, line 19: Is that expected or do we need to move it above the footer?
Image

web/src/components/MalteHelpForm.tsx
```diff
@@ -102,13 +100,11 @@ const MalteHelpForm = ({ pageTitle, languageCode, cityCode, malteHelpFormOffer }
...
-      <Alert severity='success' role='alert' action={<Link to={dashboardRoute}>{t('error:goTo.categories')}</Link>}>
-        {t('submitSuccessful')}
-      </Alert>
-    )
-  }
+  useEffect(() => {
+    if (sendingStatus === 'successful' || sendingStatus === 'failed' || sendingStatus === 'invalidEmail') {
+      setSnackbarOpen(true)
+    }
+  }, [sendingStatus])
```
bahaaTuffaha, line 107: 💭 Do we need to create a custom hook for this useEffect and the snackbarOpen state?

web/src/components/base/Snackbar.tsx
lunars97: it was `.MuiAlert-action`, dunno why it is making it small :thinking:

shared/constants/index.ts
```diff
@@ -13,5 +13,8 @@ export const INTERNAL_OFFERS = [SPRUNGBRETT_OFFER_ALIAS, MALTE_HELP_FORM_OFFER_A
 export const getChatName = (appName: string): string => `Frag${appName} (beta)`
 
 export type Rating = 'positive' | 'negative'
+
+export type SendingStatusType = 'idle' | 'sending' | 'failed' | 'successful'
+export type ExtendedSendingStatusType = SendingStatusType | 'invalidEmail'
```
lunars97, line 18: should be also adjusted in native, should I do this in this PR or better create another for native? 🤔

web/src/components/MalteHelpForm.tsx
lunars97: i dunno, so far I used it only in two components 🤔

web/src/components/Feedback.tsx
```diff
@@ -67,63 +67,68 @@ const Feedback = ({
...
+        </PrivacyFormControl>
+        <Button onClick={handleSubmit} variant='contained' startIcon={<SendIcon />}>
+          {t('send')}
+        </Button>
+      </Stack>
+      <Snackbar
+        open={snackbarOpen}
+        onClose={handleClose(setSnackbarOpen)}
+        sendingStatus={sendingStatus}
+        autoHideOnSuccess
```
lunars97, line 127: added auto hide only for success, because error message requires time to process it

web/src/components/Feedback.tsx
```diff
@@ -67,63 +67,68 @@ const Feedback = ({
...
-          {t('thanksMessage')}
-        </Alert>
-      </Stack>
-    )
-  }
+  useEffect(() => {
+    if (sendingStatus === 'successful' || sendingStatus === 'failed') {
+      setSnackbarOpen(true)
+    }
+  }, [sendingStatus])
```
steffenkleinle, line 74: ❌ Please avoid using a `useEffect` for this and instead call `setSnackbarOpen(true)` in `handleSubmit`/`onSubmit`.

web/src/components/MalteHelpForm.tsx
steffenkleinle: See comment above, please avoid using `useEffect` for this. Side effects like this make the code harder to read and understand since its not clear when what happens.

shared/constants/index.ts
steffenkleinle: I actually think that we should not add custom error messages as status type. If we actually want to show a snackbar for invalid emails, it should be set separately.
```suggestion
```

web/src/components/FeedbackToolbarItem.tsx
```diff
@@ -11,26 +11,44 @@ import { RouteType } from '../routes'
...
+  const [snackbarOpen, setSnackbarOpen] = useState(false)
   const { t } = useTranslation('feedback')
   const title = isSubmitted ? t('thanksHeadline') : t('headline')
+  const [snackbarStatus, setSnackbarStatus] = useState<SendingStatusType>('idle')
+
+  const handleFeedbackSuccess = () => {
+    setIsFeedbackOpen(false)
+    setSnackbarStatus('successful')
+    setIsSubmitted(true)
+    setSnackbarOpen(true)
```
steffenkleinle, line 36: ❓ These seem quite redundant, do we really need four different states here?

web/src/components/FeedbackToolbarItem.tsx
```diff
@@ -44,6 +62,13 @@ const FeedbackToolbarItem = ({ route, slug, rating }: FeedbackToolbarItemProps):
         text={t(rating === RATING_POSITIVE ? 'useful' : 'notUseful')}
         onClick={() => setIsFeedbackOpen(true)}
       />
+      <Snackbar
+        open={snackbarOpen}
+        onClose={handleClose(setSnackbarOpen)}
```
steffenkleinle, line 67: ❌ There is no need to pass `handleClose` to the snackbar. Instead, it should be enough to pass `setSnackbarOpen` and only define `handleClose` there:
```suggestion
        onClose={() => setSnackbarOpen(false)}
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  successMessage?: string
+  errorMessage?: string
+  errorTitle?: string
+  autoHideOnSuccess?: boolean
+}
+
+const Snackbar = ({
+  open,
+  onClose,
+  sendingStatus,
```
steffenkleinle, line 59: 🔧 This component should be kept generic (especially if put in `base`). Having a prop `sendingStatus` should therefore be avoided.

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  dashboardRoute,
+  autoHideOnSuccess = false,
+  successMessage,
+  errorMessage,
+  errorTitle,
```
steffenkleinle, line 64: ❌ What do you think about keeping the same logic as from the mui snackbar, i.e. passing a `message`, `title` and `severity` instead?

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  errorTitle?: string
+  autoHideOnSuccess?: boolean
+}
+
+const Snackbar = ({
+  open,
+  onClose,
+  sendingStatus,
+  dashboardRoute,
+  autoHideOnSuccess = false,
```
steffenkleinle, line 61: 🔧 If you use a severity below, I think you can omit this prop and set the correct hide duration directly.

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+import { ExtendedSendingStatusType, SNACKBAR_AUTO_HIDE_DURATION } from 'shared'
+
+const StyledMuiSnackbar = styled(MUISnackbar)`
+  position: fixed;
+  bottom: ${props =>
+    props.theme.dimensions.bottomNavigationHeight ?? props.theme.dimensions.visibleFooterHeight}px !important;
+  inset-inline-end: 16px;
+  margin-bottom: 16px;
+`
```
steffenkleinle, line 20: ❌ Sorry for not spotting this earlier, but please use the positioning provided by the component:
https://mui.com/material-ui/react-snackbar/#position

If necessary, it should be enough to add a custom bottom margin to put it above the bottom navigation bar.

web/src/components/MalteHelpForm.tsx
```diff
@@ -102,13 +100,11 @@ const MalteHelpForm = ({ pageTitle, languageCode, cityCode, malteHelpFormOffer }
...
 
-  if (sendingStatus === 'successful') {
-    return (
-      <Alert severity='success' role='alert' action={<Link to={dashboardRoute}>{t('error:goTo.categories')}</Link>}>
-        {t('submitSuccessful')}
-      </Alert>
-    )
-  }
+  useEffect(() => {
+    if (sendingStatus === 'successful' || sendingStatus === 'failed' || sendingStatus === 'invalidEmail') {
```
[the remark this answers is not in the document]
steffenkleinle: I think that would be the better choice imo 👍

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+) => void
+
+export const handleClose =
+  (setOpen: (open: boolean) => void): SnackbarCloseHandler =>
+  (event?, reason?) => {
+    if (reason === 'clickaway') {
+      return
+    }
+    setOpen(false)
+  }
```
steffenkleinle, line 44: 🔧 
```suggestion
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <Button component={Link} to={dashboardRoute} size='small'>
+              {t('error:goTo.categories')}
+            </Button>
+          ) : severity === 'success' ? (
+            <IconButton aria-label={t('close')} color='inherit' size='small' onClick={onClose}>
+              <CloseIcon fontSize='inherit' />
+            </IconButton>
+          ) : null
+        }>
+        {severity === 'error' && !!title && <AlertTitle>{title}</AlertTitle>}
```
steffenkleinle, line 81: 🔧 This is not very generic. If we ever want to show a non-error snackbar with title, this needs to be adjusted. Better only pass the title if you want to show one.
```suggestion
        {severity === 'error' && !!title && <AlertTitle>{title}</AlertTitle>}
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+          // eslint-disable-next-line no-nested-ternary
+          severity === 'success' && dashboardRoute ? (
+            <Button component={Link} to={dashboardRoute} size='small'>
+              {t('error:goTo.categories')}
+            </Button>
+          ) : severity === 'success' ? (
+            <IconButton aria-label={t('close')} color='inherit' size='small' onClick={onClose}>
+              <CloseIcon fontSize='inherit' />
+            </IconButton>
+          ) : null
```
steffenkleinle, line 79: ❓ Actually, what do you think of not implementing these here but instead just accepting a prop `action`? I think this would avoid quite a few of these conditionals. So something like this:

``` tsx
// Snackbar.tsx
const Snackbar = ({ action, ... }: SnackbarProps => (
  <MuiSnackbar ...>
    <Alert action={action} ... />
  </MuiSnackbar>
)

// Feedback.tsx
...
return (
  <>
    <Snackbar action={<IconButton ... />} />
  </>
)
```

~~🔧  Please avoid nested ternaries and disabling linting rules.
You can do this e.g. like this:~~

```suggestion
          const SuccessAction = ashboardRoute ? (
            <Button component={Link} to={dashboardRoute} size='small'>
              {t('error:goTo.categories')}
            </Button>
          ) : severity === 'success' ? (
            <IconButton aria-label={t('close')} color='inherit' size='small' onClick={onClose}>
              <CloseIcon fontSize='inherit' />
            </IconButton>
          )

          severity === 'success' ? SuccessAction : null
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return
+    }
+    setOpen(false)
+  }
+
+export type SnackbarProps = {
+  open: boolean
+  onClose: SnackbarCloseHandler
+  dashboardRoute?: string
+  severity: 'success' | 'error' | 'info' | 'warning'
```
steffenkleinle, line 50: 🔧 
```suggestion
  severity: AlertColor
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+}))
+
+const StyledAlert = styled(Alert)`
+  display: flex;
+  align-items: center;
+
+  .${alertClasses.action} {
+    padding: 0 8px;
+  }
```
steffenkleinle, line 29: 🤔 Is this necessary?

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { SNACKBAR_AUTO_HIDE_DURATION } from 'shared'
+
+const StyledMuiSnackbar = styled(MUISnackbar)(({ theme }) => ({
+  [theme.breakpoints.down('md')]: {
+    marginBottom: 40,
+  },
+  [theme.breakpoints.up('md')]: {
+    marginBottom: 46,
+  },
+}))
```
steffenkleinle, line 21: 🔧 
```suggestion
const StyledMuiSnackbar = styled(MUISnackbar)(({ theme }) => ({
﻿﻿﻿  marginBottom: theme.dimensions.bottomNavigationHeight,
}))
```

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const Snackbar = ({ open, onClose, severity, message, title, dashboardRoute }: SnackbarProps): ReactElement => {
+  const { t } = useTranslation('common')
+
+  return (
+    <StyledMuiSnackbar
+      open={open}
+      onClose={onClose}
+      anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
```
steffenkleinle, line 62: 🙃 Personally I think it looks better centered, but I am fine with left aligned as well.

web/src/components/base/Snackbar.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <StyledAlert
+        severity={severity}
+        role='alert'
+        onClose={onClose}
+        variant={severity === 'error' ? 'filled' : 'standard'}
+        action={
+          // eslint-disable-next-line no-nested-ternary
+          severity === 'success' && dashboardRoute ? (
+            <Button component={Link} to={dashboardRoute} size='small'>
+              {t('error:goTo.categories')}
```
steffenkleinle, line 73: 🔧 How about just `back`?
```suggestion
              {t('back')}
```

web/src/components/MalteHelpForm.tsx
```diff
@@ -188,17 +183,25 @@ const MalteHelpForm = ({ pageTitle, languageCode, cityCode, malteHelpFormOffer }
...
           {submitted && !privacyPolicyAccepted && <FormHelperText>{t('common:notePrivacyPolicy')}</FormHelperText>}
         </FormControl>
-        {(sendingStatus === 'failed' || sendingStatus === 'invalidEmail') && (
+        {invalidEmail && (
           <Alert severity='error' role='alert'>
             <AlertTitle>{t('submitFailed')}</AlertTitle>
-            {t(sendingStatus === 'invalidEmail' ? 'invalidEmailAddress' : 'submitFailedReasoning')}
+            {t('invalidEmailAddress')}
           </Alert>
         )}
```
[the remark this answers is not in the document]
lunars97: good idea :+1: Here is a good illustration for it https://m3.material.io/components/text-fields/accessibility#20694376-8235-4b06-9d0d-94f4e504a612

web/src/components/base/Snackbar.tsx
lunars97: yes, it is to align the button with the alert text

steffenkleinle: I checked and it seems correctly aligned to me. In any case, alignment should be fixed using `alignItems` and/or `justifyContent` styles. Using a padding for alignment breaks if the padding is changed on only one component.

lunars97: There was slight misalignment. I will have a look again :+1:

PR #3565, opened 2025-09-30
Fix DatePicker test to use fixed date values

bahaaTuffaha:
### Short Description

`yarn test` is failing at
```bash
 FAIL  src/components/__tests__/DatePicker.spec.tsx
  ● DatePicker › handles date change correctly

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

    - "2025-10-01T12:44:26.986+02:00"
    + "2025-09-01T12:44:26.986+02:00",
      {"validationError": null},

    Number of calls: 1
```
- Today is the 30th of September when it adds 1 day plus it will expect 1st of October not September this is clearly means that it selects the day from the same month `fireEvent.click(getByText(newValue.day))`.

### Proposed Changes

- Just used fixed dates instead of `DateTime.now()`

### Side Effects

- None.

### Testing

- `yarn test` at web.

### Resolved Issues

Fixes: #

---

### DIFF
--- a/web/src/components/__tests__/DatePicker.spec.tsx
+++ b/web/src/components/__tests__/DatePicker.spec.tsx
@@ -52,9 +52,9 @@ describe('DatePicker', () => {
   })
 
   it('handles date change correctly', () => {
-    const currentDate = DateTime.local()
-    const newValue = DateTime.now().plus({ days: 1 }).setLocale('de')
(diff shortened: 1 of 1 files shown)


steffenkleinle (approved):
Thank you!

PR #3571, opened 2025-10-01
Remove custom shadow

steffenkleinle:
### Short Description

Our custom shadows for mui lead to problems in some cases, e.g. for dialogs. Therefore I'd just use the mui defaults.

### Proposed Changes

- Use mui defaults for shadow
- Increase elevation of `Header`

### Side Effects

N/A

### Testing

Check out header, footer and dialogs.

### Resolved Issues

Fixes: N/A

---


bahaaTuffaha (commented):
I think this will affect the inverse shadows for example (top image is this PR):

Image

Image

steffenkleinle:
> I think this will affect the inverse shadows for example (left image is this PR):

Hmm yes. Personally I actually think the left looks better. What do you think?

bahaaTuffaha:
> > I think this will affect the inverse shadows for example (left image is this PR):
> 
> Hmm yes. Personally I actually think the left looks better. What do you think?

💭 It separate the button from the background because both have dark colors.

bahaaTuffaha:
Can we just make the shadow white for contrast mode and delete this custom shadow?

PR #3573, opened 2025-10-02
[Proposal] 3376: Refactor breadcrumbs

steffenkleinle:
### Short Description

Proposal PR to refactor the breadcrumbs to get rid of loads of custom logic and styling.

### Proposed Changes

- Simplify breadcrumbs logic
- Remove most of the custom breadcrumb logic
- Only show one breadcrumb on small devices
- Remove highlighting of current breadcrumb

### Side Effects

N/A

### Testing

Checkout the breadcrumbs on different pages/screen sizes.

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/web/src/components/Breadcrumbs.tsx
+++ b/web/src/components/Breadcrumbs.tsx
@@ -1,201 +1,94 @@
 import HomeOutlinedIcon from '@mui/icons-material/HomeOutlined'
 import MoreHorizIcon from '@mui/icons-material/MoreHoriz'
-import Box from '@mui/material/Box'
-import MuiBreadcrumbs from '@mui/material/Breadcrumbs'
-import IconButton from '@mui/material/IconButton'
+import MuiBreadcrumbs, { breadcrumbsClasses } from '@mui/material/Breadcrumbs'
+import IconButton, { iconButtonClasses } from '@mui/material/IconButton'
 import Menu from '@mui/material/Menu'
 import MenuItem from '@mui/material/MenuItem'
+import Stack from '@mui/material/Stack'
 import { styled } from '@mui/material/styles'
 import React, { ReactElement, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 
+import useDimensions from '../hooks/useDimensions'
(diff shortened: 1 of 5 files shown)


web/src/components/Breadcrumbs.tsx
```diff
@@ -1,201 +1,90 @@
...
-  const rest = allBreadcrumbs.slice(1)
+  const openMenu = (event: React.MouseEvent<HTMLButtonElement>) => setMenuAnchorElement(event.currentTarget)
+  const closeMenu = () => setMenuAnchorElement(null)
 
-  const ellipsis = new BreadcrumbModel({
-    title: '...',
-    pathname: '',
-    node: null,
-  })
+  const maxVisible = small ? 1 : 2
```
steffenkleinle, line 55: What do you think about this? Should we only display one item on xsmall viewports (<600px)? Or do you think we should keep the current logic?

When merging main this has to be changed to:
```suggestion
  const maxVisible = xsmall ? 1 : 2
```

bahaaTuffaha: From what we gathered from Toni we can do one or two so keep it with one item as you did (less ellipsis used).

image

web/src/components/Breadcrumbs.tsx
```diff
@@ -1,201 +1,90 @@
...
+  const [home, ...rest] = breadcrumbs
+  const hiddenBreadcrumbs = rest.slice(0, -maxVisible)
+  const visibleBreadcrumbs = rest.slice(-maxVisible)
 
-  return {
-    breadcrumbs: [home, ellipsis, ...rest.slice(lastTwoCrumbs)],
-    collapsedBreadcrumbs: rest.slice(0, lastTwoCrumbs),
+  if (breadcrumbs.length < 2 || !home) {
+    return <JsonLdBreadcrumbs breadcrumbs={breadcrumbs} />
```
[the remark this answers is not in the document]
steffenkleinle: Exactly, we previously also added breadcrumbs for SEO but only showed the breadcrumbs to users on child pages.

Issue #3562, opened 2025-09-30
Use mui list for language selector items


steffenkleinle:
### Describe the Problem

The language selector items currently don't use mui items and therefore are also not using html semantic list elements.

### Describe Your Preferred Solution

Use mui components for the language list in the language selector.
DO NOT change the overall looks nor add a search bar, this will be done in #3349. Only replace the currently used styled components with mui `ListItem`/`ListItemButton`.

### Describe Possible Alternatives

N/A

### Related Issues

#3349

### Additional Information

N/A

PR #3575, opened 2025-10-03
3562: Mui for language selector

bahaaTuffaha:
### Short Description

The language selector items currently don't use mui items and therefore are also not using html semantic list elements.

### Proposed Changes

- Used ListItem, ListItemButton and ListItemText for the language selector.
- Removed unnecessary styling like `BoldSpacer`.

### Side Effects

- None.

### Testing

- Test language selector at http://localhost:9000/landing/de.
- Test disabled text languages after reloading the page.
- Test on small screens.

### Resolved Issues

Fixes: #3562

---

### DIFF
--- a/web/src/components/PoiChips.tsx
+++ b/web/src/components/PoiChips.tsx
@@ -15,7 +15,7 @@ const PoiChips = ({ poi }: { poi: PoiModel }): ReactElement => {
   const { category } = poi
 
   return (
-    <Stack flexDirection='row' gap={1}>
+    <Stack flexDirection='row' flexWrap='wrap' gap={1}>
       <Chip icon={<Icon src={category.icon} />} label={category.name} variant='outlined' />
       {poi.organization !== null && (
         <Chip component={Link} to={poi.organization.url} label={poi.organization.name} variant='outlined' clickable />


steffenkleinle (changes requested):
I really like the new look! However, could you perhaps still remove lots of the custom styling such as legacy colors, paddings, etc please? :)

steffenkleinle (changes requested):
I really like the new look! However, could you perhaps still remove lots of the custom styling such as legacy colors, paddings, etc please? :)

web/src/components/Selector.tsx
```diff
@@ -1,62 +1,41 @@
...
 
-  ${theme.breakpoints.down('md')} {
-    width: 100%;
-    flex: 1 1 auto;
-    font-size: 1em;
+  ${props => props.theme.breakpoints.down('lg')} {
+    padding: 0 16px;
+    min-width: 96px;
+    height: 40px;
   }
```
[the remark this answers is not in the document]
bahaaTuffaha: > ❓ Are the custom paddings/margins/min-widths actually necessary?

To make sure it's above the CityContentSidebar.

> Why do we need to set a custom color for disabled buttons?

it was like that so I just passed a prop instead to minimize code as possible.

web/src/components/Selector.tsx
```diff
@@ -1,62 +1,33 @@
...
-  min-width: 90px;
-  padding: 0 5px;
-  font-size: 1.2em;
-  line-height: 64px;
-  text-align: center;
-  white-space: nowrap;
+const SelectorItemButton = styled(ListItemButton)<{ disabled?: boolean }>`
+  height: 48px;
+  min-width: 112px;
+  padding: 0 20px;
```
steffenkleinle, line 15: 🔧 
```suggestion
```

web/src/components/Selector.tsx
```diff
@@ -1,62 +1,33 @@
...
-
-const BoldSpacer = styled('div')`
-  font-weight: 700;
-  height: 0;
-  overflow: hidden;
-  visibility: hidden;
-`
+  display: flex;
+  justify-content: center;
+  margin: 4px;
```
steffenkleinle, line 19: 🙃 I think using a `gap` and `padding` on the parent (i.e. the list) would be preferable.
```suggestion
  margin: 4px;
```

web/src/components/Selector.tsx
```diff
@@ -1,62 +1,33 @@
...
+  ${props => props.disabled && `color: ${props.theme.palette.text.disabled};`}
```
steffenkleinle, line 20: 🔧 This is already handled by mui, no need to set a custom text color.
```suggestion
  ${props => props.disabled && `color: ${props.theme.palette.text.disabled};`}
```

web/src/components/Header.tsx
```diff
@@ -72,9 +74,14 @@ export const Header = ({
...
   const height = headerRect?.height ?? 0
   const landingPath = pathnameFromRouteInformation({ route: LANDING_ROUTE, languageCode: language })
+  const { xsmall, small } = useDimensions()
 
   return (
-    <Headroom scrollHeight={HEADER_HEIGHT} height={height} zIndex={2} onStickyTopChanged={onStickyTopChanged}>
+    <Headroom
+      scrollHeight={HEADER_HEIGHT}
+      height={height}
+      zIndex={xsmall || small ? 2 : Z_INDEX_ABOVE_SIDEBAR} // 2 for mobile bottom sheet
```
steffenkleinle, line 83: 🔧 
```suggestion
      // Position header above toolbar but below bottom sheet
      zIndex={mobile ? 2 : Z_INDEX_ABOVE_SIDEBAR}
```

❓ Actually, could you check if the z-index for the toolbar is still necessary? I actually don't think it still is since it was added for the sharing popup which will was removed in #3578.
Then you can just revert this:
``` suggestion
      zIndex={2}
```

bahaaTuffaha: Ok 🙂

web/src/components/Selector.tsx
```diff
@@ -1,62 +1,33 @@
...
 
-const selectorItemStyle = ({ theme }: { theme: Theme }) => css`
-  height: 64px;
-  min-width: 90px;
-  padding: 0 5px;
-  font-size: 1.2em;
-  line-height: 64px;
-  text-align: center;
-  white-space: nowrap;
+const SelectorItemButton = styled(ListItemButton)<{ disabled?: boolean }>`
```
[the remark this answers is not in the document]
bahaaTuffaha: :face_with_diagonal_mouth:  I remember there was something complaining about it :disappointed:

Issue #3574, opened 2025-10-02
Remove hardcoded locale in PoiModel


LeandraH:
### Describe the Problem

We use the hardcoded locale 'de' in the `isCurrentlyOpen` method in the `PoiModel` (https://github.com/digitalfabrik/integreat-app/blob/main/shared/api/models/PoiModel.ts#L128). We should instead use the locale of the user.

### Describe Your Preferred Solution

Use the locale of the user.

### Describe Possible Alternatives

Nope.

### Additional Information

There is a chance that at some point, we will have versions of the app where the users aren't all in Germany.

PR #3580, opened 2025-10-07
3574: Remove hardcoded locale from poi model

lunars97:
### Short Description

This PR removes hardcoded locale from poi model. It should be based on the time zone of a user

### Proposed Changes

- Remove hardcoded locale
- Preserve timezone by directly setting hours and minutes on the same day as `now`

### Side Effects

- none

### Testing

Go to Poi details in http://localhost:9000/testumgebung/de/locations and see opening hours (depending on your timezone (check with different timezone), it should show whether it is closed or open for you now) Check both native and web

### Resolved Issues

Fixes: #3574 

---

### DIFF
--- a/shared/api/models/PoiModel.ts
+++ b/shared/api/models/PoiModel.ts
@@ -128,12 +128,8 @@ class PoiModel extends ExtendedPageModel {
       }
 
       return currentDay.timeSlots.some(timeslot => {
-        // converts time string into numeric hours and minutes
-        const [startHour, startMinute] = timeslot.start.split(':').map(Number)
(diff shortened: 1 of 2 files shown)


LeandraH (commented):
Unfortunately, this breaks the display in native for whether a POI is currently open. It looks like we still need to pass in some locale. I see two possibilities: 

- We could pass in the one from the only date we have in the `PoiModel`, `lastUpdate`, and thus make it the CMS's problem if we ever have customers who aren't based in Germany. That feels a bit hacky though.
- We could pass in a locale manually when we call `isCurrentlyOpen`.

What other possibilities do you see?

Image

Issue #3581, opened 2025-10-07
Fix word-break issue in small header city names


bahaaTuffaha:
### Describe the Bug

[⚠️  this is part of `3231-design-system-web`]

On small screens the header cityName / change location button gets word-break even if there is a space.

### Steps to Reproduce

1. Go to http://localhost:9000/albdonaukreis/de
2. Switch to mobile view.

### Expected Behavior

let the name takes as much space as possible.

### Related Issues

None.

### Additional Information

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/cda32157-5d49-4468-9c10-fadcfd3d354f" />

PR #3582, opened 2025-10-07
3581: Break the words in the header, not the header itself

LeandraH:
### Short Description

The header was behaving a little weird on small screens, e.g. breaking "Alb-Donau-Kreis" into three lines when there was plenty of space to not do that. We also forced the header to break if the municipality name is too long, instead of just breaking the municipality name. That should all be fixed now.

### Proposed Changes

- The municipality name in the header breaks if it's too long to fit the entire header on one line
- The municipality name in the header doesn't break if it can fit on one line

### Side Effects

- The header doesn't break into two lines anymore, which means that at widths below ca. 300px, the icon to open the sidebar slowly starts to disappear. There shouldn't be any devices that thin though.

### Testing

Test a couple of headers on a small screen, e.g. http://localhost:9000/albdonaukreis/de and http://localhost:9000/potsdam/de

### Resolved Issues

Fixes: #3581

---

### DIFF
--- a/web/src/components/Header.tsx
+++ b/web/src/components/Header.tsx
@@ -31,7 +31,6 @@ const Row = styled('div')`
   align-items: center;
   min-height: ${HEADER_HEIGHT}px;
   justify-content: space-between;
-  flex-wrap: wrap;
   overflow-x: auto;
   padding: 0 16px;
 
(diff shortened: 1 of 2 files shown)


LeandraH:
> Sadly this leads to a scroll bar/overflow on really small screen sizes/long city names: image

Yes, I mentioned that in the side effects :) I only saw it at widths around 300px and smaller, which is smaller than the smallest somewhat common width of 360px.

steffenkleinle (approved):
Tested on firefox

lunars97 (approved):
Thank you! Looks better now 👍

Issue #3510, opened 2025-09-08
Use mui buttons for leftover buttons


steffenkleinle:
### Describe the Problem

There are still a few cases where we use custom html `button`:
- AppointmentOnlyIcon
- MobileBanner
- base/Button
 
### Describe Your Preferred Solution

We should use mui (Icon)Button everywhere.
Delete `base/Button`.

### Describe Possible Alternatives

N/A

### Related Issues

Blocked by: #3311 

### Additional Information

N/A

PR #3583, opened 2025-10-07
3510: Replace all buttons with mui buttons

steffenkleinle:
### Short Description

This PR replaces the last usages of our legacy button component with mui components. 

### Proposed Changes

- Use mui components for events date filter
- Use mui components for map attribution (incl. position/style change of the attribution)
- Use mui components for mobile banner
- Remove unused installed logic in mobile banner

### Side Effects

N/A

### Testing

Check out the events date filter, map attribution (mobile + desktop) and mobile banner.

### Resolved Issues

Fixes: #3510

---


web/src/components/EventsDateFilter.tsx
```diff
@@ -66,17 +29,24 @@ const EventsDateFilter = ({
...
   const today = DateTime.now()
   const inAWeek = DateTime.now().plus({ week: 1 })
+  const formattedStartDate = startDate?.toFormat('dd.MM.yyyy') ?? '∞'
+  const formattedEndDate = endDate?.toFormat('dd.MM.yyyy') ?? '∞'
 
   return (
     <>
-      <StyledAccordion disableGutters expanded={showDateFilter} elevation={0}>
-        <FilterToggle isDateFilterActive={showDateFilter} setToggleDateFilter={setShowDateFilter} />
+      <Accordion disableGutters expanded={showDateFilter} elevation={0}>
```
lunars97, line 42: by default Accordion uses `h3` element for heading, it should be overriden with `slotProps` to use another component

web/src/components/EventsDateFilter.tsx
```diff
@@ -66,17 +29,24 @@ const EventsDateFilter = ({
...
+        <AccordionSummary onClick={() => setShowDateFilter(!showDateFilter)} disableRipple>
```
lunars97, line 43: there is a recommendation from WAI to set those attributes
```suggestion
        <AccordionSummary onClick={() => setShowDateFilter(!showDateFilter)} disableRipple id="panel-header" aria-controls="panel-content">
```

lunars97 (changes requested):
Thank you! Almost looks good :) Just small remarks are left

web/src/components/EventsDateFilter.tsx
steffenkleinle: What should be used instead?

lunars97: I was not sure about it in our case, but I do not have any other opinion on it :+1:

Issue #3617, opened 2025-10-15
Fix background for CityListGroup letter at the landing page


bahaaTuffaha:
### Describe the Bug

⚠️ This is part of `3231-design-system-web`.

- There is no background for the letter for `CityListGroup` at the landing page.

### Steps to Reproduce

1. Go to http://localhost:9000/landing/en
2. Scroll to each alphabetical section.

### Expected Behavior

To have a background to prevent text overlapping.

### Related Issues

None.

### Additional Information

<img width="1328" height="221" alt="Image" src="https://github.com/user-attachments/assets/cc579dbc-70b0-470a-aebe-408d55a492ac" />

Issue #3496, opened 2025-08-27
Clean up unused code


LeandraH:
### Describe the Problem

We are replacing a lot of our own components while implementing the new design system, but we can't always immediately remove them. We'll try to keep a list of components that still need to be replaced here, but we'll almost certainly forget some and need to do a manual check for unused code.

### Describe Your Preferred Solution

Remove unused components:

- [x] List and ListItem

### Describe Possible Alternatives

Nope.

### Related Issues

Part of #3231

lunars97:
old legacy List and ListItem were removed here #3538

PR #3618, opened 2025-10-16
3617: Fix subheader background

steffenkleinle:
### Short Description

Remove transparent background from mui subheaders and accordions.

### Proposed Changes

Before #3496, the background color of the content was still set to a legacy color causing mui subheaders and accordions to have a slightly different color. I therefore set background to be transparent, however, that caused the problem described in the issue.
Since the colors are now correct, the transparency can just be removed.

### Side Effects

N/A

### Testing

See issue.

### Resolved Issues

Fixes: #3617

---


steffenkleinle:
> I think the only thing that could be effected by this is the poi filter overlay buttons. the hovering on these buttons now is nontransparent which I find better... what do you think?

That is not connected to this PR, see current future. They are not transparent anymore because of a fix in #3496 by specifying the hover background color. But yes, I think its better now.

Issue #3349, opened 2025-06-24
Improve language selection


svenseeberg:
### Describe the Problem

The language selection is a plain list of languages. This was and sort of is sufficient for most regions with usually less than 10 languages. For example Munich now has over 20 languages and the list is becoming problematic from an UX perspective as it is too long and not sortable or searchable.

### Describe Your Preferred Solution

I think there are (at least) 3 things that can be done to improve the list:
- Scrollable list in desktop view
- Search field
- Optionally always add the English (or German) language name (small below the original language name) and support this in the search field

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

![Image](https://github.com/user-attachments/assets/3b40203f-97ad-42f0-8515-98eac8e15ede)

bahaaTuffaha:
This is just a suggestion... what do you think @hauf-toni  ?
and we can replace the top gray text with search bar when you click on the button it will get focused.

![Image](https://github.com/user-attachments/assets/6474079f-a4b0-434e-b913-0b9795da24e8)

steffenkleinle:
Can be done after the design system.

LeandraH:
We could probably use the Mui Autocomplete component: https://mui.com/material-ui/react-autocomplete/#combo-box

PR #3622, opened 2025-10-16
3349: Improve language selection

LeandraH:
### Short Description

I rewrote the language selection to use Mui Autocomplete, in order to make it easier to find languages when there are a lot of them.

### Proposed Changes

- Used Mui Autocomplete for the language selection in desktop
- The user can search for a language in the language's own name, the current language, or the fallback language, i.e. German
- Made the language selection in mobile functionally similar (it didn't look good with the Autocomplete)
- Removed the language selection from the `CityContentHeader` if there are no languages (which shouldn't be the case)
- Removed the `Dropdown` which wasn't used anywhere else
- Removed the `SelectorItemModel` which wasn't used anywhere else

### Side Effects

- Hopefully none

### Testing

Change the language on lots of pages

### Resolved Issues

Fixes: #3349

---


web/src/routes/__tests__/PoisPage.spec.tsx
```diff
@@ -72,11 +74,10 @@ describe('PoisPage', () => {
...
     mockUseLoadFromEndpointWithData(pois)
-    const { getAllByText, getByRole } = renderPois()
-    fireEvent.click(getByRole('link', { name: poi0.title }))
+    const { getAllByText, getByRole } = renderPois('/locations/test')
+    fireEvent.click(getByRole('button', { name: 'layout:changeLanguage' }))
+
     expect(getAllByText('English')[0]?.closest('a')).toHaveAttribute('href', poi0.availableLanguages.en)
     expect(getAllByText('Deutsch')[0]?.closest('a')).toHaveAttribute('href', poi0.availableLanguages.de)
-    // Pathname is not correctly updated, therefore the pathname does not include the slug
-    expect(getAllByText('اَللُّغَةُ اَلْعَرَبِيَّة')[1]?.closest('a')).toHaveAttribute('href', '/augsburg/ar/locations')
```
LeandraH: Does anybody know why we're testing for the Arabic translation when there isn't one in the PoiModelBuilder?

web/src/components/LanguageSelectorItem.tsx
LeandraH: I don't really like the naming here because we have a `HeaderLanguageSelectorItem`, which calls a `LanguageSelector`, which calls the `LanguageSelectorItem`. It feels somewhat circular to me. Does anybody have a better name?

web/src/routes/__tests__/PoisPage.spec.tsx
LeandraH: I would say that this is a poorly written test, that half-tests something that doesn't work (I also don't really understand the comment, how is the pathname not correctly updated? Is that because there is no Arabic translation in the poi?) If we want to test for Arabic, then I think we should add Arabic test data to the test

LeandraH:
> 🤔 check this out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames#language_display_names So I think we can pass available language codes and retrun all possible language names in different languages like arabic or العربية

Thank you, the Intl object is really very cool! Unfortunately, it's not fully implemented in react-native, and the DisplayNames are one of the things that aren't 😢

LeandraH:
I played around a bit with the [polyfills from FormatJS](https://formatjs.github.io/docs/polyfills/intl-displaynames/) and got a first draft running. Since getting the actual size of the app when downloading from the app store requires uploading it to said app store (there's lots of additional caching also between different apps), these absolute numbers are not reliable, but I assume that the relative change in size is somewhat accurate, at least on iOS. For a release version on iOS, it went up from 59.1 MB to 59.5 MB. For a debug version on Android, the size stayed at 234.3 MB, I'll check if the polyfill is necessary there at all (when I'm back from vacation in a week).

Either way, I would say that it's worth adding the polyfill in. Other opinions?

LeandraH:
> This polyfill only adds the language names though, correct? So all the other intl stuff that does not yet work in react-native does not get fixed by this?

The polyfill also needs Intl.Locale ([polyfill](https://formatjs.github.io/docs/polyfills/intl-locale/) and [real deal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale)) but it's not enough to replace the other stuff. It's a good idea to check on polyfills for those, I'll create a ticket for it.

web/src/components/LanguageList.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { config } from 'translations'
+
+import useDimensions from '../hooks/useDimensions'
+import LanguageListItem from './LanguageListItem'
+import SearchInput from './SearchInput'
+
+const MobileContainer = styled('div')`
+  display: flex;
+  flex-direction: column;
+  height: 100%;
```
f1sh1918: 🔧 i don't think height and width are needed here? I didn't find any issue without it

web/src/components/LanguageList.tsx
```diff
@@ -0,0 +1,160 @@
...
+    return languageChangePaths.filter(item =>
+      filterLanguageChangePath(item, query, languageNamesInCurrentLanguage, languageNamesInFallbackLanguage),
+    )
+  }, [languageChangePaths, query, languageCode])
+
+  if (mobile) {
+    return (
+      <MobileContainer>
+        <SearchInput placeholderText={currentLanguage?.name ?? ''} filterText={query} onFilterTextChange={setQuery} />
+
```
f1sh1918: 🙃 unneeded empty line

web/src/components/LanguageList.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const filteredLanguageChangePaths = useMemo(() => {
+    const languageNamesInCurrentLanguage = new Intl.DisplayNames([languageCode], { type: 'language' })
+    const languageNamesInFallbackLanguage = new Intl.DisplayNames([config.sourceLanguage], { type: 'language' })
+    return languageChangePaths.filter(item =>
+      filterLanguageChangePath(item, query, languageNamesInCurrentLanguage, languageNamesInFallbackLanguage),
+    )
+  }, [languageChangePaths, query, languageCode])
+
+  if (mobile) {
```
f1sh1918: ❓ is there any reason why we don't use the auto complete component also for mobile?

f1sh1918 (commented):
Works fine 👍 
Just have some general question to the auto complete.
It should also be possible to use that mobile with some styling adjustments

PR #3623, opened 2025-10-16
Fix breadcrumbs direction for RTL

bahaaTuffaha:
### Short Description

This is a quick fix for the direction of the breadcrumbs for rtl languages.

### Proposed Changes

- Added direction propriety to MuiBreadcrumbs at the `ThemeContainer.tsx`

### Side Effects

- None.

### Testing

- Check breadcrumbs at any category content.
- Test rtl and ltr languages.

### Resolved Issues

Fixes: #

---

### DIFF
--- a/web/src/components/ThemeContainer.tsx
+++ b/web/src/components/ThemeContainer.tsx
@@ -96,7 +96,7 @@ const createTheme = (
         },
         MuiBreadcrumbs: {
           defaultProps: {
-            dir: theme.palette.mode,
+            dir: contentDirection,
           },
         },
         MuiTooltip: {


web/src/components/ThemeContainer.tsx
```diff
@@ -96,6 +96,11 @@ const createTheme = (
             },
           },
         },
+        MuiBreadcrumbs: {
+          defaultProps: {
+            dir: theme.palette.mode,
+          },
+        },
```
steffenkleinle: ❓ `theme.palette.mode` has possible values of `light` and `dark`, that seems wrong 🙈 Perhaps this fixes the issue by just supplying any (even invalid) values to overwrite the default. Please still use an actual valid value though (e.g. `theme.contentDirection`).

steffenkleinle (changes requested):

web/src/components/ThemeContainer.tsx
steffenkleinle: I guess setting it to an invalid value removes the default 'auto' and instead falls back to the direction set somewhere else (being ltr/rtl).

lunars97 (approved):
Thank you! Tested in Chrome, looks good 👍

Issue #3626, opened 2025-10-17
Simplify and fix resource cache


steffenkleinle:
### Describe the Problem

We currently use a very nested and complicated resource cache structure to save images and pdfs offline.
The nesting/grouping resources by which page they occur at is not necessary and adds loads of additional complexity.

We currently also have a problem that the resource cache is not correctly applied on first load of the city/language content which will lead to problems with #3550.

### Describe Your Preferred Solution

- Simplify the resource cache structure, see below
- Fix the resource cache not being applied correctly on first start

### Describe Possible Alternatives

N/A

### Related Issues

Blocks: #3522

### Additional Information

New resource cache structure:
``` json
{
  "de": {
    "https://cms.integreat-app.de/my-pdf": {
      "filePath": "/path/to/resource/cache/my-pdf-hash"
    },
    "https://cms.integreat-app.de/my-image": {
      "filePath": "/path/to/resource/cache/my-image-hash"
    }
  }
}
```

This allows us to use the resource cache anymore without needing to know the current path and passing it through everywhere.

Current resource cache structure:
``` json
{
  "de": {
    "/testumgebung/de": {
      "https://cms.integreat-app.de/my-pdf": {
        "filePath": "/path/to/resource/cache/my-pdf-hash"
      },
      "https://cms.integreat-app.de/my-image": {
        "filePath": "/path/to/resource/cache/my-image-hash"
      }
    },
    "/testumgebung/de/willkommen": {
      "https://cms.integreat-app.de/my-image": {
        "filePath": "/path/to/resource/cache/my-image-hash"
      }
    }
  }
}
```

Issue #3522, opened 2025-09-12
Make android app 16kb page size compliant


steffenkleinle:
### Describe the Problem

Google requires apps to be 16kB page size compliant starting 01.11.2025. While its possible to request an extension in the Google Play Console, we should make our app compliant.
More information: https://play.google.com/console/u/0/developers/5269030016285360668/app/4974497225878557026/policy-center/issues/4987098980781988915/details

### Describe Your Preferred Solution

Make the app 16 KB page size compliant:
Libraries that do not support 16 KB:
base/lib/arm64-v8a/libglob.so
base/lib/arm64-v8a/liblighttpd.so
base/lib/arm64-v8a/libpcre2-8.so
base/lib/arm64-v8a/libpdfiumandroid.so
base/lib/x86_64/libglob.so
base/lib/x86_64/liblighttpd.so
base/lib/x86_64/libpcre2-8.so
base/lib/x86_64/libpdfiumandroid.so

`libpdfiumandroid.so` is due to [react-native-pdf](ttps://github.com/wonday/react-native-pdf). The library has not received any update adding support, but there is an instruction for a patch in https://github.com/wonday/react-native-pdf/issues/947#issuecomment-2999541826. Alternatively, we might consider switching to https://github.com/douglasjunior/react-native-pdf-renderer.
The rest seems to stem from [@dr.pogodin/react-native-static-server](https://github.com/birdofpreyru/react-native-static-server/issues/147), which added support in [v0.24.1](https://github.com/birdofpreyru/react-native-static-server/releases/tag/v0.24.1).

### Describe Possible Alternatives

N/A

### Related Issues

Lunes: https://github.com/digitalfabrik/lunes-app/issues/1152
Entitlementcard: https://github.com/digitalfabrik/entitlementcard/issues/2494

### Additional Information

https://play.google.com/console/u/0/developers/5269030016285360668/app/4974497225878557026/policy-center/issues/4987098980781988915/details

bahaaTuffaha:
Look at this:
https://github.com/douglasjunior/react-native-pdf-renderer?tab=readme-ov-file

bahaaTuffaha:
> > Look at this: https://github.com/douglasjunior/react-native-pdf-renderer?tab=readme-ov-file
> 
> Nice find, looks very promising. Definitely worth a try!

One issue I found that it doesn't handle links like the one we have `onPressLink={url => openExternalUrl(url, showSnackbar)}`

[...]

PR #3627, opened 2025-10-17
3626: Simplify resource cache and fix first load

steffenkleinle:
### Short Description

This PR simplifies the resource cache and its usage according to the issue (#3626).
It also fixes an issue that the resource cache is not cached correctly on first app load.

### Proposed Changes

- Correctly cache resource cache on first load
- Simplify resource cache structure by omitting the grouping by page and using strings instead of object values
- Avoid passing resource cache to children and use it directly where needed

### Side Effects

N/A

### Testing

No testing necessary here, test #3522 very carefully instead.

### Resolved Issues

Fixes: #3626

---

### DIFF
--- a/native/src/components/RemoteContent.tsx
+++ b/native/src/components/RemoteContent.tsx
@@ -1,4 +1,5 @@
 import WebView, { WebViewMessageEvent, WebViewNavigation } from '@dr.pogodin/react-native-webview'
+import { mapValues } from 'lodash'
 import React, { ReactElement, useCallback, useContext, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Text, Platform, useWindowDimensions } from 'react-native'
@@ -73,13 +74,7 @@ const RemoteContent = ({
   const { t } = useTranslation()
   const { width: deviceWidth } = useWindowDimensions()
 
-  const resourceMap: { [url: string]: string } = Object.entries(resourceCache).reduce(
-    (previous, [url, { filePath }]) => ({
-      ...previous,
-      [url]: getStaticServerFileUrl(filePath, staticServerUrl),
-    }),
-    {},
-  )
+  const resourceMap = mapValues(resourceCache, filePath => getStaticServerFileUrl(filePath, staticServerUrl))
 
   useEffect(() => {
     // If it takes too long returning false in onShouldStartLoadWithRequest the webview loads the pressed url anyway on android.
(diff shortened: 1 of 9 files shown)


native/src/utils/DefaultDataContainer.ts
```diff
@@ -179,22 +141,12 @@ class DefaultDataContainer implements DataContainer {
     resourceCache: LanguageResourceCacheStateType,
   ): Promise<void> => {
     const context = new DatabaseContext(city)
-    const cache: Cache<CityResourceCacheStateType> = this.caches.resourceCache
-    const previousResourceCache = cache.getCached(context)
-
-    if (!previousResourceCache) {
-      await cache.cache(
```
steffenkleinle: This was the culprit @bahaaTuffaha of the cache not correctly updating on first load. We were setting the resource cache on the wrong cache instance since this is a local copy and not `this.caches.resourceCache` thats later read again.

native/src/components/RemoteContent.tsx
```diff
@@ -73,6 +73,14 @@ const RemoteContent = ({
...
   const { width: deviceWidth } = useWindowDimensions()
 
+  const resourceMap: { [url: string]: string } = Object.entries(resourceCache).reduce(
+    (previous, [url, { filePath }]) => ({
+      ...previous,
+      [url]: getStaticServerFileUrl(filePath, staticServerUrl),
+    }),
+    {},
+  )
+
```
bahaaTuffaha: I think it will create a new object for `...previous` each on each iteration.

Grok said it's complexity O(n²) so it gave me this as alternative (I couldn't comprehend the solution 😅 but looks clean and O(n)):
```js
const resourceMap = Object.fromEntries(
  Object.entries(resourceCache).map(([url, { filePath }]) => [
    url,
    getStaticServerFileUrl(filePath, staticServerUrl)
  ])
)
```

steffenkleinle: Good suggestion, but this actually made me think and realize that we don't even need any object values and just `{ url: path }` is enough, therefore i further reduced the complexity and made this obsolete.

native/src/utils/__tests__/helpers.spec.ts
bahaaTuffaha: I think we can add a test for `getCachedResource` in case if the cache is empty to expect the url to be the same.

bahaaTuffaha (approved):
Nice work.. Tested on Android API 36 ✅️

PR #3633, opened 2025-10-20
Fix breadcrumb icon color and remove redundant models and properties

steffenkleinle:
### Short Description

This PR removes the redundant breadcrumb model and unused properties and therefore simplifies the code.

### Proposed Changes

- Use text.secondary color for breadcrumb icons
- Remove redundant breadcrumb model
- Remove unused node property
- Remove unused TileModel

### Side Effects

N/A

### Testing

Check that the breadcrumbs still work as expected and everything has the correct color.

### Resolved Issues

Fixes: N/A

---


steffenkleinle:
> Note: on future integreat the home button was in white.

Exactly, the purpose of the PR is to make the home icon (and the menu icon with the three dots) not white but the same color as the text. I think the breadcrumbs should use a consistent color.

lunars97 (approved):
Really nice! Tested in Chrome 🚀

PR #3640, opened 2025-10-21
Prevent header action item from being focused twice

steffenkleinle:
### Short Description

Prevent header action item from being focused twice.

### Proposed Changes

- Use `component={Link}` instead of using `Button` and `Link` components

### Side Effects

N/A

### Testing

Use the tab key to navigate through the app until the language selection and see that its only selected once.

### Resolved Issues

Fixes: N/A

---


lunars97 (approved):
Thank you  for your quick fix! 👍

bahaaTuffaha (approved):
Looks good. Tested on chrome and Firefox ✅️

Issue #3635, opened 2025-10-21
Every page should include hidden links that allow skipping content blocks


hauf-toni:
### Describe the Problem

Currently users that navigate integreat via keyboard or screen reader have no way of bypassing certain content.

### Describe Your Preferred Solution

Each page should include hidden skip links that allow users (via keyboard or screen reader) to skip content and navigate directly to main landmarks like the main content, menu, or footer.

### Describe Possible Alternatives

/

### Related Issues

/

### Additional Information
Improves overall accessibility scores.
Tested here: https://accessibe.com/

Issue #3637, opened 2025-10-21
Page landmarks should be tagged and described for assestive technology


hauf-toni:
### Describe Your Preferred Solution

Landmarks like main content and footer should use their corresponding HTML5 elements (<main>, <footer>) or include an "aria-label" and "role" attribute (e.g., role="main" or role="contentinfo"). Landmarks like search and navigation are covered elsewhere.

### Describe Possible Alternatives

/

### Related Issues

/

### Additional Information

Improves overall accessibilty score.

hauf-toni:
@steffenkleinle fine with me

PR #3645, opened 2025-10-23
3635: Skip content accessibility

bahaaTuffaha:
### Short Description
- Currently users that navigate integreat via keyboard or screen reader have no way of bypassing certain content.

### Proposed Changes

- Created a component `SkipLinks` that show up at the top left with back drop similar css to www.accessibe.com.
- There is no footer for POI added a condition `showFooterSkip` and for mobile to skip to bottom nav instead.
- Wrote a variable `hrefTargetForContent` that checks for categoriesRoute to reach content directly instead of breadcrumbs.
- Added ids to let it jump to that section like an anchor.
- Added page landmarks (role) for header, layout/main, navigation tabs, and footer.
- Fixed tabs selection by adding `tabIndex={0}`.

### Side Effects

- None

### Testing

- Go to each page and test how the SkipLinks behaves.
- Test mobile screens.
- Test screen reader (on android maybe).

### Resolved Issues

Fixes: #3635, ~~#3637~~

---


web/src/components/BottomNavigation.tsx
```diff
@@ -71,7 +71,12 @@ const BottomNavigation = ({ cityModel, languageCode }: BottomNavigationProps): R
 
   return (
     <Container elevation={4}>
-      <StyledBottomNavigation id={BOTTOM_NAVIGATION_ELEMENT_ID} showLabels value={value} component='nav'>
+      <StyledBottomNavigation
+        id={BOTTOM_NAVIGATION_ELEMENT_ID}
+        showLabels
+        value={value}
+        component='nav'
+        role='navigation'>
```
steffenkleinle: ❓ Is it necessary to set the role to `navigation` on top of the component already being `nav`? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role#description

> It is preferable to use the HTML5 nav element to define a navigation landmark. If the HTML5 nav element technique is not being used, use a role="navigation" attribute to define a navigation landmark.

lunars97:
Thank you for your effort in making our project more accessible! :heart_hands: Tested with screen-reader as well, works as expected. I think we need to have only one `Skip to main content` otherwise it will lead to clutterung because we are anyway using other in-page navigation options like `regions` and `heading hierarchy` :thinking:

web/src/components/SkipLinks.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Chip from '@mui/material/Chip'
+import { styled } from '@mui/material/styles'
+import React, { ReactElement, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { MAIN_ELEMENT_ID } from '../constants'
+
+const SkipLink = styled(Button)`
+  position: fixed;
+  display: flex;
```
steffenkleinle: 🔧 Not necessary
```suggestion
```

web/src/components/SkipLinks.tsx
```diff
@@ -0,0 +1,53 @@
...
+  left: 20px;
+  top: 50px;
+  gap: 8px;
+  background-color: ${props => props.theme.palette.background.paper};
+  opacity: 0;
```
steffenkleinle: 🤔 Do we need the opacity? I don't see a difference without setting the opacity at all.

steffenkleinle (commented):
I love how it looks now, great work! Just a little cleanup :)

web/src/components/SkipToContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+import Backdrop from '@mui/material/Backdrop'
+import Button from '@mui/material/Button'
+import Chip from '@mui/material/Chip'
+import { styled } from '@mui/material/styles'
+import React, { ReactElement, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import { MAIN_ELEMENT_ID } from './Layout'
+
+const SkipLink = styled(Button)`
```
steffenkleinle:
```suggestion
const StyledButton = styled(Button)`
```

web/src/components/__tests__/SkipToContent.spec.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import { renderWithTheme } from '../../testing/render'
+import SkipToContent from '../SkipToContent'
+
+jest.mock('react-i18next')
+
+describe('SkipToContent', () => {
+  it('should render skip links', () => {
```
steffenkleinle:
```suggestion
  it('should render skip to content button', () => {
```

lunars97 (approved):
LGTM 🔥 Thank you!

Issue #3379, opened 2025-07-15
Use Mui Skeleton during loading


LeandraH:
### Describe the Problem

We can use the Mui Skeleton during loading times "to reduce load-time frustration".
Mui: https://mui.com/material-ui/react-skeleton/
Design: https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat--MUI-?node-id=138-0&p=f&m=dev

### Describe Your Preferred Solution

During loading of a page, show its skeleton. This concerns lots of pages, e.g. the category page or the region loading page. If it's unclear how the skeleton for a page should look, check back with @hauf-toni 

### Related Issues

Part of #3231

PR #3649, opened 2025-10-27
3379: Use MUI Skeleton for loading

lunars97:
### Short Description

This PR implements MUI Skeleton for loadig

### Proposed Changes

- Replaced `LoadingSpinner` with `Skeleton`
- Added styles to make it almost similar to the components (though not sure how much aligned it should be with the real components :thinking: )
- ~Deleted `LoadingSpinner` file as it is not needed anymore instead use CircleProgress from MUI if there is a need :thinking:~
- Fixes this flickering issue on search as well #3654

### Side Effects

- I hope none

### Testing
Check all the components / routes where it has been changed

### Resolved Issues

Fixes: #3379
#3654 

---

### DIFF
--- a/web/src/components/PoisDesktop.tsx
+++ b/web/src/components/PoisDesktop.tsx
@@ -36,6 +36,45 @@ const StyledMapAttribution = styled(MapAttribution)({
   left: 0,
 })
 
+const SkeletonPanelContent = () => (
+  <Stack paddingX={2}>
+    <SkeletonHeader width='60%' />
+    <SkeletonList />
+  </Stack>
+)
+
+type PanelContentProps = {
+  listRef: React.RefObject<HTMLDivElement | null>
+  canDeselect: boolean
+  deselect: () => void
+  pois: PoiModel[]
+  poi: PoiModel | undefined
+  scrollToTop: () => void
+  userLocation: LocationType | null
+  slug: string | undefined
+  switchPoi: (step: 1 | -1) => void
+}
+
+const PanelContent = ({
+  listRef,
+  canDeselect,
+  deselect,
+  pois,
+  poi,
+  scrollToTop,
+  userLocation,
+  slug,
+  switchPoi,
+}: PanelContentProps): ReactElement => (
+  <Stack justifyContent='space-between' height='100%'>
+    <ListViewWrapper ref={listRef}>
+      <PoiPanelHeader goBack={canDeselect ? deselect : null} />
+      <PoiSharedChildren pois={pois} poi={poi} scrollToTop={scrollToTop} userLocation={userLocation} slug={slug} />
+    </ListViewWrapper>
+    {poi && pois.length > 0 && <PoiPanelNavigation switchPoi={switchPoi} />}
+  </Stack>
+)
+
 type PoisDesktopProps = {
   data: PreparePoisReturn
   selectMapFeature: (mapFeature: MapFeature | null) => void
@@ -96,26 +135,23 @@ const PoisDesktop = ({
     listRef.current?.scrollTo({ top: mapFeature ? 0 : scrollOffset })
   }, [mapFeature, scrollOffset])
 
-  const PanelContent = (
-    <Stack justifyContent='space-between' height='100%'>
-      <ListViewWrapper ref={listRef}>
-        <PoiPanelHeader goBack={canDeselect ? deselect : null} />
-        <PoiSharedChildren pois={pois} poi={poi} scrollToTop={scrollToTop} userLocation={userLocation} slug={slug} />
-      </ListViewWrapper>
-      {poi && pois.length > 0 && <PoiPanelNavigation switchPoi={switchPoi} />}
-    </Stack>
-  )
-
   return (
     <>
       <PanelContainer>
         {loading ? (
-          <Stack paddingX={2}>
-            <SkeletonHeader width='60%' />
-            <SkeletonList />
-          </Stack>
+          <SkeletonPanelContent />
         ) : (
-          PanelContent
+          <PanelContent
+            listRef={listRef}
+            canDeselect={canDeselect}
+            deselect={deselect}
+            pois={pois}
+            poi={poi}
+            scrollToTop={scrollToTop}
+            userLocation={userLocation}
+            slug={slug}
+            switchPoi={switchPoi}
+          />
         )}
       </PanelContainer>
       <MapView
--- a/web/src/components/PoisMobile.tsx
+++ b/web/src/components/PoisMobile.tsx
@@ -7,6 +7,7 @@ import React, { ReactElement, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { LocationType, MapViewViewport, MapFeature, PreparePoisReturn } from 'shared'
+import { PoiModel } from 'shared/api'
 
 import useDimensions from '../hooks/useDimensions'
 import BottomActionSheet, { ScrollableBottomSheetRef } from './BottomActionSheet'
@@ -35,6 +36,34 @@ const GeocontrolContainer = styled(AttributionContainer)`
   right: 16px;
   margin-bottom: 24px;
 `
+
+const SkeletonPoiContent = () => (
+  <Stack paddingX={2}>
+    <SkeletonHeader width='40%' />
+    <SkeletonList />
+  </Stack>
+)
+
+type PoiContentProps = {
+  canDeselect: boolean
+  pois: PoiModel[]
+  poi: PoiModel | undefined
+  scrollToTop: () => void
+  userLocation: LocationType | null
+  slug: string | undefined
+  t: (key: string) => string
+}
+
+const PoiContent = ({ canDeselect, pois, poi, scrollToTop, userLocation, slug, t }: PoiContentProps): ReactElement => (
+  <Stack padding={2} gap={1}>
+    {!canDeselect && (
+      <Typography component='h1' variant='h3' alignContent='center'>
+        {t('common:nearby')}
+      </Typography>
+    )}
+    <PoiSharedChildren pois={pois} poi={poi} scrollToTop={scrollToTop} userLocation={userLocation} slug={slug} />
+  </Stack>
+)
 
 type PoisMobileProps = {
   data: PreparePoisReturn
@@ -129,25 +158,17 @@ const PoisMobile = ({
           </>
         }>
         {loading ? (
-          <Stack paddingX={2}>
-            <SkeletonHeader width='40%' />
-            <SkeletonList />
-          </Stack>
+          <SkeletonPoiContent />
         ) : (
-          <Stack padding={2} gap={1}>
-            {!canDeselect && (
-              <Typography component='h1' variant='h3' alignContent='center'>
-                {t('common:nearby')}
-              </Typography>
-            )}
-            <PoiSharedChildren
-              pois={pois}
-              poi={poi}
-              scrollToTop={scrollToTop}
-              userLocation={userLocation}
-              slug={slug}
-            />
-          </Stack>
+          <PoiContent
+            canDeselect={canDeselect}
+            pois={pois}
+            poi={poi}
+            scrollToTop={scrollToTop}
+            userLocation={userLocation}
+            slug={slug}
+            t={t}
+          />
         )}
       </BottomActionSheet>
     </>


web/src/components/Chat.tsx
```diff
@@ -27,6 +28,20 @@ const Container = styled(Stack)(({ theme }) => ({
...
+  gap: 6px;
+  width: 100%;
+`
+
+const StyledRight = styled(Box)`
+  display: flex;
+  justify-content: flex-start;
+  gap: 6px;
+  width: 100%;
+`
```
steffenkleinle: 🤔  Is this naming correct? `flex-start` should be `left` and `flex-end` `right`, shouldn't it?

web/src/components/PageSkeleton.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Stack from '@mui/material/Stack'
+import { styled } from '@mui/material/styles'
+import React, { ReactElement } from 'react'
+
+const StyledSkeletonBreadcrumb = styled(Skeleton)<SkeletonProps>`
+  margin-top: 16px;
+`
+
+const StyledStack = styled(Stack)`
+  width: 100%;
```
steffenkleinle: 🔧 No need for a styled component here, you can just directly set the width in the jsx.

web/src/components/PageSkeleton.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <StyledBox>
+      {Array.from({ length: 12 }).map((_, index) => (
+        <Skeleton
+          /* eslint-disable-next-line react/no-array-index-key */
+          key={index}
+          variant='text'
+          width={randomWidth}
+          height={30}
+        />
+      ))}
```
steffenkleinle: ❓ How about this to avoid using `index` as `key`? Same for other usages.
```suggestion
      {[...Array(NUM_SKELETONS).keys()].map(index => (
        <Skeleton key={index} variant='text' width={randomWidth} height={30} />
      ))}
```

web/src/components/PageSkeleton.tsx
```diff
@@ -0,0 +1,55 @@
...
+const MAX_WIDTH = 90
+const randomWidth = `${MIN_WIDTH + Math.random() * (MAX_WIDTH - MIN_WIDTH)}%`
+
+type PageSkeletonProps = {
+  showHeaderSkeleton?: boolean
+}
+
+const PageSkeleton = ({ showHeaderSkeleton = true }: PageSkeletonProps): ReactElement => (
+  <StyledStack direction='column' spacing={2}>
+    {showHeaderSkeleton && <StyledSkeletonBreadcrumb variant='text' component='span' width={180} height={30} />}
```
steffenkleinle: 🤔 Is this the correct naming? header vs breadcrumb?

web/src/routes/CategoriesPage.tsx
```diff
@@ -75,6 +107,26 @@ const CategoriesPage = ({ city, pathname, cityCode, languageCode }: CityRoutePro
...
+    if (!categories || !currentCategory) {
+      return false
+    }
+    try {
+      const mapModel = new CategoriesMapModel(categories)
+      return mapModel.getChildren(currentCategory).length === 0
+    } catch {
+      return false
+    }
+  }, [categories, currentCategory])
```
steffenkleinle: ❓ Can you please explain what we are doing here/why we need to simulate loading?
Why do we need a use memo? Also, `CategoriesMapModel` has an `isLeaf` method that can be used here.

web/src/routes/CategoriesPage.tsx
```diff
@@ -122,10 +174,20 @@ const CategoriesPage = ({ city, pathname, cityCode, languageCode }: CityRoutePro
...
-  if (categoriesLoading || parentsLoading || pathname !== previousPathname) {
+  if (isActuallyLoading) {
+    let skeleton = null
+
+    if (!currentCategory && !categoryId) {
+      skeleton = <RootCategorySkeleton />
+    } else if (!isLeafCategory) {
+      skeleton = <ListSkeleton />
+    } else {
+      skeleton = <PageSkeleton />
```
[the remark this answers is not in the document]
steffenkleinle: There is no way of knowing that beforehand :/ But we can just not show any skeleton before we know by using `boolean | null` type instead of just boolean and return nothing if it is null.

web/src/routes/EventsPage.tsx
```diff
@@ -81,10 +82,19 @@ const EventsPage = ({ city, pathname, languageCode, cityCode }: CityRouteProps):
     Toolbar: <CityContentToolbar slug={event?.slug} />,
   }
 
-  if (!events && loading) {
+  if (loading) {
     return (
       <CityContentLayout isLoading {...locationLayoutParams}>
-        <LoadingSpinner />
+        {!event && !eventId ? (
```
steffenkleinle: 🤔 Shouldn't the check for `eventId` be enough? Also, please don't use negations.

web/src/components/ListSkeleton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    {showHeaderSkeleton && (
+      <StyledSkeletonHeader
+        variant='text'
+        component='h1'
+        sx={{ fontSize: `${headerSize}` }}
+        width={skeletonHeaderWidth}
+      />
+    )}
+    {showSkeletonAdditionalText && <SkeletonText variant='text' width={260} sx={{ fontSize: '3rem' }} />}
+    {showSkeletonSearch && <Skeleton variant='rectangular' width='80%' height={42} />}
```
steffenkleinle: ❌ I think we should split this component. Imo it would be better if the `ListSkeleton` is only responsible to render a list skeleton, not also a breadcrumb, header, additional text and search. Those should go to separate (reusable) components instead to have a clear separation of concerns and avoid these hard to read conditions. Same goes for the `PageSkeleton`.

steffenkleinle (commented):
Thanks for all the effort put into this, looks great! 😍 I think this still needs some cleaning up, but obviously this is also still a draft PR!

Things that should still be changed:
- Avoid hardcoding heights and widths as much as possible. Instead, try to [infer the dimensions](https://mui.com/material-ui/react-skeleton/#inferring-dimensions) wherever possible. 
- `PageSkeleton` and especially `ListSkeleton` have lots of skeletons that should not be part of those components. A list skeleton should only render a list skeleton. Everything else should go to own components.
- Extract some skeletons to a separate component (e.g. PoisPage)

Minor: The list item skeletons for the events and news list are too short and don't fit the actual list item height, perhaps you could still try to improve that? Also, the poi panel width is different from the skeleton width.

One general remark: In cases like this were you are not sure if your approach is optimal or you want to get feedback first, it is usually a good idea to only start with a small piece (for example only one route) in the beginning, ask for feedback and then continue with the rest.

web/src/components/TtsContainer.tsx
```diff
@@ -178,7 +178,7 @@ const TtsContainer = ({ languageCode, children }: TtsContainerProps): ReactEleme
     <TtsContext.Provider value={ttsContextValue}>
       {children}
       <StyledBackdrop open={isInitializing}>
-        <LoadingSpinner />
+        <CircularProgress />
```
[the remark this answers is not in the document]
lunars97: I also agree with you, it looks better, but then it would be maybe too much to use it just in one or two places? :thinking: We could maybe use another one from mui

web/src/components/PageSkeleton.tsx
```diff
@@ -0,0 +1,19 @@
...
+const MAX_WIDTH = 90
+
+const PageSkeleton = (): ReactElement => (
+  <Stack direction='column' width='100%' marginTop={4}>
+    {[...Array(NUM_SKELETONS).keys()].map(index => {
+      const width = `${MIN_WIDTH + Math.random() * (MAX_WIDTH - MIN_WIDTH)}%`
+      return <Skeleton key={index} variant='text' width={width} height={30} />
+    })}
+    <Skeleton variant='text' width='30%' height={20} />
+  </Stack>
```
steffenkleinle: 💭 I am sorry, I missed that `<H1>` is actually part of a `Page`, so I think we should include it here after all! My bad :/

```suggestion
  <Stack direction='column' width='100%' marginTop={4}>
    <SkeletonHeader />
    {[...Array(NUM_SKELETONS).keys()].map(index => {
      const width = `${MIN_WIDTH + Math.random() * (MAX_WIDTH - MIN_WIDTH)}%`
      return <Skeleton key={index} variant='text' width={width} height={30} />
    })}
    <Skeleton variant='text' width='30%' height={20} />
  </Stack>
```

web/src/routes/DisclaimerPage.tsx
```diff
@@ -46,7 +47,8 @@ const DisclaimerPage = ({ cityCode, languageCode, city }: CityRouteProps): React
   if (loading) {
     return (
       <CityContentLayout isLoading {...locationLayoutParams}>
-        <LoadingSpinner />
+        <SkeletonHeader />
+        <PageSkeleton />
```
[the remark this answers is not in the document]
lunars97: you are right :)

web/src/routes/CategoriesPage.tsx
```diff
@@ -23,19 +28,47 @@ import CityContentLayout, { CityContentLayoutProps } from '../components/CityCon
...
+
+const StyledCard = styled(Card)`
+  width: 160px;
+`
+
+const RootCategorySkeleton = (): ReactElement => (
+  <Stack paddingTop={2} alignItems='center'>
+    <SkeletonHeader />
+    <StyledBox>
+      {[...Array(NUM_CARD_SKELETONS).keys()].map(index => (
```
bahaaTuffaha: here's another way without spreading:
```suggestion
      {Array.from({ length: NUM_CARD_SKELETONS }, (_, index) => (
```

web/src/CityContentSwitcher.tsx
```diff
@@ -108,9 +107,8 @@ const CityContentSwitcher = ({ languageCode }: CityContentSwitcherProps): ReactE
                 languageCode={languageCode}
                 isLoading
                 city={city}
-                pageTitle={null}>
-                <LoadingSpinner />
-              </CityContentLayout>
```
bahaaTuffaha: ❓  why you didn't use a `PageSkeleton` here instead ?

lunars97: because there is no need for this, we are anyway showing skeleton in necessary components :thinking:

bahaaTuffaha:
I don't know what happened but some pages doesn't load content:
http://localhost:9000/augsburg/de/deutsche-sprache/sprachkurse/deutschkurse-fuer-gefluechtete

web/src/components/Chat.tsx
```diff
@@ -97,7 +121,11 @@ const Chat = ({
 
   return (
     <Container justifyContent='space-between'>
-      <ChatConversation messages={messages} isTyping={isTyping} />
+      {isLoading ? (
+        renderChatConversationSkeleton()
+      ) : (
+        <ChatConversation messages={messages} isTyping={isTyping} loading={isLoading} />
+      )}
```
steffenkleinle: 🤔 Since we pass `loading` anyway to `ChatConversation`, I think it would be better to also move the rendering to the skeleton to `ChatConversation`. That way, there is a clearer separation of concerns and less complexity here and you can perhaps reuse some of the components to get the size/widths for the skeleton.

web/src/routes/CategoriesPage.tsx
```diff
@@ -73,8 +64,50 @@ const CategoriesPage = ({ city, pathname, cityCode, languageCode }: CityRoutePro
...
+
+  const isLeafPage = useMemo(() => {
+    if (!categories || !currentCategory) {
+      return null
+    }
+    try {
+      return new CategoriesMapModel(categories).isLeaf(currentCategory)
+    } catch {
+      return null
+    }
```
steffenkleinle: 🤔 Why do we need a try catch block here? If we don't need it, the logic can be simplified to:
``` tsx
  return categories && currentCategory? new CategoriesMapModel(categories).isLeaf(currentCategory) : null
```

web/src/routes/TuNewsDetailPage.tsx
```diff
@@ -69,7 +69,12 @@ const TuNewsDetailPage = ({ city, pathname, cityCode, languageCode }: CityRouteP
...
-        <LoadingSpinner />
+        <TuNewsBanner>
+          <IconContainer width={180} height='100%'>
+            <Svg src={TuNewsActiveIcon} width='100%' height='100%' />
+          </IconContainer>
+        </TuNewsBanner>
+        <SkeletonPage />
       </CityContentLayout>
     )
   }
```
steffenkleinle: 🤔 Same as for `LocalNewsPage`, what do you think of removing this `if (loading)` block and instead moving `SkeletonPage` to L111 and `SkeletonList` down to 108 in the same way we did for events and landing in #3379?
This allows us to remove some conditionals and avoids duplicating the `TuNewsBanner`.

web/src/routes/SearchPage.tsx
```diff
@@ -85,6 +98,17 @@ const SearchPage = ({ city, cityCode, languageCode }: CityRouteProps): ReactElem
...
+  // for smooth loading transition to avoid flickering
+  useEffect(() => {
+    let timeout: number | undefined
+    if (loading) {
+      setShowSkeleton(true)
+    } else {
+      timeout = window.setTimeout(() => setShowSkeleton(false), LOAD_SKELETON_TIMEOUT)
+    }
+    return () => clearTimeout(timeout)
+  }, [loading])
```
steffenkleinle: ❌ I think this can be solved in a better way without any timeouts or changes in `SearchResults` (except replacing `LoadingSpinner` with `SkeletonList`):

The problem here is that the current `loading` value is only indicating whether the `useLoadSearchDocuments` fetching the data is still in progress. However, after that is done, the `useSearch` hook still needs some time to index the documents and return results, which is currently not taken into account. In order to avoid flickering, you only need to change L167 as follows:
``` tsx
<SearchResults results={results} query={debouncedQuery} loading={loading || contentLangaugeReturn.loading} />
```

web/src/routes/PoisPage.tsx
```diff
@@ -9,12 +13,43 @@ import { CityRouteProps } from '../CityContentSwitcher'
...
+
+const MapSkeleton = styled(Grid)(({ theme }) => ({
+  flexGrow: 1,
+  width: 'auto',
+  height: '100%',
+  [theme.breakpoints.down('md')]: {
+    height: '50%',
+    width: '100%',
+  },
+}))
```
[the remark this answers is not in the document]
lunars97: I would then need to adjust for both of them: **PoisMobile** and **PoisDesktop**. And `PanelContainer` is used only for `PoisDesktop`. By using `Grid` I am adjusting it responsively :thinking: which reduces each component adjustments and avoids passing `loading` to both components. I can decrease the height of the list view in mobile so it will look similar and I guess it should not be 100% replica of all the components

web/src/routes/LocalNewsPage.tsx
```diff
@@ -71,19 +72,29 @@ const LocalNewsPage = ({ city, pathname, languageCode, cityCode }: CityRouteProp
...
       </CityContentLayout>
     )
   }
 
-  if (!localNews || (newsId && !newsModel)) {
+  const missingList = localNews == null
+  const missingDetailPage = newsId != null && newsModel == null
+  const hasDataError = !loading && (missingList || missingDetailPage)
+
+  if (hasDataError) {
```
steffenkleinle: 🔧 
```suggestion
  if (newsError) {
```

web/src/routes/SearchPage.tsx
```diff
@@ -38,26 +40,36 @@ type SearchProps = {
...
-    return <LoadingSpinner />
+    return <SkeletonList listItemHeight={64} />
   }
 
-  const items = results.map(({ title, content, path }) => (
-    <SearchListItem title={title} contentWithoutHtml={parseHTML(content)} key={path} query={query} path={path} />
-  ))
+  if (!hasResults) {
+    return <SearchFeedback noResults query={query} />
+  }
```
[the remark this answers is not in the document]
lunars97: you are right, I forgot about it completely :see_no_evil:

web/src/routes/PoisPage.tsx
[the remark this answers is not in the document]
lunars97: I passed the logic to responsible components :+1:

web/src/routes/TuNewsDetailPage.tsx
```diff
@@ -100,12 +92,18 @@ const TuNewsDetailPage = ({ city, pathname, cityCode, languageCode }: CityRouteP
...
-      <Page
-        title={newsModel.title}
-        content={newsModel.content}
-        lastUpdate={newsModel.date}
-        showLastUpdateText={false}
-      />
+      {loading ? (
+        <SkeletonPage />
+      ) : (
+        newsModel && (
```
steffenkleinle: 🔧 The check for `loading` should not be necessary.
```suggestion
      {!newsModel ? (
        <SkeletonPage />
      ) : (
```

web/src/routes/LocalNewsPage.tsx
```diff
@@ -71,19 +72,29 @@ const LocalNewsPage = ({ city, pathname, languageCode, cityCode }: CityRouteProp
...
+              tunewsEnabled={city.tunewsEnabled}
+              localNewsEnabled={city.localNewsEnabled}
+              language={languageCode}
+            />
+            <SkeletonList />
+          </>
+        )}
       </CityContentLayout>
     )
   }
```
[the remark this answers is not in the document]
steffenkleinle: Replacing the condition in L108 with the following would fix that:
``` tsx
  if (newsId) {
    if (!newsModel) {
      return (
        <CityContentLayout isLoading={false} {...locationLayoutParams}>
          <SkeletonPage />
        </CityContentLayout>
      )
    }
    // Render news detail
  }
```

PR #3664, opened 2025-10-30
Group release notes by platform

steffenkleinle:
### Short Description

There was a wish from our testers to make it visible which platforms (web/native) PRs belong to.
This PR would adjust our release notes template to group native only/web only PRs separately.

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -10,6 +10,24 @@ changelog:
         labels:
           - not-testable
           - maintenance
+          - web
+          - native
+    - title: '[web] Features, fixes, enhancements 🎉'
+      labels:
+        - web
+      exclude:
+        labels:
+          - not-testable
+          - maintenance
+          - native
+    - title: '[native] Features, fixes, enhancements 🎉'
+      labels:
+        - native
+      exclude:
+        labels:
+          - not-testable
+          - maintenance
+          - web
     - title: Maintenance 🛠
       labels:
         - not-testable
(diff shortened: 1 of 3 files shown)


steffenkleinle:
@bahaaTuffaha @LeandraH @lunars97 this would mean that we have to set the web/native labels on corresponding PRs that are web or native only. Do NOT set labels if the PR is relevant for both web AND native.

Another possible option: We could also add `[native]`/`[web]` to the PR title, but that would not sort the release notes accordingly.

What do you think?

.github/release.yml
```diff
@@ -10,6 +10,24 @@ changelog:
...
+      labels:
+        - web
+      exclude:
+        labels:
+          - not-testable
+          - maintenance
+          - native
+    - title: [native] Features, fixes, enhancements 🎉
+      labels:
+        - native
```
steffenkleinle: Do you think we should also add specific android/ios groups? Or should we handle those by using [android]/[ios] in the PR title?

steffenkleinle:
> Hmmm, looks reasonable but I think I'd prefer to have a workflow where we add `web` and `native` tags when a PR affects both. I think it's easier to remember that we always need to add them, than to remember that we need to add them only if the PR affects only one of them.

Sadly, I don't think that is possible. I think there is no way to use `AND` conditions atm which would be needed to only add PRs that have both `native` and `web` labels set to the shared section.

Issue #3683, opened 2025-11-03
Information concerning organisations are not displayed correctly


laravoelk:
### Describe the Bug

Some pages in the Integreat-App are written from external organisations. This is marked at the bottom of the relevant page. This is not shown corretly with the new design system.

### Steps to Reproduce

1. Go to one of the pages, where external organisations are named. E.G.: https://integreat.app/integreat-inhalte/de/kindergeld-und-kinderzuschlag
4. See display of the organisation at the bottom of the page

### Expected Behavior

Everything should be readable.

### Additional Information

See screenshot.

<img width="1066" height="581" alt="Image" src="https://github.com/user-attachments/assets/b54da94d-c0ee-48d7-9be6-a96cc00873dc" />

steffenkleinle:
Yes, exactly. Its included in https://github.com/digitalfabrik/integreat-app/releases/tag/2025.11.1 which is currently in beta/testing.

PR #3684, opened 2025-11-03
3683: Fix organization content info

steffenkleinle:
### Short Description

Fix organization content info.

### Proposed Changes

- Set max width on image instead of container

### Side Effects

N/A

### Testing

http://localhost:9000/integreat-inhalte/de/kindergeld-und-kinderzuschlag (production CMS)

### Resolved Issues

Fixes: #3683

---

### DIFF
--- a/web/src/components/OrganizationContentInfo.tsx
+++ b/web/src/components/OrganizationContentInfo.tsx
@@ -17,6 +17,7 @@ const StyledCard = styled(Card)({
 
 const StyledImage = styled('img')({
   width: 'auto',
+  maxWidth: 320,
   height: '100%',
   objectFit: 'contain',
 })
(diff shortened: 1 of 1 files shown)


web/src/components/OrganizationContentInfo.tsx
```diff
@@ -32,7 +33,7 @@ const OrganizationContentInfo = ({ organization }: OrganizationContentInfoProps)
   return (
     <StyledCard elevation={2}>
       <Stack direction={mobile ? 'column' : 'row'} padding={2} gap={2}>
-        <Stack maxWidth={160} height={80} alignItems='flex-start'>
+        <Stack height={80} alignItems='flex-start'>
```
bahaaTuffaha: ❓ is there a reason we used flex-start here?
we can add `alignItems='center'` to the first stack.

```suggestion
<Stack direction={mobile ? 'column' : 'row'} padding={2} gap={2} alignItems='center'>
        <Stack height={80} alignItems='flex-start'>
```

Image

bahaaTuffaha (approved):
Tested on Chrome ✅️

web/src/components/OrganizationContentInfo.tsx
steffenkleinle: On mobile it looks a lot better imo if its aligned to the start. Also on desktop `alignItems='center'` only looks better if the image has white background/padding on top. If the text is aligned to the top, `flex-start` is better imo. Since it therefore depends on the specific image in question, I'd keep it as it is.

Issue #3682, opened 2025-11-03
Malte ~ Location “Munich” is displayed as example suggestion


kalairubachandrakumaran-oss:
### Describe the Problem

In the Malte application, the location “Munich” is displayed by default (as it is in the Integrate App).
However, when the user tries to search for or select “Munich” in the location filter, it cannot be found ->  the system does not provide any selectable result for this city.

This creates confusion, since the default view suggests that “Munich” is a valid location, but it is actually not available for filtering in the Malte app.

### Describe Your Preferred Solution

Only suggest or display places that are actually available and filterable in the respective application -> in that case for Malte

### Additional Information

<img width="960" height="364" alt="Image" src="https://github.com/user-attachments/assets/33ff5a3e-9356-4d48-8d63-7264d78d53ca" />

<img width="767" height="468" alt="Image" src="https://github.com/user-attachments/assets/e4553b3f-392e-484c-a0b6-541097b2c9c2" />

Issue #3567, opened 2025-09-30
Use "München" as Placeholder in Region Search Bar


LeandraH:
### Describe the Problem

As discussed with Toni, we want to change the placeholder in the regions search bar to "München" to make it clearer that you can't find content here yet.

### Describe Your Preferred Solution

Replace the landing page search placeholder with `München`.

### Describe Possible Alternatives

N/A

### Related Issues

Discussed in: https://github.com/digitalfabrik/integreat-app/pull/3549#pullrequestreview-3285491942

### Additional Information

N/A

PR #3685, opened 2025-11-04
3682: Use actually existing region as example city

steffenkleinle:
### Short Description

Use actually existing region as example city for malte city search description.

### Proposed Changes

- Use `Landkreis München` if it exists and the first city otherwise
- Pass the example city name to the translation

### Side Effects

N/A

### Testing

Checkout the city list for both integreat and malte.

### Resolved Issues

Fixes: #3682

---

### DIFF
--- a/web/src/components/SearchInput.tsx
+++ b/web/src/components/SearchInput.tsx
@@ -49,7 +49,7 @@ const SearchInput = ({
     <StyledTextField
       id='search'
       placeholder={placeholderText}
-      aria-label={placeholderText}
+      aria-label={description ? undefined : placeholderText}
       value={filterText}
       helperText={description}
       onChange={event => onFilterTextChange(event.target.value)}


web/src/components/CitySelector.tsx
```diff
@@ -47,9 +47,9 @@ const CitySelector = ({ cities, language, stickyTop }: CitySelectorProps): React
       <Typography variant='body1'>{t('welcomeInformation')}</Typography>
       <SearchInput
         filterText={filterText}
-        placeholderText={CITY_SEARCH_PLACEHOLDER}
+        placeholderText={exampleCity?.sortingName ?? CITY_SEARCH_EXAMPLE}
```
lunars97: shouldn't be it also `exampleCity?.name` 🤔

steffenkleinle: > shouldn't be it also exampleCity?.name 🤔

I just replicated the previous state here where it was just `München` and not `Landkreis München`. Kind of makes sense, since users probably don't search for the exact term but rather just a city name.

> can we add more context it like this? Because I wonder that some users might be confused to see just name there and might think that this is for ex. about München 🤔 and this placeholder name is also used as the aria-label which just says München What do you think?

I think this is a question for @LeandraH and @hauf-toni since this seems to have been deliberately changed in https://github.com/digitalfabrik/integreat-app/pull/3569 (https://github.com/digitalfabrik/integreat-app/issues/3567). Personally I don't mind, but I do think your proposal is too long for a placeholder (at least in the UI). However, we could set the `aria-label` separately from the placeholder.

steffenkleinle: @lunars97 actually do you know if it is recommended to even set an aria-label if we have a specific helper text/label already with `aria-describedby`? Feels a little redundant to me.

lunars97: I am not sure either whether description for input `Wähle deine Region oder die nächstgrößere Stadt` suitable for input which is literary says `Choose your region or the next largest city` instead of `Search your region...` But this should be discussed probably in another issue for this :thinking:

steffenkleinle:
> ->Translation for city search description in some languages is missing, is it a CMS issue ? @steffenkleinle

No, this is actually an app issue, we just did not get translations for our constants yet. But to be expected.

PR #3691, opened 2025-11-04
3527: Add spdx format SBoM to releases

lunars97:
### Short Description
This PR adds SBoM to releases and shows link to new release where `sbom.json` is attached to assets

### Proposed Changes

- [x] Use Syft https://github.com/anchore/syft to generate SBoM with spdx format
- [x] Should be a build artifact attached to a release
- [x] Add the link to new release so that end users could see `sbom.json` file in assets (instead of showing the direct path to `sbom.json` file which automatically downloads the file, it is better to show the path to release where the end user can have a look at assets and find the `sbom` file there to download)

### Side Effects

- should be produced after build and attached to releases

### Testing
it will be triggered on release

can be tested locally by running in the root
`curl -sSfL https://get.anchore.io/syft | sh -s -- -b /usr/local/bin
export SYFT_FORMAT_SPDX_JSON_PRETTY=true
syft dir:. -o spdx-json=sbom.json`

### Resolved Issues

Fixes: #3527

---

### DIFF
--- a/.circleci/src/commands/generate_sbom.yml
+++ b/.circleci/src/commands/generate_sbom.yml
@@ -10,14 +10,6 @@ steps:
         syft version
   - run:
       name: Generate SPDX SBoM
-      command: syft dir:. -o spdx-json=sbom.json
-  - persist_to_workspace:
-      root: .
-      paths:
-        - sbom.json
+      command: syft dir:. -o spdx-json=~/sbom.json
   - store_artifacts:
-      path: sbom.json
-  - persist_to_workspace:
-      root: .
-      paths:
-        - sbom.json
+      path: ~/sbom.json
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,18 +42,10 @@ commands:
                     syft version
                 name: Install Syft
             - run:
-                command: syft dir:. -o spdx-json=sbom.json
+                command: syft dir:. -o spdx-json=~/sbom.json
                 name: Generate SPDX SBoM
-            - persist_to_workspace:
-                paths:
-                    - sbom.json
-                root: .
             - store_artifacts:
-                path: sbom.json
-            - persist_to_workspace:
-                paths:
-                    - sbom.json
-                root: .
+                path: ~/sbom.json
     notify:
         description: Send a notification (to Mattermost) at the end of a job, based on success or failure. Must be the last step in a job.
         parameters:
@@ -314,7 +306,6 @@ jobs:
                         paths:
                             - << parameters.build_config_name >>-apk-url
                         root: ./
-            - generate_sbom
             - notify
     build_ios:
         environment:
@@ -382,7 +373,6 @@ jobs:
                         paths:
                             - << parameters.build_config_name >>-ipa-url
                         root: ./
-            - generate_sbom
             - notify
     build_web:
         docker:
@@ -426,7 +416,6 @@ jobs:
                 paths:
                     - dist
                 root: web
-            - generate_sbom
             - notify
     bump_version:
         docker:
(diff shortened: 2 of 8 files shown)


.circleci/src/commands/generate_sbom.yml
```diff
@@ -0,0 +1,28 @@
...
+        echo "Installing Syft..."
+        curl -sSfL https://get.anchore.io/syft | sudo sh -s -- -b /usr/local/bin
+        export SYFT_SOURCE_NAME="integreat-app"
+        export SYFT_FORMAT_SPDX_JSON_PRETTY=true
+        syft version
+
+  - run:
+      name: Generate SPDX SBOM
+      command: |
+        export SYFT_FORMAT_SPDX_JSON_PRETTY=true
```
steffenkleinle: 🔧 You already exported this above
```suggestion
```

.circleci/src/commands/generate_sbom.yml
```diff
@@ -0,0 +1,28 @@
+description: Generate SPDX SBOM using Syft
+
```
steffenkleinle: 🙃 We don't use blank lines in yml files usually, same below
```suggestion
```

.circleci/src/jobs/generate_sbom.yml
steffenkleinle: 🤔 Why don't you use the `generate_sbom` command you created above? Also, can't we just move the sbom generation inside the `build_android`, `build_ios` and `build_web` steps to avoid needing a separate job? I also think thats the logical place to put it imo.

Advantages:
- No need for a separate job which increases credit usage (spinup, checkout, ...)
- No need to add the new job for all deliveries (delivery, native_production_delivery, native_beta_delivery, web_production_delivery, web_beta_delivery) -> less repetition

You can just use a command like this:
``` yml
steps:
  ...
  - generate_sbom
  ...
```

.circleci/src/commands/generate_sbom.yml
lunars97: :see_no_evil:

.circleci/src/jobs/generate_sbom.yml
```diff
@@ -0,0 +1,30 @@
...
+      command: |
+        echo "Installing Syft..."
+        curl -sSfL https://get.anchore.io/syft | sh -s -- -b /usr/local/bin
+        syft version
+
+  - run:
+      name: Generate SPDX SBOM
+      command: |
+        export SYFT_FORMAT_SPDX_JSON_PRETTY=true
+        syft dir:. -o spdx-json=_manifest/spdx_2.2/manifest.spdx.json
```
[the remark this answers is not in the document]
lunars97: I am not sure about it, I saw it from cms implementation and saw a definition of it, but for file :) `A manifest file contains metadata about the project such as its name, author, version-details, description and details about its components.`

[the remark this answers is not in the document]
steffenkleinle: Yes, maybe its a manifest file but for us its a sbom, so I'd prefer that name. Otherwise it is not clear what the file is about if you just see the file name imo.

.circleci/src/workflows/native_production_delivery.yml
```diff
@@ -37,7 +37,6 @@ jobs:
         - check
         - e2e_android
         - build_android-<< matrix.build_config_name >>
-
```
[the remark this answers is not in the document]
lunars97: it worked, thank you :+1:

lunars97:
@steffenkleinle could you elaborate here a bit :) I think we can now test it out `In order to test this, you'll need to temporarily adjust the commit workflow and run a job there that does execute the sbom generation`

~/.appium/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
```
lunars97: 👀 will be removed

.circleci/src/workflows/commit.yml
```diff
@@ -1,4 +1,13 @@
...
+  - bump_version:
+      prepare_delivery: true
+      context:
+        - deliverino
+  - notify_release:
+      production_delivery: false
+      context:
+        - deliverino
+      requires:
+        - bump_version
```
steffenkleinle: ❌ Please revert this
```suggestion
  - check
```

Issue #3681, opened 2025-11-03
Map ~ Filter resets as soon as the User clicks on one of the suggested results


kalairubachandrakumaran-oss:
### Describe the Bug
When the user filters by one category in the location filter, the filter resets as soon as they click on one of the suggested results :
-> it jumps back, and the default results are shown again
-> and the previously set filter is removed

Hint: If the user selects the checkbox “Show only currently open places”, this filter remains active until they manually deselect it

### Steps to Reproduce

1.    Go to Map in the Integrate App
2.    Click on Filter Location
3.    Select a category
4.    Click on one of the suggested results
5.    Observe that the selected filter resets, and the default results are displayed again

### Expected Behavior
The filter for selected category should remain active after clicking on a suggested result ( like the checkbox “Show only currently open places”)

### Additional Information

https://github.com/user-attachments/assets/a58274c8-9fda-47b3-81db-8bfd210f5a5d

PR #3698, opened 2025-11-05
3681: Fix poi filter query

bahaaTuffaha:
### Short Description

When the user filters by one category in the location filter, the filter resets as soon as they click on one of the suggested results.

### Proposed Changes

- Added queryParams to the slug at `poiListItem.tsx`
- Removed useSearchParams mocks at `PoisDesktop.spec.tsx` because `renderWithRouterAndTheme` already has AllTheProviders including `<MemoryRouter>`.

### Side Effects

None.

### Testing
- Go to Map in the Integrate App
- Click on Filter Location
- Select a category
- Click on one of the suggested results

### Resolved Issues

Fixes: #3681

---


web/src/components/PoiListItem.tsx
```diff
@@ -34,11 +35,13 @@ type PoiListItemProps = {
...
 const PoiListItem = ({ poi, distance, selectPoi }: PoiListItemProps): ReactElement => {
   const { t } = useTranslation('pois')
+  const [queryParams] = useSearchParams()
   const { title, category, slug } = poi
+  const slugWithQuery = `${slug}?${queryParams}`
 
   return (
     <ListItem disablePadding>
-      <StyledListItemButton onClick={selectPoi} id={slug} to={slug} component={Link} aria-label={title}>
+      <StyledListItemButton onClick={selectPoi} id={slug} to={slugWithQuery} component={Link} aria-label={title}>
```
bahaaTuffaha: This could be solved actually in 2 ways:
1- The way I wrote it here.
2- Just removing the `to={slug} component={Link}` will solve that issue. I think the `component={Link}` is removing the the query and without it it will use onClick but we will lose the link accessibility thing.

steffenkleinle: I think option 1. is the way to go here, its more declarative and probably better in terms of a11y. However, we now have duplicated logic (once in `to` and also in `selectPoi` (passed down from `Pois.tsx`)). I think we should delete `selectPoi` there in https://github.com/digitalfabrik/integreat-app/blob/main/web/src/components/Pois.tsx#L83. This will simplify code and reduce duplication. Then we can rename `selectPoi` here to just `onClick` as it will just adjust the scroll position.

bahaaTuffaha: > I think option 1. is the way to go here, its more declarative and probably better in terms of a11y. However, we now have duplicated logic (once in `to` and also in `selectPoi` (passed down from `Pois.tsx`)). I think we should delete `selectPoi` there in https://github.com/digitalfabrik/integreat-app/blob/main/web/src/components/Pois.tsx#L83.

I can't delete `selectPoi` because switch poi depends on it for desktop 🤔

web/src/components/PoiListItem.tsx
```diff
@@ -28,17 +29,20 @@ const StyledText = styled('p')({
...
+const PoiListItem = ({ poi, distance, onClick }: PoiListItemProps): ReactElement => {
   const { t } = useTranslation('pois')
+  const [queryParams] = useSearchParams()
   const { title, category, slug } = poi
+  const slugWithQuery = `${slug}?${queryParams}`
 
   return (
     <ListItem disablePadding>
-      <StyledListItemButton onClick={selectPoi} id={slug} to={slug} component={Link} aria-label={title}>
+      {/* onClick here is just to preserve scroll position */}
```
steffenkleinle: 🔧 It doesn't really matter what `onClick` is doing here, keeping this comment means that the comment needs to be adjusted whenever some behavior is added/changed.
```suggestion
```

Issue #3710, opened 2025-11-11
Improve the thumbnail for PoiDetails


bahaaTuffaha:
### Describe the Problem

Some images are not meant for `object-fit: cover` so some information or the logo will get cutoff so we need a way to show the full image and keep the same space intact.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/4afbb88f-a87a-4b35-9ceb-bf9a5a02c12c" />

### Describe Your Preferred Solution

- To have a div that have the image as background (cover) with blur.
- inside the div we put the thumbnail image with contain object-fit.

for ex.:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/93176ef4-ac0a-4534-ab52-a53d02bb07cf" />

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/3517b16e-16cb-4917-af04-45ac79a315dc" />

### Describe Possible Alternatives

keep doing what we doing.

### Related Issues

none.

steffenkleinle:
Looks good!

PR #3713, opened 2025-11-12
3710: Improve the thumbnail for PoiDetails

bahaaTuffaha:
### Short Description

Some images are not meant for object-fit: cover so some information or the logo will get cutoff so we need a way to show the full image and keep the same space intact.

### Proposed Changes

- Created `ThumbnailBackground` that is blurred behind the actual `Thumbnail`.
- Some Thumbnail are `png` so will cause a problem with contrast so created a conditional `isNotPng` to hide the background.
- Adjusted the `SimpleImage` to accept blurRadius props.

### Side Effects

- None.

### Testing

- Go to the poi at any region and check the thumbnail.
http://localhost:9000/staedteregion-aachen/de/locations/abbba-ev
http://localhost:9000/testumgebung/de/locations/test-handynummer
- Check the png ones like: 
http://localhost:9000/staedteregion-aachen/de/locations/ausl%C3%A4nderamt-st%C3%A4dteregion-aachen-infostelle
http://localhost:9000/staedteregion-aachen/de/locations/abbba-ev

### Resolved Issues

Fixes: #3710

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,22 @@ commands:
             - run:
                 command: circleci config validate
                 name: Validate circle config
+    generate_sbom:
+        description: Generate SPDX SBoM
+        steps:
+            - run:
+                command: |
+                    echo "Installing Syft..."
+                    curl -sSfL https://get.anchore.io/syft | sudo sh -s -- -b /usr/local/bin
+                    export SYFT_SOURCE_NAME="integreat-app"
+                    export SYFT_FORMAT_SPDX_JSON_PRETTY=true
+                    syft version
+                name: Install Syft
+            - run:
+                command: syft dir:. -o spdx-json=~/sbom.json
+                name: Generate SPDX SBoM
+            - store_artifacts:
+                path: ~/sbom.json
     notify:
         description: Send a notification (to Mattermost) at the end of a job, based on success or failure. Must be the last step in a job.
         parameters:
@@ -778,12 +794,17 @@ jobs:
             - prepare_workspace
             - restore_environment_variables
             - restore_yarn_tools_cache
+            - generate_sbom
             - run:
                 command: |
                     RELEASE=$(yarn --silent app-toolbelt v0 release create << parameters.platform >> ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} <<# parameters.production_delivery >>--production-release<</ parameters.production_delivery >>)
                     echo "export RELEASE_ID=$(echo ${RELEASE} | jq .id)" >> ${BASH_ENV}
                     echo "export RELEASE_NOTES=$(echo ${RELEASE} | jq .releaseNotes)" >> ${BASH_ENV}
                 name: Create github release
+                working_directory: tools
+            - run:
+                command: yarn app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files ~/sbom.json --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
+                name: Upload SBoM to github release
                 working_directory: tools
             - unless:
                 condition:
--- a/.circleci/src/commands/generate_sbom.yml
+++ b/.circleci/src/commands/generate_sbom.yml
@@ -1,0 +1,15 @@
+description: Generate SPDX SBoM
+steps:
+  - run:
+      name: Install Syft
+      command: |
+        echo "Installing Syft..."
+        curl -sSfL https://get.anchore.io/syft | sudo sh -s -- -b /usr/local/bin
+        export SYFT_SOURCE_NAME="integreat-app"
+        export SYFT_FORMAT_SPDX_JSON_PRETTY=true
+        syft version
+  - run:
+      name: Generate SPDX SBoM
+      command: syft dir:. -o spdx-json=~/sbom.json
+  - store_artifacts:
+      path: ~/sbom.json
--- a/.circleci/src/jobs/notify_release.yml
+++ b/.circleci/src/jobs/notify_release.yml
@@ -14,12 +14,17 @@ steps:
   - prepare_workspace
   - restore_environment_variables
   - restore_yarn_tools_cache
+  - generate_sbom
   - run:
       name: Create github release
       command: |
         RELEASE=$(yarn --silent app-toolbelt v0 release create << parameters.platform >> ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} <<# parameters.production_delivery >>--production-release<</ parameters.production_delivery >>)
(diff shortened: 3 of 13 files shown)


steffenkleinle (commented):
Thanks for addressing this! I think for the eample TaT image it looks really good, but I am not sure this is a good fit for all images, especially if they are very narrow/have a small width.

🤔 I am not sure if the `isNotPng` logic is enough, also svgs and other file types can have transparent background which would cause issues. I would therefore be inclined to just show the image aligned to the left without any background/width set, see screenshot below, I think this would also look very reasonable. Alternatively, I'd say we do this for `jpg`/`jpeg` only instead of everything except `png`.

If we decide to go with the blur background, could you move the background logic to the image component or a new component to be reusable/not clutter the poi details?

image

Furthermore, currently the thumbnails are not visible in high contrast mode, what do you think just setting a `background-color: white` until we get svgs only for those kind of images from the CMS?

image

bahaaTuffaha:
> Furthermore, currently the thumbnails are not visible in high contrast mode, what do you think just setting a `background-color: white` until we get svgs only for those kind of images from the CMS?

- This is good Idea...
- I will investigate more to see what could be done about transparent images and SVG.

bahaaTuffaha:
Update:
I did some experiments:
- For web worked well with the drop-shadow thing (applied twice for higher intensity)
- For native there is no similar affect so I used the same ThumbnailBackground as outline ... it's not perfect 🫣 what do you think?
image

native/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const CustomThumbnail = ({ src }: CustomThumbnailProps): ReactElement => {
+  const theme = useTheme()
+  const isNonTransparent = !src.toLowerCase().includes('.png') && !src.toLowerCase().includes('.svg')
+  return (
+    <ThumbnailWrapper>
+      {isNonTransparent ? (
+        <ThumbnailBackground source={src} resizeMode='cover' blurRadius={3} />
+      ) : (
+        <ThumbnailBackground style={{ tintColor: theme.colors.outline, transform: [{ scale: 1.04 }] }} source={src} />
```
steffenkleinle: 🤔 `1.04` is a little random number, is this necessary/can't we just use `1` instead?

web/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,50 @@
...
+  border-radius: 8px;
+  overflow: hidden;
+
+  ${props => props.theme.breakpoints.down('md')} {
+    order: 1;
+    margin-top: 12px;
+  }
+`
+
+const Thumbnail = styled('img')<{ isPng?: boolean }>`
```
steffenkleinle: 🔧 The naming is not correct anymore
```suggestion
const Thumbnail = styled('img')<{ opaque?: boolean }>`
```

web/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,50 @@
...
+  object-fit: cover;
+  filter: blur(10px);
+`
+
+type CustomThumbnailProps = {
+  src: string
+}
+
+const CustomThumbnail = ({ src }: CustomThumbnailProps): ReactElement => {
+  const isNonTransparent = !src.toLowerCase().includes('.png') && !src.toLowerCase().includes('.svg')
```
steffenkleinle: 🔧 See comment for native

native/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { ReactElement } from 'react'
+import styled, { useTheme } from 'styled-components/native'
+
+import SimpleImage from './SimpleImage'
+
+const ThumbnailWrapper = styled.View`
+  position: relative;
```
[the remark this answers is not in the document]
bahaaTuffaha: the default is `position: static`
https://www.w3schools.com/css/css_positioning.asp

bahaaTuffaha:
Note: none of the colors we have at the theme worked as white color 😞

native/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const ThumbnailBackground = styled(SimpleImage)`
+  position: absolute;
+  height: 100%;
+  width: 100%;
+`
+
+const Thumbnail = styled(ThumbnailBackground)`
+  position: relative;
+`
```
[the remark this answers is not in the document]
steffenkleinle: Why don't you do the same for `WhiteBackground` though? 🤔

native/src/components/HorizontalLine.tsx
```diff
@@ -3,6 +3,6 @@ import styled from 'styled-components/native'
 const HorizontalLine = styled.View`
   border-bottom-width: 1px;
   border-color: ${props => props.theme.legacy.colors.textDisabledColor};
-  margin: 20px 0;
+  margin: 8px 0;
```
steffenkleinle: 💭 This will cause side effects on other screens, but since we are going to replace this anyway before it hits production its fine.

native/src/components/PoiDetails.tsx
```diff
@@ -61,25 +47,21 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
...
-                <Contact
-                  key={
-                    contact.headline ?? contact.website ?? contact.name ?? contact.phoneNumber ?? contact.mobileNumber
-                  }
-                  contact={contact}
-                  isLastContact={contacts.length - 1 === index}
-                />
-              ))}
-            </StyledContactsContainer>
+            {contacts.map((contact, index) => (
```
steffenkleinle: Nice refactoring here, great work! ❤️

steffenkleinle (approved):
Nice work!

native/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type CustomThumbnailProps = {
+  src: string
+}
+
+const CustomThumbnail = ({ src }: CustomThumbnailProps): ReactElement => {
+  const opaque = ['.jpg', '.jpeg'].some(extension => src.toLowerCase().includes(extension))
+  return (
+    <ThumbnailWrapper>
+      {opaque ? <ThumbnailBackground source={src} resizeMode='cover' blurRadius={3} /> : <WhiteBackground />}
```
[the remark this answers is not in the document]
LeandraH: Actually, I'm not sure about the shadow either. This is Aachen > Bürgerbüro Eschweiler, I really don't think that's ideal for our users.

Image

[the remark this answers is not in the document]
LeandraH: Huh, you're right, I thought I'd never checked out the branch before, but I apparently was on an older version 😅 

Looks great!

LeandraH (approved):
I still don't love the blurred picture in the background but you're right that it's a bit tricky to know which background color to use. I'd be inclined to just say white, but I haven't seen that many JPG icons either.

Tested in simulated iOS

LeandraH:
> > but I haven't seen that many JPG icons either.
> 
> sooo do I keep the blur? for the "aesthetic" 🙂

I would prefer if you remove it but I can be convinced otherwise :)

web/src/components/CustomThumbnail.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const Thumbnail = styled('img')`
+  position: relative;
+  height: 100%;
+  width: 100%;
+  object-fit: contain;
+`
+
+const WhiteBackground = styled('div')`
+  position: absolute;
```
LeandraH: Check for Leandra: Do we need an absolute positioning without any positions?

Issue #3669, opened 2025-10-30
Replace checkbox component


steffenkleinle:
### Describe the Problem

Replace our custom checkbox implementation with https://callstack.github.io/react-native-paper/docs/components/Checkbox.

### Describe Your Preferred Solution

N/A

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3721, opened 2025-11-13
3669: Replace custom checkbox with react native paper

bahaaTuffaha:
### Short Description

Replace custom checkbox with react native paper.

### Proposed Changes

- Removed old checkbox.
- Used checkbox from `react-native-paper`.

### Side Effects

- none

### Testing

- On native go to feedback form the side menu.
- Test the checkbox there.

### Resolved Issues

Fixes: #3669

---

### DIFF
--- a/native/src/components/CalendarChoiceModal.tsx
+++ b/native/src/components/CalendarChoiceModal.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { FlatList, View } from 'react-native'
 import { Calendar } from 'react-native-calendar-events'
+import { Divider } from 'react-native-paper'
 import styled from 'styled-components/native'
 
 import Modal from './Modal'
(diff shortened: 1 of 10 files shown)


bahaaTuffaha:
It looks fine on android but there is no box for iOS. Only when you click the icon will appear.
- I think this is expected: https://callstack.github.io/react-native-paper/docs/components/Checkbox/
simulator_screenshot_1904E795-6C44-45D6-A212-B489E15DAC9F

steffenkleinle:
> It looks fine on android but there is no box for iOS. Only when you click the icon will appear.
> 
>     * I think this is expected: https://callstack.github.io/react-native-paper/docs/components/Checkbox/

Yes, with this component this seems to be expected. I think this looks bad and is quite good from a UX perspective as it is not clear for users where to press. We should probably use [Checkbox.Android](https://callstack.github.io/react-native-paper/docs/components/Checkbox/CheckboxAndroid) instead to have a consistent UI with good UX.

Perhaps it would be even better to use [Checkbox.Item](https://callstack.github.io/react-native-paper/docs/components/Checkbox/CheckboxItem#mode) since it includes a pressable label.

Issue #3675, opened 2025-10-30
Replace radio button component


steffenkleinle:
### Describe the Problem

Replace our custom radio button implementation with https://callstack.github.io/react-native-paper/docs/components/RadioButton.

### Describe Your Preferred Solution

N/A

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3722, opened 2025-11-14
3675: Replace RadioButton with react-native-paper

bahaaTuffaha:
### Short Description

Replace our custom radio button implementation with https://callstack.github.io/react-native-paper/docs/components/RadioButton.

### Proposed Changes

- Refactored `RadioButton.tsx` to use react-native-paper.
- For `FormRadioButtons.tsx` I used  `RadioButton.Group` and `RadioButton.Item` because it has a build in label.

### Side Effects

- Check the kontaktformular page at Malte.

### Testing

- Test adding events to the calendar.
- Check the kontaktformular at Malte (I couldn't find the page).

### Resolved Issues

Fixes: #3675

---


native/src/components/base/RadioButton.tsx
```diff
@@ -43,7 +19,7 @@ type RadioButtonProps = {
 
 const RadioButton = ({ children, selected, select }: RadioButtonProps): ReactElement => (
   <Container onPress={select} role='radio'>
```
steffenkleinle: ❓ Why is this container/pressable still necessary? Also, I think we should just delete this component altogether and use the react-native-paper component directly.

steffenkleinle (changes requested):
Thanks, looks good already :)

native/src/components/base/RadioButton.tsx
steffenkleinle: See https://callstack.github.io/react-native-paper/docs/components/RadioButton/RadioButtonItem/#mode:
> RadioButton.Item allows you to press the whole row (item) instead of only the RadioButton.

native/src/components/CalendarChoiceModal.tsx
```diff
@@ -20,26 +20,13 @@ const StyledList = styled(FlatList as typeof FlatList<Calendar>)`
...
   background-color: ${props => props.theme.legacy.colors.textDecorationColor};
   height: 1px;
   margin: 8px 0;
 `
 
-const StyledText = styled.Text`
-  font-family: ${props => props.theme.legacy.fonts.native.decorativeFontRegular};
-  color: ${props => props.theme.legacy.colors.textColor};
-`
+type ExportType = 'only-this-event' | 'all-future-events'
```
steffenkleinle: 🙃 It would be possible to just convert a boolean value to a string (`'true'`/`'false'`) as well. Not sure if that would be better though.

steffenkleinle (approved):
Nice work!

steffenkleinle (approved):
Tested on Android

native/src/components/CalendarChoiceModal.tsx
```diff
@@ -66,30 +51,27 @@ const CalendarChoiceModal = ({
...
-          </RadioButton>
-          <RadioButton selected={exportAllEvents} select={() => setExportAllEvents(true)}>
-            <StyledText>{t('thisAndAllFutureEvents')}</StyledText>
-          </RadioButton>
+          <RadioButton.Group
+            onValueChange={value => setExportAllEvents(value === 'true')}
+            value={exportAllEvents.toString()}>
+            <RadioButton.Item mode='android' label={t('onlyThisEvent')} value='false' />
+            <RadioButton.Item mode='android' label={t('thisAndAllFutureEvents')} value='true' />
+          </RadioButton.Group>
```
LeandraH: ❓ Shouldn't that be `value={exportAllEvents}` or `value={!exportAllEvents}`?

native/ios/Integreat.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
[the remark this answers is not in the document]
LeandraH: Ah, you should run `bundle exec pod install`, that uses our configuration

steffenkleinle (approved):

Issue #3648, opened 2025-10-24
App's logo gets re-rendered when I change tabs


bahaaTuffaha:
### Describe the Bug

- App's logo gets re-rendered for split second when I change tabs.

⚠️ from what I saw the problem could be directly from the `<Svg/>` component because I replaced it with `<img>` at `HeaderLogo` and the flickering stooped but clearly is bad solution here.
### Steps to Reproduce

1. Go to integreat.
2. Select a region and keep switching tabs.

### Expected Behavior

- The Logo should stay the same as long the props are the same.

### Related Issues

- None.

### Additional Information
https://github.com/user-attachments/assets/4a627340-6b8f-4b6b-8557-abae1ae89a69

PR #3738, opened 2025-11-24
#3648-resolved

sam2002jain:


steffenkleinle:
Hey @sam2002jain, thank you so much for all the effort and time spent on this! Sadly, your solution is relatively complicated and a little hacky. @bahaaTuffaha also took a look at this issue and came up with a more easy solution in my opinion, see #3741. I am therefore inclined to go with that solution, sorry for the trouble! 😢

PR #3742, opened 2025-11-26
Fix double encoding of URLs

steffenkleinle:
### Short Description

Fix double encoding of URLs by decoding first.

### Proposed Changes

- Decode URLs before encoding them

### Side Effects

Hopefully none.

### Testing

Go to https://integreat.app/ortenaukreis/de/antr%C3%A4ge/wohngeld-2 and check that the URL works as expected (the `Ort eintragen` field should be displayed).

Check that different URLs (e.g. with arabic chars) still work as expected.

### Resolved Issues

Fixes: N/A

---


LeandraH (approved):
Thanks for the quick fix!

Issue #3745, opened 2025-11-28
Rewrite the ChangeLanguageModal


LeandraH:
### Describe the Problem

Rewrite our ChangeLanguageModal to use components from `react-native-paper` and add a search for languages.

### Describe Your Preferred Solution

- [x] The ChangeLanguageModal should use components from `react-native-paper`
- [x] There should be a search bar with which to filter the languages

### Describe Possible Alternatives

Nope.

### Related Issues

#3349 for web issue, #3622 for web PR with some discussion about the polyfills
Part of #3666

PR #3750, opened 2025-11-29
3745: Rewrite language change modal

LeandraH:
### Short Description

This PR replaces the items in the `LanguageChangeModal` with `List.Items` from `react-native-paper`, and adds a search functionality analogous to #3622.

### Proposed Changes

- Replace the items in the `SelectorItem` with `List.Item`
- Add a `Searchbar` to the `ChangeLanguageModal`
- Add a search function to search either for the language in its own language, in the currently selected language, or the fallback language (i.e. German)
- Add a big file with polyfill directions. We don't want to import all the languages, only the currently used one plus the fallback language. Unfortunately, `metro` doesn't allow dynamic addresses (https://github.com/facebook/metro/issues/52), so it's a big file of possible imports.
- Update the documentation for adding new languages

### Side Effects

- I added elevation colors to the theme because those were necessary for the `Searchbar`. The colors will be replaced when we use the Theme from `react-native-paper`, so the background colors might also still change a little bit.

### Testing

Search for languages in all kinds of languages

### Resolved Issues

Fixes: #3745 

---


native/src/utils/importDisplayNamesPackage.ts
```diff
@@ -0,0 +1,142 @@
+// Metro does not support dynamic imports yet: https://github.com/facebook/metro/issues/52
```
LeandraH: I am watching this issue, so should it ever get resolved, I'll open an issue to replace this file but I'm not very optimistic that it'll ever happen.

bahaaTuffaha (approved):
Great Work! ✨ 
Tested on Android and iOS

Issue #3671, opened 2025-10-30
Replace icon component


steffenkleinle:
### Describe the Problem

Replace our custom icon implementation with https://callstack.github.io/react-native-paper/docs/components/Icon.

### Describe Your Preferred Solution

N/A

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3757, opened 2025-12-03
Replace legacy colors for native

bahaaTuffaha:
### Short Description

⚠️ this is sub-branch of #3671 to replace legacy colors.

### Proposed Changes
replaced the following:
- `theme.legacy.colors.backgroundColor` => `theme.colors.background`
- `theme.legacy.colors.backgroundAccentColor` => `theme.colors.surface` or `theme.colors.surfaceVariant`
- `theme.legacy.colors.textColor` => `theme.colors.onSurface`
- `theme.legacy.colors.textSecondaryColor` => `theme.colors.onSurfaceVariant`
- `theme.legacy.colors.textDisabledColor` => `theme.colors.onSurfaceDisabled`
- `theme.legacy.colors.themeColor` => `theme.colors.secondary`
- `theme.legacy.colors.themeColorLight` => `theme.colors.secondaryContainer`
- `theme.legacy.colors.linkColor` => `theme.colors.primary`
- `theme.legacy.colors.invalidInput` => `theme.colors.error`
- `theme.legacy.colors.borderColor` => `theme.colors.outlineVariant`

I left the ones with `.fonts`, `isContrastTheme`, and couldnt find a replacement for`warningColor`, `negativeHighlight` , `positiveHighlight`, `tunewsThemeColor` and `textDecorationColor`.

### Side Effects

- Alot of files changed so...

### Testing

- Test everything has a color.
- Test contrast theme.
- Test aschaffenburg and malte.
- Test snackbar and pdfView because i didn't test these 😅 

### Resolved Issues

Fixes: #

---

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -47,6 +47,8 @@
     "@dr.pogodin/react-native-fs": "^2.36.1",
     "@dr.pogodin/react-native-static-server": "^0.25.0",
     "@dr.pogodin/react-native-webview": "^13.16.4",
+    "@formatjs/intl-displaynames": "^6.8.13",
+    "@formatjs/intl-locale": "^4.2.13",
     "@gorhom/bottom-sheet": "^5.2.6",
     "@maplibre/maplibre-react-native": "^10.4.0",
     "@notifee/react-native": "^9.1.8",
--- a/native/src/@types/styled.d.ts
+++ b/native/src/@types/styled.d.ts
@@ -45,6 +45,14 @@ type MD3Theme = {
     action: {
       disabled: string
     }
+    elevation?: {
+      level0: string
+      level1: string
+      level2: string
+      level3: string
+      level4: string
+      level5: string
+    }
   }
 }
 
--- a/native/src/components/SelectorItem.tsx
+++ b/native/src/components/SelectorItem.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react'
+import { DefaultTheme, List } from 'react-native-paper'
 import styled from 'styled-components/native'
 
 import dimensions from '../constants/dimensions'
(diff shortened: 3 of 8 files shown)


LeandraH (commented):
Wow, thanks for slogging through that, it looks really good already!

One small thing:

I don't think we should use `primary` for the `linkColor`, it might be the closest we have, but the function is very different, and if we one day change the `primary` color, the links should stay blue. I'd say we add a new color to the theme, if we need it at all.

And responding to the ones that you left open:

I agree about the fonts, we can do them in another ticket.

`theme.legacy.isContrastTheme` can be replaced with `theme.dark` (first prop [here](https://oss.callstack.com/react-native-paper/docs/guides/theming/#theme-properties))

`warningColor` is an awful color that we only use in the note in the feedback screen; in the web, we don't even have a background color there anymore, we just use red for the text color. I think we should do that here too.

`textDecorationColor` seems to be a random light grey, just choose one of the existing ones, maybe `surfaceVariant`? We can also change some of the colors in the theme if we need to have more grey variations there.

Not sure about `positiveHighlight` and `negativeHighlight`, we only use that in the opening hours of POIs, for the field saying if the place is currently open or closed. We could use `error` for the `negativeHighlight` and add a `success` color?

We should definitely add `tunewsThemeColor` as a custom color.

hauf-toni:
> > > > I don't think we should use `primary` for the `linkColor`, it might be the closest we have, but the function is very different, and if we one day change the `primary` color, the links should stay blue. I'd say we add a new color to the theme, if we need it at all.
> > > 
> > > 
> > > we already use primary for web links and `at AppointmentOnlyIcon.tsx` StyledLink. Screenshot 2025-12-04 at 6 29 19 PM
> > > I can add a new color...
> > 
> > 
> > How embarrassing 😅 I would prefer a new one anyway :)
> 
> I think it would be good to discuss a change like this with @hauf-toni before going for this. In my opinion, we should keep the links in color `primary`, there is no reason to distinguish between links and other primary navigation elements like buttons for users (and us) atm. Also, the thing that links have to be blue is not something that is still the case imo, links in other examples have varying colors, too and usually the primary color. Links being blue is more of a relic of old internet days tbh. I therefore really think we should keep this exactly the same way and NOT distinguish by introducing a new (and currently redundant) color. If we see the need in the future to have different colors for links and other primary elements, then would be the right time to make that change and separate the colors.

I agree with Steffen. Please keep the Links in primary 👍 
I appreciate your thoughts on this @LeandraH. Thanks for bringing awareness to the possible complications of a future primary color change.

f1sh1918 (approved):
Nice work 👍 
Tested on android

Issue #3733, opened 2025-11-20
CalendarChoiceModal doesn't respect SafeAreaView on iOS


LeandraH:
### Describe the Bug

In the branch with the new design system, the header of the CalendarChoiceModal is partially hidden behind my phone's header.

### Steps to Reproduce

1. Be on an iPhone?
2. Go to an event
3. Click on Export Event
4. Check on the header

### Expected Behavior

The header of the Modal with the back icon and the title should be below the system bar.

### Related Issues

Seen on #3666 

### Additional Information

Please someone check if this is also an issue in real Android devices.

<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/c0d68314-bff7-4ffe-90b5-53f5142a6a0c" />

PR #3761, opened 2025-12-08
3733: SafeArea inset top in modal

f1sh1918:
### Short Description

It looks like on ios (tested on 18.4) the SafeArea does not consider the top safe area.

### Proposed Changes

- add a padding top that with the `inset.top`
- add mocks in the test

### Side Effects

- i tested on android and ios and didn't find any

### Testing

- test event export (add to calendar), back button should work
- open poi filter locations, back button should work and not be placed in the status bar

### Resolved Issues

Fixes: #3733

---

### DIFF
--- a/native/src/components/__tests__/News.spec.tsx
+++ b/native/src/components/__tests__/News.spec.tsx
@@ -20,7 +20,7 @@ const news: [TunewsModel, TunewsModel] = [
   new TunewsModel({
     id: 9902,
     title: 'Was ist ein Verein?',
-    date: DateTime.fromISO('2020-01-20T00:00:00.000Z'),
+    lastUpdate: DateTime.fromISO('2020-01-20T00:00:00.000Z'),
     tags: [],
     content:
       'Ein Verein ist eine Gruppe von Menschen. Sie haben ein gemeinsames Interesse und organisieren. https://example.com',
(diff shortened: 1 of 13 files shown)


f1sh1918:
@bahaaTuffaha okay the misplacement is really random.
After like 2 hours i didn't find the issue why sometimes its correctly aligned and sometimes the padding seems to be applied twice. For me it looks like that sometimes the Modal with the SafeArea sets the padding correctly and sometimes not. I even thought scrolling could be an issue but its really random

I have really no clue how to debug this. Even didn't find any issue in the gh repo

native/src/components/Modal.tsx
```diff
@@ -41,32 +42,35 @@ const Modal = ({
...
+  const insets = useSafeAreaInsets()
+  return (
+    // View needs to stay until https://github.com/digitalfabrik/integreat-app/issues/3331 is done
+    <View>
+      <RNModal
+        visible={modalVisible}
+        onRequestClose={closeModal}
+        animationType='fade'
+        supportedOrientations={['portrait', 'landscape']}>
+        <Container paddingTop={insets.top}>
```
bahaaTuffaha: It turns out SafeAreaView has it's own space by default when you add padding top it will get doubled so you have to exclude the top.
```suggestion
        <Container paddingTop={insets.top} edges={['right', 'bottom', 'left']}>
```

bahaaTuffaha (approved):
Tested on Android and iOS ✅

PR #3767, opened 2025-12-11
3629: React native speech

bahaaTuffaha:
### Short Description

⚠️ Not sure if this PR will solve the problem as I cant reproduce it anymore on my phone.

TTS is not functioning well for android `TTS not ready` error message.

### Proposed Changes

- Replaced `react-native-tts` with `react-native-speech`

Note: This library already has a highlight function but didn't want to put more time into it.
https://github.com/mhpdev-com/react-native-speech/blob/main/docs/USAGE.md
### Side Effects

- I didn't use `useLoadAsync` for voices as it was... not sure if that could make issue?
- I used the resume functionality that wasn't exist at the other package.

### Testing

- Test Android and iOS.
- Test anything related to the tts.
- Stress testing it like keep pressing on the forward or backward, change screens/lang while playing.

### Resolved Issues

Fixes: #3629

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -804,7 +804,7 @@ jobs:
                 name: Create github release
                 working_directory: tools
             - run:
-                command: yarn app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files ~/sbom.json --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
+                command: yarn app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files ~/sbom.json --private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
                 name: Upload SBoM to github release
                 working_directory: tools
             - unless:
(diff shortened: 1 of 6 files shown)


native/src/components/TtsContainer.tsx
```diff
@@ -124,27 +143,49 @@ const TtsContainer = ({ children }: TtsContainerProps): ReactElement => {
...
       const sentence = sentences[safeIndex]
       if (sentence !== undefined) {
-        await stopPlayer()
-        setIsPlaying(true)
-        setSentenceIndex(safeIndex)
-        Tts.addEventListener('tts-finish', () => play(safeIndex + 1))
-        Tts.setDefaultLanguage(getGenericLanguageCode(languageCode))
-        Tts.speak(sentence, TTS_OPTIONS)
+        try {
+          if (!isPlaying && subscriptionsRef.current.length > 0 && index === sentenceIndex) {
```
f1sh1918: 🔧 for better readability i would set a boolean like
```
const canResume = !isPlaying && subscriptionsRef.current.length > 0 && index === sentenceIndex
```
I would also avoid early return but use proper `if...else` 
I think its clearer to say either you "can resume" or else you stop and play in this case

native/src/components/TtsContainer.tsx
```diff
@@ -1,28 +1,29 @@
...
-const TTS_OPTIONS: Options = {
-  androidParams: {
-    KEY_PARAM_PAN: 0,
-    KEY_PARAM_VOLUME: 0.6,
-    KEY_PARAM_STREAM: 'STREAM_MUSIC',
-  },
-  // This must not be 1 on iOS
-  rate: 0.5,
-} as Options
+type EventSubscription = {
```
f1sh1918: 🔧 I think you can also use the type from react native
```
import type { EventSubscription } from 'react-native'
```

Issue #3769, opened 2025-12-12
"Events" detail page loads slowly and shows blank screen


kalairubachandrakumaran-oss:
### Describe the Bug
When opening the Events detail page, the page does not load immediately.
- On iOS, the screen freezes for a short time before the content is shown.
- On Android, the screen stays white for a few seconds and then the content appears.
-> This looks like a performance problem while loading the Events detail page.

### Steps to Reproduce

1. Go to 'Integreat App'
2. Click on 'Events'
3. Click on one of the shown Event 
4. See how the page loads

### Expected Behavior
The Events page should load smoothly and quickly.

### Additional Information

https://github.com/user-attachments/assets/a6c70b21-2a03-4884-8c29-94c3ca422b91

PR #3773, opened 2025-12-16
3769: Update react-navigation stack to fix iOS freeze while navigating

bahaaTuffaha:
### Short Description

When opening the Events detail page, the page does not load immediately.
- On iOS, the screen freezes for a short time before the content is shown.
- On Android, the screen stays white for a few seconds and then the content appears.

### Proposed Changes

- Updated `react-navigation/stack` to `"^7.6.12"` that fixed the flickering.
- Moved the EventListItem slug into `useCallback` to memorize this slug and prevent recalculation.
~~- Added a loader at `RemoteContent.tsx` to show up while loading instead of nothing.~~

### Side Effects

- I hope none.

### Testing
- Test on iOS and Android.
- Go to Events.
- Click on one of the events to see if it freezes.

### Resolved Issues

Fixes: #3769

---


kalairubachandrakumaran-oss:
Tested ✅ 
https://github.com/user-attachments/assets/40bd667d-d7ff-4427-9387-41dc6d726cdc

Issue #3759, opened 2025-12-04
Investigate RTL on Android


LeandraH:
### Describe the Problem

~`react-native-paper` claims on the splash page that they support RTL. The only information in the documentation seems to be about turning around icons. I have found some RTL behavior when on an iOS emulator with a system language that is in RTL, but not in the same cases on an Android emulator.~

It looks like our app doesn't handle RTL on Android devices set to e.g. Arabic.

### Describe Your Preferred Solution

RTL is handled incorrectly in some components, see screenshot below.
Ideally, I would like most of the components to be flipped from left to right when an RTL system language is selected without having to add anything specifically for that in the component itself.

### Describe Possible Alternatives

Alternatively, I would like to know what to do so that the behavior on iOS and Android is the same, ideally by flipping the components when the system language is RTL.

### Related Issues

~Part of #3666~
Came up in https://github.com/digitalfabrik/integreat-app/pull/3752#discussion_r2586524142

### Additional Information

<img width="200" height="2400" alt="Image" src="https://github.com/user-attachments/assets/845cfcf4-99fd-4d6f-8c8e-191266704efd" />

steffenkleinle:
This is really weird, basically has to be connected to react-native-paper, right? We never had problems with this before and RTL worked just fine on Android, too. Perhaps related to changes in #3693. No clue though.

LeandraH:
> check the `MainApplication.kt‎` at [fedd57a#diff-23eb11952de45c346ba58af3887b3e14ccfae9bfb65dabef471e754af993cf40](https://github.com/digitalfabrik/integreat-app/commit/fedd57a6c79c4ca7a2d8943c9df561ca88012510#diff-23eb11952de45c346ba58af3887b3e14ccfae9bfb65dabef471e754af993cf40) the `ReactNativeHost` could be the issue here?

No, that one was merged into main on 30 October, and things were still working with the commits from 6 Nov. But it's a good idea! Anyway, I'll keep looking today 🤞

LeandraH:
Latest status here: The RTL was correct on my Android emulator set to Arabic both on main and on the redesign branch **after restarting Android Studio**.

However, @bahaaTuffaha has also reproduced the lack of swapping direction on his real Android when setting it to Arabic and restarting it. So all in all, I have no idea.

PR #3776, opened 2025-12-18
3759: Fix system RTL for android

bahaaTuffaha:
### Short Description

`I18nManager.isRTL` keeps returning false even if it's Arabic (system language).

### Proposed Changes

- at MainApplication I added `I18nUtil.getInstance().allowRTL(this, true)`
- And forced it if the language is rtl only `I18nUtil.getInstance().forceRTL(this, isRTL)`

Note: I tried to do the same way but inside the app with `I18nManager.forceRTL` but that caused problems in some pages and flipped back button.

### Side Effects

- This is will force direction on rtl languages...

### Testing

- Switch the system language to Arabic or Persian.
- check the app.
- Try to switch back to Deutsch or english (system lang).

### Resolved Issues

Fixes: #3759 

---


f1sh1918:
@bahaaTuffaha I don't think we have to adjust kotlin code in this case.
I would try to avoid that, since it not that obvious to take a look there if there are issues
The two functions you use are also available in `react-native` that can do the trick afaics.
We just have to check where is the best place to add them.

I just did a rough test with this:

For example here (contentDirection.ts)
```
export const isContentDirectionReversalRequired = (contentLanguage: string): boolean => {
  const isRTLLanguage = config.hasRTLScript(contentLanguage)
    I18nManager.forceRTL(isRTLLanguage)
    I18nManager.allowRTL(isRTLLanguage)
  return config.hasRTLScript(contentLanguage) !== I18nManager.isRTL
}
```

It just looks like app restart is required but i think thats not a big problem, since people are not likely to switch there system language very frequently ;)

bahaaTuffaha:
> @bahaaTuffaha I don't think we have to adjust kotlin code in this case. I would try to avoid that, since it not that obvious to take a look there if there are issues The two functions you use are also available in `react-native` that can do the trick afaics. We just have to check where is the best place to add them.
> 
> I just did a rough test with this:
> 
> For example here (contentDirection.ts)
> 
> ```
> export const isContentDirectionReversalRequired = (contentLanguage: string): boolean => {
>   const isRTLLanguage = config.hasRTLScript(contentLanguage)
>     I18nManager.forceRTL(isRTLLanguage)
>     I18nManager.allowRTL(isRTLLanguage)
>   return config.hasRTLScript(contentLanguage) !== I18nManager.isRTL
> }
> ```
> 
> It just looks like app restart is required but i think thats not a big problem, since people are not likely to switch there system language very frequently ;)

I agree more with doing it in react-native but I tried this solution it will mess up other things like the CityContentMenu will open on the other side and the read-aloud player icons will flip...
Screenshot 2026-01-12 at 4 46 51 PM

Issue #3471, opened 2025-08-14
Telephone number or postcode is read out as a large digit


kalairubachandrakumaran-oss:
### Describe the Bug

telephone number or postcode is read out as a large number (compare attached video)
hint: By phone number especially if it has “+49” in front of it

### Steps to Reproduce

1. Open the integreat app on your phone.
2. Open a page with phone numbers or postal codes, e.g., "Welcome to Augsburg/Contact the Augsburg App Team."
3. Start the read-aloud function.
4. Check how the phone number is read aloud.

### Expected Behavior

Digits of telephone number or postcode must be read out individually

### Additional Information

https://github.com/user-attachments/assets/ffbdbfb9-91dd-4754-8d10-d8edfa2ec0fb

PR #3777, opened 2025-12-19
3471: Format long numbers for tts

bahaaTuffaha:
### Short Description

telephone number or postcode is read out as a large number (compare attached video)
hint: By phone number especially if it has “+49” in front of it.

### Proposed Changes

- Adjusted the HTML parser to format numbers that has 5 or more digits (to only affect phone numbers and postal codes).
- Enabled `optimizeNumbersForTts` only for Tts web and native.

### Side Effects

- I hope none.

### Testing

- For native (iOS) go to Augsburg or Munich and search for contact and look for a contact-card or text.
- Check ReadAloud on these contacts phone numbers and postal codes.
- Check Android and web just in case.

### Resolved Issues

Fixes: #3471

---

### DIFF
--- a/web/src/components/EventListItem.tsx
+++ b/web/src/components/EventListItem.tsx
@@ -39,6 +39,7 @@ export const Icon = styled('img')(({ theme }) => ({
   marginTop: 8,
   width: 96,
   height: 96,
+  objectFit: 'contain',
 
   [theme.breakpoints.down('sm')]: {
     width: 64,


shared/utils/parseHTML.ts
```diff
@@ -1,6 +1,12 @@
 import { Parser, ParserOptions } from 'htmlparser2'
 
-const parseHTML = (html: string, options: ParserOptions = { decodeEntities: true }): string => {
+const formatLongNumbers = (text: string) => text.replace(/\d{5,}/g, match => match.split('').join(' '))
```
f1sh1918: 🔧 Always add a description to regex :)

f1sh1918 (approved):
Nice improvement!

PR #3780, opened 2026-01-12
3748: Fix for Cannot read property 'localeCompare'

bahaaTuffaha:
### Short Description
this issue is rising at sentry and it could be a fix for this message:
https://chat.tuerantuer.org/digitalfabrik/pl/kjqof9qq63n6znwj6cdecfi3uo

`C++ Exception: N8facebook3jsi7JSErrorE: ExceptionsManager.reportException raised an exception: Unhandled JS Exception: TypeError: Cannot read property 'localeCompare' of undefined`

### Proposed Changes

- Created `safeLocaleCompare` to safely handle localeCompare if one of the props is undefined or null.

### Side Effects

- none.

### Testing

- Couldn't reproduces it and I don't have iOS 26.2 yet.

### Resolved Issues

Fixes: #3748

---


kalairubachandrakumaran-oss:
Was also not able to reproduce on iphone 13 mini ℹ️

Issue #3670, opened 2025-10-30
Replace chip button component and chips in poi chips


steffenkleinle:
### Describe the Problem

Replace our custom chip button implementations with https://callstack.github.io/react-native-paper/docs/components/Chip.
Also replace our custom chips in `PoiChips`.

### Describe Your Preferred Solution

N/A

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #3781, opened 2026-01-13
3670: Use paper chip

f1sh1918:
### Short Description

Replace custom chips with react native paper chips

### Proposed Changes

- replace PoiChips on Poi Detail Page
- replace filter chips on map

### Side Effects

-

### Testing

- go to map and click on the filter button and add some filters
- check that the chips will be displayed and are removable
- check poi details that the chips will be displayed correctly

### Resolved Issues

Fixes: #3670 

---

### DIFF
--- a/native/src/routes/Pois.tsx
+++ b/native/src/routes/Pois.tsx
@@ -1,10 +1,10 @@
 import { BottomSheetFlatListMethods } from '@gorhom/bottom-sheet'
 import React, { ReactElement, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { useWindowDimensions } from 'react-native'
+import { StyleSheet, useWindowDimensions } from 'react-native'
 import { Chip } from 'react-native-paper'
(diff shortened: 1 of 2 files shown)


native/src/routes/Pois.tsx
```diff
@@ -111,28 +107,47 @@ const Pois = ({ pois: allPois, cityModel, route, navigation }: PoisProps): React
...
     <>
-      <ChipButton
-        text={t('adjustFilters')}
-        Icon={<StyledIcon Icon={EditLocationIcon} />}
-        onPress={() => setShowFilterSelection(true)}
-      />
+      <Chip
+        mode='outlined'
+        style={chipStyle}
+        rippleColor='transparent'
```
bahaaTuffaha: if we keep disabling the rippleEffect on each component we can just disable it globally.
https://oss.callstack.com/react-native-paper/docs/guides/ripple-effect/#disable-ripple-effect-in-all-components

Issue #3787, opened 2026-01-14
"Add to Calendar" button is not working on reoccurring events with one calendar


bahaaTuffaha:
### Describe the Bug
At events adding to calendar button doest work properly with single local calendar.
ex. at Munchen:
https://integreat.app/muenchen/en/events/meetings-for-girls-and-women-with-disabilities-or-chronic-illnesses

### Steps to Reproduce

1. Go to integreat => Munchen then events.
2. Make sure you have one calendar. you can add many in ios calendar settings or from an app for android: https://f-droid.org/en/packages/ws.xsoh.etar
3. Go to any event and click add to events. (nothing happends on android but ios it throws an error)

### Expected Behavior

- it should add the event and show a snackbar that says "added"
- for multiple calendars it should let you select a calendar before it add one.

PR #3788, opened 2026-01-14
3787: Fix export event button for recurrent and non-recurrent events and single calendar

bahaaTuffaha:
### Short Description

"Add to calendar" button doesn't work properly with single local calendar.

### Proposed Changes

- Reused the `chooseCalendar` function for the else condition (single calendar and non-recurring).
- Adjusted the conditions to let recurrent events and multiple calendars show the CalendarChoice screen whether for selecting a calendar or one/all-future-events or both.
- created a tests for these cases under the file `ExportEventButton.spec.tsx`

### Side Effects

- Hope none.

### Testing

- Change location to `testumgebung` (it has variety of events).
- Go to events and test recurrent event, non-recurrent events.
- Test with one calendar and also with multiple calendars. ( you can add them from calendar settings at ios and for android just install this app https://f-droid.org/en/packages/ws.xsoh.etar and go to the apps settings to add one).
- Test it on Android and iOS.

### Resolved Issues

Fixes: #3787 

---

### DIFF
--- a/native/android/app/src/main/java/tuerantuer/app/integreat/MainApplication.kt
+++ b/native/android/app/src/main/java/tuerantuer/app/integreat/MainApplication.kt
@@ -1,11 +1,14 @@
 package tuerantuer.app.integreat
 
 import android.app.Application
+import android.text.TextUtils
 import com.facebook.react.PackageList
 import com.facebook.react.ReactApplication
 import com.facebook.react.ReactHost
 import com.facebook.react.ReactNativeApplicationEntryPoint.loadReactNative
 import com.facebook.react.defaults.DefaultReactHost.getDefaultReactHost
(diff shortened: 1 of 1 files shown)


kalairubachandrakumaran-oss:
Tested on ioS and andioid , reak evice and browserstack
@bahaaTuffaha Question:
after allowing calendar access on iOS via BrowserStack, I was not able to add one event on the first attempt, but on the second attempt I was able to add the same event successfully.
Is this a bug?

iOS: https://github.com/user-attachments/assets/5b2b303d-18d0-44cc-9141-d95be07b8d5e
Android: https://github.com/user-attachments/assets/80cbd5e2-3684-42f9-ad5f-e57f57f5825a

Issue #3674, opened 2025-10-30
Replace pressable and text button components with button


steffenkleinle:
### Describe the Problem

Replace our custom pressable implementation with https://callstack.github.io/react-native-paper/docs/components/Button.

### Describe Your Preferred Solution

N/A

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

Issue #3795, opened 2026-01-19
Replace any list-item with List.item that we have from react-native-paper


bahaaTuffaha:
### Describe the Problem

we need to replace the item used in lists with ones from react-native-paper similar to the way we did in web but in web we used `ListItemButton` and `ListItemText`.
https://oss.callstack.com/react-native-paper/docs/components/List/ListItem

- category list
- news list
- events list
- poi list
- city list

### Additional Information
For reference look at web:
<img width="400" alt="Image" src="https://github.com/user-attachments/assets/d4612c5c-3202-46cb-b468-21811c476942" />

Issue #3331, opened 2025-06-13
Remove Modal workaround when React-Navigation updates


LeandraH:
### Describe the Problem

With the update to React-Navigation 7, the Modals from RN were not shown anymore in Android unless wrapped in a RN-View, there is a thread about it here: https://github.com/expo/expo/issues/32991#issuecomment-2489620459 (once there is an issue in the correct repo, I'll add that but they currently need a minimal reproducible example in order to open one, and I don't have the time to make one)

This workaround is applied in the CalendarRangeModal.tsx and our Modal.tsx

### Describe Your Preferred Solution

When the underlying issue is fixed, remove our workaround.

### Describe Possible Alternatives

Wait.

### Related Issues

The workaround was added in #3158

PR #3802, opened 2026-01-22
3674: Replace pressable ,text buttons, modals and used list items from react native paper

bahaaTuffaha:
### Short Description
- Replace custom pressable.
-  Pressable also used as backdrops for modals and dialogs so I had to remove that and use Modals and dialogs from the react native paper.(this will solve #3331)
- Pressable also used for list items everywhere so I had to replace and use `List.Item`

### Proposed Changes

- Replaced `Pressable` with `TouchableRipple` the closest to what we have at web.
- `TouchableRipple`needs prop `borderless` to let the ripple work.
- I tried to minimize styled-components as much as possible and used inline styling or `StyledSheets.Create`.
- List.Item (PaperList.Item) used it in the following components: `CategoryListItem, SubCategoryListItem, NewsListItem, EventListItem, PoiListItem, SearchListItem, CityEntry`
- For `CityGroup` I used `List.Subheader`.
- Migrated the `Modal` to React Native Paper Modal + I had to put the ThemeProvider outside of PaperProvider to let the modal work properly.
- Created `useNavigationTheme` to refactor and reuse it for the modal.
- Used `Dialog` component for the following: `CalendarRangeModal` and `AppointmentOnlyOverlay`.
- Adjusted styling on `CityNotCooperatingFooter` and `CitySelector` to match what we have at web.
- I enabled the date on Tu news like web.
- Adjusted the `List.tsx` to have `Divider` instead of adding it in the item except `BottomSheetFlatList` the divider didn't render properly so I created PoiListDivider.
- Removed 2 custom components: `Pressable` and `TextButton`.

### Side Effects

- Modal from react-native-paper lacks animation like fade that we were using.
- Couldn't replace the item at `SprungbrettListItem` as I can't find it.

### Testing

Test everything including (compare with web):
- Add to calendar Modal and buttons
- AppointmentOnlyOverlay (Dialog) - Testumgebung -> poi -> Place2Be New.
- CalendarRangeModal.
- Poi filter modal
- categories / subcategories
- Any list like Poi, citySelector, News/TuNews and events.

### Resolved Issues

Fixes: #3674 ,#3795,#3331

---

### DIFF
--- a/native/src/components/Accordion.tsx
+++ b/native/src/components/Accordion.tsx
@@ -20,7 +20,7 @@ const Accordion = ({
   return (
     <List.Accordion
       title={headerContent}
-      style={{ paddingHorizontal: 0, paddingVertical: 0, paddingRight: 0, paddingLeft: 0 }}
+      style={{ paddingVertical: 0, paddingRight: 0, paddingLeft: 0 }}
       contentStyle={{ paddingLeft: 0, paddingRight: 0 }}
       titleStyle={{ fontWeight: 'bold', color: theme.colors.onBackground }}
       expanded={!collapsed}
--- a/native/src/routes/CityNotCooperating.tsx
+++ b/native/src/routes/CityNotCooperating.tsx
@@ -1,7 +1,7 @@
 import Clipboard from '@react-native-clipboard/clipboard'
 import React, { ReactElement, useCallback, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { StyleSheet } from 'react-native'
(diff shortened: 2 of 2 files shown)


f1sh1918 (approved):
Just a small nitpick which might not be related to your changes

Image
This looks not aligned horizontally center

LeandraH (approved):
This looks great! Just two small comments:

The border in the CityNotCooperating screen is above the button
Image

Also, I don't think this is from this PR, but the header shows the menu button, should be fixed at some point :)
Image

Issue #3826, opened 2026-02-07
iOS/Android - TTS ~ Forward / Rewind breaks TTS playback


kalairubachandrakumaran-oss:
### Describe the Bug
  •    Playback freezes or becomes unstable
    •    Already read text is read again
    •    Wrong paragraph is read
    •    Sometimes the app becomes unresponsive

### Steps to Reproduce

1. Go to 'Integreat-app' on mobile device
2. Click on 'open a page with text' e.g München Landeshauptstadt -> Info-Aufenthalt -> Zentrale Ausländerbehörde (ZAB)
3. open the TTs player (settings -> Read aloud)
4. Tap play and let it start reading
5. Press forward or rewind fast multiple times
6. see error
### Expected Behavior

   •    Forward / Rewind should move to the correct position
    •    TTS should continue reading from the selected paragraph
    •    No freezing, no repeated or wrong text

### Additional Information
Test on Google Pixel, I phone 12 mini and browserstack Samsung s24

https://github.com/user-attachments/assets/6206c9c5-7807-4bf1-bdbf-3dbc7e54af1d

https://github.com/user-attachments/assets/aca1d0a7-1fc4-4e2f-9b18-3c6deb9ef131

https://github.com/user-attachments/assets/040e8086-a476-4f40-a4d9-bcf4fb1623df

Issue #3828, opened 2026-02-07
Android/iOS - TTS - After screen lock, playback restarts from the beginning


kalairubachandrakumaran-oss:
### Describe the Problem
  •    Playback starts again from the beginning
    •    It does not continue from the last position

### Describe Your Preferred Solution

 •    Playback should resume from the last spoken position
    •    TTS should continue where it was paused

PR #3831, opened 2026-02-09
3826: Fixed TTS forward button

bahaaTuffaha:
### Short Description

- Playback freezes or becomes unstable
- Already read text is read again
- Wrong paragraph is read
- Sometimes the app becomes unresponsive
- Playback starts again from the beginning when you lock screen.

### Proposed Changes

- Updated `react-native-speech` to https://github.com/mhpdev-com/react-native-speech/releases/tag/v1.4.0
- Fixed the breaking changes.
- Created `playRequestIdRef` to abort if another `play()` is called while stopping.
- 
### Side Effects

- None.

### Testing

check the testing on each ticket.

### Resolved Issues

Fixes: #3826 #3828

---

### DIFF
--- a/native/android/app/src/main/java/tuerantuer/app/integreat/MainApplication.kt
+++ b/native/android/app/src/main/java/tuerantuer/app/integreat/MainApplication.kt
@@ -1,14 +1,11 @@
 package tuerantuer.app.integreat
 
 import android.app.Application
-import android.text.TextUtils
 import com.facebook.react.PackageList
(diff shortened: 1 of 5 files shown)


bahaaTuffaha:
> For me the "Uhr" issue is not fixed. Still reproduced on android I also get on android API36, the language `de` is not supported from time to time 🙈 when i open tts After rebuilding app it works again. I dunno but not related to this pr

3666 design PR is still using the old tts so please reinstall packages here.

PR #3839, opened 2026-02-11
3784: [Android] Improve location permission request

### LINKED ISSUE #3784: Location button at the POI is not showing the Snackbar when there is no permission
### Describe the Bug

At Android 36.1 when the location permission is disabled the Snackbar popup should show up when you click the yellow location button at the POI.
- I tested 35 and it worked as excpected.

### Steps to Reproduce

1. Go to POI / MAP
2. Click on the yellow button for location.
3. if the location permission is disabled the popup snackbar should show up.

### Expected Behavior

The Snackbar should show up saying "Location access required".

bahaaTuffaha:
### Short Description
At Android 36.1 when the location permission is disabled the Snackbar popup should show up when you click the yellow location button at the POI.
But here the update on the ticket:
https://github.com/digitalfabrik/integreat-app/issues/3784#issuecomment-3883753237

### Proposed Changes

- To fix this permission lockstate/raceCondtion I used `requestMultiple` and `checkMultiple` on Android which creates a single request for both Fine and Coarse permissions.
- This will let the block permission to work as intended when you deny the permission the second time.
- when the permission gets denied twice it will go blocked state and it wont show the snackbar as expected so no need to change `showSnackbarIfBlocked`

### Side Effects

- Now it will ask for the approximate or precise location in one request instead of 2.

### Testing

- Test on android 36.1 , also older like 35.
- Test if it ask permission the first time (one request).
- Test the snackbar after one deny.
- Test after 2 denys (blocked state).
- Test on ios just in case.

### Resolved Issues

Fixes: #3784

---


steffenkleinle (approved):

steffenkleinle (changes requested):
Requesting permissions and showing the snackbar works as expected. However, clicking the location button does not do anything on my emulator, it is still marked as disabled and no location is found. This might be an emulator-only problem though, but I couldn't test on a real device as I have problems with gradle atm.

bahaaTuffaha:
@steffenkleinle this issue was solved but still in beta.
https://github.com/maplibre/maplibre-react-native/issues/740

so it doesn't show the puck properly on android with the new architecture.

steffenkleinle:
> @steffenkleinle this issue was solved but still in beta. [maplibre/maplibre-react-native#740](https://github.com/maplibre/maplibre-react-native/issues/740)
> 
> so it doesn't show the puck properly on android with the new architecture.

Sorry, but thats a different issue. It is about the location button doing nothing after allowing the permission in the settings and coming back to the app. See the video. We can also have a call about this next week.

bahaaTuffaha:
> > @steffenkleinle this issue was solved but still in beta. [maplibre/maplibre-react-native#740](https://github.com/maplibre/maplibre-react-native/issues/740)
> > so it doesn't show the puck properly on android with the new architecture.
> 
> Sorry, but thats a different issue. It is about the location button doing nothing after allowing the permission in the settings and coming back to the app. See the video. We can also have a call about this next week.

How this is a different issue? the puck didn't show up:

https://github.com/user-attachments/assets/788205df-33ea-4cdd-ab49-415c4421445a

steffenkleinle (approved):
I tested this on my real device and there the problem does not seem to exist. I'd therefore say this is good to go, nice work @bahaaTuffaha! I just simplified the function and used our existing hook for the app state listener :)

native/src/hooks/useUserLocation.ts
```diff
@@ -118,24 +114,17 @@ const useUserLocation = ({ requestPermissionInitially }: UseUserLocationProps):
...
-    const subscription = AppState.addEventListener('change', nextAppState => {
-      if (nextAppState === 'active') {
-        refreshPermissionAndLocation({
-          requestPermission: false,
-          showSnackbarIfBlocked: false,
-        }).catch(reportError)
-      }
-    })
-    return () => subscription.remove()
-  }, [refreshPermissionAndLocation])
```
bahaaTuffaha: Nice refactoring!

lunars97:
I tested it on my real android 16 device and it works as intended :+1: , but on iOS simulator there is a similar problem: the location marker does not react after I went to settings through snackbar and enabled the location and went back to the map, it is not reacting at all :grimacing: Also tested on main and seems like it has been an issue for awhile :thinking:

lunars97 (approved):
i think we can create another issue for iOS. Thank you for keeping up with this  👍

steffenkleinle:
> i think we can create another issue for iOS. Thank you for keeping up with this 👍

Hmm, since its a problem on both iOS simulator and Android emulator but works on real devices, I think we can just ignore this problem. Also I think its a very problem that basically no users will ever encounter.

steffenkleinle:
@bahaaTuffaha lets merge this, nice work! 🚀

lunars97:
> > i think we can create another issue for iOS. Thank you for keeping up with this 👍
> 
> Hmm, since its a problem on both iOS simulator and Android emulator but works on real devices, I think we can just ignore this problem. Also I think its a very problem that basically no users will ever encounter.

if it only meets emulators, then no worries :+1:

Docmale:
I'm only having problems on the emulators too, but my real device(android16) works as expected:

PR #3849, opened 2026-02-18
Fix accessibility issues

lunars97:
### Short Description
This PR fixes some accessibility issues throughout the native android. 

These issues are fixed:

- Language selection is not responding and it is being said that it is disabled on touch with TalkBack
- TextInputs in Feedback form in Malte is not being focused and edited

### Proposed Changes

- Add necessary props to TouchableRipple component
- Remove accessible prop from Container which wraps WebView to enable focus on inputs

### Side Effects

- should be none

### Testing

1. Enable TalkBack and go to language selection
2. Try to go through the list of items and hear that they are announced to be disabled
3. Check MalteForm in Category **Praise and Criticism** at Bad Doberan and try to go through TextInputs, see that it is not possible to focus and edit them.
4. Also check on iOS :pray: 

### Resolved Issues

Fixes: #3843 

---


bahaaTuffaha (commented):
- There is other areas that needs to tell what is the button is about like:
    - Location button.
    - 3 dots headerMenu.
- I can navigate at headerMenu using talkback.

https://github.com/user-attachments/assets/c0048ec5-c1f4-45ea-a5ce-bad918a4af84

native/src/components/HeaderMenu.tsx
```diff
@@ -139,7 +140,8 @@ const HeaderMenu = ({
         <IconButton
           icon='dots-vertical'
           iconColor={theme.colors.onSurface}
-          onPress={() => setVisible(true)}
+          onPress={() => setVisible(!visible)}
```
bahaaTuffaha: ❓  Not sure if this `setVisible(!visible)` needed because you cant touch the button unless you close the menu.

PR #3850, opened 2026-02-18
Add maintenance label to Dependabot PRs

bahaaTuffaha:
### Proposed Changes

-

### Side Effects

-

### Resolved Issues

Fixes: #

---


f1sh1918 (commented):
I think we also need "dependencies" but not sure

Issue #3877, opened 2026-02-20
Returning from the map area is not possible.


Docmale:
### Describe the Bug

If I jump from from a contact in area 'Information' to a location in area 'Map' it's not possible to jump back to my starting point. Instead I'm stuck in 'Map' and have to go  all the way back to my original contact via 'Information' base page. This is tue for web and native.

### Steps to Reproduce

1. Select Testumgebung
2. Select Information area
3. Set language 'Deutsch'
4. Klick tile Sprache
5. Klick address 'Lange Reihe 77 ...''
6. Klick upper left back button or sys-back
7. You are stick to area Map
### Expected Behavior

As in previous versions, I want to jump back to my starting point.

bahaaTuffaha:
Each Tab in bottom nav has its own history/stack so the navigation differs from what it was before...

bahaaTuffaha:
> [@bahaaTuffaha](https://github.com/bahaaTuffaha) does it mean that it is intended to be like this?

Yes its intended ... let me check if I can change the back behaviour though...

PR #3893, opened 2026-02-26
3877: Enable back button for tab navigation history

bahaaTuffaha:
### Short Description

If I jump from from a contact in area 'Information' to a location in area 'Map' it's not possible to jump back to my starting point. 

### Proposed Changes

- Added `backBehavior= 'history'` prop for `<Tab.Navigator>` so it let us return to the previous route.
- Created `hasTabHistory` to check the Tab history.
- Updated `canGoBack` with this new `hasTabHistory`.

### Side Effects

- This affects the back-button navigation logic.

### Testing

- Switch between tabs then go back to the main tab (information) try to open sub categories and events.
- Test the logic after using the search.
- Test the physical back-button.

### Resolved Issues

Fixes: #3877 

---


lunars97 (approved):
LGTM 💛

Issue #3352, opened 2025-06-30
Inversion of svg icons on contrast theme on native


lunars97:
### Describe the Problem
Once we set all the icons to svg (currently we have not replaced the PNGs to SVGs yet), we need to invert them on first level and second level to align with contrast theme like it is done on the web. There is a problem on react native with inversion of icon colors, 
### Describe Your Preferred Solution

So far I found one solution to use https://www.npmjs.com/package/react-native-color-matrix-image-filters

### Describe Possible Alternatives

Should be done on cms side to replace color with fill="currentColor" on SVGs path

### Related Issues

#1021

osmers:
The related CMS issue is done - I'll remove the blocked label. In case it's not completely done, please re-add the label.

PR #3902, opened 2026-03-02
3352: Inversion of category icons

bahaaTuffaha:
### Short Description

we need to invert category icons/tiles color to match the contrast we have at web.

### Proposed Changes

- Used a library https://www.npmjs.com/package/react-native-color-matrix-image-filters
- Created a new component called `ContrastImage` and tried to match what we have at web invert, saturate and brightness.
### Side Effects

- None.

### Testing

- Test categories and level one categoryListItems while contrast mode enabled.
- Test different regions for verity on icons.
- Test on iOS (I didn't test on an old iOS versions).

### Resolved Issues

Fixes: #3352

---


kalairubachandrakumaran-oss:
Works as expected ✅

Issue #3903, opened 2026-03-03
Map ~ Last selected filter chip is truncated on smaller screen sizes


kalairubachandrakumaran-oss:
### Describe the Problem
When multiple filters are selected (e.g., “Open now” and a category such as “Authority/Office”), the last filter chip in the top filter bar is partially cut off on smaller displays. The text is not fully visible and appears truncated at the right edge of the screen

### Describe Your Preferred Solution
All selected filter chips should be fully visible

### Additional Information

<img width="363" height="1018" alt="Image" src="https://github.com/user-attachments/assets/745f1144-8b1c-487f-966b-c89165e975cd" />

PR #3907, opened 2026-03-03
3903: Fix filter chip's flex-wrap on small devices

bahaaTuffaha:
### Short Description

When multiple filters are selected (e.g., “Open now” and a category such as “Authority/Office”), the last filter chip in the top filter bar is partially cut off on smaller displays. The text is not fully visible and appears truncated at the right edge of the screen

### Proposed Changes

- For `OverlayContainer` I removed flex:1 and used flex-wrap to let the chips wrap plus gave right 8px like left.
- The button at `PoiFiltersModal` is under the safe area on the small screen so I added a safe area and marginBottom.

### Side Effects

- None.

### Testing

- Test poi filter on small screens (small phone - emulator).
- Test PoiFiltersModal's button (`show {count} places`) on small screens.

### Resolved Issues

Fixes: #3903 

---

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -59,8 +59,7 @@ const StyledIcon = styled(IconButton)<{ position: number | string }>`
 `
 
 const OverlayContainer = styled.View`
-  flex-direction: row;
-  flex-wrap: wrap;
+  flex-flow: row wrap;
   position: absolute;
   top: 24px;
   left: 8px;


native/src/components/MapView.tsx
```diff
@@ -59,11 +59,13 @@ const StyledIcon = styled(IconButton)<{ position: number | string }>`
 `
 
 const OverlayContainer = styled.View`
-  flex: 1;
   flex-direction: row;
+  flex-wrap: wrap;
```
lunars97: L62 and L63 can be shortend with `flex-flow: row wrap;` and the stylelint error will be fixed

lunars97 (approved):
LGTM! Thanks for fixing the bottom issue as well 💛 Tested with small phone

native/src/components/MapView.tsx
bahaaTuffaha: Thanks!

Issue #3901, opened 2026-03-02
Animate Frag Integreat "waiting for response"


svenseeberg:
### Describe the Problem

The static typing indicator seems to confuse some persons and makes them wonder if any answer is coming at all.

### Describe Your Preferred Solution

To make it even more clear that this is a typing indicator, we could animate the content of the message with moving dots. The following string sequence could be used with a change every ~0,5s:

```
...
·..
.·.
..·
```

Alternatively, instead of showing 3 dots, we can show a message "Generating answer" or "Searching for answers".

This can obviously also be combined to
```
Generating answer ...
Generating answer ·..
Generating answer .·.
Generating answer ..·
```

### Describe Possible Alternatives

Spinner, etc

### Related Issues

https://github.com/digitalfabrik/integreat-chat/issues/388

PR #3918, opened 2026-03-09
3901: Animate waiting-response for frag integreat

bahaaTuffaha:
### Short Description

The static typing indicator seems to confuse some persons and makes them wonder if any answer is coming at all.

### Proposed Changes

- ~~Moved TypingIndicator into new file.~~
- ~~It uses useEffect to start an interval when `isVisible` is True.~~
- ~~the interval will increase the index and % `dotPattern.length` so it will do the following:~~
~~`1%4 , 2%4 , 3%4 , 4%4` it corresponds to `1 , 2 , 3 , 0 `so the loop keeps going every 400 ms.~~
- ~~Added `generateAnswer` to translations for `de` and `en`.~~
- Used a styled span dot instead of text.
- `DotsWrapper` will animate the dots with keyframes.

### Side Effects

- None.

### Testing

- Go to testumgebung and open the chat.
- Type anything.
- While you wait for the response. A loading animation will appear besides "Generating answer".

### Resolved Issues

Fixes: #3901

---

### DIFF
--- a/translations/translations.json
+++ b/translations/translations.json
@@ -318,7 +318,7 @@
       "sendButton": "Senden",
       "user": "Du",
       "privacyPolicyInformation": "Bitte akzeptiere die Datenschutzerklärung, um den Chat zu nutzen.",
-      "generateAnswer": "Antwort generieren"
+      "generateAnswer": "Schreibt..."
     },
     "am": {
       "bot": "የውይይት ቦት",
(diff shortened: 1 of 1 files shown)


bahaaTuffaha:
> Nice work and thanks for taking care of this. I think using just a text-base typing indicator does not really fit the style of the rest of the app. What do you think of something like the following instead:


Thanks ... this looks good already and better than useEffect.

Issue #3732, opened 2025-11-20
RangeError: Maximum call stack size exceeded


steffenkleinle:
### Describe the Bug

We have quite a few unhandled crashes in sentry with `RangeError: Maximum call stack size exceeded`. Investigate and try to come up with a fix.

### Steps to Reproduce

Sadly no idea how to reproduce this. Seems to only happen on iOS (Chrome Mobile).
Sentry issue: https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/8314

### Expected Behavior

No crash.

### Related Issues

N/A

### Additional Information

https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/8314

bahaaTuffaha:
Could it be the circular dependencies? 🤔 
`npx madge --circular --extensions ts,tsx .`

```bash
✖ Found 39 circular dependencies!

1) shared/api/index.ts > shared/api/endpoints/createCategoriesEndpoint.ts > shared/api/mapping/mapCategoryJson.ts > shared/api/mapping/mapAvailableLanguages.ts
2) shared/api/models/CategoryModel.ts > shared/api/models/ExtendedPageModel.ts > shared/api/models/PageModel.ts > shared/utils/index.ts
3) shared/api/index.ts > shared/api/endpoints/createCategoryChildrenEndpoint.ts
4) shared/api/index.ts > shared/api/endpoints/createCategoryContentEndpoint.ts
5) shared/api/index.ts > shared/api/endpoints/createCitiesEndpoint.ts
6) shared/api/index.ts > shared/api/endpoints/createCityEndpoint.ts
7) shared/api/index.ts > shared/api/endpoints/createEventsEndpoint.ts > shared/api/models/DateModel.ts > shared/utils/date.ts
8) shared/api/models/DateModel.ts > shared/utils/date.ts
9) shared/api/index.ts > shared/api/endpoints/createPOIsEndpoint.ts > shared/api/models/PoiModel.ts > shared/utils/isCurrentlyOpen.ts
10) shared/index.ts > shared/routes/pathname.ts
11) native/src/utils/DatabaseConnector.ts > native/src/utils/helpers.ts
12) native/src/contexts/AppContextProvider.tsx > native/src/utils/PushNotificationsManager.ts
13) native/src/contexts/AppContextProvider.tsx > native/src/utils/PushNotificationsManager.ts > native/src/hooks/useCityAppContext.ts
[...]

PR #3921, opened 2026-03-10
3732: Fix mutation at categoriesPage

bahaaTuffaha:
### Short Description

We have quite a few unhandled crashes in sentry with RangeError: Maximum call stack size exceeded. Investigate and try to come up with a fix.

From sentry's breadcrumbs I found out that this could be coming from the info page:
```
from: /muenchen/de,
to: /muenchen/de/arbeit-ausbildung
```

- found a mutation at `categoriesPage.tsx`:
`categories` are coming from `useLoadFromEndpoint` and it uses `useLoadAsync` and it has `data` from a state. meaning pushing to that array will mutate and could cause issues.

### Proposed Changes

- Fixed the mutation by spread syntax.

### Side Effects

- None

### Testing

N/A (can't be reproduced locally).

### Resolved Issues

Fixes: #3732 

---


web/src/routes/CategoriesPage.tsx
```diff
@@ -67,6 +73,30 @@ const useCategoryData = (cityCode: string, languageCode: string, pathname: strin
 
   const { data: parents, loading: parentsLoading, error: parentsError } = useLoadAsync(requestParents)
 
+  // The root category is not delivered via our endpoints
+  const categories = useMemo(
```
bahaaTuffaha: I had to use useMemo to fix this:
```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate.
```

kalairubachandrakumaran-oss:
Not testable for Testing Team❗

PR #3922, opened 2026-03-10
3896: Fix filters chip flickering

### LINKED ISSUE #3896: Map ~ Filter and selected filter name disappear briefly when swiping the list
### Describe the Problem
The filter button and the selected filter name briefly disappear when swiping down from the fully expanded list.

The filter button and the selected filter name should always remain visible and should not disappear while swiping.
-> Similar to Web App

### Additional Information

https://github.com/user-attachments/assets/5a57acb0-f046-4b08-ae88-90a106be71bf

lunars97:
### Short Description

This PR fixes flickering of the filters chip in maps when the bottom sheet expands and contracts to previous state. Also added `accessible` prop to OpeningHoursListitem.ts to wrap weekdays with hours for screenreader, without it they are read separetely

### Proposed Changes

- [x] remove unmounting condition which was before used to hide the children from the screen reader focusing on them when the bottom sheet expanded. Always keep them mounted.
- [x]  fix accessibility issue when using keyboard with screen reader available by hiding it from screen reader when the modal appears and the bottom sheet expands 
- [x] add accessible prop to weekdays and hours container

### Side Effects

- I hope none

### Testing

1. Go to the maps
2. Expand and contract the bottom sheet
3. See that the filters chip is not flickering (try with additional chips)
4. Also check with screen reader also using the keyboard whether it reads out the children in MapView when the filters modal appear or when the bottom sheet expands

### Resolved Issues

Fixes: #3896 

---

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -32,9 +32,15 @@ import {
 
 import { clusterCountLayer, clusterLayer, markerLayer } from '../constants/layers'
 import useUserLocation from '../hooks/useUserLocation'
+import { conditionalA11yProps } from '../utils/helpers'
 import MapAttribution from './MapsAttribution'
 import Icon from './base/Icon'
 import IconButton from './base/IconButton'
+
+const OuterWrapper = styled.View`
+  flex: 1;
+  position: relative;
+`
 
 const MapContainer = styled.View`
   flex: 1;
@@ -46,7 +52,9 @@ const StyledMap = styled(MapLibreMapView)`
   width: 100%;
 `
 
-const StyledIcon = styled(IconButton)<{ position: number | string }>`
+const StyledIcon = styled(IconButton)<{
+  position: number | string
+}>`
   position: absolute;
   right: 0;
   bottom: ${props => props.position}${props => (typeof props.position === 'number' ? 'px' : '')};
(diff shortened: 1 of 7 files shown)


native/src/components/MapView.tsx
```diff
@@ -217,23 +223,30 @@ const MapView = ({
...
+          accessibilityElementsHidden={bottomSheetFullscreen || modalVisible}>
+          {Overlay}
+        </OverlayContainer>
+        <MapAttribution
+          importantForAccessibility={bottomSheetFullscreen || modalVisible ? 'no-hide-descendants' : 'auto'}
+          accessibilityElementsHidden={bottomSheetFullscreen || modalVisible}
+        />
+        <StyledIcon
+          importantForAccessibility={bottomSheetFullscreen || modalVisible ? 'no-hide-descendants' : 'auto'}
+          accessibilityElementsHidden={bottomSheetFullscreen || modalVisible}
```
steffenkleinle: 🔧 Can we extract those to shared constants to make this cleaner and remove duplication?

steffenkleinle (commented):
Nice work and nice solution!

native/src/routes/Pois.tsx
```diff
@@ -167,6 +167,7 @@ const Pois = ({ pois: allPois, cityModel, route, navigation }: PoisProps): React
         poisCount={pois.length}
       />
       <MapView
+        modalVisible={showFilterSelection}
```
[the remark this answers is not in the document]
lunars97: the filter chip and copyright in mapview become inaccessible this way

steffenkleinle (approved):
Tested on android, works as expected, thanks!

bahaaTuffaha (changes requested):
I can't go back on iOS 26.2:

Image
Image

native/src/components/Modal.tsx
```diff
@@ -47,13 +47,9 @@ const Modal = ({
...
   return (
-    <Portal>
-      <NavigationThemeProvider value={navigationTheme}>
-        <PaperModal
-          visible={modalVisible}
-          onDismiss={closeModal}
-          style={styles.modalStyle}
-          contentContainerStyle={styles.modalStyle}>
+    <RNModal visible={modalVisible} transparent onRequestClose={closeModal} style={styles.modalStyle}>
+      <SafeAreaView style={{ flex: 1 }} edges={['top']}>
```
bahaaTuffaha: This worked the at first then when I changed tabs then went back to the filter the safe area got broken again.

I think we had that issue before at https://github.com/digitalfabrik/integreat-app/pull/3761
Just use `useSafeAreaInsets()` and add a paddingTop to the view instead until react-native fixes the issue on ios.

bahaaTuffaha (approved):
Great work!
Tested on iOS and Android. ✅ 

one unrelated issue:
I can't select the filter buttons on iOS while the accessibility inspector is on:

Image

lunars97:
> Great work! Tested on iOS and Android. ✅
> 
> one unrelated issue: I can't select the filter buttons on iOS while the accessibility inspector is on:
> 
> Image

thank you 💚 Hmm regarding the button not reacting on onClick when accessibility inspector is enabled, for me it is working as intended on iPhone 17 Pro simulator... 🤔 

UPDATE: now I got you. Yes, seems like MapView consumes all the focus and it is read as a whole group. I will fix it here in this PR 👍

bahaaTuffaha (approved):
😃 I can select the filters now with the accessibility inspector on iOS as expected ✅ 
Thanks!! 
can you highlight what you did to make it accessable?

lunars97:
> 😃 I can select the filters now with the accessibility inspector on iOS as expected ✅ Thanks!! can you highlight what you did to make it accessable?

I took the elements above the map container outside of the MapContainer, MapContainer was consuming the focus

lunars97:
@steffenkleinle do you want to give it another round of review due to some changes after your last review? :)

steffenkleinle (approved):
Code changes look good to me, haven't tested again

Issue #3932, opened 2026-03-12
Changing location not possible


Docmale:
### Describe the Bug

The current location or municipality is displayed as a button at the top of the browser; when I click this button, the change location page opens. This is only true for the information area, other ereas have no chance to change the location. When I now switch to map or another area and than switch back to information area the possibility to click the location button is gone and I have no chance to set a new location until I delete Integreat from recent-tasks list and start from scratch.

### Steps to Reproduce

as described:

1. Start Integreat you are on Information area
2. Try location change -> works fine
3. Switch to another area
4. Switch back to Information area 
5. Try location change -> nothing happens

### Expected Behavior

All areas have the option to change location.

### Additional Information

After switching back to Information area:

<img width="396" height="316" alt="Image" src="https://github.com/user-attachments/assets/6477e4ba-48a4-47c0-8f4c-2a4e44bfb5c7" />

Docmale:
This bug was introduced in the very latest version 2026.3.1

PR #3935, opened 2026-03-13
3932: Fix location change button disappearing after tab switch

dkehne:
### Short Description

Fixes the location change button disappearing from the header after switching tabs. Tab history no longer affects the visibility of the location button.

### Proposed Changes

- Separated tab history (`hasTabHistory`) from stack navigation in the `landingPath` check
- `landingPath` now only hides the location button when there is actual stack navigation (`previousRoute`) or a root-level overlay (`hasRootHistory`)

### Side Effects

-

### Testing

- Open the app on the Information (Categories) tab — location change button is visible
- Switch to another tab (e.g. Map/POIs), then switch back to Information — location change button is still visible and tappable
- Navigate into a category detail page — location button is not shown (back button takes its place)
- Regression test covers the tab history case

### Resolved Issues

Fixes: #3932

### DIFF
--- a/native/src/components/__tests__/Header.spec.tsx
+++ b/native/src/components/__tests__/Header.spec.tsx
@@ -151,21 +151,10 @@ describe('Header', () => {
 
   it('should show location change button even when tab history exists', () => {
     mockPreviousRoute(false)
-    const tabNavigationMock = {
-      getState: jest.fn(() => ({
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):
Tested on android emulator, works as expected

PR #3936, opened 2026-03-13
3925: Fix focus outline border-radius in contrast mode

### LINKED ISSUE #3925: Focus outline radius differs from icon container radius in contrast mode (web app)
### Describe the Bug

In the web app in contrast mode, the focus outline (yellow) has a different border radius than the icon container.
This creates a visual inconsistency where the outline does not follow the shape of the icon card.

### Steps to Reproduce

1. Open the web app
2. Enable contrast mode
3. Navigate to the category tiles
4. Focus the tile

### Expected Behavior

The focus outline should have the same border radius as the icon container, so that the outline follows the exact same shape.

### Additional Information

Image
Image

dkehne:
### Short Description

Fixes mismatched focus outline border-radius on category tiles in high contrast mode. The focus ring now has the same rounded corners as the icon container on hover.

### Proposed Changes

- Combined `:hover` and `*:focus-visible > &` on `Outline` in `Tile.tsx` so the rounded outline also appears on keyboard focus
- Added `&:focus-visible { outline: none }` on `StyledButton` in contrast mode to suppress the duplicate default focus ring

### Side Effects

-

### Testing

- Open web app in contrast mode
- Tab to a category tile — focus outline around the icon should have the same rounded corners (`border-radius: 24`) as the hover outline
- Mouse hover still works correctly

### Resolved Issues

Fixes: #3925


bahaaTuffaha (changes requested):
Claude missed what the issue is about. The outline can be solved by adding the following to the Outline styling :
```
  borderRadius: 24,
  overflow: 'hidden',
```

steffenkleinle (commented):
Thank you for the PR, looks nearly perfect. It seems like there still is a small visual artifact though, a small black line/border between the icon and the yellow outline:

Image

bahaaTuffaha (approved):
Thanks for your contribution!
- I prefer the `outlineOffset` solution as I mentioned earlier but if you find it better we can go with it as it's very small artifact.

steffenkleinle (approved):
Thanks for this PR! I don't mind which solution we use.

PR #3941, opened 2026-03-15
Remove react-helmet

michael-markl:
### Short Description

Since React 19, [React ships with built-in support for rendering meta tags](https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags); react-helmet is no longer needed (and no longer maintained since 6 years). This removes the react-helmet dependency.

### Proposed Changes

See above.

I also noticed that in some cases, multiple instances of  Helmet were rendered,  resulting in ambiguous behavior.
I thus added a FailureSwitcherWithHelmet to avoid always rendering Helmet for every FailureSwitcher.

### Side Effects

* html title, meta tags (description), social media previews, etc.

### Testing

Check html title changes when navigating.

### Resolved Issues

n/a


michael-markl:
@kalairubachandrakumaran-oss 
I don't think this is a bug at all. It's probably just a bad slug/URL for this particular page (likely due to renaming / duplicating a page instead of creating a new one in the CMS). Since it's just the testumgebung, the content (which the URL is a part of) is not relevant...

https://integreat.app/testumgebung/de/gesellschaftliches-leben/alltag

Edit: We could check whether the same thing happened in a "production" instance as well... Do we know the "base" instance of the testumgebung?

kalairubachandrakumaran-oss:
> @kalairubachandrakumaran-oss I don't think this is a bug at all. It's probably just a bad slug/URL for this particular page (likely due to renaming / duplicating a page instead of creating a new one in the CMS). Since it's just the testumgebung, the content (which the URL is a part of) is not relevant...
> 
> https://integreat.app/testumgebung/de/gesellschaftliches-leben/alltag
> 
> Edit: We could check whether the same thing happened in a "production" instance as well... Do we know the "base" instance of the testumgebung?

>In production there is one example where this inconsistency also appears, but for most pages it looks correct

> image

Issue #3897, opened 2026-03-02
Map ~ List can be dragged above the header in Web App


kalairubachandrakumaran-oss:
### Describe the Problem
The list can be dragged above the header.
The header/logo disappears, and it is no longer clear which app is being used.

### Describe Your Preferred Solution
Dragging should stop at the header (like in the native app).
The header/logo should always remain visible and the list should not move above it.

### Additional Information

https://github.com/user-attachments/assets/d84a1ff7-eecf-40ae-abd6-9d74b2cb07e8

PR #3946, opened 2026-03-18
3897: Adjust Max snapPoint for bottomSheet

bahaaTuffaha:
### Short Description

The list can be dragged above the header.
The header/logo disappears, and it is no longer clear which app is being used.
Dragging should stop at the header (like in the native app).

### Proposed Changes

- Changed the max snapPoint to `height - headerHeight`.
- Adjusted the mock at `useDimensions.ts`.
 
### Side Effects

- None

### Testing

- Launch the Web-app + (small screen device).
- Go to poi/maps
- Pull the bottomSheet to the max (it should reach the header).

### Resolved Issues

Fixes: #3897

---

### DIFF
--- a/native/jest.config.ts
+++ b/native/jest.config.ts
@@ -1,3 +1,4 @@
+/** @jest-config-loader ts-node */
 const transformNodeModules = [
   'react-native',
   '@react-native',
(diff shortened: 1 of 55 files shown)


bahaaTuffaha:
We could use `expandOnContentDrag` prop for the StyledBottomSheet so it expands the sheet as we scroll like native. but in this case it will cover the filter-button from first interaction unless we adjust the max snap to the point before the filter button.

kalairubachandrakumaran-oss:
tested ✅ 

https://github.com/user-attachments/assets/3e793592-20fe-415a-8ee8-af768258cfc0

Issue #3919, opened 2026-03-10
Returning from locations in the map does not work


Docmale:
### Describe the Bug

When jumping from outside the map-area to a location on the map, it's not possible to return to the starting point.  e.g.: From the Information- or event-area it is possible to directly jump to a specific location on the map. When I now (after I looked at the place on the map) klick the backbutton to return to my event, or information tile, the app stucks in the map-area and I have to search for my original event or information again. 

### Steps to Reproduce

Reproce this behaviour with any address with a link into a specific location on the map, e.g.:

1. Go to Testumgebung, deutsch
2. Goto [Veranstaltungen - testserie2](https://webnext.integreat.app/testumgebung/de/events/testserie1)
3. Click on address [LR, Lange Reihe 77, 27211 Bassum](https://beta.integreat.app/testumgebung/de/locations/lr/)
4. Look at address on map
5. Click backbutton
6. See list of poi's in Testumgebung (not your starting point)

### Expected Behavior

My expectation is that I jump directly to my original information or event where I started my jump to the map.

### Related Issues

https://github.com/digitalfabrik/integreat-app/issues/3877

### Additional Information

Following recording illustrates the jump from atile to the map and the way back to the landing page from where I have to search my original information:
https://github.com/user-attachments/assets/911b4d71-38ca-463f-944d-ae821a646646

Issue #3914, opened 2026-03-05
Fix initial screen after going back from multiple tabs


bahaaTuffaha:
### Describe the Bug

if you navigated to the following:
info > news > events > info > events > then keep going back to last screen by the backbutton. it will get back to news as the first page (because we went to info in the middle of history).

### Steps to Reproduce

1. from info > news > events > info > events > then keep going back to last screen by the backbutton.
2. news not is the last page instead of info.

### Expected Behavior

It should get back to info tab instead.

PR #3948, opened 2026-03-20
3919: Refactor navigation to fix map back navigation

steffenkleinle:
### Short Description

This PR refactors the internal navigation logic to tackle several problems.

### Proposed Changes

- Replace custom reset navigation action with default (nested) navigation. This fixes the back navigation to be as expected (stack-like) (#3919), fixes navigating back to the initial screen after complex navigation (#3914) and fixes that pressing a tab navigates back to the tab root.
- Remove special handling in deep links if the intro was not shown yet
- Use correct typing and remove unnecessary param in PushNotificationHandler

### Side Effects

- If a deep link is opened before the intro slides are viewed, the deep link now opens directly temporarily dismissing the intro slides. The intro slides will then show up on the next app start. This gets rid of some complicated and error prone logic and is imo more user friendly.

### Testing

Thoroughly test navigating inside the app, see also the issue description.
Thoroughly test deep linking on ios and android.

### Resolved Issues

Fixes: #3919, #3914

---

### DIFF
--- a/native/src/hooks/__tests__/useNavigate.spec.tsx
+++ b/native/src/hooks/__tests__/useNavigate.spec.tsx
@@ -37,7 +37,7 @@ jest.useFakeTimers()
 
 describe('useNavigate', () => {
   const { mocked } = jest
-  const navigation = createNavigationPropMock()
+  const navigation = { ...createNavigationPropMock(), canGoBack: () => true }
   mocked(useNavigation).mockImplementation(() => navigation as never)
 
   const cityCode = 'ansbach'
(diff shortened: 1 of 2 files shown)


steffenkleinle:
> @steffenkleinle While using Sys-Back at the system navigation bar I jump directly back to the starting point (Events-testserie2) I have to click the twice the Back-button at the upper left corner to jump back to Events-testserie2 (first klick takes me to the list of pois). Is this behaviour intended?

Not necessarily intended but a different issue and sadly not very easy to fix. So for the time being, this is an accepted drawback.

PR #3954, opened 2026-03-23
2966: Upgrade yarn to v4 (berry)

steffenkleinle:
### Short Description

This PR finally upgrades yarn to v4 (berry).

### Proposed Changes

- Use corepack to install the correct yarn version
- Use `yarn patch` instead of `patch-package`
- Use `npx` instead of `yarn` to run app-toolbelt commands
- Replace `md5` with `js-sha256`

### Side Effects

- You might need to run `corepack enable` once
- I bumped the native resource cache version due to changing file names (due to the switch from md5 to sha256), which means previously downloaded resources get deleted.
- react-native does not work with hoisted dependencies in yarn 4. I therefore disabled hoisting of native dependencies.
- It seems like `corepack enable` does not work on macos in the CI, most likely the same issue as https://github.com/nodejs/corepack/issues/507. I therefore had to use `corepack yarn` instead of `yarn`.

### Testing

~~Smoke test everything on android, ios and web. Please especially test on ios as I don't have a working dev environment there yet.~~
It is probably best to test this together with #3967.

### Resolved Issues

Fixes: #2966

---


lunars97 (approved):
Thanks for taking care of it! LGTM 👍

PR #3998, opened 2026-04-06
3997: Fix push notification undefined error

### LINKED ISSUE #3997: News is not opening correctly
### Describe the Bug

New sent news which are landing in notifications in background are not opening and leading to an undefined error

### Steps to Reproduce

1. Create new news in cms
2. See that news notification is there 
3. Click on the news notification

### Expected Behavior

It should open in news

### Related Issues

might be related to #3948 

### Additional Information

Image

lunars97:
### Short Description

This PR fixes push notifications not being open when it is in the foreground/background which leads to undefined id.

### Proposed Changes

-  Move usePushNotificationListener() hook into BottomTabNavigation, which is an actual screen mounted inside the root stack and has a proper navigation context

### Side Effects

- should be none

### Testing

1. Send push notification via cms news
2. Check push notification in quit, foreground/background states
3. See that it leads opens without any error

### Resolved Issues

Fixes: #3997 

---

### DIFF
--- a/native/src/Navigator.tsx
+++ b/native/src/Navigator.tsx
@@ -31,7 +31,6 @@ import { NavigationProps, RouteProps, RoutesParamsType, RoutesType } from './con
 import buildConfig from './constants/buildConfig'
 import { useAppContext } from './hooks/useCityAppContext'
 import useLoadCities from './hooks/useLoadCities'
-import useNavigate from './hooks/useNavigate'
 import useSnackbar from './hooks/useSnackbar'
 import BottomTabNavigation from './navigation/BottomTabNavigation'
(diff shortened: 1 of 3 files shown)


steffenkleinle (approved):
Nice catch, tested on Android emulator, works as expected!

bahaaTuffaha (approved):
Nice!
Tested the push-notification while the app on foreground and also background ✅

Docmale:
@lunars97 Isn't this PR more a native fix than a maintenance PR?

steffenkleinle:
Yes, its a fix, but it was not ever released (if I am not mistaken), so I thought `maintenance` makes more sense since its not a user-visible change and therefore not testable. However, I don't mind marking this as native fix.

PR #4003, opened 2026-04-08
4002: Fix app not starting if other whitelabel is active

steffenkleinle:
### Short Description

This PR fixes the problem that the app doesn't start and just shows a white screen if another whitelabel/dev version is already active.

### Proposed Changes

- Remove hard coded static server port
- Remove deprecated `nonLocal` property

### Side Effects

N/A

### Testing

See #4002.

### Resolved Issues

Fixes: #4002

---

### DIFF
--- a/native/src/components/StaticServerProvider.tsx
+++ b/native/src/components/StaticServerProvider.tsx
@@ -8,13 +8,8 @@ type StaticServerProps = {
   children: ReactNode
 }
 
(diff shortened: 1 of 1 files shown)


bahaaTuffaha (approved):
This could fix the random annoying static server crashes 😀
LGTM ✅

lunars97 (approved):
Seems like it fixes it. Tested on android emulator v15 and real device v16 👍

PR #4011, opened 2026-04-10
4009: Fix padding for CityNotCooperatingFooter

### LINKED ISSUE #4009: Missing spacing on left side
### Describe the Problem
The text “Deine Region nicht gefunden? Werde jetzt aktiv!” is too close to the left edges of the screen.
_hint: noticed on iOS/iPhone 12 mini (probably it is a problem for small screens only)_

There should be more space (padding) on both sides of the text so it is easier to read and looks better.

### Additional Information

Image

bahaaTuffaha:
### Proposed Changes

- Added 24px on the horizontal padding for the FooterContainer at `CityNotCooperatingFooter` .
- Increased the icon size to 35%

### Side Effects

- None.

### Testing

- Launch integreat on android small device size and switch the system language to german.
- Also test this text on other language like english.

### Resolved Issues

Fixes: #4009

---


steffenkleinle (approved):
Tested on android

lunars97 (approved):
LGTM! Tested with iOS SE emulator 👍

lunars97:
Could we maybe align the words `jetzt active` to center? imho it would look better... What do you think? @ulyanahoy-cloud 🤔 
Simulator Screenshot - iPhone SE (2nd generation) - 2026-04-10 at 22 43 44

ulyanahoy-cloud:
I don’t think we need to center-align “jetzt aktiv.”
Because of different device widths and screen sizes, the line break would still vary. In other languages, the text may also be much longer, which would make the layout even less consistent. 

I tested it on an iPhone 16 Pro, and it seems to be a padding issue in Polish.

bahaaTuffaha:
> I don’t think we need to center-align “jetzt aktiv.” Because of different device widths and screen sizes, the line break would still vary. In other languages, the text may also be much longer, which would make the layout even less consistent.
> 
> I tested it on an iPhone 16 Pro, and it seems to be a padding issue in Polish.

Did you test the fix or this is on production?

steffenkleinle:
> > I don’t think we need to center-align “jetzt aktiv.” Because of different device widths and screen sizes, the line break would still vary. In other languages, the text may also be much longer, which would make the layout even less consistent.
> > I tested it on an iPhone 16 Pro, and it seems to be a padding issue in Polish.
> 
> Did you test the fix or this is on production?

I think there is no way for @ulyanahoy-cloud to test this fix, so I'd assume strongly that this is still production without your fix.

PR #4013, opened 2026-04-10
3933: Increase threshold for typos in search

### LINKED ISSUE #3933: Increase threshold for typos in search
### Describe the Problem

We currently already show search results even if the words include typos. The threshold seems rather low though, so we could think about increasing it. An example: 
Farrad returns results while Farrhad or Farhrad does not.

Increase the max distance of the fuzzy search:
https://lucaong.github.io/minisearch/classes/MiniSearch.MiniSearch.html#md:fuzzy-search
Try to find a good balance between finding the desired results and getting irrelevant results.

N/A

### Related Issues
N/A

### Additional Information

N/A

bahaaTuffaha:
### Short Description

We currently already show search results even if the words include typos. The threshold seems rather low though, so we could think about increasing it. An example:
Farrad returns results while Farrhad or Farhrad does not.

### Proposed Changes

- Increased fuzzy value from the default `0.2` to `0.25` .
- Created small test.

### Side Effects

- None.

### Testing

- Test the search with the following words that includes typos: Farrhad and Farhrad. **(it should return search results)**

### Resolved Issues

Fixes: #3933

---


steffenkleinle (approved):
Tested on web

lunars97 (approved):
Thank you! Tested on android emulator 👍

kalairubachandrakumaran-oss:
tested ✅ 
image

PR #38, opened 2020-10-20
IGAPP-327: Fix CI deliveries

klinzo:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **IGAPP**.

PR #4015, opened 2026-04-12
3966: Upgrade react-native-async-storage to v3

### LINKED ISSUE #3966: Upgrade react-native-async-storage
### Describe the Problem

react-native-async-storage has a new major release (v3.0.0) with API changes that were out of scope for #3951.

Upgrade react-native-async-storage and use the new API.
See the migration guide: https://react-native-async-storage.github.io/3.0/migration-to-3/

N/A

### Related Issues

#3951

### Additional Information

https://react-native-async-storage.github.io/3.0/migration-to-3/

lunars97:
### Short Description
This PR upgrades react-native-async-storage to v3 https://react-native-async-storage.github.io/3.0/migration-to-3/

### Proposed Changes

- Update API methods
- Create function which copies user settings from the old storage and moves to new storage to avoid losing users' data during the upgrade
- Fix test

### Side Effects

- should be none

### Testing

1. First remove awaiting for `migrateToV2()` from `loadSettings()`
2. Change the values related to async storage (for ex. setting a language, a contrast mode / disabling push notification and etc)
3. Quit the app
4. Place awaiting for `migrateToV2()` in `loadSettings()`
5. Open the app and see the settings stayed the same

### Resolved Issues

Fixes: #3966 

---

### DIFF
--- a/native/android/build.gradle
+++ b/native/android/build.gradle
@@ -19,8 +19,6 @@ buildscript {
     }
     allprojects {
         repositories {
-            google()
-            mavenCentral()
             maven {
                 url = uri("${rootDir}/../node_modules/@react-native-async-storage/async-storage/android/local_repo")
             }
--- a/native/android/build.gradle
+++ b/native/android/build.gradle
@@ -17,6 +17,15 @@ buildscript {
         classpath("com.facebook.react:react-native-gradle-plugin")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
     }
+    allprojects {
+        repositories {
+            google()
+            mavenCentral()
+            maven {
+                url = uri("${rootDir}/../node_modules/@react-native-async-storage/async-storage/android/local_repo")
+            }
+        }
+    }
 }
 
 apply plugin: "com.facebook.react.rootproject"
--- a/native/src/constants/index.ts
+++ b/native/src/constants/index.ts
@@ -2,8 +2,6 @@ export const ROOT_NAVIGATOR_ID = 'rootNavigator'
 export const TAB_NAVIGATOR_ID = 'tabNavigator'
 export type NavigatorIds = typeof ROOT_NAVIGATOR_ID | typeof TAB_NAVIGATOR_ID
 
-export const MIGRATION_KEY = 'settings_migrated_to_v3'
-
 export const SEARCH_PREVIEW_MAX_CHARS = 90
 export const EXCERPT_MAX_LINES = 4
 export const NAVIGATION_INITIALIZATION_DELAY = 100
--- a/native/src/constants/index.ts
+++ b/native/src/constants/index.ts
@@ -2,6 +2,8 @@ export const ROOT_NAVIGATOR_ID = 'rootNavigator'
 export const TAB_NAVIGATOR_ID = 'tabNavigator'
 export type NavigatorIds = typeof ROOT_NAVIGATOR_ID | typeof TAB_NAVIGATOR_ID
 
+export const MIGRATION_KEY = 'settings_migrated_to_v3'
+
 export const SEARCH_PREVIEW_MAX_CHARS = 90
 export const EXCERPT_MAX_LINES = 4
 export const NAVIGATION_INITIALIZATION_DELAY = 100
--- a/native/src/contexts/AppContextProvider.tsx
+++ b/native/src/contexts/AppContextProvider.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from 'react-i18next'
 import { hasProp } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
-import appSettings, { defaultSettings, SettingsType } from '../utils/AppSettings'
+import appSettings, { defaultSettings, copySettingsFromV2ToV3, SettingsType } from '../utils/AppSettings'
 import dataContainer from '../utils/DefaultDataContainer'
 import { subscribeNews, unsubscribeNews } from '../utils/PushNotificationsManager'
 import { reportError } from '../utils/sentry'
@@ -43,7 +43,11 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
   const uiLanguage = i18n.languages[0]
 
   useEffect(() => {
-    appSettings.loadSettings().then(setSettings).catch(reportError)
+    // Migrate v2 entries first, then load from v3 storage
+    copySettingsFromV2ToV3()
+      .then(() => appSettings.loadSettings())
+      .then(setSettings)
+      .catch(reportError)
   }, [])
 
   const updateSettings = useCallback((settings: Partial<SettingsType>) => {
--- a/native/src/contexts/AppContextProvider.tsx
+++ b/native/src/contexts/AppContextProvider.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from 'react-i18next'
 import { hasProp } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
-import appSettings, { defaultSettings, copySettingsFromV2ToV3, SettingsType } from '../utils/AppSettings'
+import appSettings, { defaultSettings, migrateSettingsToV2, SettingsType } from '../utils/AppSettings'
 import dataContainer from '../utils/DefaultDataContainer'
 import { subscribeNews, unsubscribeNews } from '../utils/PushNotificationsManager'
 import { reportError } from '../utils/sentry'
@@ -43,11 +43,8 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
   const uiLanguage = i18n.languages[0]
 
   useEffect(() => {
-    // Migrate v2 entries first, then load from v3 storage
-    copySettingsFromV2ToV3()
-      .then(() => appSettings.loadSettings())
-      .then(setSettings)
-      .catch(reportError)
+    // Migrate v1 entries first, then load from v2 storage
+    migrateSettingsToV2().then(appSettings.loadSettings).then(setSettings).catch(reportError)
   }, [])
 
   const updateSettings = useCallback((settings: Partial<SettingsType>) => {
--- a/native/src/contexts/AppContextProvider.tsx
+++ b/native/src/contexts/AppContextProvider.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from 'react-i18next'
 import { hasProp } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
-import appSettings, { defaultSettings, migrateSettingsToV2, SettingsType } from '../utils/AppSettings'
+import appSettings, { defaultSettings, SettingsType } from '../utils/AppSettings'
 import dataContainer from '../utils/DefaultDataContainer'
 import { subscribeNews, unsubscribeNews } from '../utils/PushNotificationsManager'
 import { reportError } from '../utils/sentry'
@@ -43,8 +43,7 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
   const uiLanguage = i18n.languages[0]
 
   useEffect(() => {
-    // Migrate v1 entries first, then load from v2 storage
-    migrateSettingsToV2().then(appSettings.loadSettings).then(setSettings).catch(reportError)
+    appSettings.loadSettings().then(setSettings).catch(reportError)
   }, [])
 
   const updateSettings = useCallback((settings: Partial<SettingsType>) => {
--- a/native/src/utils/AppSettings.ts
+++ b/native/src/utils/AppSettings.ts
@@ -4,7 +4,7 @@ import { mapValues } from 'lodash'
 import { ThemeKey } from 'build-configs/ThemeKey'
 import { ExternalSourcePermissions } from 'shared'
 
-import { log } from './sentry'
+import { log, reportError } from './sentry'
 
 export const ASYNC_STORAGE_VERSION = 2
 export type SettingsType = {
@@ -63,7 +63,12 @@ class AppSettings {
   }
 
   loadSettings = async (): Promise<SettingsType> => {
-    await migrateToV2()
+    try {
+      await migrateToV2()
+    } catch (error) {
+      reportError(error)
+    }
+
     const settingsKeys = Object.keys(defaultSettings) as [keyof SettingsType]
     const settings = (await this.asyncStorage.getMany(settingsKeys)) as Record<keyof SettingsType, string | null>
     return mapValues(settings, (value: string | null, key) => {
--- a/native/src/utils/AppSettings.ts
+++ b/native/src/utils/AppSettings.ts
@@ -29,38 +29,35 @@ export const defaultSettings: SettingsType = {
 }
 export const settingsStorage = createAsyncStorage('settings')
 
+const migrateToV2 = async (): Promise<void> => {
+  const currentVersion = await settingsStorage.getItem('storageVersion')
+  if (currentVersion === JSON.stringify(ASYNC_STORAGE_VERSION)) {
+    return
+  }
+
+  const keys = Object.keys(defaultSettings) as (keyof SettingsType)[]
+  const values = await Promise.all(keys.map(key => LegacyAsyncStorage.getItem(key)))
+
+  const settingsToCopy = keys.reduce<Record<string, string>>((settings, key, index) => {
+    const value = values[index]
+    if (value) {
+      return { ...settings, [key]: value }
+    }
+    return settings
+  }, {})
+
+  await settingsStorage.setMany({ ...settingsToCopy, storageVersion: JSON.stringify(ASYNC_STORAGE_VERSION) })
+}
+
 class AppSettings {
   asyncStorage: AsyncStorage
-  private migrated: Promise<void> | null = null
 
   constructor(asyncStorage: AsyncStorage = settingsStorage) {
     this.asyncStorage = asyncStorage
-    this.migrated = this.migrateToV2()
-  }
-
-  private migrateToV2 = async (): Promise<void> => {
-    const currentVersion = await this.asyncStorage.getItem('storageVersion')
-    if (currentVersion === ASYNC_STORAGE_VERSION) {
-      return
-    }
-
-    const keys = Object.keys(defaultSettings) as (keyof SettingsType)[]
-    const values = await Promise.all(keys.map(key => LegacyAsyncStorage.getItem(key)))
-
-    const settingsToCopy = keys.reduce<Record<string, string>>((settings, key, index) => {
-      const value = values[index]
-      if (value) {
-        // eslint-disable-next-line no-param-reassign
-        settings[key] = value
-      }
-      return settings
-    }, {})
-
-    await this.asyncStorage.setMany({ ...settingsToCopy, storageVersion: ASYNC_STORAGE_VERSION })
   }
 
   loadSettings = async (): Promise<SettingsType> => {
-    await this.migrated
+    await migrateToV2()
     const settingsKeys = Object.keys(defaultSettings) as [keyof SettingsType]
     const settings = (await this.asyncStorage.getMany(settingsKeys)) as Record<keyof SettingsType, string | null>
     return mapValues(settings, (value: string | null, key) => {
@@ -69,14 +66,7 @@ class AppSettings {
         return defaultSettings[key as keyof SettingsType]
       }
 
-      const parsed = JSON.parse(value)
-
-      if (parsed === null) {
-        // null means this setting does not exist
-        return JSON.parse(value) ?? defaultSettings[key as keyof SettingsType]
-      }
-
-      return parsed
+      return JSON.parse(value) ?? defaultSettings[key as keyof SettingsType]
     })
   }
 
--- a/native/src/utils/AppSettings.ts
+++ b/native/src/utils/AppSettings.ts
@@ -1,8 +1,10 @@
-import AsyncStorage from '@react-native-async-storage/async-storage'
+import AsyncStorage, { createAsyncStorage } from '@react-native-async-storage/async-storage'
 import { mapValues } from 'lodash'
 
 import { ThemeKey } from 'build-configs/ThemeKey'
 import { ExternalSourcePermissions } from 'shared'
+
+import { MIGRATION_KEY } from '../constants'
 
 export const ASYNC_STORAGE_VERSION = '1'
 export type SettingsType = {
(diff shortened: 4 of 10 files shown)


native/src/utils/AppSettings.ts
```diff
@@ -27,18 +29,42 @@ export const defaultSettings: SettingsType = {
   externalSourcePermissions: {},
   selectedTheme: 'light',
 }
+export const settingsStorage = createAsyncStorage('settings')
+
+export const copySettingsFromV2ToV3 = async (): Promise<void> => {
```
lunars97: this function should be removed after 1-2 releases to make sure that users do not need to reconfigure their settings again

native/src/contexts/AppContextProvider.tsx
```diff
@@ -43,7 +43,11 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
   const uiLanguage = i18n.languages[0]
 
   useEffect(() => {
-    appSettings.loadSettings().then(setSettings).catch(reportError)
+    // Migrate v2 entries first, then load from v3 storage
+    copySettingsFromV2ToV3()
```
lunars97: should be removed after new releases

native/android/build.gradle
```diff
@@ -17,6 +17,15 @@ buildscript {
...
+    allprojects {
+        repositories {
+            google()
+            mavenCentral()
+            maven {
+                url = uri("${rootDir}/../node_modules/@react-native-async-storage/async-storage/android/local_repo")
+            }
+        }
+    }
 }
```
bahaaTuffaha, line 27: There is a problem with the autolinking:

```bash
BUILD FAILED in 6s
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081 -PreactNativeArchitectures=arm64-v8a
/Users/digitalfabrik/Repos/integreat-app/native/android/app/build/generated/autolinking/src/main/java/com/facebook/react/PackageList.java:71: error: package com.reactnativecommunity.asyncstorage does not exist new com.reactnativecommunity.asyncstorage.AsyncStoragePackage(), ^
1 error FAILURE: Build failed with an exception. * What went wrong:
Execution failed for task ':app:compileDebugJavaWithJavac'.
```

i didn't catch it at first until I deleted the autolinking:
```
rm -rf native/android/build/generated/autolinking/
```

lunars97: hmm it did not happen to me, it ran without any issues :thinking:

lunars97: could you check again?

native/src/contexts/AppContextProvider.tsx
```diff
@@ -43,7 +43,11 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
...
+      .then(() => appSettings.loadSettings())
```
steffenkleinle: 🙃 
```suggestion
      .then(appSettings.loadSettings)
```

native/src/constants/index.ts
```diff
@@ -2,6 +2,8 @@ export const ROOT_NAVIGATOR_ID = 'rootNavigator'
 export const TAB_NAVIGATOR_ID = 'tabNavigator'
 export type NavigatorIds = typeof ROOT_NAVIGATOR_ID | typeof TAB_NAVIGATOR_ID
 
+export const MIGRATION_KEY = 'settings_migrated_to_v3'
```
steffenkleinle: ❌  We already have an intended logic for this, please use `ASYNC_STORAGE_VERSION` which is already persisted in the settings, bump it to `2` and migrate if it was still `1` (or not available in the new settings yet).

native/src/contexts/AppContextProvider.tsx
steffenkleinle: 🔧 If you update the name of the function, the comment becomes redundant. Also, internally, this is also v2 and not v3. That we now use async storage v3 is an implementation detail and you can add a comment for that in the actual migration routine.
```suggestion
    migrateSettingsToV2()
```

native/src/utils/AppSettings.ts
```diff
@@ -1,9 +1,11 @@
-import AsyncStorage from '@react-native-async-storage/async-storage'
-import { fromPairs, mapValues, toPairs } from 'lodash'
+import AsyncStorage, { createAsyncStorage } from '@react-native-async-storage/async-storage'
```
steffenkleinle: 🔧 
```suggestion
import LegacyAsyncStorage, { AsyncStorage, createAsyncStorage } from '@react-native-async-storage/async-storage'
```
This allows you to
a) mark that the default export is now legacy and should not be used anymore
b) Import the correct type which you can now use in L59 and L61

native/android/build.gradle
```diff
@@ -17,6 +17,15 @@ buildscript {
...
     }
+    allprojects {
+        repositories {
+            google()
+            mavenCentral()
+            maven {
+                url = uri("${rootDir}/../node_modules/@react-native-async-storage/async-storage/android/local_repo")
+            }
+        }
+    }
```
steffenkleinle, line 26: ❓ According to the [installation docs](https://react-native-async-storage.github.io/3.0/migration-to-3/#installation-changes) this should be/only needs to be:
```suggestion
    allprojects {
        repositories {
            maven {
                url = uri(project(":react-native-async-storage_async-storage").file("local_repo"))
            }
        }
    }
```
Did you try this or was there a problem with this?

native/src/utils/AppSettings.ts
```diff
@@ -57,8 +83,8 @@ class AppSettings {
   }
 
   setSettings = async (settings: Partial<SettingsType>): Promise<void> => {
-    const settingsArray = toPairs<string>(mapValues(settings, value => JSON.stringify(value)))
-    await this.asyncStorage.multiSet(settingsArray)
+    const entries = mapValues(settings, value => JSON.stringify(value)) as Record<string, string>
```
steffenkleinle: 🙃 The typecast is not necessary for me
```suggestion
    const entries = mapValues(settings, value => JSON.stringify(value))
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +29,42 @@ export const defaultSettings: SettingsType = {
...
+  await AsyncStorage.setItem(MIGRATION_KEY, 'true')
+}
 
 class AppSettings {
-  asyncStorage: typeof AsyncStorage
+  asyncStorage: typeof settingsStorage
 
-  constructor(asyncStorage: typeof AsyncStorage = AsyncStorage) {
+  constructor(asyncStorage: typeof settingsStorage = settingsStorage) {
     this.asyncStorage = asyncStorage
```
steffenkleinle: 🔧 See suggestion from above
```suggestion
  asyncStorage: AsyncStorage

  constructor(asyncStorage: AsyncStorage = settingsStorage) {
this.asyncStorage = asyncStorage
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +29,42 @@ export const defaultSettings: SettingsType = {
...
+    return
+  }
+
+  const keys = Object.keys(defaultSettings) as (keyof SettingsType)[]
+  const values = await Promise.all(keys.map(key => AsyncStorage.getItem(key)))
+
+  const settingsToCopy: Record<string, string> = {}
+  keys.forEach((key, index) => {
+    const value = values[index]
+    if (value !== null && value !== undefined) {
```
steffenkleinle:
```suggestion
    if (value) {
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +29,42 @@ export const defaultSettings: SettingsType = {
...
+      settingsToCopy[key] = value
+    }
+  })
```
steffenkleinle: 🤔  Could you perhaps try to use `keys.reduce()` to make this a little more elegant?

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +29,42 @@ export const defaultSettings: SettingsType = {
...
+
+  if (Object.keys(settingsToCopy).length > 0) {
+    await settingsStorage.setMany(settingsToCopy)
+  }
```
steffenkleinle: 🤔 Is this check necessary? What happens if you try to set many with an empty object?

steffenkleinle (changes requested):
Very nice work already! Tested on Android and already works as expected :)

💭 In theory it would be nice to align the settings keys with our internal naming in other places, e.g. `selectedCity` -> `cityCode`. But I think this is more complicated than it gives us a benefit, so probably we are better off just keeping it as is.

steffenkleinle:
🤔 Perhaps you could also remove L109-L111 from `Navigator`, I guess that is not needed anymore (since a long time) :D
``` tsx
    if (!storageVersion) {
      updateSettings({ storageVersion: ASYNC_STORAGE_VERSION })
    }
```

steffenkleinle:
And while this is a `maintenance` task, lets make sure to notify our testers after the next release that this gets properly tested anyway :)

native/android/build.gradle
lunars97: There was a problem with it and it also mentions that `// ... others like google(), mavenCentral()` so i think they should all be together there

lunars97: so far no error reports this way

steffenkleinle: > There was a problem with it and it also mentions that // ... others like google(), mavenCentral() so i think they should all be together there

Please add more info on what kind of problem :)

native/src/utils/AppSettings.ts
lunars97: this is not necessary, but it was just an additional check. Seems like `setMany({}) `is atomic and with no keys basically does nothing :+1:

native/android/build.gradle
lunars97: I cant reproduce that issue, apparently it was a glitch :(

native/src/contexts/AppContextProvider.tsx
```diff
@@ -43,7 +43,8 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
   const uiLanguage = i18n.languages[0]
 
   useEffect(() => {
-    appSettings.loadSettings().then(setSettings).catch(reportError)
+    // Migrate v1 entries first, then load from v2 storage
```
steffenkleinle: 🙃 Imo the code is pretty much self explanatory and the comment redundant. If you think it adds info though, feel free to keep it.
```suggestion
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +27,40 @@ export const defaultSettings: SettingsType = {
...
+  const keys = Object.keys(defaultSettings) as (keyof SettingsType)[]
+  const values = await Promise.all(keys.map(key => LegacyAsyncStorage.getItem(key)))
+
+  const settingsToCopy = keys.reduce<Record<string, string>>((accumulator, key, index) => {
+    const value = values[index]
+    if (value) {
+      accumulator[key] = value
+    }
+    return accumulator
+  }, {})
```
steffenkleinle: 🙃 I haven't been doing this in the past either, but I think replacing `accumulator` with the actual thing we create makes the code more understandable:
```suggestion
  const settingsToCopy = keys.reduce<Record<string, string>>((settings, key, index) => {
    const value = values[index]
    if (value) {
      settings[key] = value
    }
    return settings
  }, {})
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +27,40 @@ export const defaultSettings: SettingsType = {
...
+
+  await settingsStorage.setMany(settingsToCopy)
+
+  await settingsStorage.setItem(ASYNC_STORAGE_VERSION, '2')
+}
```
steffenkleinle: ❌ `ASYNC_STORAGE_VERSION` is only a stringified integer, so you'd set `"1": "2"`. What you mean to do instead:

```suggestion
  await settingsStorage.setMany({ ...settingsToCopy, storageVersion: ASYNC_STORAGE_VERSION })
}
```
and bump `ASYNC_STORAGE_VERSION` to `'2'`.

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +27,40 @@ export const defaultSettings: SettingsType = {
   externalSourcePermissions: {},
   selectedTheme: 'light',
 }
+export const settingsStorage = createAsyncStorage('settings')
+
+export const migrateSettingsToV2 = async (): Promise<void> => {
+  const currentVersion = await settingsStorage.getItem(ASYNC_STORAGE_VERSION)
+  if (currentVersion === '2') {
+    return
+  }
```
steffenkleinle: ❌ See comment above. `ASYNC_STORAGE_VERSION` has just an integer as value and resembles the current storage version (in our case `2`).
Accordingly, you should probably do something like:
```suggestion
  const migratedSettings = await this.loadSettings()
  if (migratedSettings.storageVersion === ASYNC_STORAGE_VERSION) {
    return
  }
```
For this to work, you need to move `migrateSettingsToV2` inside of `AppSettings` though (see comment in `AppContextProvider.tsx`.

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +27,40 @@ export const defaultSettings: SettingsType = {
   externalSourcePermissions: {},
   selectedTheme: 'light',
 }
+export const settingsStorage = createAsyncStorage('settings')
+
+export const migrateSettingsToV2 = async (): Promise<void> => {
```
steffenkleinle: 🙃 
```suggestion
export const migrateToV2 = async (): Promise<void> => {
```

native/src/contexts/AppContextProvider.tsx
```diff
@@ -43,7 +43,8 @@ const AppContextProvider = ({ children }: AppContextProviderProps): ReactElement
...
+    migrateSettingsToV2().then(appSettings.loadSettings).then(setSettings).catch(reportError)
```
steffenkleinle: 🔧 Sorry for not catching this earlier. Imo, you should call the migration routine directly inside `AppSettings.ts`, probably just inside the constructor. Migrating to a different version is an implementation detail of the app settings and it should not be necessary to remember this in other parts of the app.

native/src/utils/AppSettings.ts
steffenkleinle: 🙃 Not changed by you, so completely optinal:
```suggestion
      return JSON.parse(value) ?? defaultSettings[key as keyof SettingsType]
```

steffenkleinle (changes requested):

native/src/utils/AppSettings.ts
steffenkleinle: 🔧 And we need to update the storage version here, ideally lets just make this an integer to be able to compare properly:
```suggestion
const ASYNC_STORAGE_VERSION = 2
```

native/src/utils/AppSettings.ts
lunars97: I also wanted to name it like you named it, but receiving `ESLint: Assignment to property of function parameter 'settings'. (no-param-reassign) `

lunars97: no lint error when I use spread operator to return a new object like this: 
`const settingsToCopy = keys.reduce<Record<string, string>>((settings, key, index) => {
    const value = values[index]
    if (value) {
      return { ...settings, [key]: value }
    }
    return settings
  }, {})`

lunars97: I will just suppress the rule for this case

native/src/utils/AppSettings.ts
lunars97: but I could directly access it like this 
```suggestion
  const currentVersion =  await settingsStorage.getItem('storageVersion')
  if (currentVersion === ASYNC_STORAGE_VERSION) {
    return
  }
```

native/src/contexts/AppContextProvider.tsx
lunars97: good point!

native/src/utils/AppSettings.ts
steffenkleinle: Of course, however, its another unnecessary async call and not type-safe, i.e. if we ever rename `storageVersion` to something else, this breaks. Therefore I'd have a preference for my suggestion, but in the end up to you :)

native/src/utils/AppSettings.ts
steffenkleinle: Out of curiosity: Why did you go for disabling the rule instead of using the spread syntax?

native/src/utils/AppSettings.ts
```diff
@@ -49,16 +73,16 @@ class AppSettings {
 
       if (parsed === null) {
         // null means this setting does not exist
-        return defaultSettings[key as keyof SettingsType]
+        return JSON.parse(value) ?? defaultSettings[key as keyof SettingsType]
       }
 
       return parsed
```
steffenkleinle: 🔧 Sorry, but I think there was a misunderstanding:
``` suggestion
      return JSON.parse(value) ?? defaultSettings[key as keyof SettingsType]
```

native/src/utils/AppSettings.ts
```diff
@@ -27,18 +27,42 @@ export const defaultSettings: SettingsType = {
...
+        settings[key] = value
+      }
+      return settings
+    }, {})
+
+    await this.asyncStorage.setMany({ ...settingsToCopy, storageVersion: ASYNC_STORAGE_VERSION })
   }
 
   loadSettings = async (): Promise<SettingsType> => {
+    await this.migrated
```
steffenkleinle: 🤔 Hmm, if you want to await the promise here, I guess its better to just also directly call the function here and remove it from the constructor:
```suggestion
    await migrateToV2()
```

native/src/utils/AppSettings.ts
lunars97: I think in our case it is safe and it will not lead to any mutations :thinking:  but I also agree that it is a bad practice, so I will use spread operator then :)

native/src/utils/AppSettings.ts
lunars97: I think renaming it for something else is highly unlikely in the near future :)

native/src/utils/AppSettings.ts
```diff
@@ -1,10 +1,10 @@
-import AsyncStorage from '@react-native-async-storage/async-storage'
-import { fromPairs, mapValues, toPairs } from 'lodash'
+import LegacyAsyncStorage, { AsyncStorage, createAsyncStorage } from '@react-native-async-storage/async-storage'
+import { mapValues } from 'lodash'
 
 import { ThemeKey } from 'build-configs/ThemeKey'
 import { ExternalSourcePermissions } from 'shared'
 
-export const ASYNC_STORAGE_VERSION = '1'
+export const ASYNC_STORAGE_VERSION = '2'
```
steffenkleinle: 🤔 As mentioned before: Any reason why we can't just make this an integer?
```suggestion
export const ASYNC_STORAGE_VERSION = 2
```

lunars97: yes, but then we need to change the SettingsType as well

native/src/utils/AppSettings.ts
```diff
@@ -27,38 +27,52 @@ export const defaultSettings: SettingsType = {
...
   selectedTheme: 'light',
 }
+export const settingsStorage = createAsyncStorage('settings')
+
+const migrateToV2 = async (): Promise<void> => {
+  const currentVersion = await settingsStorage.getItem('storageVersion')
+  if (currentVersion === JSON.stringify(ASYNC_STORAGE_VERSION)) {
+    return
+  }
+
```
steffenkleinle, line 39: 🔧 It might be helpful to add a short log notice
```suggestion

  log('Migrating settings to v2')
```

native/src/utils/AppSettings.ts
```diff
@@ -27,38 +27,52 @@ export const defaultSettings: SettingsType = {
...
+  const settingsToCopy = keys.reduce<Record<string, string>>((settings, key, index) => {
+    const value = values[index]
+    if (value) {
+      return { ...settings, [key]: value }
+    }
+    return settings
+  }, {})
+
+  await settingsStorage.setMany({ ...settingsToCopy, storageVersion: JSON.stringify(ASYNC_STORAGE_VERSION) })
```
steffenkleinle, line 53: 🔧 
```suggestion
  await settingsStorage.setMany({ ...settingsToCopy, storageVersion: JSON.stringify(ASYNC_STORAGE_VERSION) })
  log('Migrated settings to v2')
```

steffenkleinle (changes requested):
Nice work! Tested on Android and iOS. For iOS, you'll still need to update the pods by running `pod install`. Once doing that, everything works perfectly though.

steffenkleinle (approved):
Nice work!

bahaaTuffaha (approved):
Great work! 🚀 
Tested on Android 14 , 16 ✅ 
and 14 Real device. ✅ 

Sometimes after migration to v2 it crashes not sure if its related (it runs ok when I reopened again):
```bash
04-20 16:42:17.067  9896 10065 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 10065 (mqt_v_js), pid 9896 (.integreat.test)
04-20 16:42:18.167 10101 10101 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
04-20 16:42:18.167 10101 10101 F DEBUG   : Cmdline: app.integreat.test
04-20 16:42:18.167 10101 10101 F DEBUG   : pid: 9896, tid: 10065, name: mqt_v_js  >>> app.integreat.test <<<
04-20 16:42:18.167 10101 10101 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0000000000000000
04-20 16:42:18.167 10101 10101 F DEBUG   : Cause: null pointer dereference
04-20 16:42:18.167 10101 10101 F DEBUG   : Process uptime: 21s
  #00–#06  libhermesvm.so         ← crash inside Hermes JS engine
  #07      RuntimeScheduler_Modern::performMicrotaskCheckpoint   ← RN triggered a JS microtask
  #08–#09  RuntimeScheduler_Modern::runEventLoopTick / runEventLoop
  #38      MessageQueueThreadHandler.dispatchMessage   ← JS thread dispatched a message
  #48      MessageQueueThreadImpl startNewBackgroundThread  ← on the JS background thread
```

- how about we wrap try and catch around the migrateToV2() function? 🤔

lunars97:
> Great work! 🚀 Tested on Android 14 , 16 ✅ and 14 Real device. ✅
> 
> Sometimes after migration to v2 it crashes not sure if its related (it runs ok when I reopened again):
> 
> ```shell
> 04-20 16:42:17.067  9896 10065 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 10065 (mqt_v_js), pid 9896 (.integreat.test)
> 04-20 16:42:18.167 10101 10101 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
> 04-20 16:42:18.167 10101 10101 F DEBUG   : Cmdline: app.integreat.test
> 04-20 16:42:18.167 10101 10101 F DEBUG   : pid: 9896, tid: 10065, name: mqt_v_js  >>> app.integreat.test <<<
> 04-20 16:42:18.167 10101 10101 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0000000000000000
> 04-20 16:42:18.167 10101 10101 F DEBUG   : Cause: null pointer dereference
> 04-20 16:42:18.167 10101 10101 F DEBUG   : Process uptime: 21s
>   #00–#06  libhermesvm.so         ← crash inside Hermes JS engine
>   #07      RuntimeScheduler_Modern::performMicrotaskCheckpoint   ← RN triggered a JS microtask
>   #08–#09  RuntimeScheduler_Modern::runEventLoopTick / runEventLoop
>   #38      MessageQueueThreadHandler.dispatchMessage   ← JS thread dispatched a message
>   #48      MessageQueueThreadImpl startNewBackgroundThread  ← on the JS background thread
> ```
> 
> * how about we wrap try and catch around the migrateToV2() function? 🤔

hmmm strange, I haven't faced this issue. I added try and catch block :+1:

bahaaTuffaha:
Do you need another review?

lunars97:
> Do you need another review?

it would be nice :+1:

bahaaTuffaha (approved):
Looks good...
at `Navigator.tsx`
unused import:
```
import { ASYNC_STORAGE_VERSION } from './utils/AppSettings'
```

f1sh1918 (approved):

PR #4020, opened 2026-04-14
4019: Improve header menu and use system share

### LINKED ISSUE #4019: Use system share dialog
### Describe the Problem

Since the native redesign, we also use fixed options of which apps we allow users to share with (WhatsApp, Facebook, Email) on native. This prevents users from sharing with their desired/used apps. Furthermore, this is very unusual on native.

Use the system share dialog instead of presenting pre-picked options only.

Show pre-picked options AND the native share dialog.

### Related Issues

Introduced in: https://github.com/digitalfabrik/integreat-app/issues/3804
Slightly related: #3030 

### Additional Information

N/A

### LINKED ISSUE #4027: Fix settings on landing page first install
### Describe the Bug

On landing page after fresh install the settings page can be accessed by the menuHeader opening it without selecting a city will cause a crash.

### Steps to Reproduce

1. Launch the native app (fresh install).
2. before selecting a region just open the header menu and select settings.

### Expected Behavior

~~Expected the settings menuItem to hide or use a default city code without errors.~~
Allow opening settings without a selected city.

### Additional Information

https://github.com/user-attachments/assets/b8ce1038-e357-4463-874a-66432d21e0e4

steffenkleinle:
### Short Description

This PR cleans up the header menu, both in code as well as shown items. We had a lot of duplicates that were already available in other places (e.g. settings).
On top, we use the system share dialog instead of hardcoded in-app options.

### Proposed Changes

- Use system dialog for sharing and remove hardcoded sharing options
- Extract HeaderMenuItem
- Remove contrast mode header item (available in settings)
- Remove legal header items (available in settings)
- Remove copy url header item (available in system sharing dialog)
- Show settings last
- Add accessibility and imprint to settings
- Move version to end of settings
- Fix opening settings without a selected city
- Fix main disclaimer content direction at http://webnext.integreat.app/main-disclaimer/ar (web)

### Side Effects

- ~~The accessibility setting also links to integreat for malte, aschaffenburg and obdach for now. I'll ask what should be the actual value for the whitelabels. In the meantime, this should be okay.~~ As discussed in https://chat.tuerantuer.org/digitalfabrik/pl/cpnjirh6g7rr3gmrf8doe6356o, we'll just use integreat's declaration for all whitelabels.
- The main disclaimer (i.e. the disclaimer if no city is selected) is only opened from the web app in the inapp browser. Should be improved in #4023.

### Testing

Thoroughly test all header overflow items (i.e. three dot/burger menu) on all screens (including categories, landing, settings, ...), especially sharing.
Test the new accessibility and imprint setting options.
Test the content direction at http://webnext.integreat.app/main-disclaimer/ar.

### Resolved Issues

Fixes: #4019, #4027

---


native/src/components/HeaderMenu.tsx
```diff
@@ -1,117 +1,79 @@
...
+    ...(shareUrl ? [<HeaderMenuItem key='share' title={t('share')} onPress={share} icon='share-variant' />] : []),
+    <HeaderMenuItem
+      key='settings'
+      title={t('settings')}
+      onPress={() => navigation.navigate(SETTINGS_ROUTE)}
+      icon='cog-outline'
+    />,
+  ].map((element: ReactElement<{ onPress?: () => void }>) =>
+    // Close header menu on item press
+    cloneElement(element, {
```
bahaaTuffaha: https://react.dev/reference/react/cloneElement

`Using cloneElement is uncommon and can lead to fragile code.`

There is alternatives:
https://react.dev/reference/react/cloneElement#passing-data-with-a-render-prop

native/src/components/Header.tsx
```diff
@@ -230,25 +163,28 @@ const Header = ({
...
       innerText={forceText ? currentLanguageName : undefined}
     />,
   ]
 
-  const overflowItems = showOverflowItems
-    ? [
-        ...(route.name !== NEWS_ROUTE
-          ? [renderMenuItem(HeaderButtonTitle.Feedback, navigateToFeedback, 'comment-text-outline')]
-          : []),
-        renderMenuItem('contrastTheme', toggleContrastTheme, 'contrast-circle'),
```
bahaaTuffaha, line 247: why did you remove the contrast from the menu?

bahaaTuffaha (approved):
I'm not a fan of this change cuz its different from the new design and the webapp.
These changes runs great overall.
- Tested integreat, Malte, and Aschaffenburg on Android ✅ 
- Tested integreat and Aschaffenburg on iOS ✅ 
- Tested disclamer page on web ✅ 

- Feedback menuItem at the landing page doesn't navigate to anything (fresh install)

native/src/components/HeaderMenu.tsx
steffenkleinle: Yeah, you are right. Thanks for the hint :)

native/src/components/Header.tsx
steffenkleinle: Because it is already available in the settings and there is no need for a duplication. This allows us to keep the menu more clean and therefore the user to focus more on the important things. On top, the contrast mode setting is not frequently necessary and users expect it in the settings.

steffenkleinle:
> I'm not a fan of this change cuz its different from the new design and the webapp. These changes runs great overall.

I definitely get your point, I am usually the first that is aiming for consistency. However, it shouldn't be a self-purpose, i.e. consistency just for the goal of consistency. The goal is in the end always to improve the user experience and most of the time, consistency helps to achieve that by creating memorability for users. However, there are other factors to consider and sometimes other alternatives are better for the user experience. In this case, its more important to
a) achieve consistency not with our web-app, but with other apps users are familiar with. Users expect to use the system share dialog (and to be able to choose whatever app they want), at least I don't know a single app that only presents me with preselected app options to share with. Also legal options are mostly part of the settings in some way or another, same goes for high contrast mode.
b) avoid a cluttered menu and let users focus on the important and frequently used things. Users will basically never press any item in the legal section and rarely (or usually exactly zero to one time) use the high contrast mode setting. On the other hand, we want to promote the feedback and share features, which become more visible if there are less items in the menu.

In the end, there are just differences between a web app and native apps and I think we have to accept that sometimes. And even better if that allows us to reduce code and complexity.

>     * Feedback menuItem at the landing page doesn't navigate to anything (fresh install)

Thanks, fixed.

lunars97:
not sure, whether it is related to this, but there is a bug in header title. If you go from landing page to settings... 
image

lunars97:
> not sure, whether it is related to this, but there is a bug in header title. If you go from landing page to settings...

also in `Suggest Your Region`
image

lunars97:
Also in Aschaffenburg there is a header title in settings: `pdf`, if you go to Hallo!->Aschaffenburg in Zahlen->Broschure zum Familienleben (pdf file) -> Settings

steffenkleinle:
> not sure, whether it is related to this, but there is a bug in header title. If you go from landing page to settings...

Yes, our current `getHeaderText` logic is not sufficient. I fixed the text for settings from landing (as I think that is the most confusing and most common). I think the rest is a little out of scope here and not very critical. I'll open a separate issue for that. Good catch!

lunars97 (approved):
LGTM! Tested on real device android 16 integreat and aschaffenburg. Also 💯 agree with your points above 👍

bahaaTuffaha:
> > I'm not a fan of this change cuz its different from the new design and the webapp. These changes runs great overall.
> 
> I definitely get your point, I am usually the first that is aiming for consistency. However, it shouldn't be a self-purpose, i.e. consistency just for the goal of consistency. The goal is in the end always to improve the user experience and most of the time, consistency helps to achieve that by creating memorability for users. However, there are other factors to consider and sometimes other alternatives are better for the user experience. In this case, its more important to a) achieve consistency not with our web-app, but with other apps users are familiar with. Users expect to use the system share dialog (and to be able to choose whatever app they want), at least I don't know a single app that only presents me with preselected app options to share with. Also legal options are mostly part of the settings in some way or another, same goes for high contrast mode. b) avoid a cluttered menu and let users focus on the important and frequently used things. Users will basically never press any item in the legal section and rarely (or usually exactly zero to one time) use the high contrast mode setting. On the other hand, we want to promote the feedback and share features, which become more visible if there are less items in the menu.
> 
> In the end, there are just differences between a web app and native apps and I think we have to accept that sometimes. And even better if that allows us to reduce code and complexity.
> 
> > ```
> > * Feedback menuItem at the landing page doesn't navigate to anything (fresh install)
> > ```
> 
> Thanks, fixed.

Thank you for this long response. 😅
Yes the user experience is the end goal 👍

PR #4031, opened 2026-04-15
4021: Fix search back navigation

### LINKED ISSUE #4021: Navigating back to search results is impossible
### Describe the Bug

When I click on a search result, I want to easily navigate back to the list of search results and possibly look at another result. This is currently not possible on native, instead I get redirected to the site before the search.

### Steps to Reproduce

1. Go to search
2. Enter a term
3. Click a search result
4. Go back
5. See the previous page, not the search results

### Expected Behavior

The back behavior should be consistent and always the previous page.

### Related Issues

N/A

### Additional Information

N/A

steffenkleinle:
### Short Description

This PR fixes back navigation after clicking on a search result to the list of search results.

### Proposed Changes

- Fix back navigation to search results
- Rename `SearchModal` to `Search` as it is not a modal anymore

### Side Effects

N/A

### Testing

See issue.

### Resolved Issues

Fixes: #4021

---

### DIFF
--- a/native/src/utils/navigation.ts
+++ b/native/src/utils/navigation.ts
@@ -34,6 +34,7 @@ export const navigateNested = <T extends RoutesType, S extends keyof NestedRoute
 ): void => {
   if (navigation.getId() === ROOT_NAVIGATOR_ID) {
     const bottomTabRouteOpened = navigation.getState().routes.some(({ name }) => name === BOTTOM_TAB_NAVIGATION_ROUTE)
+
     if (redirect) {
       // Allow going back to the dashboard if opening a deep link
       navigation.replace(BOTTOM_TAB_NAVIGATION_ROUTE, {
@@ -44,7 +45,7 @@ export const navigateNested = <T extends RoutesType, S extends keyof NestedRoute
       })
     }
 
-    const navigate = bottomTabRouteOpened ? navigation.replace : navigation.push
+    const navigate = bottomTabRouteOpened && redirect ? navigation.replace : navigation.push
     navigate(BOTTOM_TAB_NAVIGATION_ROUTE, {
       screen: tabRoutes[route],
       params: {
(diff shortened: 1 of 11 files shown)


native/src/utils/navigation.ts
```diff
@@ -33,7 +33,8 @@ export const navigateNested = <T extends RoutesType, S extends keyof NestedRoute
   redirect: boolean,
 ): void => {
   if (navigation.getId() === ROOT_NAVIGATOR_ID) {
-    const bottomTabRouteOpened = navigation.getState().routes.some(({ name }) => name === BOTTOM_TAB_NAVIGATION_ROUTE)
+    const routes = navigation.getState().routes
```
steffenkleinle: This is the actual fix, the search is part of the root navigator and there is always a bottom tab route present for the search. However, we only need to replace the bottom tab route if that is also the top route. Otherwise push the new route.

lunars97: Really nice!

lunars97 (approved):
Works as expected! Tested with iOS simulator 👍

bahaaTuffaha (commented):
I think we are back to the same issue as before with deeplinks
```
npx uri-scheme open "integreat://integreat.app/lkaugsburg/de/willkommen-lkaugsburg/willkommen" --android
```
another back after going back from a deeplink (foreground)

Image

steffenkleinle:
> I think we are back to the same issue as before with deeplinks
> 
> ```
> npx uri-scheme open "integreat://integreat.app/lkaugsburg/de/willkommen-lkaugsburg/willkommen" --android
> ```
> 
> another back after going back from a deeplink (foreground)

Very good idea to check that 🙈 Now it should hopefully be fixed...

bahaaTuffaha (approved):
Nice work!
Tested the search modal on Android and iOS. ✅ 
Tested the deeplink ✅ 

one small thing :
The back title should be "search" right? because now we get back to the search modal not the previous page.

Image

steffenkleinle:
> one small thing : The back title should be "search" right? because now we get back to the search modal not the previous page.

True, I'll fix that in #4032.

PR #4033, opened 2026-04-15
3984: Announce new chat answers with screen reader

### LINKED ISSUE #3984: Announce new answers to screen reader
### Describe the Problem

As a user using Frag Integreat with a screen reader, I'd like to get notified about new answers from the bot or counselors. The answers should then also be read aloud.

Announce new answers to the screen reader.

N/A

### Related Issues

N/A

### Additional Information

N/A

lunars97:
### Short Description

This PR implements announcement of chat answers

### Proposed Changes

- create LiveAnnouncer.tsx component which stays mounted. Screen-readers announce changes only inside already existing regions and hide it visually
- set `status` role which implicitly has aria-live='polite' and aria-atomic='true'
- strip html tags for screen-reader (links were containing html tags)

### Side Effects

should be none

N/A

### Testing
```
on Linux - use Orca
on MacOS - use VoiceOver
on Windows - use JAWS / NVDA
```

- Go to the chat -> enable screen-reader -> send a message -> wait for response

### Resolved Issues

Fixes: #3984 

---

### DIFF
--- a/web/src/components/ChatConversation.tsx
+++ b/web/src/components/ChatConversation.tsx
@@ -2,12 +2,13 @@ import List from '@mui/material/List'
 import Stack from '@mui/material/Stack'
 import Typography from '@mui/material/Typography'
 import { styled } from '@mui/material/styles'
-import React, { ReactElement, useEffect, useRef, useState } from 'react'
+import React, { ReactElement, useEffect, useMemo, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { ChatMessageModel } from 'shared/api'
 
 import ChatMessage from './ChatMessage'
+import LiveAnnouncer from './LiveAnnouncer'
 import SkeletonChatConversation from './SkeletonChatConversation'
 import TypingIndicator from './TypingIndicator'
 
@@ -45,9 +46,19 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
+  const lastMessageText = useMemo(() => {
+    if (!messages.length) {
+      return ''
+    }
+
+    const last = messages[messages.length - 1]
+    return last?.content
+  }, [messages])
+
   if (messages.length === 0 && !loading) {
     return (
       <Stack paddingInline={3} gap={1}>
+        <LiveAnnouncer message='' />
         <Typography variant='subtitle1'>{t('conversationText')}</Typography>
       </Stack>
     )
@@ -55,11 +66,12 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
 
   return (
     <Stack padding={2} gap={2} overflow='auto'>
+      <LiveAnnouncer message={lastMessageText ?? ''} />
       {loading ? (
         <SkeletonChatConversation />
       ) : (
         <>
-          <StyledList disablePadding aria-relevant='additions'>
+          <StyledList disablePadding>
             {messages.map((message, index) => (
               <ChatMessage message={message} key={message.id} previousMessage={messages[index - 1]} />
             ))}
--- a/web/src/components/ChatConversation.tsx
+++ b/web/src/components/ChatConversation.tsx
@@ -5,6 +5,7 @@ import { styled } from '@mui/material/styles'
 import React, { ReactElement, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import { parseHTML } from 'shared'
 import { ChatMessageModel } from 'shared/api'
 
 import ChatMessage from './ChatMessage'
@@ -46,14 +47,8 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
-  const extractText = (html: string): string => {
-    const div = document.createElement('div')
-    div.innerHTML = html
-    return div.textContent || div.innerText
-  }
-
   const last = messages[messages.length - 1]
-  const lastMessageText = last ? extractText(last.content) : ''
+  const lastMessageText = last ? parseHTML(last.content) : ''
 
   if (messages.length === 0 && !loading) {
     return (
--- a/web/src/components/ChatConversation.tsx
+++ b/web/src/components/ChatConversation.tsx
@@ -2,7 +2,7 @@ import List from '@mui/material/List'
 import Stack from '@mui/material/Stack'
 import Typography from '@mui/material/Typography'
 import { styled } from '@mui/material/styles'
-import React, { ReactElement, useEffect, useMemo, useRef, useState } from 'react'
+import React, { ReactElement, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { ChatMessageModel } from 'shared/api'
@@ -46,19 +46,18 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
-  const lastMessageText = useMemo(() => {
-    if (!messages.length) {
-      return ''
-    }
+  const extractText = (html: string): string => {
+    const div = document.createElement('div')
+    div.innerHTML = html
+    return div.textContent || div.innerText
+  }
 
-    const last = messages[messages.length - 1]
-    return last?.content
-  }, [messages])
+  const last = messages[messages.length - 1]
+  const lastMessageText = last ? extractText(last.content) : ''
 
   if (messages.length === 0 && !loading) {
     return (
       <Stack paddingInline={3} gap={1}>
-        <LiveAnnouncer message='' />
         <Typography variant='subtitle1'>{t('conversationText')}</Typography>
       </Stack>
     )
@@ -66,7 +65,7 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
 
   return (
     <Stack padding={2} gap={2} overflow='auto'>
-      <LiveAnnouncer message={lastMessageText ?? ''} />
+      <LiveAnnouncer message={lastMessageText} />
       {loading ? (
         <SkeletonChatConversation />
       ) : (
--- a/web/src/components/ChatConversation.tsx
+++ b/web/src/components/ChatConversation.tsx
@@ -59,7 +59,7 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
         <SkeletonChatConversation />
       ) : (
         <>
-          <StyledList disablePadding>
+          <StyledList disablePadding aria-relevant='additions'>
             {messages.map((message, index) => (
               <ChatMessage message={message} key={message.id} previousMessage={messages[index - 1]} />
             ))}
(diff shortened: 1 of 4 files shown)


web/src/components/ChatConversation.tsx
```diff
@@ -45,16 +46,27 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
+  const lastMessageText = useMemo(() => {
```
steffenkleinle: ❓ Why do we need a memo here? `useMemo` [should only be used for resource-/calculation-intensive code](https://react.dev/reference/react/useMemo#skipping-expensive-recalculations).

> You should only rely on useMemo as a performance optimization. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

web/src/components/ChatConversation.tsx
```diff
@@ -45,16 +46,27 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
...
+    }
+
+    const last = messages[messages.length - 1]
+    return last?.content
+  }, [messages])
+
   if (messages.length === 0 && !loading) {
     return (
       <Stack paddingInline={3} gap={1}>
+        <LiveAnnouncer message='' />
```
steffenkleinle: 🔧 I'd very much prefer if we'd explicitly pass `null` in case there is no new message.

🤔 Is it necessary that we render this here at all?

web/src/components/LiveAnnouncer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  height: '1px',
+  width: '1px',
+  overflow: 'hidden',
+}))
+
+type LiveAnnouncerProps = {
+  message: string
+}
+const LiveAnnouncer = ({ message }: LiveAnnouncerProps): ReactElement => {
+  const [announced, setAnnounced] = useState('')
```
steffenkleinle: ❓ Why is it necessary that we save it to a separate state if it just gets updated to the passed prop anyway?

web/src/components/LiveAnnouncer.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  useEffect(() => {
+    if (!message) {
+      return
+    }
+
+    setAnnounced('')
+    setAnnounced(message)
```
steffenkleinle: ❓ Why do we need to set `announced` to an empty string first?

web/src/components/LiveAnnouncer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }, [message])
+
+  return <VisuallyHidden role='status'>{announced}</VisuallyHidden>
```
steffenkleinle: 🤔 Couldn't we just directly add `role='status'` to the last message instead?

steffenkleinle:
> strip html tags for screen-reader (links were containing html tags)

Where is this happening? Maybe I am just missing it, but I don't see it in your changes 🤔

lunars97:
> > strip html tags for screen-reader (links were containing html tags)
> 
> Where is this happening? Maybe I am just missing it, but I don't see it in your changes 🤔

seems like I overwritten it with rebase by accident :(

web/src/components/ChatConversation.tsx
lunars97: erased by accident new changes by rebase and it shouldn't be there 👍

web/src/components/ChatConversation.tsx
lunars97: not needed anymore 👍

web/src/components/LiveAnnouncer.tsx
lunars97: yes, it is not necessary. I have been trying to find out why it was not working with screen readers. I have cleaned up everthing 👍

web/src/components/LiveAnnouncer.tsx
lunars97: not needed anymore

web/src/components/LiveAnnouncer.tsx
lunars97: no, it does not work for dynamic content. It should already exist

web/src/components/ChatConversation.tsx
```diff
@@ -45,6 +46,15 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
+  const extractText = (html: string): string => {
+    const div = document.createElement('div')
+    div.innerHTML = html
```
bahaaTuffaha: ❓ could execute scripts (XSS).
or considered safe?

bahaaTuffaha:
> * Go to the chat -> enable screen-reader -> send a message -> wait for response

I enabled voice over with ctrl + F5 then waited for a response but it didn't read it 🤔

lunars97:
> > * Go to the chat -> enable screen-reader -> send a message -> wait for response
> 
> I enabled voice over with ctrl + F5 then waited for a response but it didn't read it 🤔

hmmm interesting, I also did the same and it is reading out the messages 👀

web/src/components/ChatConversation.tsx
```diff
@@ -45,6 +46,15 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
...
+    return div.textContent || div.innerText
+  }
+
+  const last = messages[messages.length - 1]
+  const lastMessageText = last ? extractText(last.content) : ''
```
steffenkleinle: ❌ We already have helpers for that, see `parseHTML`.

web/src/components/LiveAnnouncer.tsx
```diff
@@ -0,0 +1,19 @@
+import { styled } from '@mui/material/styles'
+import React, { ReactElement } from 'react'
+
+const VisuallyHidden = styled('div')(() => ({
+  position: 'absolute',
+  left: '-9999px',
+  height: '1px',
+  width: '1px',
+  overflow: 'hidden',
```
steffenkleinle: 🤔 `visibility: 'hidden'` doesn't work? Is it actually necessary to set a height and width?

steffenkleinle (changes requested):

web/src/components/ChatConversation.tsx
lunars97: not sure, we are not appending here anything :thinking:

steffenkleinle: But we set users messages as HTML which could be malicious :)

web/src/components/LiveAnnouncer.tsx
lunars97: sadly no, it removes completely from accessibility tree and yes, height and width should be there to be detectable for screen readers

web/src/components/LiveAnnouncer.tsx
```diff
@@ -0,0 +1,23 @@
...
+const VisuallyHidden = styled('div')(() => ({
+  border: 0,
+  clip: 'rect(0 0 0 0)',
+  clipPath: 'inset(50%)',
+  height: '1px',
+  width: '1px',
+  margin: 0,
+  overflow: 'hidden',
+  position: 'absolute',
+  whiteSpace: 'nowrap',
```
lunars97, line 15: preffered clipping method https://www.a11y-collective.com/blog/visually-hidden/
`clip-path: inset(50%) clips the visible area to almost nothing.
white-space: nowrap stops text wrapping that might expose content.
`

web/src/components/ChatConversation.tsx
lunars97: replace with parseHTML which tackles it :+1:

web/src/components/ChatConversation.tsx
```diff
@@ -45,6 +47,9 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
+  const last = messages[messages.length - 1]
+  const lastMessageText = last ? parseHTML(last.content) : ''
```
bahaaTuffaha: ❓ do you think `optimizeNumbersForTts` parameter could be needed here?

web/src/components/LiveAnnouncer.tsx
bahaaTuffaha, line 1: it says 9 items when I focused on the input:

Image

bahaaTuffaha, line 1: It repeats the user message twice:

https://github.com/user-attachments/assets/12b77c30-e8b7-4987-b08d-722d713bdaa2

web/src/components/LiveAnnouncer.tsx
steffenkleinle: What do you think of adding that link as a comment?

web/src/components/ChatConversation.tsx
lunars97: I think we dont need it here, tts is also not available on the chat dialog, so it makes no sense? 👍

web/src/components/LiveAnnouncer.tsx
lunars97: Seems like it is an expected behavior because of the voiceOver settings, if you go to `verbose` and set it to medium instead of high, it won't announce it... It says 9 items in a dialog which are focusable

lunars97: I think it is because of the ui and then aria-live region. 
UPDATE: hmm but with orca it does not announce it twice, so maybe it is a default behavior of the voiceOver 🤔

bahaaTuffaha: https://stackoverflow.com/questions/61244988/aria-live-content-read-twice-by-mac-chrome-voiceover

lunars97: but in our case, there is no iframe...

bahaaTuffaha: > but in our case, there is no iframe...

Yes correct 🤔

bahaaTuffaha (approved):
The repeated message looks unsolvable. we can just skip the userIsAuthor messages 🤔 

Tested Firefox and chrome ✅

lunars97:
> The repeated message looks unsolvable. we can just skip the userIsAuthor messages 🤔
> 
> Tested Firefox and chrome ✅

I would keep the announcement of users' message, imho it keeps the flow...

web/src/components/ChatConversation.tsx
steffenkleinle: @lunars97 I'd say this depends a lot on how numbers in chat messages are now read aloud? If they are read the same as with our own tts function (i.e. reading postcodes or phone numbers as one number instead of multiple single numbers), then we should also use the parameter here. We should also check this on all platforms. I'd be a little in favor of just using the parameter anyway, since it definitely does not cause any problems but might fix problems on some platforms. Nice catch @bahaaTuffaha!

web/src/components/ChatConversation.tsx
```diff
@@ -45,6 +47,9 @@ const ChatConversation = ({ messages, isTyping, loading }: ChatConversationProps
     }
   }, [isTyping])
 
+  const last = messages[messages.length - 1]
+  const lastMessageText = last ? parseHTML(last.content, true) : ''
```
steffenkleinle: 🙃 
```suggestion
  const lastMessage = messages[messages.length - 1]
  const lastMessageText = lastMessage ? parseHTML(lastMessage.content, true) : ''
```

steffenkleinle (approved):
Tested on MacOS Firefox, nice work!

PR #4044, opened 2026-04-16
3838: Show search results in users language

### LINKED ISSUE #3838: German search term should lead to result in set language
### Describe the Problem

A user navigates Integreat e.g. in English. He/she types a German term into the search bcs it is a word they read in a letter or know from previous migration contect e.g. Aufenthaltstitel. Currently they are directed to the German page in the search results.

The search results should lead to a page in the language that the user is using - possibly with some kind of indication that this page was found based on the German translation and its matching with the search term (if deemed necessary)

Let user change the language back to their originally set language (status quo)

steffenkleinle:
### Short Description

Show all search results in the users language (if possible).

### Proposed Changes

- Add `useMultiLanguageSearch` hook to search in multiple languages and return results in the users language (if possible)
- Return results in German if there is no translation of the result

### Side Effects

N/A

### Testing

Search for a German term that is not included in the users language version and make sure that the result is still included (in the users language). Also make sure that the German search still works as expected.

### Resolved Issues

Fixes: #3838

---

### DIFF
--- a/release-notes/unreleased/3838-search-language.yml
+++ b/release-notes/unreleased/3838-search-language.yml
@@ -1,0 +1,8 @@
+issue_key: 3838
+show_in_stores: true
+platforms:
+  - web
+  - android
+  - ios
+en: The search now returns all results with translation in your language.
+de: Die Suche liefert nun alle Ergebnisse mit einer Übersetzung in deiner Sprache.
--- a/shared/hooks/useSearch.ts
+++ b/shared/hooks/useSearch.ts
@@ -1,4 +1,4 @@
-import MiniSearch from 'minisearch'
+import MiniSearch, { SearchResult } from 'minisearch'
 import { useEffect, useState } from 'react'
 
 import CategoriesMapModel from '../api/models/CategoriesMapModel'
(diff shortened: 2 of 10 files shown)


release-notes/unreleased/3838-search-language.yml
```diff
@@ -0,0 +1,8 @@
+issue_key: 3838
+show_in_stores: true
+platforms:
+  - web
+  - android
+  - ios
+en: The search now returns returns all results with translation in your language.
```
bahaaTuffaha: 🔧  
```suggestion
en: The search now returns all results with translation in your language.
```

shared/hooks/useSearch.ts
```diff
@@ -72,19 +73,63 @@ const useSearch = (documents: ExtendedPageModel[], query: string): UseSearchRetu
...
+  error: Error | null
+  loading: boolean
+}
+
+// Performs a search in documents of multiple languages (usually the users language and the source language (de))
+// If the search term is found in a non-user language document, we still return the document translated in the users language
+// If no translation to the users language exists, we use the original language of the search result
+const useMultiLanguageSearch = ({
+  userLanguageDocuments,
+  moreDocuments,
```
bahaaTuffaha, line 101: I think `moreDocuments` is a bit ambiguous 🤔

shared/hooks/useSearch.ts
```diff
@@ -72,19 +73,63 @@ const useSearch = (documents: ExtendedPageModel[], query: string): UseSearchRetu
...
+  const moreSearch = useSearch(moreDocuments, query)
+
+  const userLanguageResults = userLanguageSearch.data
+  const moreResults = moreSearch.data.map(result => ({
+    ...result,
+    id: result.availableLanguages[userLanguageCode] ?? result.id,
+  }))
+
+  const results = [...userLanguageResults, ...moreResults].sort((a, b) => b.score - a.score).map(result => result.id)
+  const uniqueResults = results.filter((result, index) => index === results.indexOf(result))
```
bahaaTuffaha: You can use `Set()` to reduce complexity:

```suggestion
  const uniqueResults = [...new Set(results)]
```

bahaaTuffaha (approved):
LGTM ✅ 
Tested on Android and Chrome.

shared/hooks/useSearch.ts
steffenkleinle: Yeah I wasn't to sure about that one, just thought its clear by `userLanguageDocuments` and `moreDocuments` and couldn't come up with a good alternative. Do you have a suggestion?

lunars97: for me it is also self-explanatory and I can't think about any other alternative :)

lunars97 (approved):
Nice! Tested on android emularor and chrome 👍

Issue #4045, opened 2026-04-16
Streamline internal naming


steffenkleinle:
### Describe the Problem

We use quite a few terms very inconsistently, both in our own codebase and in regards to the CMS and other teams. This can lead to problems in understanding for both team members and other teams and is not very good practice.

### Describe Your Preferred Solution

Streamline and adjust the following naming:
- [ ] `city` -> `region`: #4071 
- [ ] `landing` -> `regionSelection`
- [ ] `poi` -> `place`
- [x] `disclaimer` -> `imprint`: #4046 
- [x] `cityNotCooperating` -> `suggestToRegion`: #4047 
- [x] `PageModel` -> `DocumentModel`: #4062
- [x] `Switcher` -> `Navigator`: #4063 and #4068 
- [ ] TBD: `category` -> `?`

### Describe Possible Alternatives

N/A

### Related Issues

#4038 

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/ugxogojzdf85zfudgszrn81asw
https://chat.tuerantuer.org/digitalfabrik/pl/khbpkozqrinrxesf3uktjrhtie
https://chat.tuerantuer.org/digitalfabrik/pl/igu1zf1xxtnhigwninsjiykqrc

PR #4047, opened 2026-04-16
4045: Rename city not cooperating to suggest to region

steffenkleinle:
### Short Description

Rename cityNotCooperating to suggestToRegion

### Proposed Changes

- Rename city not cooperating to suggest to region everywhere

### Side Effects

N/A

### Testing

Since all of these changes are just find + replace, no need to check the individual changes carefully. Perhaps check more just the files if they should be modified.
Smoke test the app on web and native and check out the suggest to region feature.

### Resolved Issues

Partly fixes: #4045

---


native/src/constants/buildConfig.ts
```diff
@@ -11,14 +11,15 @@ import aschaffenburgLoadingImage from 'build-configs/aschaffenburg/assets/app-ic
 import aschaffenburgAppIcon from 'build-configs/aschaffenburg/assets/app-icon-round.svg'
 import integreatLoadingImage from 'build-configs/integreat/assets/app-icon-inverted.svg'
 import integreatAppIcon from 'build-configs/integreat/assets/app-icon-round.svg'
-import integreatCityNotCooperatingIcon from 'build-configs/integreat/assets/city-not-cooperating.svg'
 import malteLoadingImage from 'build-configs/malte/assets/app-icon-circle.svg'
 import malteAppIcon from 'build-configs/malte/assets/app-icon-round.svg'
 
+import integreatSuggestToRegionIcon from '../../../build-configs/integreat/assets/suggest-to-region.svg'
```
bahaaTuffaha: why you wrote the path relatively instead of using`build-configs/integreat/...`?

bahaaTuffaha (approved):
Nice work!

- How about `recommendRegion`?
- Native border still yellow.
- `CityNotCooperatingFooter.spec.tsx` -> `SuggestToRegionFooter.spec.tsx`
I smoke tested the app ✅

native/src/constants/buildConfig.ts
steffenkleinle: Not on purpose, good catch. I'll fix it!

steffenkleinle:
>     * How about `recommendRegion`?

Thought about that too, but its a little ambiguous and sounds more like we are recommending a region to the user. I am open for that as well though.

>     * Native border still yellow.

I fixed that in #4048, wanted to keep this PR just about renaming.
 
>     * `CityNotCooperatingFooter.spec.tsx` -> `SuggestToRegionFooter.spec.tsx`

Another good catch, thanks!

lunars97 (approved):
LGTM!

PR #4050, opened 2026-04-16
2550: Show info on unavailable languages

### LINKED ISSUE #2550: More information when translation is not available for certain languages
I have received this feedback from the project "Digital Active Women": 
They have emphasized that they notice when a page is available in German but not in their own language. It makes them feel as if others would decide what is relevant to them and what they can or can not know. 

We could change this for example by including a short information text, when no translation is available such as: 

"Sorry. This page does not exist in your language as it has not been translated yet. Click here to read the [German/English] page."

bahaaTuffaha:
### Short Description

1. When a user selects a language that lacks available translations we should continue to grey out unavailable languages to signal their status, ensuring users understand which languages are accessible.

2. If the user clicks on a greyed-out language (e.g., "Polski"), a modal window appears with a message.
   The modal provides context: "Language currently not available. We are actively expanding our language options based on user needs. If you'd like to request a specific language, please use the feedback feature."

### Proposed Changes

Web:
- Added `LanguageNotFoundItem` component in LanguageList.tsx that appears when search returns no matching languages.
- used `noOptionsText` so it shows the item when there is no search result.
 - Adjusted `AlertDialog`  to hide the close button on mobile screen as it takes space when the dialog message is too big. 
 

Native:
- Created a base component `AlertDialog` for native and used text instead of `Dialog.text` to satisfy AppointmentOnlyOverlay dialog that uses text Variant.
- Added LanguageNotFoundItem using SelectorItem that appears when search returns no results

### Side Effects

None.

### Testing

- Launch the webapp.
- Check the search results after you enter unavailable language.
- Test on mobile screens sizes.
- Test the same thing for native (android and ios).
- Test AppointmentOnlyOverlay.

### Resolved Issues

Fixes: #2550

---


native/src/components/base/AlertDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const AlertDialog = ({ visible, onDismiss, title, children, titleTextVariant }: AlertDialogProps): ReactElement => {
+  const { t } = useTranslation('common')
+  const theme = useTheme()
+
+  return (
+    <Portal>
+      <Dialog visible={visible} onDismiss={onDismiss} style={{ backgroundColor: theme.colors.surfaceVariant }}>
+        <Text variant={titleTextVariant} style={{ marginTop: 22, marginBottom: 18, marginHorizontal: 24 }}>
```
steffenkleinle: ❓ Do we now just use inline styles? Was there a decision for this at some point? Please use style sheets or styled components instead :)

native/src/components/base/AlertDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useTranslation } from 'react-i18next'
+import { Button, Dialog, Portal, useTheme } from 'react-native-paper'
+
+import { TypographyVariant } from 'build-configs/TypographyType'
+
+import Text from './Text'
+
+type AlertDialogProps = {
+  visible: boolean
+  onDismiss: () => void
```
steffenkleinle: 🙃  So far we usually used the naming `close`
```suggestion
  close: () => void
```

native/src/components/base/AlertDialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+  title: string
+  children: ReactElement
+  titleTextVariant?: TypographyVariant
```
steffenkleinle: 🔧 Imo the better solution is to just allow passing a custom `title` component instead, i.e. `title: string | ReactElement`. Otherwise, we need many different props if more than just the variant changes (e.g. color, paddings, ...).

native/src/components/AppointmentOnlyOverlay.tsx
```diff
@@ -18,27 +17,20 @@ const AppointmentOnlyOverlay = ({
...
+      visible={isVisible}
+      onDismiss={closeOverlay}
+      title={t('appointmentNecessary')}>
+      <Text>
+        <Trans i18nKey='pois:makeAppointmentTooltipWithLink'>
+          This gets replaced
+          {appointmentUrl ? <Link url={appointmentUrl}>by react-i18next</Link> : <Text>by react-i18next</Text>}
+        </Trans>
+      </Text>
+    </AlertDialog>
```
steffenkleinle, line 32: 👍 Nice refactoring and reusing of components 🎉

web/src/components/LanguageListItem.tsx
```diff
@@ -43,11 +49,14 @@ const LanguageListItem = ({ code, path, name, close, selectedLanguageCode }: Sel
       </SelectorItemButton>
     </ListItem>
   ) : (
-    <Tooltip key={code} title={t('noTranslation')} placement={contentDirection === 'ltr' ? 'right' : 'left'}>
-      <ListItem disablePadding>
-        <SelectorItemButton disabled>{name}</SelectorItemButton>
-      </ListItem>
-    </Tooltip>
+    <ListItem disablePadding>
```
steffenkleinle: 🔧 
```suggestion
    <ListItem key={code} disablePadding>
```

translations/translations.json
```diff
@@ -8132,6 +8134,8 @@
       "news": "News",
       "noLanguages": "Not available in other languages",
       "noTranslation": "No translation available",
+      "languageNotAvailableTitle": "Language currently not available",
```
steffenkleinle: 🤔 I think this is quite lengthy as title, why not just
```suggestion
      "languageNotAvailableTitle": "Language not available",
```
🤔 Also, can't we just use `noTranslation`, i..e `No translation available`?

translations/translations.json
```diff
@@ -8132,6 +8134,8 @@
...
+      "languageNotAvailableMessage": "Content may not be available in all languages yet. We are actively expanding our language options based on user needs. If you'd like to request a specific language, please use the feedback feature.",
```
steffenkleinle: 🔧 
```suggestion
      "languageNotAvailableMessage": "This content is not yet available in all languages. We are actively expanding our language options based on user needs. If you'd like to request a specific language, please use the feedback feature.",
```

web/src/components/base/AlertDialog.tsx
```diff
@@ -40,9 +48,11 @@ const AlertDialog = ({ title, close, children, className, actions }: DialogProps
         <DialogTitle component='h2' variant='h4' textOverflow='ellipsis' whiteSpace='nowrap' overflow='hidden'>
           {title}
         </DialogTitle>
-        <IconButton aria-label={t('common:close')} onClick={close}>
-          <CloseIcon />
-        </IconButton>
+        {!hideCloseButton && (
```
steffenkleinle: 🤔 Shouldn't we rather just disable this in general on mobile then instead of making it configurable and thus inconsistent?
```suggestion
        {desktop && (
```

native/src/components/SelectorItem.tsx
```diff
@@ -22,41 +51,50 @@ const SelectorItem = ({
...
-        </Text>
-      }
-    />
-  )
+  const infoIcon =
+    !enabled && !selected ? (
+      <View style={styles.infoIconWrapper}>
+        <Icon
+          source='information-outline'
+          size={INFO_ICON_SIZE}
```
steffenkleinle: 🔧 That is the default icon size, isn't it?
```suggestion
```

native/src/components/SelectorItem.tsx
```diff
@@ -22,41 +51,50 @@ const SelectorItem = ({
...
+          color={theme.colors.onSurfaceDisabled}
+          label='informationIcon'
```
steffenkleinle: 🤔 Does this label add any meaningful info for users?
```suggestion
```

native/src/components/SelectorItem.tsx
```diff
@@ -1,10 +1,39 @@
...
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  text: {
+    fontSize: 20,
+    textAlign: 'center',
+  },
+  infoIconWrapper: {
+    marginLeft: 4,
```
steffenkleinle: 🔧 Let's use a gap on the parent instead

native/src/routes/ChangeLanguageModal.tsx
```diff
@@ -53,24 +56,34 @@ const ChangeLanguageModal = ({ navigation, route }: ChangeLanguageModalProps): R
...
 
   useEffect(() => {
     loadPolyfillIfNeeded(languageCode).then(() => setPolyfillLoaded(true))
   }, [languageCode])
 
+  const allLanguages = useMemo(() => {
+    const cityCodes = new Set(languages.map(language => language.code))
+    const unsupported = Object.entries(config.supportedLanguages)
+      .filter(([code]) => !cityCodes.has(code))
+      .map(([code, { name }]) => ({ code, name }) as LanguageModel)
```
steffenkleinle: 🔧 No typecasts please
```suggestion
      .map(([code, { name }]) => new LanguageModel(code, name))
```

native/src/routes/ChangeLanguageModal.tsx
```diff
@@ -53,24 +56,34 @@ const ChangeLanguageModal = ({ navigation, route }: ChangeLanguageModalProps): R
...
   const currentLanguageName = languages.find(lang => lang.code === languageCode)?.name
 
   useEffect(() => {
     loadPolyfillIfNeeded(languageCode).then(() => setPolyfillLoaded(true))
   }, [languageCode])
 
+  const allLanguages = useMemo(() => {
+    const cityCodes = new Set(languages.map(language => language.code))
+    const unsupported = Object.entries(config.supportedLanguages)
+      .filter(([code]) => !cityCodes.has(code))
```
steffenkleinle: 🤔 Why do we need to convert languages to a set to filter?
```suggestion
      .filter(([code]) => !languages.find(language => language.code === code))
```

web/src/components/LanguageListItem.tsx
```diff
@@ -43,11 +49,14 @@ const LanguageListItem = ({ code, path, name, close, selectedLanguageCode }: Sel
...
+      <SelectorItemButton style={{ gap: 8 }} onClick={onUnavailableLanguageClick}>
```
steffenkleinle: 🔧 Please don't use inline styles

web/src/components/LanguageList.tsx
```diff
@@ -64,25 +65,35 @@ const LanguageList = ({
...
 }: LanguageListProps): ReactElement => {
   const [query, setQuery] = useState('')
   const { t } = useTranslation('layout')
   const { mobile } = useDimensions()
 
-  const allOptions = useMemo(
-    () => languageChangePaths.filter(item => !availableOnly || !!item.path),
-    [languageChangePaths, availableOnly],
-  )
+  const languageOptions = useMemo(() => {
```
steffenkleinle: ❓ Why do we need a memo here? `useMemo` [should only be used for resource-/calculation-intensive code](https://react.dev/reference/react/useMemo#skipping-expensive-recalculations).

> You should only rely on useMemo as a performance optimization. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

steffenkleinle (changes requested):
Nice work already, looks great! I opened a discussion in https://github.com/digitalfabrik/integreat-app/issues/2550#issuecomment-4279647880 about whether we should actually display all languages. In any case, we should probably sort the languages somehow, maybe just alphabetically.

native/src/components/SelectorItem.tsx
bahaaTuffaha: Yes

web/src/components/LanguageList.tsx
bahaaTuffaha: 🤔  Not sure if it affects performance I just changed the name

steffenkleinle:
Sorry, I deleted the branch by mistake :(

web/src/components/LanguageList.tsx
steffenkleinle: Please remove the `useMemo`, same as above. The docs clearly state that it should only be used in case you experience performance issues otherwise. Since this is not the case here (at least not for me), we also should not use this. The amount of languages are in the low two digit area, there should definitely not be any performance issues.

translations/translations.json
```diff
@@ -7814,15 +7814,17 @@
...
-      "imprint": "Impressum",
+      "disclaimer": "Impressum",
       "privacy": "Datenschutz",
       "localInformation": "Lokale Informationen",
       "localInformationLabel": "Informationen",
       "events": "Veranstaltungen",
       "news": "Nachrichten",
       "noLanguages": "Nicht verfügbar in anderen Sprachen",
       "noTranslation": "Keine Übersetzung verfügbar",
+      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv basierend auf den Bedürfnissen unserer Nutzer. Wenn Sie eine bestimmte Sprache anfragen möchten, nutzen Sie bitte die Feedback-Funktion.",
```
steffenkleinle: ❌ We usually use the personal `du`:)
```suggestion
      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv basierend auf den Bedürfnissen unserer Nutzer*innen. Wenn Sie eine bestimmte Sprache anfragen möchten, nutze bitte die Feedback-Funktion.",
```
@ulyanahoy-cloud do you perhaps have any better idea for the text here? I am not that happy with `basierend auf den Bedürfnissen unserer Nutzer*innen`, that sounds like usual marketing slang (like the person reading is the user).

translations/translations.json
```diff
@@ -7814,15 +7814,17 @@
...
       "noLanguageFound": "Keine Sprache gefunden",
+      "languageNotFoundQuestion": "Ihre Sprache nicht gefunden?",
```
steffenkleinle: ❌ 
```suggestion
      "languageNotFoundQuestion": "Deine Sprache nicht gefunden?",
```

native/src/routes/ChangeLanguageModal.tsx
```diff
@@ -93,21 +99,40 @@ const ChangeLanguageModal = ({ navigation, route }: ChangeLanguageModalProps): R
...
+          value={query}
+          right={() => undefined}
+          inputStyle={{ color: theme.dark ? theme.colors.background : theme.colors.onBackground }}
+          style={{
+            marginHorizontal: 16,
+            backgroundColor: theme.dark ? theme.colors.tertiary : theme.colors.surfaceVariant,
+          }}
+        />
+        {filteredLanguages.length === 0 ? (
+          <SelectorItem model={languageNotFoundItem} selected={false} />
```
steffenkleinle: ❓ Hmm, I don't think we should use a `SelectorItem` for this. Can we just use a normal button instead at the end of the list? I think we can also always display it, not just when there are no languages found.

native/src/routes/ChangeLanguageModal.tsx
```diff
@@ -93,21 +99,40 @@ const ChangeLanguageModal = ({ navigation, route }: ChangeLanguageModalProps): R
...
+      <Wrapper contentContainerStyle={{ flexGrow: 1, justifyContent: 'flex-start', gap: 4 }}>
+        <Searchbar
+          placeholder={currentLanguageName}
+          onChangeText={setQuery}
+          value={query}
+          right={() => undefined}
+          inputStyle={{ color: theme.dark ? theme.colors.background : theme.colors.onBackground }}
+          style={{
+            marginHorizontal: 16,
+            backgroundColor: theme.dark ? theme.colors.tertiary : theme.colors.surfaceVariant,
```
steffenkleinle: 💭 The contrast of the mag icon and the placeholder text with high contrast mode is not very good imo. Would you mind adjusting it?

Image

web/src/components/LanguageList.tsx
```diff
@@ -88,24 +105,22 @@ const LanguageList = ({
...
-                  close={close}
-                  selectedLanguageCode={currentLanguage?.code}
-                />
-              ))}
-            </List>
-          )}
-        </>
+        {filteredLanguageChangePaths.length === 0 ? (
+          <LanguageNotFoundItem onClick={onUnavailableLanguageClick} label={t('languageNotFoundQuestion')} />
+        ) : (
```
steffenkleinle: ❌  Same as for native, imo we should not use a list item for this and instead just use a regular button (similar to the result not found button in the normal search, except that the question is the button text).

Image

translations/translations.json
ulyanahoy-cloud: @steffenkleinle How about a little shorter: "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Sie können diese Sprache über die Feedback-Funktion anfragen."
Is it possible to make the Feedback text a link?

steffenkleinle: Then lets use the following (`du` instead of `Sie`):
```suggestion
      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Sie können diese Sprache über die Feedback-Funktion anfragen.",
```
>  Is it possible to make the Feedback text a link?

Not yet, but I'll create an issue for it, good idea and thought the same.

native/src/routes/ChangeLanguageModal.tsx
bahaaTuffaha: Yes I agree 100%

bahaaTuffaha:
> [2550: Open unavailable dialog on unavailable language click](https://github.com/digitalfabrik/integreat-app/pull/4050/commits/095795efce2a675e0a1a7c40824bbb965e53cf68)

I forgot that some pages could have unavailable languages for certain pages for ex.:
http://localhost:9000/testumgebung/de/sprache

translations/translations.json
```diff
@@ -7814,15 +7814,17 @@
...
-      "imprint": "Impressum",
+      "disclaimer": "Impressum",
       "privacy": "Datenschutz",
       "localInformation": "Lokale Informationen",
       "localInformationLabel": "Informationen",
       "events": "Veranstaltungen",
       "news": "Nachrichten",
       "noLanguages": "Nicht verfügbar in anderen Sprachen",
       "noTranslation": "Keine Übersetzung verfügbar",
+      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Sie können diese Sprache über die Feedback-Funktion anfragen.",
```
steffenkleinle: ❌ 
```suggestion
      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Du kannst diese Sprache über die Feedback-Funktion anfragen.",
```

native/src/routes/ChangeLanguageModal.tsx
```diff
@@ -53,61 +69,82 @@ const ChangeLanguageModal = ({ navigation, route }: ChangeLanguageModalProps): R
...
+            { backgroundColor: theme.dark ? theme.colors.tertiary : theme.colors.surfaceVariant },
+          ]}
+        />
+        {filteredLanguages.length === 0 ? (
+          <Button mode='outlined' onPress={() => setIsUnavailableDialogOpen(true)} style={styles.horizontalMargin}>
+            {t('languageNotFoundQuestion')}
+          </Button>
+        ) : (
+          <Selector selectedItemCode={languageCode} items={selectorItems} />
+        )}
```
steffenkleinle: 🤔 Can we just always display the button at the bottom? Even if there are still languages, I think that would provide a more consistent UI and I think it happens that you already realize before you finish your search that the language you are looking for is not there.

translations/translations.json
```diff
@@ -8124,15 +8126,17 @@
...
       "privacy": "Data protection",
       "localInformation": "Local information",
       "localInformationLabel": "Information",
       "events": "Events",
       "news": "News",
       "noLanguages": "Not available in other languages",
       "noTranslation": "No translation available",
+      "languageNotAvailableMessage": "This content is not yet available in all languages. We are actively expanding our language options based on user needs. If you'd like to request a specific language, please use the feedback feature.",
       "noLanguageFound": "No language found",
+      "languageNotFoundQuestion": "Your language not found?",
```
steffenkleinle: 🔧 Perhaps we can keep this even shorter?
```suggestion
      "languageNotFoundQuestion": "Language not found?",
```

web/src/components/LanguageList.tsx
```diff
@@ -64,48 +66,45 @@ const LanguageList = ({
...
+                key={language.code}
+                code={language.code}
+                path={language.path}
+                name={language.name}
+                close={close}
+                selectedLanguageCode={currentLanguage?.code}
+                onUnavailableLanguageClick={onUnavailableLanguageClick}
+              />
+            ))}
+          </List>
```
steffenkleinle: 🤔 Same as for native, what do you think of just always showing the button at the end of the list?

translations/translations.json
```diff
@@ -7822,9 +7822,9 @@
       "news": "Nachrichten",
       "noLanguages": "Nicht verfügbar in anderen Sprachen",
       "noTranslation": "Keine Übersetzung verfügbar",
-      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Sie können diese Sprache über die Feedback-Funktion anfragen.",
+      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Du kannst diese Sprache über die Feedback-Funktion anfragen.",
```
steffenkleinle: 🤔 Sorry for the repeat suggestions :( Last one, promised! 🤞 
``` suggestion
      "languageNotAvailableMessage": "Dieser Inhalt ist noch nicht in allen Sprachen verfügbar. Wir erweitern unser Sprachangebot aktiv. Du kannst deine Sprache über die Feedback-Funktion anfragen.",
```

web/src/components/LanguageList.tsx
```diff
@@ -60,6 +60,12 @@ type LanguageListProps = {
   onUnavailableLanguageClick?: () => void
 }
 
+const UnavailableLanguageButton = ({ onClick, label }: { onClick?: () => void; label: string }) => (
+  <Button style={{ margin: '2px 8px' }} variant='outlined' onClick={onClick}>
+    {label}
+  </Button>
+)
+
```
steffenkleinle: 🔧 Please avoid inline styles
🔧 Please extract the prop types to a separate type

translations/translations.json
```diff
@@ -8124,15 +8126,17 @@
...
-      "imprint": "Imprint",
+      "disclaimer": "Imprint",
       "privacy": "Data protection",
       "localInformation": "Local information",
       "localInformationLabel": "Information",
       "events": "Events",
       "news": "News",
       "noLanguages": "Not available in other languages",
       "noTranslation": "No translation available",
+      "languageNotAvailableMessage": "This content is not yet available in all languages. We are actively expanding our language options based on user needs. If you'd like to request a specific language, please use the feedback feature.",
```
steffenkleinle: 🔧 The English translations needs to be adjusted as well:
```suggestion
      "languageNotAvailableMessage": "This content is not yet available in all languages. We are actively expanding our language options. You can request your language via the feedback feature.",
```

steffenkleinle (approved):
Tested on firefox and android emulator, great work!

translations/translations.json
bahaaTuffaha: Did you create one or do you want me to do so?

web/src/components/LanguageList.tsx
```diff
@@ -123,17 +132,25 @@ const LanguageList = ({
...
-        <LanguageListItem
-          code={language.code}
-          path={language.path}
-          name={language.name}
-          close={close}
-          selectedLanguageCode={currentLanguage?.code}
-          key={language.code}
-        />
+      renderOption={(_, language, { index }) => (
+        <>
```
f1sh1918: 🔧 i think the key should be set in the outer html item so the fragment or we should get a missing key warning
Sth like should work
```
<React.Fragment key={language.code}>
```

web/src/components/base/AlertDialog.tsx
```diff
@@ -30,6 +31,7 @@ type DialogProps = {
 
 const AlertDialog = ({ title, close, children, className, actions }: DialogProps): ReactElement => {
   const { t } = useTranslation('layout')
+  const { desktop } = useDimensions()
```
f1sh1918, line 35: 🔧 i just found a general issue here which can lead to unexpected issues or misuse of the component.
For mobile we have to close button and only use `actions` but actions are optional. So if we would not pass `actions` the dialog can't be closed on mobile.
This is a component design issue and may be also fixed in a separate issue.

web/src/components/LanguageListItem.tsx
```diff
@@ -8,7 +8,7 @@ import { useTranslation } from 'react-i18next'
 
 import Link from './base/Link'
 
-const SelectorItemButton = styled(ListItemButton)`
+export const SelectorItemButton = styled(ListItemButton)`
```
f1sh1918: ❓ Where is this component imported? Can't find it

f1sh1918 (approved):
Looks and works good.
A few minor improvements/remarks

I know this is a button issue with caps. not sure if that "Nicht" shouldn't be "nicht" in this case
Image

Not sure if this is our general shadow but imo could be improved (grey instead of black)

Image

web/src/components/base/AlertDialog.tsx
bahaaTuffaha: Yes I hid the close button on mobile because it  takes alot of space and there is another button for closing.
You can also close it by clicking on the backdrop `onClose`.

web/src/components/LanguageListItem.tsx
bahaaTuffaha: Nice catch.
I used it before and forgot to revert this part.

web/src/components/base/AlertDialog.tsx
steffenkleinle: I guess it should be mandatory to pass at least one action to the dialog, e.g. by throwing an exception if there are none.

bahaaTuffaha: Do we need to throw an exception I think mandatory action prop is enough right?

kalairubachandrakumaran-oss:
tested ✅ 
72734E70-2DE5-4F36-862D-21B28E1920356FF4EFCF-8A54-4509-9621-4989A316961E

PR #4065, opened 2026-04-20
Troubleshooting nvm symlinks

f1sh1918:
### Short Description

Since also @steffenkleinle faced this issue and i was not able to fix it with an Automator script, i was investigating what could be the cause of the issue that XCode can not properly get your node version.
It appears to be a symlinking issue.

### Proposed Changes

- Add documentation how to fix that

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

---


native/docs/troubleshooting.md
```diff
@@ -163,3 +164,37 @@ If the app connection to the android device closes frequently during hot reload
 
 1. `yarn start` and `run android` to run and install the app
 2. Press `d` to open the Dev Menu in your running terminal and change the bundler location
+
+## PreAction scripts fail in XCode
+This issue happens most likely because you are using nvm and updated your node version at some point.
```
steffenkleinle: 🤔 There are many reasons why the pre action scripts can ail. Could you add the exact error message here and where to find it (log files in `native/ios`)? Thanks :)

f1sh1918: if i could find one in any log, i would definitly. XCode is a mess

steffenkleinle: Thats what I mean, we save logs to the integreat-app/native/ios folder (you can just open them in IntelliJ). And definitely +1 for xcode is a mess, especially in regards to logs/error messages.

steffenkleinle: For me there always were errors like node/yarn/nvm not found, but okay

steffenkleinle (approved):

native/docs/troubleshooting.md
f1sh1918: okay now i found them. thx

lunars97 (approved):
👍

PR #4066, opened 2026-04-22
3968: Fix space navbar ios chrome

### LINKED ISSUE #3968: Investigate the space under the bottom navigation on iOS
### Describe the Bug

While scrolling through the page, the bottom navigation bar stays in the same position and leaves a gap underneath.
Device: iPhone 16 Pro (390 × 844 px)
Browser: Chrome.

### Steps to Reproduce

1- Open the web app on an iPhone (Chrome)
2- Go to the main page (also reproducible on other pages).
3- Scroll down.

### Expected Behavior

The bottom navigation bar should stay attached to the bottom of the viewport without leaving any gap while scrolling.

### Related Issues

N/A

### Additional Information

https://github.com/user-attachments/assets/5ed86ed8-e675-453c-8211-a03c33559ee1

f1sh1918:
### Short Description

Chrome on ios has a bottom toolbar. It looks like `100vh` does not respect that and generates an empty space.

### Proposed Changes

- use `100lvh` to respect browsers toolbar
- mind that its mentioned not to work for firefox. I couldn't find any issue, but i simply added a fallback for safety reasons. See: https://caniuse.com/?search=lvh 

Explanation:

> lvh = Large Viewport Height — defined as the viewport height with all browser toolbars retracted. This is always the maximum possible size. So min-height: 100lvh means the layout background always fills at least the full large viewport, regardless of what the toolbar is doing. There's never a case where the 
>   visible area is larger than the layout background, so nothing can show through below the nav.

### Side Effects

- none afaics (tested iphone with safari and chrome and android with different browsers)
- 

### Testing

1. Use a iPhone real device and open the app in chrome. Choose a location and scroll up and down on a page. The bottom bar should be displayed without a gap
2. Please also test on android and different browsers
3. Also check in desktop mode that nothing is broken

### Resolved Issues

Fixes: #3968

---


lunars97 (approved):
I was not able to reproduce it on iPhone 11 iOS v18.3.1, checked on mobile Chrome and desktop. It definitely did not cause any issues 👍

steffenkleinle (approved):
Welcome back @f1sh1918 🎉 
Couldn't reproduce the original issue but also could not find any new ones, so lets try this :)

Issue #4037, opened 2026-04-15
Bottom Sheet disappears sometimes


lunars97:
### Describe the Bug

Bottom Sheet sometimes disappears from the view all of a sudden and it is not that consistent. 

### Steps to Reproduce

1. Go to Maps
2. See no bottom sheet, map view takes the whole height

### Expected Behavior

Bottom Sheet should be there

### Environment

- so far noticed on android device, might be also on iOS
- android version 16, 17 (might be also in other versions)
- App version: 2026.4.2

### Related Issues

might be because of the #3951 

### Additional Information
Also see discussion in mattermost https://chat.tuerantuer.org/digitalfabrik/pl/foxymqxwptgcbkqmchn53waope
<img width="373" height="834" alt="Image" src="https://github.com/user-attachments/assets/8208e8d2-0b90-431e-8cca-a1ccde2242ca" />

steffenkleinle:
Possibly fixed by #4082.

bahaaTuffaha:
I found a way to reproduce it : If you just waited around 3 seconds before clicking on the permission dialog (deny or allow) the sheet will vanish to the dark realm.

PR #4073, opened 2026-04-23
3985: Fix poi back navigation

### LINKED ISSUE #3985: Directly navigate back from pois after link click
### Describe the Problem

As a user I'd like to have a consistent history, i.e. navigating back should also always lead to the previous screen. Currently, this is not the case for linking to a poi detail (neither from the content or from deep links). Instead, the first back navigation only navigates back to the poi list.

On web this already works as expected.

Navigating back after linking to a poi detail should directly lead back to the previous page.
E.g. when clicking on `[LR, Lange Reihe 77, 27211 Bassum](https://webnext.integreat.app/testumgebung/de/locations/lr/)` at https://webnext.integreat.app/testumgebung/de/events/testserie1, navigating back should lead back to the event.
Make sure that for normal poi selection navigating back still only clears the selection.

N/A

### Related Issues

N/A

### Additional Information

https://github.com/digitalfabrik/integreat-app/pull/3948#issuecomment-4161168978

steffenkleinle:
### Short Description

This PR implements a custom stack history independent of react-navigation to allow for proper history on the map.
Using the react-navigation stack history was not possible, as this would lead to rerendering the whole map and bottom sheet due to new screens being pushed.

### Proposed Changes

- Use new custom stack history for map
- Override back navigation for map

### Side Effects

- I am not totally convinced that this is a good idea as its quite a bit of custom logic and workarounds that might break. However, I think from a users perspective this allows a better experience on the map with seamless back/forth navigation.

### Testing

Thoroughly test navigating back/forth/inside the maps tab on native. Especially check switching tabs, navigating via deep link to the map (e.g. from an event with linked poi) and navigate back and forth using both the header back button and system buttons/gestures. Check the history inside the poi tab by selecting/unselecting pois, multipois and filters and navigating back.

### Resolved Issues

Fixes: #3985

---


f1sh1918:
Not sure if this is only happening to me, but i don't see any bottomSheet on android @steffenkleinle 
image

bahaaTuffaha:
> Not sure if this is only happening to me, but i don't see any bottomSheet on android @steffenkleinle 

check this out: https://github.com/digitalfabrik/integreat-app/issues/4037

f1sh1918:
sorry again to bother you. Can you please result merge conflicts @steffenkleinle

steffenkleinle:
> sorry again to bother you. Can you please result merge conflicts @steffenkleinle

Of course, no worries. Done @f1sh1918

native/src/components/MapView.tsx
```diff
@@ -178,7 +179,9 @@ const MapView = ({
     const feature = featureCollection.features.find((it): it is MapFeature => it.geometry.type === 'Point')
     selectFeature(feature ?? null)
 
-    zoomOnClusterPress(pressedCoordinates)
+    if (selectedFeature && isMultipoi(selectedFeature)) {
```
f1sh1918:
```suggestion
    if (feature && isMultipoi(feature)) {
```
❓ wouldn't it be safer to check the current `feature` since the `selectFeature` just triggers a re-render and might use a stale state in the if clause.
Actually minor issue

native/src/hooks/useLocalStackHistory.ts
```diff
@@ -0,0 +1,67 @@
...
+  const pushReset = (newHistory: Partial<T>) =>
+    setHistory(previous => [...previous, { ...resetHistory, ...newHistory }])
+
+  const pop = useCallback(() => setHistory(previous => [...previous.slice(0, previous.length - 1)]), [])
+
+  const reset = useCallback((newHistory?: T[]) => setHistory(newHistory ?? [resetHistory]), [resetHistory])
+
+  useEffect(
+    () =>
+      BackHandler.addEventListener('hardwareBackPress', () => {
```
f1sh1918: ❌  i think this BackHandler makes issues.
Every BackHandler call adds a listener to a global list.
`useLocalStackHistory` is mounted inside `PoisContainer`, which stays mounted the entire time the POIs tab exists — even when you navigate to a different tab like Categories. Because it uses useEffect (not useFocusEffect), the BackHandler listener is always active.

Example how to reproduce:
When you now go to map -> select a poi
Switch the tab to categories and go to a sublevel and try the android back navigation it will happen nothing for the first try, because it does not go back for the category but for the poi detail.

f1sh1918 (changes requested):
Really nice approach looks way cleaner but i unfortunately found two issues.
Not sure if the second one is really related but the first has to be resolved!

The BackHandler causes side effects.
You see that the back navigation happens not in the current stack but in the POI.
See explanation in my review comment.
https://github.com/user-attachments/assets/2eb1f562-755b-44b1-9e6f-60fb01a3c071

Not sure if this is related. I might have faced this issue also in the maplibre update pr
Some camera blocking issue
https://github.com/user-attachments/assets/bb3605e2-07cd-4ffd-a560-4e27d68b17e9

native/src/components/MapView.tsx
steffenkleinle: You are absolutely right, thanks

native/src/hooks/useLocalStackHistory.ts
steffenkleinle: Thanks a lot, very nice catch and great explanation/reproduction description! Fixed :)

steffenkleinle:
> Not sure if this is related. I might have faced this issue also in the maplibre update pr Some camera blocking issue https://github.com/user-attachments/assets/bb3605e2-07cd-4ffd-a560-4e27d68b17e9

I can't open your attachment (404). But I guess if you have it in other PRs as well, it is not related? 😄

native/src/routes/PoisContainer.tsx
```diff
@@ -33,24 +72,39 @@ const PoisContainer = ({ navigation, route }: PoisContainerProps): ReactElement
     slug,
     multipoi,
     poiCategoryId,
-    zoom,
+    zoom: route.params.zoom,
   })
-  useHeader({ navigation, route, availableLanguages, data, shareUrl })
+
+  const goBack =
```
f1sh1918, line 70: 🔧 If you pass goBack function to a hook that includes states changes, i think the function should be wrapped in a `useCallback`

native/src/routes/Pois.tsx
```diff
@@ -31,80 +31,51 @@ const Container = styled.View`
...
 
-const Pois = ({ pois: allPois, regionModel, route, navigation }: PoisProps): ReactElement => {
-  const { slug, multipoi, poiCategoryId, zoom } = route.params
-  const [deselectOnBackNavigation, setDeselectOnBackNavigation] = useState(slug === undefined && multipoi === undefined)
-  const [poiCurrentlyOpenFilter, setPoiCurrentlyOpenFilter] = useState(false)
-  const [showFilterSelection, setShowFilterSelection] = useState(false)
+const Pois = ({ localHistory, initialZoom, pois: allPois, regionModel }: PoisProps): ReactElement => {
   const [userLocation, setUserLocation] = useState<LocationType | null>(null)
   const [bottomSheetSnapPointIndex, setBottomSheetSnapPointIndex] = useState(1)
-  const [listScrollPosition, setListScrollPosition] = useState(0)
```
f1sh1918, line 47: ❓ why did you drop that functionality?

f1sh1918 (commented):
Works good so far. 👍 

Just one question about the functionality drop.

native/src/routes/Pois.tsx
steffenkleinle: I think I dropped it when trying to use the default react-navigation for poi navigation. Will add this again.

native/src/routes/PoisContainer.tsx
steffenkleinle: Hmm, why? Are there any side-effects?

native/src/routes/__tests__/PoisContainer.spec.tsx
```diff
@@ -0,0 +1,181 @@
+import { fireEvent, waitFor } from '@testing-library/react-native'
+import React from 'react'
+
+import { POIS_ROUTE, PoisRouteType } from 'shared'
+import { RegionModelBuilder, LanguageModelBuilder, PoiModelBuilder } from 'shared/api'
+
+import useHeader from '../../hooks/useHeader'
+import useLoadCityContent from '../../hooks/useLoadRegionContent'
```
bahaaTuffaha: 🔧 
```suggestion
import useLoadRegionContent from '../../hooks/useLoadRegionContent'
```

native/src/routes/PoisContainer.tsx
```diff
@@ -1,27 +1,66 @@
...
+
+  useEffect(
+    () =>
+      navigation
+        .getParent<BottomTabNavigationProp<ParamListBase>>(TAB_NAVIGATOR_ID)
+        // Reset the history when we press the pois tab while already focused
+        // Keep the history while just switching between tabs
+        .addListener('tabPress', () => (navigation.isFocused() ? localHistory.reset() : null)),
+    [navigation, localHistory],
+  )
```
bahaaTuffaha, line 54: ❓  Does it need a cleanup function for the listener?

native/src/routes/Pois.tsx
bahaaTuffaha: (unrelated) can we put this StyleSheet outside of the component ?

native/src/hooks/useLocalStackHistory.ts
```diff
@@ -0,0 +1,70 @@
...
+}: UseLocalHistoryParams<S, T>): UseLocalHistoryReturn<T> => {
+  const [history, setHistory] = useState<T[]>(historyFromParams(params))
+  const current = history[history.length - 1] ?? resetHistory
+
+  const push = (newHistory: Partial<T>) => setHistory(previous => [...previous, { ...current, ...newHistory }])
+
+  const pushReset = (newHistory: Partial<T>) =>
+    setHistory(previous => [...previous, { ...resetHistory, ...newHistory }])
+
+  const pop = useCallback(() => setHistory(previous => [...previous.slice(0, previous.length - 1)]), [])
```
bahaaTuffaha:
```suggestion
  const pop = useCallback(() => setHistory(previous => previous.slice(0, -1)), [])
```

native/src/hooks/useLocalStackHistory.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const reset = useCallback((newHistory?: T[]) => setHistory(newHistory ?? [resetHistory]), [resetHistory])
+
+  useFocusEffect(
+    useCallback(
+      () =>
+        BackHandler.addEventListener('hardwareBackPress', () => {
+          const canGoBack = history.length > 1
+          if (canGoBack) {
```
bahaaTuffaha:
```suggestion
          if (history.length > 1) {
```

native/src/routes/PoisContainer.tsx
steffenkleinle: It is already provided by returning the result of `addListener` which is a function to remove the listener:
https://reactnavigation.org/docs/material-top-tab-navigator/#tabpress

native/src/routes/Pois.tsx
steffenkleinle: Seems like this already doesn't work on main anymore. I tried for a few hours now but could not get it to work again. @f1sh1918 do you want to give it a try or can you create an issue?

steffenkleinle: https://github.com/digitalfabrik/integreat-app/issues/4113

native/src/routes/__tests__/PoisContainer.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const pois = new PoiModelBuilder(3).build()
+const city = new RegionModelBuilder(1).build()[0]!
+const languages = new LanguageModelBuilder(2).build()
+
+const data = {
+  pois,
+  city,
+  languages,
+  cities: [city],
```
bahaaTuffaha: ❓ should it be regions instead of cities?

bahaaTuffaha (approved):
Nice work!
Tested on Android and iOS ... poi history works as expected ✅

native/src/routes/PoisContainer.tsx
f1sh1918: i mean afaics this is a react convention to do that. But if you think its not needed you can leave it

f1sh1918 (approved):
Good Work 👍

kalairubachandrakumaran-oss:
Open a Poi from Events section and go back works fine, I return to the event back ✅ 
But when I open a poi from a tile (e.g Social Lie) and close the web/map overlay using the X , I'm taken back to the Information overview instead of the previous page (Social Life)

@steffenkleinle Is that expected?
https://github.com/user-attachments/assets/96354f57-3165-4eef-a49d-543d576ab786

steffenkleinle:
Definitely not expected, could you please create an issue?

Issue #3920, opened 2026-03-10
App sometimes freezes


kalairubachandrakumaran-oss:
### Describe the Bug
The app sometimes freezes on both iOS and Android. It might be related to quickly switching between tabs or tapping multiple tiles in a short period of time, when navigating back and forth quickly between different sections, for example opening the map, going back to the information page, opening a tile, then switching back to the map again
-> the exact cause is unclear

### Steps to Reproduce

1.    Open the app "Integreat" on mobile device
2.    Quickly switch between tabs, navigate between different sections (e.g., Information -> open the map -> go back to information ->open a tile -> go back to the map)
3.    The app may freeze

### Expected Behavior
The app should respond normally while navigating between sections.

### Additional Information
Device Info:
-  iOS version 26.3.1 (iPhone 12 mini)
- Android version 12 (Google Pixel 3 XL)
<img width="400" height="860" alt="Image" src="https://github.com/user-attachments/assets/eb1d90fe-d3a4-426f-9fd4-c0296dbb47f7" />

bahaaTuffaha:
My findings on this issue:
can be reproduced on android 12 and problem could be from the maplibre package when switching between tabs rapidly : https://github.com/maplibre/maplibre-react-native/issues/873.
I tried all forms of optimization from conditional rendering to `useMemo` or even `freezeOnBlur`, `lazy` screenOptions.
I couldn't reproduce it on api 36.1 (not sure if 16kb is the difference here)

https://github.com/user-attachments/assets/cbf086f8-bf5b-42bd-b007-a91365558816

steffenkleinle:
~~@bahaaTuffaha could you recheck if this got fixed in #3951 please?~~
I just saw that this will only be fixed with the next major version v11, so we'll have to wait some more.

PR #4074, opened 2026-04-23
4056: Upgrade maplibre to v11

### LINKED ISSUE #4056: Upgrade maplibre-react-native to v11
### Describe the Problem

Upgrade `@maplibre/maplibre-react-native` to [v11](https://github.com/maplibre/maplibre-react-native/releases/tag/v11.0.0). This might fix some of our problems:
https://github.com/digitalfabrik/integreat-app/issues/3920
https://github.com/digitalfabrik/integreat-app/issues/4018 (Patch added in #4025, Android 12)

Upgrade using the [migration guide](https://github.com/maplibre/maplibre-react-native/blob/main/docs/content/setup/migrations/v11.md) to [v11](https://github.com/maplibre/maplibre-react-native/releases/tag/v11.0.0) and check if #3920 is still reproducible and the patch from #4025 is still necessary.

N/A

### Related Issues

#3920
#4018

### Additional Information

https://github.com/maplibre/maplibre-react-native/releases/tag/v11.0.0

bahaaTuffaha:
### Short Description

Upgrade @maplibre/maplibre-react-native to [v11](https://github.com/maplibre/maplibre-react-native/releases/tag/v11.0.0).

### Proposed Changes

Followed the migration setup at https://github.com/maplibre/maplibre-react-native/blob/main/docs/content/setup/migrations/v11.md

- `MapView` => `Map`
- `MapViewRef` => `MapRef`
- `ShapeSource` => `GeoJSONSource`
- `<CircleLayer/>` , `<SymbolLayer/>` => `<Layer type=""/>`
~~- `UserTrackingMode` => `trackUserLocation`  (now it uses value default instead of follow).~~
- Replaced `UserLocation` onUpdate callback with `useCurrentPosition()` and the useEffect will update the userLocation.
- `<NativeUserLocation/>` used as replacement to `<UserLocation/>` because UserLocation’s puck not visible.
- Added `toLngLat()` helper: Converts Position arrays to LngLat tuples.
- `logoEnabled` => `logo`
- `attributionEnabled` => `attribution`
- `Shape` => `data`
- Padding: `paddingBottom` => `bottom`
~~- Used a `camera ref` a separate camera ref.~~
- Removed `zoomEnabled` (deprecated)
- Replaced `cameraSettings` state’s types with `CameraStop` type.
- Prevent centering on the currently selected feature when minimizing the bottom sheet
- Use initial camera settings
- Move user location hook up to Pois
- Keep previous user location on reload
- Rename coordinates to userLocation

⚠️  There is error in the console:
This issue could be from the server misconfiguration 🤔 
```js
MapLibre Native [ERROR] [Mbgl] {RenderThread 12}[Style]: Failed to load tile 5/17/11=>5 for source natural_earth_shaded_relief: bitmap decoding: couldn't get bitmap info
```

### Side Effects

- Different looking puck 👀 .

### Testing

Go to Map/poi and test the following:
- Press on Pois on the map.
- Test location button (precise and approximate).
- Test iOS and android.
- Test interaction between following user location, (un-)selecting features (list and map) and expanding/collapsing the bottom sheet.

### Resolved Issues

Fixes: #4056 , #3920

---


f1sh1918:
For me the gps icon does not change, even i allow gps. Not sure if its related (fresh install required) @bahaaTuffaha 
android tested
I think there also changed something from v10 to v11

native/src/components/MapView.tsx
```diff
@@ -99,96 +105,114 @@ const MapView = ({
...
-    ])
-
-    const feature = featureCollection.features.find((it): it is MapFeature => it.geometry.type === 'Point')
+    const pressedCoordinates = event.nativeEvent.point
+    const featureCollection: Feature<Geometry, GeoJsonProperties>[] = await mapRef.current.queryRenderedFeatures(
+      pressedCoordinates,
+      { layers: [featureLayerId, 'selected-marker'] },
+    )
+
+    const feature = featureCollection.find((it): it is MapFeature => it.geometry.type === 'Point')
```
f1sh1918, line 181: 🔧 zoomOnClusterPress should be awaited since its async

native/src/components/MapView.tsx
```diff
@@ -99,96 +105,114 @@ const MapView = ({
...
 
   const onRequestLocation = useCallback(async () => {
-    const newUserLocation = userLocation ?? (await refreshPermissionAndLocation())?.coordinates
+    const currentUserLocation: LocationType | null =
+      currentPosition?.coords == null ? null : [currentPosition.coords.longitude, currentPosition.coords.latitude]
+    const newUserLocation = currentUserLocation ?? userLocation ?? (await refreshPermissionAndLocation())?.coordinates
     if (newUserLocation) {
       setUserLocation(newUserLocation)
       moveTo(newUserLocation)
+      cameraRef.current?.easeTo({
```
f1sh1918, line 133: 🔧 since we use `moveTo` which has an `ease` and we do it here again. Is it not used twice now? Could lead to visual issues. Why is this neccesary?

bahaaTuffaha:
> For me the gps icon does not change, even i allow gps. Not sure if its related (fresh install required) @bahaaTuffaha
> android tested

it worked without fresh install 🤔 did the camera move without a puck showing up or non of them?

f1sh1918:
> > For me the gps icon does not change, even i allow gps. Not sure if its related (fresh install required) @bahaaTuffaha
> > android tested
> 
> it worked without fresh install 🤔 did the camera move without a puck showing up or non of them?

no i meant, if you do a fresh install of the app. the location permissions will not be reflected on the map after you permitted. I guess you have to kill the app and start it again to reflect the permissions
But it should work directly

bahaaTuffaha:
> > > For me the gps icon does not change, even i allow gps. Not sure if its related (fresh install required) @bahaaTuffaha
> > > android tested
> > 
> > 
> > it worked without fresh install 🤔 did the camera move without a puck showing up or non of them?
> 
> no i meant, if you do a fresh install of the app. the location permissions will not be reflected on the map after you permitted. I guess you have to kill the app and start it again to reflect the permissions But it should work directly

I had that issue before 🤔 what and android version do you use?
does it work as expected on a real device?

f1sh1918:
> > > > For me the gps icon does not change, even i allow gps. Not sure if its related (fresh install required) @bahaaTuffaha
> > > > android tested
> > > 
> > > 
> > > it worked without fresh install 🤔 did the camera move without a puck showing up or non of them?
> > 
> > 
> > no i meant, if you do a fresh install of the app. the location permissions will not be reflected on the map after you permitted. I guess you have to kill the app and start it again to reflect the permissions But it should work directly
> 
> I had that issue before 🤔 what and android version do you use? does it work as expected on a real device?

android 16 so recent

f1sh1918:
removed and reinstalled all node modules but cannot get it run @bahaaTuffaha 

image

bahaaTuffaha:
> removed and reinstalled all node modules but cannot get it run @bahaaTuffaha

I deleted node_modules and reinstalled. It worked without crashing

f1sh1918:
Well for a fresh install it still not works. 
And i also think that the disappear of the bottom sheet was not fixed by my pr (out of scope) 
As a user the gps location is still disabled on android 16 even i allowed permissions
image

bahaaTuffaha:
> And i also think that the disappear of the bottom sheet was not fixed by my pr (out of scope)

Did you pull to get the last merge/(sync) commit?

f1sh1918:
okay i retested on real device android 16 and it seem to work. (without your last commit) so you can remove the workaround. But on android emulator i still get no position update even i granted permission. No idea why.

I think rebasing is the better option instead of merging (main) at least if the pr is not fully approved and there might be additional commits. No its difficult to get rid of the unneeded workaround commit

f1sh1918 (approved):
Tested on android and ios. 
On real device works great 👍 
On android emulator the gps location does not work well. But not sure if its related

Please remove this workaround commit here: c3c7dece3a24cc3961e1f37773840dfaf57d5d89

Not need afaics @bahaaTuffaha

steffenkleinle:
> On real device works great 👍 On android emulator the gps location does not work well. But not sure if its related

This has been the case for a bit sadly. But since its only on emulators, I guess no use putting time into it.

native/package.json
```diff
@@ -47,7 +47,7 @@
     "@formatjs/intl-displaynames": "^6.8.13",
     "@formatjs/intl-locale": "^4.2.13",
     "@gorhom/bottom-sheet": "^5.2.8",
-    "@maplibre/maplibre-react-native": "patch:@maplibre/maplibre-react-native@npm%3A10.4.2#~/.yarn/patches/@maplibre-maplibre-react-native-npm-10.4.2-a5da575b03.patch",
+    "@maplibre/maplibre-react-native": "11.0.1",
```
steffenkleinle: 🤔 Any reason why you are requiring the exact version?
```suggestion
    "@maplibre/maplibre-react-native": "^11.0.2",
```

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
+  const toLngLat = (position: Position): LngLat => {
+    const [longitude, latitude] = position
+
+    if (longitude === undefined || latitude === undefined) {
+      reportError('Expected two-dimensional map coordinates.')
+      return [0, 0]
+    }
+
+    return [longitude, latitude]
+  }
```
steffenkleinle: ~~🔧 Lets move this to utils (or at least out of the `MapView` component.
🤔 I think we should just throw an error instead of just reporting one (which is more for debugging in sentry).~~

Sorry, only now got what this is actually doing. In my opinion, in this case it would be a lot easier and clearer to understand to just type cast it with `as LngLat` in L131 and L140:
``` ts
    ...(coordinates !== undefined ? { center: coordinates as LngLat } : { bounds }),
```

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
+
+  const mapRef = useRef<MapRef>(null)
   const [followUserLocation, setFollowUserLocation] = useState<boolean>(false)
   const { refreshPermissionAndLocation } = useUserLocation({ requestPermissionInitially: true })
+  const currentPosition = useCurrentPosition()
```
steffenkleinle: 🤔 Hmm, it feels a little redundant to have both two hooks to request the location. Is this necessary? And on top, we get `userLocation` already passed as prop. I think we should clean this up a bit.

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
-    animationDuration,
+  const [cameraSettings, setCameraSettings] = useState<CameraStop>({
+    ...(lngLatCoordinates !== undefined ? { center: lngLatCoordinates } : { bounds }),
+    zoom: zoom ?? defaultZoom,
+    padding: { bottom: bottomSheetHeight },
   })
 
   const moveTo = useCallback(
-    (location: Position, zoomLevel = normalDetailZoom) =>
+    (location: LngLat, zoomLevel = normalDetailZoom) => {
```
steffenkleinle: 🙃 or center
```suggestion
    (position: LngLat, zoomLevel = normalDetailZoom) => {
```

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
+      })
+    },
+    [bottomSheetHeight],
   )
 
   const onRequestLocation = useCallback(async () => {
-    const newUserLocation = userLocation ?? (await refreshPermissionAndLocation())?.coordinates
+    const currentUserLocation: LocationType | null =
+      currentPosition?.coords == null ? null : [currentPosition.coords.longitude, currentPosition.coords.latitude]
+    const newUserLocation = currentUserLocation ?? userLocation ?? (await refreshPermissionAndLocation())?.coordinates
```
steffenkleinle: ❌ Imo this is too many different user locations now 🙈

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
     if (newUserLocation) {
+      selectFeature(null)
```
steffenkleinle: 🤔 Why do we deselect the feature if there is a new user location?

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
+    }
+
+    const currentUserLocation: LocationType = [currentPosition.coords.longitude, currentPosition.coords.latitude]
+
+    // Avoid frequent rerenders if distance only changes minimally
+    if (!userLocation || calculateDistance(userLocation, currentUserLocation) > MIN_DISTANCE_THRESHOLD) {
+      setUserLocation(currentUserLocation)
+      moveTo(currentUserLocation)
+    }
+  }, [currentPosition, followUserLocation, moveTo, setUserLocation, userLocation])
```
steffenkleinle: ❓ Shouldn't we only do this if `followUserLocation` is true? Why is it part of the dependency array even though its not used? 
Can we use the library functionality for this instead?
https://maplibre.org/maplibre-react-native/docs/components/camera#trackuserlocation

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
+
   const zoomOnClusterPress = async (pressedCoordinates: [number, number]) => {
-    const clusterCollection = await mapRef.current?.queryRenderedFeaturesAtPoint(pressedCoordinates, undefined, [
-      clusterLayerId,
-    ])
-    if (clusterCollection && 0 in clusterCollection.features && mapRef.current) {
-      const feature = clusterCollection.features[0] as MapFeature
-      moveTo(feature.geometry.coordinates, (await mapRef.current.getZoom()) + clusterClickZoomFactor)
+    const clusterCollection: Feature<Geometry, GeoJsonProperties>[] | undefined =
```
steffenkleinle: 🔧 The type annotation is not needed and removing it makes it more readable imo.
```suggestion
    const clusterCollection =
```

native/src/components/MapView.tsx
```diff
@@ -99,94 +104,110 @@ const MapView = ({
...
-    ]
-    const featureCollection = await mapRef.current.queryRenderedFeaturesAtPoint(pressedCoordinates, undefined, [
-      featureLayerId,
-      'selected-marker',
-    ])
-
-    const feature = featureCollection.features.find((it): it is MapFeature => it.geometry.type === 'Point')
-    selectFeature(feature ?? null)
+    const pressedCoordinates = event.nativeEvent.point
+    const featureCollection: Feature<Geometry, GeoJsonProperties>[] = await mapRef.current.queryRenderedFeatures(
```
steffenkleinle: 🔧 Same as before
```suggestion
    const featureCollection = await mapRef.current.queryRenderedFeatures(
```

steffenkleinle (changes requested):
Nice work!

native/package.json
bahaaTuffaha: I was having issues with 11.0.0 so I switched to 11.0.1 and probably I missed the `^`

native/src/components/MapView.tsx
bahaaTuffaha: Then I will use our hook instead

native/src/components/MapView.tsx
bahaaTuffaha: I will remove it and just use our hook `useUserLocation`

native/src/components/MapView.tsx
bahaaTuffaha: Without it will cause problems with previous selected poi:

https://github.com/user-attachments/assets/ef6511aa-a727-42df-a0d8-31e79529be78

steffenkleinle: Yes, because you added `bottomSheetHeight` as dependency to `moveTo` callback which in turn triggers the `useEffect` to center to the selected feature on every adjustment of the bottom sheet height. This will also have the side effect that if I select a feature, move the map to somewhere else and then adjust the bottom sheet, I also get sent back to the selected feature, which I would be annoyed about. So we should fix that the moveTo feature is not triggered again when adjusting the bottom sheet height.

Possible fix:
- Create a ref with the current `bottomSheetHeight`:
``` tsx
  const bottomSheetHeightRef = useRef(bottomSheetHeight)                                                                                                                                                                                                                                                                                                                          
  bottomSheetHeightRef.current = bottomSheetHeight     
```
- Use that instead of the prop in `moveTo`:
```tsx
  const moveTo = useCallback(                                                                                                                                                                                                                                                                                                                                                     
    (position: LngLat, zoomLevel = normalDetailZoom) => {
      cameraRef.current?.easeTo({                                                                                                                                                                                                                                                                                                                                                 
        center: position,
        zoom: zoomLevel,                                                                                                                                                                                                                                                                                                                                                          
        duration: animationDuration,
        padding: { bottom: bottomSheetHeightRef.current },
      })                                                                                                                                                                                                                                                                                                                                                                          
    },
    [],                                                                                                                                                                                                                                                                                                     
  ) 
```

native/src/components/MapView.tsx
```diff
@@ -99,97 +101,84 @@ const MapView = ({
...
+      selectFeature(null)
       setUserLocation(newUserLocation)
       moveTo(newUserLocation)
-      setFollowUserLocation(true)
+      setShowUserLocation(true)
     }
-  }, [refreshPermissionAndLocation, moveTo, setUserLocation, userLocation])
+  }, [showUserLocation, refreshPermissionAndLocation, moveTo, setUserLocation, selectFeature])
 
+  // Recenter on the selected marker.
```
steffenkleinle: 🙃 Unnecessary comment imo
```suggestion
```

native/src/components/MapView.tsx
```diff
@@ -99,97 +101,84 @@ const MapView = ({
   bottomSheetFullscreen,
   zoom,
 }: MapViewProps): ReactElement => {
-  const mapRef = useRef<MapViewRef>(null)
-  const [followUserLocation, setFollowUserLocation] = useState<boolean>(false)
+  const mapRef = useRef<MapRef>(null)
+  const cameraRef = useRef<CameraRef>(null)
+  const [showUserLocation, setShowUserLocation] = useState<boolean>(false)
```
steffenkleinle: ❓ Why did you rename this? Imo the previous naming `followUserLocation` was a lot clearer on what it does. Maybe `trackUserLocation` would be even better and reflect the internal maplibre naming.
❓ And shouldn't we actually also set `trackUserLocation` then?
https://maplibre.org/maplibre-react-native/docs/components/camera#trackuserlocation

native/src/components/MapView.tsx
```diff
@@ -99,97 +101,84 @@ const MapView = ({
...
+    [bottomSheetHeight],
   )
 
   const onRequestLocation = useCallback(async () => {
-    const newUserLocation = userLocation ?? (await refreshPermissionAndLocation())?.coordinates
+    // Toggle off if already showing user location.
+    if (showUserLocation) {
+      setShowUserLocation(false)
+      return
+    }
```
steffenkleinle: 🤔 Is this really the standard behavior in other maps?
🙃 Redundant comment

native/src/components/MapView.tsx
```diff
@@ -99,97 +101,84 @@ const MapView = ({
...
-    animationDuration: number
-  }>({
-    zoomLevel: zoom ?? defaultZoom,
-    centerCoordinate: coordinates,
-    bounds: coordinates ? undefined : bounds,
-    animationDuration,
+  const [cameraSettings] = useState<CameraStop>({
+    ...(coordinates !== undefined ? { center: coordinates as LngLat } : { bounds }),
+    zoom: zoom ?? defaultZoom,
+    padding: { bottom: bottomSheetHeight },
```
steffenkleinle: 🔧 If we never update these, do we actually need the `useState`?

native/src/components/MapView.tsx
```diff
@@ -198,33 +187,26 @@ const MapView = ({
...
-          zoomEnabled
           onPress={onPress}
           ref={mapRef}
-          attributionEnabled={false}
-          logoEnabled={false}>
-          <UserLocation visible={!!userLocation} onUpdate={updateUserLocation} />
-          <ShapeSource
+          attribution={false}
+          logo={false}>
+          {showUserLocation && <NativeUserLocation />}
```
steffenkleinle: ❓ Why do we use the `NativeUserLocation` instead of the standard `UserLocation` component?
❓ Why do we not always show the user location?

bahaaTuffaha: I wrote that in the description:
`<NativeUserLocation/>` used as replacement to `<UserLocation/>` because UserLocation’s puck not visible.

bahaaTuffaha: > ❓ Why do we not always show the user location?

why there is a button then?

native/src/components/MapView.tsx
bahaaTuffaha: `useState` here freezes the initial value but const will recompute on every render

native/src/components/MapView.tsx
steffenkleinle: > > ❓ Why do we not always show the user location?
> 
> why there is a button then?

Its to track/follow the user location, not to hide/show it

native/src/components/MapView.tsx
bahaaTuffaha: `<Camera trackUserLocation={followUserLocation ? 'default' : undefined} ref={cameraRef} {...cameraSettings} />`
I added `trackUserLocation` to the camera and removed `moveTo()` from onRequestLocation but now the userlocation cant zoom in it just pans to the puck without zooming.

bahaaTuffaha: I kept `moveTo()` and `trackUserLocation`
- `moveTo()` for animation + zoom in (one time).
- `trackUserLocation` : it continuously follows the user.

native/src/components/MapView.tsx
steffenkleinle: Why is it necessary that it keeps the initial value? Seems like its not used afterwards anyway anymore? Otherwise we'd not be able to zoom in/out or move the map, right?

steffenkleinle (commented):
@bahaaTuffaha I found a way to avoid deselecting a feature on track user location, have a look at https://github.com/digitalfabrik/integreat-app/pull/4120 and let me know what you think of it. I also included a few other improvements.

bahaaTuffaha:
> @bahaaTuffaha I found a way to avoid deselecting a feature on track user location, have a look at #4120 and let me know what you think of it. I also included a few other improvements.

Thank you 😄 I will check it out!

steffenkleinle (approved):

steffenkleinle:
Adding a `Native` label since there have been some user facing changes, mainly in regards to the user location and bottom sheet handling. For more information see https://github.com/digitalfabrik/integreat-app/pull/4120.

f1sh1918 (approved):
Tested on android

kalairubachandrakumaran-oss:
tested ✅ 
related issue: #4190

PR #4076, opened 2026-04-26
4055: Move easter egg to version in settings

### LINKED ISSUE #4055: Move the easter egg yellow marker's logic to settings
### Describe the Problem

Move the easter egg yellow map-marker logic to the version button in settings at native.
The version button is being used for sentry error report but we don't need that anymore.
- Discuss if we gonna need to adjust the placement of the text for `Currently using API:`

- Remove the yellow map marker at landing page.
- Use the same clickCount logic at the version button.
- Update documentation.

Keep it as it is.

lunars97:
### Short Description

Move easter egg to version in settings. I left switch to default api button in landing page, imho it makes it visually more noticeable there. 
### Proposed Changes

- Add easter egg logic to app version
- Rename SwitchCmsUrlIcon -> SwitchCmsUrlButton
- Leave switch button in landing page
- Adjust test
- Adjust documentation

### Side Effects

- none

### Testing

Go to settings -> scroll down to version -> tap it 10 times - see the snackbar -> see the switch button at the top

### Resolved Issues

Fixes: #4055 

---


native/src/utils/createSettingsSections.ts
```diff
@@ -27,150 +28,177 @@ export type SettingsSectionType = {
...
+  setTapStart,
+}: CreateSettingsSectionsProps): (SettingsSectionType | null)[] => {
+  const { cmsUrl, switchCmsUrl } = buildConfig()
+  const { apiUrlOverride } = settings
+
+  const setApiUrl = (newApiUrl: string) => {
+    updateSettings({ apiUrlOverride: newApiUrl })
+    setTapCount(0)
+    setTapStart(null)
+  }
```
bahaaTuffaha: ❓ is it missing a `clearResourcesAndCache()` ?

native/src/components/SwitchCmsUrlButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+import buildConfig from '../constants/buildConfig'
+import { useAppContext } from '../hooks/useCityAppContext'
+import Text from './base/Text'
+
+const Container = styled.View`
+  display: flex;
+  align-items: center;
+`
+
+type LandingIconProps = {
```
bahaaTuffaha: 🔧 
```suggestion
type SwitchCmsUrlButtonProps = {
```

native/src/components/SwitchCmsUrlButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const setApiUrl = (newApiUrl: string) => {
+    updateSettings({ apiUrlOverride: newApiUrl })
+    clearResourcesAndCache()
+  }
+
+  return (
+    <Container>
+      <Text style={{ paddingTop: 12, color: theme.colors.error }}>{`Currently using API: ${apiUrlOverride}`}</Text>
+      <Button mode='contained' style={{ marginTop: 16 }} onPress={() => setApiUrl(cmsUrl)}>
```
bahaaTuffaha: 🔧  Move it to StyleSheet. Keep the dynamic variables as inline.

bahaaTuffaha (commented):
- Clicking rapidly again after switching to cms will show the snackbar with text `Switched to CMS url` again.

lunars97:
> * Clicking rapidly again after switching to cms will show the snackbar with text `Switched to CMS url` again.

Hmm the logic here that it can be switched to deafult through the button click in landing, so I am not sure whether this behaviour is incorrect... :thinking:

native/src/utils/createSettingsSections.ts
lunars97: I dont think we need to clear it in settings, the api switch to default happens in landing :thinking:

bahaaTuffaha: could it leave any old objects if not cleared? also if switched to default from settings by clicking again it wont clear Resources

native/src/components/SwitchCmsUrlButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  const setApiUrl = (newApiUrl: string) => {
+    updateSettings({ apiUrlOverride: newApiUrl })
+    clearResourcesAndCache()
+  }
+
+  return (
+    <Container>
+      <Text style={{ paddingTop: 12, color: theme.colors.error }}>{`Currently using API: ${apiUrlOverride}`}</Text>
```
steffenkleinle: 🔧 No need for a string template:
```suggestion
      <Text style={{ paddingTop: 12, color: theme.colors.error }}>Currently using API: {apiUrlOverride}</Text>
```

native/src/components/SwitchCmsUrlButton.tsx
steffenkleinle: Even better would be to just use padding/margin on the parent component (i.e. `Container`) and use a `gap` for padding between components.

native/src/utils/createSettingsSections.ts
```diff
@@ -27,150 +28,177 @@ export type SettingsSectionType = {
   getSettingValue?: (settings: SettingsType) => boolean | null
 }
 
-const volatileValues = {
-  versionTaps: 0,
-}
-
-const TRIGGER_VERSION_TAPS = 25
```
steffenkleinle, line 33: ❓ Is there a reason why we can't use exactly the same logic we used before for sentry? I think that was a lot simpler/required less code.

native/src/utils/createSettingsSections.ts
steffenkleinle: Agreeing with @bahaaTuffaha, I think we need to clear the resources on every CMS switch, not just if switching back to the default. We had the same implementation before.

steffenkleinle (changes requested):

native/README.md
```diff
@@ -61,7 +61,7 @@ run the app.
 
 In development builds, i.e. `yarn start`, the test cms is used as default.
 
-To switch between production and test cms, press the yellow location marker on the regions page 10 times.
+To switch between production and test cms, go to **Settings**, scroll down to the **Version** and tap it 10 times. To switch back to default cms url, go to landing page (Change location) and press the button **Switch back to default API**.
```
steffenkleinle: 🔧 
```suggestion
To switch between production and test cms, go to **Settings**, scroll down to the **Version** and tap it 10 times. To switch back to default cms url, go to the region selection and press the button **Switch back to default API**.
```

native/src/utils/createSettingsSections.ts
```diff
@@ -163,7 +165,21 @@ const createSettingsSections = ({
...
+
+        const { apiUrlOverride } = settings
+        if (apiUrlOverride === switchCmsUrl) {
+          return
+        }
+
+        updateSettings({ apiUrlOverride: switchCmsUrl })
+        clearResourcesAndCache()
+        log(`Switching to new API-Url: ${switchCmsUrl}`)
+        showSnackbar({ text: 'Switched to CMS url' })
```
steffenkleinle: 🔧
```suggestion
        showSnackbar({ text: `Switched to CMS ${switchCmsUrl}. Go back and pull to refresh to update.` })
```

native/src/utils/createSettingsSections.ts
```diff
@@ -163,7 +165,21 @@ const createSettingsSections = ({
...
+        }
+
+        const { apiUrlOverride } = settings
+        if (apiUrlOverride === switchCmsUrl) {
+          return
+        }
+
+        updateSettings({ apiUrlOverride: switchCmsUrl })
+        clearResourcesAndCache()
+        log(`Switching to new API-Url: ${switchCmsUrl}`)
```
steffenkleinle: 🙃 
```suggestion
        log(`Switching to CMS ${switchCmsUrl}`)
```

native/src/routes/Regions.tsx
```diff
@@ -32,12 +32,18 @@ type RegionsProps = {
...
   const { t } = useTranslation('regions')
+  const { apiUrlOverride } = settings
 
   // The regions are otherwise only updated by pull to refresh
   useEffect(refresh, [refresh])
 
+  // Refresh whenever the API url override changes
+  useEffect(() => {
+    refresh()
+  }, [apiUrlOverride, refresh])
```
steffenkleinle: ❓ Why do we need a separate refresh here? Isn't this already handled with the refresh in L55?

native/src/routes/Settings.tsx
```diff
@@ -20,10 +22,17 @@ type SettingsProps = {
...
   const appContext = useAppContext()
+  const { refresh } = useLoadRegions()
   const showSnackbar = useSnackbar()
   const { t } = useTranslation('settings')
   const { settings } = appContext
 
+  const clearResourcesAndCache = useCallback(() => {
+    dataContainer.clearInMemoryCache()
+    dataContainer._clearOfflineCache().catch(reportError)
+    refresh()
```
steffenkleinle: 🤔 This only updates the regions, not the region content. So the user still needs to manually do a pull to refresh. Therefore, I think I'd just remove this and instead include that in the snackbar message. Then we also don't need to pass this function as prop.
```suggestion
```

native/src/routes/Regions.tsx
```diff
@@ -54,7 +60,7 @@ const Regions = ({ navigation }: RegionsProps): ReactElement => {
       {regions && (
         <>
           <Wrapper {...testID('Regions-Page')}>
-            <SwitchCmsUrlIcon clearResourcesAndCache={clearResourcesAndCache} />
+            <SwitchCmsUrlButton clearResourcesAndCache={clearResourcesAndCache} />
```
steffenkleinle: 🤔 From a user flow I'd find it clearer if we'd display this button in the settings above the other settings section instead of the region selection if we use a switch cms url. What do you think of moving it there?

native/src/components/SwitchCmsUrlButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { cmsUrl } = buildConfig()
+  const { apiUrlOverride } = settings
+
+  if (!apiUrlOverride || apiUrlOverride === cmsUrl) {
+    return null
+  }
+
+  const setApiUrl = (newApiUrl: string) => {
+    updateSettings({ apiUrlOverride: newApiUrl })
+    clearResourcesAndCache()
```
steffenkleinle, line 35: 🔧 Can we display the same snackbar message and log the cms url switch as on version tap?

native/src/components/SwitchCmsUrlButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return (
+    <Container>
+      <Text style={{ color: theme.colors.error }}>Currently using API: {apiUrlOverride}</Text>
```
steffenkleinle: 🔧 
```suggestion
      <Text style={{ color: theme.colors.error }}>Currently using CMS: {apiUrlOverride}</Text>
```

steffenkleinle (commented):
Thanks, it now looks really clean! I have just a few remarks regarding UX/user flow and consistency.

native/src/routes/Regions.tsx
lunars97: yes, this sounds better 👍

native/src/routes/Regions.tsx
lunars97: unfortunately, it does not react on url override...

lunars97: anyway I will remove it and move the component to settings

native/src/routes/Settings.tsx
```diff
@@ -24,6 +26,11 @@ const Settings = ({ navigation }: SettingsProps): ReactElement => {
   const { t } = useTranslation('settings')
   const { settings } = appContext
 
+  const clearResourcesAndCache = useCallback(() => {
+    dataContainer.clearInMemoryCache()
+    dataContainer._clearOfflineCache().catch(reportError)
+  }, [])
```
steffenkleinle: 🔧  There is no need having this as a `useCallback` and passing it since there are no dependencies to the props/component. You can just directly call these two functions where they are used.

native/src/utils/createSettingsSections.ts
```diff
@@ -163,7 +165,21 @@ const createSettingsSections = ({
...
+
+        const { apiUrlOverride } = settings
+        if (apiUrlOverride === switchCmsUrl) {
+          return
+        }
+
+        updateSettings({ apiUrlOverride: switchCmsUrl })
+        clearResourcesAndCache()
+        log(`Switching to CMS: ${switchCmsUrl}`)
+        showSnackbar({ text: `Switched to CMS ${switchCmsUrl}. Go back and pull to refresh to update.` })
```
steffenkleinle: ~~🙃 This logic is duplicated, so we could extract it to a hook `useSwitchApiUrl` that returns a function doing this and accepting only a param `newApiUrl`.~~
Edit: Nvm, this is not a hook/component so hooks are not usable here.

steffenkleinle (approved):
Nice work! Tested on iOS and Android.

native/src/routes/Settings.tsx
```diff
@@ -42,22 +49,29 @@ const Settings = ({ navigation }: SettingsProps): ReactElement => {
...
+        <Divider />
+        <FlatList
+          data={sections}
+          extraData={appContext.settings}
+          renderItem={renderItem}
+          ItemSeparatorComponent={Divider}
+          ListFooterComponent={Divider}
+        />
+      </Layout>
+    </>
```
steffenkleinle: ❌  While this is mostly just visible to devs, I still think we should try to have a good UI. Please move the switch cms button below the caption and make sure the whole thing (including the `Caption`) is scrollable by replacing `Layout` with `LayoutedScrollView`.

steffenkleinle (changes requested):

native/src/routes/Settings.tsx
```diff
@@ -42,22 +50,31 @@ const Settings = ({ navigation }: SettingsProps): ReactElement => {
...
+          data={sections}
+          extraData={appContext.settings}
+          renderItem={renderItem}
+          ItemSeparatorComponent={Divider}
+          ListFooterComponent={Divider}
+          // Fixes VirtualizedList should never be nested inside plain ScrollViews
+          scrollEnabled={false}
+        />
+      </Layout>
+    </LayoutedScrollView>
```
steffenkleinle, line 75: 🔧 The `Layout` can be removed and pls move the switch button below the caption.

steffenkleinle (approved):
Not tested again, great work!

native/src/routes/Settings.tsx
```diff
@@ -24,6 +26,11 @@ const Settings = ({ navigation }: SettingsProps): ReactElement => {
   const { t } = useTranslation('settings')
   const { settings } = appContext
 
+  const clearResourcesAndCache = () => {
+    dataContainer.clearInMemoryCache()
+    dataContainer._clearOfflineCache().catch(reportError)
+  }
+
```
bahaaTuffaha, line 34: ❓ How about forcing a reset to regionsPage that will fetch data again instead of a pull ?

```suggestion
  const clearResourcesAndCache = () => {
    dataContainer.clearInMemoryCache()
    dataContainer._clearOfflineCache().catch(reportError)
    navigation.reset({ index: 0, routes: [{ name: REGIONS_ROUTE }] })
  }

```

bahaaTuffaha (approved):
Nice work ✨ 
Tested on Android and iOS ✅

native/src/routes/Settings.tsx
```diff
@@ -42,22 +49,29 @@ const Settings = ({ navigation }: SettingsProps): ReactElement => {
...
       <Divider />
       <FlatList
         data={sections}
         extraData={appContext.settings}
         renderItem={renderItem}
         ItemSeparatorComponent={Divider}
         ListFooterComponent={Divider}
+        // Fixes VirtualizedList should never be nested inside plain ScrollViews
+        scrollEnabled={false}
       />
```
bahaaTuffaha, line 74: ❓  Do you think FlatList here could be replaced with .map() without affecting performance? or this workaround is fine?

native/src/routes/Settings.tsx
lunars97: if it makes resfrsh without manual refresh, I think we could add it :+1: , what do you think @steffenkleinle ?

steffenkleinle: Good idea!

native/src/routes/Settings.tsx
steffenkleinle: I think either is fine, both the workaround and just using map.

PR #4081, opened 2026-04-27
3478: Improve accessibility by Zoom-in Zoom-out buttons

### LINKED ISSUE #3478: Add Zoom in and Zoom out buttons for Poi map
### Describe the Problem

This zoom in and zoom out buttons will allow keyboard users to be able to zoom just by focusing on the these buttons.

- Add 2 buttons on the right of the screen on top of the map one for zooming in the map camera and one for zooming out. (Design needs discussion)
- Make sure the buttons are responsive when the bottom-sheet minimize/maximize.

N/A.

### Additional Information
Check the keyboard Focus implementation  #3369

bahaaTuffaha:
### Short Description
This zoom in and zoom out buttons will allow keyboard users to be able to zoom just by focusing on the these buttons.

⚠️  This is a sub-branch of `#4056 Upgrade maplibre to v11`

### Proposed Changes

- Created a state `mapViewRef` at web to control the zoom.
- New component called `MapZoomControls.tsx`:
   - For web I used `:focus` and right to show the button like skip to content. (for desktop it already has `NavigationControl` and it's not hidden).
   - For native used `Pressable` that has `onFocus` and `onBlur` and  its hidden by opacity (I cant use right to move it offscreen because the keyboard tab will not focus on it).
   - I had an issue at native that I can't reach the Zoom button while using keyboard tab so I adjusted the `BottomSheetHandle` to forward the focus (nextFocusForward) to the zoom button next using TouchableOpacity.

### Side Effects

- At native the area where the buttons are located while not focused on (opacity 0) are rendered and could affect pressing on the map on that area.
- This focus forwarding workaround on native doesn't work on iOS. (Settings -> Accessibility -> Keyboards -> Full Keyboard Access -> On).

### Testing

- Launch Integreat and go to poi/map.
- Press tab on keyboard to reach the zoom buttons (on native press down arrow to reach the minus  after you reach the plus)
- Test web on desktop and mobile screens (on web it should be visible all the time).
- Test contrast mode.

### Resolved Issues

Fixes: #3478 

---

### DIFF
--- a/native/src/components/BottomSheetHandle.tsx
+++ b/native/src/components/BottomSheetHandle.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react'
 import styled from 'styled-components/native'
 
-const Handle = styled.View`
+const Handle = styled.TouchableOpacity`
   width: 34px;
   border: 1px solid ${props => props.theme.colors.onSurfaceVariant};
   background-color: ${props => props.theme.colors.onSurfaceVariant};
@@ -10,6 +10,12 @@ const Handle = styled.View`
   margin: 20px 0;
 `
 
-const BottomSheetHandle = (): ReactElement => <Handle />
+type BottomSheetHandleProps = {
+  nextFocusForward?: number
+}
+
+const BottomSheetHandle = ({ nextFocusForward }: BottomSheetHandleProps): ReactElement => (
+  <Handle focusable nextFocusForward={nextFocusForward} />
+)
 
 export default BottomSheetHandle
--- a/native/src/components/BottomSheetHandle.tsx
+++ b/native/src/components/BottomSheetHandle.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react'
 import styled from 'styled-components/native'
 
-const Handle = styled.TouchableOpacity`
+const Handle = styled.Pressable`
   width: 34px;
   border: 1px solid ${props => props.theme.colors.onSurfaceVariant};
   background-color: ${props => props.theme.colors.onSurfaceVariant};
@@ -15,6 +15,7 @@ type BottomSheetHandleProps = {
 }
 
 const BottomSheetHandle = ({ nextFocusForward }: BottomSheetHandleProps): ReactElement => (
+  // @ts-expect-error Pressable doesn't have a type for nextFocusForward but it is a valid prop
   <Handle focusable nextFocusForward={nextFocusForward} />
 )
 
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -14,9 +14,9 @@ import {
   useCurrentPosition,
 } from '@maplibre/maplibre-react-native'
 import type { BBox, Feature, GeoJsonProperties, Geometry, Position } from 'geojson'
-import React, { ReactElement, useCallback, useEffect, useRef, useState } from 'react'
+import React, { ReactElement, type Ref, useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import type { NativeSyntheticEvent } from 'react-native'
+import type { NativeSyntheticEvent, View } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 
 import {
@@ -39,6 +39,7 @@ import { clusterCountLayer, clusterLayer, markerLayer } from '../constants/layer
 import useUserLocation from '../hooks/useUserLocation'
 import { conditionalA11yProps } from '../utils/helpers'
 import { reportError } from '../utils/sentry'
+import MapZoomControls from './MapZoomControls'
 import MapAttribution from './MapsAttribution'
 import Icon from './base/Icon'
 import IconButton from './base/IconButton'
@@ -91,6 +92,7 @@ type MapViewProps = {
   bottomSheetFullscreen: boolean
   zoom: number | undefined
   Overlay?: ReactElement
+  zoomInRef?: Ref<View>
 }
 
 const MapView = ({
@@ -104,6 +106,7 @@ const MapView = ({
   bottomSheetHeight,
   bottomSheetFullscreen,
   zoom,
+  zoomInRef,
 }: MapViewProps): ReactElement => {
   const toLngLat = (position: Position): LngLat => {
     const [longitude, latitude] = position
@@ -213,6 +216,20 @@ const MapView = ({
     zoomOnClusterPress(pressedCoordinates)
   }
 
+  const zoomBy = useCallback(async (delta: number) => {
+    if (mapRef.current === null) {
+      return
+    }
+
+    const currentZoom = await mapRef.current.getZoom()
+
+    setCameraSettings({
+      zoom: currentZoom + delta,
+      duration: animationDuration,
+      easing: 'ease',
+    })
+  }, [])
+
   const locationPermissionGrantedIcon = followUserLocation ? 'crosshairs-gps' : 'crosshairs'
   const locationPermissionIcon = userLocation ? locationPermissionGrantedIcon : 'crosshairs-off'
 
@@ -259,6 +276,12 @@ const MapView = ({
         onPress={onRequestLocation}
         position={bottomSheetHeight}
         accessibilityLabel={t('showOwnLocation')}
+      />
+      <MapZoomControls
+        onZoomIn={() => zoomBy(1)}
+        onZoomOut={() => zoomBy(-1)}
+        bottomSheetHeight={bottomSheetHeight}
+        zoomInRef={zoomInRef}
       />
     </OuterWrapper>
   )
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -91,7 +91,7 @@ type MapViewProps = {
   bottomSheetFullscreen: boolean
   zoom: number | undefined
   Overlay?: ReactElement
-  zoomInRef?: Ref<View>
+  zoomRef?: Ref<View>
 }
 
 const MapView = ({
@@ -106,7 +106,7 @@ const MapView = ({
   bottomSheetMidHeight,
   bottomSheetFullscreen,
   zoom,
-  zoomInRef,
+  zoomRef,
 }: MapViewProps): ReactElement => {
   const mapRef = useRef<MapRef>(null)
   const cameraRef = useRef<CameraRef>(null)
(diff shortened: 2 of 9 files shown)


native/src/components/MapZoomControls.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { ReactElement, Ref, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Pressable, StyleSheet, type StyleProp, type View, type ViewStyle } from 'react-native'
+import { useTheme } from 'styled-components/native'
+
+import Icon from './base/Icon'
+
+// 50px button height + 16px margin
```
f1sh1918: 🙃 the button height is 48 not 50 afaics

web/src/components/PoisMobile.tsx
```diff
@@ -37,6 +38,11 @@ const GeocontrolContainer = styled(AttributionContainer)`
   margin-bottom: 24px;
 `
 
+const ZoomControlsContainer = styled(AttributionContainer)`
+  right: 8px;
+  margin-bottom: 66px;
```
f1sh1918: 🔧 this margin should be commented and maybe also calculated. It looks like a random value

f1sh1918 (approved):
In general works great 👍 
Just some minor things.

As far as i understood this is only for external keyboard users?
Since TalkBack and VoiceOver have their separate zooming system and don't need this buttons at all.
Just because i had some a11y concerns with hidden buttons.

web/src/components/PoisMobile.tsx
bahaaTuffaha: Yes `29px` for geolocate button height and `24px` for GeocontrolContainer margin-bottom so in total `53px`

53 + (8 space margin) = 61 then to make it dividable by 8 we can use closest number that is 64

native/src/components/BottomSheetHandle.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { ReactElement } from 'react'
 import styled from 'styled-components/native'
 
-const Handle = styled.View`
+const Handle = styled.TouchableOpacity`
```
steffenkleinle: 🤔 Any reason not to use a `Pressable`?
From the [docs](https://reactnative.dev/docs/touchableopacity):
> If you're looking for a more extensive and future-proof way to handle touch-based input, check out the [Pressable](https://reactnative.dev/docs/pressable) API.

native/src/components/MapView.tsx
```diff
@@ -251,6 +277,12 @@ const MapView = ({
         position={bottomSheetHeight}
         accessibilityLabel={t('showOwnLocation')}
       />
+      <MapZoomControls
+        onZoomIn={() => zoomBy(1)}
+        onZoomOut={() => zoomBy(-1)}
+        bottomSheetHeight={bottomSheetHeight}
+        zoomInRef={zoomInRef}
```
steffenkleinle: 🔧 Let's keep naming consistent here:
```suggestion
        zoomIn={() => zoom(1)}
        zoomOut={() => zoom(-1)}
        bottomSheetHeight={bottomSheetHeight}
        ref={zoomRef}
```
We could even merge `zoomIn` and `zoomOut` and handle it internally in `MapZoomControls`:
```suggestion
        zoomIn={zoom}
        bottomSheetHeight={bottomSheetHeight}
        ref={zoomRef}
```

translations/translations.json
```diff
@@ -11561,7 +11561,9 @@
       "poisCount_few": "{{count}} Orte",
       "poisCount_many": "{{count}} Orte",
       "poisCount_other": "{{count}} Orte",
-      "contacts": "Kontakte"
+      "contacts": "Kontakte",
+      "zoomIn": "Zoom in",
+      "zoomOut": "Zoom out"
```
steffenkleinle: 🔧 Either of the two, not sure which one:
```suggestion
      "zoomIn": "Hereinzommen",
      "zoomOut": "Herauszoomen"
```
or
```suggestion
      "zoomIn": "Vergrößern",
      "zoomOut": "Verkleinern"
```

native/src/components/MapView.tsx
```diff
@@ -206,9 +213,23 @@ const MapView = ({
     const feature = featureCollection.find((it): it is MapFeature => it.geometry.type === 'Point')
     selectFeature(feature ?? null)
 
-    await zoomOnClusterPress(pressedCoordinates)
+    zoomOnClusterPress(pressedCoordinates)
   }
 
+  const zoomBy = useCallback(async (delta: number) => {
```
steffenkleinle: 🤔 Is the `useCallback` really necessary?

native/src/components/MapZoomControls.tsx
```diff
@@ -0,0 +1,79 @@
...
+        style={{ bottom: bottomSheetHeight + buttonSpacing }}
+        onPress={onZoomOut}
+        accessibilityLabel={t('zoomOut')}>
+        <Icon source='minus' color={theme.colors.onSurface} />
+      </FocusedButton>
+      <FocusedButton
+        style={{ bottom: bottomSheetHeight + buttonSpacing * 2 }}
+        onPress={onZoomIn}
+        accessibilityLabel={t('zoomIn')}
+        ref={zoomInRef}>
```
steffenkleinle: ❓ Why do we put the ref (and therefore the next focus on the second (zoom in) button? Doesn't that mean that we can't really reach the zoom out button with the keyboard?

native/src/components/PoisBottomSheet.tsx
```diff
@@ -89,6 +91,11 @@ const PoisBottomSheet = ({
     bottomSheetRef.current?.snapToIndex(1)
   }
 
+  const HandleComponent = useCallback(
```
steffenkleinle, line 90: 🤔 Is the `useCallback` really necessary?

native/src/components/MapZoomControls.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <>
+      <FocusedButton
+        style={{ bottom: bottomSheetHeight + buttonSpacing }}
+        onPress={onZoomOut}
+        accessibilityLabel={t('zoomOut')}>
+        <Icon source='minus' color={theme.colors.onSurface} />
+      </FocusedButton>
+      <FocusedButton
+        style={{ bottom: bottomSheetHeight + buttonSpacing * 2 }}
```
steffenkleinle: 🔧 Wouldn't it make more sense to just add one `bottom` to the parent component instead of setting it to both buttons individually?

web/src/components/PoisMobile.tsx
```diff
@@ -37,6 +38,14 @@ const GeocontrolContainer = styled(AttributionContainer)`
   margin-bottom: 24px;
 `
 
+// margin-bottom for ZoomControlsContainer is calculated as follows:
+// The Geocontrol height with its margin-bottom = 29px + 24px
+// = 53px + 8px gap rounded up to 64px
+const ZoomControlsContainer = styled(AttributionContainer)`
+  right: 8px;
+  margin-bottom: 64px;
+`
```
steffenkleinle: 🔧 Wouldn't it be easier to create one container for both geocontrol and zoom controls so a calculation like this is not necessary?

web/src/components/MapZoomControls.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    :focus {
+      inset-inline-end: 0;
+    }
+  }
+`
+
+type MapZoomControlsProps = {
+  onZoomIn: () => void
+  onZoomOut: () => void
```
steffenkleinle: 🔧 
```suggestion
  zoomIn: () => void
  zoomOut: () => void
```
or even
``` suggestion
  zoom: (zoom: number) => void
```

steffenkleinle (changes requested):
Tested on firefox and android. On android, the focus shape is a square and I can't reach the zoom out button.

Image

bahaaTuffaha:
> Tested on firefox and android. On android, the focus shape is a square and I can't reach the zoom out button.

"(on native press down arrow to reach the minus after you reach the plus button)"

steffenkleinle:
> > Tested on firefox and android. On android, the focus shape is a square and I can't reach the zoom out button.
> 
> "(on native press down arrow to reach the minus after you reach the plus button)"

Isn't that kind of counterintuitive?

native/src/components/MapZoomControls.tsx
bahaaTuffaha: hmm they can but with arrows... I can add another `nextFocusForward` just to move from zoomin to zoomout

native/src/components/BottomSheetHandle.tsx
bahaaTuffaha: because it has a nextFocusForward. I tested nextFocusForward on Pressable it works but I have to suppress typescript because there is no type for it.

native/src/components/PoisBottomSheet.tsx
bahaaTuffaha:
```
Do not define components during render. React will see a new component type on every render and destroy the entire subtree’s DOM nodes and state (https://reactjs.org/docs/reconciliation.html#elements-of-different-types). Instead, move this component definition out of the parent component “PoisBottomSheet” and pass data as props. If you want to allow component creation in props, set allowAsProps option to true.eslint[react/no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/no-unstable-nested-components.md)
```

native/src/components/MapView.tsx
```diff
@@ -218,6 +221,7 @@ const MapView = ({
         position={bottomSheetHeight}
         accessibilityLabel={t('showOwnLocation')}
       />
+      <MapZoomControls mapRef={mapRef} cameraRef={cameraRef} bottomSheetHeight={bottomSheetHeight} ref={zoomRef} />
```
steffenkleinle: 🔧 I guess it would also be better to move this in a container together with `StyledIcon` (i.e. the user location button) above to avoid the bottom calculation.

native/src/components/MapZoomControls.tsx
```diff
@@ -0,0 +1,99 @@
...
+type FocusedButtonProps = {
+  children: ReactElement
+  accessibilityLabel: string
+  ref?: Ref<View>
+  style?: StyleProp<ViewStyle>
+  onPress: () => void
+  nextFocusForward?: number
+}
+
+const FocusedButton = ({ children, accessibilityLabel, ref, style, onPress, nextFocusForward }: FocusedButtonProps) => {
```
steffenkleinle: 🤔 Maybe `FocusedOnlyButton`

steffenkleinle (approved):
Not tested again, nice work

PR #4083, opened 2026-04-29
3842: Autofocus search bar

### LINKED ISSUE #3842: Autofocus search bar in region selection
### Describe the Problem

When changing the region, I first have to manually select the search field for region names. As far as I can see there would be no downside to having the search field directly focused so that I can immediately start typing.

Focus search field when opening region selection.

Keep as is.

### Related Issues

None.

### Additional Information

Not really important, most users probably do not change the region that often.

f1sh1918:
### Short Description

The main goal for the user is to find the right location which usually start with searching for it. Therefore the search field should be auto focused.

### Proposed Changes

- add autofocus for web

### Side Effects

- none

### Testing

- Open the region selects and check that the search field is auto focused by default

### Resolved Issues

Fixes: #3842

---


steffenkleinle (approved):
Tested on firefox, works as expected

bahaaTuffaha (approved):
Tested on Chrome based browser ✅

PR #4084, opened 2026-04-29
3487: [iOS] External keyboard navigation

### LINKED ISSUE #3487: Support external keyboard usage - iOS ~ Navigation with arrow keys only works inside text fields
### Describe the Bug

Navigation with arrow keys only works inside text fields
-> Within a text field, arrow keys (up, down, left, right, tab) work correctly
-> But outside of text fields (e.g., in menus or on the map), the arrow keys do not allow navigation between menu items or moving the map up and down

### Steps to Reproduce

1.    Connect an external keyboard to an iOS device
2.    Open the integreat app
3.    Navigate to a page with multiple menu items or a map
4.    Try to navigate using the arrow keys

### Expected Behavior

Arrow keys should allow navigation between menu items and map movement, not only inside text fields

f1sh1918:
### Short Description

With the introduction of react-native paper, we also introduced an a11y trap

### Proposed Changes

- use standard view instead of AppBar.Header to avoid a11y trap

### Side Effects

- none

### Testing

1. use an iphone real device, connect external keyboard. 
2. enable `Settings -> Accessibility -> Keyboards -> Full Keyboard Access -> On`
3. open information page and navigation by using TAB to the tile list and then you should be able to navigate with arrow buttons inside
4. Please check that the app bar header still looks and works as expected

It also works on android
### Resolved Issues

Fixes: #3487 

---


steffenkleinle (approved):
LGTM, not tested as I neither have an iOS real device nor a keyboard.

bahaaTuffaha (approved):
Tested on Andorid ✅ 
On iOS : it works better now with lists except the poiDetails's bottomSheet. the onFocus prop doesn't work on iOS it seems.

f1sh1918:
> Tested on Andorid ✅ On iOS : it works better now with lists except the poiDetails's bottomSheet. the onFocus prop doesn't work on iOS it seems.

We can create a separate issue for that

kalairubachandrakumaran-oss:
The keyboard navigation works much better now. It is no longer trapped in the header and  I can navigate to the tiles and the bottom navigation using the arrow keys. ✅ 

@steffenkleinle @f1sh1918 
But  I noticed the following behavior:
- If I open a tile using touch, I can navigate through the individual menu items on the next page with the arrow keys
- If I open the same tile using the arrow keys (keyboard), I can open the page, but I cannot navigate through the individual menu items on that page using the arrow keys

I'm not sure but it still looks like an issue, how should I proceed now? Should I create a separate task for this?
compare https://github.com/user-attachments/assets/5d43580a-cb7e-4a70-88d4-fadd4db3e433

steffenkleinle:
Yes, please create an issue, thanks for testing! :)

PR #4090, opened 2026-04-30
4087: Detect and fix web's circular dependencies

### LINKED ISSUE #4087: Detect and fix circular dependencies (web)
### Describe the Problem

This is a sub-issue of #3146

Detect and fix circular dependencies at web.

N/A

bahaaTuffaha:
### Short Description

Detect and fix web circular dependencies

### Proposed Changes

- Adjusted the script  [check_circular_deps.sh](https://github.com/digitalfabrik/entitlementcard/blob/811a0e269292b8301b64a319f2f2727c0a931934/administration/scripts/check_circular_deps.sh) to show number of circular-deps for each workspace and it has verbose optional argument to show the lists.
- Added another option called `--current` to show the list at the current workspace and exclude others.
- Created the following new files:
    - constants/dimensions.ts
    - components/TtsContext.ts (not sure if its good here we could create a new directory like native called `/contexts`)
#### Fixed the following circular dependencies:
`1) components/BottomNavigation.tsx > hooks/useDimensions.ts`
- Moved BOTTOM_NAVIGATION_ELEMENT_ID from BottomNavigation.tsx to constants/dimensions.ts

`2) components/Layout.tsx > components/SkipToContent.tsx`
- Moved MAIN_ELEMENT_ID from Layout.tsx to constants/dimensions.ts
- Updated SkipToContent.tsx to import from constants/dimensions

`3) hooks/useDimensions.ts > components/RegionContentToolbar.tsx > components/FeedbackToolbarItem.tsx > components/base/Dialog.tsx > components/Layout.tsx`
`4) hooks/useDimensions.ts > components/RegionContentToolbar.tsx > components/FeedbackToolbarItem.tsx > components/base/Dialog.tsx`
- Moved LAYOUT_ELEMENT_ID from Layout.tsx to constants/dimensions.ts
- Updated Dialog.tsx, AlertDialog.tsx, Sidebar.tsx to import from constants/dimensions

`5) hooks/useDimensions.ts > components/ThemeContainer.tsx`
- Moved BREAKPOINTS from ThemeContainer.tsx to constants/dimensions.ts

`6) hooks/useDimensions.ts > components/TtsContainer.tsx > components/TtsPlayer.tsx`
- Created components/TtsContext.ts - extracted TtsContext + TtsContextType
- Moved TTS_PLAYER_ELEMENT_ID from TtsPlayer.tsx to constants/dimensions.ts
- Updated TtsContainer.tsx, TtsPlayer.tsx, useTtsPlayer.ts, RegionContentMenu.tsx, ChatContainer.tsx to import from components/TtsContext

`7) components/RemoteContent.tsx > utils/iframes.ts`
- Moved IframeSources type + IFRAME_BLANK_SOURCE constant from RemoteContent.tsx to utils/iframes.ts`

`8) RegionContentNavigator.tsx > routes/CategoriesPage.tsx`
`9) RegionContentNavigator.tsx > routes/CategoriesPage.tsx > components/CategoriesContent.tsx`
`10) RegionContentNavigator.tsx > routes/CategoriesPage.tsx > components/CategoriesContent.tsx > components/EmbeddedOffers.tsx`
`11) RegionContentNavigator.tsx > routes/EventsPage.tsx`
`12) RegionContentNavigator.tsx > routes/ImprintPage.tsx`
`13) RegionContentNavigator.tsx > routes/LocalNewsPage.tsx`
`14) RegionContentNavigator.tsx > routes/PoisPage.tsx`
`15) RegionContentNavigator.tsx > routes/SearchPage.tsx`
`16) RegionContentNavigator.tsx > routes/TuNewsDetailPage.tsx`
`17) RegionContentNavigator.tsx > routes/TuNewsPage.tsx`
- Moved RegionRouteProps type from RegionContentNavigator.tsx to routes/index.ts
- And adjusted the rest of the files that depends on RegionRouteProps

### Side Effects

- none.

### Testing

- `npm install --global madge`
- Test `yarn check-circular-deps` at `Root` , `web`, `native` and `shared`.
- web should return `No circular dependency found!`
- Launch web and do a smoke test .

### Resolved Issues

Fixes: #4087

---


f1sh1918 (approved):

tools/checkCircularDependencies.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+# Checks circular dependencies via madge.
```
steffenkleinle: ❓ Did you write this yourself? If not, could we add a comment with the source?

native/package.json
```diff
@@ -28,6 +28,7 @@
     "android:reverse": "adb reverse tcp:8081 tcp:8081",
     "android:reinit": "adb shell pm clear com.integreat && adb shell am start com.integreat/com.integreat.MainActivity",
     "android:start-avd": "bash -c '$ANDROID_HOME/emulator/emulator -avd $($ANDROID_HOME/emulator/emulator -list-avds | head -n 1)'",
+    "check-circular-deps": "bash ../scripts/checkCircularDeps.sh --current",
```
steffenkleinle: 🙃 Personally I'd prefer to avoid abbreviations, same for others.
🔧 Also, lets move this to tools to avoid having another dir.
🔧 `bash` is not necessary (you might need to run `chmod +x check-circular-dependencies` though and commit the changes.
```suggestion
    "check-circular-dependencies": "../tools/checkCircularDependencies.sh --current",
```

shared/package.json
```diff
@@ -9,6 +9,7 @@
     "test": "jest",
     "test:changed": "yarn run test -o",
     "test:watch": "yarn run test --watchAll",
+    "check-circular-deps": "bash ../scripts/checkCircularDeps.sh --current 0 ./index.ts",
```
steffenkleinle: ❓ Why do we here and for the root but use `0` but not for web and native?

web/src/constants/layout.ts
steffenkleinle, line 1: 🤔 Not sure if `dimensions` is really fitting, perhaps rather `layout.ts`?

steffenkleinle (changes requested):
Really nice work, thanks for baring with this tedious task! 💚

steffenkleinle:
❌ The commands to check the deps currently don't work in the workspaces (`Madge is not installed`)

web/src/contexts/TtsContext.ts
steffenkleinle, line 1: 🔧 This is not a component and should probably be moved to `contexts/TtsContext.ts`?

tools/checkCircularDependencies.sh
```diff
@@ -0,0 +1,101 @@
...
+    check_workspace "$(pwd)" "$entry_point" "$workspace_name" "$exclude_pattern"
+else
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+    check_workspace "$PROJECT_ROOT/web" ./src/index.tsx "web" "\.\.\/shared\/"
+    echo ""
+    check_workspace "$PROJECT_ROOT/native" ./src/index.tsx "native" "\.\.\/shared\/"
+    echo ""
+    check_workspace "$PROJECT_ROOT/shared" ./index.ts "shared"
```
steffenkleinle: ❓ Will you do `build-configs` and `translation` in separate PRs?

tools/checkCircularDependencies.sh
bahaaTuffaha: Mix between this article https://dev.to/greenroach/detecting-circular-dependencies-in-a-reacttypescript-app-using-madge-229 and ai 😅

bahaaTuffaha:
> The commands to check the deps currently don't work in the workspaces (`Madge is not installed`)

did you install Madge ?

shared/package.json
bahaaTuffaha: I will remove the zeros cuz it's the default 👍

tools/checkCircularDependencies.sh
bahaaTuffaha: sure I will create a sub-issue for that

steffenkleinle:
> > The commands to check the deps currently don't work in the workspaces (`Madge is not installed`)
> 
> did you install Madge ?

Yes, in the root it works.

package.json
```diff
@@ -55,6 +56,7 @@
     "husky": "^9.1.7",
     "jest": "^30.3.0",
     "jest-junit": "^16.0.0",
+    "madge": "^8.0.0",
```
steffenkleinle, line 59: 🤔 Does it perhaps add more sense to install madge to the tools dir since we use it there? And then add a `yarn` command in tools and use that here and in web/native/shared?

tools/checkCircularDependencies.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
steffenkleinle, line 1: ❓ Maybe we should rather simplify this as much as possible without options we don't need/use? I think we can use the `verbose` and `current` as default and get rid of the `entry_point` and `limit` parameters as we are not going to use them.

steffenkleinle: ❓ The question is actually if we even need to ever execute this command separately just for one workspace (i.e. web, shared, ...). I think it would be enough to always just execute this in the root for all workspaces. This allows to a) not require installing `madge` globally b) helps keep the package.json simple.

package.json
bahaaTuffaha: now it runs from root so no need to add anything or the need for installing it globally

steffenkleinle (approved):
Nice work!

PR #4093, opened 2026-05-04
2404: Switch to esm

### LINKED ISSUE #2404: Convert to ESM
Some dependencies nowadays migrated their packages to ES modules (ESM). It is trivial to use commonjs modules in ESM, but not the other way around. For us this means we cannot use those dependencies anymore. 
Following [this blogpost](https://molily.de/ecmascript-modules/) most browsers should support ESM modules, with Edge being the last one in 2020. We should at least discuss whether we also want to adapt to ESM or whether we want to wait and see how this develops. 
ESM supports tree-shaking, lazy imports and is compatible with webpack. Regarding react-native compatibility/Mobile compatibility I don't have any information.

Discuss whether we should/want to migrate to ESM. Most of the work be done automatically

**Additional context**
Short summary of the main differences: https://dev.to/abbeyperini/tldr-commonjs-vs-esm-47dk
Longer summary: https://www.hoeser.dev/blog/2023-02-21-cjs-vs-esm/

lunars97:
### Short Description

This PR migrates the codebase to ES modules while keeping Jest running tests in CommonJS mode. After some trials and fails in jest esm hell :grimacing:, I was forced to keep jest run in CommonJS because of the memory leaks especially when i use it in native and web for dynamic imports as well as error prone test rewrites and import issues, see the open issue for the memory leak issue https://github.com/jestjs/jest/issues/14605 Seems like jest esm is unstable to use it in our project as the poor "documentation"(blog page) for esm support shows it https://jestjs.io/docs/ecmascript-modules

### Proposed Changes

- [x] for build-configs added and .js extension as the `moduleResolution`: `NodeNext` requires it
- [x] for translations  added .js extension as the `moduleResolution`:`NodeNext` requires it
- [x] for shared added exports map as node enforces the exports map for different entries which solves the `... is not defined by exports` issue and  .js extension as the `moduleResolution`: `NodeNext` requires it
- [x] for web added `moduleResolution: bundler` which is more loose and does not require .js extension usage and settings explicitly type as well as rewrote CommonJS imports/exports to esm
- [x] for native added `moduleResolution: bundler` which is more loose and does not require .js extension usage and settings explicitly type as well as rewrote CommonJS imports/exports to esm

Also added explicitly `type` in some places because of the import error `should be listed in the project's dependencies, not devDependencies. (import-x/no-extraneous-dependencies)`

### Side Effects

I hope none :see_no_evil: 

### Testing
- Smoke test the web and native on ios and android (checkout malte, aschaffenburg and obdach as well :pray: )
- Run yarn ts:check, yarn lint and yarn test

### Resolved Issues

Fixes: #2404 

---


lunars97:
I tried to run it on iOS and faced this issue with constants I guess, should I do something here as well? 🤔 
Screenshot 2026-06-01 at 09 15 39

lunars97:
> I tried to run it on iOS and faced this issue with constants I guess, should I do something here as well? 🤔 Screenshot 2026-06-01 at 09 15 39

It was because of `lodash` esm compatibility issue in translations, replaced `lodash` with `lodash-es`

steffenkleinle:
❓ Is it possible to also use `moduleResolution: bundler` for `build-configs`, `shared` and `translations` to avoid having to use `.js` extensions?

lunars97:
> ❓ Is it possible to also use `moduleResolution: bundler` for `build-configs`, `shared` and `translations` to avoid having to use `.js` extensions?

I am not sure about, I think it can lead to import issue on native and web 🤔 , as I understood it is used for code that gets bundled, in our case `native` with metro and `web` with webpack, I might also lack there some understanding, and others are consumed by node and they are used as libraries and it requires them to have strict rules e.g having js extension and explicitly adding type for type imports

bahaaTuffaha (commented):
I can't start web (I deleted all node_modules and reinstalled):
```bash
[webpack-cli] Failed to load 'tools/webpack.config.ts' config
▶ ESM (`import`) failed:
  Unknown file extension ".ts" for /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts

▶ CJS (`require`) failed:
  Must use import to load ES Module: /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts
  require() of ES modules is not supported.
  require() of /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts from /Users/digitalfabrik/Repos/integreat-app/node_modules/webpack-cli/lib/webpack-cli.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
  Instead change the requiring code to use import(), or remove "type": "module" from /Users/digitalfabrik/Repos/integreat-app/web/package.json.
```

and tests fails:
```bash
 FAIL  src/utils/__tests__/createSettingsSections.spec.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation, specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation
```

lunars97:
> I can't start web (I deleted all node_modules and reinstalled):
> 
> ```shell
> [webpack-cli] Failed to load 'tools/webpack.config.ts' config
> ▶ ESM (`import`) failed:
>   Unknown file extension ".ts" for /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts
> 
> ▶ CJS (`require`) failed:
>   Must use import to load ES Module: /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts
>   require() of ES modules is not supported.
>   require() of /Users/digitalfabrik/Repos/integreat-app/web/tools/webpack.config.ts from /Users/digitalfabrik/Repos/integreat-app/node_modules/webpack-cli/lib/webpack-cli.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
>   Instead change the requiring code to use import(), or remove "type": "module" from /Users/digitalfabrik/Repos/integreat-app/web/package.json.
> ```
> 
> and tests fails:
> 
> ```shell
>  FAIL  src/utils/__tests__/createSettingsSections.spec.ts
>   ● Test suite failed to run
> 
>     Jest encountered an unexpected token
> 
>     Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
> 
>     Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.
> 
>     By default "node_modules" folder is ignored by transformers.
> 
>     Here's what you can do:
>      • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
>      • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
>      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
>      • If you need a custom transformation, specify a "transform" option in your config.
>      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
> 
>     You'll find more details and examples of these config options in the docs:
>     https://jestjs.io/docs/configuration
>     For information about custom transformations, see:
>     https://jestjs.io/docs/code-transformation
> ```

Thank you for checking it out :green_heart: Did you try running `yarn ts:check` before running the web? Because of stale compiled `dist` folders it can lead to mismatch and fail

bahaaTuffaha:
> Thank you for checking it out 💚 Did you try running `yarn ts:check` before running the web? Because of stale compiled `dist` folders it can lead to mismatch and fail

I tried to run `yarn ts:check` before but didn't work. Let me check if there is another way...

bahaaTuffaha:
try  `NODE_OPTIONS='--import tsx/esm'`
```json
  "scripts": {
    "webpack-ts": "cross-env NODE_OPTIONS='--import tsx/esm' webpack --config tools/webpack.config.ts --env bundle_analyzer",
    "webpack-dev-server-ts": "cross-env NODE_OPTIONS='--import tsx/esm' webpack-dev-server --config tools/webpack.config.ts --progress",
```

steffenkleinle:
> > ❓ Is it possible to also use `moduleResolution: bundler` for `build-configs`, `shared` and `translations` to avoid having to use `.js` extensions?
> 
> I am not sure about, I think it can lead to import issue on native and web 🤔 , as I understood it is used for code that gets bundled, in our case `native` with metro and `web` with webpack, I might also lack there some understanding, and others are consumed by node and they are used as libraries and it requires them to have strict rules e.g having js extension and explicitly adding type for type imports

Makes sense. Did you try it though? I think it would be preferable, so trying would be a nice thing. If you don't want to, I can also take care of that.

lunars97:
> > > ❓ Is it possible to also use `moduleResolution: bundler` for `build-configs`, `shared` and `translations` to avoid having to use `.js` extensions?
> > 
> > 
> > I am not sure about, I think it can lead to import issue on native and web 🤔 , as I understood it is used for code that gets bundled, in our case `native` with metro and `web` with webpack, I might also lack there some understanding, and others are consumed by node and they are used as libraries and it requires them to have strict rules e.g having js extension and explicitly adding type for type imports
> 
> Makes sense. Did you try it though? I think it would be preferable, so trying would be a nice thing. If you don't want to, I can also take care of that.

i will have a look :+1:

lunars97:
> > > > ❓ Is it possible to also use `moduleResolution: bundler` for `build-configs`, `shared` and `translations` to avoid having to use `.js` extensions?
> > > 
> > > 
> > > I am not sure about, I think it can lead to import issue on native and web 🤔 , as I understood it is used for code that gets bundled, in our case `native` with metro and `web` with webpack, I might also lack there some understanding, and others are consumed by node and they are used as libraries and it requires them to have strict rules e.g having js extension and explicitly adding type for type imports
> > 
> > 
> > Makes sense. Did you try it though? I think it would be preferable, so trying would be a nice thing. If you don't want to, I can also take care of that.
> 
> i will have a look 👍

Seems like it leads to `Error [ERR_MODULE_NOT_FOUND]: Cannot find module` and `Error [ERR_UNSUPPORTED_DIR_IMPORT]: ` It means that we cannot avoid adding `.js` extensions  and there is a runtime need for the exports map

native/src/components/Highlighter.tsx
```diff
@@ -26,7 +26,7 @@ const Highlighter = ({ search, text, style }: HighlighterProps): ReactElement =>
     <Text style={style}>
       {chunks.map(chunk => {
         const matchedText = text.substring(chunk.start, chunk.end)
-        return chunk.highlight ? (
+        return chunk.highlight === true ? (
```
steffenkleinle, line 29: 🤔 Why is this necessary? Are the types somehow not correctly loaded anymore?

native/metro.config.ci.js
```diff
@@ -1,7 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
```
steffenkleinle: 🤔 Can we remove the eslint directive?
```suggestion
```

native/metro.config.js
```diff
@@ -1,9 +1,14 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
```
steffenkleinle: 🔧 
```suggestion
```

shared/api/endpoints/testing/EventModelBuilder.ts
```diff
@@ -1,11 +1,13 @@
 import { sha256 } from 'js-sha256'
 import { DateTime } from 'luxon'
-import { rrulestr } from 'rrule'
+import * as rrule from 'rrule'
```
steffenkleinle: 🤔 Why was this change necessary? Same for other usages
```suggestion
import { rrulestr } from 'rrule'
```

shared/api/endpoints/createCitiesEndpoint.ts
steffenkleinle, line 1: 🔧 This file should be deleted

shared/api/endpoints/createCityEndpoint.ts
steffenkleinle, line 1: 🔧 This file should be deleted

shared/api/endpoints/createPOIsEndpoint.ts
steffenkleinle, line 1: 🔧 This file should be deleted

shared/utils/pois.ts
steffenkleinle, line 1: 🔧 This file should be deleted

steffenkleinle (commented):
Thanks for your work, this looks like a lot of effort 🙈 💚

package.json
andrew8er, line 27: We should run all `postinstall` scripts for this workspace and build the TS sources:

```suggestion
    "postinstall": "yarn exec tsc --build && yarn workspaces foreach --all run postinstall"
```

shared/api/endpoints/testing/EventModelBuilder.ts
lunars97: yes, there was an error saying `SyntaxError: Named export 'RRule' not found. The requested module 'rrule' is a CommonJS module, which may not support all module.exports as named exports.` There is a workaround for it https://github.com/jkbrzt/rrule/issues/478

native/src/components/Highlighter.tsx
lunars97: yes, there was `ESLint: Unexpected any value in conditional.` error

andrew8er (changes requested):
So there are two ways to use code from different packages:
- Just directly import sources with `import { foo } from '../../build-configs/module'
- Link packages together with the package manager, build all packages, import with `import { foo } from `build-config/module'

Each has some pros and cons, this project uses the latter. But for this two work, there are a few things missing:

### Problems

- When running `tsc --build` from root, I get this error:
  ```
   TS2353: Object literal may only specify known properties, and ''appium:autoGrantPermissions'' does not exist in type 'Capabilities'.
   ```
   Fixed this by commenting out the offending line in `/e2e-tests/native/wdio.conf.ts`.
- The TS sources need to be built, for convenience, could be done in the `postinstall` script. `.yarnrc.yml` also needs this to enable postinstall scripts:
  ```
  enableScripts: true
  ```
- Running tests, I get this error (probably unrelated to this PR?):
  ```
    FAIL  native/src/components/__tests__/MalteHelpFormOffer.spec.tsx (20.331 s)
  ● MalteHelpFormOffer › should show an error if there is another error while sending to Zammad
   ```

lunars97:
> So there are two ways to use code from different packages:
> 
> * Just directly import sources with `import { foo } from '../../build-configs/module'
> * Link packages together with the package manager, build all packages, import with `import { foo } from `build-config/module'
> 
> Each has some pros and cons, this project uses the latter. But for this two work, there are a few things missing:
> 
> ### Problems
> * When running `tsc --build` from root, I get this error:
>   ```
>    TS2353: Object literal may only specify known properties, and ''appium:autoGrantPermissions'' does not exist in type 'Capabilities'.
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   Fixed this by commenting out the offending line in `/e2e-tests/native/wdio.conf.ts`.
> * The TS sources need to be built, for convenience, could be done in the `postinstall` script. `.yarnrc.yml` also needs this to enable postinstall scripts:
>   ```
>   enableScripts: true
>   ```
> * Running tests, I get this error (probably unrelated to this PR?):
>   ```
>FAIL  native/src/components/__tests__/MalteHelpFormOffer.spec.tsx (20.331 s)
>   ● MalteHelpFormOffer › should show an error if there is another error while sending to Zammad
>   ```

Thank you a lot for your review and clarification! :green_heart: yes, the latter one is about importing like a real package, and the first one does not treat them as a real packages thus direct access.

**About the problems you found** 

1. `TS2353: Object literal may only specify known properties, and ''appium:autoGrantPermissions'' does not exist in type 'Capabilities'.` Hmm I was not able to reproduce it, I ran `tsc --build` in root and in e2e-tests, no error at all.
2. `enableScripts: true` isn't it enough just to build them when we run `yarn ts:check` ? :thinking: 
3.  `FAIL  native/src/components/__tests__/MalteHelpFormOffer.spec.tsx (20.331 s)
● MalteHelpFormOffer › should show an error if there is another error while sending to Zammad`
Yes, this error is unrelated to it, when you run it in native, there is no fail, but it fails in root

package.json
lunars97: I tried it and seems like it is not working when i run `yarn`, I also set enableScripts to true, nothing happens. There is an open issue for that https://github.com/yarnpkg/yarn/issues/5476 and possible workaround https://github.com/yarnpkg/yarn/issues/5476#issuecomment-2647688927

andrew8er:
> TS2353: Object literal may only specify known properties, and ''appium:autoGrantPermissions'' does not exist in type 'Capabilities'. Hmm I was not able to reproduce it, I ran tsc --build in root and in e2e-tests, no error at all.

Mh, strange. I have:
- **tsc**: 5.9.3
- **@wdio/types**: 8.41.0 (the origin of this definition)

I fixed it "properly" with:
```ts
import { Capabilities } from '@wdio/types'

const androidCapabilities: WebdriverIO.Capabilities | Capabilities.AppiumAndroidCapabilities = {
  platformName: 'android',
  'appium:app': '../native/android/app/build/outputs/apk/debug/app-debug.apk',
  'appium:automationName': 'UiAutomator2',
  'appium:autoGrantPermissions': true,
}
```


> enableScripts: true isn't it enough just to build them when we run yarn ts:check ? 🤔

Ah yes, this is enough. I did not saw this script, because I expected `ts:check` to not actually emit code, though. Check for me implies the `--no-emit` flag.

Still, I'm a fan of making projects easy to use, and having to manually run a command (which is easily forgotten) creates friction, especially for newcomers/volunteers.

A more comprehensive solution would also start a background watcher that compiles every shared library in the code, because otherwise changes in those will not immediately be reflected in the web/native apps. I can add this when working on #4166.


> FAIL  native/src/components/__tests__/MalteHelpFormOffer.spec.tsx (20.331 s) ● MalteHelpFormOffer › should show an error if there is another error while sending to Zammad
> Yes, this error is unrelated to it, when you run it in native, there is no fail, but it fails in root

Ok, we should fix the test at some point, but this should not block this PR.

lunars97:
> > TS2353: Object literal may only specify known properties, and ''appium:autoGrantPermissions'' does not exist in type 'Capabilities'. Hmm I was not able to reproduce it, I ran tsc --build in root and in e2e-tests, no error at all.
> 
> Mh, strange. I have:
> 
> * **tsc**: 5.9.3
> * **@wdio/types**: 8.41.0 (the origin of this definition)
> 
> I fixed it "properly" with:
> 
> ```ts
> import { Capabilities } from '@wdio/types'
> 
> const androidCapabilities: WebdriverIO.Capabilities | Capabilities.AppiumAndroidCapabilities = {
>   platformName: 'android',
>   'appium:app': '../native/android/app/build/outputs/apk/debug/app-debug.apk',
>   'appium:automationName': 'UiAutomator2',
>   'appium:autoGrantPermissions': true,
> }
> ```
> 
> > enableScripts: true isn't it enough just to build them when we run yarn ts:check ? 🤔
> 
> Ah yes, this is enough. I did not saw this script, because I expected `ts:check` to not actually emit code, though. Check for me implies the `--no-emit` flag.
> 
> Still, I'm a fan of making projects easy to use, and having to manually run a command (which is easily forgotten) creates friction, especially for newcomers/volunteers.
> 
> A more comprehensive solution would also start a background watcher that compiles every shared library in the code, because otherwise changes in those will not immediately be reflected in the web/native apps. I can add this when working on #4166.
> 
> > FAIL  native/src/components/**tests**/MalteHelpFormOffer.spec.tsx (20.331 s) ● MalteHelpFormOffer › should show an error if there is another error while sending to Zammad
> > Yes, this error is unrelated to it, when you run it in native, there is no fail, but it fails in root
> 
> Ok, we should fix the test at some point, but this should not block this PR.

I also have **tsc** 5.9.3, but we have * **@wdio/types**: 9.27.0 in web, native and e2e-tests. Could this be an incompatability issue? :thinking: You have a valid point about reducing extra step. There is a workaround for postinstall https://github.com/yarnpkg/yarn/issues/5476#issuecomment-2647688927 I am not sure which option to implement would be better :thinking: I will open an issue about the failing test.

andrew8er (approved):
> I also have tsc 5.9.3, but we have * @wdio/types: 9.27.0 in web, native and e2e-tests. Could this be an incompatability issue? 

Indeed, I had a `node_modules` directory in `/e2e-tests`, probably because I once installed the project with npm. Apparently, yarn does not clean up stale `node_modules` directories and hoists all packages to the install root (which is a really bad idea, IMHO), so I had a mix of packages installed.

> 🤔 You have a valid point about reducing extra step. There is a workaround for postinstall https://github.com/yarnpkg/yarn/issues/5476#issuecomment-2647688927 I am not sure which option to implement would be better 🤔 I will open an issue about the failing test.

I get why they do not support all the lifecycle scripts, but if we compile and watch all changes to the libraries in parallel with running the bundlers for web and native, we don't need it anyway. I'll add that when working on #4166. So I'd say, from my side, all issues are resolved 👍

shared/utils/rrule.ts
```diff
@@ -0,0 +1,5 @@
+// Workaround for CJS - ESM interop
+// Fixes SyntaxError: The requested module 'rrule' does not provide an export named 'rrulestr'
+import * as rrule from 'rrule'
+
+export const { RRule, rrulestr } = rrule
```
steffenkleinle: ❓ I can't reproduce any problem without this actually. I ran all tests, checks and tested web and android without reproducing this when replacing these usages with the normal imports. Perhaps you could check again if this is actually necessary.

web/src/components/Header.tsx
```diff
@@ -11,6 +11,18 @@ import useUpdateDimensions from '../hooks/useUpdateDimensions'
...
+const Headroom = StickyHeadroom as unknown as React.ComponentClass<{
+  children: React.ReactNode
+  scrollHeight: number
+  height: number
+  onStickyTopChanged?: (stickyTop: number) => void
+  positionStickyDisabled?: boolean
+  parent?: HTMLElement | null
+  zIndex?: number
+  className?: string
+}>
```
steffenkleinle, line 24: ❓ This could probably fixed directly in the react-sticky-headroom repo as well, right? Would you mind opening an issue there to add support for esm there?

steffenkleinle (commented):
Really nice work! Just two questions to the workarounds here :) Thank you so much for tackling this!

web/src/components/Header.tsx
lunars97: oh yes, I also wanted to ask whether to open an issue for that. I will leave it like this for now till it will be updated with esm support

shared/utils/rrule.ts
lunars97: Yes, you right. Good catch! Not sure what really fixed this error during a back and forth implementation :thinking:

web/src/components/Header.tsx
lunars97: Done https://github.com/digitalfabrik/react-sticky-headroom/issues/87 https://github.com/digitalfabrik/integreat-app/issues/4173

steffenkleinle (approved):
Let's go 🚀

PR #4098, opened 2026-05-05
4037: [Android] Fix bottomSheet's disappearance

### LINKED ISSUE #4037: Bottom Sheet disappears sometimes

bahaaTuffaha:
### Short Description

Bottom Sheet sometimes disappears from the view all of a sudden and it is not that consistent.
(To reproduce this issue you should wait for the location permission 3 seconds before selecting an action.)

### Proposed Changes

- This is a workaround to remount the bottomSheet when the UI thread stops when it gets interrupted by the permission dialog.
- Added a `useEffect` that runs once and subscribes to `AppState` changes. When the app returns to `'active'` (e.g. after the permission dialog closes), it increments `remountKey`.
- Assigned `key={remountKey}` to `<StyledBottomSheet>` so it remounts the bottomSheet again when its incremented.

### Side Effects

- The AppState could re-render the sheet when the app is in the background.

### Testing

- Make sure that the app is deleted from the emulator.
- Launch integreat on android and go to Poi/map.
- Wait for the location permission dialog around 3 second before selecting deny or "only while using the app".
- just to make sure repeat the test multiple time with different waiting time.
- Test iOS.

### Resolved Issues

Fixes: #4037

---

### DIFF
--- a/native/src/components/PoisBottomSheet.tsx
+++ b/native/src/components/PoisBottomSheet.tsx
@@ -91,6 +91,7 @@ const PoisBottomSheet = ({
 
   // Workaround for bottomSheet gets hidden after permissions dialog on Android so we force remounting after app comes back to foreground.
   // reanimated's shared values gets affected by the permissions dialog (UI thread is stopped when the permission dialog is displayed).
+  // For more details https://github.com/digitalfabrik/integreat-app/issues/4037 and https://github.com/gorhom/react-native-bottom-sheet/issues/1791#issuecomment-2060957019
   useAppStateListener(nextAppState => {
     if (nextAppState === 'active') {
       setRemountKey(previous => previous + 1)
--- a/native/src/components/PoisBottomSheet.tsx
+++ b/native/src/components/PoisBottomSheet.tsx
@@ -3,9 +3,9 @@ import BottomSheet, {
   BottomSheetFlatListMethods,
   BottomSheetScrollView,
 } from '@gorhom/bottom-sheet'
-import React, { memo, ReactElement, Ref, useCallback, useRef } from 'react'
+import React, { memo, ReactElement, Ref, useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { NativeScrollEvent, NativeSyntheticEvent, View } from 'react-native'
+import { AppState, AppStateStatus, NativeScrollEvent, NativeSyntheticEvent, View } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 
 import { LocationType } from 'shared'
@@ -74,6 +74,7 @@ const PoisBottomSheet = ({
   const { t } = useTranslation('pois')
   const theme = useTheme()
   const bottomSheetRef = useRef<BottomSheet>(null)
+  const [remountKey, setRemountKey] = useState(0)
 
   const handlePoiFocus = useCallback(() => {
     if (!isFullscreen && bottomSheetRef.current) {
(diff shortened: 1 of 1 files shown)


native/src/components/PoisBottomSheet.tsx
```diff
@@ -87,6 +88,18 @@ const PoisBottomSheet = ({
...
+  // reanimated's shared values gets affected by the permissions dialog (UI thread is stopped when the permission dialog is displayed).
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'active') {
+        setRemountKey(prev => prev + 1)
+      }
+    }
+    const subscription = AppState.addEventListener('change', handleAppStateChange)
+    return () => subscription.remove()
+  }, [])
```
steffenkleinle: 🔧 We have a hook for this: `useAppStateListener`

native/src/components/PoisBottomSheet.tsx
```diff
@@ -87,6 +88,18 @@ const PoisBottomSheet = ({
     bottomSheetRef.current?.snapToIndex(1)
   }
 
+  // Workaround for bottomSheet gets hidden after permissions dialog on Android so we force remounting after app comes back to foreground.
+  // reanimated's shared values gets affected by the permissions dialog (UI thread is stopped when the permission dialog is displayed).
```
steffenkleinle, line 93: ❓ Nice find! Do you have an issue you could link here/when we could remove this workaround again?

steffenkleinle (approved):
Very nice! Tested on android and iOS :) Could reproduce the issue and verify that with your fix it did not occur anymore.

native/src/components/PoisBottomSheet.tsx
```diff
@@ -87,6 +88,18 @@ const PoisBottomSheet = ({
...
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'active') {
+        setRemountKey(prev => prev + 1)
```
steffenkleinle: 🔧 
```suggestion
        setRemountKey(previous => previous + 1)
```

native/src/components/PoisBottomSheet.tsx
bahaaTuffaha: someone talked about similar issue:
https://github.com/gorhom/react-native-bottom-sheet/issues/1791#issuecomment-2060957019

bahaaTuffaha: hmm I can create a new ticket about removing this workaround and with tag blocked until a proper fix.

steffenkleinle: Not necessary to create a ticket imo, but I'd add a comment here with a link to the issue.

steffenkleinle (approved):

f1sh1918 (approved):
Nice find. works well 👍

PR #4117, opened 2026-05-13
4113: Fix position scroll for BottomSheetFlatList

### LINKED ISSUE #4113: Persist scroll position for pois
### Describe the Problem

We previously had functionality to restore the scroll position in the pois list after selecting and deselecting a poi.
This was not working anymore and has therefore been removed with the refactoring to a local stack history in #4073 (54d4e2b61cc75dda815ca24546fa32f0c6c7f843).

Reimplement the persisting and restoring of the list scroll position. It seems like there are problems with the different snap points of the bottom sheet.
In theory it would be enough to just save the list scroll position in the `PoisBottomSheet`:
``` ts
const [listScrollPosition, setListScrollPosition] = useState(0)

<BottomSheetFlatList
  contentOffset={{ y: listScrollPosition, x: 0 }}
  onMomentumScrollBegin={(event: NativeSyntheticEvent<NativeScrollEvent>) =>
    setListScrollPosition(event.nativeEvent.contentOffset.y)
  }
  ...
/>
```

N/A

### Related Issues

#3985

### Additional Information

N/A

bahaaTuffaha:
### Short Description

We previously had functionality to restore the scroll position in the pois list after selecting and deselecting a poi.
This was not working anymore and has therefore been removed with the refactoring to a local stack history in https://github.com/digitalfabrik/integreat-app/pull/4073 (https://github.com/digitalfabrik/integreat-app/commit/54d4e2b61cc75dda815ca24546fa32f0c6c7f843).

### Proposed Changes

- I went for a cleaner approach with less refs and useEffects so I moved the `PoiDetails` view to be on top of the `flatList` instead of conditional that re-renders the list.
- Wrapped `BottomSheetScrollView` with view `PoiDetailsOverlay` that has absolute positioning.
- Disabled accessibility on the list while the poiDetails is shown.

### Side Effects

- none.

### Testing

- Launch the native app and go to map/poi.
- Scroll down in the list to any poi item and press it.
- The list should have the same scroll position when you go back to the list.
- Test on iOS.

### Resolved Issues

Fixes: #4113

---


steffenkleinle (approved):
Very nice solution! Can't see any downsides atm :) Tested on android and iOS.

💭 Minor/unrelated: Since you need to open the bottom sheet modal to fullscreen to be able to scroll and clicking on a poi reduces the bottom sheet to the middle snappoint again, the scroll still seems off if the opened poi is in the bottom half of the visible list.
Two possible solutions I could think of:
a) Scroll to the selected poi on press, such that the selected poi is on top of the bottom sheet even at mid snappoint
b) Restore the bottom sheet size to fullscreen when navigating back to the list if it was fullscreen before
I haven't checked how easy it is to implement the two options and if there are any downsides, but perhaps you are motivated to investigate that a bit? If not, the current solution is good enough imo.

native/src/components/PoisBottomSheet.tsx
```diff
@@ -122,28 +129,30 @@ const PoisBottomSheet = ({
...
-                style={{
-                  alignSelf: 'center',
-                  marginTop: 20,
-                }}>
-                {t('noPois')}
-              </Text>
-            }
-            ItemSeparatorComponent={PoiListDivider}
-          />
+        <BottomSheetFlatList
```
f1sh1918, line 151: 🙃 i think the solution may fail with background/foreground transition, since you fixed that bottom sheet disappearing with the remount key in df94400f0
This remount key implementation may kill this solution when a user is on detail list then app in the background and app in the foreground -> everything will be remounted and scroll position lost.

I think this i really a minor case but i wanted to mention that
I even dunno a better solution

f1sh1918:
> Very nice solution! Can't see any downsides atm :) Tested on android and iOS.
> 
> 💭 Minor/unrelated: Since you need to open the bottom sheet modal to fullscreen to be able to scroll and clicking on a poi reduces the bottom sheet to the middle snappoint again, the scroll still seems off if the opened poi is in the bottom half of the visible list. Two possible solutions I could think of: a) Scroll to the selected poi on press, such that the selected poi is on top of the bottom sheet even at mid snappoint b) Restore the bottom sheet size to fullscreen when navigating back to the list if it was fullscreen before I haven't checked how easy it is to implement the two options and if there are any downsides, but perhaps you are motivated to investigate that a bit? If not, the current solution is good enough imo.

i vote for solution a) if its not too complicated

f1sh1918 (approved):
Tested on android & ios. Nice work 👍 
Just added one concern.

native/src/components/PoisBottomSheet.tsx
bahaaTuffaha: I test it again it only switch to 1 when this happens so I just set to work one time (set it to 1)

bahaaTuffaha:
> > Very nice solution! Can't see any downsides atm :) Tested on android and iOS.
> > 💭 Minor/unrelated: Since you need to open the bottom sheet modal to fullscreen to be able to scroll and clicking on a poi reduces the bottom sheet to the middle snappoint again, the scroll still seems off if the opened poi is in the bottom half of the visible list. Two possible solutions I could think of: a) Scroll to the selected poi on press, such that the selected poi is on top of the bottom sheet even at mid snappoint b) Restore the bottom sheet size to fullscreen when navigating back to the list if it was fullscreen before I haven't checked how easy it is to implement the two options and if there are any downsides, but perhaps you are motivated to investigate that a bit? If not, the current solution is good enough imo.
> 
> i vote for solution a) if its not too complicated

Done.
I added a ref called flatListRef that will scroll the list.

native/src/components/PoisBottomSheet.tsx
```diff
@@ -83,6 +95,10 @@ const PoisBottomSheet = ({
   const handlePoiSelection = (poi: PoiModel) => {
     selectPoi(poi)
     bottomSheetRef.current?.snapToIndex(1)
+    const index = pois.findIndex(it => it.path === poi.path)
+    if (index !== -1) {
+      flatListRef.current?.scrollToIndex({ index, animated: false })
```
steffenkleinle: 🤔 Should we do `- 0.5` so it is clear that the list is still scrollable? Not sure about that though.
```suggestion
      flatListRef.current?.scrollToIndex({ index: index - 0.5, animated: false })
```

steffenkleinle (approved):
Perfect, nice!

native/src/components/PoisBottomSheet.tsx
bahaaTuffaha: Good idea 👍

PR #4119, opened 2026-05-14
4092: Fix poiDetails keyboard accessibility

### LINKED ISSUE #4092: POI details in bottom sheet not accessable
### Describe the Bug

A user with an external keyboard kann not access the poi bottom sheet details 

### Steps to Reproduce

1. Go to map
2. navigate with tab and try to access poi details in bottom sheet
3. See error

### Expected Behavior

A user can access the bottom sheet via external keyboard

### Environment

- Device/Browser:  iOS and android
- OS version: all recent versions
- App version: 2026.4.4

### Related Issues

#3487

bahaaTuffaha:
### Short Description
A user with an external keyboard can't access the poi details at bottomSheet.

⚠️  This is a sub-branch of 4113

### Proposed Changes

- Because the poiDetails is above the list I had to disable the focus when I press tab on the poiDetails so I added a prop to  `PoiListItem` called `isKeyboardFocusable` to disable accessible if there is there is no slug.

- Removed `rippleColor='transparent'` from Accordion because it affects the keyboard focus highlight.

- At PoiDetailsContainer I changed `focusable` to `screenReaderFocusable` to prevent it from focusing on non pressable content and keep it just for the readers.
- There was an issue with Accordion and the location copy button that the focus gets stuck in a loop between them so I tried many solution to solve that but non worked other than an experimental accessibility prop called `experimental_accessibilityOrder` and gave it an array of strings that represents the current order.
https://reactnative.dev/docs/accessibility#experimental_accessibilityorder
- Added a prop `nativeID` to the Accordion and wrapped it with a view just for nativeID.
### Side Effects

- None (other than that experimental prop I used).

### Testing

- Test PoiDetails on Android and keep pressing tab to navigate.
- Test the talkback functionality.
- Test iOS.

### Resolved Issues

Fixes: #4092

---

### DIFF
--- a/native/src/components/PoiDetails.tsx
+++ b/native/src/components/PoiDetails.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
+import { Platform } from 'react-native'
 import { Divider } from 'react-native-paper'
 import styled from 'styled-components/native'
 
@@ -39,22 +40,12 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
   const { t } = useTranslation('pois')
   const { title, content, contacts, openingHours, temporarilyClosed, isCurrentlyOpen, category, appointmentUrl } = poi
 
-  // [4092] accessibilityOrder feature is experimental and could change in the future
-  // Used here to fix the the loop that happens between the description accordion and the address copy button.
-  // https://reactnative.dev/docs/accessibility#experimental_accessibilityorder
-  const accessibilityOrder = [
-    ...(content.length > 0 ? ['accessibility-order-description'] : []),
-    'accessibility-order-address',
-    ...(contacts.length > 0 ? ['accessibility-order-contacts'] : []),
-    'accessibility-order-opening-hours',
-  ]
-
   return (
     <PoiDetailsContainer
       accessibilityLabel={`${title} - ${category.name}`}
       onFocus={onFocus}
       screenReaderFocusable
-      experimental_accessibilityOrder={accessibilityOrder}>
+      focusable={Platform.OS === 'ios' ? true : undefined}>
       <Text variant='h5' style={{ paddingBottom: 4 }}>
         {title}
       </Text>
@@ -68,7 +59,7 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
       <StyledDivider />
       {content.length > 0 && (
         <>
-          <Accordion nativeID='accessibility-order-description' headerContent={t('description')}>
+          <Accordion headerContent={t('description')}>
             <Page content={content} language={language} padding={false} />
           </Accordion>
           <StyledDivider />
@@ -78,7 +69,7 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
       <StyledDivider />
       {contacts.length > 0 && (
         <>
-          <Accordion nativeID='accessibility-order-contacts' headerContent={t('contacts')} initialCollapsed>
+          <Accordion headerContent={t('contacts')} initialCollapsed>
             <StyledContactsContainer>
               {contacts.map((contact, index) => (
                 <Contact
--- a/native/src/components/PoiDetails.tsx
+++ b/native/src/components/PoiDetails.tsx
@@ -14,7 +14,7 @@ import Page from './Page'
 import PoiChips from './PoiChips'
 import Text from './base/Text'
 
-const PoiDetailsContainer = styled.View`
+const PoiDetailsContainer = styled.View<{ experimental_accessibilityOrder?: string[] }>`
   flex: 1;
   background-color: ${props => props.theme.colors.background};
   gap: 16px;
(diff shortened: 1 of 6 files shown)


native/src/components/PoisBottomSheet.tsx
```diff
@@ -111,6 +111,7 @@ const PoisBottomSheet = ({
       navigateToPoi={() => handlePoiSelection(poi)}
       distance={userLocation && poi.distance(userLocation)}
       onFocus={expandFullscreen}
+      isKeyboardFocusable={!slug}
```
steffenkleinle: 🔧 I think this would be more descriptive what this is actually about. Same for other props.
```suggestion
      visible={!slug}
```

native/src/components/PoiDetails.tsx
```diff
@@ -39,8 +39,22 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
...
+    ...(contacts.length > 0 ? ['accessibility-order-contacts'] : []),
+    'accessibility-order-opening-hours',
+  ]
+
   return (
-    <PoiDetailsContainer accessibilityLabel={`${title} - ${category.name}`} onFocus={onFocus} focusable>
+    <PoiDetailsContainer
+      accessibilityLabel={`${title} - ${category.name}`}
+      onFocus={onFocus}
+      screenReaderFocusable
```
steffenkleinle: 🤔 Why did you switch from `focusable` to `screenReaderFocusable`?

native/src/components/PoiDetails.tsx
```diff
@@ -39,8 +39,22 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
...
+  // [4092] accessibilityOrder feature is experimental and could change in the future
+  // Used here to fix the the loop that happens between the description accordion and the address copy button.
+  // https://reactnative.dev/docs/accessibility#experimental_accessibilityorder
+  const accessibilityOrder = [
+    ...(content.length > 0 ? ['accessibility-order-description'] : []),
+    'accessibility-order-address',
+    ...(contacts.length > 0 ? ['accessibility-order-contacts'] : []),
+    'accessibility-order-opening-hours',
+  ]
+
```
steffenkleinle: ❓ I can't reproduce an endless loop here, but maybe I also just don't know how to test. Perhaps you can share a video?

steffenkleinle (commented):
Tested on Android only, looks good so far, just curious about the endless loop.

native/src/components/PoiDetails.tsx
bahaaTuffaha: > At PoiDetailsContainer I changed focusable to screenReaderFocusable to prevent it from focusing on non pressable content and keep it just for the readers.

bahaaTuffaha: if I removed both... the text in the description will not be selected by the reader.

native/src/components/PoiDetails.tsx
bahaaTuffaha: https://github.com/user-attachments/assets/d4987ec8-3758-44e5-b59c-b17003a03975

native/src/components/PoiDetails.tsx
steffenkleinle: Well... I guess I really have to be more careful of reading the PR descriptions, sorry! Thanks for your patience 💚

native/src/components/PoiDetails.tsx
bahaaTuffaha: hmm what I found is I had to remove all of the nativeIDs to reproduce it. I think we can remove the `experimental_accessibilityOrder` and keep some IDs just to fix this problem.

native/src/components/PoiDetails.tsx
steffenkleinle: And for iOS this is not necessary (as `screenReaderFocusable` is an Android-only prop)?

bahaaTuffaha: > And for iOS this is not necessary (as `screenReaderFocusable` is an Android-only prop)?

Thanks for pointing out ... I will keep focusable true if platform is iOS

native/src/components/PoiDetails.tsx
```diff
@@ -14,7 +15,7 @@ import Page from './Page'
 import PoiChips from './PoiChips'
 import Text from './base/Text'
 
-const PoiDetailsContainer = styled.View`
+const PoiDetailsContainer = styled.View<{ experimental_accessibilityOrder?: string[] }>`
```
steffenkleinle: 🔧 Please revert

steffenkleinle (approved):

native/src/components/PoiDetails.tsx
```diff
@@ -40,7 +41,11 @@ const PoiDetails = ({ poi, language, distance, onFocus }: PoiDetailsProps): Reac
   const { title, content, contacts, openingHours, temporarilyClosed, isCurrentlyOpen, category, appointmentUrl } = poi
 
   return (
-    <PoiDetailsContainer accessibilityLabel={`${title} - ${category.name}`} onFocus={onFocus} focusable>
+    <PoiDetailsContainer
+      accessibilityLabel={`${title} - ${category.name}`}
+      onFocus={onFocus}
+      screenReaderFocusable
+      focusable={Platform.OS === 'ios' ? true : undefined}>
```
f1sh1918: ❓ why do we need this?
`focusable` is an android only property. So for ios there should nothing happen?

https://reactnative.dev/docs/view#focusable-android

f1sh1918 (approved):
Tested on android and ios. Works well 👍

native/src/components/PoiDetails.tsx
bahaaTuffaha: Screenshot 2026-05-18 at 4 47 36 PM


I thought it would work for ios too... then I will remove it 👍

PR #4120, opened 2026-05-15
4056: [Proposal] Fix recentering

steffenkleinle:
### Short Description

Proposal for #4074 to avoid unselecting the poi on track user location.
Also removes duplicated location state and fixes the initial user location.

### Proposed Changes

- Prevent centering on the currently selected feature when minimizing the bottom sheet 
- Use initial camera settings
- Move user location hook up to `Pois`
- Keep previous user location on reload
- Rename `coordinates` to `userLocation`

### Side Effects

N/A

### Testing

Test interaction between following user location, (un-)selecting features (list and map) and expanding/collapsing the bottom sheet.

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/native/src/components/MapView.tsx
+++ b/native/src/components/MapView.tsx
@@ -85,6 +85,7 @@ type MapViewProps = {
   refreshPermissionAndLocation: () => Promise<LocationStateType | null>
   selectFeature: (feature: MapFeature | null) => void
   bottomSheetHeight: number
+  bottomSheetMidHeight: number
   bottomSheetFullscreen: boolean
   zoom: number | undefined
   Overlay?: ReactElement
@@ -99,6 +100,7 @@ const MapView = ({
   selectFeature,
   Overlay,
   bottomSheetHeight,
+  bottomSheetMidHeight,
   bottomSheetFullscreen,
   zoom,
 }: MapViewProps): ReactElement => {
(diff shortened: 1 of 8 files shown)


bahaaTuffaha (commented):
Thank you for the help!
I noticed some differences in the behavior for example:
- expanding/minimizing the bottomSheet doesn't recenter the camera.

https://github.com/user-attachments/assets/6199af1d-9716-4b95-85d1-59c2230c0c94

steffenkleinle:
> Thank you for the help! I noticed some differences in the behavior for example:
> 
>     * expanding/minimizing the bottomSheet doesn't recenter the camera.
> 
> 
> 2026-05-15.16-00-46.mov

I thought thats what you wanted to avoid in https://github.com/digitalfabrik/integreat-app/pull/4074#discussion_r3193362929?

bahaaTuffaha:
> I thought thats what you wanted to avoid in [#4074 (comment)](https://github.com/digitalfabrik/integreat-app/pull/4074#discussion_r3193362929)?

The link doesn't navigate to the comment but in any case we need that re-centring behaviour.
I think the issue was from the userlocation gets disabled/un-toggled when I minimize the bottomSheet.

steffenkleinle:
> > I thought thats what you wanted to avoid in [#4074 (comment)](https://github.com/digitalfabrik/integreat-app/pull/4074#discussion_r3193362929)?
> 
> The link doesn't navigate to the comment but in any case we need that re-centring behaviour. I think the issue was from the userlocation gets disabled/un-toggled when I minimize the bottomSheet.

This was your comment that no recentering should happen if changing the bottom sheet height:

> Without it will cause problems with previous selected poi:
>
> https://github.com/user-attachments/assets/ef6511aa-a727-42df-a0d8-31e79529be78

Anyway, I fixed it now in a way that only recenters if expanding the bottom sheet, not when minimizing.

bahaaTuffaha (approved):
LGTM! ✅

PR #4121, opened 2026-05-18
2691: Add caching for web requests

### LINKED ISSUE #2691: Caching API results in web
As discussed in https://github.com/digitalfabrik/integreat-app/pull/2630#discussion_r1495953995, it would make sense to cache some results of API calls.

I would like a caching layer that we can use to cache data for a while. We have to discuss what a good amount of time for caching would be to balance the fastest user experience with the most up-to-date results. 

We could leave it the way it is.

steffenkleinle:
### Short Description

Add caching for web requests to reduce loading times and network traffic.

### Proposed Changes

- Add `useQueryFromEndpoint` hook and replace `useLoadFromEndpoint` usages in web
- Use `@tanstack/react-query` for caching
- Keep native as it is

### Side Effects

N/A

### Testing

Check web and all changed endpoints. Verify that caching works and loading times are reduced.

### Resolved Issues

Fixes: #2691

---

### DIFF
--- a/web/src/components/ChatController.tsx
+++ b/web/src/components/ChatController.tsx
@@ -6,29 +6,21 @@ import { cmsApiBaseUrl } from '../constants/urls'
 import useIsTabActive from '../hooks/useIsTabActive'
 import useLocalStorage from '../hooks/useLocalStorage'
 import useQueryFromEndpoint from '../hooks/useQueryFromEndpoint'
-import { generateChatId } from '../utils/chat'
 import Chat from './Chat'
 
 type ChatControllerProps = {
   region: RegionModel
   languageCode: string
-  chatId: string | null
-  updateChatId: (newChatId: string | null) => void
+  chatId: string
 }
 
 const LOCAL_STORAGE_ITEM_CHAT_PRIVACY_POLICIES = 'Chat-Privacy-Policies'
 const DEFAULT_POLLING_INTERVAL = 15000
 const TYPING_POLLING_INTERVAL = 3000
 
-const ChatController = ({ region, languageCode, chatId, updateChatId }: ChatControllerProps): ReactElement => {
+const ChatController = ({ region, languageCode, chatId }: ChatControllerProps): ReactElement => {
   const [sendingError, setSendingError] = useState<Error | null>(null)
   const isBrowserTabActive = useIsTabActive()
-
-  const initializeChat = (): string => {
-    const newId = generateChatId()
-    updateChatId(newId)
-    return newId
-  }
 
   const {
     data,
@@ -42,7 +34,7 @@ const ChatController = ({ region, languageCode, chatId, updateChatId }: ChatCont
     {
       regionCode: region.code,
       language: languageCode,
-      deviceId: chatId ?? '',
+      deviceId: chatId,
     },
     { cached: false },
   )
@@ -69,7 +61,7 @@ const ChatController = ({ region, languageCode, chatId, updateChatId }: ChatCont
       regionCode: region.code,
       language: languageCode,
       message,
-      deviceId: chatId ?? initializeChat(),
+      deviceId: chatId,
     })
 
     if (data !== null) {
--- a/web/src/components/ChatController.tsx
+++ b/web/src/components/ChatController.tsx
@@ -1,17 +1,11 @@
 import React, { ReactElement, useEffect, useState } from 'react'
 
-import { SendingStatusType } from 'shared'
-import {
-  RegionModel,
-  createChatMessagesEndpoint,
-  createSendChatMessageEndpoint,
-  NotFoundError,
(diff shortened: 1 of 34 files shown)


web/src/components/SprungbrettOffer.tsx
```diff
@@ -15,14 +16,18 @@ type SprungbrettOfferPageProps = {
...
     return <SkeletonList />
   }
 
-  if (!data) {
-    return <FailureSwitcher error={error ?? new Error('Data missing')} />
+  if (error) {
+    return <FailureSwitcher error={error} />
   }
 
   const items = data.map(job => <SprungbrettListItem key={job.url} job={job} />)
```
bahaaTuffaha, line 33: ❓  could data return undefined here?

f1sh1918:
@steffenkleinle do you have any idea if that may affect the chat, since afaics all query are cached for 5 minutes (stale times)

web/src/components/ChatController.tsx
bahaaTuffaha, line 1: There is something wrong with the refetch:

https://github.com/user-attachments/assets/4539d0fc-096b-4fdc-acac-1e6c4d6ea7a4

bahaaTuffaha (commented):
Great work!

The caching work great as long as we don't refresh the page which an improvement to the search functionality.
- Search result are cached ✅ 
- Regions loaded faster when I get to landing page again ✅ 

💭  We can make it persistent but not sure if there is cons to this:
https://tanstack.com/query/v4/docs/framework/react/plugins/persistQueryClient

web/src/components/SprungbrettOffer.tsx
steffenkleinle: I don't think so, why do you ask?

steffenkleinle:
> The caching work great as long as we don't refresh the page which an improvement to the search functionality.
>
> 💭 We can make it persistent but not sure if there is cons to this: https://tanstack.com/query/v4/docs/framework/react/plugins/persistQueryClient

Personally, I'd say the current behavior is what we want, i.e. cached requests inside one session, but redo the requests on a browser refresh.

web/src/hooks/useQueryFromEndpoint.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return {
+    ...useQuery({
+      queryKey,
+      queryFn: async () => loadFromEndpoint(createEndpoint, baseUrl, params),
+      staleTime: cached ? defaultStaleTime : 0,
+    }),
+    setData: data => queryClient.setQueryData(queryKey, data),
+  }
+}
```
bahaaTuffaha, line 32: refetchOnWindowFocus is enabled by default
https://tanstack.com/query/latest/docs/framework/react/guides/window-focus-refetching

❓ could it affect caching?

web/src/components/SprungbrettOffer.tsx
bahaaTuffaha: because the the typing of data says `SprungbrettModel[] | undefined`
Screenshot 2026-05-20 at 1 54 05 PM

but didn't complaint maybe because we check for error before it...

web/src/components/ChatController.tsx
bahaaTuffaha, line 1: ❌  Chat still not showing the first message.
when `chatId = null` the queryKey is empty string. 🤔

web/src/components/SprungbrettOffer.tsx
steffenkleinle: Yep, so nothing to be done here?

bahaaTuffaha (approved):

lunars97:
thank you for your PR :heart_hands: There is one issue I found: if you go to search and search for something and open the chat, it does not stay opened, instead it closes back immediately, it does not open with and without search results

steffenkleinle:
@lunars97 maybe its #4126 (already fixed in https://github.com/digitalfabrik/integreat-app/pull/4137)? Did you check if this also occurs on main?

lunars97:
> @lunars97 maybe its #4126 (already fixed in #4137)? Did you check if this also occurs on main?

oh sorry, yes, this bug is on main. So it is good to go then :+1:

lunars97:
> > @lunars97 maybe its #4126 (already fixed in #4137)? Did you check if this also occurs on main?
> 
> oh sorry, yes, this bug is on main. So it is good to go then 👍

I will open an issue for that

lunars97 (approved):

lunars97:
> > > @lunars97 maybe its #4126 (already fixed in #4137)? Did you check if this also occurs on main?
> > 
> > 
> > oh sorry, yes, this bug is on main. So it is good to go then 👍
> 
> I will open an issue for that

Oops seems like there is a fix for that :+1:

PR #4130, opened 2026-05-20
3944: Improve error handling for blocked regions

### LINKED ISSUE #3944: Improve error screen for IP-blocked regions
### Describe the Bug
When you select the region `Wickede - Wimbern` or other IP-blocked regions it shows "The request could not be processed." without a possibility to go back.

While this error is expected, it should be possible to go back to the region selection in that case.
~~Possibly, we also want the error message more descriptive.~~

### Steps to Reproduce

1. Launch Malte.
2. from landing page select `Wickede - Wimbern` the last item in the list.

### Expected Behavior

- It should be possible to go back/show the change location button at the top in case there is a failure. It might be quite difficult/impossible to implement the usual header go back button as we replace the route. But perhaps we can add a normal option somewhere on the error screen.

### Related Issues

N/A

### Additional Information

Image

steffenkleinle:
### Short Description

Improve error handling both for blocked regions and in general.

### Proposed Changes

- Show a explicit message if an endpoint returns status 403 forbidden
- Always show a go back button for error screens (native)
- Always (except for RemoteContent) show a retry button for error screens(native)
- Use the primary blue as outline color for outlined buttons (native)

### Side Effects

N/A

### Testing

See the issue description for the initial issue (web and native).
Check different possible error screens (e.g. disabling internet) and the retry/go back button (native).

### Resolved Issues

Fixes: #3944

---


lunars97:
Almost looks good :heart_eyes:  , just one remark: I think having a `Go back` button is a bit misleading, because a user cant go back when an internet is disabled, see the screenshot 
image

steffenkleinle:
> Almost looks good 😍 , just one remark: I think having a `Go back` button is a bit misleading, because a user cant go back when an internet is disabled, see the screenshot

Hmm, good point, but I actually thinks this is a valid usecase. Since our app has offline content, in some cases (e.g. switching a region without internet), the user can navigate back and recover a working app state.

lunars97:
> > Almost looks good 😍 , just one remark: I think having a `Go back` button is a bit misleading, because a user cant go back when an internet is disabled, see the screenshot
> 
> Hmm, good point, but I actually thinks this is a valid usecase. Since our app has offline content, in some cases (e.g. switching a region without internet), the user can navigate back and recover a working app state.

hmm then I think there is a bug, because `Go back` does not react at all on tap :eyes:

steffenkleinle:
> > > Almost looks good 😍 , just one remark: I think having a `Go back` button is a bit misleading, because a user cant go back when an internet is disabled, see the screenshot
> > 
> > 
> > Hmm, good point, but I actually thinks this is a valid usecase. Since our app has offline content, in some cases (e.g. switching a region without internet), the user can navigate back and recover a working app state.
> 
> hmm then I think there is a bug, because `Go back` does not react at all on tap 👀

Can you please give me more information on what do/how to reproduce this?

native/src/components/Failure.tsx
```diff
@@ -3,51 +3,61 @@ import { useTranslation } from 'react-i18next'
...
+      <Text>{t(code === ErrorCode.RegionUnavailable ? 'notFound.region' : code)}</Text>
+      {retry && (
+        <Button mode='contained' onPress={retry}>
+          {t('tryAgain')}
         </Button>
       )}
-    </ViewContainer>
+      <Button
+        mode='outlined'
+        buttonColor='primary'
```
f1sh1918: ❓ is the button color really needed? I think primary is default. At least i can't see any difference if i spare it out

lunars97:
> > > > Almost looks good 😍 , just one remark: I think having a `Go back` button is a bit misleading, because a user cant go back when an internet is disabled, see the screenshot
> > > 
> > > 
> > > Hmm, good point, but I actually thinks this is a valid usecase. Since our app has offline content, in some cases (e.g. switching a region without internet), the user can navigate back and recover a working app state.
> > 
> > 
> > hmm then I think there is a bug, because `Go back` does not react at all on tap 👀
> 
> Can you please give me more information on what do/how to reproduce this?

if you go to region selection and disable an internet and try to choose any region, then it will pop up as well as if you go to tunews without the internet, it will also show up...

f1sh1918 (approved):
Nicely done. Works fine
Tested on web and android

Only minor thing couldn't find a difference with or without setting `buttonColor`

steffenkleinle:
@lunars97 good catch, should be fixed now. Could you please check again?

native/src/components/Failure.tsx
```diff
@@ -1,53 +1,62 @@
...
   return (
-    <ViewContainer>
-      <Icon size={150} source={ErrorIcon} />
-      <Text style={{ margin: 12 }}>{t(code === ErrorCode.RegionUnavailable ? 'notFound.region' : code)}</Text>
-      {buttonAction && (
-        <Button mode='contained' onPress={buttonAction}>
-          {t(buttonLabel ?? 'tryAgain')}
+    <Container>
+      <Icon size={160} source={getErrorIcon(code)} />
+      <Text>{t(code === ErrorCode.RegionUnavailable ? 'notFound.region' : code)}</Text>
```
lunars97: what do you think about adding the `role='alert'` here

lunars97 (approved):
Thank you for your PR! LGTM, left a suggestion comment below

Docmale:
When I click the back button from the error screen, I don't return to my previous region but instead land on the "Set Region" screen. is this intendet?

steffenkleinle:
> When I click the back button from the error screen, I don't return to my previous region but instead land on the "Set Region" screen. is this intendet?

Good question. Not necessarily intended, but technically not trivially possible otherwise (as we'd also need to save the slug of the previous region), which we don't do at the moment. Furthermore, I think this behavior is actually consistent for the user, since the region selection was actually the previous page. Do you think it would be better UX otherwise?

Docmale:
I initially expected to end up in my original region, but after thinking about it (following your reply), I can no longer say which I prefer; since the user wanted to change regions, the current solution makes sense to me too. I'll close my test here.

PR #4135, opened 2026-05-21
4134: Improve and refactor useLocalStorage

### LINKED ISSUE #4134: Improve and refactor useLocalStorage
### Describe the Problem

Refactor useLocalStorage and its usages to make it simpler to use and more consistent.

- [x] Use array return value analog to useState for easier usage
- [x] Move storage keys to central place
- [x] Allow passing a function to update receiving the old value
- [x] Streamline naming

N/A

### Related Issues

Preparation for: #4129

### Additional Information

N/A

steffenkleinle:
### Short Description

Refactor useLocalStorage and its usages.

### Proposed Changes

- Use array return value analog to useState for easier usage
- Move storage keys to central place
- Allow passing a function to update receiving the old value
- Streamline naming of storage keys and (set) state vars

### Side Effects

N/A

### Testing

Check that everything storage related still works.

### Resolved Issues

Fixes: #4134

---

### DIFF
--- a/web/src/hooks/useLocalStorage.ts
+++ b/web/src/hooks/useLocalStorage.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react'
+import { useCallback, useState } from 'react'
 
 import { reportError } from '../utils/sentry'
 
@@ -16,7 +16,7 @@ type UseLocalStorageProps<T> = {
   isSessionStorage?: boolean
 }
 
-type UseLocalStorageReturn<T> = [value: T, updateLocalStorageItem: (newValue: T) => void]
+type UseLocalStorageReturn<T> = [value: T, updateLocalStorageItem: (newValue: T | ((oldValue: T) => T)) => void]
 
 const useLocalStorage = <T>({
   key,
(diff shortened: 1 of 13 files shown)


web/src/hooks/useLocalStorage.ts
```diff
@@ -38,24 +48,39 @@ const useLocalStorage = <T>({
...
+  useEffect(() => {
+    const syncValue = (event: Event) => {
+      const { key: changedKey, value: newValue } = (event as CustomEvent<StorageSyncEventDetail<T>>).detail
+      if (changedKey === key) {
+        setValue(newValue)
+      }
+    }
+    window.addEventListener(STORAGE_SYNC_EVENT, syncValue)
+    return () => window.removeEventListener(STORAGE_SYNC_EVENT, syncValue)
+  }, [key])
```
bahaaTuffaha: 🤔   what this useEffect does?

bahaaTuffaha (approved):
LGTM ✅  tested on Chrome.

web/src/hooks/useLocalStorage.ts
steffenkleinle: Listen to our custom event we send when the state changes. This syncs the state between multiple usages of the hook.

steffenkleinle: JFYI: I removed this functionality again, this breaks the concept of React's top-down data flow. The right approach here is to just pass state down.
If passing state is not possible or too cumbersome, the right approach would be to insert another layer (most likely in the form of a React context) that can then be used by the components instead directly using the `useLocalStorage` hook. However, this would be a little over-engineered atm.

bahaaTuffaha: Do you need another round of testing?

steffenkleinle: Don't think so, thanks. This only removes functionality and doesn't change/add existing.

lunars97 (approved):
Thanks a lot for refactoring! Found no issues 👍

web/src/hooks/useLocalStorage.ts
```diff
@@ -1,17 +1,22 @@
...
 import { reportError } from '../utils/sentry'
 
+export const CHAT_HINT_VISIBLE_STORAGE_KEY = 'showChatHint'
+export const CHAT_PRIVACY_POLICIES_STORAGE_KEY = 'Chat-Privacy-Policies'
+export const CHAT_ID_STORAGE_KEY = 'Chat-Device-Id'
+
+export const EXTERNAL_SOURCES_STORAGE_KEY = 'Opt-In-External-Sources'
+export const APP_BANNER_HIDDEN_EXPIRATION_DATE_STORAGE_KEY = 'showBanner'
+export const THEME_STORAGE_KEY = 'theme'
+
```
lunars97, line 12: shouldn't these keys be in a separate constant file? 💭

steffenkleinle: In this case I put it here on purpose to make it more visible when adding a new value to the local storage. If it is in some constant file, I think this might be missed.

PR #4141, opened 2026-05-25
3030: Share Integreat pages as QR-Code

### LINKED ISSUE #3030: Share Integreat pages as QR Code
### User Story

As a counselor I want to be able to quickly share a page link with a client to provide additional information for reading later.

### Details

Integreat already has a share function. However, it is difficult to share a page / link when the app users do not yet have each other in the contact list. A QR code which can easily be scanned by any other smart phone user can improve this. This could be achieved by adding a new dedicated "Share as QR Code function" to the share menu.

bahaaTuffaha:
### Short Description

As a counselor I want to be able to quickly share a page link with a client to provide additional information for reading later.

### Proposed Changes

For Web:
- Created a component called `QrCodeDialog` that uses mui components like `AlertDialog`
- Used the suggested package https://github.com/paulmillr/qr with svg encoding.
- `QrCodeDialog` has `description` and `qrDetails` as props to match the issue 3031 design.
- Added a new `MenuItem` for the QR-Code at HeaderMenu.tsx and inside the `sharingItems`.

For Native:
- Created a component called `QrCodeModal` that uses our Modal component.
- Used a popular package `react-qr-code` and increased error correction Level to H to match the QR code we have at web.
- Added a new `HeaderMenuItem` for the QR-Code at HeaderMenu.tsx.
- Added translations.

### Side Effects

None.

### Testing

- Go to any page and open the headerMenu.
- Click on the QR-Code button
- Test the QR code with your phone.
- Check contrast mode.
- Test Native.

### Resolved Issues

Fixes: #3030

---


native/package.json
```diff
@@ -93,6 +93,7 @@
     "react-native-svg": "^15.15.4",
     "react-native-url-polyfill": "^3.0.0",
     "react-native-worklets": "^0.8.1",
+    "react-qr-code": "^2.0.21",
```
steffenkleinle: ❓ Why do you choose to use different libraries? Doesn't https://github.com/paulmillr/qr also work on react-native since it has no dependencies? Did you try to use it?

web/src/components/QrCodeDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+import AlertDialog from './base/AlertDialog'
+import Svg from './base/Svg'
+
+type QrCodeDialogProps = {
+  open: boolean
+  close: () => void
+  title: string
+  description: string
+  url: string
+  qrDetails: string
```
steffenkleinle: ❓ Shouldn't the url and the qr details always be the same thing?
```suggestion
  content: string
```

web/src/components/QrCodeDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+      actions={
+        <DialogActions>
+          <Button onClick={close} variant='outlined'>
+            {t('common:close')}
+          </Button>
+        </DialogActions>
+      }>
+      <Stack alignItems='center' gap={2}>
+        <Typography variant='body2'>{description}</Typography>
+        <Box borderRadius={2} sx={{ backgroundColor: '#FFF' }}>
```
steffenkleinle: 🔧 Avoid inline styles and using hardcoded colors

web/jest.config.ts
```diff
@@ -3,7 +3,7 @@ import { JestConfigWithTsJest } from 'ts-jest'
 
 import { webIntegreatTestCmsBuildConfig } from 'build-configs/integreat-test-cms'
 
-const transformNodeModules = ['shared', 'build-configs', 'translations']
+const transformNodeModules = ['shared', 'build-configs', 'translations', 'qr']
```
steffenkleinle, line 6: 🤔 Why is this necessary?

web/src/components/HeaderMenu.tsx
```diff
@@ -45,6 +47,7 @@ const HeaderMenu = ({ children, pageTitle, fitScreen, ref }: HeaderMenuProps): R
   const [menuAnchorElement, setMenuAnchorElement] = React.useState<HTMLElement | null>(null)
   const [expandedAccordion, setExpandedAccordion] = React.useState<'share' | 'legal' | null>(null)
   const [urlCopied, setUrlCopied] = React.useState<boolean>(false)
+  const [qrDialogOpen, setQrDialogOpen] = React.useState<boolean>(false)
```
steffenkleinle: 🙃 
```suggestion
  const [qrShareOpen, setQrShareOpen] = React.useState<boolean>(false)
```

web/src/components/base/Svg.tsx
```diff
@@ -28,4 +28,4 @@ const Svg = ({
   />
 )
 
-export default Svg
+export default memo(Svg)
```
steffenkleinle, line 31: 🤔 Why was this necessary?

native/src/components/QrCodeModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+})
+
+type QrCodeModalProps = {
+  modalVisible: boolean
+  closeModal: () => void
+  title: string
+  description: string
+  url: string
+  qrDetails: string
```
steffenkleinle: 🤔 Same as for web:
```suggestion
  content: string
```

translations/translations.json
```diff
@@ -7815,7 +7815,10 @@
       "voiceUnavailable": "Keine Stimme installiert",
       "voiceUnavailableMessage": "Es wurde keine Stimme gefunden. Um die Vorlesefunktion nutzen zu können, muss eine Stimme installiert sein. Bitte beachten Sie die folgenden Anweisungen für Ihr Betriebssystem.",
       "openExternal": "Extern öffnen",
-      "skipToContent": "Zum Inhalt springen"
+      "skipToContent": "Zum Inhalt springen",
+      "shareQrCodeTitle": "Über QR-Code teilen",
+      "shareQrCodeDescription": "Teilen Sie diese Seite einfach, indem Sie den QR-Code auf einem anderen Gerät scannen.",
```
steffenkleinle: 🔧 
```suggestion
      "shareQrCodeDescription": "Teile diese Seite einfach, indem Sie den QR-Code auf einem anderen Gerät scannen.",
```

steffenkleinle (changes requested):
Nice work, works as expected!

native/package.json
bahaaTuffaha: Hmm let me try...

web/src/components/QrCodeDialog.tsx
bahaaTuffaha: because I'm considering the other ticket about Zammad with chat-id:
https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=5696-1735

I guess the url probably will have char id query 🤔

web/src/components/QrCodeDialog.tsx
bahaaTuffaha: I will add a color for white not sure what to call it ... `whiteBackground` ?

web/jest.config.ts
bahaaTuffaha: it causes errors for `yarn test` due to untranspiled code at node_modules you can read about it here:
https://jestjs.io/docs/configuration#transformignorepatterns-arraystring

web/src/components/base/Svg.tsx
bahaaTuffaha: It was flickering/re-renders when I scroll while the dialog is open.

build-configs/common/theme/colors.ts
```diff
@@ -40,6 +40,7 @@ export const commonLightColors: CommonColorPalette = {
     default: '#FFFFFF',
     paper: '#FFFFFF',
     accent: '#EAEEF9',
+    qrCode: '#FFFFFF',
```
steffenkleinle: ❌  Sorry, I think my comment was not clear enough. Instead of adding a new color (both inline or here), you should use the background color as background color and the text color as QR code color imo.

translations/translations.json
```diff
@@ -7733,7 +7733,10 @@
       "voiceUnavailable": "Keine Stimme installiert",
       "voiceUnavailableMessage": "Es wurde keine Stimme gefunden. Um die Vorlesefunktion nutzen zu können, muss eine Stimme installiert sein. Bitte beachten Sie die folgenden Anweisungen für Ihr Betriebssystem.",
       "openExternal": "Extern öffnen",
-      "skipToContent": "Zum Inhalt springen"
+      "skipToContent": "Zum Inhalt springen",
+      "shareQrCodeTitle": "Über QR-Code teilen",
+      "shareQrCodeDescription": "Teile diese Seite einfach, indem Sie den QR-Code auf einem anderen Gerät scannen.",
```
steffenkleinle: 🔧 Sorry, I missed the second `Sie`:
```suggestion
      "shareQrCodeDescription": "Teile diese Seite einfach, indem du den QR-Code auf einem anderen Gerät scannst.",
```

web/src/components/QrCodeDialog.tsx
steffenkleinle: Yes, in which case the chat id should imo be the content of both the QR code and the text below the QR code? What am I missing?

bahaaTuffaha: From the design it looks like just the id was passed to the content.

steffenkleinle: Ahh now I see. TBH I'd just omit the text `Chat-ID: ` and only display the actual uuid.

bahaaTuffaha: so you want to `.replace("Chat-ID:")` before passing it to qrCode?
also we need to remove the rest of the path except the id for the qrDetails

bahaaTuffaha:
Any idea why when I scan the QR with my phone it consider it as a deeplink and it opens the app but it shows a failure page.
Screenshot_2026-06-03-13-09-48-89_a78c51a44716047e2d17fb7ada84ae24

steffenkleinle:
> Any idea why when I scan the QR with my phone it consider it as a deeplink and it opens the app but it shows a failure page.

Perhaps test cms vs production cms? Hard to say without knowing the actual deep link.

web/src/components/QrCodeDialog.tsx
steffenkleinle: Sorry, I really don't get the question. If the qr code contains an url, the hint below should show exactly the same URL (no hint/explanation). If the qr code contains a chat id, the hint below should show exactly the same chat id. If we need a path/slug/URL in the qr code to share the chat (I think this has not been clarified yet), the hint should be exactly the same thing.
So in my opinion, the value of the qr code should always be exactly the same as the hint text below the qr code.

web/src/components/QrCodeDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+  title: string
+  description: string
+  url: string
+  qrDetails: string
+}
+
+const QrCodeDialog = ({ open, close, title, description, url, qrDetails }: QrCodeDialogProps): ReactElement | null => {
+  const { t } = useTranslation('layout')
+  const theme = useTheme()
+  const svgSrc = `data:image/svg+xml,${encodeURIComponent(encodeQR(url, 'svg'))}`
```
steffenkleinle: ❓ Why do we use a different way here to encode the qr code than on native?

bahaaTuffaha: I added it for a reason but I cant remember why. I tested it again without `encodeURIComponent` it seems redundant 👍

bahaaTuffaha:
> > Any idea why when I scan the QR with my phone it consider it as a deeplink and it opens the app but it shows a failure page.
> 
> Perhaps test cms vs production cms? Hard to say without knowing the actual deep link.

yeah because it's localhost on my phone and production cms. 😅

native/src/components/QrCodeModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  modalVisible: boolean
+  closeModal: () => void
+  title: string
+  description: string
+  content: string
+}
+
+const QrCodeModal = ({ modalVisible, closeModal, title, description, content }: QrCodeModalProps): ReactElement => {
+  const { t } = useTranslation('layout')
+  const theme = useTheme() as DefaultTheme
```
steffenkleinle: 🔧 Why do you need a typecast here?
```suggestion
  const theme = useTheme()
```

web/src/components/QrCodeDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  close: () => void
+  title: string
+  description: string
+  content: string
+}
+
+const QrCodeDialog = ({ open, close, title, description, content }: QrCodeDialogProps): ReactElement | null => {
+  const { t } = useTranslation('layout')
+  const theme = useTheme()
+  const svgSrc = `data:image/svg+xml,${encodeQR(content, 'svg')}`
```
steffenkleinle, line 28: 🤔 This is still different from native (which is just `const svgXml = encodeQR(content, 'svg')`). Not saying you should change this, just curious why this is necessary.

web/src/components/QrCodeDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <Typography variant='body2'>{description}</Typography>
+        <Box
+          sx={{
+            color: theme.palette.text.primary,
+            '& svg *': {
+              fill: 'currentColor',
+            },
+          }}>
+          <Svg src={svgSrc} width={QR_CODE_SIZE} height={QR_CODE_SIZE} />
+        </Box>
```
steffenkleinle: 🔧 So far we don't use inline styles for web. Please avoid this. Also, try to style `Svg` directly instead of using a child selector.

Edit: Actually, neither the `Box` nor the `color: theme.palette.text.primary` is necessary, the following is enough:

```suggestion
        <StyledSvg src={svgSrc} width={QR_CODE_SIZE} height={QR_CODE_SIZE} />
```

with 

``` tsx
const StyledSvg = styled(Svg)({
  fill: 'currentColor',
})
```

steffenkleinle (approved):
Tested on firefox, android and ios.

web/src/components/QrCodeDialog.tsx
bahaaTuffaha: the svgXml from `react-native-svg` can takes raw SVG but for web it expects a url.
If I changed it to  `const svgSrc = encodeQR(content, 'svg')` it wont show a QR code at web.

web/src/components/QrCodeDialog.tsx
bahaaTuffaha: Yeah. thanks this is much cleaner.

steffenkleinle (approved):
Nice work!

lunars97 (approved):
Very nicely done! Thank you for your effort 👍 LGTM. Tested with Chrome, Safari, iOS real device and Android real device

PR #4148, opened 2026-05-31
4094: Add space between last update and date

### LINKED ISSUE #4094: Add space between the last updated text and date
### Describe the Bug

Currently last updated text and date are squished together without any gap between them.
### Steps to Reproduce

Go to any page and see the last updated below. See the screenshot below.

### Expected Behavior

There should be a space between them

### Environment

- Device/Browser:
- OS version:
- App version:

### Additional Information

Image

lunars97:
### Short Description

This PR adds space between last update text and date in native

### Proposed Changes

- Combine `Text` into one and add space between the last update and the date
- Adjust test

### Side Effects

- none

### Testing

Check pages where last update and date exist in native

### Resolved Issues

Fixes: #4094

---

### DIFF
--- a/native/src/components/TimeStamp.tsx
+++ b/native/src/components/TimeStamp.tsx
@@ -30,7 +30,8 @@ export const TimeStamp = ({ lastUpdate, showText = true, format = 'DDD' }: TimeS
   return (
     <DirectionContainer language={i18n.language}>
       <Text variant='caption' style={styles.timeStampText}>
-        {showText && `${t('lastUpdate')} `}
+        {showText && t('lastUpdate')}
+        {showText && ' '}
         {lastUpdate.setLocale(i18n.language).toFormat(format)}
       </Text>
     </DirectionContainer>
--- a/native/src/components/TimeStamp.tsx
+++ b/native/src/components/TimeStamp.tsx
@@ -29,12 +29,8 @@ export const TimeStamp = ({ lastUpdate, showText = true, format = 'DDD' }: TimeS
 
   return (
     <DirectionContainer language={i18n.language}>
-      {showText && (
-        <Text variant='caption' style={styles.timeStampText}>
-          {t('lastUpdate')}
-        </Text>
-      )}
       <Text variant='caption' style={styles.timeStampText}>
+        {showText && `${t('lastUpdate')} `}
         {lastUpdate.setLocale(i18n.language).toFormat(format)}
       </Text>
     </DirectionContainer>
--- a/native/src/components/__tests__/TimeStamp.spec.tsx
+++ b/native/src/components/__tests__/TimeStamp.spec.tsx
@@ -21,15 +21,13 @@ describe('TimeStamp', () => {
     render(<TimeStamp lastUpdate={lastUpdate} format={format ?? undefined} showText={showText ?? undefined} />)
 
   it('should display last update text and formatted timestamp', () => {
-    const { findByText } = renderTimeStamp(null, null)
-    expect(findByText(/lastUpdate/)).toBeTruthy()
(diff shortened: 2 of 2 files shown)


native/src/components/__tests__/TimeStamp.spec.tsx
```diff
@@ -21,15 +21,15 @@ describe('TimeStamp', () => {
     render(<TimeStamp lastUpdate={lastUpdate} format={format ?? undefined} showText={showText ?? undefined} />)
 
   it('should display last update text and formatted timestamp', () => {
-    const { getByText } = renderTimeStamp(null, null)
-    expect(getByText(/lastUpdate/)).toBeTruthy()
-    expect(getByText(lastUpdate.setLocale('en').toFormat('DDD'))).toBeTruthy()
+    const { findByText } = renderTimeStamp(null, null)
+    expect(findByText(/lastUpdate/)).toBeTruthy()
+    expect(findByText(lastUpdate.setLocale('en').toFormat('DDD'))).toBeTruthy()
```
steffenkleinle: 🔧 This only fixes the symptom (i.e. failing tests), not the root cause of the problem (i.e. text not found). The tests now only succeed because `findByText` returns a promise which is truthy, even if the promise would return false if properly awaited. You can fix the actual problem by using the option `{ exact: false }` on the `getByText` matcher. However, I think it would be even better to try to match the whole text, i.e. `'${lastUpdate} ${timestamp}'`. This would also ensure that the bug (no space between) does not occur again.

native/src/components/TimeStamp.tsx
```diff
@@ -29,12 +29,8 @@ export const TimeStamp = ({ lastUpdate, showText = true, format = 'DDD' }: TimeS
 
   return (
     <DirectionContainer language={i18n.language}>
-      {showText && (
-        <Text variant='caption' style={styles.timeStampText}>
-          {t('lastUpdate')}
-        </Text>
-      )}
       <Text variant='caption' style={styles.timeStampText}>
+        {showText && `${t('lastUpdate')} `}
```
steffenkleinle: 🤔 Maybe the following would be more explicit such that we don't remove it by accident again:
```suggestion
        {showText && t('lastUpdate')} 
        {showText && ' '}
```

steffenkleinle (commented):
Tested on android, works as expected

steffenkleinle (approved):

native/src/components/TimeStamp.tsx
```diff
@@ -29,12 +29,9 @@ export const TimeStamp = ({ lastUpdate, showText = true, format = 'DDD' }: TimeS
 
   return (
     <DirectionContainer language={i18n.language}>
-      {showText && (
-        <Text variant='caption' style={styles.timeStampText}>
-          {t('lastUpdate')}
-        </Text>
-      )}
       <Text variant='caption' style={styles.timeStampText}>
+        {showText && t('lastUpdate')}
```
f1sh1918, line 33: ❓ Any reason not using backtick interpolation?
```
{showText && `${t('lastUpdate')} `}
```

f1sh1918 (approved):
LGTM 👍 

just some cosmetic optional improvement

native/src/components/TimeStamp.tsx
steffenkleinle: That was my suggestion as this is more explicit in the sense of that it is intended and not just a typo. No hard opinion on this though.

lunars97: I agree with @steffenkleinle otherwise it might get removed by accident I guess

kalairubachandrakumaran-oss:
tested ✅ 
1DC5061D-1605-4C38-B260-31DCDB78BD4C

PR #4156, opened 2026-06-02
4030: Show detailed error message for rate limit

### LINKED ISSUE #4030: Improve Chat Rate Limit Handling
### Describe the Problem

To prevent overloading the back end with too many chat messages and also to limit scanning for valid chat tokens, the chat API has a rate limit. The app currently shows a generic error message for all HTTP status codes, that are not okay. We should add a dedicated error message for users in case of HTTP 429 status codes.

Instead of "an error ocurred", the error message for users in case of a 429 status code should be "You sent too many messages. Please slow down".

Additionally, the chat messages disappear in case of an empty API response. We should keep the last known valid state in the app.

None

### Related Issues

https://github.com/digitalfabrik/integreat-chat/issues/460

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/jjcpcrgxd3f5ukp486i4inrb3r

steffenkleinle:
### Short Description

Show detailed error message for rate limit (e.g. in the chat).

### Proposed Changes

- Add a detailed error message for rate limit
- Add error handling for error code 429 rate limit

### Side Effects

N/A

### Testing

Trigger a rate limit and see the error message.

### Resolved Issues

Fixes: #4030

---


lunars97 (approved):
Tested with Chrome, works as expected 💯

bahaaTuffaha (commented):
❓  The loading keeps going after the limit hits. do you think we need to stop that?

Image

- Also If you refreshed the page the chat doesn't show any messages (due to the limit) do we need to show something like get back in 10 mins or adjust the backend to prevent post requests and keeps the get for the messages or something.

steffenkleinle:
> ❓ The loading keeps going after the limit hits. do you think we need to stop that?
>     * Also If you refreshed the page the chat doesn't show any messages (due to the limit) do we need to show something like get back in 10 mins or adjust the backend to prevent post requests and keeps the get for the messages or something.

@svenseeberg what do you think of that? Can we keep the get endpoint working and only block the post endpoint for new messages?

@bahaaTuffaha since these are API issues, could you still approve? Thanks :)

bahaaTuffaha (approved):

PR #4161, opened 2026-06-07
4160: Fix out of range error for index in poi list

### LINKED ISSUE #4160: Fix Uncaught Error at poi/places
### Describe the Bug

At poi/places if you clicked on the first element in the list it will throw an error:
```
scrollTolndex out of range: requested index -0.5 but minimum is 0
```

### Steps to Reproduce

1. Go to map/places.
2. Click on the first place on the list.
3. Check the error.

### Expected Behavior

- an experience without any crashes. (check the value before calling the scrollTolndex)

### Environment

- Device/Browser: Native
- OS version: Android 16
- App version: main branch

### Related Issues

Introduced in: #4113 

### Additional Information

Image

lunars97:
### Short Description

This PR fixes `scrollTolndex out of range: requested index -0.5 but minimum is 0` when a user chooses a first item from poi list

### Proposed Changes

- Fix by clamping `index - SCROLL_OFFSET` to a valid non negative integer

### Side Effects

- none

### Testing

Go to poi list and choose a first item

### Resolved Issues

Fixes: #4160 

---

### DIFF
--- a/native/src/components/PlacesBottomSheet.tsx
+++ b/native/src/components/PlacesBottomSheet.tsx
@@ -103,7 +103,10 @@ const PlacesBottomSheet = ({
     bottomSheetRef.current?.snapToIndex(1)
     const index = places.findIndex(it => it.path === place.path)
     if (index !== -1) {
-      flatListRef.current?.scrollToIndex({ index: index - SCROLL_OFFSET, animated: false })
+      flatListRef.current?.scrollToIndex({
+        index: Math.max(0, Math.round(index - SCROLL_OFFSET)),
+        animated: false,
+      })
     }
   }
 
@@ -157,6 +160,7 @@ const PlacesBottomSheet = ({
           role='list'
           {...conditionalA11yProps({ hidden: !!slug })}
           accessibilityLabel={t('placesCount', { count: places.length })}
+          keyExtractor={(item: PoiModel) => item.path}
           renderItem={renderPlaceListItem}
           showsVerticalScrollIndicator={false}
           ListHeaderComponent={<Text variant='h5'>{t('common:nearby')}</Text>}
--- a/native/src/components/PlacesBottomSheet.tsx
+++ b/native/src/components/PlacesBottomSheet.tsx
@@ -5,7 +5,7 @@ import BottomSheet, {
 } from '@gorhom/bottom-sheet'
 import React, { memo, ReactElement, useCallback, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { View } from 'react-native'
+import { Platform, View } from 'react-native'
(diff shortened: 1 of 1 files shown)


native/src/components/PoisBottomSheet.tsx
```diff
@@ -103,7 +103,10 @@ const PoisBottomSheet = ({
     bottomSheetRef.current?.snapToIndex(1)
     const index = pois.findIndex(it => it.path === poi.path)
     if (index !== -1) {
-      flatListRef.current?.scrollToIndex({ index: index - SCROLL_OFFSET, animated: false })
+      flatListRef.current?.scrollToIndex({
+        index: Math.max(0, Math.round(index - SCROLL_OFFSET)),
```
steffenkleinle: ❓ Why do we need `Math.round()`? We deliberately use `- 0.5` at the moment.

lunars97: but the issue is because of it :eyes:  the minimum should be 0 apparently and non negative integer, so far I did not see any issue with it

steffenkleinle: Yes, but the `Math.max()` already sufficiently solves this problem, doesn't it? The goal of the `0.5` was explicitly to clearly indicate the list is scrollable/the current item is not the first item in the list.

lunars97: sorry, I misunderstood you. Yes, this should be sufficient, no need to have extra method :+1:

steffenkleinle (approved):

bahaaTuffaha (commented):
Tested on Android ✅ 

on iOS the persisted list is not working (I don't think its from your changes though) 🤔

bahaaTuffaha:
> on iOS the persisted list is not working (I don't think its from your changes though) 🤔

found a workaround:
for `<StyledBottomSheet` 
```js
 enableContentPanningGesture={Platform.OS !== 'ios'}
```
and at `<BottomSheetFlatList`
```js
 onScrollBeginDrag={Platform.OS === 'ios' ? expandFullscreen : undefined}
```

steffenkleinle (approved):
Tested on android and ios

lunars97:
> > on iOS the persisted list is not working (I don't think its from your changes though) 🤔
> 
> found a workaround: for `<StyledBottomSheet`
> 
> ```js
>  enableContentPanningGesture={Platform.OS !== 'ios'}
> ```
> 
> and at `<BottomSheetFlatList`
> 
> ```js
>  onScrollBeginDrag={Platform.OS === 'ios' ? expandFullscreen : undefined}
> ```

Thank you for catching it! Tested on android as well, works as expected :heart_eyes:

bahaaTuffaha (approved):
Looks good ✅  tested on Android and iOS.

PR #4155, opened 2026-06-02
4045: Rename pois to places

steffenkleinle:
### Short Description

Rename pois to places.

### Proposed Changes

- Rename pois to places
- Rename multipoi to multiPlaces

### Side Effects

N/A

### Testing

Thoroughly test the maps/places feature on both web and native. Also test multiplaces, e.g. http://localhost:9000/muenchen/de/locations?multiplace=5.

### Resolved Issues

Part of: #4045

---

PR #4162, opened 2026-06-08
4045: Change places path to places

steffenkleinle:
### Short Description

This PR aligns the path of the places to the internal and user facing naming `places` (follow up for #4155).
The new slug is added to the CMS in https://github.com/digitalfabrik/integreat-cms/pull/4328.

### Proposed Changes

- Change the path of the places from https://integreat.app/testumgebung/de/locations to https://integreat.app/testumgebung/de/places
- Add redirects for the legacy path `locations`

### Side Effects

N/A

### Testing

Check that the places feature still works as expected in both native and web. Check that both the new `places` path as well as the legacy `locations` path works in both web and native.

### Resolved Issues

Part of: #4045

---

### DIFF
--- a/web/src/RootNavigator.tsx
+++ b/web/src/RootNavigator.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement, useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
-import { Navigate, Route, Routes, useMatch } from 'react-router'
+import { Navigate, Route, Routes, useLocation, useMatch, useParams } from 'react-router'
 
 import {
   SUGGEST_TO_REGION_ROUTE,
@@ -12,6 +12,9 @@ import {
   NOT_FOUND_ROUTE,
   pathnameFromRouteInformation,
   RESERVED_REGION_CONTENT_SLUGS,
+  LEGACY_REGIONS_ROUTE,
+  LEGACY_PLACES_ROUTE,
+  PLACES_ROUTE,
 } from 'shared'
 
 import FixedRegionContentNavigator from './FixedRegionContentNavigator'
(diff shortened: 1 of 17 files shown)


web/src/RootNavigator.tsx
bahaaTuffaha, line 1: http://localhost:9000/augsburg/de/locations
it redirects to http://localhost:9000/augsburg/de/places/
🤔  Which is good but there is a trailing slash... so far its not causing issues.

web/src/__tests__/RootNavigator.spec.tsx
```diff
@@ -60,18 +60,21 @@ describe('RootNavigator', () => {
...
+      ${'/augsburg'}                         | ${'/augsburg/de'}
+      ${'/augsburg/de'}                      | ${'/augsburg/de'}
+      ${'/augsburg/events'}                  | ${'/augsburg/de/events'}
+      ${'/augsburg/events/event-1'}          | ${'/augsburg/de/events/event-1'}
+      ${'/augsburg/news'}                    | ${'/augsburg/de/news'}
+      ${'/augsburg/news/local'}              | ${'/augsburg/de/news/local'}
+      ${'/augsburg/news/tu-news'}            | ${'/augsburg/de/news/tu-news'}
+      ${'/augsburg/de/locations'}            | ${'/augsburg/de/places'}
+      ${'/augsburg/de/locations/some-place'} | ${'/augsburg/de/places/some-place'}
+      ${'/augsburg/locations'}               | ${'/augsburg/de/places'}
```
bahaaTuffaha: 🙃  `'/augsburg/locations'` doesn't keep the multiplace query for example. Not sure if this edge case needs fixing.

bahaaTuffaha (approved):
Nice work! ✨ 
Tested web : integreat, malte and aschaffenburg ✅ 
Tested : native ✅ 
Tested `/landing`, `/locations` under different queries. ✅

web/src/__tests__/RootNavigator.spec.tsx
steffenkleinle: Good point! I fixed the keeping of query params for language-independent redirects. The persisting of multiplace queries won't help, because they have been renamed from `multipoi` anyway. I don't see the usecase where someone would share a multipoi, but its important for other query params imo (e.g. zoom).

web/src/RootNavigator.tsx
steffenkleinle: Should not be an issue, this is also the case for the redirect of language-independent redirects. Fixed it nevertheless now for both examples.

lunars97 (approved):
LGTM! 👍

kalairubachandrakumaran-oss:
tested (native+web+deeplink) ✅ 
https://github.com/user-attachments/assets/429263af-161a-4252-9054-8f2a3ddfe920

PR #4165, opened 2026-06-08
4154: Improve region search to match exact and prefix strings

### LINKED ISSUE #4154: Improve region search relevance by prioritizing exact and prefix matches
### Describe the Problem

Currently, the region search returns matches whenever the entered text appears anywhere within a region name or associated municipality name. For example, searching for "Mü" also returns results where the character sequence appears in the middle of a word. This leads to longer result lists and makes it harder for users to find the intended region.

Search results should use prefix matching on:
- the displayed region name
- associated municipality/city names

Substring matches within words should not be returned.

For example, searching for "Mü" should return:
- München
- Münster
- Mühlheim

The search should not return results where the search term only appears in the middle of a word, for example:
- Dortmund
- Schwabmünchen
- Altmünster

N/A

### Related Issues

Discussed in: #4144 

### Additional Information

Image

lunars97:
### Short Description
This PR improves region search by matching exact regions or aliases by using `startsWith` string method to match whether this string begins with the exact string

### Proposed Changes

- Adjust search.ts to always return beginning string matches for region names or aliases
- Create shared helpers `matchesWordStart` and `getMatchingAliases`
- Adjust `findNormalizedMatches` to match at the beginning or in the middle where the new word starts
- Add additional test cases

### Side Effects

- none

### Testing

Go to region selection and look for `Mu` and see that it returns exact region or alias beginning names

### Resolved Issues

Fixes: #4154 

---

### DIFF
--- a/native/src/components/RegionEntry.tsx
+++ b/native/src/components/RegionEntry.tsx
@@ -3,7 +3,7 @@ import { StyleSheet, View } from 'react-native'
 import { List as PaperList } from 'react-native-paper'
 import styled, { useTheme } from 'styled-components/native'
 
-import { getMatchingAliases, MAX_NUMBER_OF_ALIASES_SHOWN, normalizeString } from 'shared'
+import { getMatchingAliases, MAX_NUMBER_OF_ALIASES_SHOWN, findWordStartMatches } from 'shared'
 import { RegionModel } from 'shared/api'
 
 import { AppContext } from '../contexts/AppContext'
@@ -39,8 +39,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps): 
       color: theme.colors.onSurfaceVariant,
     },
   })
-  const normalizedQuery = normalizeString(query)
-  const matchingAliases = getMatchingAliases(region.aliases, normalizedQuery)
+  const matchingAliases = getMatchingAliases(region.aliases, query)
   const aliases = matchingAliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN)
   const { languageCode } = useContext(AppContext)
 
@@ -49,7 +48,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps): 
       <AliasesWrapper>
         {aliases.map((it, index) => (
           <Fragment key={it}>
-            <AliasLabel search={normalizedQuery} text={it} />
+            <AliasLabel search={query} text={it} findChunks={findWordStartMatches} />
             {index !== aliases.length - 1 && (
               <Text variant='body3' style={styles.separator}>
                 ,{' '}
@@ -71,7 +70,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps): 
       titleNumberOfLines={0}
       title={
         <View>
-          <Label search={normalizedQuery} text={region.name} />
+          <Label search={query} text={region.name} findChunks={findWordStartMatches} />
         </View>
       }
       description={Aliases}
--- a/native/src/components/RegionEntry.tsx
+++ b/native/src/components/RegionEntry.tsx
@@ -3,15 +3,13 @@ import { StyleSheet, View } from 'react-native'
 import { List as PaperList } from 'react-native-paper'
 import styled, { useTheme } from 'styled-components/native'
 
-import { normalizeString } from 'shared'
+import { getMatchingAliases, MAX_NUMBER_OF_ALIASES_SHOWN, normalizeString } from 'shared'
 import { RegionModel } from 'shared/api'
 
 import { AppContext } from '../contexts/AppContext'
 import testID from '../testing/testID'
 import Highlighter from './Highlighter'
 import Text from './base/Text'
-
-const MAX_NUMBER_OF_ALIASES_SHOWN = 3
 
 const Label = styled(Highlighter)`
   color: ${props => props.theme.colors.onSurface};
@@ -42,10 +40,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps): 
     },
   })
   const normalizedQuery = normalizeString(query)
-  const matchingAliases =
-    region.aliases && normalizedQuery.length >= 1
-      ? Object.keys(region.aliases).filter(alias => normalizeString(alias).startsWith(normalizedQuery))
-      : []
+  const matchingAliases = getMatchingAliases(region.aliases, normalizedQuery)
   const aliases = matchingAliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN)
   const { languageCode } = useContext(AppContext)
 
--- a/native/src/components/RegionEntry.tsx
+++ b/native/src/components/RegionEntry.tsx
@@ -44,7 +44,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps): 
   const normalizedQuery = normalizeString(query)
   const matchingAliases =
     region.aliases && normalizedQuery.length >= 1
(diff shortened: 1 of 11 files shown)


native/src/components/RegionEntry.tsx
```diff
@@ -44,7 +44,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps):
   const normalizedQuery = normalizeString(query)
   const matchingAliases =
     region.aliases && normalizedQuery.length >= 1
-      ? Object.keys(region.aliases).filter(alias => normalizeString(alias).includes(normalizedQuery))
+      ? Object.keys(region.aliases).filter(alias => normalizeString(alias).startsWith(normalizedQuery))
```
steffenkleinle: 🔧 I think we also need the same algorithm here to split the words, otherwise the matched aliases are not displayed.
Please also add a test case testing this case. Same for web (perhaps we can move the logic to shared).

shared/utils/__tests__/search.spec.ts
```diff
@@ -107,7 +107,7 @@ describe('search', () => {
         }),
         region({ sortingName: 'Nürnberg' }),
       ]
-      expect(filterSortRegions(regions, 'äch')).toEqual([regions[0], regions[3], regions[4], regions[5]])
+      expect(filterSortRegions(regions, 'ba')).toEqual([regions[3], regions[4], regions[5]])
```
steffenkleinle, line 110: 🔧 Could you please include one region that asserts that a region is not returned if `ba` is found in the middle?

native/src/components/RegionEntry.tsx
```diff
@@ -42,10 +40,7 @@ const RegionEntry = ({ region, query, navigateToDashboard }: RegionEntryProps):
     },
   })
   const normalizedQuery = normalizeString(query)
-  const matchingAliases =
-    region.aliases && normalizedQuery.length >= 1
-      ? Object.keys(region.aliases).filter(alias => normalizeString(alias).includes(normalizedQuery))
-      : []
+  const matchingAliases = getMatchingAliases(region.aliases, normalizedQuery)
```
steffenkleinle: 🔧 What do you think of also moving normalizing the query to `getMatchingAliases` in `shared`?

shared/utils/__tests__/search.spec.ts
```diff
@@ -107,7 +107,24 @@ describe('search', () => {
...
+    })
+
+    it('should not return live regions when the string is mid-word of the names or aliases', () => {
+      const regions = [
+        region({ sortingName: 'Aichach' }),
+        region({ sortingName: 'Stadtbaden' }),
```
steffenkleinle, line 116: 🏊

shared/utils/findNormalizedMatches.ts
```diff
@@ -11,16 +14,21 @@ const charsAddedByNormalization = (text: string, until: number) => {
...
+      .map(chunk => {
+        const charsAddedBeforeMatch = charsAddedByNormalization(props.textToHighlight, chunk.start)
+        const charsAddedIncludingMatch = charsAddedByNormalization(props.textToHighlight, chunk.end)
+        return {
+          ...chunk,
+          start: chunk.start - charsAddedBeforeMatch,
+          end: chunk.end - charsAddedIncludingMatch,
+        }
+      })
+  )
```
steffenkleinle, line 41: ❌ This will break behavior for the normal region content search where we also match parts of words inside the content. I think it is okay that we also show inline matches in the region search even if we don't actually filter by them. Especially since I don't think there is that many cases when a reasonable query is entered where one word contains the query twice. However, if you think this actually should be fixed, I think you'd need a parameter here :)

shared/utils/search.ts
```diff
@@ -26,3 +36,13 @@ export const filterSortRegions = (
   filterText: string,
   developerFriendly = false,
 ): RegionModel[] => regions.filter(regionFilter(filterText, developerFriendly)).sort(regionSort)
+
+export const getMatchingAliases = (
+  aliases: Record<string, CoordinateType> | null,
```
steffenkleinle: 🙃 Since you never use the actual coordinates, you could also just use this:
```suggestion
  aliases: Record<string, unknown> | null,
```

steffenkleinle (changes requested):

shared/utils/findNormalizedMatches.ts
lunars97: I created wrappers with a flag to align with needs of region content search and region search. Thank you for catching this point! :+1:

shared/utils/findNormalizedMatches.ts
```diff
@@ -10,19 +13,32 @@ const charsAddedByNormalization = (text: string, until: number) => {
...
+      .filter(chunk => chunk.start === 0 || MATCH_WHITESPACE_AND_DASHES.test(normalized[chunk.start - 1] ?? ''))
+      .map(normalizedMatches)
+  )
 }
 
 export const findAllMatches = findAll
 
-export default findNormalizedMatches
+export const findNormalizedMatches = (props: FindChunks): Chunk[] => buildMatches(props, false)
+export const findWordStartMatches = (props: FindChunks): Chunk[] => buildMatches(props, true)
```
steffenkleinle: 🔧 Can we have only one function for both use cases and have an option prop to allow only match at word starts? Also, for options like this I'd always use objects instead of simple values because it directly adds context, i.e. it is clear what the param does. With simple values, you have to go to the definition to see what the parameter is actually about.

```suggestion
export const findNormalizedMatches = (props: FindChunks, { wordStartOnly = false }: FindNormalizedMatchesOptions = {}): Chunk[] {
```

Then you also don't need a separate function `buildMatches` but can just keep the previous code (with the addition of wordStartOnly matching).

steffenkleinle (commented):
Nice work and great implementation. Only one minor improvement to avoid having to different functions for the same task :)

shared/utils/findNormalizedMatches.ts
```diff
@@ -1,5 +1,12 @@
 import { Chunk, FindAll, findAll, findChunks, FindChunks } from 'highlight-words-core'
 
+import normalizeString from './normalizeString.js'
+import { MATCH_WHITESPACE_AND_DASHES } from './search.js'
+
+type FindNormalizedMatchesOptions = {
+  wordStartOnly?: boolean
+}
```
steffenkleinle: 🙃 I am always in favor to putting the types for props/parameters directly next to the actual function/component to see what I need/can pass at first glance without needing to search for the corresponding type.

steffenkleinle (approved):
Well done!

shared/utils/search.ts
```diff
@@ -26,3 +36,11 @@ export const filterSortRegions = (
   filterText: string,
   developerFriendly = false,
 ): RegionModel[] => regions.filter(regionFilter(filterText, developerFriendly)).sort(regionSort)
+
+export const getMatchingAliases = (aliases: Record<string, unknown> | null, filterText: string): string[] => {
```
bahaaTuffaha, line 40:
```suggestion
export const getMatchingAliases = (aliases: Record<string, CoordinateType> | null, filterText: string): string[] => {
```
and export the `CoordinateType` to import it here.

shared/utils/search.ts
```diff
@@ -26,3 +36,11 @@ export const filterSortRegions = (
...
+  const normalizedFilter = normalizeString(filterText)
```
bahaaTuffaha: ❓ How about moving `normalizedFilter` line under the if statement  `if (!aliases)` just to not compute this function if there is an early return.

bahaaTuffaha (approved):
Great work ✨ 
Tested web and native ✅

shared/utils/search.ts
lunars97: @steffenkleinle mentioned that it is better to have it unknown while we are not using actual coordinates anyway

PR #4168, opened 2026-06-09
4103: Open feedback via URL query

### LINKED ISSUE #4103: Open feedback via URL query parameter
### Describe the Problem

We currently open the feedback via a local state parameter. While this works good for our feedback buttons, it is not possible to open the feedback from other routes (e.g. a link in a popup) without including the whole feedback component in the tree again, which is complicated, error prone and leads to duplication.

- Instead of using a local state, use a query parameter `feedback=[positive|negative]` to open the feedback (check if you can reuse `useQueryParamVisibility`)
- Render the `FeedbackContainer` once in the app (compare chat container)
- Remove `noResults` parameter and replace it with `feedback=negative`
- Think of a nice way to handle navigating back/forth
- Replace existing usages of FeedbackContainer in toolbar and search with query solution

N/A

### Related Issues

Blocks: #4104 
Discussed in: https://github.com/digitalfabrik/integreat-app/pull/4050#discussion_r3116164093

### Additional Information

N/A

bahaaTuffaha:
### Short Description

We currently open the feedback via a local state parameter. While this works good for our feedback buttons, it is not possible to open the feedback from other routes (e.g. a link in a popup) without including the whole feedback component in the tree again, which is complicated, error prone and leads to duplication.

### Proposed Changes

- Created new query key `FEEDBACK_QUERY_KEY` and new param to `toQueryParams`.
- Reused the  `useQueryParamVisibility` and adjusted the open function to accept an optional parameter `value`.
- Moved the `FeedbackContainer` from `FeedbackToolbarItem` into RegionContentLayout.tsx to behave like chat so it can be open form any where.
- Adjusted `FeedbackContainer` to wrap the feedback with dialog instead and removed all of the props because we will depend on the queries to get the value for the feedback and the search query.
- Removed `FeedbackContainer` from `FeedbackToolbarItem.tsx` and `RegionContentMenu.tsx`.
- To extract search terms from quires I created a hook called `useSearchFeedback.ts` and path to extract slug.
- Adjusted tests.

### Side Effects

- Now the search feedback will show up as pop up instead of embedded within the page.

### Testing

- Test feedback across the whole web-app.
- Test search feedback (with/without results)
- Test how it behaves with other queries like categories filter at the map , search and chat.
- Check malte and aschaffenburg.

### Resolved Issues

Fixes: #4103

---


web/src/hooks/useSearchFeedback.ts
```diff
@@ -0,0 +1,36 @@
...
+  const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
+
+  useEffect(() => {
+    setRating(initialRating)
+  }, [initialRating])
+
+  useEffect(() => {
+    setSearchTerm(query)
+  }, [query])
+
```
steffenkleinle: ❓ Why do we need to persist those values in the state and use useEffects instead of just returning them directly?

I am actually not sure if a hook is needed here, we basically just compute simple values from `route` and `queryParams`.
This can also be made a lot simpler by using `parseQueryParams` and applying my suggestions for `useQueryParamVisibility`.

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -3,19 +3,20 @@ import { useSearchParams } from 'react-router'
...
   const [queryParams, setQueryParams] = useSearchParams()
-  const visible = parseQueryParams(queryParams)[key] ?? false
+  const paramValue = parseQueryParams(queryParams)[key]
+  const visible = typeof paramValue === 'boolean' ? paramValue : paramValue !== undefined
 
-  const open = () => {
+  const open = (value?: string) => {
     const newQueryParams = queryParams
-    newQueryParams.set(key, 'true')
+    newQueryParams.set(key, typeof value === 'string' ? value : 'true')
```
steffenkleinle: 🔧 
```suggestion
    newQueryParams.set(key, value ?? 'true')
```

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -3,19 +3,20 @@ import { useSearchParams } from 'react-router'
...
-  openUrl: (url: string | null) => string | null
+  openUrl: (url: string | null, value?: string) => string | null
   visible: boolean
 }
 
 const useQueryParamVisibility = (key: keyof VisibilityQueryParams): UseQueryParamVisibilityReturn => {
   const [queryParams, setQueryParams] = useSearchParams()
-  const visible = parseQueryParams(queryParams)[key] ?? false
+  const paramValue = parseQueryParams(queryParams)[key]
+  const visible = typeof paramValue === 'boolean' ? paramValue : paramValue !== undefined
```
steffenkleinle: 🤔 What do you think of the following?
```suggestion
  const value = parseQueryParams(queryParams)[key]
  const visible = Boolean(value)
```

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -3,19 +3,20 @@ import { useSearchParams } from 'react-router'
...
 type UseQueryParamVisibilityReturn = {
-  open: () => void
+  open: (value?: string) => void
   close: () => void
-  openUrl: (url: string | null) => string | null
+  openUrl: (url: string | null, value?: string) => string | null
   visible: boolean
 }
 
 const useQueryParamVisibility = (key: keyof VisibilityQueryParams): UseQueryParamVisibilityReturn => {
```
steffenkleinle: 🔧 I think we should correctly type this and it would be nice to directly add the `value` in the return type to avoid the need to manually parse the query params again:
``` suggestion
type UseQueryParamVisibilityReturn<T extends keyof VisibilityQueryParams> = {
  open: (value?: VisibilityQueryParams[T]) => void
  close: () => void
  openUrl: (url: string | null, value?: VisibilityQueryParams[T]) => string | null
  visible: boolean
  value: VisibilityQueryParams[T]
}

const useQueryParamVisibility = <T extends keyof VisibilityQueryParams>(key: T): UseQueryParamVisibilityReturn<T> => {
```

web/src/components/RegionContentMenu.tsx
```diff
@@ -68,18 +65,9 @@ const RegionContentMenu = ({ slug, category, pageTitle, fitScreen }: RegionConte
...
-        <Dialog
-          title={feedbackSubmitted ? t('feedback:thanksHeadline') : t('feedback:headline')}
-          close={() => setFeedbackOpen(false)}>
-          <FeedbackContainer onSubmit={() => setFeedbackSubmitted(true)} slug={slug} initialRating={null} />
-        </Dialog>
-      )}
-    </>
+    <HeaderMenu pageTitle={pageTitle} fitScreen={fitScreen} ref={ref}>
+      {items}
+    </HeaderMenu>
```
steffenkleinle, line 70: 💪 very nice refactoring, I love it!

web/src/components/FeedbackToolbarItem.tsx
```diff
@@ -1,73 +1,23 @@
...
-            <CloseIcon />
-          </IconButton>
-        }
-      />
-    </>
+    <ToolbarItem
+      icon={rating === RATING_POSITIVE ? <SentimentSatisfiedOutlinedIcon /> : <SentimentDissatisfiedOutlinedIcon />}
+      text={t(rating === RATING_POSITIVE ? 'useful' : 'notUseful')}
+      onClick={() => open(rating ?? undefined)}
+    />
```
steffenkleinle, line 20: 😍

web/src/components/SearchFeedback.tsx
```diff
@@ -32,20 +33,12 @@ type SearchFeedbackProps = {
...
-  useEffect(() => setShowFeedback(false), [query])
-
-  if (showFeedback) {
-    return (
-      <Container>
-        <FeedbackContainer query={query} initialRating={null} noResults={noResults} />
-      </Container>
-    )
-  }
+  const openFeedback = () => open(noResults ? RATING_NEGATIVE : undefined)
```
steffenkleinle: 🤔 I think I would perhaps even just assume that feedback for the search is always negative. If not, the user is still able to switch it to positive manually.
```suggestion
  const openFeedback = () => open(RATING_NEGATIVE)
```

web/src/components/FeedbackContainer.tsx
```diff
@@ -1,46 +1,47 @@
...
-  }, [query])
+  // Extract search from query params if available
+  const { pathname } = useLocation()
+  const { rating, setRating, searchTerm, setSearchTerm, query } = useSearchFeedback(queryParams)
+  const pathParts = pathname.split('/').filter(Boolean)
+  const pathAfterLanguage = pathParts.slice(3)
+  const slug =
+    (route === CATEGORIES_ROUTE || route === EVENTS_ROUTE || route === POIS_ROUTE) && pathAfterLanguage.length > 0
+      ? decodeURIComponent(pathAfterLanguage[pathAfterLanguage.length - 1] ?? '')
+      : undefined
```
steffenkleinle: ❌ Instead of doing something like this which is imo quite error prone and hacky, I'd suggest to pass the `slug` down to `FeedbackContainer` through `RegionContentLayout`, similar to how it was previously implemented in `RegionContentToolbar`/`RegionContentMenu`.

web/src/components/SearchFeedback.tsx
```diff
@@ -32,20 +33,12 @@ type SearchFeedbackProps = {
   noResults: boolean
```
steffenkleinle: 🔧 `query` is not used anymore, so lets remove it. Also, if you apply my suggestion below, `noResults` is not used anymore and could be removed.

steffenkleinle (changes requested):
Really really nice refactoring, gerat work! I love it! I think we can make a few things even simpler/more consistent, but otherwise this looks great already!
Tested on firefox, works as expected!

web/src/components/SearchFeedback.tsx
bahaaTuffaha: for search Feedback there is no toggles for rating... so I kept the same logic: no rating if there is a result and negative if there is non.

web/src/hooks/useQueryParamVisibility.ts
bahaaTuffaha: Much better 🙈

web/src/hooks/useQueryParamVisibility.ts
bahaaTuffaha: where do we need to use this value for?

web/src/hooks/useSearchFeedback.ts
bahaaTuffaha: I thought we could depend on the search query to refactor it more but it looks like I made it more messy

web/src/hooks/useQueryParamVisibility.ts
steffenkleinle: You currently use the value in `useSearchFeedback` which could then be replaced directly with this value and after edits you'll still need this value somewhere. More of an optional suggestion though but I think it might be handy. Should be typed in any case though imo.

web/src/components/SearchFeedback.tsx
steffenkleinle: Oh, good point. I think the question still stands though, in what case do you want to sent non-negative feedback from the search? I think its much more likely that it is negative feedback, even if there are some results. I'd therefore still set it always to negative.

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -2,20 +2,22 @@ import { useSearchParams } from 'react-router'
...
   const [queryParams, setQueryParams] = useSearchParams()
-  const visible = parseQueryParams(queryParams)[key] ?? false
+  const value = parseQueryParams(queryParams)[key]
+  const visible = Boolean(value)
 
-  const open = () => {
+  const open = (value?: string) => {
     const newQueryParams = queryParams
-    newQueryParams.set(key, 'true')
+    newQueryParams.set(key, value ?? 'true')
```
bahaaTuffaha: at ChatContainer.tsx‎:

I had to change `onClick(open)` to :
`<ChatFab onClick={() => open()} />`

Any idea why the chat didnt open the the url turned into:
http://localhost:9000/testumgebung/de?chat=%5Bobject+Object%5D

web/src/components/FeedbackContainer.tsx
```diff
@@ -2,42 +2,49 @@ import CloseIcon from '@mui/icons-material/Close'
...
-  const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
   const { route, regionCode, languageCode } = useRegionContentParams()
 
+  const { feedback, searchText } = parseQueryParams(queryParams)
+  const initialRating = feedback === RATING_POSITIVE || feedback === RATING_NEGATIVE ? feedback : null
+  const query = route === SEARCH_ROUTE ? searchText : undefined
+  const [rating, setRating] = useState<Rating | null>(initialRating)
+  const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
+
+  useEffect(() => setRating(initialRating), [initialRating])
```
bahaaTuffaha: couldn't remove this useEffect the only other way to solve this by creating another component that has all of the feedback stuff and pass the initialRating as prop so any change in props will cause rerender.

web/src/hooks/useQueryParamVisibility.ts
steffenkleinle: Yes, because `onClick` of the actual `Fab` component passes an event ` to the function. I guess we should try to find some way to prevent this from happening with type checking such that an error is raised instead of this.

web/src/components/FeedbackContainer.tsx
steffenkleinle: Could you please explain why this `useEffect` is necessary/what problem does it solve?

bahaaTuffaha: it triggers a re-render so when we hit one of the ratings from the toolbar it should toggle that when you open the feedback modal/alert pop up.

steffenkleinle: The right way would be to reduce duplicated state, in this case remove the `rating` `useState` altogether and instead only rely on the query param/update that if the user changes the rating.

web/src/hooks/useQueryParamVisibility.ts
bahaaTuffaha: with type checking like it was before?
`visible = typeof paramValue === 'boolean' ? paramValue : paramValue !== undefined`

steffenkleinle: This would only fix the symptom, not the problem of the query param with "object object" in the query param. I have to give it another thought later...

steffenkleinle: Probably the best way is actually to make `open` take a required value. That way, the types have to match and we get an error. It then requires to always pass a value, but I think its way safer that way (even for the feedback).

shared/routes/query.ts
```diff
@@ -39,6 +40,7 @@ export const queryStringFromRouteInformation = (
 
 export type VisibilityQueryParams = {
   chat?: boolean
```
steffenkleinle: 🔧 Let's restrict this to only `true`, `false` is not really a valid (i.e. supported) value:
```suggestion
  chat?: true
```

web/src/components/FeedbackContainer.tsx
```diff
@@ -2,42 +2,47 @@ import CloseIcon from '@mui/icons-material/Close'
...
 import useRegionContentParams from '../hooks/useRegionContentParams'
 import { reportError } from '../utils/sentry'
 import Feedback from './Feedback'
+import Dialog from './base/Dialog'
 import Snackbar from './base/Snackbar'
 
 type FeedbackContainerProps = {
-  query?: string
-  noResults?: boolean
   slug?: string
```
steffenkleinle: 🔧 I'd also make this required here (with null or undefined):
```suggestion
  slug: string | null
```

web/src/components/RegionContentLayout.tsx
```diff
@@ -23,6 +24,7 @@ export type RegionContentLayoutProps = {
   fitScreen?: boolean
   category?: CategoryModel
   pageTitle: string | null
+  slug?: string
```
steffenkleinle: 🔧 Let's make this required (but allow setting `null`) to avoid forgetting it:
```suggestion
  slug: string | null
```

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -2,32 +2,38 @@ import { useSearchParams } from 'react-router'
 
 import { parseQueryParams, VisibilityQueryParams, toQueryParams } from 'shared'
 
-type UseQueryParamVisibilityReturn = {
+type UseQueryParamVisibilityReturn<T extends keyof VisibilityQueryParams> = {
   open: () => void
+  openWith: (value: string) => void
```
steffenkleinle: ❌ I don't think we should introduce a second method here. Instead, lets rather just have one with a required value.
```suggestion
  open: (value: string) => void
```

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -2,32 +2,38 @@ import { useSearchParams } from 'react-router'
 
 import { parseQueryParams, VisibilityQueryParams, toQueryParams } from 'shared'
 
-type UseQueryParamVisibilityReturn = {
+type UseQueryParamVisibilityReturn<T extends keyof VisibilityQueryParams> = {
```
steffenkleinle, line 5: ❌ Please apply the correct typing as mentioned in another comment: https://github.com/digitalfabrik/integreat-app/pull/4168#discussion_r3386338078

steffenkleinle (changes requested):

web/src/hooks/useQueryParamVisibility.ts
bahaaTuffaha: 🤔  I applied it exactly as you suggested?

steffenkleinle: Yes, but you don't actually use the type. Check my comment again, there was more than just this one change :D Or rather: `open` and `openUrl` should also use it.

web/src/components/FeedbackContainer.tsx
```diff
@@ -2,42 +2,47 @@ import CloseIcon from '@mui/icons-material/Close'
...
-export const FeedbackContainer = ({
-  query,
-  noResults,
-  slug,
-  onSubmit,
-  onError,
-  initialRating,
-}: FeedbackContainerProps): ReactElement => {
+const FeedbackContainer = ({ slug }: FeedbackContainerProps): ReactElement | null => {
+  const { visible, open, close, value } = useQueryParamVisibility(FEEDBACK_QUERY_KEY)
```
steffenkleinle: 🔧 
```suggestion
  const { visible, open, close, value: rating } = useQueryParamVisibility(FEEDBACK_QUERY_KEY)
```

shared/routes/query.ts
```diff
@@ -38,7 +39,8 @@ export const queryStringFromRouteInformation = (
 }
 
 export type VisibilityQueryParams = {
-  chat?: boolean
+  chat?: true
+  feedback?: string
```
steffenkleinle: 🔧 Let's actually move the correct parsing here. This ensures typesafety when opening the feedback and removes duplicated code if we need the rating in multiple places.
```suggestion
  feedback?: Rating
```

shared/routes/query.ts
```diff
@@ -50,11 +52,12 @@ type QueryParams = VisibilityQueryParams & {
 
 export const parseQueryParams = (queryParams: URLSearchParams): QueryParams => {
   const searchText = queryParams.get(SEARCH_QUERY_KEY) ?? undefined
-  const chat = queryParams.get(CHAT_QUERY_KEY) ? queryParams.get(CHAT_QUERY_KEY) === 'true' : undefined
+  const chat = queryParams.get(CHAT_QUERY_KEY) === 'true' || undefined
+  const feedback = queryParams.get(FEEDBACK_QUERY_KEY) ?? undefined
```
steffenkleinle: 🔧 
```suggestion
  const feedbackQuery = queryParams.get(FEEDBACK_QUERY_KEY) ?? undefined
  const feedback = feedbackQuery === RATING_POSITIVE || feedbackQuery === RATING_NEGATIVE ? feedbackQuery : undefined
```

web/src/components/FeedbackContainer.tsx
```diff
@@ -2,42 +2,47 @@ import CloseIcon from '@mui/icons-material/Close'
...
   const [sendingStatus, setSendingStatus] = useState<SendingStatusType>('idle')
   const [snackbarOpen, setSnackbarOpen] = useState(false)
-  const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
   const { route, regionCode, languageCode } = useRegionContentParams()
 
+  const { searchText } = parseQueryParams(queryParams)
+  const query = route === SEARCH_ROUTE ? searchText : undefined
+  const rating = value === RATING_POSITIVE || value === RATING_NEGATIVE ? value : null
+  const setRating = (newRating: Rating | null) => open(newRating ?? undefined)
+  const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
```
steffenkleinle: 🙃  Would you mind moving that back to the top to the rest of the `useState`s? I think it increases readability if we keep them together.

web/src/components/FeedbackToolbarItem.tsx
steffenkleinle, line 1: 🤔 We could think of removing this component now altogether and moving its logic directly inside of `RegionContentToolbar` since your refactoring made it so much simpler 🎉  But this is optional, whatever you think is best :)

shared/routes/query.ts
bahaaTuffaha: I like this idea 👍

web/src/components/FeedbackToolbarItem.tsx
bahaaTuffaha: I will keep it as is.

steffenkleinle (approved):
Nice work! Not tested again :)

web/src/routes/PlacesPage.tsx
```diff
@@ -53,6 +53,7 @@ const PlacesPage = ({ regionCode, languageCode, region, pathname }: RegionRouteP
     languageChangePaths,
     languageCode,
     pageTitle: null,
+    slug: null,
```
lunars97: shouldn't this be with a slug? Otherwise it will not be clear which location it is 🤔 
```suggestion
    slug: place?.slug ?? null,
```

lunars97 (changes requested):
Looks almost perfect to me 😻 , except for one question below.

web/src/routes/PlacesPage.tsx
bahaaTuffaha: Thank you this is a good find 😅

bahaaTuffaha: Yes without it the places won't pass the name of the place/poi when I send a feedback 👍

lunars97 (approved):
LGTM! Thank you for cleaning up and simplifying everything. Not tested again, tested before in Chrome as well as for Malte and Aschaffenburg 👍

PR #4171, opened 2026-06-09
4146: Show badge for new chat messages

### LINKED ISSUE #4146: Show badge for new chat messages
### Describe the Problem

As a user, I'd like to see if there are new chat messages without necessarily having to open the chat modal itself.

Add a badge with a counter of new messages to the chat fab.
With #4145, the chat messages are now always polled in the background.
See https://mui.com/material-ui/react-badge.

N/A

### Related Issues

#4129

### Additional Information

https://mui.com/material-ui/react-badge/

bahaaTuffaha:
### Short Description

As a user, I'd like to see if there are new chat messages without necessarily having to open the chat modal itself.

### Proposed Changes

- Created new key called `CHAT_NEW_MESSAGES_COUNT_STORAGE_KEY`
- Used `useLocalStorage` to store the new messages count into local storage I added I appended the region code just to differentiate count for each region.
- Added new prop to ChatFab called `unreadMessageCount` and passes it to the badge `incomingMessageCount - lastSeenMessageCount`.
- Added simple useEffect to set `setLastSeenMessageCount` when we open the chat.
- Added new test at `ChatContainer.spec.tsx`

### Side Effects

- none.

### Testing

- Test chat at web-app by submitting a message then close the chat and wait for the badge to appear.
- Test different regions like Munich and testumgebung to check different counts.
- Check contrast mode.
- Check small screens.

### Resolved Issues

Fixes: #4146

---


web/src/components/ChatContainer.tsx
```diff
@@ -50,6 +53,10 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
     key: CHAT_UNSYNCED_MESSAGES_STORAGE_KEY,
     initialValue: [],
   })
+  const [lastSeenMessageCount, setLastSeenMessageCount] = useLocalStorage<number>({
+    key: `${CHAT_NEW_MESSAGES_COUNT_STORAGE_KEY}-${region.code}`,
+    initialValue: 0,
+  })
```
steffenkleinle: 🔧 Let's keep the naming consistent and name it was it actually is:
```suggestion
  const [seenMessageCount, setSeenMessageCount] = useLocalStorage<number>({
    key: `${CHAT_SEEN_MESSAGE_COUNT_STORAGE_KEY}-${region.code}`,
    initialValue: 0,
  })
```

Alternatively, we could make it even shorter and omit `Count` and do `seenMessages` instead.

🙃 Perhaps we event want to create a reusable helper function to create the key, similar to `chatIdKey`.

web/src/components/ChatContainer.tsx
```diff
@@ -64,15 +71,22 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
...
+
+  useEffect(() => {
+    if (visible && incomingMessageCount > lastSeenMessageCount) {
+      setLastSeenMessageCount(incomingMessageCount)
+    }
+  }, [incomingMessageCount, lastSeenMessageCount, setLastSeenMessageCount, visible])
 
   useEffect(() => {
-    if (!isBrowserTabActive || messageCount === 0) {
+    if (!isBrowserTabActive || incomingMessageCount === 0) {
```
steffenkleinle: ❓  This will prevent polling before the first message is received, won't it? I think this should be changed back 
```suggestion
    if (!isBrowserTabActive || messageCount === 0) {
```

web/src/components/ChatFab.tsx
```diff
@@ -23,17 +24,20 @@ const ChatButtonContainer = styled('div')<{ bottom: number }>`
 
 type ChatButtonProps = {
   onClick: () => void
+  unreadMessageCount?: number
```
steffenkleinle: 🔧 Since we always pass this prop:
```suggestion
  unreadMessageCount: number
```

web/src/hooks/useLocalStorage.ts
```diff
@@ -6,6 +6,7 @@ export const CHAT_HINT_VISIBLE_STORAGE_KEY = 'Chat-Hint-Visible'
 export const CHAT_PRIVACY_POLICIES_STORAGE_KEY = 'Chat-Privacy-Policies'
 export const CHAT_ID_STORAGE_KEY = 'Chat-Device-Id'
 export const CHAT_UNSYNCED_MESSAGES_STORAGE_KEY = 'Chat-Unsynced-Messages'
+export const CHAT_NEW_MESSAGES_COUNT_STORAGE_KEY = 'Chat-New-Messages-Count'
```
steffenkleinle:
```suggestion
export const CHAT_SEEN_MESSAGE_COUNT_STORAGE_KEY = 'Chat-Seen-Message-Count'
```

web/src/components/ChatContainer.tsx
```diff
@@ -64,15 +71,22 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
 
   const { data, refetch } = response
   const botTyping = data?.botTyping
-  const messageCount = data?.messages.length ?? 0
+  const incomingMessageCount = data?.messages.filter(message => !message.userIsAuthor).length ?? 0
+  const unreadMessageCount = Math.max(0, incomingMessageCount - lastSeenMessageCount)
```
steffenkleinle: 🤔 Since you only check if `unreadMessageCount > 0`, the `Math.max()` could be omitted.

web/src/components/ChatContainer.tsx
```diff
@@ -64,15 +71,22 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
...
+
+  useEffect(() => {
+    if (visible && incomingMessageCount > lastSeenMessageCount) {
+      setLastSeenMessageCount(incomingMessageCount)
+    }
+  }, [incomingMessageCount, lastSeenMessageCount, setLastSeenMessageCount, visible])
```
steffenkleinle: 🤔 I am not sure if a `useEffect` is the best solution here, in theory the last seen message count should probably only be updated in case of an actual user interaction. Not sure how though. Otherwise this is probably fine.

steffenkleinle (commented):
Nice work, tested on firefox and works as expected!

web/src/components/ChatContainer.tsx
bahaaTuffaha: - when you open the chat it will get read so the `incomingMessageCount` turns into lastSeen.
- You will not receive any messages if you didn't interact/send any messages.
- I think you get 1 message incoming most of the time because of the bot replaying usually with 1 message except the welcome message at first.

web/src/components/ChatContainer.tsx
bahaaTuffaha: Yeah... nice find!

web/src/components/ChatContainer.tsx
```diff
@@ -77,6 +89,7 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
   const resetChat = () => {
     setUnsyncedMessages([])
     setChatId(generateChatId())
+    setSeenMessages(0)
```
steffenkleinle, line 92: 👍

web/src/components/ChatFab.tsx
```diff
@@ -23,17 +24,20 @@ const ChatButtonContainer = styled('div')<{ bottom: number }>`
 
 type ChatButtonProps = {
   onClick: () => void
+  unreadMessageCount: number
 }
 
-const ChatFab = ({ onClick }: ChatButtonProps): ReactElement => {
+const ChatFab = ({ onClick, unreadMessageCount = 0 }: ChatButtonProps): ReactElement => {
```
steffenkleinle: 🔧 
```suggestion
const ChatFab = ({ onClick, unreadMessageCount }: ChatButtonProps): ReactElement => {
```

steffenkleinle (approved):
Nice!

web/src/components/ChatFab.tsx
```diff
@@ -23,17 +24,20 @@ const ChatButtonContainer = styled('div')<{ bottom: number }>`
...
+  unreadMessageCount: number
 }
 
-const ChatFab = ({ onClick }: ChatButtonProps): ReactElement => {
+const ChatFab = ({ onClick, unreadMessageCount }: ChatButtonProps): ReactElement => {
   const { desktop, visibleFooterHeight, bottomNavigationHeight } = useDimensions()
   const chatName = getChatName(buildConfig().appName)
 
   return (
     <ChatButtonContainer bottom={bottomNavigationHeight ?? visibleFooterHeight}>
```
lunars97, line 38: I think we can reuse the LiveAnnouncer component here for screen-readers, we could add sth like in L36, what do you think?
```suggestion
    <LiveAnnouncer
        message={
          unreadMessageCount > 0
            ? `${unreadMessageCount} unread chat messages`
            : ''
        }
      />
```

lunars97 (changes requested):
Looks really nice 💛  I tested in Safari and Chrome. What lacks is live announcement for screen readers. I left a suggestion above :)

web/src/components/ChatFab.tsx
bahaaTuffaha: Yes great idea... I will try it 👍

translations/translations.json
```diff
@@ -416,7 +417,8 @@
       "newChat": "Start new chat",
       "newChatConfirmation": "Do you really want to start a new chat?",
       "newChatConfirmationMessage": "The current chat history will be lost.",
-      "chatOptions": "Chat options"
+      "chatOptions": "Chat options",
+      "unreadMessages": "{{count}} unread chat messages"
```
lunars97:
```suggestion
      "unreadMessages": "{{count}} unread chat message"
```

translations/translations.json
```diff
@@ -320,7 +320,8 @@
       "newChat": "Neuen Chat starten",
       "newChatConfirmation": "Möchtest du wirklich einen neuen Chat starten?",
       "newChatConfirmationMessage": "Der aktuelle Chatverlauf wird gelöscht.",
-      "chatOptions": "Chat-Optionen"
+      "chatOptions": "Chat-Optionen",
+      "unreadMessages": "{{count}} ungelesene Chat-Nachrichten"
```
lunars97:
```suggestion
      "unreadMessages": "{{count}} ungelesene Chat-Nachricht"
```

lunars97 (changes requested):
LGTM except for plural form of the chat message :)

translations/translations.json
steffenkleinle: Good catch! However, I think we need to make use of the i18next plural feature, see e.g. `showPlaces_zero` and so on :)

lunars97: doesn't it anyway return one message per question? 🤔

lunars97: but to make it future-proof, it would be nice to have it 👍

steffenkleinle: You could ask multiple questions in a row without waiting for an answer. Also, a human advisor could answer in multiple messages.

bahaaTuffaha: Nice find! I forgot about the `showPlaces_zero`

translations/translations.json
```diff
@@ -320,7 +320,13 @@
       "newChat": "Neuen Chat starten",
       "newChatConfirmation": "Möchtest du wirklich einen neuen Chat starten?",
       "newChatConfirmationMessage": "Der aktuelle Chatverlauf wird gelöscht.",
-      "chatOptions": "Chat-Optionen"
+      "chatOptions": "Chat-Optionen",
+      "unreadMessages_zero": "0 ungelesene Chat-Nachrichten",
```
lunars97: sorry again 🙈 but do we need zero here? 👀

bahaaTuffaha: No I don't think so. I will remove it

lunars97 (approved):
Thank you! Looks really good 👍  Not tested again

PR #4175, opened 2026-06-13
2404: Fix circleci dist folder build

lunars97:
### Short Description

Fixes the error where the file is not found in dist folder of the packages by circleci caused by #2404 
### Proposed Changes

- Add ts:check to circleci builds to generate dist folder for workspaces

### Side Effects

- none

### Testing

N/A

### Resolved Issues

Fixes: #4093 

---

### DIFF
--- a/.circleci/src/jobs/build_android.yml
+++ b/.circleci/src/jobs/build_android.yml
@@ -30,7 +30,7 @@ steps:
       command: bundle exec fastlane android keystore
       working_directory: native/android
   - run:
-      name: Check TypeScript
+      name: Build files
       command: yarn ts:check
   - run:
       name: '[FL] << parameters.build_config_name >> Android Build'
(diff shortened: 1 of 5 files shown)


.circleci/src/jobs/build_android.yml
```diff
@@ -29,6 +29,9 @@ steps:
       name: '[FL] Prepare Android Keystore'
       command: bundle exec fastlane android keystore
       working_directory: native/android
+  - run:
+      name: Check TypeScript
```
steffenkleinle: 🔧 I'd rename this here to `Build Files` or similar, same for other usages. Otherwise people might be tempted in the future to remove this as technically its duplicated.

steffenkleinle (commented):
❓ Aren't we missing `e2e_web` as we need to run the bundler there as well which requires the files to be built?

lunars97:
> ❓ Aren't we missing `e2e_web` as we need to run the bundler there as well which requires the files to be built?

Good catch :+1:

steffenkleinle (approved):

bahaaTuffaha (approved):
Not tested

PR #4181, opened 2026-06-16
2404: Fix concurrent upstream jobs persisted the same file(s)

lunars97:
### Short Description

It might fix this error `Concurrent upstream jobs persisted the same file(s)` https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/16693/workflows/32152574-84c1-461b-b47b-8ad79e055603/jobs/76881  by creating one job `build_files` 

### Proposed Changes

- Create `build_files` job, instead of running it for every build, maybe we could run it once? :thought_balloon: 

### Side Effects

I hope none :see_no_evil: 

### Checklist

N/A

### Testing

N/A

### Resolved Issues

Fixes: #4093 

---

### DIFF
--- a/.circleci/src/jobs/build_web.yml
+++ b/.circleci/src/jobs/build_web.yml
@@ -15,9 +15,6 @@ steps:
   - prepare_workspace
   - restore_environment_variables
   - run:
-      name: Build files
-      command: yarn ts:check
-  - run:
       name: << parameters.build_config_name >> build
       command: yarn workspace web build:<< parameters.build_config_name >> --env commit_sha=${CIRCLE_SHA1} --env version_name=${NEW_VERSION_NAME}
   - store_artifacts:
--- a/.circleci/src/jobs/build_web.yml
+++ b/.circleci/src/jobs/build_web.yml
@@ -32,5 +32,5 @@ steps:
   - persist_to_workspace:
       root: web
       paths:
-        - dist
+        - dist/<<parameters.build_config_name >>
   - notify
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -427,7 +427,7 @@ jobs:
                 path: web/dist/<< parameters.build_config_name >>.tar.gz
             - persist_to_workspace:
                 paths:
-                    - dist
+                    - dist/<<parameters.build_config_name >>
                 root: web
             - notify
     bump_version:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,6 +191,9 @@ commands:
             - run:
                 command: corepack yarn --immutable
                 name: Yarn
+            - run:
+                command: corepack yarn ts:check
+                name: Build files
             - save_cache:
                 key: 11-yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum "package.json" }}-{{ checksum "native/package.json" }}-{{ checksum "web/package.json" }}
                 name: Save Yarn Package Cache
@@ -265,9 +268,6 @@ jobs:
                 command: bundle exec fastlane android keystore
                 name: '[FL] Prepare Android Keystore'
                 working_directory: native/android
-            - run:
-                command: yarn ts:check
-                name: Build files
             - run:
                 command: bundle exec fastlane android build build_config_name:<< parameters.build_config_name >> version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
                 name: '[FL] << parameters.build_config_name >> Android Build'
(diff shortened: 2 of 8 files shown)


.circleci/src/jobs/build_files.yml
```diff
@@ -0,0 +1,16 @@
...
+      name: Build files
+      command: yarn ts:check:ci
+  - persist_to_workspace:
+      root: .
+      paths:
+        - build-configs/dist
+        - shared/dist
+        - translations/dist
+        - e2e-tests/dist
+        - native/dist
```
lunars97: I did not add for web while in build_web it already persists to workspace

lunars97:
Seems like it does not fix it, but break it and it does not build dist files :( https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app?branch=fix-concurrent-upstream-jobs-persistent-same-file

.circleci/src/jobs/build_files.yml
bahaaTuffaha, line 1: it's not used in any workflow 🤔

lunars97: I added it to e2e_tests workflow now :+1:

.circleci/src/jobs/build_web.yml
```diff
@@ -16,7 +16,11 @@ steps:
   - restore_environment_variables
   - run:
       name: Build files
-      command: yarn ts:check
+      command: yarn ts:check:ci
+  - persist_to_workspace:
+      root: web
+      paths:
+        - dist/<<parameters.build_config_name >>
```
steffenkleinle: ❌ There is already a persist job (see L39), no need to introduce a new one. This would not work in any case since the artifact is only available after building (L24ff).

steffenkleinle (changes requested):

steffenkleinle (commented):
Looks good, thanks for addressing this! Shouldn't we use `yarn ts:check:ci` everywhere consistently?

lunars97:
> Looks good, thanks for addressing this! Shouldn't we use `yarn ts:check:ci` everywhere consistently?

It leads to `error This project's package.json defines "packageManager": "yarn@4.13.0". However the current global version of Yarn is 1.22.22.` I think this is because of `ts:check:ci`, there was no such error before :thinking:

steffenkleinle:
> > Looks good, thanks for addressing this! Shouldn't we use `yarn ts:check:ci` everywhere consistently?
> 
> It leads to `error This project's package.json defines "packageManager": "yarn@4.13.0". However the current global version of Yarn is 1.22.22.` I think this is because of `ts:check:ci`, there was no such error before 🤔

Where did this happen? Can you send me the CI job pls? Maybe we should then just use the normal `yarn ts:check` everywhere and just remove the command from `package.json`?

lunars97:
> > > Looks good, thanks for addressing this! Shouldn't we use `yarn ts:check:ci` everywhere consistently?
> > 
> > 
> > It leads to `error This project's package.json defines "packageManager": "yarn@4.13.0". However the current global version of Yarn is 1.22.22.` I think this is because of `ts:check:ci`, there was no such error before 🤔
> 
> Where did this happen? Can you send me the CI job pls? Maybe we should then just use the normal `yarn ts:check` everywhere and just remove the command from `package.json`?

when I was testing and triggered the e2e_tests pipeline, it failed for build_ios with the error above, I will trigger it again to test this one
UPDATE: It happened again, look at the [build_ios fail](https://app.circleci.com/pipelines/github/digitalfabrik?created_after=2026-06-03T00%3A00%3A00.000Z&created_preset=Last+2+weeks&selectedPipeline=gh%2Fdigitalfabrik%2Fintegreat-app%2F16730&job=14e11295-4c5f-44a0-8050-371919a5edbc&buildNumber=76966&jobType=build&workflowId=b078cddc-2fac-4acf-bbb2-562810530176) I guess there should be extra `enable corepack `command before building the files? or just use normal one. But `restore_yarn_cache `and `restore_yarn_tools_cache` are doing it already :thinking:

lunars97:
> > > Looks good, thanks for addressing this! Shouldn't we use `yarn ts:check:ci` everywhere consistently?
> > 
> > 
> > It leads to `error This project's package.json defines "packageManager": "yarn@4.13.0". However the current global version of Yarn is 1.22.22.` I think this is because of `ts:check:ci`, there was no such error before 🤔
> 
> Where did this happen? Can you send me the CI job pls? Maybe we should then just use the normal `yarn ts:check` everywhere and just remove the command from `package.json`?

In check.yml there is yarn `ts:check:ci ` as well :eyes:

steffenkleinle:
@lunars97 I took the liberty to just fix the problem. I added the build files step to `restore_yarn_cache` since that is directly executed everywhere we need it. Also, I used `corepack yarn ts:check` instead of `yarn ts:check` to fix the problem with `build_ios` on mac.

steffenkleinle (approved):

bahaaTuffaha (approved):

Issue #4166, opened 2026-06-09
Switch from webpack to vite


steffenkleinle:
### Describe the Problem

Our current bundler webpack is quite slow and not really state of the art anymore. Our config is complicated and most likely outdated. As a dev, this is a suboptimal experience.

### Describe Your Preferred Solution

[Vite](https://vite.dev) currently seems to be the de-facto standard choice in modern web development and allows for easy switch to server-side rendering (SSR) and integrates well with react-router.
Therefore switch from webpack to vite.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

https://vite.dev

lunars97:
won't #2404 block it for now? 🤔

andrew8er:
@lunars97 
I can base my PR on top of yours, if it is somewhat ready. But #4093 needs to be rebased before that.

lunars97:
> [@lunars97](https://github.com/lunars97) I can base my PR on top of yours, if it is somewhat ready.

I need to do some adjustments there. Yes, I will rebase it

steffenkleinle:
Since this is not urgent, we could also just postpone this until we are done with ESM. Personally I don't mind, so up to you :)

PR #4184, opened 2026-06-16
4149: Improve back navigation from the regions page

### LINKED ISSUE #4149: Improve navigation back from the region selection page
### Describe the Problem

Observed during user testing on Android (WebApp)

While reading content, a participant clicked the chevron next to the current city name, expecting a dropdown menu to expand. Instead, the app navigated to the region selection page.

As a result, the participant lost the context of the page they were reading and did not know how to return. No obvious back navigation was visible on the screen.

Add a visible "Back to content" action on the region selection page (mobile only).

This would provide users with a clear recovery path when they unintentionally open the region selection flow and help them return to the page they were previously reading.

N/A

### Related Issues

N/A

### Additional Information

After being redirected to the region selection page, the participant was unable to find a clear way back to the previously viewed content. The device was using gesture navigation, meaning no visible browser back button was present on screen. 
Although Android's system back gesture could have been used to return to the previous page, the participant did not attempt this because they expected an in-app navigation pattern and did not realize that browser/system navigation was required to recover from the action.

bahaaTuffaha:
### Short Description

Add a visible "Back to content" action on the region selection page (mobile only).

This would provide users with a clear recovery path when they unintentionally open the region selection flow and help them return to the page they were previously reading.

### Proposed Changes

- Created new component just for BackToContentButton.
- `window.history.state?.idx` reads React Router's current position in the browser history stack.
- edge-case: If you changed the language while at region page the last history page becomes the language change so to ignore this page: we save an entryIndex and subtract it from the history index so we remove any history added while we are still on the same page.
- Created a small test for BackToContentButton.
- While running jest tests on VsCode i encountered some weird issues on the IDE but it's passing the test on cli's `yarn test` (It assumes one Jest at the workspace root) so I added `jest.virtualFolders` to vsCode settings.

### Side Effects

N/A

### Checklist

### Testing

- Go to categories then click on the changeRegion button while testing on small screen.
- Test the back button on large screens.
- Test the back button after language change.
- Test contrast mode.

### Resolved Issues

Fixes: #4149

---


web/src/components/BackToContentButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+  if (!mobile || entryIndex === 0) {
+    return null
+  }
+
+  return (
+    <StyledButton
+      onClick={() => navigate(entryIndex - historyIndex() - 1)}
+      startIcon={<DirectionDependentBackIcon />}
+      color='inherit'
+      aria-label={t('backToContent')}>
```
steffenkleinle: 🤔 I think we shouldn't set an aria-label if the text content is the same.

web/src/components/BackToContentButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const historyIndex = (): number => window.history.state?.idx ?? 0
+
+const BackToContentButton = (): ReactElement | null => {
+  const navigate = useNavigate()
+  const { mobile } = useDimensions()
+  const { t } = useTranslation('layout')
+
+  // Index of the content page we arrived from so that we can navigate back even if the language changed.
+  const entryIndex = useRef(historyIndex()).current
```
steffenkleinle: 🔧 We usually use `useState` to persist values, any reason you chose to use a `useRef` here?

```suggestion
  // Initial history index to account for language changes or other user interactions on this page
  const [initialHistoryIndex] = useState(historyIndex)
```

web/src/components/BackToContentButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import useDimensions from '../hooks/useDimensions'
+import { DirectionDependentBackIcon } from './base/Dialog'
+
+const StyledButton = styled(Button)({
+  textTransform: 'none',
+  alignSelf: 'flex-start',
+})
+
+const historyIndex = (): number => window.history.state?.idx ?? 0
```
steffenkleinle: 🙃 No need for this helper (we can just use a const in the component) and naming with `currentHistoryIndex` would be a little clearer:
``` suggestion
  const currentHistoryIndex = window.history.state?.idx ?? 0
```

web/src/components/BackToContentButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const BackToContentButton = (): ReactElement | null => {
```
steffenkleinle: 🔧 Let's use a consistent internal naming:
```suggestion
const BackToRegionButton = (): ReactElement | null => {
```

translations/translations.json
```diff
@@ -7748,7 +7738,8 @@
       "skipToContent": "Zum Inhalt springen",
       "shareQrCodeTitle": "Über QR-Code teilen",
       "shareQrCodeDescription": "Teile diese Seite einfach, indem du den QR-Code auf einem anderen Gerät scannst.",
-      "qrCode": "QR-Code"
+      "qrCode": "QR-Code",
+      "backToContent": "Zurück zum Inhalt"
```
steffenkleinle: @ulyanahoy-cloud is Back to content/Zurück zum Inhalt necessary or would a simple Go Back/Zurück be enough? That would save us another translation :)

steffenkleinle (approved):
Nice work! Works as expected :)

web/src/components/BackToContentButton.tsx
bahaaTuffaha: Because it's set once and it doesn't trigger a re-render..

steffenkleinle: Using a useState instead also doesn't trigger a re-render by itself. There is only a difference if you want to use it inside a `useEffect` or `useCallback`.

web/src/components/BackToContentButton.tsx
bahaaTuffaha: what if we want to reuse it in the future?

steffenkleinle: Then we should just name it `BackButton` since `ToContent` is also quite specific already. But I think this could be a future problem.

web/src/components/__tests__/BackToRegionButton.spec.tsx
steffenkleinle, line 1: 🔧 The name of the test file is still wrong

steffenkleinle (approved):

bahaaTuffaha:
@steffenkleinle I noticed that I need to run `yarn ts:check` sometimes  to get the new translations working. did we change something in that regard ?

steffenkleinle:
> @steffenkleinle I noticed that I need to run `yarn ts:check` sometimes to get the new translations working. did we change something in that regard ?

I think it is always necessary due to the switch to esm in #2404

bahaaTuffaha:
> > @steffenkleinle I noticed that I need to run `yarn ts:check` sometimes to get the new translations working. did we change something in that regard ?
> 
> I think it is always necessary due to the switch to esm in #2404

How about we run it once after a `yarn install` or before `yarn start`?

steffenkleinle:
> > > @steffenkleinle I noticed that I need to run `yarn ts:check` sometimes to get the new translations working. did we change something in that regard ?
> > 
> > 
> > I think it is always necessary due to the switch to esm in #2404
> 
> How about we run it once after a `yarn install` or before `yarn start`?

We could do that, but I think it would not fix your problem. You still need to run it manually after adjusting the translations.

lunars97:
> > > > @steffenkleinle I noticed that I need to run `yarn ts:check` sometimes to get the new translations working. did we change something in that regard ?
> > > 
> > > 
> > > I think it is always necessary due to the switch to esm in #2404
> > 
> > 
> > How about we run it once after a `yarn install` or before `yarn start`?
> 
> We could do that, but I think it would not fix your problem. You still need to run it manually after adjusting the translations.

@andrew8er mentioned it before here https://github.com/digitalfabrik/integreat-app/pull/4093#issuecomment-4690032244 and he said that he will implement this step as well in #4166 if I understood correctly

andrew8er:
> @andrew8er mentioned it before here https://github.com/digitalfabrik/integreat-app/pull/4093#issuecomment-4690032244 and he said that he will implement this step as well in https://github.com/digitalfabrik/integreat-app/issues/4166 if I understood correctly

I'll see how I can implement that. What needs to be run for the translations?

bahaaTuffaha:
> > @andrew8er mentioned it before here [#4093 (comment)](https://github.com/digitalfabrik/integreat-app/pull/4093#issuecomment-4690032244) and he said that he will implement this step as well in #4166 if I understood correctly
> 
> I'll see how I can implement that. What needs to be run for the translations?

`yarn ts:check`

translations/translations.json
steffenkleinle: @bahaaTuffaha after talking to Ulyana: Just `Go back`/`Zurück` is enough, so you can reuse `common.back`.

lunars97 (approved):
LGTM!

PR #4205, opened 2026-06-22
2986: Improve events sorting

### LINKED ISSUE #2986: Improve Sorting of Events
the current sorting system for events causes long-term (ongoing) events to appear prominently at the top of the list. This reduces the visibility of single (one-time) events, which may be less noticeable to users. As a result, users may miss out on important individual events, leading to a less optimal user experience

An improved sorting mechanism where single (one-time) events are prioritized and displayed at the top of the feed, followed by long-term (ongoing) events. This adjustment would ensure better visibility for time-specific events, making it easier for users to discover and engage with them

An alternative solution could involve visually differentiating long-term events with distinct labels, making them easier to distinguish. However, this approach might not fully resolve the issue of reduced visibility for single events

**Additional context**
The implementation is dependent on the backend release of the option to categorize events in single events and ongoing events (📆 [#3070](https://github.com/digitalfabrik/integreat-cms/issues/3070))

lunars97:
### Short Description

This PR implements events sorting by grouping them by date (today, tomorrow, this week, this month, further). Inside each group one-time events are shown first and recurring after them. Each group has corresponding tittle like `today, tomorrow, this week, this month, further`. I did not implement the logic on filtered events because it makes no sense while grouping is related to today, filtered events are not grouped.

### Proposed Changes

- Remove icons for today and for today recurring
- Create dedicated shared helper for sorting events `events.ts`
- Move filteringEventsLogic.ts into events.ts
- Create separate EventList.tsx component
- Use SectionList in native instead of FlatList in native as it is semantically close to date grouping
- Adjust each component accordingly
- Add new constants to translation.json
- Add/Adjust test cases

### Side Effects

N/A

### Checklist

### Testing

- Go to events
- See whether events grouped accordingly
- Also check by filtering events (there should be no grouping)

### Resolved Issues

Fixes: #2986 

---


native/src/components/EventListItem.tsx
```diff
@@ -109,8 +75,8 @@ const EventListItem = ({
           </Text>
         </View>
       }
-      left={isRtl ? DateIcon : renderThumbnail}
-      right={isRtl ? renderThumbnail : DateIcon}
+      left={isRtl ? undefined : renderThumbnail}
+      right={isRtl ? renderThumbnail : undefined}
```
lunars97: I guess we dont need these? 🤔

steffenkleinle: ❌ The recurring icon is supposed to stay,  only today and today-recurring should be removed 🙈

bahaaTuffaha:
Does the filter changes part of this ticket? or turned into grouping feature?

translations/translations.json
```diff
@@ -4078,9 +4066,11 @@
       "chooseCalendar": "Calendar choice",
       "added": "Added",
       "goToCalendar": "Go to calendar",
-      "recurring": "recurring",
-      "today": "today",
-      "todayRecurring": "today, recurring",
+      "today": "Today",
+      "tomorrow": "Tomorrow",
+      "thisWeek": "This week",
+      "thisMonth": "Next month",
```
bahaaTuffaha:
```suggestion
      "thisMonth": "this month",
```

web/src/routes/EventsPage.tsx
```diff
@@ -143,15 +152,51 @@ const EventsPage = ({ region, pathname, languageCode, regionCode }: RegionRouteP
...
+          filterStartDate={startDate}
+          filterEndDate={endDate}
+        />
+      ))}
+      noItemsMessage='events:currentlyNoEvents'
     />
-  ))
+  )
+
+  const groupedList = GROUP_ORDER.some(key => groupedEvents[key].length > 0)
```
bahaaTuffaha: I think this needs a better naming because its boolean.

native/src/routes/Events.tsx
```diff
@@ -42,6 +55,19 @@ const Events = ({ regionModel, language, navigateTo, events, slug, refresh }: Ev
   const event = events.find(it => it.slug === slug)
   useTtsPlayer(event)
 
+  const groupedEvents = groupEvents(events)
+
+  const items: (string | EventModel)[] = useMemo(() => {
```
bahaaTuffaha:
```suggestion
  const items: EventListEntry[] = useMemo(() => {
```

native/src/routes/Events.tsx
```diff
@@ -42,6 +55,19 @@ const Events = ({ regionModel, language, navigateTo, events, slug, refresh }: Ev
   const event = events.find(it => it.slug === slug)
   useTtsPlayer(event)
 
+  const groupedEvents = groupEvents(events)
```
bahaaTuffaha: 🤔  I think its a good idea to move this into the useMemo so it wont run on every render.

bahaaTuffaha (commented):
❓  Is it possible to keep the grouping while filtering or this is how it's meant to behave?

native/src/routes/Events.tsx
steffenkleinle: https://react.dev/reference/react/useMemo#how-to-tell-if-a-calculation-is-expensive

steffenkleinle: > In general, unless you’re creating or looping over thousands of objects, it’s probably not expensive.

`useMemo` should only be used if it is expensive which I think is not the case here :)

bahaaTuffaha: I mean the one used under L60 not in a new one

lunars97:
> ❓ Is it possible to keep the grouping while filtering or this is how it's meant to behave?

I guess if we keep filtering, then it might be a bit confusing because if a user inserts date for several months ahead then there will be only `further` group, so imho it makes no sense to do it for filtering :thought_balloon:

native/src/routes/Events.tsx
lunars97: there is already memoization in L60 👀

translations/translations.json
lunars97: nice catch! :+1:

native/src/components/EventListItem.tsx
```diff
@@ -31,15 +26,10 @@ const styles = StyleSheet.create({
 const placeholderThumbnails = [EventThumbnailPlaceholder1, EventThumbnailPlaceholder2, EventThumbnailPlaceholder3]
 
 const getDateIcon = (date: DateModel): { icon: string | JSXElementConstructor<SvgProps>; label: string } | null => {
```
steffenkleinle: 🔧 This method is now redundant, please remove it. Same for web. Also check `date.getDateIcon()` if it is still required/should be adjusted.

shared/api/models/DateModel.ts
```diff
@@ -192,17 +192,9 @@ class DateModel {
 
   getDateIcon(): { icon: DateIcon; label: string } | null {
```
steffenkleinle: 🔧 I think this method is now redundant and does not really make sense anymore since we need to map the icon again, both for native and web. Let's remove it.

shared/utils/groupEventsByDate.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import EventModel from '../api/models/EventModel.js'
+
+const WEEK = 7
+
+export type DateGroupKey = 'today' | 'tomorrow' | 'thisWeek' | 'thisMonth' | 'further'
+
+export const GROUP_ORDER: DateGroupKey[] = ['today', 'tomorrow', 'thisWeek', 'thisMonth', 'further']
+
+const isRecurring = (event: EventModel): boolean => event.date.hasMoreRecurrencesThan(1)
```
steffenkleinle: 🔧 Can be moved to be a direct method/getter of `EventModel`.

release-notes/unreleased/2986-improve-events-sorting.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: 2986
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: Events are now sorted according to the group they belong to.
+de: Die Veranstaltungen sind nun nach der Gruppe sortiert, zu der sie gehören.
```
steffenkleinle:
```suggestion
en: Events are now grouped by their next recurrence.
de: Veranstaltungen werden nun nach ihrer nächsten Wiederholung gruppiert.
```

shared/utils/groupEventsByDate.ts
```diff
@@ -0,0 +1,68 @@
...
+  const today = now.startOf('day')
+  const startDay = start.startOf('day')
+  const diffDays = Math.round(startDay.diff(today, 'days').days)
+
+  if (diffDays <= WEEK) {
+    return 'thisWeek'
+  }
+
+  const endOfMonth = now.endOf('month').startOf('day')
+  if (startDay <= endOfMonth) {
```
steffenkleinle: ❌ This is the literal calendar month, so e.g. end of June now. In the issue, this group was intended to contain all events in the next 30 days.
```suggestion
  if (event.date.startDate <= now.plus({ days: 30 }).endOf('day')) {
```

shared/utils/groupEventsByDate.ts
```diff
@@ -0,0 +1,68 @@
...
+  return 'further'
+}
+
+const sortGroup = (events: EventModel[]): EventModel[] => {
+  const byDate = (a: EventModel, b: EventModel): number => a.date.startDate.toMillis() - b.date.startDate.toMillis()
+
+  const oneTimeEvent = events.filter(event => !isRecurring(event)).sort(byDate)
+  const recurringEvent = events.filter(event => isRecurring(event)).sort(byDate)
+
+  return [...oneTimeEvent, ...recurringEvent]
```
steffenkleinle: 🔧 There is no filtering necessary here, this can be merged into a single sort call. When comparing two events, first check if only one event is recurring and return an according integer and only sort by `startDate` otherwise.

shared/utils/groupEventsByDate.ts
```diff
@@ -0,0 +1,68 @@
...
+    return 'thisWeek'
+  }
+
+  const endOfMonth = now.endOf('month').startOf('day')
+  if (startDay <= endOfMonth) {
+    return 'thisMonth'
+  }
+
+  return 'further'
+}
```
steffenkleinle: 🔧 What do you think of
```suggestion
const eventGroups: Record<Exclude<DateGroupKey, 'further'>, number> = {
  today: 0,
  tomorrow: 1,
  thisWeek: 7,
  thisMonth: 30,
}

const getGroupKey = (event: EventModel): DateGroupKey => {
  const now = DateTime.now()
  const labels = Object.keys(groups) as Exclude<DateGroupKey, 'further'>[]
  return labels.find(label => event.date.startDate <= now.plus({ days: groups[label] }).endOf('day')) ?? 'further'
}
```

native/src/routes/Events.tsx
```diff
@@ -27,6 +30,16 @@ const PageDetailsContainer = styled.View`
...
+type EventListEntry = string | EventModel
+
+const EventSeparator = ({
+  leadingItem,
+  trailingItem,
+}: {
+  leadingItem: EventListEntry
+  trailingItem: EventListEntry
+}) => (typeof leadingItem === 'string' || typeof trailingItem === 'string' ? null : <Divider />)
+
```
steffenkleinle: 🔧 I think we should move all this logic to a separate `EventList` component that handles the filtering, grouping and sorting. Same for web.

native/src/routes/Events.tsx
```diff
@@ -104,7 +130,15 @@ const Events = ({ regionModel, language, navigateTo, events, slug, refresh }: Ev
     return <Failure code={fromError(error)} retry={refresh} />
   }
 
-  const renderEventListItem = ({ item }: { item: EventModel }) => {
+  const renderEventListItem = ({ item }: { item: string | EventModel }) => {
+    if (typeof item === 'string') {
+      return (
+        <Text variant='body2' style={{ paddingTop: 8, paddingRight: 8, textAlign: contentAlignment(language) }}>
```
steffenkleinle: 🙃 Is `paddingRight` actually necessary?
```suggestion
        <Text variant='body2' style={{ paddingTop: 8, textAlign: contentAlignment(language) }}>
```

native/src/routes/Events.tsx
```diff
@@ -42,6 +55,19 @@ const Events = ({ regionModel, language, navigateTo, events, slug, refresh }: Ev
   const event = events.find(it => it.slug === slug)
   useTtsPlayer(event)
 
+  const groupedEvents = groupEvents(events)
+
+  const items: EventListEntry[] = useMemo(() => {
```
steffenkleinle: ❓ Is this `useMemo` actually necessary?

native/src/routes/Events.tsx
```diff
@@ -42,6 +55,19 @@ const Events = ({ regionModel, language, navigateTo, events, slug, refresh }: Ev
...
+    if (startDate || endDate) {
+      return filteredEvents ?? []
+    }
+
+    return GROUP_ORDER.flatMap(key => {
+      const bucket = groupedEvents[key]
+      return bucket.length > 0 ? [t(key), ...bucket] : []
+    })
```
steffenkleinle: 🔧 Let's use a consistent and meaningful naming:
```suggestion
    return EVENT_DATE_GROUPS.flatMap(key => {
      const dateGroup = groupedEvents[key]
      return dateGroup.length > 0 ? [t(key), ...dateGroup] : []
    })
```

shared/utils/groupEventsByDate.ts
```diff
@@ -0,0 +1,68 @@
...
+  events.forEach(event => {
+    groups[getGroupKey(event)].push(event)
+  })
+
+  GROUP_ORDER.forEach(key => {
+    groups[key] = sortGroup(groups[key])
+  })
+
+  return groups
+}
```
steffenkleinle: 🔧 
```suggestion
export const groupEventsByDate = (events: EventModel[]): Record<DateGroupKey, EventModel[]> =>
  Object.fromEntries(
    GROUP_ORDER.map(key => [key, sortGroup(events.filter(event => getGroupKey(event) === key))]),
  ) as Record<DateGroupKey, EventModel[]>
```

shared/utils/groupEventsByDate.ts
steffenkleinle, line 1: 🔧 I'd keep this file generic, i.e. `events.ts` so we can avoid scattered logic.

web/src/routes/EventsPage.tsx
```diff
@@ -143,15 +152,51 @@ const EventsPage = ({ region, pathname, languageCode, regionCode }: RegionRouteP
...
+  const isGroupedList = GROUP_ORDER.some(key => groupedEvents[key].length > 0)
+
+  let eventsList
+  if (startDate || endDate) {
+    eventsList = filteredListItems
+  } else if (isGroupedList) {
+    eventsList = groupedListSections
+  } else {
+    eventsList = <List items={[]} noItemsMessage='events:currentlyNoEvents' />
+  }
```
steffenkleinle: 🔧 Same as for native: Let's move this to a separate `EventList` component

web/src/routes/EventsPage.tsx
```diff
@@ -143,15 +152,51 @@ const EventsPage = ({ region, pathname, languageCode, regionCode }: RegionRouteP
...
-    <EventListItem
-      event={event}
-      languageCode={languageCode}
-      key={event.path}
-      filterStartDate={startDate}
-      filterEndDate={endDate}
+  const groupedListSections = GROUP_ORDER.map((key: DateGroupKey) => {
+    const eventsGroup = groupedEvents[key]
+    if (eventsGroup.length === 0) {
+      return []
```
steffenkleinle: 🔧 
```suggestion
      return null
```

web/src/components/__tests__/EventListItem.spec.tsx
```diff
@@ -57,61 +51,60 @@ describe('EventListItem', () => {
...
-      const { queryByText, getByText } = renderWithRouterAndTheme(
-        <EventListItem event={event} languageCode={language} />,
-      )
-
-      expect(queryByText('events:todayRecurring')).toBeFalsy()
-      expect(getByText('events:recurring')).toBeTruthy()
-      expect(queryByText('events:today')).toBeFalsy()
+    it('should put an event 3 weeks after into this month group', () => {
+      const event = createEvent('2023-10-25T15:00:00.000+02:00')
+      expect(getGroupKey(event)).toBe('thisMonth')
```
steffenkleinle: 🤔 Hmm, these tests seem to be in the wrong place, imo this should be part of a new test suite for the shared utils? Same for native.

steffenkleinle (changes requested):
Thank you! This looks very good already, but I think we could still simplify a few utils and extract the list/grouping logic to own components.
Furthermore, the divider logic is inconsistent between web and native (personally I think we should go with web (no divider):
native (divider): 

Image

web (no divider): 

Image

shared/utils/groupEventsByDate.ts
lunars97: looks way better and also less code :+1:

native/src/routes/Events.tsx
lunars97: this was for rtl :see_no_evil: ,forgot to adjust it properly.  I think it is also fine without it

lunars97:
@ulyanahoy-cloud what do you think about these grouping look? Currently it looks like this:
web :point_down: 
image
native: :point_down: 
image

ulyanahoy-cloud:
Looks good!
You can also filter by these groups, right?

steffenkleinle:
> Looks good! You can also filter by these groups, right?

Not yet, nope. Wasn't a requirement so far and if that is actually necessary, it should be done in a separate issue. But I am not sure if that is actually necessary?

ulyanahoy-cloud:
In that case, I think it works for now.

steffenkleinle:
> > Looks good! You can also filter by these groups, right?
> 
> Not yet, nope. Wasn't a requirement so far and if that is actually necessary, it should be done in a separate issue. But I am not sure if that is actually necessary?

@ulyanahoy-cloud adding to that, it is currently already possible to filter by date, so I think filtering by those group would be very duplicating and redundant imo.

native/src/components/List.tsx
```diff
@@ -30,6 +30,7 @@ const List = <T,>({
   scrollEnabled,
   style,
   keyboardShouldPersistTaps = 'never',
+  itemSeparatorComponent,
```
steffenkleinle: 🤔 What do you think of a default:
```suggestion
  itemSeparatorComponent = Divider,
```

shared/utils/events.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const getGroupKey = (event: EventModel): DateGroupKey => {
+  const now = DateTime.now()
+  const labels = Object.keys(eventGroups) as Exclude<DateGroupKey, 'further'>[]
+  return labels.find(label => event.date.startDate <= now.plus({ days: eventGroups[label] }).endOf('day')) ?? 'further'
+}
+
+const sortGroup = (events: EventModel[]): EventModel[] =>
+  [...events].sort((a, b) => {
```
steffenkleinle: 🤔 Why is the destructuring necessary?
```suggestion
  events.sort((a, b) => {
```

native/src/components/EventList.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const EventList = ({ events, regionModel, language, navigateTo, refresh }: EventListProps): ReactElement => {
+  const { t } = useTranslation('events')
+  const { startDate, setStartDate, endDate, setEndDate, filteredEvents, startDateError } = useDateFilter(events)
+
+  const groupedEvents = groupEventsByDate(events)
+
+  const items =
+    startDate || endDate
+      ? (filteredEvents ?? [])
```
steffenkleinle: 💭 Optional since not done in this PR, but if we would adjust `useDateFilter` and `filteringEventsLogic` to only accept and return non-nullable events (i.e. `EventModel[]` instead of `EventModel[] | null`), these `?? []` would not be necessary. Would you mind adjusting it?

shared/src/utils/events.ts
steffenkleinle, line 1: 💭 Thank you for renaming this! Optional since it wasn't done in your PR, but would you mind moving the `filteringEventsLogic` function here as well and rename it to `filterEvents`? Thanks ❤️

web/src/components/EventList.tsx
```diff
@@ -0,0 +1,83 @@
...
+        startDate={startDate}
+        setStartDate={setStartDate}
+        endDate={endDate}
+        setEndDate={setEndDate}
+        startDateError={startDateError}
+      />
+      {eventsList}
+    </>
+  )
+}
```
steffenkleinle: 🔧 Thanks for moving this to an own component, I think this is a great refactoring. I have one more suggestion though to make it more readable and cluster related logic together:
What do you think of leveraging early returns, such as we do in `EventsPage.tsx` with the checks for error, eventId, etc? This would avoid executing unnecessary filtering/grouping while making the code more structured/readable.
So something like this:
``` tsx
  const dateFilter = <EventsDateFilter ... />

  if (startDate || endDate) {
    const filteredListItems = ...
    return <>{dateFilter}{filteredListItems}</>
  }

  ...
```

web/src/components/EventList.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return (
+      <section key={key}>
+        <Typography component='span' variant='body1' dir='auto'>
+          {t(key)}
+        </Typography>
+        <List
+          items={dateGroup.map(event => (
+            <EventListItem event={event} languageCode={languageCode} key={event.path} />
+          ))}
+        />
```
steffenkleinle, line 77: ❌ I think it is not recommended to use multiple lists with items of the same type in terms of semantic html/a11y. Instead, have a look at `RegionSelector.tsx` and `RegionListGroup`, I think that pattern should be usable pretty much analogous here.

native/src/components/EventList.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type EventListProps = {
+  events: EventModel[]
+  regionModel: RegionModel
+  language: string
+  navigateTo: (routeInformation: RouteInformationType) => void
+  refresh: () => void
+}
+
+const EventList = ({ events, regionModel, language, navigateTo, refresh }: EventListProps): ReactElement => {
```
steffenkleinle, line 34: ❓ I think similarly to web, we should use the corresponding react-native-paper components for sections and subheaders. See e.g. https://oss.callstack.com/react-native-paper/docs/components/List/ListSection and https://oss.callstack.com/react-native-paper/docs/components/List/ListSubheader.

steffenkleinle (changes requested):
Looks a lot cleaner already! Nice work :)

native/src/components/EventList.tsx
lunars97: I did not check it, nice find! 👍

native/src/components/List.tsx
lunars97: it will be applied everywhere...

native/src/components/EventList.tsx
lunars97: what do you think of using SectionList instead of FlatList, we could use it only for EventList? :thinking: I think semantically it is more suitable... See the example https://reactnative.dev/docs/sectionlist#example

steffenkleinle: We should use the components of our component library, i.e. https://oss.callstack.com/react-native-paper/docs/components/List/ListSection. If that makes sense together with react-natives SectionList, yes. But if I am not mistaken, this is not the case and I think they don't work well together.

lunars97: yes, I know. I tried with SectionList in combination with react-native-paper and it works well so far :eyes: I will push soon changes

native/src/components/EventList.tsx
```diff
@@ -38,7 +39,7 @@ type EventListProps = {
 }
 
 const EventList = ({ events, regionModel, language, navigateTo, refresh }: EventListProps): ReactElement => {
-  const { t } = useTranslation('events')
+  const { t } = useTranslation('events', { lng: language })
```
lunars97, line 36: I wonder whether we want to have group titles being translated based on the content language or system language? 🤔

translations/src/translations.json
```diff
@@ -3919,8 +3919,6 @@
       "goToCalendar": "إلى التقويم",
       "recurring": "مكرر",
```
bahaaTuffaha, line 3842: why did you remove "recurring" if we still using it at EventListItem? and there is one translation for recurring in Arabic left.

translations/src/translations.json
```diff
@@ -3919,8 +3919,6 @@
...
       "today": "اليوم",
-      "todayRecurring": "اليوم، مكرر",
-      "onlyThisEvent": "هذه الفعالية فقط",
```
bahaaTuffaha, line 3843: ❓ why "onlyThisEvent" removed here only?

native/src/components/EventList.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  const renderSectionHeader = ({ section }: { section: EventSection }) =>
+    section.title ? (
+      <PaperList.Subheader style={{ textAlign: contentAlignment(language) }}>{section.title}</PaperList.Subheader>
+    ) : null
+
+  return (
+    <ListContainer>
+      <SectionList
```
bahaaTuffaha, line 75: On iOS the sticky headers got enabled. you can disabled that or find a way to give it a background to make it readable.
Image

shared/src/utils/__tests__/events.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    const now = DateTime.local()
+
+    const events = [
+      createEvent(1, now.minus({ days: 5 }), now.minus({ days: 5 }).plus({ hours: 2 })), // start: 2024-11-02, end: start + plus 2H
+      createEvent(2, now.minus({ days: 2 }), now.minus({ days: 2 }).plus({ hours: 2 })), // start: 2024-11-05, end: start + plus 2H
+      createEvent(3, now.plus({ days: 1 }), now.plus({ days: 1 }).plus({ hours: 2 })), // start: 2024-11-08, end: start + plus 2H
+      createEvent(4, now.plus({ days: 4 }), now.plus({ days: 4 }).plus({ hours: 2 })), // start: 2024-11-11, end: start + plus 2H
+    ]
```
bahaaTuffaha, line 58: These comments needs updating because the FakeTimer is at 2023

web/src/components/EventList.tsx
```diff
@@ -0,0 +1,83 @@
...
+            <EventListItem
+              event={event}
+              languageCode={languageCode}
+              key={event.path}
+              filterStartDate={startDate}
+              filterEndDate={endDate}
+            />
+          ))}
+          noItemsMessage='events:currentlyNoEvents'
+        />
```
bahaaTuffaha, line 77: 🤔  hmm the `if (startDate || endDate) {` logic seems different from native. one is handling the array and the other at the return section ...

bahaaTuffaha (approved):
Nice work ✨ 
Tested web and native. ✅

translations/src/translations.json
lunars97: this is from the previous changes, I forgot to revert it 👍

native/src/components/EventList.tsx
lunars97: :question: The question whether it is better to have it as a sticky header in web and native, for me it looks better with it and it lets users know where they are in the grouping located :thinking:

web/src/components/EventList.tsx
lunars97: well, they have different render structures, native renders `SectionList` in both cases, only the data differs, and web renders different components :eyes:

native/src/components/EventList.tsx
steffenkleinle: I'd say content language since its part of the content imo.

web/src/components/EventListItem.tsx
```diff
@@ -71,45 +64,31 @@ const getEventPlaceholder = (path: string): string => {
...
-
 const EventListItem = ({
   event,
   languageCode,
   filterStartDate = null,
   filterEndDate = null,
 }: EventListItemProps): ReactElement => {
-  const dateIcon = getDateIcon(event.date)
   const { t } = useTranslation('events')
+  const isRecurringDateIcon = event.isRecurring ? (
```
steffenkleinle: 🙃 This is not a boolean (which the `is` would imply)
```suggestion
  const recurringDateIcon = event.isRecurring ? (
```

web/src/components/EventListItem.tsx
```diff
@@ -71,45 +64,31 @@ const getEventPlaceholder = (path: string): string => {
...
+    <StyledListItem dir='auto' disablePadding secondaryAction={isRecurringDateIcon}>
       <StyledListItemButton component={Link} to={event.path}>
         <ListItemIcon>
           <Icon src={thumbnailSrc} alt='' />
         </ListItemIcon>
         <ListItemText
           disableTypography
           primary={
-            <Typography component='h2' variant='subtitle1'>
+            <Typography component='h3' variant='subtitle1'>
```
steffenkleinle, line 91: 🤔 Did you change this on purpose?

shared/src/utils/events.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (a.isRecurring !== b.isRecurring) {
+      return a.isRecurring ? 1 : -1
+    }
+    return a.date.startDate.toMillis() - b.date.startDate.toMillis()
+  })
+
+export const groupEventsByDate = (events: EventModel[]): Record<DateGroupKey, EventModel[]> =>
+  Object.fromEntries(
+    EVENT_DATE_GROUPS.map(key => [key, sortGroup(events.filter(event => getGroupKey(event) === key))]),
+  ) as Record<DateGroupKey, EventModel[]>
```
steffenkleinle: 🔧 I have another proposal for improvement which eliminates the need for a typecast, includes the filtering for empty groups and allows us to directly use the values without needing to destructure an object:

```suggestion
export const groupEventsByDate = (events: EventModel[]): [DateGroupKey, EventModel[]][] =>
  EVENT_DATE_GROUPS.map((key): [DateGroupKey, EventModel[]] => [
    key,
    sortGroup(events.filter(event => getGroupKey(event) === key)),
  ]).filter(([_, events]) => events.length > 0)
```

web/src/components/EventList.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </>
+    )
+  }
+
+  const groupedEvents = groupEventsByDate(events)
+
+  const dateGroups = EVENT_DATE_GROUPS.filter(key => groupedEvents[key].length > 0).map(key => (
+    <EventListGroup key={key} title={t(key)} events={groupedEvents[key]} languageCode={languageCode} />
+  ))
```
steffenkleinle: 🔧 With my other suggestion, this can be simplified to:

```suggestion
  const dateGroups = groupEventsByDate(events).map(([key, events]) => (
    <EventListGroup key={key} title={t(key)} events={events} languageCode={languageCode} />
  ))
```

native/src/components/EventList.tsx
```diff
@@ -0,0 +1,120 @@
...
+  if (startDate || endDate) {
+    sections = filteredEvents.length > 0 ? [{ title: null, data: filteredEvents }] : []
+  } else {
+    const groupedEvents = groupEventsByDate(events)
+    sections = EVENT_DATE_GROUPS.filter((key: DateGroupKey) => groupedEvents[key].length > 0).map(
+      (key: DateGroupKey) => ({
+        title: t(key),
+        data: groupedEvents[key],
+      }),
+    )
```
steffenkleinle: 🔧 With my other suggestion, this could be simplified to:
```suggestion
    sections = groupEventsByDate(events).map(([key, events]) => ({ title: t(key), data: events }))
```

native/src/components/EventList.tsx
steffenkleinle: I think sticky headers are quite nice, but only if they are
a) used everywhere consistently
b) have a white background, i.e. don't get obscured by underlying content/images

native/src/components/EventList.tsx
```diff
@@ -0,0 +1,120 @@
...
+            />
+          </>
+        }
+        ListEmptyComponent={
+          <Text
+            variant='body2'
+            style={{
+              alignSelf: 'center',
+              marginTop: 20,
+            }}>
```
steffenkleinle: 🙃 
```suggestion
          <Text variant='body2' style={{ alignSelf: 'center', marginTop: 20 }}>
```

steffenkleinle (commented):
❓ Why did you choose not to use https://oss.callstack.com/react-native-paper/docs/components/List/ListSection? I am still convinced that we should try to use components from our component library wherever applicable to adhere to material UI defaults and consistent UIs.

web/src/components/EventListItem.tsx
lunars97: yes, because the grouping title has `h2` now

native/src/components/EventList.tsx
lunars97: this is how it was formated :)

steffenkleinle: See https://prettier.io/docs/rationale#multi-line-objects
TLDR: If there is a newline after the opening curly brace, it is formatted as multiline object, otherwise in one line (if the line is not too long).

steffenkleinle: 🔧 Since this is actually also the third usage of this, lets make this a reusable component `ListEmptyComponent` or similar. I think there is no need to create a new file for this, I'd just add it to our `List` component as named export.

lunars97:
> ❓ Why did you choose not to use https://oss.callstack.com/react-native-paper/docs/components/List/ListSection? I am still convinced that we should try to use components from our component library wherever applicable to adhere to material UI defaults and consistent UIs.

We used before List.tsx which is `FlatList` with its virtualization, `SectionList` also has virtualization. Afaik `ListSection` is just a `View` under the hood :eyes:

native/src/components/EventList.tsx
lunars97: much cleaner now :heart_eyes:

shared/src/utils/groupEventsByDate.ts
bahaaTuffaha, line 1: This file is empty.

shared/src/utils/__tests__/events.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      createEvent(3, now.plus({ days: 1 }), now.plus({ days: 1 }).plus({ hours: 2 })), // start: 2023-11-08, end: start + plus 2H
+      createEvent(4, now.plus({ days: 4 }), now.plus({ days: 4 }).plus({ hours: 2 })), // start: 2023-11-11, end: start + plus 2H
+    ]
+
+    it('returns all events when startDate and endDate are null', () => {
+      const result = filterEvents(events, null, null)
+      expect(result).toEqual(events)
+    })
+
+    it('returns null when startDate is after endDate', () => {
```
bahaaTuffaha:
```suggestion
    it('returns empty array when startDate is after endDate', () => {
```

shared/src/utils/events.ts
```diff
@@ -0,0 +1,51 @@
...
+      return a.isRecurring ? 1 : -1
+    }
+    return a.date.startDate.toMillis() - b.date.startDate.toMillis()
+  })
+
+export const groupEventsByDate = (events: EventModel[]): [DateGroupKey, EventModel[]][] =>
+  EVENT_DATE_GROUPS.map((key): [DateGroupKey, EventModel[]] => [
+    key,
+    sortGroup(events.filter(event => getGroupKey(event) === key)),
+  ]).filter(([_, events]) => events.length > 0)
```
bahaaTuffaha, line 51: 🤔  `getGroupKey` ran 5× per event and each call it will re-computed DateTime.now() plus the date math.

bahaaTuffaha (approved):
Tested Web and native. ✅ 
Android and iOS looks great with these sticky headers ✅

shared/src/utils/groupEventsByDate.ts
lunars97: :see_no_evil: thanks for catching it, I remember deleting it...

web/src/components/EventList.tsx
```diff
@@ -0,0 +1,85 @@
...
+import EventsDateFilter from '../components/EventsDateFilter'
+import List from './base/List'
+
+type EventListGroupProps = {
+  title: string
+  events: EventModel[]
+  languageCode: string
+}
+
+export const EventListGroup = ({ title, events, languageCode }: EventListGroupProps): ReactElement => {
```
steffenkleinle: 🤔 Is this export necessary?
```suggestion
const EventListGroup = ({ title, events, languageCode }: EventListGroupProps): ReactElement => {
```

steffenkleinle (commented):
❌ There are currently no dividers at all on native. On web, there is only a divider before the subheader, where imo there should not be any.
- On native, the fix is pretty easy by just adding `ItemSeparatorComponent={Divider}` to the `SectionList` as well
- On web, you need to add the `events.map` in L28 with `withDividers()` and disable the dividers for `List` in L81 of `EventList`

lunars97:
> ❌ There are currently no dividers at all on native. On web, there is only a divider before the subheader, where imo there should not be any.
> 
> * On native, the fix is pretty easy by just adding `ItemSeparatorComponent={Divider}` to the `SectionList` as well
> * On web, you need to add the `events.map` in L28 with `withDividers()` and disable the dividers for `List` in L81 of `EventList`

ohh good catch! I havent' noticed it going back and forth between web and native. I will check it :+1: 
EDIT: I think for web it looks correct sematically, I guess? Because it is separating based on the grouping, there is a divider between event groups... :thinking:

steffenkleinle:
> > ❌ There are currently no dividers at all on native. On web, there is only a divider before the subheader, where imo there should not be any.
> > 
> > * On native, the fix is pretty easy by just adding `ItemSeparatorComponent={Divider}` to the `SectionList` as well
> > * On web, you need to add the `events.map` in L28 with `withDividers()` and disable the dividers for `List` in L81 of `EventList`
> 
> ohh good catch! I havent' noticed it going back and forth between web and native. I will check it 👍 EDIT: I think for web it looks correct sematically, I guess? Because it is separating based on the grouping, there is a divider between event groups... 🤔

There should be a divider between items, not between groups. Compare also the regions list.

lunars97:
> > > ❌ There are currently no dividers at all on native. On web, there is only a divider before the subheader, where imo there should not be any.
> > > 
> > > * On native, the fix is pretty easy by just adding `ItemSeparatorComponent={Divider}` to the `SectionList` as well
> > > * On web, you need to add the `events.map` in L28 with `withDividers()` and disable the dividers for `List` in L81 of `EventList`
> > 
> > 
> > ohh good catch! I havent' noticed it going back and forth between web and native. I will check it 👍 EDIT: I think for web it looks correct sematically, I guess? Because it is separating based on the grouping, there is a divider between event groups... 🤔
> 
> There should be a divider between items, not between groups. Compare also the regions list.

hmm in region list it is also the same :thinking: Should I fix it?

web/src/components/base/List.tsx
```diff
@@ -9,15 +9,22 @@ type ListProps = {
...
+  withDividers?: boolean
 }
 
-const List = ({ items, noItemsMessage, disablePadding, className }: ListProps): ReactElement | null => {
+const List = ({
+  items,
+  noItemsMessage,
+  disablePadding,
+  className,
+  withDividers: showDividers = true,
```
steffenkleinle: 🔧 Let's just name the prop `showDividers`
``` suggestion
  showDividers = true,
```

steffenkleinle (approved):
Not tested again, very nice work!

steffenkleinle:
>> There should be a divider between items, not between groups. Compare also the regions list.
>
> hmm in region list it is also the same 🤔 Should I fix it?

Ohh, you are right. I still think its better this way. The region list should stay as it is imo. So the current state would be my preferred solution.

lunars97:
@bahaaTuffaha do yo want to have a look again or is it okay to merge?

bahaaTuffaha:
> @bahaaTuffaha do yo want to have a look again or is it okay to merge?

sorry. I will have a look now

bahaaTuffaha (commented):
at web:
- The main header is covering the events headers.
- The events sticky headers are above the chat button.

https://github.com/user-attachments/assets/105feaab-9f5d-42ec-b798-588bdbb96978

bahaaTuffaha:
> at web:
> 
> * The main header is covering the events headers.
> * The events sticky headers are above the chat button.

I can fix them if you want...

lunars97:
> > at web:
> > 
> > * The main header is covering the events headers.
> > * The events sticky headers are above the chat button.
> 
> I can fix them if you want...

Wow good catch! Feel free, otherwise I will fix it next week :+1:

steffenkleinle (approved):
Nice work @lunars97 and thanks a lot for the fixes @bahaaTuffaha! And very nice solution to add `stickyTop` to `useDimensions` 😍

PR #4217, opened 2026-06-27
4199: Fix unsanitized content

### LINKED ISSUE #4199: Security: JavaScript executes in WebView from unsanitized page content and pushnotification text served by public API
### Describe the Bug

The Integreat app renders page content from the public API in a WebView with JavaScript enabled. The CMS does not sanitize `<script>` tags or JavaScript event handler attributes (e.g. `onerror`) before storing or serving content. Both attack vectors have been confirmed to execute on Android.

An attacker with CMS editor access can inject a script payload into a page's content by bypassing TinyMCE's client-side sanitization via a direct HTTP request. The payload is stored in the database, served raw via the public API, and executed by the app's WebView when any user opens that page.

EDIT: The same is true for Pushnotifications shown in the native app under the tab ` News` / `Nachrichten`

### Steps to Reproduce

1. A CMS editor submits a page edit with a script payload in the `content` field (bypassing TinyMCE via direct POST)
2. The payload is stored and served by the public API, e.g.:
   ```
   GET /api/v3/testumgebung/de/pages/
   ```
   ```json
   { "content": "<p><img src='x' onerror='alert(1)'><script>alert(2)</script></p>" }
   ```
3. Open the affected page in the Integreat Android app
4. Both the `onerror` handler and the `<script>` tag execute in the WebView

For pushnotification:
1. Create a pushnotification with `<script>` tag in the text field
2. Send it
3. Open the sent pushnotification in the "News" tab in the native app
4. See that the script executes

### Expected Behavior

Script tags and event handler attributes should never be present in API content. The app's WebView should not execute JavaScript from page content.

### Environment

- Device/Browser: Android
- OS version: confirmed on Android
- App version: current

### Related Issues

- digitalfabrik/integreat-cms#4373 — Root cause: missing server-side sanitization in `clean_content()`. Fix in progress.

### Additional Information

**Severity:** The fix belongs in the CMS (`clean_content()` should strip scripts and event handlers before storage). Once digitalfabrik/integreat-cms#4373 is resolved, the API will no longer serve executable content.

**Defense-in-depth option for the app:** Sanitize HTML content from the API before passing it to the WebView, or set `javaScriptEnabled={false}` on the WebView if JavaScript is not required for content rendering. This would prevent execution even if unsanitized content reaches the app.

**Scope:** Affects all pages/events/POIs — any content type served via the public API. The web app (React) is not affected as it sanitizes before DOM insertion.

lunars97:
### Short Description
This PR fixes unsanitized remote content which is a vulnerability to XSS. Requires fix only in the RemoteContent.tsx, seems like in push notification there is no need to add sanitization explicitly again because it is not raw html -> it's just visually there, no execution, safe to leave unsanitized I guess, but we can add additionally to remove these tags?

### Proposed Changes

- Add a `sanitizeContent` shared helper using `sanitize-html `(`dompurfy` needs a DOM and doesn't run in the RN runtime).
- Adjusted both native and web RemoteContent.tsx
- Added test
- Removed `node-modules` from modules in webpack.config.ts because of the error `isPlainObject is not a function` in web. Seems like it fixes it, there is a discussion https://stackoverflow.com/questions/48825061/webpack-sub-modules-not-resolved-correctly with a solution https://github.com/webpack/webpack/issues/6505#issuecomment-366502560

### Side Effects

N/A

### Checklist

### Testing
Go to Mülltrennung Category in Testumgebung http://localhost:9001/testumgebung/de/m%C3%BClltrennung-3 and see no script trigger
Go to local news http://localhost:9001/testumgebung/de/news/local/19598 and http://localhost:9001/testumgebung/de/news/local/19599 (no scripts are fired anymore)
Please smoke test native and web

### Resolved Issues

Fixes: #4199 

---


native/src/components/RemoteContent.tsx
```diff
@@ -149,6 +149,8 @@ const RemoteContent = ({
     [staticServerUrl],
   )
 
+  const sanitizedContent = useMemo(() => sanitizeContent(content), [content])
```
steffenkleinle: ❓ Is this `useMemo` really necessary?

native/src/utils/helpers.ts
```diff
@@ -117,3 +118,20 @@ export const getCachedResource = (resourceUrl: string, { resourceCache }: GetCac
...
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'allowfullscreen'],
+      a: ['href', 'name', 'target'],
+      '*': ['class', 'style', 'dir'],
+    },
+    allowedIframeHostnames: buildConfig().supportedIframeSources,
+  })
+}
```
steffenkleinle: ❓ `dompurify` which we use in web obviously doesn't work in native without a DOM, but what about the other way? Any reason why we don't switch to one shared sanitizing option?

lunars97: yes, we could use it for web as well :+1:

native/src/utils/helpers.ts
```diff
@@ -117,3 +118,20 @@ export const getCachedResource = (resourceUrl: string, { resourceCache }: GetCac
...
+    return ''
+  }
+
+  return sanitizeHtml(content, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'allowfullscreen'],
+      a: ['href', 'name', 'target'],
+      '*': ['class', 'style', 'dir'],
```
bahaaTuffaha: 'style' from what I have read could let an attacker to preform css injection.
I tried the app without 'style' it works as expected.

native/src/utils/helpers.ts
```diff
@@ -117,3 +118,20 @@ export const getCachedResource = (resourceUrl: string, { resourceCache }: GetCac
...
 }
+
+export const sanitizeContent = (content: string): string => {
+  if (!content) {
+    return ''
+  }
+
+  return sanitizeHtml(content, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
```
bahaaTuffaha: ❓  do we need to add attributes for img ?

lunars97: yes, because it is not included in default `allowedTags`

bahaaTuffaha: I meant to include `allowedAttributes` for img or this includes all of them by default ?

web/tools/webpack.config.ts
```diff
@@ -150,7 +150,6 @@ const createConfig = (
 
   const root = resolve(dirname, '../..')
   const configAssets = resolve(root, `node_modules/build-configs/${buildConfigName}/assets`)
-  const rootNodeModules = resolve(root, 'node_modules')
```
lunars97: Removed because of the error` isPlainObject is not a function` in web. Seems like it fixes it, there is a discussion https://stackoverflow.com/questions/48825061/webpack-sub-modules-not-resolved-correctly with a solution https://github.com/webpack/webpack/issues/6505#issuecomment-366502560

web/src/components/RemoteContent.tsx
```diff
@@ -118,15 +113,8 @@ const RemoteContent = ({ html, centered = false, smallText = false }: RemoteCont
...
-    () => ({
-      __html: Dompurify.sanitize(html, {
-        ADD_TAGS: [DOMPURIFY_TAG_IFRAME],
-        ADD_ATTR: [DOMPURIFY_ATTRIBUTE_FULLSCREEN, DOMPURIFY_ATTRIBUTE_TARGET],
-      }),
-    }),
-    [html],
-  )
+  const sanitizedContent = sanitizeContent(html)
+  const dangerouslySetInnerHTML = useMemo(() => ({ __html: sanitizedContent }), [sanitizedContent])
```
steffenkleinle: 🔧 If you already use a `useMemo`, you should probably also move the whole sanitation inside.
```suggestion
  const dangerouslySetInnerHTML = useMemo(() => ({ __html: sanitizeContent(html) }), [html])
```
Edit: This sounded wrong, the `useMemo` is necessary here to avoid causing rerenders in the child component. The point with including the sanitation still stands though.

native/src/utils/helpers.ts
lunars97: nice find! :+1:

shared/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  return sanitizeHtml(content, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'title', 'allowfullscreen'],
+      img: ['src', 'srcset', 'alt', 'title', 'width', 'height', 'loading'],
+      '*': ['class'],
+    },
+    allowedIframeHostnames: ['vimeo.com'],
```
steffenkleinle: ❌ Instead of hardcoding this here, please pass the corresponding sources from buildConfig:
``` tsx
sanitizeContent(html, { supportedIframeSources: buildConfig().supportedIframeSources })
```

shared/src/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,18 @@
+import sanitizeHtml from 'sanitize-html'
+
+export const sanitizeContent = (content: string): string => {
+  if (!content) {
+    return ''
+  }
+
+  return sanitizeHtml(content, {
```
steffenkleinle, line 10: ❓ The defaults between dompurify and sanitize-html differ quite a lot, have you checked whether these differences have any impact/break things for us?
  - style — DOMPurify's defaults allow inline style (with CSS-level sanitization). The shared config only whitelists class globally. Any CMS content with inline styles (custom colors, margins, etc.) now gets stripped. Notably RemoteContent.tsx:67–76 still reads and mutates element.style in a useEffect, but it now runs against always-empty inline styles.
  - dir — DOMPurify's defaults allow it. sanitize-html defaults don't, and the shared config doesn't add it. The interim native version had '*': ['class', 'style', 'dir']; the shared version dropped style and dir. Affects RTL rendering on inline elements.
  - aria-* / data-* — DOMPurify allows them by default. sanitize-html doesn't unless you configure allowedSchemesAppliedToAttributes or list them; the shared config doesn't. Accessibility hooks and CMS data-attrs get stripped.
  - iframe attributes beyond src/title/allowfullscreen — width, height, frameborder, allow, loading etc. no longer survive.

web/tools/webpack.config.ts
steffenkleinle: Hopefully gets irrelevant with #4221 anyway.

shared/src/utils/sanitizeContent.ts
lunars97: I thought there is not much difference and I didn't noticed any difference, about accessibility hooks and aria attributes I need to check them 🤔 I will check the points above 👍

lunars97: I extended the config to allow the points you mentioned above, I hope I did not miss anything :see_no_evil:

shared/src/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,22 @@
...
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'title', 'width', 'height', 'allow', 'allowfullscreen', 'loading', 'style'],
+      img: ['src', 'srcset', 'alt', 'title', 'width', 'height', 'loading', 'style'],
+      '*': ['class', 'style', 'dir', 'aria-*', 'data-*'],
+    },
+    ...(supportedIframeSources && supportedIframeSources.length > 0
+      ? { allowedIframeHostnames: supportedIframeSources }
+      : {}),
```
steffenkleinle: 🤔 Is there a problem if the array is empty?
```suggestion
    allowedIframeHostnames: supportedIframeSources,
```

shared/src/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,22 @@
+import sanitizeHtml from 'sanitize-html'
+
+export const sanitizeContent = (content: string, options: { supportedIframeSources?: string[] }): string => {
```
steffenkleinle: 🔧 The `supportedIframeSources` are required in the build config, so any reason to make this optional?
```suggestion
export const sanitizeContent = (content: string, options: { supportedIframeSources: string[] }): string => {
```

web/tools/webpack.config.ts
steffenkleinle, line 1: 🔧 I think this file should be deleted

shared/src/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  return sanitizeHtml(content, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'title', 'width', 'height', 'allow', 'allowfullscreen', 'loading', 'style'],
+      img: ['src', 'srcset', 'alt', 'title', 'width', 'height', 'loading', 'style'],
+      '*': ['class', 'style', 'dir', 'aria-*', 'data-*'],
+    },
+    allowedIframeHostnames: supportedIframeSources,
```
bahaaTuffaha: The allowed iframe hostNames are matching with exact hostname for example we currently use `vimeo.com` but and actual embedded link could look like:
`https://player.vimeo.com/video/{VIDEO_ID}

I think you need to use `allowedIframeDomains` instead.

shared/src/utils/sanitizeContent.ts
```diff
@@ -0,0 +1,20 @@
...
+    return ''
+  }
+
+  const { supportedIframeSources } = options
+
+  return sanitizeHtml(content, {
+    allowedTags: sanitizeHtml.defaults.allowedTags.concat(['iframe', 'img', 'details', 'summary']),
+    allowedAttributes: {
+      ...sanitizeHtml.defaults.allowedAttributes,
+      iframe: ['src', 'title', 'width', 'height', 'allow', 'allowfullscreen', 'loading', 'style'],
```
bahaaTuffaha, line 14: hmm what if it uses `id` or `role` will it stay?

native/src/components/RemoteContent.tsx
```diff
@@ -33,6 +35,11 @@ const RemoteContent = ({ onLoad, content, language, loading }: RemoteContentProp
 
   const resourceMap = mapValues(resourceCache, filePath => getStaticServerFileUrl(filePath, staticServerUrl))
 
+  const sanitizedContent = useMemo(
```
bahaaTuffaha, line 38: ❓  is this useMemo needed?

bahaaTuffaha (approved):
Nice!
It works as expected. Tested web and native ✅ 
I left some comments...

native/src/components/RemoteContent.tsx
lunars97: yes, to avoid rerenders in child components, see the comment above about `useMemo()`

shared/src/utils/sanitizeContent.ts
lunars97: `id` is already in default attributes, I dont think we have `role` attribute in a remote content, we have it in our components

steffenkleinle (approved):

Issue #4198, opened 2026-06-18
E2E tests for native failing


lunars97:
### Describe the Bug

Currently running e2e tests locally for native failing. All test cases fail. 
### Steps to Reproduce

1. Follow [running tests locally documentation](https://github.com/digitalfabrik/integreat-app/blob/main/native/docs/e2e-testing.md#running-the-tests-locally) to test native android and ios
2. See tests cases failing

### Expected Behavior
Test cases should be successfull

### Environment

- Device/Browser: N/A
- OS version: N/A
- App version: N/A

### Related Issues

N/A

### Additional Information

N/A

PR #4221, opened 2026-06-29
4166: Switch from webpack to vite

### LINKED ISSUE #4166: Switch from webpack to vite

andrew8er:
### Short Description

Replaces webpack with vite bundler in the `/web` module.

### Proposed Changes

- Add a `tsconfig.base.config`, that sets base config for all child `tsconfig.json` files.
- The root `/tsconfig.json` now only pulls in sub-modules as references.
- Move sources from `/e2e-tests`, `/build-configs`, `/translations`, `/shared` into `src` dirs. These package layouts are now consistent with `/web` and `/native`. This necessitated an update of app-toolbelt. Also modify IntelliJ directory config accordingly.
- Removes all `main` entries from all `package.json` files. These are for CJS modules only.
- The "frontend" modules (`/native` and `/web`) now do _not_ emit compiled code from tsc. Shared modules only emit declaration files, which is enough to type check the whole project.
- The sub-modules `/e2e-tests`, `/build-configs`, `/translations`, `/shared` now are configured in such a way, that metro and vite can live-reload changes from them. This is possible since metro can import TS files directly, but cannot do a real compilation step as `tsc` would do. It only strips type information. To achieve this, a few things changed:
  - The `exports` entries in `package.json` files were consolidated and now point to the original source files.
  - The sources now import the original `.ts` files. `tsc` allows this with the combination of `allowImportingTsExtensions` and `rewriteRelativeImportExtensions`.
  - The two enums `ErrorCode` and `FeedbackType` were replaced with objects and corresponding types. Enums are [discouraged](https://www.angularspace.com/breaking-the-enum-habit-why-typescript-developers-need-a-new-approach/) in TS anyway. Compatibility with type stripping is now enforced with the `erasableSyntaxOnly` option.
- Small fixes to cross-module imports.
- Small improvements to the `/native/metro-config.js`
- Fixed typos and unify wording (build-config) in documentation.
- Link depending npm packages with the yarn workspace protocol.
- Updates yarn to the latest stable version

### Open questions

- I did not test the whole e2e-test package, because I get some error when running Chrome. Please test this and inform me of anything that needs to be fixed.

### Side Effects

N/A

### Checklist

### Testing

As this PR touches a lot of different parts, the whole build, test and develop workflow should be tested.

### Resolved Issues

Fixes: #4166

Related: #4149:

> @bahaaTuffaha: I noticed that I need to run yarn ts:check sometimes to get the new translations working. did we change something in that regard ?

The translations are a bit complicated: the have a "source" aspect (the source files, which need to be compiled), which should now be resolved. But there is also a "run" aspect (the scripts in `/translation/tools`), which is not touched by this PR. But since your problem was solved by running `yarn run ts:check`, this should actually be resolved with this PR.

PS: 
- ~~I do not get why the `commit` check is failing.~~ Fixed now.
- The `Dependency Review` test is failing because the dependencies ARE in fact really outdated, but this PR does not address this.
- The `CodeScene` health checks are IMHO completely bogus and worst of all, not actionable. We removed it in Entitlementcard and I would suggest to do this here as well.

---


native/src/components/RemoteContent.tsx
```diff
@@ -39,11 +39,11 @@ const ORIGIN_WHITELIST = ['*'] as const
 
 export const renderWebviewError = (
   errorDomain: string | null | undefined,
-  errorCode: number,
+  errorCodes: number,
   errorDesc: string,
 ): React.ReactElement => (
   <Text>
-    ${errorDomain} ${errorCode} ${errorDesc}
+    ${errorDomain} ${errorCodes} ${errorDesc}
```
steffenkleinle: 🙃 
```suggestion
  errorCode: number,
errorDesc: string,
): React.ReactElement => (
<Text>
    ${errorDomain} ${errorCode} ${errorDesc}
```

shared/src/constants/map.ts
```diff
@@ -76,7 +76,7 @@ export const isMultiPlace = (placeFeature: MapFeature): boolean => placeFeature.
 
 export const mapConfig: MapConfigProps = {
   styleJSON: 'https://maps.tuerantuer.org/styles/integreat/style.json',
-  rtlPluginUrl: '/plugins/mapbox-gl-rtl-text-0.2.3.min.js',
+  rtlPluginUrl: '/plugins/mapbox-gl-rtl-text-0.2.3.min',
```
steffenkleinle: 🤔 Is this change intended?

shared/tsconfig.json
```diff
@@ -1,14 +1,30 @@
...
+    "emitDeclarationOnly": true,
+
     "types": ["jest", "node"],
-    "typeRoots": ["./@types", "node_modules/@types", "../node_modules/@types"]
+    "typeRoots": [
+      "./src/@types",
+      "node_modules/@types",
+      // TODO This assumes a certain layout of node_modules, which may break in the
+      //   future. We should add the "@types/jest" package to this package's dependencies.
+      "../node_modules/@types"
```
steffenkleinle: ❓ Are you still planning to do this here? Would you mind opening a ticket otherwise?

tools/package.json
```diff
@@ -5,6 +5,6 @@
   "type": "module",
   "devDependencies": {
     "@sentry/cli": "^3.4.0",
-    "app-toolbelt": "github:digitalfabrik/app-toolbelt#semver:1.0.0"
+    "app-toolbelt": "catalog:"
```
steffenkleinle, line 8: TIL, nice!

translations/tools/manage.ts
steffenkleinle, line 1: ❓ Did you forget to readd this file after moving this to src?

tsconfig.base.json
```diff
@@ -0,0 +1,18 @@
+{
+  // This file provides a base configuration, providing only correctness and language options
```
steffenkleinle, line 2: ❓ Do comments like this work in json? Are they stripped somehow?

web/index.html
steffenkleinle, line 1: ❓ Isn't this basically the same file as src/index.html.ejs? Do we need it twice?

web/package.json
```diff
@@ -61,46 +61,34 @@
...
     "schema-dts": "^2.0.0",
-    "style-loader": "^4.0.0",
-    "terser-webpack-plugin": "5.4.0",
-    "ts-loader": "^9.5.4",
-    "tsconfig-paths": "^4.2.0",
-    "url-loader": "^4.1.1",
-    "webpack": "^5.105.4",
-    "webpack-bundle-analyzer": "^5.3.0",
-    "webpack-cli": "^7.0.2",
-    "webpack-dev-server": "5.2.4"
```
steffenkleinle, line 104: 🎉

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
+/* eslint-disable prefer-arrow/prefer-arrow-functions */
```
steffenkleinle: ❓ Any reason to not use arrow functions here?

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
...
+
+export default defineConfig(async configEnv => {
+  const buildConfigName = env.CONFIG_NAME
+  if (!buildConfigName) {
+    throw new Error('Please specify a build config name via the CONFIG_NAME env var')
+  }
+
+  const buildConfig = loadBuildConfig(buildConfigName, WEB)
+  const mode = configEnv.mode as ViteMode
+  const isDev = configEnv.command === 'serve'
```
steffenkleinle: 🔧 Seems to be unused
```suggestion
```

mise.toml
steffenkleinle, line 1: ❓ This is not a complete switch yet, but allows devs to manage their toolchain with mise, right?

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
...
+    },
+    preview: {
+      open: true,
+      port: 9001,
+    },
+    server: {
+      port: 9001,
+      host: true,
+    },
+  } satisfies UserConfig
```
steffenkleinle, line 91: 👍 TIL 😍

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
...
+function buildConfigAssetsPlugin(assetsDir: string): Plugin {
+  let outDir: string
+
+  return {
+    name: 'integreat:build-config-assets',
+    configResolved(config) {
+      outDir = resolve(config.root, config.build.outDir)
+    },
+    configureServer(server) {
+      server.middlewares.use((req, res, next) => {
```
steffenkleinle: 🙃 
```suggestion
      server.middlewares.use((request, result, next) => {
```

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
...
+        if (!req.url) {
+          return next()
+        }
+
+        const requested = decodeURIComponent(req.url.split('?')[0] ?? '')
+        const filePath = join(assetsDir, requested)
+
+        if (!filePath.startsWith(`${assetsDir}/`) || !existsSync(filePath) || !statSync(filePath).isFile()) {
```
steffenkleinle: 🤔 Is it ever possible that `filePath` does not start with `assetsDir`?
```suggestion
        if (!existsSync(filePath) || !statSync(filePath).isFile()) {
```

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
+/* eslint-disable prefer-arrow/prefer-arrow-functions */
+/* eslint-disable no-magic-numbers */
+/* eslint-disable no-console */
+/* eslint-disable no-else-return */
```
steffenkleinle: ❓ Any reason to disable this rule?

eslint.config.js
```diff
@@ -39,6 +39,7 @@ export default defineConfig([
   ...reactConfig,
   ...typescriptConfig,
   ...prettierConfig,
+  globalIgnores(['**/vite.config.ts', '**/jest.*.ts']),
```
steffenkleinle, line 42: ❓ Is this to be able to circumvent linting rules or because linting those files is not trivially possible?

web/vite.config.ts
```diff
@@ -0,0 +1,293 @@
...
+
+        const requested = decodeURIComponent(req.url.split('?')[0] ?? '')
+        const filePath = join(assetsDir, requested)
+
+        if (!filePath.startsWith(`${assetsDir}/`) || !existsSync(filePath) || !statSync(filePath).isFile()) {
+          return next()
+        } else {
+          res.setHeader('Content-Type', mimeTypeFor(filePath))
+          createReadStream(filePath).pipe(res)
+          return undefined
```
steffenkleinle, line 130: 🤔 Is this return necessary?
``` suggestion
```

steffenkleinle (changes requested):
Thank you so much for this huge effort! There are so many DX and speed improvements here, crazy!
I still found a few (smaller) problems though:
1. on native, I can't run the app, in dev I just get a 404 while on production the build fails. Most likely, you still have to adjust `entryFile` in `build.gradle`
```
/gradlew app:installRelease -PreactNativeDevServerPort=8081 FAILURE: Build failed with an exception. * What went wrong:
A problem was found with the configuration of task ':app:createBundleReleaseJsAndAssets' (type 'BundleHermesCTask'). - In plugin 'com.facebook.react' type 'com.facebook.react.tasks.BundleHermesCTask' property 'entryFile' specifies file '/Users/st/digitalfabrik/integreat-app/native/index.ts' which doesn't exist. Reason: An input file was expected to be present but it doesn't exist. Possible solutions: 1. Make sure the file exists before the task is called. 2. Make sure that the task which produces the file is declared as an input. For more information, please refer to https://docs.gradle.org/9.0.0/userguide/validation_problems.html#input_file_does_not_exist in the Gradle documentation. * Try:
> Make sure the file exists before the task is called
> Make sure that the task which produces the file is declared as an input
> Run with --scan to generate a Build Scan (Powered by Develocity). BUILD FAILED in 13s.
```
Image

2. The whole app (web) has a light blue border:
Image

3. translations `manage` program seems to be missing
4. tools/checkCircularDependencies.sh fails silently:
```
✖ Error: ENOENT: no such file or directory, stat '/Users/st/digitalfabrik/integreat-app/shared/index.ts'
✖ Error: ENOENT: no such file or directory, stat '/Users/st/digitalfabrik/integreat-app/build-configs/index.ts'
```

steffenkleinle:
> The CodeScene health checks are IMHO completely bogus and worst of all, not actionable. We removed it in Entitlementcard and I would suggest to do this here as well.

Yeah, I guess everybody just ignores them :D
https://github.com/digitalfabrik/integreat-app/issues/4226

steffenkleinle:
> The Dependency Review test is failing because the dependencies ARE in fact really outdated, but this PR does not address this.

Issue for dependency upgrades (blocked by this PR): https://github.com/digitalfabrik/integreat-app/issues/4233

native/src/components/RemoteContent.tsx
andrew8er: After rebase, this code does not exist anymore.

shared/src/constants/map.ts
andrew8er: Good catch, that was a search/replace error.

shared/tsconfig.json
andrew8er: I tried it, but yarn (in its current configuration) hoists packages, so it's not really clear where they'll end up. With `nodeLinker: pnpm`, this would be an error, as packages can only access their declared dependencies. I would recommend it in the long run, since it's then impossible to accidentally add dependencies.

For now, I removed the comment.

See: https://yarnpkg.com/configuration/yarnrc#nodeLinker

tools/package.json
andrew8er: It's a feature ported from pnpm, very useful and sorely missed in npm.

tsconfig.base.json
andrew8er: tsc (and its compiler library that is used by other packages) actually use JSONC as the config language, so yes, this is legal. Also trailing commas work. Everyone that uses JSON as a config language should use JSONC or be punished with public impaling.

web/index.html
andrew8er: Vite will use `index.html`, I removed the other file.

web/vite.config.ts
andrew8er: There are a few, but the main reason for me is that it forces me to layout the functions/variables _upside-down_:

I want to have the top-level functionality at the top of the file and the utility functions at the bottom, but when using [function expressions (as opposed to function declarations)](https://medium.com/geekculture/javascript-functions-for-dummies-part-1-function-declarations-vs-expressions-b11ad65c050f) I cannot do this. Instead the layout gets reversed: lots of irrelevant (to the reader that wants a quick glance) cruft at the top, the meat of the file (often times the things you also want to export) at the bottom.

Besides, a function declaration is a perfectly fine way to have a function in JS.

mise.toml
andrew8er: It helps with my setup, because I cannot install yarn through brew, and I cannot use my global Node.js (which is too new).

But it's not really necessary for now, I can remove it, if you want.

web/vite.config.ts
andrew8er: I found a write-up of quiet nifty TS tricks here, there are some things I didn't know: https://github.com/AllThingsSmitty/typescript-tips-everyone-should-know

web/vite.config.ts
andrew8er: This is the terminology vite uses for its middleware, but I agree, these abbreviations are stupid. I replaced it.

web/vite.config.ts
andrew8er: Yes, the middleware will receive all requests for every module, and this check serves as a kind of shortcut: it is faster to check the string than to actually hit the disk and check for the existence of a file.

eslint.config.js
andrew8er: I excluded it, because eslint gives me an error about these files not being included in any project. 

Indeed, these files are intentionally not covered by the `input` specifiers in `tsconfig.json` files, because these are config files, not source files that contribute to a module, and they have different semantics, especially when it comes to resolving imports. E.g. imports in `vite.config.ts` need to be type strippable by Node.js, because vite will not add any further processing to files. This is the reason, I removed the enums from `shared`, because these are not loadable by raw Node.js and we need to load them in `vite.config.ts`.

I tried to include them by eslint's [allowdefaultproject](https://typescript-eslint.io/packages/parser/#allowdefaultproject) directive, but then again, I hit other errors:

https://typescript-eslint.io/troubleshooting/typed-linting/#i-get-errors-telling-me-having-many-files-run-with-the-default-project-is-known-to-cause-performance-issues-and-slow-down-linting

So I opted to just ignore config files from eslint, I think it's a reasonable trade off.

web/vite.config.ts
andrew8er: > /* eslint-disable prefer-arrow/prefer-arrow-functions */

Already explained above.

> /* eslint-disable no-magic-numbers */

Sometimes, you DO have magic numbers.

> /* eslint-disable no-console */

Sometimes, you do actually want to output to the console.

> /* eslint-disable no-else-return */

This takes a bit longer to explain, but I really hate the "trailing-else" handling. It obscures the conditions an (now invisible) else block gets executed, where a "true" else block provides the (IMHO correct) indentation. I can explain it in more depth separately if there is any interest. 

All in all, I realized that config files are now ignored by eslint anyway (see below), so I could remove these.

web/vite.config.ts
andrew8er: It is not strictly necessary, but in these cases I tend to include it. When I have an `if () {} else {}` construct in the tail position (i.e. the last thing that a function will execute), I keep both branches uniform and add both return statements so it is immediately clear, what this function returns given the `if` condition. I do avoid implicit `return undefined`s as much as possible.

I would usually prefer a ternary operator, but in this case it is not possible, because of the side effects in the else block.

andrew8er:
@steffenkleinle 
> on native, I can't run the app, in dev I just get a 404 while on production the build fails. Most likely, you still have to adjust entryFile in build.gradle

This is fixed now, turns out, metro really needs this `index.ts` file in `/native`.

> The whole app (web) has a light blue border:

Looks like a CSS margin setting on the body and/or html node is missing. I don't know, why this is different than on webpack. I'll have a look.

> translations manage program seems to be missing

The file moved into `src`. From where is it started?

> tools/checkCircularDependencies.sh fails silently:

Should be fixed now.

shared/tsconfig.json
steffenkleinle: Thanks for explaining!

tsconfig.base.json
steffenkleinle: Haha yeah. I actually never followed up on this, just thought its not possible because it got flagged in the IDE. Definitely good to know, thank you!

web/vite.config.ts
steffenkleinle: Makes sense, thanks for explaining!

> Besides, a function declaration is a perfectly fine way to have a function in JS.

For sure, so you'd use function declarations everywhere? Or is this just a special case here? I don't really mind which one and never really questioned the linting rule, I think its just nice to choose either for the sake of consistency/avoid refactorings bc of personal preference in the matter.

mise.toml
steffenkleinle: I personally use nix, so I won't use this. But generally it makes sense I guess, so feel free to leave it. If nobody of us uses in the end, we can still remove it at a later point again.

web/vite.config.ts
steffenkleinle: But don't you just add `assetsDir` directly before? Or what am I missing? But obviously very minor, so we can also just keep the check. 🤔 
``` ts
        const filePath = join(assetsDir, requested)
```

eslint.config.js
steffenkleinle: Agreed, thanks for trying. Definitely reasonable.

web/vite.config.ts
steffenkleinle: Thanks for explaining once again, great reasoning.

web/vite.config.ts
steffenkleinle: Yep, sure, thanks for the explanations. The question was specifically for `no-else-return` bc so far I really liked the rule since (just in my personal opinion) its more readable. But I am very interested in hearing some more about your opinion on different rules and why you think they are counterproductive. I think you had a few other rules you were not that convinced of and/or missing? Maybe you can (only if you are motivated) put those to a list and let us know your reasoning? I'd be very happy to learn and improve here :)

> Sometimes, you DO have magic numbers.
> Sometimes, you do actually want to output to the console.

Definitely, but we previously disabled these rules in the eslint.config.js for tools/config files.

web/vite.config.ts
```diff
@@ -0,0 +1,287 @@
...
+        templateManifest(buildConfig, versionName),
+        ...templateAssetLinks(buildConfig),
+        ...templateAppleAppSiteAssociations(buildConfig),
+      ]),
+    ],
+    build: {
+      outDir: join('build', buildConfigName),
+      emptyOutDir: true,
+      // TODO Do we really need source maps in the final build? Sentry can upload these.
+      sourcemap: true,
```
steffenkleinle, line 63: I guess we currently upload the source-maps in `deliver_web` with sentry. Maybe the paths still have to be adjusted there though, the builds are now in `build` instead of `dist`, right?

steffenkleinle: Same for the general release process in `deliver_web`

steffenkleinle:
> > The whole app (web) has a light blue border:
> 
> Looks like a CSS margin setting on the body and/or html node is missing. I don't know, why this is different than on webpack. I'll have a look.

Thanks! Let me know if I should have a look as well.

> > translations manage program seems to be missing
> 
> The file moved into `src`. From where is it started?

I neither see the file in the diff nor locally when checking out the branch. Are you sure you added it? 🤔 I think you fixed all usages already. But its used in xcode prebuilt scripts and in scripts from the commandline via the package.json (which I think you all fixed already).

translations/tools/manage.ts
andrew8er: Ah, it was indeed removed instead of moved… my bad…

web/vite.config.ts
andrew8er: I use function expressions sometimes, but usually only for inline functions, otherwise I prefer declarations. Not a hard rule though.

>  I think its just nice to choose either for the sake of consistency/avoid refactorings bc of personal preference in the matter.

I would say, in this case, it is not just a personal preference, I genuinely think this rule makes the code layout worse.

mise.toml
andrew8er: I'm in the process of preparing a small knowledge exchange on mise, maybe next week.

web/vite.config.ts
andrew8er: Ah you're right. This file was partly created by Claude, I tweaked it a lot but missed that.

web/vite.config.ts
andrew8er: Yeah, I will write up a list (I already have a small list with some explanations). 

The console log here is necessary, because developers to immediately see what config values where chosen.

web/vite.config.ts
andrew8er: I'll have a look at the paths. I choose a distinct directory for the vite output, because I wanted to separate it from the `dist` output that tsc uses. These are to completely different kinds of output and should not be mixed.

andrew8er:
> I neither see the file in the diff nor locally when checking out the branch. Are you sure you added it? 🤔 I think you fixed all usages already. But its used in xcode prebuilt scripts and in scripts from the commandline via the package.json (which I think you all fixed already).

I accidentally removed – instead of moved – it. I'll guess I need to fix up a few file paths.

andrew8er:
Ok
- The CSS issue is fixed (may need a hard reload in the browser).
- Paths to `manage.ts` (and the file itself) is fixed.
- Build paths in CircleCI config should all be fixed.

Did I miss anything?

web/package.json
```diff
@@ -4,21 +4,21 @@
...
+    "start:aschaffenburg": "cross-env CONFIG_NAME=aschaffenburg vite dev",
+    "start:integreat-e2e": "cross-env CONFIG_NAME=integreat-e2e vite dev",
+    "start:obdach": "cross-env CONFIG_NAME=obdach vite dev",
+    "build": "vite build",
+    "build:integreat": "cross-env CONFIG_NAME=integreat vite dev",
+    "build:integreat-test-cms": "cross-env CONFIG_NAME=integreat-test-cms vite dev",
+    "build:malte": "cross-env CONFIG_NAME=malte vite dev",
+    "build:malte-test-cms": "cross-env CONFIG_NAME=malte-test-cms vite dev",
+    "build:aschaffenburg": "cross-env CONFIG_NAME=aschaffenburg vite dev",
+    "build:obdach": "cross-env CONFIG_NAME=obdach vite build",
```
bahaaTuffaha: ❓ why `build:obdach` using `vite build` and the rest uses vite dev

web/package.json
```diff
@@ -4,21 +4,21 @@
...
-    "build:obdach": "yarn run webpack-ts --env config_name=obdach",
+    "start": "cross-env CONFIG_NAME=integreat-test-cms vite dev",
+    "start:integreat": "cross-env CONFIG_NAME=integreat vite dev",
+    "start:integreat-test-cms": "cross-env CONFIG_NAME=integreat-test-cms vite dev",
+    "start:malte": "cross-env CONFIG_NAME=malte vite dev",
+    "start:malte-test-cms": "cross-env CONFIG_NAME=malte-test-cms vite dev",
+    "start:aschaffenburg": "cross-env CONFIG_NAME=aschaffenburg vite dev",
+    "start:integreat-e2e": "cross-env CONFIG_NAME=integreat-e2e vite dev",
+    "start:obdach": "cross-env CONFIG_NAME=obdach vite dev",
+    "build": "vite build",
```
bahaaTuffaha: missing config name here?
```suggestion
    "build": "vite build:integreat",
```

steffenkleinle (commented):
One more thing that still needs to be adjusted:
There are still a few cases of a wrong `translation.json` path in xcode prebuilt scripts which causes iOS builds to fail. Replacing all occurrences of `translations/translations.json` with `translations/src/translations.json` did the trick for me. I'll still trigger the e2e test workflow, but otherwise I tested on Android, iOS and web and everything works as expected now.
Edit: Obviously still fails due to the translation.json paths: https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/16900/details?workflowId=abce7b89-9198-42cf-9afb-5f2717f82f8e

build-configs/package.json
```diff
@@ -1,47 +1,57 @@
...
-      "default": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      "import": "./src/index.ts",
+      "types": "./src/index.d.ts",
+      "default": "./src/index.ts"
+    },
+    "./aschaffenburg/build-config-name": {
+      "import": "./src/integreat/build-config-name/index.ts",
+      "types": "./src/integreat/build-config-name/index.d.ts",
+      "default": "./src/integreat/build-config-name/index.ts"
```
bahaaTuffaha, line 34:
```suggestion
    "./aschaffenburg/build-config-name": {
      "import": "./src/aschaffenburg/build-config-name/index.ts",
      "types": "./src/aschaffenburg/build-config-name/index.d.ts",
      "default": "./src/aschaffenburg/build-config-name/index.ts"
```

web/package.json
```diff
@@ -4,21 +4,21 @@
...
+    "build:integreat": "cross-env CONFIG_NAME=integreat vite dev",
+    "build:integreat-test-cms": "cross-env CONFIG_NAME=integreat-test-cms vite dev",
+    "build:malte": "cross-env CONFIG_NAME=malte vite dev",
+    "build:malte-test-cms": "cross-env CONFIG_NAME=malte-test-cms vite dev",
+    "build:aschaffenburg": "cross-env CONFIG_NAME=aschaffenburg vite dev",
+    "build:obdach": "cross-env CONFIG_NAME=obdach vite build",
     "check:build": "yarn es-check es2022 ./dist/**/*.js --verbose --module",
```
bahaaTuffaha: "check:build" points to `./dist/**/*.js`
Vite outputs to `build/<configName>/`

lunars97:
Thank you a lot for your effort and enourmous contribution :heart_eyes: Regarding the light blue frame in web browser. it is still there after hard load :thinking: Tested with Chrome...

web/package.json
```diff
@@ -61,46 +61,34 @@
...
     "stylis": "^4.3.6",
-    "translations": "0.0.1"
+    "translations": "workspace:^"
   },
   "devDependencies": {
-    "@pmmmwh/react-refresh-webpack-plugin": "^0.6.2",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "6.9.1",
     "@testing-library/react": "^16.3.2",
     "@types/assets-webpack-plugin": "^7.1.6",
```
bahaaTuffaha: - @types/assets-webpack-plugin
- @types/webpack-bundle-analyzer

do we still need them?

bahaaTuffaha (commented):
at `.circleci/src/jobs/build_web.yml` still there is `--env` flags

build-configs/package.json
```diff
@@ -1,47 +1,57 @@
...
     },
     "./integreat/build-config-name": {
-      "import": "./dist/integreat/build-config-name/index.js",
-      "default": "./dist/integreat/build-config-name/index.js",
-      "types": "./dist/integreat/build-config-name/index.d.ts"
+      "import": "./src/integreat-test-cms/index.js",
+      "types": "./src/integreat-test-cms/index.d.ts",
+      "default": "./src/integreat-test-cms/index.js"
```
bahaaTuffaha: ❓ should it be `/build-config-name` ?

steffenkleinle:
> Thank you a lot for your effort and enourmous contribution 😍 Regarding the light blue frame in web browser. it is still there after hard load 🤔 Tested with Chrome...

For me the frame is now gone, so probably a caching problem?

build-configs/package.json
bahaaTuffaha, line 1: There is no one for malte:
`"./malte/build-config-name": {`

lunars97:
- There is no frame in web browser anymore :+1:  I forgot to pull new changes before :grimacing: 
- I ran e2e-tests web locally, and it runs and for native there is the error related to file not being found in exports map, I guess. See the screenshot below: 
image

- Regarding the tests failing on e2e-tests native, I am fixing them in my PR #4198

web/package.json
andrew8er: Oh good catch, that was an error…

andrew8er:
@steffenkleinle 
> There are still a few cases of a wrong translation.json path in xcode prebuilt scripts which causes iOS builds to fail. Replacing all occurrences of translations/translations.json with translations/src/translations.json did the trick for me.

The Xcode files should be fixed, however I still get a compile error (unrelated to this PR):
Screenshot 2026-07-06 at 11 47 00

…why the hell doesn't Xcode give me a proper log? I found [this](https://github.com/react/react-native/issues/42221), but the solutions did not help.

@bahaaTuffaha 
> at .circleci/src/jobs/build_web.yml still there is --env flags

Fixed. Also the missing/wrong scripts in the `/web/package.json` and the missing `export` entries in the `/build-configs/package.json` were fixed. I removed the redundant `viewport` meta entries in `index.html`, two of them were identical, but the third was a bit different. I'm not sure, if I picked the correct one though.

@lunars97 
> I ran e2e-tests web locally, and it runs and for native there is the error related to file not being found in exports map, I guess. See the screenshot below:

I couldn't run the e2e tests yet, I got some error running Chrome. Can I do anything in this PR now, or will this be fixed in your PR?

steffenkleinle:
@andrew8er both android and ios builds (at least for integreat-e2e) are failing, see https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/16908/workflows/ff49a180-ca3f-48cd-9f5b-c46cb30f9b00/jobs/77366 and https://app.circleci.com/pipelines/github/digitalfabrik/integreat-app/16908/workflows/ff49a180-ca3f-48cd-9f5b-c46cb30f9b00/jobs/77363.
Perhaps you are missing the exports in `build-configs/package.json` for `integreat-e2e`?

> …why the hell doesn't Xcode give me a proper log? I found https://github.com/react/react-native/issues/42221, but the solutions did not help.

Very good question, haven't figured that out either yet. One of the many reasons why I don't like xcode :D
Locally, I am able to build successfully.

> I couldn't run the e2e tests yet, I got some error running Chrome. Can I do anything in this PR now, or will this be fixed in your PR?

I guess if they are succeeding in the CI it's enough, no need to run them locally.

lunars97:
`I couldn't run the e2e tests yet, I got some error running Chrome. Can I do anything in this PR now, or will this be fixed in your PR?` As @steffenkleinle mentioned above, if it is succeding in ci, it is enough. I ran it also on web e2e and it passed test, but there is an issue with build for integreat-e2e. Regarding the test cases failing, I will fix them in my PR :+1:

steffenkleinle:
> I ran it also on web e2e and it passed test, but there is an issue with build for integreat-e2e. Regarding the test cases failing, I will fix them in my PR 👍

Already fixed, at least succeeds locally for me and in the CI now.

bahaaTuffaha:
I can't run `yarn start` at native

```
/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/util.ts:17
  throw new Error(`${name} not found in ${path.resolve(process.cwd(), directory)} or parent directories!`)
        ^


Error: integreat-test-cms not found in /Users/digitalfabrik/Repos/integreat-app/native/build-configs or parent directories!
    at findPathInParents (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/util.ts:17:9)
    at loadBuildConfig (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/build-config/loader.ts:25:37)
    at Command.<anonymous> (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/build-config/program-to-bash.ts:19:33)
    at Command.listener [as _actionHandler] (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:568:17)
    at /Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1604:14
    at Command._chainOrCall (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1488:12)
    at Command._parseCommand (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1603:27)
    at /Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1367:27
    at Command._chainOrCall (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1488:12)
    at Command._dispatchSubcommand (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1363:25)

Node.js v22.13.0
node:internal/errors:983
  const err = new Error(message);
              ^

Error: Command failed: npx --no app-toolbelt v0 build-config to-bash integreat-test-cms common
/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/util.ts:17
  throw new Error(`${name} not found in ${path.resolve(process.cwd(), directory)} or parent directories!`)
        ^


Error: integreat-test-cms not found in /Users/digitalfabrik/Repos/integreat-app/native/build-configs or parent directories!
    at findPathInParents (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/util.ts:17:9)
    at loadBuildConfig (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/build-config/loader.ts:25:37)
    at Command.<anonymous> (/Users/digitalfabrik/Repos/integreat-app/node_modules/app-toolbelt/src/build-config/program-to-bash.ts:19:33)
    at Command.listener [as _actionHandler] (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:568:17)
    at /Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1604:14
    at Command._chainOrCall (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1488:12)
    at Command._parseCommand (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1603:27)
    at /Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1367:27
    at Command._chainOrCall (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1488:12)
    at Command._dispatchSubcommand (/Users/digitalfabrik/Repos/integreat-app/node_modules/commander/lib/command.js:1363:25)

Node.js v22.13.0
```

steffenkleinle (approved):
Nice work, LGTM! e2e tests also succeed now 🚀

steffenkleinle:
> I can't run `yarn start` at native

Works for me, perhaps you have to force rebuild the workspaces:
```
yarn tsc --build --force build-configs translations shared
```

lunars97 (approved):
Really great work! LGTM 👍 Tested with ios emulator and safari

web/index.html
```diff
@@ -0,0 +1,187 @@
...
+    <link rel="icon" type="image/png" sizes="16x16" href="<%= config.icons.favicons %>favicon-16x16.png" />
+    <link rel="icon" type="image/png" sizes="48x48" href="<%= config.icons.favicons %>favicon-48x48.png" />
+    <link rel="icon" type="image/png" sizes="96x96" href="<%= config.icons.favicons %>favicon-96x96.png" />
+    <link rel="icon" type="image/png" sizes="192x192" href="<%= config.icons.favicons %>favicon-192x192.png" />
+    <link rel="icon" type="image/png" sizes="512x512" href="<%= config.icons.favicons %>favicon-512x512.png" />
+
+    <link rel="apple-touch-icon" type="image/png" sizes="180x180" href="<%= config.icons.appleTouchIcon %>" />
+    <link rel="shortcut icon" href="<%= config.icons.favicons %>favicon.ico" />
+
+    <link rel="mask-icon" href="/safari-pinned-tab.svg" color="<%= config.lightTheme.palette.secondary.main %>" />
```
bahaaTuffaha, line 106: ❓  I don't think we have that svg at the project files...

.circleci/src/jobs/build_web.yml
```diff
@@ -16,7 +16,7 @@ steps:
   - restore_environment_variables
   - run:
       name: << parameters.build_config_name >> build
-      command: yarn workspace web build:<< parameters.build_config_name >> --env commit_sha=${CIRCLE_SHA1} --env version_name=${NEW_VERSION_NAME}
+      command: yarn workspace web build:<< parameters.build_config_name >>
   - store_artifacts:
       name: Store output of bundle analyzer
```
bahaaTuffaha, line 21: 🤔  do we need an equivalent to `BundleAnalyzerPlugin` in vite like `vite-bundle-analyzer` ?

bahaaTuffaha (commented):
Nice work!
Tested web and native ✅

web/index.html
andrew8er: I removed this in #4241.

.circleci/src/jobs/build_web.yml
andrew8er: There is https://github.com/nonzzz/vite-bundle-analyzer that could be integrated. But does anyone actually look at these graphs? The vite output is already pretty detailed.

Looking at the numbers, I noticed that we have a lot of optimization options with our svg assets.

PR #4230, opened 2026-07-02
4225: Fix endless loop in PlaceFiltersModal

### LINKED ISSUE #4225: Fix back navigation loop when places filter is toggled
### Describe the Bug

At locations when you toggle a filter and try to go back or after pressing the show button at the filterLocation and go back it will keep opening the filter modal again.

### Steps to Reproduce

1. Launch integreat and go to places/map.
2. Click on filter locations button and toggle one of the filter then go back.
3. keep going back from places/map to previous screens.
4. it will keep opening the filter modal.

### Expected Behavior

we should exclude the filter modal from the history.

### Environment

- Device/Browser: Android and iOS
- OS version: for android 36.1
- App version: (main branch)

### Related Issues

https://github.com/digitalfabrik/integreat-app/pull/4212

### Additional Information

https://github.com/user-attachments/assets/3e718e3b-50e2-435a-b260-56d2116449a4

steffenkleinle:
### Short Description

Fix endless loop with navigating back after opening the place filter modal.

### Proposed Changes

- Add temporary state for the filter modal and only update the filters when the modal is closed
- Use the correct disabled background color

### Side Effects

N/A

### Checklist

### Testing

See issue.

### Resolved Issues

Fixes: #4225

---

### DIFF
--- a/native/src/components/PlaceFiltersModal.tsx
+++ b/native/src/components/PlaceFiltersModal.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react'
+import React, { ReactElement, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Button } from 'react-native-paper'
 import { SafeAreaView } from 'react-native-safe-area-context'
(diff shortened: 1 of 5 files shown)


native/src/components/PlaceFiltersModal.tsx
```diff
@@ -56,30 +56,36 @@ const StyledSvgUri = styled(SvgUri)<{ active: boolean }>`
...
-  selectedPlaceCategory,
-  setSelectedPlaceCategory,
+  placeCategoryFilter,
   currentlyOpenFilter,
-  setCurrentlyOpenFilter,
-  placesCount,
+  getPlacesCount,
 }: PlaceFiltersModalProps): ReactElement => {
+  const [tempPlaceCategory, setTempPlaceCategory] = useState(placeCategoryFilter)
+  const [tempCurrentlyOpen, setTemCurrentlyOpen] = useState(currentlyOpenFilter)
```
bahaaTuffaha:
```suggestion
  const [tempCurrentlyOpen, setTempCurrentlyOpen] = useState(currentlyOpenFilter)
```

native/src/routes/Places.tsx
```diff
@@ -98,6 +104,11 @@ const Places = ({ refresh, localHistory, initialZoom, places: allPlaces, regionM
     }
   }
 
+  const closeModal = ({ placeCategoryFilter, currentlyOpenFilter }: PlaceFilters) => {
+    localHistory.pop()
+    localHistory.pushReset({ placeCategoryId: placeCategoryFilter?.id, currentlyOpen: currentlyOpenFilter })
+  }
```
lunars97, line 110: 👍

lunars97 (approved):
Tested on android emulator, works as expected 👍 Thank you for fixing it 💚

native/src/routes/Places.tsx
bahaaTuffaha, line 110: pushReset will affect the selected place. when I go back it will rest.
How about push instead so it preserves current slug:


```suggestion
  const closeModal = ({ placeCategoryFilter, currentlyOpenFilter }: PlaceFilters) => {
    localHistory.pop()
   localHistory.push({
    placeCategoryId: placeCategoryFilter?.id,
    currentlyOpen: currentlyOpenFilter,
    showFilterSelection: false,
  })
  }
```

native/src/components/PlaceFiltersModal.tsx
bahaaTuffaha, line 1: when we toggle the currently open then remove it form the overlay and then open the filter again it will stay enabled.

I think it needs a useEffect to update the temp values.

bahaaTuffaha (changes requested):

native/src/routes/Places.tsx
steffenkleinle: This is on purpose! If I select filters, I want to deselect the currently open place such that I see places that match my filters. Why would I otherwise want to adjust my filters?

native/src/components/PlaceFiltersModal.tsx
steffenkleinle: Good catch! Fixed by unmounting when the modal gets closed.

native/src/routes/Places.tsx
```diff
@@ -138,16 +149,15 @@ const Places = ({ refresh, localHistory, initialZoom, places: allPlaces, regionM
...
-        modalVisible={showFilterSelection}
-        closeModal={() => updateShowFilterSelection(false)}
-        placeCategories={placeCategories}
-        selectedPlaceCategory={placeCategory}
-        setSelectedPlaceCategory={updatePlaceCategoryFilter}
-        currentlyOpenFilter={currentlyOpen}
-        setCurrentlyOpenFilter={updatePlaceCurrentlyOpenFilter}
-        placesCount={places.length}
-      />
+      {showFilterSelection && (
```
bahaaTuffaha, line 152: This is genius 🧐 ... better than useEffect I guess.

bahaaTuffaha (approved):
Thank you for the quick fix 👍 
Tested on Android and iOS. ✅

native/src/routes/Places.tsx
steffenkleinle: Thanks! :D I was very torn between both, since useEffects would be explicit while this is a very implicit solution. But I'll add a unit test :)

PR #4242, opened 2026-07-07
4104: Not available language dialogs feedback link

### LINKED ISSUE #4104: Link to feedback in translation not available dialog
### Describe the Problem

We currently show a dialog when the users presses on a language with no translation in the language change dialog, which tells the user to use the feedback to request a new language. However, since there is no direct option to go to the feedback, its more complicated to send feedback then it would need to be.

Link to feedback in language not available dialog (See https://github.com/digitalfabrik/integreat-app/pull/4050#discussion_r3116164093) with `feedback=negative`
 
N/A

### Related Issues

Blocked by: #4103 
Discussed in: https://github.com/digitalfabrik/integreat-app/pull/4050#discussion_r3116164093

### Additional Information

N/A

bahaaTuffaha:
### Short Description

We currently show a dialog when the users presses on a language with no translation in the language change dialog, which tells the user to use the feedback to request a new language.

### Proposed Changes

- Wrapped feedback in `<1></1>` tags inside translation.
- Created a new component for both web and native called `LanguageNotAvailableMessage` just to handle the react-i18next text replacement.
- Adjusted `SimpleAlertDialogProps` so I can also pass a `ReactElement` for LanguageNotAvailableMessage.
- For native :
    - I had to pass `slug` and `previousRouteType` so I adjusted `RootRoutesParamsType` at `NavigationTypes.ts`.
    - At Header.tsx => goToLanguageChange function: I passed previousRouteType and slug when navigating to languages so it knows from which screen that feedback came from.
- To render the link when feedback is possible I created `feedbackAvailable` prop for web and for native I just checks for canGiveFeedback if regionCode and previousRouteType are available.
- Enabled feedbackAvailable just for chat (language change) and RegionContentHeader and the rest should be disabled due to the unavailability of the feedback like at the landing/regions page.

### Side Effects

N/A

### Checklist

### Testing

- Launch integreat. go to any region.
- Test language not available button and it should show a feedback link.
- Test where the feedback is not available : regions page, langauge failure page (when you select a language and switch to region that doesn't have that selected language.
- Test native and web.

### Resolved Issues

Fixes: #4104

---


web/src/components/LanguageNotAvailableMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Trans } from 'react-i18next'
+
+import { RATING_NEGATIVE, toQueryParams } from 'shared'
+
+import Link from './base/Link'
+
+const StyledLink = styled(Link)(({ theme }) => ({
+  color: theme.palette.primary.main,
+  textDecoration: 'underline',
+}))
```
steffenkleinle: 🔧 Please use the `highlighted` prop instead of applying the styling yourself.

web/src/components/LanguageNotAvailableMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+  close: () => void
+}
+
+const LanguageNotAvailableMessage = ({
+  feedbackAvailable = false,
+  close,
+}: LanguageNotAvailableMessageProps): ReactElement => (
+  <Trans i18nKey='layout:languageNotAvailableMessage'>
+    This gets replaced
+    {feedbackAvailable ? (
```
steffenkleinle, line 19: 💭  The problem is that we are still promoting sending feedback, even though it is not technically possible. I guess we should come up with a way to send feedback, even though no valid region or language is selected. However, I can see that this is out of scope for this PR. Do you perhaps want to open an issue?

native/src/components/LanguageNotAvailableMessage.tsx
```diff
@@ -0,0 +1,61 @@
...
+import styled from 'styled-components/native'
+
+import { FEEDBACK_MODAL_ROUTE, LanguagesRouteType } from 'shared'
+import { FeedbackRouteType } from 'shared/api'
+
+import { NavigationProps } from '../constants/NavigationTypes'
+import { AppContext } from '../contexts/AppContext'
+import Text from './base/Text'
+
+const FeedbackLink = styled(Text)`
```
steffenkleinle: ❌ Please use the `Link` component instead for consistent styling and a11y behavior. You obviously need to adjust the corresponding component, but something like the following should work:
``` tsx
type LinkTargetProps = { url: string; onPress?: never } | { url?: never; onPress: () => void }

type LinkProps = LinkTargetProps & {
  children: string
  style?: StyleProp<TextStyle>
}

...

return <Text onPress={onPress ?? (() => openExternalUrl(url, showSnackbar))} />
```

native/src/constants/NavigationTypes.ts
```diff
@@ -129,6 +129,8 @@ export type RootRoutesParamsType = {
   [LANGUAGES_ROUTE]: {
     languages: LanguageModel[]
     availableLanguages: string[]
+    previousRouteType?: FeedbackRouteType
```
steffenkleinle: 🤔 I'd keep the naming consistent and just use `routeType`

web/src/components/LanguageNotAvailableMessage.tsx
bahaaTuffaha: we can pass default values or we create new general feedback endpoint.
either way I will create an issue for it 👍

lunars97 (commented):
Almost looks good 👍 Currently on native, the negative feedback is not set by default like it is done on web.

native/src/components/LanguageNotAvailableMessage.tsx
```diff
@@ -33,6 +33,7 @@ const LanguageNotAvailableMessage = ({
         language: languageCode,
         regionCode,
         slug,
+        noResults: true,
```
bahaaTuffaha: what do you think of this fix? @steffenkleinle

steffenkleinle: No objections to the solution itself, but then we should also use corresponding naming imo:
🔧 
```suggestion
        rating: 'negative',
```

steffenkleinle (approved):
Nice work!

lunars97 (approved):
Thank you for your effort! LGTM. Tested in native and web 👍

PR #4251, opened 2026-07-13
4250: Add query for Contrast mode

### LINKED ISSUE #4250: Use correct theme for chat on native
### Describe the Problem

Since the chat on native is only an embedded webview with the web chat, the native theme is not correctly applied and always uses the light theme. This should be fixed.

- [ ] Add a query param `theme` with values `light`/`dark` on web
- [ ] Use the theme specified in the query param and only resort to the local storage value as fallback
- [ ] Pass the selected native theme to the chat

N/A

### Related Issues

#4209 

### Additional Information

https://chat.tuerantuer.org/digitalfabrik/pl/6msi99mq1p895p9qd7nxt7y33h

bahaaTuffaha:
### Short Description

Since the chat on native is only an embedded webview with the web chat, the native theme is not correctly applied and always uses the light theme. This should be fixed.

### Proposed Changes

- Adjusted `query.ts` to add a new key for theme.
- Added `theme` to `ParamsType` in `RouteInformationTypes.ts`.
- Extended `getChatUrl` in `/utils/url.ts` to accept an optional `theme` parameter.
- Added a `useEffect` that runs once: if a valid `?theme=` query param is present, it persists it to localStorage via `setThemeType` and strips it from the URL with `replaceState`.
- Removed the theme query after it gets processed to it wont intercept with the current localStorage logic.

### Side Effects

N/A

### Checklist

### Testing

- Launch Native and Web (integreat).
- Replace the Url at `/native/src/utils/url.ts` with what you have from the web yarn start ip for example: `http://192.168.1.34:9001`.
- At native go to settings and change contrast then open chat at testumgebung.
- Test iOS.
- Test web with `?theme=contrast` or `?theme=light`.

### Resolved Issues

Fixes: #4250

---


native/src/utils/url.ts
```diff
@@ -21,17 +21,20 @@ export const getChatUrl = ({
   regionCode,
   languageCode,
   chatId,
+  theme,
 }: {
   regionCode: string
   languageCode: string
   chatId?: string
+  theme?: string
```
steffenkleinle: 🔧 Please restrict to possible values
```suggestion
  theme?: ThemeType
```

shared/src/routes/query.ts
```diff
@@ -49,6 +53,7 @@ export type VisibilityQueryParams = {
 
 type QueryParams = VisibilityQueryParams & {
   chatId?: string
+  theme?: string
```
steffenkleinle: 🔧 Please restrict to allowed values
```suggestion
  theme?: ThemeType
```

web/src/components/ThemeContainer.tsx
```diff
@@ -152,11 +153,23 @@ type ThemeContainerProps = {
 
 const ThemeContainer = ({ children, contentDirection }: ThemeContainerProps): ReactElement => {
   const dimensions = useDimensions()
+  const url = new URL(window.location.href)
+  const { theme: queryTheme } = parseQueryParams(url.searchParams)
   const [themeType, setThemeType] = useLocalStorage<ThemeKey>({
```
steffenkleinle: 🔧 
```suggestion
  const [storageThemeType, setThemeType] = useLocalStorage<ThemeKey>({
```

web/src/components/ThemeContainer.tsx
```diff
@@ -152,11 +153,23 @@ type ThemeContainerProps = {
...
 
+  useEffect(() => {
+    // process theme from query param into localStorage and remove it from the URL
+    if (queryTheme === 'light' || queryTheme === 'contrast') {
+      setThemeType(queryTheme)
+      url.searchParams.delete('theme')
+      window.history.replaceState({}, '', url.href)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
steffenkleinle: 🔧 No need to persist this to storage:
```suggestion
  const themeType = queryTheme ?? storageTheme
```

steffenkleinle (changes requested):
Nice work! When opening the chat, there is initially a white screen showing up. Could you perhaps fix this by setting a background color in the WebView?

steffenkleinle:
Perhaps you could add a release note?

web/src/components/ThemeContainer.tsx
bahaaTuffaha: 🤔 do I change `setThemeType` to `setStorageThemeType` ?

web/src/components/ThemeContainer.tsx
bahaaTuffaha: Without it if I added ?theme=contrast and then changed the theme using the contrast button it won't change.
here I'm registering the theme first run then remove the theme query.

steffenkleinle: Hmm I see. From my point of view, this was really only meant as a solution for the native chat, where there is no way to change the theme since the header is disabled, so this would not be necessary for that use case. As long as we don't have another use case where passing a custom theme is necessary, I'd keep it as simple as possible.

bahaaTuffaha: sure

steffenkleinle (approved):
Not tested again

lunars97 (approved):
LGTM! Tested on iOS and android emulators 👍

Docmale:
I can't see any change, the chat is still not affected by the contrast mode. Is the right version included in this release? tested on Android real device and emulator

lunars97:
Seems like it is not working in main branch :thinking: I tested it with ios simulator
UPDATE: Tested by setting network ip and it worked. I wonder why it is not working on real device :thinking:

bahaaTuffaha:
> I can't see any change, the chat is still not affected by the contrast mode. Is the right version included in this release? tested on Android real device and emulator

The web changes are not applied to production yet and I think the native is fetching the production one in the webview unless I'm wrong...

lunars97:
> > I can't see any change, the chat is still not affected by the contrast mode. Is the right version included in this release? tested on Android real device and emulator
> 
> The web changes are not applied to production yet and I think the native is fetching the production one in the webview unless I'm wrong...

might be :eyes:

steffenkleinle:
What @bahaaTuffaha says, this can only be tested once the changes are on production.

Docmale:
I will check this on production

PR #4265, opened 2026-07-16
3621: Improve the UI for frag integreat

### LINKED ISSUE #3621: Improve the ui for FragIntegreat (chat modal)
### Describe the Problem

For the chat popup I think there is a plenty of room to modernize it more. Also it's position in the middle of the screen feels like it's in the way.

An example for inspiration: https://dribbble.com/shots/25543081-Ai-Assistant

- Still under discussion.

- Still under discussion.

### Related Issues

N/A.

bahaaTuffaha:
### Short Description

Our current chat interface has grown organically and is not really consistent/optimal in regards to the UI. Some parts take excessive space making it harder to display all meaningful information while other features are not highlighted properly.

### Proposed Changes

- Added 2 props to `Dialog` component `subtitle`, `icon` and restructured the title area into a `Stack` of `icon`.
- Restyled `StyledDialog` so the header has rounded top corners, no box-shadow and  a `divider` border to look like the design as much as possible.
- For the chat messages : I changed the background colors and padding from `8` to `16`, added border and adjusted the line-height to match 14/20.
- Switched the bot avatar to render the app logo instead and kept `SmartToyOutlinedIcon` as fallback.
- Removed the "ask one complete question per message" info `Alert` and it's `CHAT_HINT_VISIBLE_STORAGE_KEY`.
- Added new translations: `subtitle`, `welcomeText`.
- Added `chat` and `chatMessage` background colors.
- Added an optional `minimize` prop that switches the desktop close button to a `RemoveIcon`.
- Added a new `app-icon-inverted.svg` for Malte.

### Side Effects

N/A

### Checklist

### Testing

- Launch Integreat's web app.
- Open the frag integreat chat and check the design at [Design](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=6175-1899&t=OQdJ7sL2LLX2DJ41-1) for comparison .
- Check mobile screens and contrast mode.
- Test malte's chat Icon at testumgebung.

### Resolved Issues

Fixes: #3621

---


build-configs/src/common/theme/colors.ts
```diff
@@ -40,6 +40,8 @@ export const commonLightColors: CommonColorPalette = {
     default: '#FFFFFF',
     paper: '#FFFFFF',
     accent: '#EAEEF9',
+    chat: '#FFFFFF',
+    chatMessage: '#EAEEF9',
```
steffenkleinle: 🔧 Please move this to its own block similar to `ttsPlayer`

build-configs/src/malte/assets/app-icon-inverted.svg
steffenkleinle, line 1: 🤔 I am not sure about this, I think it looks a little weird/not correctly aligned: 

Image

@ulyanahoy-cloud any other ideas for an icon for the chat for malte?

web/src/components/HeaderActionItem.tsx
```diff
@@ -9,10 +9,12 @@ import Link from './base/Link'
 const StyledButton = styled(MuiButton)`
   padding: 2px 12px;
   border: 1px solid;
+  background-color: ${props => (props.theme.isContrastTheme ? undefined : props.theme.palette.background.chat)};
```
steffenkleinle: ❓ This looks really weird, why are we setting the chat color to a general header action item, but only on light theme? If this is actually intended, the color should a) not be named `chat` and b) be used for both light and dark theme.
Same below.

web/src/components/ChatMessage.tsx
```diff
@@ -13,25 +12,30 @@ import { useTranslation } from 'react-i18next'
...
-  border: `1px solid ${theme.palette.divider}`,
+  padding: 16,
   borderRadius: 8,
   wordBreak: 'break-word',
+  ...(theme.isContrastTheme
+    ? {
+        backgroundColor: userIsAuthor ? theme.palette.background.accent : theme.palette.background.chatMessage,
+        border: `1px solid ${theme.palette.divider}`,
+      }
+    : { backgroundColor: userIsAuthor ? theme.palette.grey[100] : theme.palette.background.accent }),
```
steffenkleinle: ❌ I think we should try to avoid these `if (isContrastTheme)` checks. Our theme should be considered in a way, that the colors can be used for both themes without any check.

web/src/components/ChatMessage.tsx
```diff
@@ -13,25 +12,30 @@ import { useTranslation } from 'react-i18next'
...
+import { AppLogoIcon, ChatAvatar } from './ChatAvatar'
 import RemoteContent from './RemoteContent'
 
-export const Message = styled('div')(({ theme }) => ({
+export const Message = styled('div', { shouldForwardProp })<{ userIsAuthor: boolean }>(({ theme, userIsAuthor }) => ({
   maxWidth: '70%',
   width: 'max-content',
-  padding: 8,
-  border: `1px solid ${theme.palette.divider}`,
+  padding: 16,
```
steffenkleinle: ❓ @ulyanahoy-cloud do we actually want a bigger padding for all messages? I think this takes up valuable space and looks especially weird for short messages. I'd be in favor of keeping the old `8px` padding. Same for the new `16px` gap.

Image

web/src/components/ChatMessage.tsx
```diff
@@ -42,16 +46,26 @@ type MessageAvatarProps = {
 const MessageAvatar = ({ userIsAuthor, isAutomaticAnswer, visible }: MessageAvatarProps): ReactElement => {
```
steffenkleinle: 🔧 Let's move this component to `ChatAvatar`, I think it belongs there better and makes this file more readable.

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  '& svg': {
+    display: 'block',
+  },
+})
+
+type AppLogoIconProps = {
+  size: number
+}
+
+export const AppLogoIcon = ({ size }: AppLogoIconProps): ReactElement | null => {
```
steffenkleinle: 🔧 To avoid confusion perhaps rather
```suggestion
export const ChatLogo = ({ size }: ChatLogoProps): ReactElement | null => {
```

web/src/components/ChatContainer.tsx
```diff
@@ -25,17 +29,37 @@ import useQueryParamVisibility from '../hooks/useQueryParamVisibility'
...
+      padding: theme.spacing(1),
+    },
+  },
   [`.${dialogContentClasses.root}`]: {
-    padding: '0 0 16px',
+    padding: theme.spacing(0, 0, 2),
+    backgroundColor: theme.palette.background.chat,
   },
-})
+}))
```
steffenkleinle, line 54: ❓ @ulyanahoy-cloud should we apply (some of) this styling from the chat to all dialogs in the app if we want to keep a consistent UI?

web/src/components/ChatContainer.tsx
```diff
@@ -25,17 +29,37 @@ import useQueryParamVisibility from '../hooks/useQueryParamVisibility'
...
 import { LanguageChangePath } from './LanguageSelection'
 import Dialog from './base/Dialog'
 
-const StyledDialog = styled(Dialog)({
+const HEADER_BORDER_RADIUS = 12
+const AVATAR_ICON_SIZE = 40
+
+const StyledDialog = styled(Dialog)(({ theme }) => ({
+  [`.${dialogClasses.paper}:not(.${dialogClasses.paperFullScreen})`]: {
+    borderRadius: `${HEADER_BORDER_RADIUS}px ${HEADER_BORDER_RADIUS}px 0 0`,
```
steffenkleinle: 🤔 In the design all four dialog corners are rounded (which imo also looks a lot better)

web/src/components/ChatContainer.tsx
```diff
@@ -25,17 +29,37 @@ import useQueryParamVisibility from '../hooks/useQueryParamVisibility'
...
+    boxShadow: 'none',
+    border: `1px solid ${theme.palette.divider}`,
+  },
+  [`.${dialogClasses.paper} > .${stackClasses.root}`]: {
+    margin: 0,
+    padding: theme.spacing(1),
+    backgroundColor: theme.isContrastTheme ? theme.palette.background.chat : theme.palette.background.accent,
```
steffenkleinle: 🔧 Same as before, please avoid these conditionals

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,37 @@
...
+}))
+
+const CenteredSvg = styled(Svg)({
+  '& svg': {
+    display: 'block',
+  },
+})
+
+type AppLogoIconProps = {
+  size: number
```
steffenkleinle: 🤔 What do you think of setting the default to `40` to avoid the constants and needing to pass it? Same above for `ChatAvatar`.

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,37 @@
+import Avatar from '@mui/material/Avatar'
+import { styled, useTheme } from '@mui/material/styles'
+import React, { ReactElement } from 'react'
+
+import buildConfig from '../constants/buildConfig'
+import Svg from './base/Svg'
+
+export const ChatAvatar = styled(Avatar, { shouldForwardProp: prop => prop !== 'size' })<{
```
steffenkleinle: 🔧 I am getting
```
Received `true` for a non-boolean attribute `visible`.

If you want to write it to the DOM, pass a string instead: visible="true" or visible={value.toString()}.
```
Please use the helper from `@emotion/is-prop-valid` instead:
```suggestion
export const ChatAvatar = styled(Avatar, { shouldForwardProp })<{
```

steffenkleinle: 🔧 Also, shouldn't this be the default export at the end of the file?

web/src/components/ChatContainer.tsx
```diff
@@ -119,14 +144,24 @@ const ChatContainer = ({ region, languageCode, languageChangePaths }: ChatContai
...
+    const appLogo = buildConfig().icons.appLogoMobileInverted
     const chatLanguageChangePaths =
       languageChangePaths?.map(({ path, ...rest }) => ({ ...rest, path: openUrl(path) })) ?? []
 
     return (
       <StyledDialog
         title={chatName}
+        subtitle={t('subtitle')}
+        icon={
+          appLogo ? (
```
steffenkleinle: 🤔 Could we just make this a required prop and set it for all build configs?
🔧 Alternatively, since these are only ever used together, what do you think of changing `AppLogoIcon` to directly include it, i.e. change the former `AppLogoIcon` to:
``` tsx
export const ChatLogoAvatar = ({ size, visible }: ChatLogoAvatarProps): ReactElement | null => {
  const theme = useTheme()
  const appLogo = buildConfig().icons.appLogoMobileInverted

  if (!appLogo) {
    return null
  }

  return (
    <ChatAvatar size={size} visible={visible}>
      <CenteredSvg src={appLogo} width={size} height={size} overrideFillColors={theme.palette.background.paper} />
    </ChatAvatar>
  )
}
```

steffenkleinle (changes requested):
Thanks a lot for your work on this, I think it looks great! Tested on firefox.

web/src/components/HeaderActionItem.tsx
bahaaTuffaha: Yes this is intended. As you can see in the design the change language button's background color becomes white while it's light mode and transparent when it's contrast.

bahaaTuffaha: if I used `props.theme.palette.chat.background` it will affect the button at the main header

web/src/components/ChatMessage.tsx
bahaaTuffaha: The trade-off will look like this:


Screenshot 2026-07-17 at 3 26 34 PM

web/src/components/ChatContainer.tsx
bahaaTuffaha: I got confused with the header... I will adjust it 👍

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  if (!appLogo) {
+    return null
+  }
+
+  return (
+    <ChatAvatar size={size} visible={visible}>
+      <ChatLogo size={size} />
+    </ChatAvatar>
+  )
+}
```
steffenkleinle: 🔧 These two components can be merged, right?

web/src/components/HeaderActionItem.tsx
```diff
@@ -9,10 +9,12 @@ import Link from './base/Link'
 const StyledButton = styled(MuiButton)`
   padding: 2px 12px;
   border: 1px solid;
+  background-color: ${props => props.theme.palette.chat.headerActionBackground};
```
steffenkleinle: 🔧 To avoid using a `chat` color for a generally used component, I'd inline the color in this case:
```suggestion
  background-color: ${props => props.theme.isContrastTheme ? theme.palette.background.default : 'transparent'};
```
Same below. The color can then be removed from the theme.

build-configs/src/common/theme/colors.ts
```diff
@@ -55,6 +55,14 @@ export const commonLightColors: CommonColorPalette = {
     background: '#dedede',
     playIconColor: '#007aa8',
   },
+  chat: {
+    background: '#FFFFFF',
+    headerBackground: '#EAEEF9',
+    messageBackground: '#EAEEF9',
+    userMessageBackground: '#F5F5F5',
```
steffenkleinle: 🙃 For consistency/explicitness:
```suggestion
    answerMessageBackground: '#EAEEF9',
    userMessageBackground: '#F5F5F5',
```

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  if (userIsAuthor) {
+    return <ChatAvatar visible={visible} aria-label={t('user')} />
+  }
+
+  let avatarIcon: ReactElement
+  if (isAutomaticAnswer && appLogo) {
+    avatarIcon = <ChatLogo />
+  } else if (isAutomaticAnswer) {
+    avatarIcon = <SmartToyOutlinedIcon />
```
steffenkleinle: 🤔 Wouldn't it make more sense to just make `appLogoMobileInverted` required, add an inverted logo for all build configs and avoid this fallback?

web/src/components/ChatAvatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+type MessageAvatarProps = {
+  userIsAuthor: boolean
+  isAutomaticAnswer: boolean
+  visible: boolean
+}
+
+export const MessageAvatar = ({ userIsAuthor, isAutomaticAnswer, visible }: MessageAvatarProps): ReactElement => {
+  const { t } = useTranslation('chat')
+  const label = t(isAutomaticAnswer ? 'bot' : 'consultant')
+  const appLogo = buildConfig().icons.appLogoMobileInverted
```
steffenkleinle: 🤔 Why is this called `mobile`? Isn't it used on both mobile and desktop?

web/src/components/HeaderActionItem.tsx
bahaaTuffaha: I thought we are avoiding conditionals?
Screenshot 2026-07-20 at 1 52 33 PM

web/src/components/ChatAvatar.tsx
bahaaTuffaha: yes that's a good question.

web/src/components/HeaderActionItem.tsx
steffenkleinle: Yeah, but we should also avoid setting a `chat` color on a general component, thats worse imo. I couldn't think of another easy solution, so I'd keep the conditional in this case.

web/src/components/ChatMessage.tsx
ulyanahoy-cloud: We can keep 8px, I'm good with it.

build-configs/src/malte/assets/app-icon-inverted.svg
ulyanahoy-cloud: Are we allowed to change it? Maybe we could just use the star on its own...

steffenkleinle: @FritjofK @svenseeberg do you know if this would be an acceptable change to the icon?

svenseeberg: Generally, I would assume that we are not allowed without explicit consent. Usually, these brands are protected. In this specific case however, the logo is not protected. So yes, we could do that. In this case it would mostly be about satisfying the customer.

ulyanahoy-cloud: In that case, I'd suggest using a generic MUI icon or creating a simple custom icon instead of reusing the brand icon. A simple chat bubble could work, or even just an "F" for Frag Integreat if we want to keep it minimal.

bahaaTuffaha: How about we keep the logo for integreat and the rest just the "F" ?
do you have a design for that "F" for example (shape of the letter):
Screenshot 2026-07-21 at 1 19 10 PM

steffenkleinle: Let's just wait until we hear back from the maltesers (some time next week) instead of spending time on this. Until then, we can either
a) put this PR on hold
b) remove the icon here and add the solution they want in a follow-up PR
I'd be in favor of b).

web/src/components/ChatContainer.tsx
ulyanahoy-cloud: I think it makes sense to apply the general dialog styling to all dialogs to keep the UI consistent.

build-configs/src/malte/assets/app-icon-inverted.svg
osmers: Discussion on the icon here: https://chat.tuerantuer.org/digitalfabrik/pl/147s31uhnpfwpewqu5ij6h9o5y

svenseeberg: @ulyanahoy-cloud you can choose any icon you like. But we are not allowed to use the cross without the shield.

ulyanahoy-cloud: What do you think about using an abstract, neutral icon instead?

image

steffenkleinle (approved):
Great work, looks amazing!

lunars97 (approved):
Looks really nice! Tested with Safari 😻 Thank youu

kalairubachandrakumaran-oss:
tested ✅ 
image

PR #4285, opened 2026-07-28
4064: Announce dynamic content updates

### LINKED ISSUE #4064: Announce dynamic content updates to screen reader users
### Describe the Problem

Currently dynamic content updates are not announced to screen readers, for ex. contrast mode being changed, link being copied, language or city being changed and so on.

Check where it is necessary and apply screen-reader announcement

N/A

### Related Issues

similar to #3984

### LINKED ISSUE #4223: Accessibility issues with screen reader
### Describe the Problem
Android: 
- The screen reader says “Frag Integreat” several times before reading the chatbot content
- The whole chatbot area is treated as one element instead of separate elements
    -> Because of this, users cannot move through the chatbot response or select individual text
    -> The question input field is not accessible, so users cannot enter a question with the screen reader

iOS: 
Some buttons, such as the three-dot menu, are difficult to select
 -> The focus sometimes moves to another button instead
 ->  In some cases, the Back button gets the focus, and activating it closes the chatbot by mistake

Windows:
- When the chatbot is opened, the screen reader first reads several header elements instead of the chatbot
- After a question is entered, the welcome message is shown but is not announced by the screen reader
- When the focus returns to the question input field, the screen reader sometimes reads the header again instead of staying in the chatbot
- When navigating the chatbot header, the screen reader sometimes announces content from the page behind the chatbot (for example, the page title “Local Information”) instead of staying within the chatbot

Android:
- The chatbot content, buttons, and input field should be accessible as separate elements
- Screen reader users should be able to move through the content and use the chatbot without problems

iOS:
- All buttons should receive the correct screen reader focus
- All buttons should be easy to select and activate

 Windows:
- The screen reader should focus on the chatbot when it is opened
- New chatbot messages should be announced automatically
- The focus should stay inside the chatbot while it is open
- The screen reader should not read content from the page behind the chatbot

### Related Issues
#3984 

### Additional Information
During testing of PR #3984  , the following accessibility observations were found when using a screen reader

https://github.com/user-attachments/assets/c3d28c9b-3fb4-4463-bd2c-800767a0f243
https://github.com/user-attachments/assets/95df9d15-5ade-478b-96aa-aac96d9f2d47
https://github.com/user-attachments/assets/7f854ab5-a415-4a31-8ad8-10a16f8b085b

### LINKED ISSUE #4297: Fix TTS player covering the bottom nav
### Describe the Bug

Currently the TTS player is not above the bottom nav like web.

### Steps to Reproduce

1. Launch Integreat on native.
2. Open any page.
3. Toggle the tts from the side menu.
4. The tts player is covering the bottom nav.

### Expected Behavior

The tts player should be above the bottom nav.

### Environment

- Device/Browser: Android
- OS version: API 36.1
- App version: Main

### Related Issues

N/A

### Additional Information

Image

lunars97:
### Short Description

This PR implements announcement for dynamic content which is either updated automatically or in response to a user action, to make sure that users are notified of this update in an accessible way.

### Proposed Changes

- apply additional aria attributes where it is necessary
- add announcement hook to language change in web and native and make use of LiveAnnouncer.tsx component
- useTouchableOpacity instead of react-native-paper `FAB` to fix the chat fab button not being recognized by screen-reader also relates to #4223 

### Side Effects

Uses react-native TouchableOpacity instead of react-native-paper FAB for Chat Button

### Checklist

### Testing

- Enable screen-reader
- Go to language selection
- Select a language and wait for language selection announcement through screen-reader
- Go to Feedback in native and choose useful/not useful toggle button, the state should be announced and after the feedback is sent, the success message should be announced
- Header menu in native and web should announce when the dropdown menu extended 
- Also check TTS player, it should not cover the bottom navigation anymore
- Settings item list should properly announce a change of the toggle state, the same for External Sources settings screen in settings

### Resolved Issues

Fixes: #4064 #4297 
Parially fixes #4223 

---


web/src/components/GeneralHeader.tsx
```diff
@@ -30,6 +33,14 @@ const GeneralHeader = ({ languageCode, regionLanguages }: GeneralHeaderProps): R
     path: `/${slug}/${language.code}`,
   }))
 
+  useEffect(() => {
+    if (previousLanguage.current !== languageCode) {
+      const languageName = languageChangePaths.find(l => l.code === languageCode)?.name ?? languageCode
+      setAnnouncement(`${t('languageChangedTo')} ${languageName}`)
+      previousLanguage.current = languageCode
+    }
+  }, [languageCode, languageChangePaths, t])
```
bahaaTuffaha: I think there is a potential for it to be a hook 🤔

lunars97: really nice idea 👍

native/src/components/base/ToggleButton.tsx
```diff
@@ -63,7 +63,12 @@ const ToggleButton = ({ text, onPress, icon, active, style }: ToggleButtonProps)
         },
         style,
       ]}>
-      <TouchableRipple role='switch' onPress={onPress} style={styles.TouchableRippleStyle} borderless>
+      <TouchableRipple
+        role='button'
```
lunars97, line 67: Replaced to button to know the state via screen-reader, otherwise the state is not known

native/src/components/ChatFab.tsx
```diff
@@ -60,17 +72,16 @@ const ChatFab = ({ style }: ChatFabProps): ReactElement => {
   return (
     <Container style={style}>
       <ChatHighlightPopup chatName={getChatName(buildConfig().appName)} />
-      <FAB
-        icon='forum-outline'
+      <PressableFab
```
lunars97: Switched to Pressable while with screen-reader enabled it is not possible to click on chat fab...

native/src/constants/index.ts
```diff
@@ -1,4 +1,6 @@
 export const ROOT_NAVIGATOR_ID = 'rootNavigator'
+
+export const TAB_HEIGHT = 60
```
bahaaTuffaha: hmm how about moving it to `dimensions from './constants/dimensions'` like bottomNavigationHeight: 60 .

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -55,6 +55,8 @@ const HeaderLanguageSelectorItem = ({
       text={isOpen ? '' : t('changeLanguage') /* to not cover the dropdown with the tooltip */}
       icon={<TranslateOutlinedIcon />}
       innerText={forceText || desktop ? currentLanguageName : undefined}
+      aria-haspopup='listbox'
```
bahaaTuffaha: ❓ does it actually reach the DOM... I'm not sure if `HeaderActionItem` can take this prop.

native/src/components/ChatFab.tsx
```diff
@@ -21,6 +23,16 @@ const Container = styled.View`
   align-items: flex-end;
 `
 
+const PressableFab = styled(Pressable)`
```
bahaaTuffaha: ❓  Does the TouchableRipple work with the screen reader?

web/src/hooks/useAnnounceLanguageChange.ts
bahaaTuffaha, line 1: The name of the file is a bit different than `useLanguageChangeAnnouncement`.

bahaaTuffaha (commented):
- At `page.tsx` we have `SpaceForTts` this needs to be adjusted because currently it covers the last update date. I think you need just add to the ttsPlayerHeight + bottomNavigationHeight .

Image

- Not sure about `aschaffenburg` cuz it doesn't have a bottom nav.

native/src/components/ChatFab.tsx
lunars97: I havent checked it, I will have a look :+1:

web/src/components/HeaderLanguageSelectorItem.tsx
lunars97: hmm when I was testing with screen-reader, i think there was a difference :eyes: I will have a look :+1:

native/src/constants/index.ts
lunars97: good idea!

web/src/hooks/useAnnounceLanguageChange.ts
lunars97: oops :eyes:

native/src/components/ChatFab.tsx
lunars97: I think Pressable is more robust and future-proof. See react-native highlighted that `If you're looking for a more extensive and future-proof way to handle touch-based input, check out the [Pressable](https://reactnative.dev/docs/pressable) API.` https://reactnative.dev/docs/touchableopacity Maybe we should also slowly switch completely to Pressable :thinking:

bahaaTuffaha: But Pressable doesn't show any feedback when you press on it...

web/src/components/HeaderLanguageSelectorItem.tsx
lunars97: you are right, it is redundant

native/src/routes/LanguageSelection.tsx
```diff
@@ -77,6 +77,7 @@ const LanguageSelection = ({ navigation, route }: LanguageSelectionProps): React
             if (code !== languageCode) {
               changeLanguageCode(code)
             }
+            AccessibilityInfo.announceForAccessibility(`${t('languageChangedTo')} ${name}`)
```
bahaaTuffaha: This is already good but english voice for example can't speak some languages so how about we use what is already at accessibilityLabel : `userLanguageNames?.of(code) ?? name`

native/src/components/ChatFab.tsx
bahaaTuffaha, line 1: Let's hide the chatFab when the Tts player is open because it's under the player and can't be open anyway.

native/src/components/Page.tsx
```diff
@@ -12,7 +12,7 @@ const Container = styled.View<{ $padding: boolean }>`
   ${props => props.$padding && `padding: 0 ${dimensions.pageContainerPaddingHorizontal}px 8px;`}
 `
 const SpaceForTts = styled.View<{ $ttsPlayerVisible: boolean }>`
-  height: ${props => (props.$ttsPlayerVisible ? dimensions.ttsPlayerHeight : 0)}px;
+  height: ${props => (props.$ttsPlayerVisible ? dimensions.ttsPlayerHeight + dimensions.bottomNavigationHeight : 0)}px;
```
bahaaTuffaha, line 15: There is a space at the bottom at aschaffenburg:
Image

bahaaTuffaha (changes requested):

native/src/components/Page.tsx
lunars97: Thanks for catching it :see_no_evil:

native/src/routes/LanguageSelection.tsx
lunars97: I will check it :+1:

native/src/components/ChatFab.tsx
lunars97: let's ask @ulyanahoy-cloud, whether she has other ideas? @ulyanahoy-cloud what do you think about removing chat button or fitting them together by making the dialog smaller when the `read aloud player` is activated? Currently chat button is underneath of the `read aloud player` dialog :thought_balloon: 
image

native/src/components/Page.tsx
lunars97: hmm I dont have any solution for this, except for removing additional height and leaving it to cover the last update date :thinking:

bahaaTuffaha: sorry I think the issue is not from here its from the `BottomTabNavigator.tsx`
```js
sceneStyle: bottomTabsVisible
            ? undefined
            : { paddingBottom: insets.bottom + dimensions.bottomNavigationHeight },
```

lunars97: yes, I mean this. I added it to add a height to not cover the last update date, but then it breaks the whole view when there are no navigation tabs...

bahaaTuffaha: I can think of 2 possible solutions for the spacing at aschaffenburg:

1) at TtsPlayer.tsx
```js
  insetBottom={bottom + (bottomTabsVisible ? dimensions.bottomNavigationHeight : 0)}
  ```
  
  now the difficulty is where can we get the `bottomTabsVisible`:
  ```js
  const { regionCode } = useAppContext()
  const { data: regions } = useLoadRegions()
  
   const region = regions?.find(it => it.code === regionCode)
  const bottomTabsVisible = !!(
    region?.placesEnabled ||
    region?.localNewsEnabled ||
    region?.tuNewsEnabled ||
    region?.eventsEnabled
  )
  ```
  
2) Or create a custom hook for bottom tab navigation visibility that has a context.
3) OR maybe `buildConfig().featureFlags.fixedRegion` but a fixed region could have a bottom nav right?

File: native/src/components/ChatFab.tsx
Comment (ulyanahoy-cloud) [reply native/src/components/ChatFab.tsx:1]: I'd also hide the FAB while the player is expanded.

File: native/src/components/Page.tsx (hunk 8 above)
Comment (lunars97) [reply native/src/components/Page.tsx:15]: I will try this out, thank you :green_heart:

Comment (lunars97) [reply native/src/components/Page.tsx:15]: I guess the 3rd option is the simplest and robust :)

Comment (bahaaTuffaha) [reply native/src/components/Page.tsx:15]: Unless they add a tab 👀 ... so it needs a comment just incase.

Comment (lunars97) [reply native/src/components/Page.tsx:15]: hmm i have't thougth about, good point :+1:

File: native/src/components/TtsPlayer.tsx [hunk 9]
```diff
@@ -87,6 +87,7 @@ const TtsPlayer = ({
 
   return (
     <StyledTtsPlayer
+      // There shouldn't be an additional bottom navigation height for a fixed region, for ex. aschaffenburg, as it does not have bottom tab navigation
```
lunars97, line 90: I am not sure whether it delivers the message 😬

bahaaTuffaha: yeah I think it's enough

bahaaTuffaha (approved):
Great work ✅ 
Tested on Android and iOS ✅

web/src/hooks/useAnnounceLanguageChange.ts
```diff
@@ -0,0 +1,22 @@
...
+  languageChangePaths: { code: string; name: string }[] | null,
+  t: (key: string) => string,
+): string => {
+  const previousLanguage = useRef(languageCode)
+  const [announcement, setAnnouncement] = useState('')
+
+  useEffect(() => {
+    if (previousLanguage.current !== languageCode) {
+      const languageName = languageChangePaths?.find(l => l.code === languageCode)?.name ?? languageCode
+      setAnnouncement(`${t('languageChangedTo')} ${languageName}`)
```
steffenkleinle: 🔧 The `languageName` should be directly included in the translation and passed as parameter. The placement of the name might be different in other languages.

web/src/hooks/useAnnounceLanguageChange.ts
```diff
@@ -0,0 +1,22 @@
...
+  languageCode: string,
+  languageChangePaths: { code: string; name: string }[] | null,
+  t: (key: string) => string,
+): string => {
+  const previousLanguage = useRef(languageCode)
+  const [announcement, setAnnouncement] = useState('')
+
+  useEffect(() => {
+    if (previousLanguage.current !== languageCode) {
+      const languageName = languageChangePaths?.find(l => l.code === languageCode)?.name ?? languageCode
```
steffenkleinle: 🔧 Please avoid abbreviations like this

web/src/hooks/useAnnounceLanguageChange.ts
```diff
@@ -0,0 +1,22 @@
+import { useRef, useEffect, useState } from 'react'
+
+const useAnnounceLanguageChange = (
+  languageCode: string,
+  languageChangePaths: { code: string; name: string }[] | null,
+  t: (key: string) => string,
```
steffenkleinle: 🔧 No need to pass `t` here, just use the hook directly here.

web/src/hooks/useAnnounceLanguageChange.ts
```diff
@@ -0,0 +1,22 @@
+import { useRef, useEffect, useState } from 'react'
+
+const useAnnounceLanguageChange = (
```
steffenkleinle: 🔧 We already have a `usePreviousProps` hook that should be used here. Perhaps no custom hook is necessary then anymore.

web/src/components/GeneralHeader.tsx
steffenkleinle, line 40: ❓ Wouldn't it be more fitting to include the announcement directly in `HeaderLanguageSelectorItem`, which is in the end the place where the language changing happens? This way, we also would not need the same code twice.

web/src/components/HeaderMenu.tsx
```diff
@@ -113,10 +113,15 @@ const HeaderMenu = ({ children, pageTitle, fitScreen, ref }: HeaderMenuProps): R
...
+      <IconButton
+        onClick={openMenu}
+        aria-label={t('sideBarOpenAriaLabel')}
+        aria-expanded={open}
+        aria-haspopup='menu'
+        aria-controls={open ? 'side-menu' : undefined}>
         <MoreVertIcon />
       </IconButton>
-      <StyledMenu anchorEl={menuAnchorElement} open={open} onClose={closeMenu}>
+      <StyledMenu role='menu' id='side-menu' anchorEl={menuAnchorElement} open={open} onClose={closeMenu}>
```
steffenkleinle: 🤔 Why `side-menu`? AFAICS, we don't have a sidebar menu anymore, only a popup :)
```suggestion
      <StyledMenu role='menu' id='header-menu' anchorEl={menuAnchorElement} open={open} onClose={closeMenu}>
```

translations/src/translations.json
```diff
@@ -7623,6 +7624,7 @@
       "getOnPlayStore": "Im Google Play Store anschauen",
       "accessibility": "Barrierefreiheit",
       "readAloud": "Vorlesefunktion",
+      "readAloudPlayer": "Vorlese-Player",
```
steffenkleinle: 🤔 What do you think of changing `readAloud` to `Vorlesen` and using that as a11y label as well? 
```suggestion
      "readAloud": "Vorlesen",
```

translations/src/translations.json
```diff
@@ -7912,6 +7914,7 @@
       "localInformationLabel": "Information",
       "events": "Events",
       "news": "News",
+      "languageChangedTo": "Language changed to",
```
steffenkleinle: 🔧 See my other comment
```suggestion
      "languageChanged": "Language changed to {{name}}",
```

native/src/components/ChatFab.tsx
```diff
@@ -60,17 +72,18 @@ const ChatFab = ({ style }: ChatFabProps): ReactElement => {
...
+        accessibilityRole='button'
+        activeOpacity={0.9}
         onPress={() => navigation.navigate(CHAT_ROUTE)}
         accessibilityLabel={getChatName(buildConfig().appName)}
-        variant='primary'
-        style={{ backgroundColor: theme.colors.primary }}
-        size='medium'
-        aria-label={getChatName(buildConfig().appName)}
-      />
+        testID='fab'>
```
steffenkleinle: 🤔 Is the `testID` really necessary?

native/src/components/ChatFab.tsx
```diff
@@ -21,6 +23,16 @@ const Container = styled.View`
   align-items: flex-end;
 `
 
+const TouchableFab = styled(TouchableOpacity)`
```
steffenkleinle, line 26: 🤔 Would this also work with `Pressable`?
From the react-native docs:
> If you're looking for a more extensive and future-proof way to handle touch-based input, check out the [Pressable](https://reactnative.dev/docs/pressable) API.

steffenkleinle (commented):
Looks very good! Just a few minor comments :)

web/src/hooks/useAnnounceLanguageChange.ts
lunars97: nice point! :100:

native/src/components/ChatFab.tsx
lunars97: yes, I also wanted to have `Pressable` as it says it is more future-proof in the documentation, see the comment at the very above from @bahaaTuffaha about it not having a feedback...

native/src/components/ChatFab.tsx
lunars97: it is used in the test file, I think I can replace it with accessibility label :thinking:

web/src/components/GeneralHeader.tsx
lunars97: good point, I think I tried it before but something was off. Seems like it works now :+1:

native/src/components/ChatFab.tsx
```diff
@@ -60,17 +72,15 @@ const ChatFab = ({ style }: ChatFabProps): ReactElement => {
...
+        activeOpacity={0.9}
         onPress={() => navigation.navigate(CHAT_ROUTE)}
-        accessibilityLabel={getChatName(buildConfig().appName)}
-        variant='primary'
-        style={{ backgroundColor: theme.colors.primary }}
-        size='medium'
-        aria-label={getChatName(buildConfig().appName)}
-      />
+        accessibilityLabel={getChatName(buildConfig().appName)}>
+        <Icon size={24} source='forum-outline' color={theme.colors.background} />
```
bahaaTuffaha: I think the icon was in white color at contrast mode before.

Image

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -85,6 +95,7 @@ const HeaderLanguageSelectorItem = ({
 
   return (
     <>
+      <LiveAnnouncer message={announcement} />
```
bahaaTuffaha, line 100: ❓  So it will render only for large screens?

bahaaTuffaha (approved):
LGTM! ✅ 
I have found some small issues:

1- The settings works well if moved by swiping horizontally but not by hovring like other components do:

https://github.com/user-attachments/assets/6442a0a8-5dcb-46db-aceb-f9aea022e99d

2- I cant reach the checkbox at frag integreat while enabling talkback (maybe because its a webview 🤔 )

Image

native/src/constants/index.ts
```diff
@@ -1,4 +1,5 @@
 export const ROOT_NAVIGATOR_ID = 'rootNavigator'
+
```
bahaaTuffaha: there is an empty space here

lunars97:
> LGTM! ✅ I have found some small issues:
> 
> 1- The settings works well if moved by swiping horizontally but not by hovring like other components do:
> 
>  2026-09-01.19-33-54.mov 
> 
> 2- I cant reach the checkbox at frag integreat while enabling talkback (maybe because its a webview 🤔 )
> 
> Image

Thank you for your review :)

- Regarding the `1- The settings works well if moved by swiping horizontally but not by hovring like other components do:` I will check it 👍 
- Regarding the `2- I cant reach the checkbox at frag integreat while enabling talkback (maybe because it's a webview 🤔 )` Sadly it is because of the webview...

lunars97:
> LGTM! ✅ I have found some small issues:
> 
> 1- The settings works well if moved by swiping horizontally but not by hovring like other components do:
> 
>  2026-09-01.19-33-54.mov 
> 2- I cant reach the checkbox at frag integreat while enabling talkback (maybe because its a webview 🤔 )
> 
> Image

- The issue `1- The settings works well if moved by swiping horizontally but not by hovring like other components do: ` has been resolved by removing ScrollView, I noticed that there is a similar issue with nested FlatList in category list items (created an issue #4360 for it)

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -47,6 +51,11 @@ const HeaderLanguageSelectorItem = ({
   }
 
   const currentLanguageName = languageChangePaths.find(item => item.code === languageCode)?.name
+  useEffect(() => {
+    if (previousLanguageCode !== languageCode) {
+      setAnnouncement(t($ => $.layout.languageChanged, { name: currentLanguageName ?? languageCode }))
+    }
+  }, [previousLanguageCode, languageCode, currentLanguageName, t])
```
steffenkleinle: 🔧 This should also work without an additional useEffect/useState:

```suggestion
  const announcement = t($ => $.layout.languageChanged, { name: currentLanguageName ?? languageCode })
```
If you actually would need a `useState` for some reason, you could use the `onPropChange` prop of `usePreviousProp` to at least get rid of the `useEffect`. However, I don't think either is necessary.

steffenkleinle (approved):
LGTM, not tested

web/src/components/HeaderLanguageSelectorItem.tsx
lunars97: replaced it with `usePreviousProp`, works like a charm :+1:

native/src/components/Feedback.tsx
```diff
@@ -61,7 +61,7 @@ const Feedback = ({
 
   if (sendingStatus === 'successful') {
     return (
-      <Wrapper>
+      <Wrapper accessible accessibilityRole='alert' accessibilityLiveRegion='assertive'>
```
bahaaTuffaha, line 64: ❓  will this get announced on iOS?

native/src/BottomTabNavigator.tsx
bahaaTuffaha, line 50: unused constant.

web/src/components/HeaderMenu.tsx
```diff
@@ -113,10 +113,15 @@ const HeaderMenu = ({ children, pageTitle, fitScreen, ref }: HeaderMenuProps): R
...
+      <IconButton
+        onClick={openMenu}
+        aria-label={t($ => $.layout.sideBarOpenAriaLabel)}
+        aria-expanded={open}
+        aria-haspopup='menu'
+        aria-controls={open ? 'header-menu' : undefined}>
         <MoreVertIcon />
       </IconButton>
-      <StyledMenu anchorEl={menuAnchorElement} open={open} onClose={closeMenu}>
+      <StyledMenu role='menu' id='header-menu' anchorEl={menuAnchorElement} open={open} onClose={closeMenu}>
```
bahaaTuffaha, line 124: `role='menu'` is already set by MUI.

bahaaTuffaha (approved):
Tested on Android. ✅ 

Is there a way to test iOS voiceOver? the accessibility inspector wont capture these announcements.

web/src/components/HeaderLanguageSelectorItem.tsx
```diff
@@ -70,6 +81,7 @@ const HeaderLanguageSelectorItem = ({
   if (mobile) {
     return (
       <>
+        <LiveAnnouncer message={announcement} />
```
steffenkleinle, line 84: 🤔 Does it make sense to show a snackbar instead of just announcing this to a11y? Maybe user feedback would generally be a good thing here and it would also be announced automatically to screen readers, right? Same for native.

lunars97: Hmm not really sure about this idea, it might also be disturbing for a user who doesn't use screen-reader :thinking: :thought_balloon:

native/src/components/Feedback.tsx
lunars97: Hmm I assume it should work but I will try to check it with ios real device, thank you for mentioning it :green_heart:

web/src/components/HeaderLanguageSelectorItem.tsx
steffenkleinle: Why is it disturbing for users without screen readers to show a normal confirmation snackbar that the language change was successful? But I am okay with not showing one, so lets just keep it as it is.

translations/src/translations/de.json
```diff
@@ -58,7 +58,8 @@
       "hideFilters": "Filter verbergen"
     },
     "a11y": {
-      "skipToContent": "Zum Inhalt springen"
+      "skipToContent": "Zum Inhalt springen",
+      "languageChanged": "Sprache wurde auf {{languageName}} umgestellt"
```
steffenkleinle: 🔧 The translation should be moved as `changed` to the `languages` namespace

native/src/components/Feedback.tsx
lunars97: @bahaaTuffaha Unfortunately I was not able to test it with real device, but VoiceOver works in web, so this might be a good sign for native as well? :thinking:

bahaaTuffaha: Then lets keep it...

PR #4290, opened 2026-07-28
3689: Adjust display of date for recurring events

### LINKED ISSUE #3689: Adjust display of date for recurring events
### Describe the Problem

https://github.com/digitalfabrik/integreat-app/issues/2978 introduced a new design for events. Part of that change includes that weekly recurring events are shown with "ab dd.mm.yyyy" where dd.mm.yyyy is the closest date in the future when the event takes place. 

This implies that it is the first time that event takes place and also does not really show well how often it is recurring.

Display a collapsible "Weitere Termine" section for recurring events directly in the event list.

The section should show the next 5 upcoming occurrences. If more future occurrences exist, display (...) after the last visible date.

Design [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=6427-1542&t=PDeP2WFKCgzMgHFS-1)

### Related Issues

Duplicated in: https://github.com/digitalfabrik/integreat-app/issues/3299
Related: #2986 

### Additional Information
N/A

bahaaTuffaha:
### Short Description

Display a collapsible "Weitere Termine" section for recurring events directly in the event list.

The section should show the next 5 upcoming occurrences. If more future occurrences exist, display (...) after the last visible date. Design [here](https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=6427-1542&t=PDeP2WFKCgzMgHFS-1)

### Proposed Changes

- Added a `shortFormat` parameter to `formatMonthlyOrYearlyRecurrence`, defaulting to `false`; when `true`, it uses `dateFormatShort` instead of `dateFormatWithWeekday`.

Web:
- Added a new component `EventFurtherDates` that renders the upcoming recurrence dates.
-  Used the existing `Accordion` component for `EventFurtherDates.tsx`.
- Showed up to `MAX_FURTHER_DATES` (6) recurrences via `date.recurrences()`, then dropped the first one with `.slice(1)` since it's already shown in the eventList Item.
- Used `preventDefault`/`stopPropagation` on the wrapper's click handler so toggling the accordion doesn't trigger the parent list item's navigation link.
- Added a new icons `EventNoteIcon`,  `LocationOnOutlinedIcon` for the date and location at `EventListItem.tsx`.

Native:
- Built the toggle with a `TouchableRipple` instead of using the accordion because it needs alot of changes to match the design.
- Used the same `hasMoreRecurrencesThan` logic like web.
- Imported and rendered `EventFurtherDates` directly below the event's date line.
- New Icons like web: `calendar-text-outline`, `map-marker-outline` for location and date.
------------
- Added a `furtherDates` translation key.
- Create new tests for web and native:`EventFurtherDates.spec.tsx`

### Side Effects

N/A

### Checklist

### Testing

- Launch integreat.
- Go to Events list and check the ones that has the recurrence date icon.
- Check web and native.
- Check contrast and RTL languages alignment.

### Resolved Issues

Fixes: #3689

---

### DIFF
--- a/native/src/components/EventFurtherDates.tsx
+++ b/native/src/components/EventFurtherDates.tsx
@@ -9,8 +9,6 @@ import { contentAlignment, contentDirection } from '../constants/contentDirectio
 import Icon from './base/Icon'
 import Text from './base/Text'
 
-const MAX_FURTHER_DATES = 6
-
 const Toggle = styled(TouchableRipple)<{ language: string }>`
   align-self: ${props => (contentDirection(props.language) === 'row-reverse' ? 'flex-end' : 'flex-start')};
   padding-block: 8px;
@@ -23,7 +21,8 @@ const ToggleContent = styled.View<{ language: string }>`
 `
 
 const Dates = styled.View`
-  padding-inline-start: 32px;
+  padding-inline-start: 24px;
+  gap: 4px;
 `
 
 type EventFurtherDatesProps = {
@@ -36,15 +35,9 @@ const EventFurtherDates = ({ date, language }: EventFurtherDatesProps): ReactEle
   const theme = useTheme()
   const [expanded, setExpanded] = useState(false)
 
-  const furtherDates = date.isMonthlyOrYearlyRecurrence() ? date.recurrences(MAX_FURTHER_DATES).slice(1) : []
+  const furtherDates = date.furtherDates()
   if (furtherDates.length === 0) {
     return null
-  }
-  const hasMoreDates = date.hasMoreRecurrencesThan(MAX_FURTHER_DATES)
-
-  const formatFurtherDate = (recurrence: DateModel): string => {
-    const { date: formattedDate, time } = recurrence.formatMonthlyOrYearlyRecurrence(language, t, true)
-    return `${formattedDate} · ${time}`
   }
 
   return (
@@ -65,18 +58,16 @@ const EventFurtherDates = ({ date, language }: EventFurtherDatesProps): ReactEle
       </Toggle>
       {expanded && (
         <Dates>
-          {furtherDates.map((recurrence, index) => {
-            const isLast = index === furtherDates.length - 1
-            const formattedDate = formatFurtherDate(recurrence)
-            return (
-              <Text
-                key={recurrence.startDate.toISO()}
-                variant='body3'
-                style={{ textAlign: contentAlignment(language) }}>
-                {isLast && hasMoreDates ? `${formattedDate} …` : formattedDate}
-              </Text>
-            )
-          })}
+          {furtherDates.map(recurrence => (
+            <Text key={recurrence.startDate.toISO()} variant='body3' style={{ textAlign: contentAlignment(language) }}>
+              {recurrence.formatFurtherDate(language, t)}
+            </Text>
+          ))}
+          {date.hasMoreFurtherDates() && (
+            <Text variant='body3' style={{ textAlign: contentAlignment(language) }}>
+              …
+            </Text>
+          )}
         </Dates>
       )}
     </>
--- a/native/src/components/EventFurtherDates.tsx
+++ b/native/src/components/EventFurtherDates.tsx
@@ -1,0 +1,85 @@
+import React, { ReactElement, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { TouchableRipple } from 'react-native-paper'
+import styled, { useTheme } from 'styled-components/native'
+
+import { DateModel } from 'shared/api'
+
+import { contentAlignment } from '../constants/contentDirection'
+import Icon from './base/Icon'
+import Text from './base/Text'
+
+const MAX_FURTHER_DATES = 6
+
+const Toggle = styled(TouchableRipple)`
+  align-self: flex-start;
+  padding: 8px;
+`
+
+const ToggleContent = styled.View`
+  flex-direction: row;
+  align-items: center;
+  gap: 4px;
+`
+
+const Dates = styled.View`
+  padding-inline-start: 32px;
+`
+
+type EventFurtherDatesProps = {
+  date: DateModel
+  language: string
+}
+
+const EventFurtherDates = ({ date, language }: EventFurtherDatesProps): ReactElement | null => {
+  const { t } = useTranslation('events', { lng: language })
+  const theme = useTheme()
+  const [expanded, setExpanded] = useState(false)
+
+  const furtherDates = date.isMonthlyOrYearlyRecurrence() ? date.recurrences(MAX_FURTHER_DATES).slice(1) : []
+  if (furtherDates.length === 0) {
+    return null
+  }
+  const hasMoreDates = date.hasMoreRecurrencesThan(MAX_FURTHER_DATES)
+
+  const formatFurtherDate = (recurrence: DateModel): string => {
+    const { date: formattedDate, time } = recurrence.formatMonthlyOrYearlyRecurrence(language, t, true)
+    return `${formattedDate} · ${time}`
+  }
+
+  return (
+    <>
+      <Toggle
+        borderless
+        accessibilityRole='button'
+        accessibilityState={{ expanded }}
+        onPress={() => setExpanded(previous => !previous)}>
+        <ToggleContent>
+          <Icon source='repeat' color={theme.colors.primary} />
+          <Text variant='body3' style={{ color: theme.colors.primary }}>
+            {t('furtherDates')}
+          </Text>
+          <Icon source={expanded ? 'chevron-up' : 'chevron-down'} color={theme.colors.primary} />
+        </ToggleContent>
+      </Toggle>
+      {expanded && (
+        <Dates>
+          {furtherDates.map((recurrence, index) => {
+            const isLast = index === furtherDates.length - 1
+            const formattedDate = formatFurtherDate(recurrence)
+            return (
+              <Text
+                key={recurrence.startDate.toISO()}
+                variant='body3'
+                style={{ textAlign: contentAlignment(language) }}>
+                {isLast && hasMoreDates ? `${formattedDate} …` : formattedDate}
+              </Text>
+            )
+          })}
+        </Dates>
+      )}
+    </>
+  )
+}
+
+export default EventFurtherDates
--- a/native/src/components/EventFurtherDates.tsx
+++ b/native/src/components/EventFurtherDates.tsx
@@ -5,19 +5,19 @@ import styled, { useTheme } from 'styled-components/native'
 
 import { DateModel } from 'shared/api'
 
-import { contentAlignment } from '../constants/contentDirection'
+import { contentAlignment, contentDirection } from '../constants/contentDirection'
 import Icon from './base/Icon'
 import Text from './base/Text'
 
 const MAX_FURTHER_DATES = 6
 
-const Toggle = styled(TouchableRipple)`
-  align-self: flex-start;
-  padding: 8px;
+const Toggle = styled(TouchableRipple)<{ language: string }>`
+  align-self: ${props => (contentDirection(props.language) === 'row-reverse' ? 'flex-end' : 'flex-start')};
+  padding-block: 8px;
 `
 
-const ToggleContent = styled.View`
-  flex-direction: row;
+const ToggleContent = styled.View<{ language: string }>`
+  flex-direction: ${props => contentDirection(props.language)};
   align-items: center;
   gap: 4px;
 `
@@ -50,12 +50,13 @@ const EventFurtherDates = ({ date, language }: EventFurtherDatesProps): ReactEle
   return (
     <>
       <Toggle
+        language={language}
         borderless
         accessibilityRole='button'
         accessibilityState={{ expanded }}
         onPress={() => setExpanded(previous => !previous)}>
-        <ToggleContent>
-          <Icon source='repeat' color={theme.colors.primary} />
+        <ToggleContent language={language}>
+          <Icon source='repeat' size={16} color={theme.colors.primary} />
           <Text variant='body3' style={{ color: theme.colors.primary }}>
             {t('furtherDates')}
           </Text>
--- a/native/src/components/EventFurtherDates.tsx
+++ b/native/src/components/EventFurtherDates.tsx
@@ -1,8 +1,10 @@
 import React, { ReactElement, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { View } from 'react-native'
 import { TouchableRipple } from 'react-native-paper'
 import styled, { useTheme } from 'styled-components/native'
 
+import { MAX_FURTHER_DATES_MOBILE } from 'shared'
 import { DateModel } from 'shared/api'
 
 import { contentAlignment, contentDirection } from '../constants/contentDirection'
@@ -22,7 +24,7 @@ const ToggleContent = styled.View<{ language: string }>`
 
 const Dates = styled.View`
   padding-inline-start: 24px;
-  gap: 4px;
+  gap: 8px;
 `
 
 type EventFurtherDatesProps = {
@@ -35,10 +37,17 @@ const EventFurtherDates = ({ date, language }: EventFurtherDatesProps): ReactEle
   const theme = useTheme()
   const [expanded, setExpanded] = useState(false)
 
-  const furtherDates = date.furtherDates()
+  const furtherDates = date.furtherDates(MAX_FURTHER_DATES_MOBILE)
   if (furtherDates.length === 0) {
     return null
   }
+
+  const formattedRecurrences = furtherDates.map(recurrence => ({
+    key: recurrence.startDate.toISO(),
+    ...recurrence.formatMonthlyOrYearlyRecurrence(language, t, true),
+  }))
+  const firstRecurrenceTime = formattedRecurrences[0]?.time
+  const hasVaryingTimes = formattedRecurrences.some(recurrence => recurrence.time !== firstRecurrenceTime)
 
   return (
     <>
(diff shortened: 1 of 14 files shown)


web/src/components/EventListItem.tsx
```diff
@@ -93,9 +98,18 @@ const EventListItem = ({
             </Typography>
           }
           secondary={
-            <StyledTypography variant='body1' flexDirection='column' component='div'>
-              <p>{dateToDisplay.formatEventDateInOneLine(languageCode, t)}</p>
-              {event.location && <p>{event.location.name}</p>}
+            <StyledTypography variant='body1' flexDirection='column' component='div' dir={contentDirection}>
+              <Stack direction='row' alignItems='center' gap={1} component='p'>
+                <EventNoteOutlinedIcon fontSize='small' />
```
steffenkleinle: 🔧 I'd stick to the normal calendar icon, so `import EventIcon from '@mui/icons-material/Event'`

web/src/components/EventListItem.tsx
```diff
@@ -93,9 +98,18 @@ const EventListItem = ({
...
+                <span>{dateToDisplay.formatEventDateInOneLine(languageCode, t)}</span>
```
steffenkleinle, line 108: ❌ I think this still needs to be adjusted, we currently still only show something like `From August 4`, not the complete first recurrence as it is in the design. That looks especially weird if expanding more dates, since the next recurrence is than the only one without a time.

Image

Image

web/src/components/EventListItem.tsx
```diff
@@ -93,9 +98,18 @@ const EventListItem = ({
...
+              </Stack>
+              <EventFurtherDates date={event.date} languageCode={languageCode} />
```
steffenkleinle, line 110: 🔧 I think this uses too much vertical padding

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,94 @@
...
+type EventFurtherDatesProps = {
+  date: DateModel
+  languageCode: string
+}
+
+const EventFurtherDates = ({ date, languageCode }: EventFurtherDatesProps): ReactElement | null => {
+  const { t } = useTranslation('events')
+  const { contentDirection } = useTheme()
+
+  const furtherDates = date.isMonthlyOrYearlyRecurrence() ? date.recurrences(MAX_FURTHER_DATES).slice(1) : []
```
steffenkleinle: ❓ Why only for monthly or yearly recurrences?

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,94 @@
...
+              {t('furtherDates')}
+            </StyledText>
+          </Stack>
+        }>
+        {furtherDates.map((recurrence, index) => {
+          const isLast = index === furtherDates.length - 1
+          const formattedDate = formatFurtherDate(recurrence)
+          return (
+            <StyledText key={recurrence.startDate.toISO()} variant='body1'>
+              {isLast && hasMoreDates ? `${formattedDate} …` : formattedDate}
```
steffenkleinle: 🤔 I know this is like this in the design, but what do you think of instead just rendering a separate `StyledText` for the dots in a separate row instead of appending it to the last recurrence?
```tsx
            {hasMoreDates ? <StyledText>…</StyledText>
```

native/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { TouchableRipple } from 'react-native-paper'
+import styled, { useTheme } from 'styled-components/native'
+
+import { DateModel } from 'shared/api'
+
+import { contentAlignment, contentDirection } from '../constants/contentDirection'
+import Icon from './base/Icon'
+import Text from './base/Text'
+
+const MAX_FURTHER_DATES = 6
```
steffenkleinle: 🔧 Please move this to a shared constant in `shared`

native/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const furtherDates = date.isMonthlyOrYearlyRecurrence() ? date.recurrences(MAX_FURTHER_DATES).slice(1) : []
+  if (furtherDates.length === 0) {
+    return null
+  }
+  const hasMoreDates = date.hasMoreRecurrencesThan(MAX_FURTHER_DATES)
+
+  const formatFurtherDate = (recurrence: DateModel): string => {
+    const { date: formattedDate, time } = recurrence.formatMonthlyOrYearlyRecurrence(language, t, true)
+    return `${formattedDate} · ${time}`
+  }
```
steffenkleinle: 🤔 Couldn't this be shared between web and native in `shared` if you directly return the formatted further dates?

steffenkleinle (changes requested):

web/src/components/EventFurtherDates.tsx
bahaaTuffaha: because the weekly dosen't show any dates when you open it's EventPage. but I think we can implement it too for weekly.

steffenkleinle: I think I'd keep it consistent, at least I can't see any reason why this should be handled differently.

web/src/components/EventListItem.tsx
bahaaTuffaha: EventNoteOutlinedIcon matches more but I don't mind.

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const stopLinkNavigation = (event: MouseEvent): void => {
+    event.preventDefault()
+    event.stopPropagation()
+  }
+
+  return (
+    <AccordionWrapper dir={contentDirection} onClick={stopLinkNavigation}>
+      <Accordion
+        id='further-dates'
```
lunars97, line 87: could we make the heading component of `./base/Accordion` dynamic by using maybe in our case just a `span`? Currently it is set to `h2` and it used in our case after `h3` which is against semantic elements. See the screenshot below: 

Image

lunars97 (changes requested):
Almost perfect! Tested with real android device and Chrome. I found one issue, please see my comment 👍

bahaaTuffaha:
@ulyanahoy-cloud 
For German it looks great but in english it looks like this:

Screenshot 2026-08-04 at 7 23 01 PM

Screenshot 2026-08-04 at 7 23 14 PM

Is this ok?

ulyanahoy-cloud:
What if we only show the date on mobile if the time stays the same, and limit the list to the next 3 occurrences? This would keep the card much more compact.
I would also slightly increase the spacing between the event title and the date.


image

bahaaTuffaha:
> What if we only show the date on mobile if the time stays the same, and limit the list to the next 3 occurrences? This would keep the card much more compact. I would also slightly increase the spacing between the event title and the date.

How about this for mobile and one line for desktop:
Screenshot 2026-08-05 at 12 34 39 PM

ulyanahoy-cloud:
If the recurring event always has the same time, I'd still prefer showing only the dates on mobile. It keeps the card much more compact and avoids repeating the same information.

 If the time differs between occurrences, then this two-line layout works well.

bahaaTuffaha:
> If the recurring event always has the same time, I'd still prefer showing only the dates on mobile. It keeps the card much more compact and avoids repeating the same information.
> 
> If the time differs between occurrences, then this two-line layout works well.

Ok and for the occurrences keep them 5 or 3 just for mobile?

ulyanahoy-cloud:
for mobile 3

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,123 @@
...
+  if (furtherDates.length === 0) {
+    return null
+  }
+
+  const formattedRecurrences = furtherDates.map(recurrence => ({
+    key: recurrence.startDate.toISO(),
+    ...recurrence.formatMonthlyOrYearlyRecurrence(languageCode, t, true),
+  }))
+  const firstRecurrenceTime = formattedRecurrences[0]?.time
+  const hasVaryingTimes = formattedRecurrences.some(recurrence => recurrence.time !== firstRecurrenceTime)
```
lunars97: do we need to check for varying times? Don't we have the same time for recurring events and not recurring as well? 🤔

bahaaTuffaha: hmm I checked again at the production and cms. You are correct I will remove this logic. 🙈

lunars97: thank you, I also got confused and checked cms :)

lunars97 (approved):
Not tested again. LGTM 👍

web/src/components/EventFurtherDates.tsx
lunars97: @bahaaTuffaha fyi https://github.com/digitalfabrik/integreat-cms/issues/2815 There is an open issue to add the varying times so maybe we need to keep it 🤔

bahaaTuffaha: > @bahaaTuffaha fyi [digitalfabrik/integreat-cms#2815](https://github.com/digitalfabrik/integreat-cms/issues/2815) There is an open issue to add the varying times so maybe we need to keep it 🤔

🤔 sure I will put it back 👍

web/src/components/EventListItem.tsx
```diff
@@ -45,6 +49,10 @@ const StyledTypography = styled(Typography)<TypographyProps>`
     margin: 0;
   }
 
+  p:first-of-type {
+    margin-top: 8px;
+  }
+
   p:nth-of-type(2) {
     margin-bottom: 4px;
   }
```
steffenkleinle: :question: Is there any reason why we don't just set the margin on the parent component? Imo this would be a lot clearer.
```suggestion
  margin-top: 8px;
```
Edit: The `margin-bottom` is intended to add some space between the second last item and the excerpt/description. However, this approach also only works if both location and a date is set. For that to work, we should instead use a custom spacing/padding just on the description.
``` tsx
  const Excerpt = styled('p')({
    margin-top: 4px;
  })
```

web/src/hooks/useQueryParamVisibility.ts
```diff
@@ -18,13 +18,14 @@ const useQueryParamVisibility = <T extends keyof VisibilityQueryParams>(key: T):
   const visible = queryParams.has(key) && queryParams.get(key) !== 'false'
 
   const open = (value?: VisibilityQueryParams[T]) => {
-    const newQueryParams = queryParams
+    // The query params of the current location must not be mutated with new URLSearchParams, as they are memoized by react-router
```
steffenkleinle: ❓ Why does the comment say `must not be mutated with new URLSearchParams`, but that is exactly what is then done afterwards? Or am I missing something? Do you have some link or description what happens otherwise?

shared/src/api/models/DateModel.ts
```diff
@@ -129,14 +131,27 @@ class DateModel {
...
+  }
+
+  hasMoreFurtherDates(count = MAX_FURTHER_DATES): boolean {
+    return this.hasMoreRecurrencesThan(count)
+  }
+
+  hasVaryingTimes(locale: string, t: TranslateFunction, count = MAX_FURTHER_DATES): boolean {
+    const times = this.recurrences(count).map(recurrence => formatTime(locale, recurrence, t))
+    return times.some(time => time !== times[0])
+  }
```
steffenkleinle: ❓ Is it really necessary to pass `locale` and `t` to decide if all recurrences have the same time? 🤔

shared/src/constants/index.ts
```diff
@@ -1,4 +1,6 @@
 export const MAX_DATE_RECURRENCES = 3
+export const MAX_FURTHER_DATES = 6
+export const MAX_FURTHER_DATES_MOBILE = 4
```
steffenkleinle: 🔧 I think this is very misleading, the constants should reflect how many dates are actually displayed. Should be changed to 5 and 3 imo and handled programmatically to still return the correct amount of dates.
```suggestion
export const MAX_FURTHER_DATES = 5
export const MAX_FURTHER_DATES_MOBILE = 3
```

shared/src/api/models/DateModel.ts
```diff
@@ -173,19 +188,12 @@ class DateModel {
...
     }
-    const finalDate = this.getFinalDate(this)
-    if (finalDate) {
-      return this.formatDateInterval(locale, finalDate, { showYear })
-    }
-    return t('startingFrom', {
-      date: this.startDate.toLocaleString(format, { locale }),
-    })
+
+    // long-term event
```
steffenkleinle, line 197: 🙃 Imo `long-term event` is very unspecific and depending on subjective criteria. At least I would not expect a two day event to be `long-term`. Additionally, we also use `isSingleOneDayEvent` above. Would be nice if you could align the naming in other comments as well. 🙏 
```suggestion
    // Multi-day event
```

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,121 @@
...
+  '& .MuiAccordionSummary-root': {
+    minHeight: 40,
+  },
+  '& .MuiAccordionSummary-content': {
+    margin: 0,
+  },
+  '& .MuiAccordionSummary-expandIconWrapper': {
+    color: theme.palette.primary.main,
+  },
+  '& .MuiAccordionDetails-root': {
```
steffenkleinle: 🤔 Are there predefined keys you could use instead of hardcoding the classes?

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const formattedRecurrences = furtherDates.map(recurrence => ({
+    key: recurrence.startDate.toISO(),
+    ...recurrence.formatMonthlyOrYearlyRecurrence(languageCode, t, true),
+  }))
+  const showTime = desktop || date.hasVaryingTimes(languageCode, t, maxFurtherDates)
+
+  const stopLinkNavigation = (event: MouseEvent): void => {
+    event.preventDefault()
+    event.stopPropagation()
```
steffenkleinle: 🤔 Is both `preventDefault` and `stopPropagation` needed? I can't see a difference with or without `stopPropagation`.

web/src/components/EventFurtherDates.tsx
```diff
@@ -0,0 +1,121 @@
...
+        defaultCollapsed
+        headingComponent='span'
+        title={
+          <Stack direction='row' alignItems='center' gap={1}>
+            <RepeatIcon color='primary' fontSize='small' />
+            <StyledText color='primary' variant='body1'>
+              {t('furtherDates')}
+            </StyledText>
+          </Stack>
+        }>
```
steffenkleinle, line 100: 🔧 The title gets wider if the content is wider, which leads to a not that nice spacing between the title and the caret. Furthermore, the dot between date and time is missing.
Image
Image

steffenkleinle (commented):
Nice work, looks very good! A few more comments :)

web/src/hooks/useQueryParamVisibility.ts
bahaaTuffaha: Yes this comment is terrible.
before: it mutates the react-router object so it will affect the memoized instance.
After: it mutates a new object.

I added a test for this case at `useQueryParamVisibility.spec.tsx`:
`should stay visible when the same query param is set again after closing`

I found this article read the pitfalls part: 
https://www.dhiwise.com/blog/design-converter/a-complete-guide-to-setsearchparams-in-react-router

also https://reactrouter.com/api/hooks/useSearchParams#notes

shared/src/api/models/DateModel.ts
bahaaTuffaha: hmm do you mean to format it to just time like: `recurrence.startDate.toFormat('HH:mm')`?

steffenkleinle: Would be a possibility, but maybe too complicated. But you could also just pass a hardcoded locale and a `key => key` `t` function.

web/src/components/EventFurtherDates.tsx
bahaaTuffaha: Isn't the accordion expected to expand with the content?

steffenkleinle: Hmm, I guess that is a question for @ulyanahoy-cloud then, in the design its at least different:
image

shared/src/api/models/DateModel.ts
```diff
@@ -140,15 +140,15 @@ class DateModel {
...
 
   hasMoreFurtherDates(count = MAX_FURTHER_DATES): boolean {
-    return this.hasMoreRecurrencesThan(count)
+    return this.hasMoreRecurrencesThan(count + 1)
   }
 
-  hasVaryingTimes(locale: string, t: TranslateFunction, count = MAX_FURTHER_DATES): boolean {
-    const times = this.recurrences(count).map(recurrence => formatTime(locale, recurrence, t))
+  hasVaryingTimes(count = MAX_FURTHER_DATES): boolean {
+    const times = this.recurrences(count + 1).map(recurrence => recurrence.startDate.toFormat('HH:mm'))
```
steffenkleinle: ❌ This only checks the `startDate` (i.e. time), but not the end. So if one event recurrence would take longer/shorter, this would not be caught.

web/src/hooks/useQueryParamVisibility.ts
steffenkleinle: Thanks for explaining. But then lets adjust the comment in a more meaningful way. Or, maybe even better: Let's just remove the comment, we can't comment every usage of this :)

steffenkleinle (approved):
Nice work! Not tested again :)

kalairubachandrakumaran-oss:
@bahaaTuffaha , @steffenkleinle 
Is it intentonal that the year is shown on mobile but not on web?
image

bahaaTuffaha:
> @bahaaTuffaha , @steffenkleinle Is it intentonal that the year is shown on mobile but not on web? image

can you check if the System's date is correct and matching current year?

steffenkleinle:
> > @bahaaTuffaha , @steffenkleinle Is it intentonal that the year is shown on mobile but not on web? image
> 
> can you check if the System's date is correct and matching current year?

Yes, the year should only show if it is different than the current calendar year

kalairubachandrakumaran-oss:
> > > @bahaaTuffaha , @steffenkleinle Is it intentonal that the year is shown on mobile but not on web? image
> > 
> > 
> > can you check if the System's date is correct and matching current year?
> 
> Yes, the year should only show if it is different than the current calendar year

For me the year is also shown for the current year

9B179AAB-34C1-4C68-A946-682C62143B1D
72F8124E-C517-4AE0-ABA2-A77C4A3016CB
41DB6A46-5798-42A5-9346-F9F0B8AD28BA

steffenkleinle:
> > > > @bahaaTuffaha , @steffenkleinle Is it intentonal that the year is shown on mobile but not on web? image
> > > 
> > > 
> > > can you check if the System's date is correct and matching current year?
> > 
> > 
> > Yes, the year should only show if it is different than the current calendar year
> 
> For me the year is also shown for the current year

Hmm, on current main this is fixed for me. In the latest version I can reproduce it. Perhaps you can check if this is fixed in the next release? Thanks!

PR #4313, opened 2026-08-06
4310: Improve region search results for dashes

### LINKED ISSUE #4310: Region search results does not appear for dash
### Describe the Bug

If a user writes a region which has a dash in between the region name then nothing returns, it returns only the word before the dash, without including dash and after dash word. For ex, Rems-Murr-Kreis, Landkreis Alzey-Worms

### Steps to Reproduce

1. Go to region search
2. Search for Rems-Murr-Kreis 
3. See no search result
4. Try searching just for Rems and you will see the result returning Rems-Murr-Kreis

### Expected Behavior

It should return all the occurrences including the regions which have dashes in-between the region name

### Environment

- Device/Browser: N/A
- OS version: N/A
- App version: N/A

### Related Issues

N/A

### Additional Information

N/A

bahaaTuffaha:
### Short Description

If a user writes a region which has a dash in between the region name then nothing returns, it returns only the word before the dash, without including dash and after dash word. For ex, Rems-Murr-Kreis, Landkreis Alzey-Worms.
It should return all the occurrences including the regions which have dashes in-between the region name

### Proposed Changes

- Split `normalizedFilter` into `normalizedSearchWords` on `MATCH_WHITESPACE_AND_DASHES`, so a search containing dashes is no longer compared as a single string against single words.
- Updated `matchesWordStart` to split the query on whitespace/dashes as well, then use `every`/`some` so each search word must be a prefix of some text word. This allows matches for names like `Rems-Murr-Kreis` and `Landkreis Alzey-Worms` while still rejecting mid-word hits.
- For the Highlighter's (web and native) `search` prop I changed the typing to `string | string[]` letting each caller decide between highlighting a string or highlighting an array.
- Changed the native `Highlighter` to split `search` with `MATCH_WHITESPACE_AND_DASHES` before passing `searchWords` to `findAllMatches`
- Switched `searchWords` in native and web `Highlighter` from `[search]` to `search.split(MATCH_WHITESPACE_AND_DASHES)` so both platforms split the query the same way.
- Replaced the `test` script's `NODE_OPTIONS=--experimental-vm-modules jest` with `yarn run -T jest`, matching `web` and `native`.

### Side Effects

- This could affect the search and Highlighting at the searchPage.

### Checklist

### Testing

- Launch integreat and test the regions page.
- Type `rems-murr-kreis` for example also `murr` or `murr-kreis`.
- Test web and native.
- Highlighting should work as expected.

### Resolved Issues

Fixes: #4310

---

### DIFF
--- a/shared/src/utils/__tests__/search.spec.ts
+++ b/shared/src/utils/__tests__/search.spec.ts
@@ -126,6 +126,19 @@ describe('search', () => {
       ]
       expect(filterSortRegions(regions, 'ba')).toEqual([])
     })
+
+    it('should match regions with dashes in their name', () => {
+      const regions = [
+        region({ sortingName: 'Rems-Murr-Kreis', prefix: '' }),
+        region({ sortingName: 'Alzey-Worms', prefix: 'Landkreis' }),
+        region({ sortingName: 'Breisgau-Hochschwarzwald', prefix: 'Landkreis' }),
+      ]
+      expect(filterSortRegions(regions, 'Rems-Murr-Kreis')).toEqual([regions[0]])
+      expect(filterSortRegions(regions, 'rems-murr')).toEqual([regions[0]])
+      expect(filterSortRegions(regions, 'Alzey-Worms')).toEqual([regions[1]])
+      expect(filterSortRegions(regions, 'alzey')).toEqual([regions[1]])
+      expect(filterSortRegions(regions, 'Landkreis Breisgau-Hochschwarzwald')).toEqual([regions[2]])
+    })
   })
 })
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1477,4 +1477,3 @@ workflows:
             equal:
                 - << pipeline.parameters.workflow_type >>
                 - web_promotion
-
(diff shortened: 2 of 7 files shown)


shared/src/utils/__tests__/search.spec.ts
```diff
@@ -126,6 +126,19 @@ describe('search', () => {
...
+        region({ sortingName: 'Rems-Murr-Kreis', prefix: '' }),
+        region({ sortingName: 'Alzey-Worms', prefix: 'Landkreis' }),
+        region({ sortingName: 'Breisgau-Hochschwarzwald', prefix: 'Landkreis' }),
+      ]
+      expect(filterSortRegions(regions, 'Rems-Murr-Kreis')).toEqual([regions[0]])
+      expect(filterSortRegions(regions, 'rems-murr')).toEqual([regions[0]])
+      expect(filterSortRegions(regions, 'Alzey-Worms')).toEqual([regions[1]])
+      expect(filterSortRegions(regions, 'alzey')).toEqual([regions[1]])
+      expect(filterSortRegions(regions, 'Landkreis Breisgau-Hochschwarzwald')).toEqual([regions[2]])
+    })
```
lunars97, line 141: 😍

lunars97 (approved):
LGTM! Thank you for a quick fix! 💚

lunars97:
Probably we need to merge it asap to fix the circle ci config update or create a separate PR to merge it...

kalairubachandrakumaran-oss:
tested ✅ 
image

PR #4320, opened 2026-08-12
4308: Fix selected filter being reset

### LINKED ISSUE #4308: Selected event filters are reset when returning to the Events page
### Describe the Problem
When a user applies filters on the Events page, opens an event from the filtered results, and then returns to the Events tab/page, all previously selected filters are reset.
-> As a result, the user has to reapply the filters, which interrupts the workflow and creates an inconsistent user experience.

- Previously selected filters should remain applied when returning to the Events page.
- The filtered list should be restored so the user can continue browsing without having to reapply the filters.

N/A

### Related Issues

N/A

### Additional Information
Steps to Reproduce:
1. Open the Events page.
2. Apply one or more filters.
3. Open an event from the filtered results.
4. Return to the Events tab/page.
5. Observe the filter state

https://github.com/user-attachments/assets/a87a5cd7-aa8a-422c-b610-65f8db81413b

N/A

lunars97:
### Short Description

This PR fixes filtered events being reset after going into event details and going back. 

### Proposed Changes

- Persist event date filters on web and native
- Persist news source filter on web and native
- Always persist ALL query params on language change

Technical:
- Create separate `useDateFilter` hook for `web` and `native` with route params on native and query params on web
- Refactor `useQueryParamVisibility` to `useQueryParam` for generic usages

### Side Effects

N/A

### Checklist

### Testing

1. Go to events (web and native)
2. Adjust filter
3. Go to event detail and back to events list
4. See the filtered dates being there (see the url)
5. Switch language and see the filters still being set
6. Open the chat
7. Switch the language
8. Close the chat
9. See the filters still being set

Check the same things for news source fitler (web and native).
Also check that e.g. feedback, chat and other functions with query parameters still work as expected.

### Resolved Issues

Fixes: #4308 

---


web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,66 @@
+import { DateTime } from 'luxon'
+import { useCallback, useMemo } from 'react'
+import { useSearchParams } from 'react-router'
+
+import { filterEvents } from 'shared'
+import { EventModel } from 'shared/api'
+
+const parseDate = (value: string | null) => (value ? DateTime.fromISO(value) : null)
```
bahaaTuffaha: ❓  Do we need to check for the invalid dates?

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,66 @@
...
+
+type UseSharedDateFilterReturn = {
```
bahaaTuffaha: 🔧  It's no longer shared ...

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,66 @@
...
+  filteredEvents: EventModel[]
+  startDateError: string | null
+  resetDates: () => void
+}
+
+const useDateFilter = (events: EventModel[]): UseSharedDateFilterReturn => {
+  const [searchParams, setSearchParams] = useSearchParams()
+
+  const startDate = parseDate(searchParams.get('startDate'))
+  const endDate = parseDate(searchParams.get('endDate'))
```
bahaaTuffaha: I think we need to put the keys in one place like at `query.ts` instead of hard coding it here.
`const START_DATE_QUERY_KEY = 'startDate'`
`const END_DATE_QUERY_KEY = 'endDate'`

bahaaTuffaha (commented):
- Changing the language removes the query 🤔 this could be pre-existing problem.
- Does it need a test?

web/src/hooks/useDateFilter.ts
lunars97: yep, I forgot to rename :grimacing:

web/src/hooks/useDateFilter.ts
lunars97: i will have a look

web/src/hooks/useDateFilter.ts
lunars97: I left constant keys in this file, I don't think it needs to be shared because it is used just here...

web/src/hooks/useDateFilter.ts
lunars97: there is already a check for it in L63

lunars97:
> * Changing the language removes the query 🤔 this could be pre-existing problem.
> * Does it need a test?

Thank you for point it out :+1: 

- I fixed the changing a language issue
- Also added tests for hooks

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,69 @@
...
+import { useCallback, useMemo } from 'react'
+import { useSearchParams } from 'react-router'
+
+import { filterEvents } from 'shared'
+import { EventModel } from 'shared/api'
+
+const START_DATE_QUERY_KEY = 'startDate'
+const END_DATE_QUERY_KEY = 'endDate'
+
+const parseDate = (value: string | null) => (value ? DateTime.fromISO(value) : null)
```
steffenkleinle: ❌ Please move all logic related to parsing and query params to `query.ts` in `shared`. This will allow us to consistently use the query params accross web and native and among different features in the future while keeping this file lean and clean.

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const startDate = parseDate(searchParams.get(START_DATE_QUERY_KEY))
+  const endDate = parseDate(searchParams.get(END_DATE_QUERY_KEY))
+
+  const setDate = useCallback(
+    (key: string, date: DateTime | null) => {
+      setSearchParams(
+        prevParams => {
+          const params = new URLSearchParams(prevParams)
+          const isoDate = date?.isValid ? date.toISODate() : null
```
steffenkleinle: 🤔 I think this check is redundant, our dates are always valid since we set `throwOnInvalid` to true.
```suggestion
```

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,69 @@
...
+          return params
+        },
+        { replace: true },
+      )
+    },
+    [setSearchParams],
+  )
+
+  const setStartDate = (date: DateTime | null) => setDate('startDate', date)
+  const setEndDate = (date: DateTime | null) => setDate('endDate', date)
```
steffenkleinle: 🔧 Please use the key constants instead of hardcoding them here

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const useDateFilter = (events: EventModel[]): UseDateFilterReturn => {
+  const [searchParams, setSearchParams] = useSearchParams()
+
+  const startDate = parseDate(searchParams.get(START_DATE_QUERY_KEY))
+  const endDate = parseDate(searchParams.get(END_DATE_QUERY_KEY))
+
+  const setDate = useCallback(
+    (key: string, date: DateTime | null) => {
```
steffenkleinle: 🔧 `key` should possibly only be a valid query param key, not a generic string. Compare to `query.ts`.

web/src/hooks/useDateFilter.ts
```diff
@@ -0,0 +1,69 @@
...
+    setSearchParams(
+      prevParams => {
+        const params = new URLSearchParams(prevParams)
+        params.delete(START_DATE_QUERY_KEY)
+        params.delete(END_DATE_QUERY_KEY)
+        return params
+      },
+      { replace: true },
+    )
+  }
```
steffenkleinle, line 35: 🙃 No braces necessary here
```suggestion
  const resetDates = () =>
    setSearchParams(
      prevParams => {
        const params = new URLSearchParams(prevParams)
        params.delete(START_DATE_QUERY_KEY)
        params.delete(END_DATE_QUERY_KEY)
        return params
      },
      { replace: true },
    )
```

steffenkleinle (commented):
Nice work, tested on firefox and works as expected.

web/src/hooks/useDateFilter.ts
lunars97: 🙈

web/src/hooks/useDateFilter.ts
lunars97: I think a dedicated DateQueryKey would be probably the cleanest...

steffenkleinle (approved):
Not tested again 🚀

web/src/routes/EventsPage.tsx
```diff
@@ -67,7 +68,7 @@ const EventsPage = ({ region, pathname, languageCode, regionCode }: RegionRouteP
         languageCode: code,
       })
     return {
-      path: isCurrentLanguage ? pathname : path,
+      path: isCurrentLanguage ? pathname : `${path}${search}`,
```
bahaaTuffaha: I found small bug (I think its harmless):
- Go to http://localhost:9001/muenchen/de/events
- Open chat then change language within chat popup to any other language.
- Notice twice chat=true?chat=true

bahaaTuffaha (approved):
Thank you for your efforts on this...
found some small issues:
- Changing the language on native still resets the dates.

shared/src/routes/query.ts
```diff
@@ -62,35 +62,46 @@ export const queryStringFromRouteInformation = (
 
 export type QueryParams = {
   chat: true | undefined
-  feedback: Rating | undefined
+  feedback: Rating | true | undefined
```
bahaaTuffaha: why there is a need for `true` here?

web/src/components/FeedbackContainer.tsx
```diff
@@ -20,27 +20,28 @@ type FeedbackContainerProps = {
...
+  const rating = typeof feedbackQueryParam === 'string' ? feedbackQueryParam : null
 
   const [comment, setComment] = useState<string>('')
   const [contactMail, setContactMail] = useState<string>('')
   const [sendingStatus, setSendingStatus] = useState<SendingStatusType>('idle')
   const [snackbarOpen, setSnackbarOpen] = useState(false)
   const [searchTerm, setSearchTerm] = useState<string | undefined>(query)
 
-  const setRating = (newRating: Rating | null) => open(newRating ?? undefined)
+  const setRating = (newRating: Rating | null) => setFeedbackQueryParam(newRating ?? undefined)
```
bahaaTuffaha: Deselecting one of the feedback buttons will close the modal 🤔

bahaaTuffaha (approved):
Nice work
Tested again on web and android ✅ 

There is one issue found at the issue modal 👇

web/src/components/FeedbackContainer.tsx
steffenkleinle: Good catch! ❤️

kalairubachandrakumaran-oss:
tested ✅ 
https://github.com/user-attachments/assets/8e1c070c-db73-4551-b4ef-a085feafc366

PR #4330, opened 2026-08-20
Refactor feedback rating to remove inconsistent handling

steffenkleinle:
### Short Description

We currently use three different ways to represent the users positive/negative feedback:
- `postive` and `negative` in `RATING_POSITIVE` and `RATING_NEGATIVE`
- `true` and `false` in `isPositiveRating`
- `up` and `down` in `POSITIVE_RATING` and `NEGATIVE_RATING` (only used in `createFeedbackEndpoint`/API)

While we need `up`/`down` for the API, `isPostiveRating` is not necessary, implicit and error prone.
Therefore remove all usages of a boolean rating.

### Proposed Changes

- Remove boolean `isPositiveRating`
- Only use `up` and `down` in the endpoint and make that clear
- Streamline the naming

### Side Effects

N/A

### Checklist

### Testing

Check that sending feedback still works as expected.

### Resolved Issues

Fixes: N/A

---

### DIFF
--- a/native/src/components/Feedback.tsx
+++ b/native/src/components/Feedback.tsx
@@ -4,7 +4,7 @@ import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view
 import { Button } from 'react-native-paper'
 import styled from 'styled-components/native'
 
-import { DEFAULT_ROWS_NUMBER, SendingStatusType } from 'shared'
+import { DEFAULT_ROWS_NUMBER, Rating, SendingStatusType } from 'shared'
 
 import buildConfig from '../constants/buildConfig'
 import useNavigate from '../hooks/useNavigate'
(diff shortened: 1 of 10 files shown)


native/src/components/Feedback.tsx
```diff
@@ -52,7 +52,7 @@ const Feedback = ({
 
   const isSearchFeedback = searchTerm !== undefined
   const [privacyPolicyAccepted, setPrivacyPolicyAccepted] = useState(false)
-  const feedbackFilled = feedbackRating === null && comment.trim().length === 0 && !searchTerm
+  const feedbackFilled = rating === null && comment.trim().length === 0 && !searchTerm
```
bahaaTuffaha: I'm confused with the feedbackFilled logic. the comment is optional so why are we checking the comment?

bahaaTuffaha (approved):

native/src/components/Feedback.tsx
steffenkleinle: The naming is the wrong way round, it should be `feedbackMissing` or something. Basically we are checking that the user can't send empty feedback, so either a rating or a comment or a search term needs to be set.

PR #4331, opened 2026-08-21
4321: Use consistent event dates

### LINKED ISSUE #4321: Make event dates clear and understandable
### User Story

As a user of Integreat, I want to see when a recurring event next takes place so that I know, whethrer I will have time to attend without having to calculate

### Problem Description

Currently, recurring events are displayed in two ways
- Recurring with end date as xx.xx.xxxx - xx.xx.xxxx making them seem like a very long event taking place every day
- recurring without end date as "ab xx.xx.xxxx" where the date shown is always the next occurence. Which makes it seem like the event is only just starting its recurrence, which is not true. 

This should hopefully make the chronology apparent again as well, paying towards fixing this comment from a municipality: 
> seit kurzem sind Veranstaltungen im Kalender anders sortiert als bisher, scheinbar nicht mehr streng chronologisch
Ich persönlich finde die neue Sortierung verwirrend und nicht intuitiv.

### Acceptance Criteria

- [ ] Recurring events with end date are clearly marked as having an end date 
- [x] recurring events in general show the next and subsequent occurences
- [ ] If using ab for future recurring events, the word should change to seit, if the recurrence has already started
- [ ] The day of the recurrence, or in general the recurrence rule should be clearly visible
- [x] The design is followed

### Related Issues

N/A

### Technical Notes

N/A

### Additional Context

See Mattermost discussion here: https://chat.tuerantuer.org/digitalfabrik/pl/6eb95hf5qtgfzbya1uduqpqdcw
See screenshot of current state here: 

Image

steffenkleinle:
### Short Description

Part 1 of #4321: Use consistent event dates in the event list and detail. 

The recurrence start, end and frequency will be added in a separate PR.

### Proposed Changes

- Create `EventDates` components for both web and native used for the list and detail
- Refactor formatting functions in `DateModel` to only `formatDateInterval` and `formatTimeInterval` that is consistently used for all date types (recurring, single day, multi day, all day, weekly, ...)

### Side Effects

For now, the recurrence start/end are not visible at all anymore, only the upcoming event dates.

### Checklist

### Testing

Check varying event dates on web and native and in the list and detail.

### Resolved Issues

Fixes: N/A

---


native/src/components/EventDates.tsx
```diff
@@ -0,0 +1,125 @@
...
+  size = 'medium',
+}: EventDatesProps): ReactElement => {
+  const { t } = useTranslation('events', { lng: language })
+  const theme = useTheme()
+  const [expanded, setExpanded] = useState(false)
+
+  const date = firstDateInRange(event, filterStartDate, filterEndDate)
+
+  const maxFurtherDates = size === 'medium' ? MAX_FURTHER_DATES : MAX_FURTHER_DATES_MOBILE
+  const furtherDates = event.isRecurring ? event.date.furtherDates(maxFurtherDates) : []
```
steffenkleinle: @osmers @ulyanahoy-cloud which further dates should be displayed if a custom date range filter is set?
E.g. for a weekly event that has
- upcoming recurrences 0, 1, 2, ... 10
- and the date range filter only includes recurrence 5 and 6
-> a displayed date `5`

Should we show as further dates recurrences
a) 0, 1, 2, 3, 4 (i.e. the next/upcoming dates without the displayed date `5` -> the second recurrence in the time range (`6`) would not be included in this example) or
b) 6, 7, 8, 9, 10 (i.e. dates following the displayed date `5`)

shared/src/api/models/DateModel.ts
```diff
@@ -148,54 +118,40 @@ class DateModel {
   }
 
   hasVaryingTimes(count = MAX_FURTHER_DATES): boolean {
-    const times = this.recurrences(count + 1).map(
-      recurrence => `${recurrence.startDate.toFormat('HH:mm')}-${recurrence.endDate?.toFormat('HH:mm')}`,
+    const times = this.recurrences(count + 1).map(recurrence =>
+      // If the times vary with locale 'de' and a hardcoded translation, they will also vary with the real locale and translation
+      recurrence.formatTimeInterval('de', { allDayLabel: 'allDay' }),
```
steffenkleinle: The previous implementation did not work/was not defined for events with `allDay` set to true

native/src/components/EventDates.tsx
osmers: Or something like - the event takes place every {recurrence rule}, also on {dates in time range selected}?

Not sure if that is possible with the information you are getting from the CMS.

If not, a) is weird to me bcs it shows dates as further occurences before the date it shows above, so I think I would go for b) but I find either option not really that understandable. Maybe option c) would be, only dates that fall within the selectd time range.

shared/src/api/models/DateModel.ts
```diff
@@ -148,54 +128,40 @@ class DateModel {
...
     )
     return times.some(time => time !== times[0])
   }
 
-  formatEventDate(locale: string, t: TranslateFunction): FormattedEventDate {
-    const time = formatTime(locale, this, t)
-    const mondayTranslation = getWeekdayFromIndex(0, locale)
-    const fridayTranslation = getWeekdayFromIndex(4, locale)
-    const weekday = this.onlyWeekdays ? `${mondayTranslation} - ${fridayTranslation}` : undefined
```
bahaaTuffaha, line 161: ❓  so we don't need `onlyWeekdays` anymore? 🤔

shared/src/utils/date.ts
```diff
@@ -1,39 +1,16 @@
...
+  time.toLocaleString({ hour: 'numeric', minute: '2-digit' }, { locale })
 
-type DateType = {
-  startDate: DateTime
-  endDate: DateTime | null
-  allDay: boolean
+type FormatDateOptions = {
+  locale: string
+  showYear?: boolean
+  short?: boolean
```
bahaaTuffaha: ❓  is it used anymore? (the short format)

web/src/components/EventDates.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <>
+            <span aria-hidden>{HORIZONTAL_TEXT_DIVIDER}</span>
+            {timeInterval}
+          </>
+        </TextRow>
+      </Stack>
+      {furtherDates.length > 0 && (
+        <AccordionWrapper dir={contentDirection} onClick={stopLinkNavigation}>
+          <Accordion
+            id='further-dates'
```
bahaaTuffaha: I just noticed this we currently use the same id for each of the accordions.

bahaaTuffaha (commented):
Good refactoring but replacing `DatesPageDetails` I find it a bit odd because it should show all dates and the more you click on "show more" the more it will give you dates which is intuitive and eventFurtherDates should act as a small preview. now they are the identical (details and the eventsfurtherDates)  plus there is no year in the format and doesn't show time for each future event at desktop screens (we have space for time on big screens).

- The event details also shows 3 future events on mobile screens but native shows 5 at eventDetails...

shared/src/api/models/DateModel.ts
steffenkleinle: Will be re-added with the recurrence rule in the follow-up PR.

web/src/components/EventDates.tsx
steffenkleinle: Is that a problem? They are never rendered at the same time?

bahaaTuffaha: in the EventListItem.tsx‎ in the list? 🤔

steffenkleinle: Ooooh 🤦

steffenkleinle:
> Good refactoring but replacing `DatesPageDetails` I find it a bit odd because it should show all dates and the more you click on "show more" the more it will give you dates which is intuitive and eventFurtherDates should act as a small preview. now they are the identical (details and the eventsfurtherDates)

I readded the show more button while still keeping the same component. In the event details, it is now expanded as default.

> plus there is no year in the format and doesn't show time for each future event at desktop screens (we have space for time on big screens).

Personally, I think it looks a lot clearer and easier to read this way, so I would continue to only show
- the year, if different than the current calendar year
- the time, if different than the time of the main date

>     * The event details also shows 3 future events on mobile screens but native shows 5 at eventDetails...

Fixed

native/src/components/EventDates.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <DateRow language={language}>
+        <Icon source='calendar-text-outline' size={iconSize} />
+        <InlineWrap language={language} style={{ flex: 1 }}>
+          <Text variant={textVariant}>{date.formatDateInterval(language)}</Text>
+          <Text variant={textVariant} aria-hidden>
+            {HORIZONTAL_TEXT_DIVIDER}
+          </Text>
+          <Text variant={textVariant}>{date.formatTimeInterval(language, { allDayLabel })}</Text>
+        </InlineWrap>
+      </DateRow>
```
bahaaTuffaha, line 85: App language is english and the system is Arabic.
Both further dates and the date above it should align (text-direction) with the app language.

Image

web/src/components/base/Accordion.tsx
```diff
@@ -46,7 +46,10 @@ const Accordion = ({
       disableGutters
       elevation={0}
       expanded={expanded}
-      onChange={(_, isExpanded) => setExpanded(isExpanded)}
+      onChange={(event, isExpanded) => {
+        event.preventDefault()
+        setExpanded(isExpanded)
+      }}
```
bahaaTuffaha, line 52: ❓  is it a good idea to make it globally for all Accordions?

shared/src/api/models/DateModel.ts
```diff
@@ -1,21 +1,11 @@
-import { DateTime, DateTimeFormatOptions, Duration } from 'luxon'
+import { DateTime, Duration } from 'luxon'
 import { RRule as RRuleType, rrulestr } from 'rrule'
 
 import { MAX_FURTHER_DATES } from '../../constants/index.ts'
```
bahaaTuffaha: unused import

shared/src/api/models/DateModel.ts
```diff
@@ -85,14 +75,18 @@ class DateModel {
...
+    // The rrule package treats all times as UTC
+    // Subtracting the duration also includes events that are happening right now.
+    const minDate = rangeStart
       .minus(duration ?? 0)
-      .minus({ minutes: startDate.offset })
+      .setZone('utc', { keepLocalTime: true })
+      .toJSDate()
+    const maxDate = (filterEndDate ?? now.plus({ years: MAX_RECURRENCE_YEARS }))
+      .setZone('utc', { keepLocalTime: true })
       .toJSDate()
```
bahaaTuffaha, line 88: ❓  why it needs a setZone now?

bahaaTuffaha (approved):
Great work ✨ 

Tested on web and Android ✅

web/src/components/base/Accordion.tsx
steffenkleinle: I was wondering the same, but at least at the moment this doesn't break anything. So I would keep it like this for now at least.

shared/src/api/models/DateModel.ts
steffenkleinle: The function did not work correctly before when filtering dates. Try to set the filter to 28.08 and see that some events with 27.08 are still displayed. These changes fix that problem. Basically the offset was the wrong way round and setting the zone to utc while keeping the local time is basically another way to apply an offset.

native/src/components/EventDates.tsx
steffenkleinle: We will display the recurrence rule in any case with a follow-up PR. I went with a) now actually and I think that makes the most sense and is most likely what users care most of the time about. We could also change `Weitere Termine` to `Andere Termine` or something to make that more clear.
c) would mean that we don't display any dates if the filter is set accordingly.

native/src/components/EventDates.tsx
steffenkleinle: Turns out that `flex-flow` does not work correctly on native. Should all be correctly aligned now. Has it always been that complicated to do on native? It seems like it is basically necessary to adjust every single component...

Issue #4038, opened 2026-04-15
Clean up translations


steffenkleinle:
### Describe the Problem

Our translations experienced quite a lot of (wild) growth in the path, which led to many inconsistencies, duplication and confusion on what the different namespaces are for.
On top, many translations are probably not used anymore.

### Describe Your Preferred Solution

- Try to clean up inconsistencies in translation keys and values
- Merge namespaces `layout`, `common` and `settings`
- Make sure namespaces and constants follow our naming

### Describe Possible Alternatives

N/A

### Related Issues

Blocks: #4032

### Additional Information

Glossary of currently used terms: https://nextcloud.tuerantuer.org/index.php/apps/files/files/6711789?dir=/Projects/Integreat&editing=false&openfile=true

PR #4337, opened 2026-08-26
4329: Add typechecking for translation types

### LINKED ISSUE #4329: Improve DX for translations
### Describe the Problem

Our current translations are quite difficult to handle in terms of DX. On top, there is no type safety, so its quite error prone. Autocompletion is also missing.

- Setup i18n correctly to allow for typechecking
- Use selector API
- Use per-language json files

N/A

### Related Issues

N/A

### Additional Information

https://www.i18next.com/overview/typescript

steffenkleinle:
### Short Description

Add typechecking for translations by switching to selector-based translate functions.
Example:
``` tsx
// previous
t('events')
t('common:back')

// new
const { t } = useTranslation()
t($ => $.events.events)
t($ => $.common.back)
```

This has the following advantages for devs:
- TS errors if using a wrong key -> translations are less error-prone
- Autocompletion for keys
- Preview of the translation value in the IDE (e.g. `$.events.currentlyNoEvent` -> `No events are currently available.`.

### Proposed Changes

- Set `enableSelector: 'strict'` to enable the selector-based translate function and therefore typechecking
- Set `defaultNS` to all namespaces and therefore allow using `useTranslation()` without passing namespaces
- Auto-generate `translations/src/resources.gen.ts` to allow typechecking params and IDE value previews
- Pass translated strings to `showSnackbar` and `useOpenExternalUrl` hook among others
- Fix all usages and tests

### Side Effects

I decided to directly go for `enableSelector: 'strict'`, which forces us to always pass the whole key, e.g. `$.events.currentlyNoEvents` instead of `useTranslation('events') and `$.currentlyNoEvents`. This leads to more consistent and explicit usage of the translations. However, it would also be possible to use typechecking without this. See https://www.i18next.com/overview/typescript#custom-type-options -> `enableSelector` for more info.

Tests are fixed in #4340 for (slightly better) readability of this PR.

### Checklist

### Testing

Try to test the whole app very thoroughly and try to have a look for broken translations.
IMPORTANT: Checking either German or English thoroughly is enough, previously missing translations in other languages will only be added in #4038. Basically, it is enough that actual translations are rendered and not just the keys, e.g. something like `events.currentlyNoEvents`.

### Resolved Issues

Fixes: #4329

---


native/src/components/ChatHighlightPopup.tsx
```diff
@@ -72,18 +72,19 @@ const ChatHighlightPopup = ({ chatName }: ChatHighlightPopupProps): ReactElement
...
           onPress={() => updateSettings({ chatHighlightPopupDismissed: true })}
-          accessibilityLabel={t('common:close')}
+          accessibilityLabel={t($ => $.common.close)}
         />
       </Header>
       <Text variant='body2'>
         <Trans
-          i18nKey='chat:welcomeText'
+          ns='chat'
+          i18nKey={$ => $.chat.welcomeText}
```
bahaaTuffaha, line 87: The PDF button and the highlightPopup are having issues:
Image

bahaaTuffaha (changes requested):
- `contributing.md` needs updating (the note at Pull Requests section).
- For iOS the file at `native/ios/Integreat.xcodeproj/xcshareddata/xcschemes/` they still use the old path: 
```
--translations &quot;../translations/src/translations.json`
```
- Do we need to add `enableSelector` strict at `i18nextInstance.init()` for web and native (I18nProvider.tsx)
- language not found for web and native:

Image

steffenkleinle:
> * `contributing.md` needs updating (the note at Pull Requests section).
> 
>     * For iOS the file at `native/ios/Integreat.xcodeproj/xcshareddata/xcschemes/` they still use the old path:

Done, good catches!

>     * Do we need to add `enableSelector` strict at `i18nextInstance.init()` for web and native (I18nProvider.tsx)

Good catch again, that was missing and was the cause for the other two problems. Seems like I accidentally reverted that at some point.

bahaaTuffaha (approved):
Thanks! LGTM ✅ 
Tested web and native.

native/src/components/EventDates.tsx
```diff
@@ -54,11 +54,12 @@ const EventDates = ({
   compact = false,
 }: EventDatesProps): ReactElement => {
   const [expansionCount, setExpansionCount] = useState(compact ? 0 : 1)
-  const { t } = useTranslation('events', { lng: language })
+  const { t } = useTranslation()
```
bahaaTuffaha: does it now follows the device language instead of content language?

web/src/components/PlaceSharedChildren.tsx
bahaaTuffaha: I think it still uses the old format here.

PR #4341, opened 2026-08-28
4321: Add weekdays for multi-day and end date for recurring events

### LINKED ISSUE #4321: Make event dates clear and understandable

steffenkleinle:
### Proposed Changes

- Show `Monday - Friday` for multi-day events with the corresponding option enabled
- Show end date for recurring events
- Show a short weekday for all dates (e.g. `Tue` or `Fri.`)
- Allow usage of translation function `t` in shared with type support

### Side Effects

N/A

### Checklist

### Testing

Check both the event list and event details on native and web. Check multi-day events with `Monday to Friday only` enabled. Check recurring events with an end date.

### Resolved Issues

Fixes: #4321

---


lunars97 (approved):
Thank you for your effort! LGTM 💛 Tested in Safari and iOS

shared/src/api/models/DateModel.ts
```diff
@@ -151,6 +152,27 @@ class DateModel {
...
+    return null
+  }
+
+  formatRecurrenceEnd(locale: string, { t }: { t: TFunction }): string | null {
+    const until = this.recurrenceRule?.options.until
+    if (!until || !this.hasMoreRecurrencesThan(1)) {
+      return null
+    }
+
+    const lastRecurrenceJsDate = this.recurrenceRule.before(until, true)
```
bahaaTuffaha, line 170: Claude found a case where an event could show up in the next day if the end was after midnight. Check this test out `DateModelMidnightBug.spec.ts`:

```js
import { TFunction } from 'i18next'
import { DateTime, Settings } from 'luxon'
import { rrulestr } from 'rrule'

import DateModel from '../DateModel.ts'

// Pin the zone to a positive offset (Europe/Berlin in July = CEST, +2) so the machine running the
// test does not affect the result. The bug only surfaces at a non-zero offset.
Settings.defaultZone = 'Europe/Berlin'
jest.useFakeTimers({ now: new Date('2025-07-01T12:00:00.000+02:00') })

describe('formatRecurrenceEnd', () => {
  const baseDates = {
    startDate: DateTime.fromISO('2025-07-04T23:30:00+02:00'),
    endDate: DateTime.fromISO('2025-07-04T23:30:00+02:00'),
    allDay: false,
    onlyWeekdays: false,
  }

  it('should report the real last occurrence date for a near-midnight recurrence', () => {
    // Weekly on Fridays (4/11/18/25 July) at 23:30; the real last occurrence is Friday, 25 July 2025.
    const date = new DateModel({
      ...baseDates,
      recurrenceRule: rrulestr('DTSTART:20250704T233000\nRRULE:FREQ=WEEKLY;BYDAY=FR;UNTIL=20250725T233000'),
    })
    const interpolatingT = ((_: unknown, options: { date: string }) => `bis ${options.date}`) as unknown as TFunction

    // The last occurrence is Friday, 25 July 2025. Currently fails because the +2h offset rolls
    // 23:30 forward to 01:30 the next day and reports 'bis Sa., 26. Juli' instead.
    expect(date.formatRecurrenceEnd('de', { t: interpolatingT })).toBe('bis Fr., 25. Juli')
  })
})

```

native/src/components/EventDates.tsx
```diff
@@ -68,9 +62,14 @@ const EventDates = ({
   const hasMoreRecurrences = date.hasMoreRecurrencesThan(maxVisibleRecurrences + 1)
   const expanded = expansionCount > 0
 
+  const extraDateText = date.formatRecurrenceEnd(language, { t }) ?? date.formatWeekdays(language)
```
bahaaTuffaha, line 65: ❓  I think it need to depend on the app language not the system to match the rest if the content 🤔

native/src/components/EventDates.tsx
```diff
@@ -68,9 +62,14 @@ const EventDates = ({
...
   const textVariant = compact ? 'body3' : 'body2'
   const iconSize = compact ? SMALL_ICON_SIZE : DEFAULT_ICON_SIZE
 
+  const Styles = StyleSheet.create({
+    indentation: { [isContentDirectionReversalRequired(language) ? 'paddingRight' : 'paddingLeft']: iconSize + 4 },
+  })
```
bahaaTuffaha, line 71: 🔧  I think it's better defining it out of the component

web/src/components/EventDates.tsx
```diff
@@ -35,7 +37,7 @@ const AccordionWrapper = styled('div')(({ theme }) => ({
     color: theme.palette.primary.main,
   },
   [`& .${accordionDetailsClasses.root}`]: {
-    padding: theme.spacing(0, 0, 0, 4),
+    padding: theme.spacing(0, 0, 0, compact ? SMALL_ICON_INDENTATION : 4),
```
bahaaTuffaha, line 40: ❓  why it has different indentation when its compact?

bahaaTuffaha (approved):
Great work!
tested on web and Android ✅

web/src/components/EventDates.tsx
steffenkleinle: Because the icon is smaller then, so in order to align it with the text in the first row after the icon the indentation needs to be accordingly.

native/src/components/EventDates.tsx
steffenkleinle: Since we need the language it does not really make sense.

Docmale:
I encountered the following problem during testing: for a scheduled period — say, from Saturday, Sept. 26 to Saturday, Nov. 28 — I cannot tell whether the appointment takes place monthly, weekly, daily, or at some other frequency. I would have liked to see a frequency specification, such as daily, weekly, etc.

steffenkleinle:
> I encountered the following problem during testing: for a scheduled period — say, from Saturday, Sept. 26 to Saturday, Nov. 28 — I cannot tell whether the appointment takes place monthly, weekly, daily, or at some other frequency. I would have liked to see a frequency specification, such as daily, weekly, etc.

It is very difficult to render user-facing frequencies in natural language. Additionally, it is sometimes quite difficult and error-prone to calculate all future occurrences manually. Since it is possible to load every single future recurrence, it was a deliberate decision not to show a recurrence rule to users.

PR #4371, opened 2026-09-09
4347: Optimize events list loading time

### LINKED ISSUE #4347: Event details lags for couple seconds when going back to events list
### Describe the Bug

Event detail page lagging for couple seconds when going back to events list. Occurs depending on the  amount of events.

### Steps to Reproduce

1. Go to 'Events' in Stadt Augsburg / Landeshauptstadt München
2. Click on any event
3. Navigate back from event detail to events list
4. See lagging for couple seconds before it navigates back

### Expected Behavior
The behaviour should be smooth

### Environment

- Device/Browser: N/A
- OS version: N/A
- App version: N/A

### Related Issues

Might be useful #4317 

### Additional Information

N/A

lunars97:
### Short Description

This PR fixes events list loading performance especially after navigating back from event details to events list by using virtualization from [react-virtuoso]( https://www.npmjs.com/package/react-virtuoso) which mounts only event items in the viewport and therefore improves the performance especially if events list is large.

### Proposed Changes

- use `Virtuoso` for the filtered view
- use `GroupedVirtuoso` for grouping the events
- the `key='filteredEvents' `and `key='groupedEvents'` on the Virtuoso instances is intentional while it forces a fresh mount when switching between filter and grouped mode so state doesn't leak between them
- rearrange grouped data for `GroupedVirtuoso` as it needs flat data
- add scroll restoration for back navigation using `restoreStateFrom` prop

### Side Effects

Keyboard behaviour is different now because of `useWindowScroll` prop, it requires to use arrow-down / arrow-up to load more events by moving the scroll and use Tab to focus on the item

### Checklist

### Testing

- Go to events (Testumgebung)
- Scroll the event list items
- Click on any event
- See event details
- Go back to events list

Check also filtering and mobile view as well 

### Resolved Issues

Fixes: #4347

---

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -59,6 +59,7 @@
     "react-router": "^7.18.2",
     "react-spring-bottom-sheet": "^3.4.1",
     "react-svg": "^19.1.4",
+    "react-virtuoso": "^4.18.13",
     "shared": "workspace:^",
     "stylis": "4.2.0",
     "translations": "workspace:^"
--- a/web/src/components/EventList.tsx
+++ b/web/src/components/EventList.tsx
@@ -3,9 +3,9 @@ import Divider from '@mui/material/Divider'
 import Stack from '@mui/material/Stack'
 import { styled } from '@mui/material/styles'
 import { DateTime } from 'luxon'
-import React, { ReactElement, useMemo } from 'react'
+import React, { ReactElement, useLayoutEffect, useMemo, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
-import { GroupedVirtuoso, Virtuoso } from 'react-virtuoso'
+import { GroupedVirtuoso, GroupedVirtuosoHandle, StateSnapshot, Virtuoso, VirtuosoHandle } from 'react-virtuoso'
 
 import { groupEventsByDate, eventGroupTitle } from 'shared'
 import { EventModel } from 'shared/api'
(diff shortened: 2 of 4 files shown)


web/src/components/EventList.tsx
```diff
@@ -9,27 +14,74 @@ import EventListItem from '../components/EventListItem'
...
-type EventListGroupProps = {
-  title: string
-  events: EventModel[]
+const Container = styled(Stack)(() => ({
+  height: '90%',
+}))
+
+const EventsListWrapper = styled(Box)(() => ({
+  flex: 1,
+  minHeight: 0,
```
lunars97: to make it shrink when the filter accordion expands, otherwise it overflows the footer container

web/package.json
```diff
@@ -59,6 +59,7 @@
     "react-router": "^7.18.2",
     "react-spring-bottom-sheet": "^3.4.1",
     "react-svg": "^19.1.4",
+    "react-virtuoso": "^4.18.13",
```
bahaaTuffaha, line 62: hmm MUI has a virtual list :
https://mui.com/material-ui/react-list/#virtualized-list

so the question is does this package has an advantage?

web/src/components/EventList.tsx
```diff
@@ -51,43 +142,59 @@ const EventList = ({ events, languageCode }: EventListProps): ReactElement | nul
...
-  })
+  }
 
   return (
-    <>
+    <Container>
       {dateFilter}
-      <List items={dateGroups} noItemsMessage={t($ => $.events.currentlyNoEvents)} showDividers={false} />
-    </>
+      <GroupedListWrapper $stickyTop={stickyTop}>
```
bahaaTuffaha, line 184: at console it says:
```
Invalid attribute name: `$stickyTop`
```

web/src/components/EventList.tsx
```diff
@@ -51,43 +142,59 @@ const EventList = ({ events, languageCode }: EventListProps): ReactElement | nul
...
+        <GroupedVirtuoso
```
bahaaTuffaha, line 185: You can add `increaseViewportBy` prop with like 400px to increase the rendered items off screen so we can tab with keyboard without jumping to the footer.

web/package.json
steffenkleinle: AFAICS, the virtualized list section in the docs only recommends `react-window`, which is not included in mui itself. So another library is necessary in any case if we need virtualization.

However, at the moment, testumgebung has 600+ events, while our actual regions have a lot less (e.g. munich or augsburg only have ~50). So the question is whether we actually need virtualization in general.

Much of our performance problems with events seem to persist with this PR, and using react-virutoso introduces new problems (e.g. a11y, find in page, list semantics, another library, sticky top hack, ...).

I therefore tried to optimize deeper lying problems in #4384, perhaps you could have a look if you think react-virtuoso is still necessary?

lunars97: @bahaaTuffaha as @steffenkleinle mentioned above, MUI is just suggesting to use `react-window` with their list component which is rather primitive, but in our case, we have grouped events and `react-virtuoso` has grouped virtualization with sticky header which aligns well with SectionList, so that is why i went for it. 
@steffenkleinle thank you for creating a PR to tackle it, I will have a look. Yes, you are right. It has accessibility issues and the point about whether it is really needed if real regions' don't have usually that much events is also valid :thinking:

PR #4323, opened 2026-08-14
4010: Implement the tour feature

bahaaTuffaha:
### Short Description
[Reactour](https://docs.reactour.dev/) will allow us to highlight each section with description like holding the user's hand through the website.
the design: https://www.figma.com/design/cA4F2MwHs2LNGviOWUjfE9/Integreat-Tickets-Frontend?node-id=6677-1040&p=f&t=NhuMIUFpZRK6sINl-0

### Proposed Changes

- Installed `@reactour/tour` at web.
- The tour is structured as follows: `TourContainer` -> `TourProvider`(from the package) -> `TourDialog` (welcome/finish) and `TourPopover`-> `TourStepContent`.
- Added `HEADER_ELEMENT_ID`, `HEADER_TITLE_ELEMENT_ID`, `HEADER_ACTIONS_ELEMENT_ID`, `HEADER_MENU_ELEMENT_ID`, `NAVIGATION_TABS_ELEMENT_ID`, `TILES_ELEMENT_ID` and `CHAT_FAB_ELEMENT_ID` as selectors for the highlighted elements.
- At  `useLocalStorage.ts` I added `TOUR_DIALOG_VISIBLE_STORAGE_KEY` set to `'Tour-Dialog-Visible'`.
- Extended `StepType` to `TourStepType` with `content`, `arrowAlignment` and an `offset`.
- At `TourStepContent.tsx` : Rendered the title, description and an optional image.
- Rendered the `TourContainer` at `RegionContentLayout.tsx`
- Created custom `TourDialog` that has content welcome or finished.
- I divided the steps into `tourStepsDesktop` and `tourStepsMobile`. Both returns an array of steps that includes the `content` , `position`, `selector`, `offset` and `arrowAlignment`.
- These will get hidden dynamically: the location, navigation, and chat steps when `featureFlags.fixedRegion` is set, `getNavigationItems` is empty, or chat is off in the build config.
- Replaced the default Reactour popover with a custom `TourPopover` that renders content , dots for steps and next/previous buttons.
- Added `tourMenuPreview: '/tour-menu-preview.png'` to the web icons of `aschaffenburg`, `integreat`, `malte`, and `obdach`.
- Used `position: relative` with `insetInlineStart` and `insetBlockStart` for the offset (i didn't use margins since margins change the size of the popover).
- Used `positionBelowElement` and per-step `offset`s to keep the popover on screen and aligned with the highlighted target.
[...]

Issue #4032, opened 2026-04-15
Fix header title


steffenkleinle:
### Describe the Bug

We currently have a few broken header titles when navigating to the settings from less common routes (e.g. licenses, pdf, image, recommend to city, ...
There is no corresponding translation.

### Steps to Reproduce

1. Go to settings > licenses > settings
2. See header title

### Expected Behavior

The header title is always correctly translated.

### Related Issues

https://github.com/digitalfabrik/integreat-app/pull/4020#issuecomment-4250260411

### Additional Information

N/A

PR #4372, opened 2026-09-10
4038: Cleanup and restructure translations

### LINKED ISSUE #4038: Clean up translations

steffenkleinle:
### Short Description

Our translations were a mess, with many inconsistencies, weird namespace choices and legacy naming.
This PR restructures it to clear feature namespaces (e.g. categories, events, news, places, ...) that match our feature names and one extra `common` namespace, which is only used for things that are actual used commonly.

Tests and validation are added/fixed in #4374.

### Proposed Changes

- Restructure namespaces (e.g. remove `layout` namespace`)
- Cleanup
- Remove duplicates and redundant or overly specific translations
- Streamline translations keys and values
A little unrelated:
- Split suggest to region steps into four steps
- Use a default `nothingFound` message for lists
- Always show `Go back` as message on errors

### Side Effects

There are currently only translations in English and German. Once I get approvals here, I will forward our translations to @osmers to translate the files in all our languages. They should then be added in a follow-up PR. Perhaps we can combine translating with #4323.
Header titles on native might still be broken, I'll fix this in a follow-up PR in #4381.
Tests and validation will be added/fixed in #4374.

### Checklist

### Testing

Test the whole app thoroughly and check for missing/wrong translations. Check especially the (user-facing) things in the proposed changes section.

### Resolved Issues

Fixes: #4038

---


native/src/components/TimeStamp.tsx
bahaaTuffaha, line 1: There is an issue here.
Image

web/src/components/EventsDateFilter.tsx
bahaaTuffaha, line 1: Found one here at the date filter (also at native)
Image

web/src/routes/RegionsPage.tsx
```diff
@@ -23,8 +23,8 @@ const RegionsPage = ({ languageCode }: RegionsPageProps): ReactElement => {
   const { data: regions, isPending, error } = useQueryFromEndpoint(createRegionsEndpoint, cmsApiBaseUrl, undefined)
   const { t } = useTranslation()
 
-  const pageTitle = t($ => $.regions.pageTitle)
-  const metaDescription = t($ => $.regions.metaDescription, { appName: buildConfig().appName })
+  const pageTitle = t($ => $.regions.title)
```
bahaaTuffaha, line 26: `regions.title` has {{appName}} not sure if it's the correct translation...

native/src/components/AppointmentOnlyOverlay.tsx
bahaaTuffaha, line 1: Image

bahaaTuffaha (changes requested):
- The search is showing no results found...

Image

- The header is not showing the language correctly (switched system language to DE):

Image

lunars97 (commented):
Thank you for your PR! :)
I found the same issues as Bahaa. Regarding his point about now having `No search results` with icon in web, before it wasn't there and in native it doesn't exist. We could maybe leave it as it is and add it to native...

Image

lunars97:
I just found one issue on main: 

- The variables in OrganizationContentInfo.tsx are not being translated. See the screenshot:
Screenshot_20260912-084317

- Is it also possible to update the link in Rules.md in translations, the included link to page does not exist, I believe it should be replaced with this one: https://wiki.tuerantuer.org/service-team/translations-internal ? or we could open an issue to update it...

steffenkleinle:
>     * The header is not showing the language correctly (switched system language to DE):

@bahaaTuffaha I assume you mean that the header title is not translated correctly? This will be addressed in #4032.

@bahaaTuffaha @lunars97 I think I fixed all issues otherwise, please have another look.

bahaaTuffaha:
> @bahaaTuffaha I assume you mean that the header title is not translated correctly? This will be addressed in #4032.

Yes exactly 👍

web/src/components/BottomActionSheet.tsx
```diff
@@ -95,7 +95,7 @@ const BottomActionSheet = ({ children, sibling, title, ref }: BottomActionSheetP
         <HeaderButton
           dir={contentDirection}
           onClick={toggleFullscreen}
-          aria-label={t($ => $.common.handle)}
+          aria-label={t($ => $.common.actions.back)}
```
bahaaTuffaha: Why it has back translation for a handle ?

native/src/components/PlacesBottomSheet.tsx
```diff
@@ -170,7 +170,7 @@ const PlacesBottomSheet = ({
           data={places}
           role='list'
           {...conditionalA11yProps({ hidden: !!slug })}
-          accessibilityLabel={t($ => $.places.placesCount, { count: places.length })}
+          accessibilityLabel={t($ => $.places.filter.show, { count: places.length })}
```
bahaaTuffaha: `placesCount` was "{{count}} locations" or "{{count}} places" but without "show"

web/src/components/Sidebar.tsx
```diff
@@ -69,7 +66,7 @@ const Sidebar = ({ children, open, setOpen, footer, openButton, className }: Sid
         disableRestoreFocus>
         <Header>
           <Stack sx={{ minHeight: headerHeight, justifyContent: 'center', alignItems: 'flex-end', paddingInline: 1 }}>
-            <IconButton onClick={() => setOpen(false)} aria-label={t($ => $.layout.sideBarCloseAriaLabel)}>
+            <IconButton onClick={() => setOpen(false)} aria-label={t($ => $.common.actions.close)}>
```
bahaaTuffaha: We were using a dedicated translation for the sidebar now its a generic open is that expected?

web/src/components/PlacePanelNavigation.tsx
```diff
@@ -28,16 +28,16 @@ const PlacePanelNavigation = ({ switchPlace }: PlacePanelNavigationProps): React
         startIcon={<DirectionDependentBackIcon />}
         tabIndex={0}
         color='inherit'
-        aria-label={t($ => $.places.previousPlace)}>
-        {t($ => $.places.detailsPreviousPlace)}
+        aria-label={t($ => $.common.actions.previous)}>
```
bahaaTuffaha: was it `Vorheriger Ort` before?

bahaaTuffaha (approved):
Thank you for the huge effort.
Tested web ✅ 
Tested iOS ✅ 

I left some questions 👇 

-  At `translations/src/manage.ts` L211 still using .native I think we use ios now.

web/src/components/BottomActionSheet.tsx
lunars97: good catch, it should be `show less` or `show more` like in native...

web/src/components/Sidebar.tsx
steffenkleinle: This is not used anymore, I removed it.

web/src/components/PlacePanelNavigation.tsx
steffenkleinle: Yes, I think thats okay.

native/src/components/Feedback.tsx
```diff
@@ -76,20 +76,20 @@ const Feedback = ({
...
-            title={t($ => $.feedback.searchTermDescription)}
+            title={t($ => $.feedback.search.searchTermDescription)}
             value={searchTerm}
             onChange={setSearchTerm}
             accessibilityRole='search'
           />
         ) : (
           <>
-            <Caption title={t($ => $.feedback.headline)} />
+            <Caption title={t($ => $.feedback.title)} />
```
lunars97, line 86: Hmm now it is just `Feedback`, before it was `Wie findest du diese Seite?`

native/src/components/News.tsx
```diff
@@ -124,16 +124,16 @@ const News = ({ news, id, languageCode, region, refresh, sourceFilter, setSource
...
         <ListHeaderContainer>
-          <Caption title={t($ => $.news.news)} />
+          <Caption title={t($ => $.news.title)} />
           {showNewsSourceFilter && (
             <ToggleTextButtonGroup
               setValue={setSourceFilter}
               value={sourceFilter}
               options={NEWS_SOURCE_FILTERS}
-              getLabel={value => t($ => $.news[value])}
+              getLabel={value => t($ => $.news.sources[value])}
```
lunars97: I think we need to use short versions like it is done for web mobile view 👀 
`getLabel={value => t($ => $.news.sources[`${value}Short`])}`

lunars97 (changes requested):
Tested in real android device 17 and Chrome. Everything looks almost perfect, just found some minor misalignments 💛 

- Currently for export button in German we have `Export` instead of `Exportieren`. See the screenshot below:

Image

native/src/components/Feedback.tsx
steffenkleinle: Yes, I would say that is a change that makes sense. I am open to keep the previous `What do you think of this page?` as well though.

lunars97: I think it is also okay to keep it as it is...

lunars97 (approved):
LGTM 👍

PR #4382, opened 2026-09-14
4360: Fix list focus for screen-reader

### LINKED ISSUE #4360: Fix FlatList not focusing on tap with TalkBack in category list items
### Describe the Bug

Currently we are using nested FlatList in category list items which is a bad practice as it leads to `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead.` if `scrollEnabled` is not set to `false`. Seems like the fix with `scrollEnabled={false}` just hides the error warning and does not solves it.
`scrollEnabled={false}` should be avoided to not break the tap behaviour when TalkBack is activated (could be also the case with VoiceOver in ios)

### Steps to Reproduce

1. Activate TalkBack on android 
2. Go to any category
3. Try to tap on the category list items
4. See that it is not focusing on the tapped list item

### Environment

- Device/Browser: N/A
- OS version: N/A
- App version: N/A

### Related Issues

could be relatable #4080 

### Additional Information

N/A

lunars97:
### Short Description

This PR fixes focus/hover issue in category list items when screen-reader (Talkback) is activated by replacing FlatList with [SectionList](https://reactnative.dev/docs/sectionlist) as it seems more semantically close than FlatList (events list also uses SectionList) and there is no need to include a workaround with `scrollEnabled={false}` for the error `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead.` 

### Proposed Changes

- Replace Page.tsx component and List.tsx component with SectionList
- Make adjustments in CategoryListItem.tsx and SubCategoryListItem.tsx

### Side Effects

N/A

### Checklist

### Testing
Try first without screen-reader and then with TalkBack or VoiceOver

1. Press on any category
2. Check category list items with sub categories (check whether everything is there and nothing is missing)

### Resolved Issues

Fixes: #4360 

---


native/src/components/Categories.tsx
```diff
@@ -43,41 +58,84 @@ const Categories = ({
...
+      accessibilityRole='list'
+      ListHeaderComponent={renderPageHeader}
+      ListFooterComponent={<SpaceForTts $ttsPlayerVisible={ttsPlayerVisible} />}
+      renderSectionHeader={({ section }) => (
+        <CategoryListItem category={section.category} language={language} onItemPress={navigateToCategory} />
+      )}
+      renderItem={({ item, section, index }) => (
+        <>
+          <SubCategoryListItem subCategory={item} onItemPress={navigateToCategory} language={language} />
+          {index < section.data.length - 1 && <Divider />}
```
bahaaTuffaha: This will create another divider on top of `ItemSeparatorComponent`.

bahaaTuffaha (commented):
- There is no subcategory indent anymore:
Image

native/src/components/Categories.tsx
lunars97: oops forgot to remove :grimacing:

native/src/components/CategoryListItem.tsx
```diff
@@ -56,24 +53,7 @@ const CategoryListItem = ({ language, category, subCategories, onItemPress }: Ca
         onPress={() => onItemPress({ path: category.path })}
         accessibilityLanguage={language}
       />
-      {subCategories.length > 0 && (
-        <>
-          <Divider />
-          <List
-            items={subCategories}
-            style={isRTLText(subCategories[0]?.title ?? category.title) ? { marginRight: 56 } : { marginLeft: 56 }}
```
bahaaTuffaha: it was margin not padding for the subCategories so this will affect the dividers too

bahaaTuffaha (commented):
There is a divider at the end of the list after last subCategory:

Image

native/src/components/CategoryListItem.tsx
lunars97: with margins it was looking a bit weird, but I will check it out 👍

lunars97: with margins it breaks, so I used paddings

PR #4384, opened 2026-09-15
4347: Optimize event list

### LINKED ISSUE #4347: Event details lags for couple seconds when going back to events list

steffenkleinle:
### Short Description

Proposal for optimization for deeper lying problems with our theme, dimensions and event grouping that are responsible for a big part of the current performance problems with the event list.
This PR is meant to be used instead (or additionally) to virtualization with react-virtuoso in #4371.

### Proposed Changes

- Optimize `useDimensions` to avoid frequent rerendering and dom queries
- Memoize event list items and groups
- Memoize query param values
- Remove dimensions from theme
- ~~Add lazy loading for icons and [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/content-visibility) to reduce browser rendering work in long lists~~
- Set body background color to default instead of accent 

### Side Effects

- ~~`content-visibility` does not work in older browsers~~

### Checklist

### Testing

- Go to events (try both testumgebung (very many events) and augsburg/münchen)
- Scroll the event list items
- Click on any event
- See event details
- Go back to events list

### Resolved Issues

Fixes: #4347

---


web/src/components/ThemeContainer.tsx
```diff
@@ -184,20 +184,15 @@ const ThemeContainer = ({ children, contentDirection }: ThemeContainerProps): Re
...
-    }
-
+  const theme = useMemo(() => {
+    const toggleTheme = () => setStorageThemeType(themeType === 'light' ? 'contrast' : 'light')
     const theme = createTheme(themeType, contentDirection)
-    document.body.style.backgroundColor = theme.palette.background.accent
-    return { ...theme, toggleTheme }
-  }, [themeType, setStorageThemeType, contentDirection])
+    return { ...theme, toggleTheme, dimensions }
+  }, [themeType, setStorageThemeType, contentDirection, dimensions])
```
bahaaTuffaha, line 191: so it will recompute if the layout changes right?

web/src/components/EventList.tsx
bahaaTuffaha, line 1: It doesn't go back to the same scroll position anymore...

https://github.com/user-attachments/assets/4766e816-ba3b-4739-8128-4a978b6a7ce9

web/src/hooks/useDimensions.ts
bahaaTuffaha, line 1: Claude suggested a cleaner solution could be used here:
using a new react hook called  `useSyncExternalStore()` instead of `useDimensionsSnapshot()`

Add  `measure()` at the end of `subscribe` function and replace `const metrics = useDimensionsSnapshot(getMetrics)` with `const metrics = useSyncExternalStore(subscribe, getMetrics)`

same for scrollMetrics `const scrollMetrics = useSyncExternalStore(subscribe, getScrollMetrics)`

web/src/components/EventList.tsx
lunars97: yes, seems like it appears slightly above the necessary scroll position...

lunars97 (commented):
Thank you a lot for finding a really nice solution, I think with your changes there is no need at all to use `react-virtuoso` at all. The performance has really drastically improved 😻 
One issue from my observation so far:

- When you use the filter to set a `from` date, it fires immediately without giving a chance to set `to` date

bahaaTuffaha:
> * When you use the filter to set a `from` date, it fires immediately without giving a chance to set `to` date

I think `to` should be infinity in this case so this is expected 🤔

lunars97:
> > * When you use the filter to set a `from` date, it fires immediately without giving a chance to set `to` date
> 
> I think `to` should be infinity in this case so this is expected 🤔

Hmm from users' perspective, I would say that I would expect to have a chance to fill in the `to` and it only breaks in this PR

web/src/components/ThemeContainer.tsx
steffenkleinle: It will, thats the cost of providing access to the dimensions in the theme/styled components. Before this PR, the resulting theme was not even memoized though, so this is already an improvement. However, you are right to raise the concern and I think the performance implications still don't justify the convenience of having the dimensions in the theme (and its not really the right place for it), so I removed it completely.

web/src/components/EventList.tsx
steffenkleinle: Caused by virtualization with `content-visibility`, so I removed it again.

web/src/components/EventListItem.tsx
```diff
@@ -88,7 +88,7 @@ const EventListItem = ({
     <StyledListItem dir='auto' disablePadding secondaryAction={recurringDateIcon}>
       <StyledListItemButton component={Link} to={event.path}>
         <ListItemIcon>
-          <Icon src={thumbnailSrc} alt='' />
+          <Icon src={thumbnailSrc} alt='' loading='lazy' decoding='async' />
```
lunars97, line 91: 👍

lunars97 (commented):
Thank you for the refactoring 💚  Hmm after removing `content-visibility` it seems like the performance improvement is not that much noticable 🤔 

- Good thing there is no weird blue color and skip to content button when scrolling 😻

web/src/hooks/useDimensions.ts
```diff
@@ -96,29 +127,88 @@ const getDimensions = (): Dimensions => {
...
+/**
+ * Dimensions including the scroll position. Rerenders the consuming component on every scroll.
+ * Use {@link useDimensions} instead if the scroll position is not needed.
+ */
+export const useScrollDimensions = (): ScrollDimensions => {
+  const dimensions = useDimensions()
+  const scrollMetrics = useDimensionsSnapshot(getScrollMetrics)
+  const { visible } = useContext(TtsContext)
+
+  useEffect(measure, [visible])
```
bahaaTuffaha: `ttsPlayerHeight` stays stale after closing and `Layout` keeps the extra bottom space until the next scroll or resize so I think we can keep the `useEffect(measure, [visible])` and `useContext(TtsContext)` in useDimensions.

web/src/hooks/useDimensions.ts
```diff
@@ -49,20 +45,57 @@ export type Dimensions = {
...
   const footerHeight = document.querySelector('footer')?.offsetHeight ?? 0
   const documentHeight = document.body.offsetHeight
-  const visibleFooterHeight = Math.max(0, height + scrollY + footerHeight - documentHeight)
+  const stickyTop = Math.max(0, document.querySelector('header')?.getBoundingClientRect().bottom ?? 0)
 
-  const headerElement = document.querySelector('header')
-  const stickyTop = Math.max(0, headerElement?.getBoundingClientRect().bottom ?? 0)
+  return {
+    scrollX,
+    scrollY,
```
bahaaTuffaha: ❓  Are these used (scrollX and scrollY)?

bahaaTuffaha (commented):
The flicker after scolling fast is reduced. ✅

web/src/hooks/useDimensions.ts
bahaaTuffaha: I thought about this again ... maybe its better this way cuz its related to scrolling 🤔 

Also how about putting `useScrollDimensions` in a new file?

